PR #40, opened 2020-08-20
Allow logger to be configured via CLI flags

### LINKED ISSUE #39: Allow logger to be configured via CLI
The kubebuilder bootstrap code comes with a default logger configuration for development purposes, but allows flags to be bound, so that users can configure it. We should be using that.

jpkrohling:
Fixes #39


objectiser (approved):
LGTM

PR #52, opened 2020-08-31
Implemented pod injector webhook

### LINKED ISSUE #45: Sidecar injection using admission webhooks
Kubebuilder provides a facility to provision certs and install webhooks. We can use that when doing the sidecar injection, instead of the approach followed by the jaeger-operator.

jpkrohling:
Closes #45.

### DIFF
--- a/internal/podinjector/main.go
+++ b/internal/podinjector/main.go
@@ -1,0 +1,181 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package podinjector
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"net/http"
+	"strings"
+
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/open-telemetry/opentelemetry-operator/api/v1alpha1"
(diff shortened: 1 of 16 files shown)


jpkrohling:
@rubenvp8510, could you please review this one? You are probably more familiar with the Jaeger's injection logic than I am, and your review here would be greatly appreciated!

jpkrohling:
@objectiser: if you have time, your review would also be welcome!

internal/podinjector/main.go
```diff
@@ -0,0 +1,181 @@
...
+	if otelcol.Spec.Mode != v1alpha1.ModeSidecar {
+		return v1alpha1.OpenTelemetryCollector{}, ErrInstanceNotSidecar
+	}
+
+	return otelcol, nil
+}
+
+func (p *podSidecarInjector) selectCollectorInstance(ctx context.Context, ns corev1.Namespace, ann string) (v1alpha1.OpenTelemetryCollector, error) {
+	otelcols := v1alpha1.OpenTelemetryCollectorList{}
+	if err := p.client.List(ctx, &otelcols, client.InNamespace(ns.Name)); err != nil {
```
rubenvp8510, line 164: Is this only looking for collectors in the same namespace? Is not possible to have the collector on another namespace and use true as an annotation value ?

pkg/sidecar/pod_test.go
```diff
@@ -0,0 +1,141 @@
...
+		cfg := config.New(config.WithCollectorImage("some-default-image"))
+
+		// test
+		changed, err := sidecar.Add(cfg, logger, otelcol, pod)
+
+		// verify
+		Expect(err).ToNot(HaveOccurred())
+		Expect(changed.Spec.Containers).To(HaveLen(2))
+		Expect(changed.Spec.Volumes).To(HaveLen(2))
+		Expect(changed.Labels["sidecar.opentelemetry.io/injected"]).To(Equal("some-app.otelcol-sample"))
```
rubenvp8510, line 60: Can we use the label constant?

internal/podinjector/main.go
jpkrohling: For now, yes. @objectiser gave a nice idea about adding a list of namespaces to the CR for Mode=Sidecar, so that we could have all otelcols in the same namespace, each taking care of a set of namespaces. But I still have to figure out RBAC for that, so, it might come only in a future PR.

pkg/sidecar/pod_test.go
jpkrohling: We could, but the actual text is part of the contract: once someone adds this specific string to their deployments, we can't simply change the constant without breaking them (and our test, as a canary).

rubenvp8510: Ok, that makes sense.

PR #91, opened 2020-10-13
Added 'operator-sdk' and 'tools' targets to Makefile

ibraimgm:


Makefile
```diff
@@ -149,12 +149,30 @@ else
 KUSTOMIZE=$(shell which kustomize)
 endif
 
+operator-sdk:
+ifeq (, $(shell which operator-sdk))
+	@{ \
```
jpkrohling, line 154: What if we just download the available binary? 

https://github.com/operator-framework/operator-sdk/releases/download/v1.0.1/operator-sdk-v1.0.1-x86_64-linux-gnu

ibraimgm: We could do that, but if we do it thisway we end up with two little problems:
1. We now depends on `curl` being available.
1.  We need to know the OS/Arch of the user when the makefile is running, to avoid e. g. downloading `linux` binaries to a `darwin` system.

For the first problem, it is not that absurd (IMHO) to just assume `curl` is available.

The second problem is a bit more involved. We can get the OS/Arch pair through `go env` and `grep`, but this will assume `go` is installed. And for OS/Arch combinations that don't have an available binary (e. g. `windows`) can we just fail?

jpkrohling: You know what, it might not be worth pursuing it: for CI, we can have a simple curl in place. In the makefile, we can have an `ensure-operator-sdk-exists` target that is called before targets that require the Operator SDK and would fail + point users to the installation instructions of the operator-sdk. I'm not sure we should really add this complexity to our Makefile...

What do you think?

ibraimgm: I guess 38b4c9ab003139e18e309d346df624924aa1fdaf is as good as it gets. I just changed the "install" logic to an error message pointing to the official website. Now `tools` will install everything and complains if `operator-sdk` doesn't exists, and the same check is repeated on `bundle` (because it uses `operator-sdk`).

Take a look and let me knowifyou need anything else.

jpkrohling (approved):
LGTM

jpkrohling:
@ibraimgm, could you please address the CLA check?

ibraimgm:
@jpkrohling done!

PR #92, opened 2020-10-15
Use Patch instead of Update in the reconciliation

### LINKED ISSUE #54: Change calls to Update() to Patch()
Instead of doing `Update()` calls for the changes we detect, we should attempt to use `Patch()`, to reduce the number of failures caused by stale objects.

From:
```go
		if err := params.Client.Update(ctx, updated); err != nil {
			return fmt.Errorf("failed to apply changes: %w", err)
		}
```

To:
```go
	// at the beginning of the reconciliation function, store a copy of `params.Instance` as `changed`
	patch := client.MergeFrom(&params.Instance)
	if err := params.Client.Patch(ctx, &changed, patch); err != nil {
		return fmt.Errorf("failed to apply changes: %w", err)
	}
```

treasurechristain:
Fixes #54 by changing the reconciliation calls to use Patch instead of Update.

### DIFF
--- a/pkg/collector/reconcile/configmap.go
+++ b/pkg/collector/reconcile/configmap.go
@@ -117,18 +117,11 @@ func expectedConfigMaps(ctx context.Context, params Params, expected []corev1.Co
 			updated.ObjectMeta.Labels[k] = v
 		}
 
-		// editting Update with Patch
 		patch := client.MergeFrom(&params.Instance)
 
-		// if err := params.Client.Patch(ctx, &changed, patch); err != nil {
 		if err := params.Client.Patch(ctx, updated, patch); err != nil {
 			return fmt.Errorf("failed to apply changes: %w", err)
 		}
-
-		
-		// if err := params.Client.Update(ctx, updated); err != nil {
-		// 	return fmt.Errorf("failed to apply changes: %w", err)
-		// }
 
 		params.Log.V(2).Info("applied", "configmap.name", desired.Name, "configmap.namespace", desired.Namespace)
 	}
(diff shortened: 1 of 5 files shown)


jpkrohling:
Looks like the code isn't compiling. Can you sort it out?

pkg/collector/reconcile/configmap.go
```diff
@@ -117,10 +117,19 @@ func expectedConfigMaps(ctx context.Context, params Params, expected []corev1.Co
 			updated.ObjectMeta.Labels[k] = v
 		}
 
-		if err := params.Client.Update(ctx, updated); err != nil {
+		// editting Update with Patch
```
jpkrohling, line 120: You can remove this comment, the code is simple enough to be understood without it.

pkg/collector/reconcile/configmap.go
```diff
@@ -117,10 +117,19 @@ func expectedConfigMaps(ctx context.Context, params Params, expected []corev1.Co
...
+		patch := client.MergeFrom(&params.Instance)
+
+		// if err := params.Client.Patch(ctx, &changed, patch); err != nil {
```
jpkrohling, line 123: You can also remove the commented out code. If we need it later, we can browse the git history.

jpkrohling (commented):
Looks good! Just remove the comments, which aren't really necessary in this case, and take a look at the CLA check that is failing. I'll try it out manually soon and merge if everything works as expected!

treasurechristain:
I just removed the comments I put in those files

jpkrohling (approved):
Looks good to me. Once you sort out the CLA check, it will get merged.

treasurechristain:
I have checked the CLA. Hopefully it's merged now

PR #94, opened 2020-10-20
Allow for spec.Env to be set on the OTEL Colelctor Spec

ekarlso:
* Some of the extensions like Kubernetes Observer requires you to set some
  ENV vars
* Be useful for things like getting API Tokens via Vault etc.


pkg/collector/container.go
```diff
@@ -63,5 +63,6 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
 		Image:        image,
 		VolumeMounts: volumeMounts,
 		Args:         args,
+		Env:          otelcol.Spec.Env,
```
jpkrohling, line 66: Could you add a test that exercises this?

jpkrohling (commented):
Apart from the missing test, looks good to me.

pkg/collector/container.go
ekarlso: Yes I can look at it.

jpkrohling (approved):
LGTM

jpkrohling:
Once you sort the EasyCLA check, it should be merged automatically by mergify. Thanks for the contribution!

PR #117, opened 2020-11-14
Explicitly set Service Port Protocol for Jaeger Receivers

KingJ:
The service port created for each Jaeger receiver does not currently set a Protocol, and therefore defaults to TCP. Some Jaeger receiver protocols (i.e. thrift_binary and thrift_compact) are UDP based rather than TCP, and therefore the operator currently creates service port entries that cannot be used due to the protocol mismatch. For example, the following manifest will currently create service port entries with the wrong protocol for thrift_compact and thrift_binary protocols;

```
kind: OpenTelemetryCollector
metadata:
  name: simplest
spec:
  config: |
    receivers:
      jaeger:
        protocols:
          grpc:
          thrift_compact:
          thrift_binary:
          thrift_http:
    processors:
      queued_retry:
    exporters:
      logging:
        loglevel: debug
    service:
      pipelines:
        traces:
          receivers: [jaeger]
          processors: [queued_retry]
          exporters: [logging]
```

```
Type:              ClusterIP
IP:                10.233.59.118
Port:              jaeger-grpc  14250/TCP
TargetPort:        14250/TCP
Endpoints:         10.233.110.188:14250
Port:              jaeger-thrift-http  14268/TCP
TargetPort:        14268/TCP
Endpoints:         10.233.110.188:14268
Port:              jaeger-thrift-compact  6831/TCP
TargetPort:        6831/TCP
Endpoints:         10.233.110.188:6831
Port:              jaeger-thrift-binary  6832/TCP
TargetPort:        6832/TCP
Endpoints:         10.233.110.188:6832
```

This PR fixes this behavior by extending the protocol struct to also include each Jaeger receiver protocol's protocol and set the service port's protocol when a new Jaeger receiver is created. I have also updated the Jaeger receiver test cases to expect the correct service port protocol for each Jaeger receiver protocol, and they all pass. I have also tested this manually against my own cluster and confirmed that the service port protocol is set for UDP for thrift_binary and thrift_compact both when an endpoint is explicitly defined and when left unset.

I'm still quite new to OpenTelemetry, so i'm open to feedback if i've misunderstood things!

### DIFF
--- a/pkg/collector/parser/receiver_jaeger.go
+++ b/pkg/collector/parser/receiver_jaeger.go
@@ -46,29 +46,29 @@ func (j *JaegerReceiverParser) Ports() ([]corev1.ServicePort, error) {
 	ports := []corev1.ServicePort{}
 
 	for _, protocol := range []struct {
-		name        string
-		defaultPort int32
-		protocol    corev1.Protocol
+		name              string
+		defaultPort       int32
+		transportProtocol corev1.Protocol
 	}{
 		{
-			name:        "grpc",
-			defaultPort: defaultGRPCPort,
-			protocol:    corev1.ProtocolTCP,
+			name:              "grpc",
+			defaultPort:       defaultGRPCPort,
+			transportProtocol: corev1.ProtocolTCP,
 		},
 		{
-			name:        "thrift_http",
-			defaultPort: defaultThriftHTTPPort,
-			protocol:    corev1.ProtocolTCP,
+			name:              "thrift_http",
+			defaultPort:       defaultThriftHTTPPort,
+			transportProtocol: corev1.ProtocolTCP,
 		},
 		{
-			name:        "thrift_compact",
-			defaultPort: defaultThriftCompactPort,
-			protocol:    corev1.ProtocolUDP,
+			name:              "thrift_compact",
+			defaultPort:       defaultThriftCompactPort,
+			transportProtocol: corev1.ProtocolUDP,
 		},
 		{
-			name:        "thrift_binary",
-			defaultPort: defaultThriftBinaryPort,
-			protocol:    corev1.ProtocolUDP,
+			name:              "thrift_binary",
+			defaultPort:       defaultThriftBinaryPort,
+			transportProtocol: corev1.ProtocolUDP,
 		},
 	} {
 		// do we have the protocol specified at all?
@@ -92,8 +92,8 @@ func (j *JaegerReceiverParser) Ports() ([]corev1.ServicePort, error) {
 				}
 			}
 
-			// set the appropriate TCP/UDP protocol for this kind of receiver protocol
-			protocolPort.Protocol = protocol.protocol
+			// set the appropriate transport protocol (i.e. TCP/UDP) for this kind of receiver protocol
+			protocolPort.Protocol = protocol.transportProtocol
 
 			// at this point, we *have* a port specified, add it to the list of ports
 			ports = append(ports, *protocolPort)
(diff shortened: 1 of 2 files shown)


KingJ:
I have now signed the CLA, but I don't know why the lint timed out and failed unfortunately.

jpkrohling:
Thanks for the PR! I'm unable to review it today, but I'll take a look soon.

pkg/collector/parser/receiver_jaeger.go
```diff
@@ -82,9 +87,13 @@ func (j *JaegerReceiverParser) Ports() ([]corev1.ServicePort, error) {
...
 				protocolPort = &corev1.ServicePort{
-					Name: portName(nameWithProtocol, protocol.defaultPort),
-					Port: protocol.defaultPort,
+					Name:     portName(nameWithProtocol, protocol.defaultPort),
+					Port:     protocol.defaultPort,
+					Protocol: protocol.protocol,
 				}
+			} else {
+				// port from configuration block has been used, but protocol needs to be set
+				protocolPort.Protocol = protocol.protocol
```
jpkrohling, line 96: You could simplify this code by setting the protocol after this conditional. The current code would then be the same, but you'd just set the protocol before appending it to the list of ports.

jpkrohling (commented):
The changes look good and there's only one thing that could be improved. This also needs a rebase.

pkg/collector/parser/receiver_jaeger.go
KingJ: Thanks for spotting this @jpkrohling! I've broken this out of the conditional so that the port's protocol is now consistently set regardless of if a port is specified in the supplied configuration or not.

KingJ:
Thanks for the review! I've resolved the issue you found and re-based with the latest changes. I've re-run the tests locally and re-confirmed that it also works as expected on my cluster.

pkg/collector/parser/receiver_jaeger.go
```diff
@@ -87,6 +92,9 @@ func (j *JaegerReceiverParser) Ports() ([]corev1.ServicePort, error) {
 				}
 			}
 
+			// set the appropriate TCP/UDP protocol for this kind of receiver protocol
+			protocolPort.Protocol = protocol.protocol
```
jpkrohling, line 96: I'm slightly annoyed that we have the same term (protocol) with two meanings here as part of the same struct: the data protocol (gRPC/protobuf, Thrift-compact, Thrift-binary, ...) and the transport protocol (TCP/UDP). Your comment above saves this ambiguity from confusion. Given that I don't have a better proposal, I think it's OK to leave it as is, but if you have another idea, I'm open to consider it.

jpkrohling (commented):
LGTM. I'll wait for your OK to merge, as you might have a better name for the `protocol` property in the `protocol` struct.

pkg/collector/parser/receiver_jaeger.go
KingJ: I agree it's confusing, and preferably i'd like to make it unambiguous by default to avoid the need for a clarifying comment. Referring to it as the transport protocol in the struct would be an improvement, although `protocolPort.Protocol` will remain somewhat ambiguous regardless. It's still an overall improvement though, so i'll get another commit prepared.

jpkrohling (approved):
LGTM. Thank you for your patience and for your contribution!

KingJ:
> LGTM. Thank you for your patience and for your contribution!

Awesome! Thank you for your review and constructive feedback too!

PR #122, opened 2020-11-16
Convert tests to plain go tests

jpkrohling:


jpkrohling:
@kearls, @objectiser, could one of you please review this? I'm mostly AFK today, but wanted to open this PR anyway as this might take a while to review.

Most of the changes are quite mechanic and there shouldn't be any changes in behavior. There was one test removed, as it was exactly the same as another existing one.

kevinearls (approved):
Given the length of this PR I did not do an in-depth line by line review, but did a quick pass on almost everything and it looks good to me.

Since you're already using stretchr it might be worth using their suites package for these tests two.  I might be able to work on that if you want to open an issue for it.

jpkrohling:
> Since you're already using stretchr it might be worth using their suites package for these tests two

I confess that I don't know stretchr as much as I should, so, I'm not quite sure I understand the benefits. Do you mean to group the envtests ?

kevinearls:
@jpkrohling The suites package gives you things you'd typically have in an xUnit type framework, notably beforeSuite, beforeTest, afterTest, and afterSuite functions.  If I remember correctly for the Jaeger Operator E2E tests it also made it easier to do reporting.

PR #141, opened 2020-12-14
Add login step to publishing of images

jpkrohling:


jpkrohling:
Now it works:

image

objectiser:
@jpkrohling  Just wondering whether we actually need the sha tag at the moment?

Would it also be worth changing master -> main before 1.x?

jpkrohling:
We don't need the SHA. Using master here comes from the branch name, but it might indeed be a good idea to change it soon to avoid breaking users.

jpkrohling:
#143 to track the branch rename

PR #171, opened 2021-01-22
Fixed testing image from being used in non-test artifacts (fixes #170)

### LINKED ISSUE #170: Latest release (0.18.0) points to wrong image (local/opentelemetry-operator:e2e)
The latest 0.18.0 release is pointing to the wrong image. Expectation is that it points to `quay.io/opentelemetry/opentelemetry-operator`. 

```
Events:
  Type     Reason     Age                From               Message
  ----     ------     ----               ----               -------
...
  Normal   Pulling    19s (x3 over 58s)  kubelet            Pulling image "local/opentelemetry-operator:e2e"
  Warning  Failed     18s (x3 over 57s)  kubelet            Failed to pull image "local/opentelemetry-operator:e2e": rpc error: code = Unknown desc = Error response from daemon: pull access denied for local/opentelemetry-operator, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
  Warning  Failed     18s (x3 over 57s)  kubelet            Error: ErrImagePull
  Normal   BackOff    6s (x3 over 56s)   kubelet            Back-off pulling image "local/opentelemetry-operator:e2e"
  Warning  Failed     6s (x3 over 56s)   kubelet            Error: ImagePullBackOff
```

gramidt:
Signed-off-by: Granville Schmidt <1246157+gramidt@users.noreply.github.com>

Resolves Issue: #170

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,7 +102,7 @@ prepare-e2e: kuttl set-test-image-vars set-image-controller container
 	$(KUSTOMIZE) build config/crd -o tests/_build/crds/
 
 set-test-image-vars:
-	IMG=local/opentelemetry-operator:e2e
+	$(eval IMG=local/opentelemetry-operator:e2e)
 
 # Build the container image, used only for local dev purposes
 container:
--- a/Makefile
+++ b/Makefile
@@ -102,7 +102,7 @@ prepare-e2e: kuttl set-test-image-vars set-image-controller container
 	$(KUSTOMIZE) build config/crd -o tests/_build/crds/
 
 set-test-image-vars:
-IMG=local/opentelemetry-operator:e2e
+	IMG=local/opentelemetry-operator:e2e
 
 # Build the container image, used only for local dev purposes
 container:


gramidt:
Looking into e2e failure now.

Makefile
```diff
@@ -102,7 +102,7 @@ prepare-e2e: kuttl set-test-image-vars set-image-controller container
 	$(KUSTOMIZE) build config/crd -o tests/_build/crds/
 
 set-test-image-vars:
-	IMG=local/opentelemetry-operator:e2e
+	$(eval IMG=local/opentelemetry-operator:e2e)
```
gramidt, line 105: Eval is typically not a good solution within Makefiles, but given that this is only used for setting up tests, it should be an okay solution for the time being.

gramidt:
The latest error from e2e:

```
 logger.go:42: 00:34:58 |  | Error from server: error when creating "./tests/_build/manifests/01-opentelemetry-operator.yaml": conversion webhook for cert-manager.io/v1alpha2, Kind=Issuer failed: Post "https://cert-manager-webhook.cert-manager.svc:443/convert?timeout=30s": x509: certificate signed by unknown authority

```

@jpkrohling - I may be able to also look into this issue either tonight or tomorrow.

gramidt:
Appears the very issue didn't appear after re-running the e2e tests. Maybe we should open an issue and look into it further.

Makefile
jpkrohling: Could you try running the release workflow locally without and with this change? I don't really understand this fix (vs. the current behavior), but there are a few things about Makefiles that I don't understand anyway :-)

gramidt: @jpkrohling - Confirmed. Below are the two cases to be aware of:

1. Without any changes to the make file, the release process uses the "local/opentelemetry-operator:e2e" image.
Screen Shot 2021-01-23 at 1 33 33 PM

2. With `IMG=local/opentelemetry-operator:e2e` indented under `set-test-image-vars`, the release artifacts correctly use the quay.io image; however, now the e2e workflow is broken https://github.com/open-telemetry/opentelemetry-operator/blob/e23224ee92ac86d9b31d70dc580a25c1883d2d9d/.github/workflows/e2e.yaml#L21 as `IMG` is still set to the global quay.io image, since the assignment of `IMG` within `set-test-image-vars` does not change the value of `IMG` for the reset of the target prerequisites (set-image-controller container).
Screen Shot 2021-01-23 at 1 34 05 PM
Screen Shot 2021-01-23 at 1 35 57 PM


With this PR, both of the issues identified above are resolved. However, we may want to revisit portions of this Makefile in the future.

jpkrohling: I'll merge this and do a dry-run for 0.18.1.

jpkrohling (approved):

PR #210, opened 2021-02-19
fix: collector selection should not fail if there is a single sidecar

### LINKED ISSUE #200: Injection when annotation=true fails when if there are other instances but only one in sidecar mode.
Side injection when `sidecar.opentelemetry.io/inject: true` doesn't work when there are other  opentelemetry collectors but only one in _sidecar_ mode.

Failed with the following message:

`Failed to select an OpenTelemetry Collector instance for this pod's sidecar","namespace":"","name":"","error":"multiple OpenTelemetry Collector instances available, cannot determine which one to select`

It is true that there are other opentelemetry collector instances, but should not fail if there is only one instance with `mode:sidecar`.

vbehar:
fixes #200

if there are multiple collectors, but a single sidecar, the collector selection should not fail.
this fix ensures we filter the list of collectors first.

Signed-off-by: Vincent Behar <v.behar@free.fr>


jpkrohling (approved):
Thank you for this fix!

Issue #211, opened 2021-02-24
Failed reconciliation on OpenShift


rubenvp8510:
I tried to use the operator on OpenShift and when I created a CR the operator reconciliation fails with the following message:

`{"level":"error","ts":1614142166.5476513,"logger":"controller-runtime.manager.controller.opentelemetrycollector","msg":"Reconciler error","reconciler group":"opentelemetry.io","reconciler kind":"OpenTelemetryCollector","name":"opentelemetrycollector-sample","namespace":"default","error":"failed to reconcile the expected configmaps: failed to create: configmaps \"opentelemetrycollector-sample-collector\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/pkg/mod/github.com/go-logr/zapr@v0.2.0/zapr.go:132\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.8.2/pkg/internal/controller/controller.go:301\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.8.2/pkg/internal/controller/controller.go:252\nsigs.k8...`

This error only happens on OpenShift, plain kubernetes is fine.

PR #212, opened 2021-02-24
Add permissions for opentelemetry finalizer resource

rubenvp8510:
Still need to test this changes on an OCP cluster, but this should be enough for fix https://github.com/open-telemetry/opentelemetry-operator/issues/211

See other similar issues here:
https://github.com/operator-framework/operator-sdk/issues/1736
https://github.com/spotahome/redis-operator/issues/98

Signed-off-by: Ruben Vargas <ruben.vp8510@gmail.com>

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -97,6 +97,14 @@ rules:
 - apiGroups:
   - opentelemetry.io
   resources:
+  - opentelemetrycollectors/finalizers
+  verbs:
+  - get
+  - patch
+  - update
+- apiGroups:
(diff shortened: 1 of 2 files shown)


jpkrohling:
Don't you need to generate the manifests again? I would have expected at least the following file to be changed: config/rbac/role.yaml

rubenvp8510:
> Don't you need to generate the manifests again? I would have expected at least the following file to be changed: config/rbac/role.yaml

Yes, I forgot to re-generate the manifest

jpkrohling (approved):

PR #218, opened 2021-03-02
Upgrade controller-gen to v0.4.0 and change CRDs apiVersion to v1

bhiravabhatla:

Change Description

- Upgrade controller-gen version to v0.4.0. It by default creates CRDs using apiVersion apiextensions.k8s.io/v1
- Fix webhook patches to work with apiextensions.k8s.io/v1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,7 +127,7 @@ ifeq (, $(shell which controller-gen))
 	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
 	cd $$CONTROLLER_GEN_TMP_DIR ;\
 	go mod init tmp ;\
-	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.3.0 ;\
+	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.0 ;\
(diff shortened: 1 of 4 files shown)


bhiravabhatla:
Need to move webhook manifests to v1 as well. For that would need to confirm, if I can mark sideAffects as None. SideAffects is a mandatory field in v1 apiVersion.

https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#side-effects

jpkrohling:
There are no side-effects, ie, out-of-band changes executed as part of any of our webhooks. The only webhook that changes anything is the pod injector, but that change isn't out-of-band.

jpkrohling (approved):

bhiravabhatla:
@jpkrohling  - Just curious what mutations would be come under out-of-band changes. I have gone through the docs, dint get clear understanding.

jpkrohling:
If we would auto-provision some resource based on the CR, this would be an out-of-band change.

bhiravabhatla:
@jpkrohling Would need to modify the webhook manifests as well. They are in v1beta1 now. Would raise another PR.

PR #248, opened 2021-04-15
feat: allow resource configuration on collector spec

### LINKED ISSUE #247: Add support for configuring the K8s resources for injected collectors
It would be nice if K8s resources of the injected collectors (for sidecar mode) could be configured via the operator CRD. This is an important requirement when using horizontal pod autoscalers later on. Containers in a pod that have no resource requests can cause issues with HPA.

jnodorp-jaconi:
Fixes #247


jpkrohling (approved):
Thanks!

PR #250, opened 2021-04-19
[Part 1]Add unit tests for reconcile methods

### LINKED ISSUE #35: Tests for reconcilers
The reconcilers (`pkg/collector/reconcile`) are currently untested. This issue is to track building the unit tests for them.

bhiravabhatla:

Resolves #35 
* Add unit tests for configmap, daemonset, deployment reconcilers
* Fix Patch calls for  configmap, daemonset, deployment reconcilers

Raised https://github.com/open-telemetry/opentelemetry-operator/pull/261 for service and service account reconcilers

### DIFF
--- a/pkg/collector/reconcile/configmap.go
+++ b/pkg/collector/reconcile/configmap.go
@@ -120,7 +120,7 @@ func expectedConfigMaps(ctx context.Context, params Params, expected []corev1.Co
 			updated.ObjectMeta.Labels[k] = v
 		}
 
-		patch := client.MergeFrom(&params.Instance)
+		patch := client.MergeFrom(existing)
 
 		if err := params.Client.Patch(ctx, updated, patch); err != nil {
 			return fmt.Errorf("failed to apply changes: %w", err)
--- a/pkg/collector/reconcile/configmap_test.go
+++ b/pkg/collector/reconcile/configmap_test.go
@@ -19,19 +19,11 @@ import (
 	"fmt"
 	"testing"
 
-	"github.com/open-telemetry/opentelemetry-operator/api/v1alpha1"
-	"github.com/open-telemetry/opentelemetry-operator/internal/config"
-
 	"github.com/stretchr/testify/assert"
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/tools/record"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"k8s.io/apimachinery/pkg/types"
 )
-
-var logger = logf.Log.WithName("unit-tests")
 
 func TestDesiredConfigMap(t *testing.T) {
 	t.Run("should return expected config map", func(t *testing.T) {
@@ -43,18 +35,23 @@ func TestDesiredConfigMap(t *testing.T) {
 }
 
 func TestExpectedConfigMap(t *testing.T) {
+
+	cm := configMap("test-collector")
+	deletecm := configMap("test")
+
 	t.Run("should create config map", func(t *testing.T) {
-		err := expectedConfigMaps(context.Background(), params(), []v1.ConfigMap{configMap("test-collector")}, true)
+		err := expectedConfigMaps(context.Background(), params(), []v1.ConfigMap{cm}, true)
 		assert.NoError(t, err)
 
-		actual, err := getCM("test-collector")
+		exists, err := populateObjectIfExists(t, &v1.ConfigMap{}, types.NamespacedName{Namespace: "default", Name: "test-collector"})
 
 		assert.NoError(t, err)
-		assert.NotNil(t, actual)
+		assert.True(t, exists)
 	})
 
 	t.Run("should update config map", func(t *testing.T) {
-		createCMIfNotExists(t, "test-collector")
+
+		createObjectIfNotExists(t, "test-collector", &cm)
 
 		_ = expectedConfigMaps(context.Background(), params(), []v1.ConfigMap{configMap("test-collector")}, true)
 		//assert.NoError(t, err)
@@ -66,14 +63,13 @@ func TestExpectedConfigMap(t *testing.T) {
 	})
 
 	t.Run("should delete config map", func(t *testing.T) {
-		createCMIfNotExists(t, "test")
-
+		createObjectIfNotExists(t, "test", &deletecm)
 		err := deleteConfigMaps(context.Background(), params(), []v1.ConfigMap{configMap("test-collector")})
 		assert.NoError(t, err)
 
-		_, err = getCM("test")
+		exists, err := populateObjectIfExists(t, &v1.ConfigMap{}, types.NamespacedName{Namespace: "default", Name: "test"})
 
-		assert.True(t, errors.IsNotFound(err))
+		assert.False(t, exists)
 	})
 }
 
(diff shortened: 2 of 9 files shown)


bhiravabhatla:
@jpkrohling  - I am unable to test Patch call (updating existing configmap) with envtest.

`go
		err = expectedConfigMaps(context.Background(), params(), []v1.ConfigMap{configMap()}, true)
		assert.NoError(t, err)
`

While running expectedConfigMaps with existing configmap, I am seeing below error returned. 

`
failed to apply changes: the name of the object (test-collector based on URL) was undeterminable: name must be provided
`
While running on debug - i could see the control reaching the Patch statement - 

`go
if err := params.Client.Patch(ctx, updated, patch); err != nil {
			return fmt.Errorf("failed to apply changes: %w", err)
		}
`
And Patch is returning above error. Any clues what might be causing this. Current PR has this particular test commented.

alolita:
@humivo @IrisTuntun can you take a look at this PR too for the Operator enhancements we're working on?

bhiravabhatla:
I would need help with above blocker. If this is resolved, I can add tests for all the other reconcilers similarly.

bhiravabhatla:
@jpkrohling  - I guess there is a bug in Patch logic - 

https://github.com/open-telemetry/opentelemetry-operator/blob/db40230d99040201ba7b85a64700788392e96b62/pkg/collector/reconcile/configmap.go#L123

We are generating patch from Instance, (not sure why). If I generate the patch from existing configmap [ which should be the case, as we are pacthing configmap - not openTelemetryCollector CR ] , like below - 

patch := client.MergeFrom(existing)

Patch Tests are  passing. 

Same is the case for all the patch calls ( configmap, daemonset, deployment , etc). Kindly confirm

bhiravabhatla:
> I ll update this thread after testing the operator using above fix

Works well. I could see deployment being patched if I manually edit the deployment. I am pushing the above change with this PR if its ok. @jpkrohling  - Thoughts?

jpkrohling:
Sorry for taking so long to look into this, I'll review this today or, at the latest, tomorrow.

IrisTuntun:
Hi, I'm working on adding StatefulSet to reconcile and have a test added in this [PR](https://github.com/open-o11y/opentelemetry-operator/pull/1). The test works without changing the `MergeFrom` logic. I was wondering if my test strategy makes sense here.

pkg/collector/reconcile/configmap.go
```diff
@@ -120,7 +120,7 @@ func expectedConfigMaps(ctx context.Context, params Params, expected []corev1.Co
 			updated.ObjectMeta.Labels[k] = v
 		}
 
-		patch := client.MergeFrom(&params.Instance)
+		patch := client.MergeFrom(existing)
```
jpkrohling, line 123: This was indeed a bug! Looks like the same problem exists for the Service and ServiceAccount.

pkg/collector/reconcile/configmap_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestDesiredConfigMap(t *testing.T) {
+	t.Run("should return expected config map", func(t *testing.T) {
+		expected := configMap("test-collector")
```
jpkrohling, line 30: I would skip the local configMap, and make explicit assertions about certain aspects of the returned config map, like checking the returned labels and the name, and possibly the config.

pkg/collector/reconcile/configmap_test.go
```diff
@@ -0,0 +1,95 @@
...
+		actual := desiredConfigMap(context.Background(), params())
+		assert.Equal(t, expected, actual)
+	})
+
+}
+
+func TestExpectedConfigMap(t *testing.T) {
+
+	cm := configMap("test-collector")
```
jpkrohling, line 39: You can call the desiredConfigMap here, as you are testing its correctness in another test.

pkg/collector/reconcile/daemonset_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+		actual := v1.DaemonSet{}
+		exists, _ := populateObjectIfExists(t, &actual, types.NamespacedName{Namespace: "default", Name: "dummy"})
+
+		assert.False(t, exists)
+
+	})
+}
+
+func daemonset(name string) v1.DaemonSet {
```
jpkrohling, line 75: Same as the configmap: if you have a test for the desiredDaemonSet, this one here could probably be removed in favor of the official one.

pkg/collector/reconcile/deployment_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+		actual := v1.Deployment{}
+		exists, _ := populateObjectIfExists(t, &actual, types.NamespacedName{Namespace: "default", Name: "dummy"})
+
+		assert.False(t, exists)
+
+	})
+}
+
+func deployment(name string) v1.Deployment {
```
jpkrohling, line 76: Ditto :-)

jpkrohling (commented):
It's looking great! Would you like to get this one merged first, before working on the next tests? This way, we don't get huge PRs in the queue.

bhiravabhatla:
> Would you like to get this one merged first, before working on the next tests?

Makes sense. I am half way through with service tests - I ll raise another PR for it.

pkg/collector/reconcile/configmap.go
bhiravabhatla: Yes it does. I ll raise seperate PR for it.

pkg/collector/reconcile/daemonset_test.go
bhiravabhatla: Yep makes sense.

pkg/collector/reconcile/deployment_test.go
bhiravabhatla: Will do

bhiravabhatla: For delete test cases - I cant use desiredDaemonSet/desiredDeployment - As I would need to a resource with same labels as the one in Params but with different name. 

i.e.

Lables should be 

Labels: map[string]string{
				"app.kubernetes.io/instance":   fmt.Sprintf("%s.%s", params.Instance.Namespace, params().Instance.Name),
			},
			
but name should be different than params.Instance.Name. Its not possible using  desiredDaemonSet/desiredDeployment  functions. So I would have to retain the daemonset and deployment helper functions.

pkg/collector/reconcile/daemonset_test.go
bhiravabhatla: Same as below

pkg/collector/reconcile/configmap_test.go
bhiravabhatla: Done

pkg/collector/reconcile/configmap_test.go
bhiravabhatla: Done

pkg/collector/reconcile/deployment_test.go
jpkrohling: I think all you need for those cases is a deployment/daemonset with one or two labels (managed-by, I believe). This is what we have in the jaeger-operator:

https://github.com/jaegertracing/jaeger-operator/blob/4693d119aabb550a7a651315cbfce1833bb786fc/pkg/controller/jaeger/daemonset_test.go#L104-L114

The good thing about doing it like this is that we see exactly what has influence on the decision to remove the CR.

jpkrohling:
Let me know once this is ready for another review.

pkg/collector/reconcile/deployment_test.go
bhiravabhatla: In our case - we are actually creating the object. So would need the entire spec, only namespacedName and labels wont suffice. Basically have to move everything in daemonset function to actual test. Will do the same then

bhiravabhatla:
> Let me know once this is ready for another review.

Done. Also added one more test case each for daemonset and deployment to make sure we dont delete resources which are not owned by operator. Kindly review @jpkrohling

bhiravabhatla:
> Let me know once this is ready for another review.

 Done. Also added one more test case each for daemonset and deployment to make sure we dont delete resources which are not owned by operator. Kindly review @jpkrohling

bhiravabhatla:
> > Would you like to get this one merged first, before working on the next tests?
> 
> Makes sense. I am half way through with service tests - I ll raise another PR for it.

Raised second PR for this. Could you also take a look at - https://github.com/open-telemetry/opentelemetry-operator/pull/261 @jpkrohling

jpkrohling (approved):

jpkrohling:
CodeQL seems stuck. Would you be able to do a "git commit --amend" and push it again?

PR #261, opened 2021-05-02
[Part 2]Add unit tests for reconcile methods 

### LINKED ISSUE #35: Tests for reconcilers

bhiravabhatla:
Second PR for the same issue. As the first one - https://github.com/open-telemetry/opentelemetry-operator/pull/250, was getting big, have split the work into two PRs

This PR covers -
* Add unit tests for service and service account reconcile functions
* Fix patch call for service and service account.

Fixes #35 

### DIFF
--- a/pkg/collector/reconcile/service_test.go
+++ b/pkg/collector/reconcile/service_test.go
@@ -95,7 +95,7 @@ func TestFilterPort(t *testing.T) {
 }
 
 func TestDesiredService(t *testing.T) {
-	t.Run("should return nil service for unknown reciever and protocol", func(t *testing.T) {
+	t.Run("should return nil service for unknown receiver and protocol", func(t *testing.T) {
 		params := Params{
 			Config: config.Config{},
 			Client: k8sClient,
--- a/pkg/collector/reconcile/service_test.go
+++ b/pkg/collector/reconcile/service_test.go
@@ -1,0 +1,231 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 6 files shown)


bhiravabhatla:
Resolved conflicts. @jpkrohling  Kindly help review.

pkg/collector/reconcile/service_test.go
```diff
@@ -0,0 +1,231 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/open-telemetry/opentelemetry-operator/api/v1alpha1"
+	"github.com/open-telemetry/opentelemetry-operator/internal/config"
+	"github.com/open-telemetry/opentelemetry-operator/pkg/collector"
+)
+
+func TestExtractPortNumbersAndNames(t *testing.T) {
+	t.Run("should return extracted port names and numbers", func(t *testing.T) {
```
jpkrohling, line 32: Sorry, should have caught this before on the other PR as well, but why are you wrapping the test in a test?

jpkrohling (commented):
Looks good, just wondering why you are wrapping some tests in other tests when the func has only one test.

pkg/collector/reconcile/service_test.go
bhiravabhatla: Just give a proper name to the test - t.Run gives a way to describe the test properly using name field and to group few tests which belong together.

I understand that for function with only  one test it might look odd. I can remove it

jpkrohling: That's fine, not sure I quite like the fact that we now have two patterns in the test suite, but we can make this the standard for the reconciliation tests.

jpkrohling (approved):

PR #265, opened 2021-05-06
Add StatefulSet support in OpenTelemetry Operator

IrisTuntun:
Add StatefulSet support in Operator so that the Operator can start collectors in StatefulSet mode.  Also added unit tests and e2e tests. Fixes  [wg-prometheus#31](https://github.com/open-telemetry/wg-prometheus/issues/31) and [wg-prometheus#47](https://github.com/open-telemetry/wg-prometheus/issues/47).


api/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -51,6 +51,12 @@ type OpenTelemetryCollectorSpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
 	ServiceAccount string `json:"serviceAccount,omitempty"`
 
+	// VolumeClaimTemplates will provide stable storage using PersistentVolumes.
```
jpkrohling, line 54:
```suggestion
	// VolumeClaimTemplates will provide stable storage using PersistentVolumes. Only available when the mode=statefulset.
```

pkg/collector/reconcile/statefulset.go
```diff
@@ -0,0 +1,135 @@
...
+		// it exists already, merge the two if the end result isn't identical to the existing one
+		updated := existing.DeepCopy()
+		if updated.Annotations == nil {
+			updated.Annotations = map[string]string{}
+		}
+		if updated.Labels == nil {
+			updated.Labels = map[string]string{}
+		}
+
+		updated.Spec = desired.Spec
```
jpkrohling, line 83: Could you double-check that you can copy the entire Spec here? Some resources have properties that can't be changed, or that we don't want to change.

pkg/collector/volumeclaim.go
```diff
@@ -0,0 +1,52 @@
...
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/open-telemetry/opentelemetry-operator/api/v1alpha1"
+	"github.com/open-telemetry/opentelemetry-operator/internal/config"
+)
+
+// VolumeClaimTemplates builds the volumeClaimTemplates for the given instance,
+// including the config map volume mount.
+func VolumeClaimTemplates(cfg config.Config, otelcol v1alpha1.OpenTelemetryCollector) []corev1.PersistentVolumeClaim {
```
jpkrohling, line 29: Question: given that this is only used when the mode=statefulset, should this function here be checking that condition and return an empty slice when that condition is not satisfied?

tests/e2e/smoke-statefulset/00-assert.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: stateful-collector
+status:
+  replicas: 1
```
jpkrohling, line 6: Do Statefulsets have `readyReplicas` as well?

jpkrohling (commented):
This looks great! The design doc was also very nice, thank you for that. I left a few comments, but nothing major. I'd be more than happy to review your future docs in the future before they become PRs. As I'm not part of the Prometheus WG, feel free to ping me directly.

jpkrohling:
One extra comment: we have a validating webhook that can be used to ensure that the VolumeClaimTemplate is only used when the mode is statefulset:

https://github.com/open-telemetry/opentelemetry-operator/blob/b6ac997b1ba13d756f8679870d49ee497b699c6f/api/v1alpha1/opentelemetrycollector_webhook.go#L57-L67

pkg/collector/reconcile/statefulset.go
IrisTuntun: I have double checked [StatefulSetSpec](https://github.com/kubernetes/api/blob/3a2d6b5bb7f84daa2125131865eb81b151448d5b/apps/v1beta1/types.go#L162), and didn't see any fields that we do not want to change. Any ideas?

api/v1alpha1/opentelemetrycollector_types.go
humivo: Applied in latest commit: "Fix volumeclaim function and small bugs"

pkg/collector/volumeclaim.go
humivo: Applied this check to the function in commit: "Fix volumeclaim function and small bugs"

tests/e2e/smoke-statefulset/00-assert.yaml
IrisTuntun: `readyReplicas` assertions added

humivo:
@jpkrohling Also added the logic for ensuring volumeClaimTemplates only used for statefulset modes in the validation webhook

jpkrohling (approved):
Thank you for this PR!

PR #299, opened 2021-05-27
Copy desired Service ports when reconciling

### LINKED ISSUE #256: expectedServices function not copying desired spec
https://github.com/open-telemetry/opentelemetry-operator/blob/db40230d99040201ba7b85a64700788392e96b62/pkg/collector/reconcile/service.go#L182-L199

I guess, we should be copying desired.Spec to updated.Spec if I am not wrong. - 

```go
  updated.Spec = desired.Spec
```
If we dont do that service spec wont be updated if some one adds a new receiver in config or explicit port in Spec.Ports of collector.

### LINKED ISSUE #295: OpenCensus port is not exposed by the Collector Service
Hello,

I am deploying an OpenTelemetry Collector with an OpenCensus receiver.
According to [the comments on the `ports` parameter](https://github.com/open-telemetry/opentelemetry-operator/blob/36d59730af234c2bfc4f909fa2ff8a4659b90c86/api/v1alpha1/opentelemetrycollector_types.go#L72-L78), the operator should be able to detect well-known protocols in the `config` and expose their port. However, after deploying it, the collector initializes the OpenCensus receiver, but the service is not updated to expose the OpenCensus port.

It should be noted that the collector was first declared and deployed with only an `oltp` receiver, and I later updated it to add the OpenCensus collector. Maybe the operator simply didn't update the service.

thib92:
Copy services from the desired service when reconciling.

For now it only copies the `ports` attribute. We might need to copy additional fields as well. Let me know!

Fixes #256 
Fixes #295

### DIFF
--- a/pkg/collector/reconcile/service.go
+++ b/pkg/collector/reconcile/service.go
@@ -194,6 +194,7 @@ func expectedServices(ctx context.Context, params Params, expected []corev1.Serv
 		for k, v := range desired.ObjectMeta.Labels {
 			updated.ObjectMeta.Labels[k] = v
 		}
+		updated.Spec.Ports = desired.Spec.Ports
 
 		patch := client.MergeFrom(existing)
 
--- a/pkg/collector/reconcile/service_test.go
+++ b/pkg/collector/reconcile/service_test.go
@@ -22,6 +22,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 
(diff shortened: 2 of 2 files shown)


jpkrohling (approved):

jpkrohling:
Looks good to me: once you confirm that you also don't see any other fields that should be copied, I'll merge.

jpkrohling (approved):

thib92:
The [`ServiceSpec`](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/#ServiceSpec) does have its fair share of options, most of them I'm not too familiar with.

However, I think that using the OpenTelemetry Operator, only the `ports` field can be messed with (at least at the moment), so I think it's safe to keep it like that.

jpkrohling:
Thank you for this PR!

Issue #186, opened 2021-01-31
Add support for tolerations?


markfink-splunk:
In daemonset mode, it installs to worker nodes only, not master, which is normal and we can usually workaround with a toleration.  But I can't figure out how to apply a toleration to this operator and support for it may need to be added.  The use-case is for the kubelet metrics receiver that is available with otel-collector-contrib.  I'd like to get kubelet metrics on the master node.  I'm creating a separate deployment just for the master now.  It would be convenient to roll support for this into the operator.

jpkrohling:
Sounds like a valid use case. Would you like to send a PR adding support to the operator?

PR #302, opened 2021-06-01
Tolerations support in otelcol CRD

VineethReddy02:
add support for tolerations in otelcol CRD, This helps in running Otel collectors on tainted nodes for both deployments and daemon sets.

fixes: #186


api/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -87,6 +87,11 @@ type OpenTelemetryCollectorSpec struct {
 	// +optional
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
 	Resources v1.ResourceRequirements `json:"resources,omitempty"`
+
+	// Toleration to schedule OpenTelemetry Collector pods.
```
jpkrohling, line 91: Please, document under which modes this is relevant: `daemonset` and `deployment`. How about statefulsets? Similarly, you might need to add some validation, so that we don't accept tolerations for unsupported modes (sidecar comes to my mind, for instance).

jpkrohling (commented):
Looks good, just need some input validation and a more complete godoc for the new field.

jpkrohling:
Tests are missing as well.

api/v1alpha1/opentelemetrycollector_types.go
VineethReddy02: This is relevant in `daemonsets`, `deployments` & `statefulsets` (I didn't know this existed). I thought to add validation but noticed we do not have validation in place for example `replicas` are relevant only to `deployments` & `statefulsets` but not `daemonsets` and `sidecars`. Currently, the CRD accepts replicas for unsupported modes too. 

Could you point me to where exactly do you think CRD validation makes sense in the operator flow?

jpkrohling: Good catch about the replica size. Do you mind creating an issue for it? The appropriate place for such a validation would be here: 

https://github.com/open-telemetry/opentelemetry-operator/blob/4ead5fe1079b6d83c5c1796ee45ef93c6f88671f/api/v1alpha1/opentelemetrycollector_webhook.go#L59-L81

VineethReddy02: Added the replicas validation in my latest commit.

pkg/collector/deployment_test.go
```diff
@@ -17,6 +17,8 @@ package collector_test
 import (
 	"testing"
 
+	v1 "k8s.io/api/core/v1"
```
jpkrohling, line 20: The import order is odd here.

api/v1alpha1/opentelemetrycollector_webhook.go
```diff
@@ -62,6 +62,17 @@ func (r *OpenTelemetryCollector) ValidateCreate() error {
 	if r.Spec.Mode != ModeStatefulSet && len(r.Spec.VolumeClaimTemplates) > 0 {
 		return errors.New("Can only specify VolumeClaimTemplates for statefulsets")
 	}
+
+	// validate replicas for sidecar and daemonsets mode
```
jpkrohling, line 66: I think it might be worth extracting the common validation checks to a specific function. Right now, ValidateCreate and ValidateUpdate are pretty much the same.

api/v1alpha1/opentelemetrycollector_webhook.go
```diff
@@ -62,6 +62,17 @@ func (r *OpenTelemetryCollector) ValidateCreate() error {
...
+	if (r.Spec.Mode == ModeSidecar || r.Spec.Mode == ModeDaemonSet) && *r.Spec.Replicas > 0 {
+		return errors.New("Can only specify Replicas for deployments and statefulsets")
```
jpkrohling, line 68: It might be better to align the message with the check, like: `the OpenTelemetry Collector mode is set to %s, which does not support the attribute "replicas"`

jpkrohling (commented):
Looks good, but there's an import order issue. Because of that, I requested two other minor changes. Other than that, looks good to me!

jpkrohling (approved):

jpkrohling:
Could you please check the test failure?

jpkrohling (approved):

VineethReddy02:
@jpkrohling it looks like mergify task post all checks is failing. :)

PR #306, opened 2021-06-08
Update project to use operator-sdk v1.8.0

owais:
Latest operator-sdk (v1.8.0) does not work with the project as the schema for the PROJECT file has changed. Updated PROJECT file and then ran `operator-sdk bundle` to update config files so we can review what next bundle will look like.


PROJECT
```diff
@@ -1,11 +1,7 @@
-projectName: opentelemetry-operator
```
jpkrohling, line 1: Shouldn't the projectName be kept?

jpkrohling (commented):
LGTM, just a question about the changes to the project file.

PROJECT
owais: I thought the operator-sdk doesn't like it anymore.  However, looking at the source code, it looks like kubebuilder v2 does not have projectName defined but v3 does. I think we were already using v3 schema but version number was set to v2? I'll dig deeper into this and try to figure it out.

--------
Error we get if we keep projectName:
```

FATA[0001] unable to load the configuration: unable to unmarshal config at "PROJECT": error unmarshalling project configuration: error unmarshaling JSON: while decoding JSON: json: unknown field "projectName"
make: *** [bundle] Error 1
```

owais: I'm not sure how it was working before but "v2" projects cannot have `projectName`, `layout` or `plugins` keys in the PROJECT file. I updated the file to change the version to "v3" and now it accepts these values. 

I'm not sure what other implications updating to "v3" might have but the generated files look fine to me.

owais: I'll compare the generated bundle with previously published one and see if there are any important differences.

owais: https://gist.github.com/owais/537b9416ff7564f5f022aa35d50a3da8

Just took a cursory look and looked fine to me. I've to go right now. Will take another detailed look in a while and update here.

jpkrohling:
@owais, could you rebase this? It will make the new tests run, which should make the test checks pass.

PROJECT
jpkrohling: I think I had problems with webhooks with v3, but let's see if the CI passes. I believe we do have tests exercising the webhooks.

config/manifests/bases/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -19,7 +19,8 @@ spec:
   apiservicedefinitions: {}
   customresourcedefinitions:
     owned:
-    - description: OpenTelemetryCollector is the Schema for the opentelemetrycollectors API.
+    - description: OpenTelemetryCollector is the Schema for the opentelemetrycollectors
+        API.
```
VineethReddy02, line 23: Is this a type? `API` in new line?

jpkrohling: It's auto-generated code.

owais:
@jpkrohling Rebased. I'm trying to run the latest generated bundle on OpenShift locally but it is failing so please don't merge this just yet.

config/manifests/bases/opentelemetry-operator.clusterserviceversion.yaml
owais: auto generated by `make bundle`

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -21,11 +21,11 @@ metadata:
     containerImage: quay.io/opentelemetry/opentelemetry-operator
     createdAt: "2020-12-16T13:37:00+00:00"
     description: Provides the OpenTelemetry components, including the Collector
-    operators.operatorframework.io/builder: operator-sdk-v1.3.0
-    operators.operatorframework.io/project_layout: go
+    operators.operatorframework.io/builder: operator-sdk-v1.7.1+git
```
jpkrohling, line 24: Shouldn't it be 1.8.0?

jpkrohling (dismissed):

jpkrohling:
Looks good to me. I marked this as a draft so that I won't accidentally merge it. Once you confirm this to work on OpenShift, let me know and I'll merge this.

PROJECT
```diff
@@ -3,9 +3,13 @@ domain: opentelemetry.io
...
+  group: core
   kind: OpenTelemetryCollector
+  path: k8s.io/api/core/v1alpha1
   version: v1alpha1
-version: 2
+version: "3"
 plugins:
-  go.operator-sdk.io/v2-alpha: {}
+  manifests.sdk.operatorframework.io/v2: {}
+  scorecard.sdk.operatorframework.io/v2: {}
```
owais, line 15: as recommended by: https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.4.0/#gov2-change-go-project-plugin-object-to-manifests-and-scorecard-objects

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: quay.io/opentelemetry/opentelemetry-operator
+  newTag: v0.27.0
```
owais, line 8: This was recommended by the migration/upgrade guide and without it I could not get the bundle to use the operator image. It would try to find an image called `controller` which would obviously absent.

config/samples/core_v1alpha1_opentelemetrycollector.yaml
```diff
@@ -9,15 +9,11 @@ spec:
         protocols: 
           grpc:
 
-    processors:
-      queued_retry:
```
owais, line 13: Since latest collector does not have this anymore

owais:
@jpkrohling I went through the upgrade guides from v1.3 to v1.8 and made all changes which seemed applicable to this project. 

I verified by running this on a local OpenShift cluster. I built `make bundle-build BUNDLE_IMG=img-name` to create bundle image and then ran it with `operator-sdk run bundle img-name`. Then I created one instance of collector from the operator. Everything seemed to be functioning properly.

config/manager/kustomization.yaml
jpkrohling: There is a target in the Makefile that sets the image name to what we have as vars there:

https://github.com/open-telemetry/opentelemetry-operator/blob/aaabe4ef12e9e445c165f19385dfc5b431c781ef/Makefile#L69-L71

But if they do have a new recommendation about this, we should definitely follow it.

jpkrohling:
You are the hero!

jpkrohling:
Would you be open to creating a doc at the root of the product with the steps necessary to update the operator-sdk? Which docs and changelogs to check, how to test, and so on? Not as part of this PR, though.

config/prometheus/monitor.yaml
```diff
@@ -11,6 +11,10 @@ spec:
   endpoints:
     - path: /metrics
       port: https
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        insecureSkipVerify: true
```
jpkrohling, line 17: Is this also part of the migration? Or did you have to set this manually?

jpkrohling (commented):
Looks good, just need clarification about the TLS configuration for Prometheus. Given that the e2e tests for Kubernetes passed, I don't think this change broke anything for plain Kubernetes.

Thank you for this PR!

config/prometheus/monitor.yaml
owais: Yes and yes. It was suggested by migration guide and had to be done manually. https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.6.1/#gov2-gov3-ansiblev1-helmv1-add-scheme-token-and-tls-config-to-the-prometheus-servicemonitor-metrics-endpoint

jpkrohling: I'm a bit uneasy about this, as we have all the pieces there to make a proper secure connection. I'm opening an issue to track this and merge as is, as it's certainly an improvement to what we have today (no TLS at all).

jpkrohling: Created #308

jpkrohling (approved):

PR #307, opened 2021-06-08
add go lint rule in Makefile

VineethReddy02:
We have the go-lint checks in place in CI. But we do not have it in Makefile to test the linting locally before committing. So added it into Makefile for a better development experience.

### DIFF
--- a/config/crd/bases/opentelemetry.io_opentelemetrycollectors.yaml
+++ b/config/crd/bases/opentelemetry.io_opentelemetrycollectors.yaml
@@ -282,6 +282,7 @@ spec:
                 type: string
               tolerations:
                 description: Toleration to schedule OpenTelemetry Collector pods.
+                  This is only relevant to daemonsets, statefulsets and deployments
                 items:
                   description: The pod this Toleration is attached to tolerates any
                     taint that matches the triple <key,value,effect> using the matching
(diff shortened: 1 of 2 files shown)


config/crd/bases/opentelemetry.io_opentelemetrycollectors.yaml
```diff
@@ -282,6 +282,7 @@ spec:
                 type: string
               tolerations:
                 description: Toleration to schedule OpenTelemetry Collector pods.
+                  This is only relevant to daemonsets, statefulsets and deployments
```
jpkrohling, line 285: Could you revert this change? Otherwise it might conflict with another PR that is fixing this.

VineethReddy02: I didn't see this in other PRs. I mean even in your recent PR (you reverted it, so I was thinking to add it with a separate PR).

jpkrohling: This is being changed here: https://github.com/open-telemetry/opentelemetry-operator/pull/306/files#diff-c5e8337ee47f0f73be5c1b6eb6e52fff3c9cde0a18b59d696dc827e1fba75e10R285

jpkrohling (approved):

Issue #116, opened 2020-11-13
Refresh go proxy module information on release


jpkrohling:
When releasing the opentelemetry-operator, we might want to make an HTTP request to the Go module proxy, so that it knows about the new version. This should then cause the pkg.go.dev website to refresh the Godoc for this operator.

This is the URL that we should call for new releases: https://proxy.golang.org/github.com/open-telemetry/opentelemetry-operator/@v/v0.14.0.info , replacing `0.14.0` with the released version.

More info: https://go.dev/about

KKelvinLo:
Hey I'd like to pick this up

jpkrohling:
Here's our release script: https://github.com/open-telemetry/opentelemetry-operator/blob/main/.github/workflows/release.yaml

That script could just make a curl call to the URL mentioned earlier, based on the tag information. Check the GitHub Actions documentation to see which env var contains the tag, and use that when building the URL.

VineethReddy02:
I can take this up! :)

owais:
Should we call the URL directly or just run `go get/install` with the newly released version? I think the later is a bit more "future proof"?

jpkrohling:
These are the options:

![image](https://user-images.githubusercontent.com/13387/123131221-c9e91c80-d44d-11eb-8529-4e1299624ec8.png)

I would be fine with the `go get` approach :-)

owais:
Ah, so making request to the specific URL is officially supported. I suggested `go get` just in case the underlying URL scheme changes in future but if it is officially supported then either way should be fine.

PR #315, opened 2021-06-17
Refresh go proxy module info on release

VineethReddy02:
fixes: #116

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -29,3 +29,8 @@ jobs:
       env:
         GH_WRITE_TOKEN: ${{ secrets.GH_WRITE_TOKEN }}
       run: ./.ci/create-release-github.sh
+
+    - name: "refresh go proxy module info on release"
+      run: |
+        OPERATOR_VERSION=$(git describe --tags)
+        curl https://proxy.golang.org/github.com/open-telemetry/opentelemetry-operator/@v/${OPERATOR_VERSION}.info


.github/workflows/release.yaml
```diff
@@ -29,3 +29,8 @@ jobs:
       env:
         GH_WRITE_TOKEN: ${{ secrets.GH_WRITE_TOKEN }}
       run: ./.ci/create-release-github.sh
+
+    - name: "refresh go proxy module info on release"
+      run: |
+        OPERATOR_VERSION=$(git describe --tags)
+        curl https://proxy.golang.org/github.com/open-telemetry/opentelemetry-operator/@v/${OPERATOR_VERSION}.info
```
owais, line 36: How about running `go get` or `get install` to fetch the new version instead?

VineethReddy02: I am not sure how to fetch the latest version using `go get` do you mind sharing the cmd to fetch the version? Currently, I am using the same method used to create release tag which is already being used in prior step of release flow [here](https://github.com/open-telemetry/opentelemetry-operator/blob/main/.ci/create-release-github.sh#L3)

owais:
```
go get github.com/open-telemetry/opentelemetry-operator@{VERSION}
```

should work. For example,

```
go get github.com/open-telemetry/opentelemetry-operator@v0.27.0
```

jpkrohling: Did it work to use the notation that @owais mentioned?

VineethReddy02: the `go get` just pulls the dep, but I am unable to test the refresh thing as it happens after the releases. How about letting it be `curl` as suggested in go docs?

jpkrohling: I think `curl` is more explicit about the intentions. It would also remove the doubts on whether the call actually got to the destination. I was referring to the env var with the version, though :-)

VineethReddy02: Switched it back to `curl` and I noticed during the release process we are creating operator release in `create-release-github.sh` using the same git cmd [here](https://github.com/open-telemetry/opentelemetry-operator/blob/main/.ci/create-release-github.sh#L3) so now during the release we create a new version and perform the curl on the same version with current changes.

jpkrohling (approved):

Issue #322, opened 2021-06-24
Bump to Operator SDK v1.9.0


jpkrohling:
Similar to #298, but for v1.9.0. Changelog: https://github.com/operator-framework/operator-sdk/releases/tag/v1.9.0

jpkrohling:
@owais, do you remember which commands you had to run to update the bundles in https://github.com/open-telemetry/opentelemetry-operator/pull/311 ?

owais:
IIRC, I only ran `make bundle` to generate the bundle.

I used the following to build a docker image from the bundle so I could test locally. 

```
 make bundle-build BUNDLE_IMG=quay.io/repo/image-name VERSION=0.27.0
```

VineethReddy02:
I am working on this!

PR #327, opened 2021-06-25
Bump operator-sdk to v1.9.0, change v0.27.0 tags to v0.28.0

VineethReddy02:
FIxes: #322 

Note: I tested the operator-sdk version bump by running the operator locally & created a sample CR.

### DIFF
--- a/bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
@@ -21,11 +21,11 @@ metadata:
     containerImage: quay.io/opentelemetry/opentelemetry-operator
     createdAt: "2020-12-16T13:37:00+00:00"
     description: Provides the OpenTelemetry components, including the Collector
-    operators.operatorframework.io/builder: operator-sdk-v1.8.0
+    operators.operatorframework.io/builder: operator-sdk-v1.9.0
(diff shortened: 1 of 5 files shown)


bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -297,15 +297,15 @@ spec:
...
+  version: 0.28.0
   webhookdefinitions:
   - admissionReviewVersions:
     - v1
     - v1beta1
     containerPort: 443
     deploymentName: opentelemetry-operator-controller-manager
     failurePolicy: Fail
-    generateName: vopentelemetrycollector.kb.io
+    generateName: vopentelemetrycollectorcreateupdate.kb.io
```
VineethReddy02, line 308: @jpkrohling I just realized I need to run `make bundle` after making a change to webhooks (missed these changes in previous PR). Usually, I run `make manifests container container-push deploy` as the manifests are generated and I test the deployed changes before I create the PR.

jpkrohling: In another PR, could you add a note about it in the contributing.md? Something like: "when the CRD changes, run ...". 

For the Jaeger Operator, we have a target that will fail the build in case `make generate` would yield changes, implying that the PR has changes that impact the auto-generated files. Perhaps we could have something like this here as well?

jpkrohling (approved):

PR #332, opened 2021-07-01
Allow customizing security context

owais:
This allows users to customize the container security context settings. It's helpful when some collector features need elevated privileges.

### DIFF
--- a/PROJECT
+++ b/PROJECT
@@ -12,4 +12,3 @@ resources:
 version: "3"
 plugins:
   manifests.sdk.operatorframework.io/v2: {}
-  scorecard.sdk.operatorframework.io/v2: {} 
--- a/api/v1alpha1/opentelemetrycollector_types.go
+++ b/api/v1alpha1/opentelemetrycollector_types.go
@@ -51,6 +51,11 @@ type OpenTelemetryCollectorSpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
 	ServiceAccount string `json:"serviceAccount,omitempty"`
 
+	// SecurityContext will be set as the container security context.
+	// +optional
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	SecurityContext *v1.SecurityContext `json:"securityContext,omitempty"`
+
 	// VolumeClaimTemplates will provide stable storage using PersistentVolumes. Only available when the mode=statefulset.
 	// +optional
 	// +listType=atomic
--- a/bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
@@ -21,7 +21,7 @@ metadata:
     containerImage: quay.io/opentelemetry/opentelemetry-operator
     createdAt: "2020-12-16T13:37:00+00:00"
     description: Provides the OpenTelemetry components, including the Collector
-    operators.operatorframework.io/builder: operator-sdk-v1.8.0+git
+    operators.operatorframework.io/builder: operator-sdk-v1.9.0
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v2
     repository: github.com/open-telemetry/opentelemetry-operator
     support: OpenTelemetry Community
--- a/pkg/collector/container.go
+++ b/pkg/collector/container.go
@@ -69,11 +69,12 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
 	}
 
 	return corev1.Container{
-		Name:         naming.Container(),
-		Image:        image,
-		VolumeMounts: volumeMounts,
-		Args:         args,
-		Env:          envVars,
-		Resources:    otelcol.Spec.Resources,
+		Name:            naming.Container(),
+		Image:           image,
+		VolumeMounts:    volumeMounts,
+		Args:            args,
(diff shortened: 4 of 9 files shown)


PROJECT
```diff
@@ -12,4 +12,3 @@ resources:
 version: "3"
 plugins:
   manifests.sdk.operatorframework.io/v2: {}
-  scorecard.sdk.operatorframework.io/v2: {}
```
owais, line 15: we don't have scorecard tests so there is no need to include this plugin.

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -21,7 +21,7 @@ metadata:
     containerImage: quay.io/opentelemetry/opentelemetry-operator
     createdAt: "2020-12-16T13:37:00+00:00"
     description: Provides the OpenTelemetry components, including the Collector
-    operators.operatorframework.io/builder: operator-sdk-v1.8.0+git
```
owais, line 24: Current version was also generated with 1.9 but the build was tagged 1.8.0+git wrongly

api/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -51,6 +51,11 @@ type OpenTelemetryCollectorSpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
 	ServiceAccount string `json:"serviceAccount,omitempty"`
 
+	// SecurityContext will be set as the container security context
+	// +optional
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	SecurityContext v1.SecurityContext `json:"securityContext,omitempty"`
```
jpkrohling, line 57: You have to use it here, don't you? 

https://github.com/open-telemetry/opentelemetry-operator/blob/aff74858930ceaddf8a972db455c716906f1bf39/pkg/collector/container.go#L71-L78

owais: Ah my bad. I thought it worked as expected in my tests but may be I didn't look hard enough. Let me spin up local openshfit and report back in 78 hours once it has started :)

jpkrohling: > in 78 hours once it has started

ROFLMAO

For the record, I use `crc` for local testing, and a cluster starts relatively quick (under 5 mins).

owais: Yeah, it does take roughly the same amount of time. I was just exaggerating. It's not that bad. It does shut down occasionally on it's own though and if i keep it running for longer than an hour, I start seeing a ton of i/o timeout network errors. Other than that it is pretty solid. I think I should be able to finish the test and update here in the next couple of hours or so if I don't get distracted. :)

owais: Fixed and verified on a KIND cluster locally. Thanks for catching this.

pkg/collector/container.go
```diff
@@ -69,11 +69,12 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
...
-		Args:         args,
-		Env:          envVars,
-		Resources:    otelcol.Spec.Resources,
+		Name:            naming.Container(),
+		Image:           image,
+		VolumeMounts:    volumeMounts,
+		Args:            args,
+		Env:             envVars,
+		Resources:       otelcol.Spec.Resources,
+		SecurityContext: &otelcol.Spec.SecurityContext,
```
jpkrohling, line 78: Could you add a unit test, ensuring that a security context set at the CR will propagate down to the container?

jpkrohling (commented):
LGTM, and is ready to be merged as soon as a unit test is added.

pkg/collector/container.go
owais: Sure. Im out for the weekend. Will do on Monday.

owais: Followed existing test for volumes and added a new one for security context. LMK if you think that's enough.

api/v1alpha1/opentelemetrycollector_types.go
jpkrohling, line 57: Sorry for catching this only now, but this should be a pointer. Otherwise, we won't know when it's been specified explicitly by the user. We need this signal if we ever decide to change the default value we pass down to the containers.

pkg/collector/container.go
jpkrohling: They look good!

VineethReddy02: @owais Currently we are manually maintaining the Otelcol spec in this [doc](https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/otelcol_cr_spec.md) to let users know what can be configured in the CR. Can you add `SecurityContext` to this doc?

VineethReddy02:
LGTM!

pkg/collector/container.go
jpkrohling: Totally forgot about the doc... Another reason to automate it as soon as possible :-)

api/v1alpha1/opentelemetrycollector_types.go
owais: Updated

pkg/collector/container.go
owais: Updated

jpkrohling (approved):

Issue #331, opened 2021-07-01
Validate auto-generated files in CI


VineethReddy02:
For the Jaeger Operator, we have a target that will fail the build in case make generate would yield changes, implying that the PR has changes that impact the auto-generated files.

VineethReddy02:
I'm working on this!

PR #336, opened 2021-07-07
Add CI check for auto-generated files

VineethReddy02:
add CI validation check to make sure all auto-generated files are part of the PR.

fixes: #331

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,7 +28,7 @@ $ gh pr create
 
 Build the manifests, install the CRD and run the operator as a local process:
 ```
-$ make manifests install run
+$ make bundle install run
 ```
 
 ### Deployment with webhooks
@@ -47,7 +47,7 @@ make cert-manager
 Once it's ready, the following can be used to build and deploy a manager, along with the required webhook configuration:
 
 ```
-make manifests container container-push deploy
+make bundle container container-push deploy
 ```
 
 By default, it will generate an image following the format `quay.io/${USER}/opentelemetry-operator:${VERSION}`. You can set the following env vars in front of the `make` command to override parts or the entirety of the image:
--- a/Makefile
+++ b/Makefile
@@ -44,8 +44,8 @@ endif
 KUBE_VERSION ?= 1.21
 KIND_CONFIG ?= kind-$(KUBE_VERSION).yaml
 
+ensure-generate-is-noop: VERSION=$(OPERATOR_VERSION)
 ensure-generate-is-noop: generate bundle
-	cat ./bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
 	@git diff -s --exit-code api/v1alpha1/zz_generated.*.go || (echo "Build failed: a model has been changed but the generated resources aren't up to date. Run 'make generate' and update your PR." && exit 1)
 	@git diff -s --exit-code bundle config || (echo "Build failed: the bundle, config files has been changed but the generated bundle, config files aren't up to date. Run 'make bundle' and update your PR." && exit 1)
 
@@ -206,7 +206,7 @@ endif
 # Generate bundle manifests and metadata, then validate generated files.
 bundle: kustomize operator-sdk manifests
 	$(OPERATOR_SDK) generate kustomize manifests -q
-	$(KUSTOMIZE) build config/manifests | $(OPERATOR_SDK) generate bundle -q --overwrite --manifests --version $(OPERATOR_VERSION) $(BUNDLE_METADATA_OPTS)
+	$(KUSTOMIZE) build config/manifests | $(OPERATOR_SDK) generate bundle -q --overwrite --manifests --version $(VERSION) $(BUNDLE_METADATA_OPTS)
 	$(OPERATOR_SDK) bundle validate ./bundle
 
 # Build the bundle image, used only for local dev purposes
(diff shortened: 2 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -28,7 +28,7 @@ $ gh pr create
 
 Build the manifests, install the CRD and run the operator as a local process:
 ```
-$ make manifests install run
+$ make bundle install run
```
VineethReddy02, line 31: changing `manifests` to `bundle` because I see bundle internally runs `manifests` already, so running `bundle` also generates CRD level changes if any made.  

Also, If my understanding is right based on commit history. `make manifests` is added in docs prior to adding the `make bundle` target in Makefile. So now changing it to `make bundle` makes more sense for auto-generating the files.

Makefile
```diff
@@ -200,7 +205,7 @@ endif
 # Generate bundle manifests and metadata, then validate generated files.
 bundle: kustomize operator-sdk manifests
 	$(OPERATOR_SDK) generate kustomize manifests -q
-	$(KUSTOMIZE) build config/manifests | $(OPERATOR_SDK) generate bundle -q --overwrite --manifests --version $(VERSION) $(BUNDLE_METADATA_OPTS)
+	$(KUSTOMIZE) build config/manifests | $(OPERATOR_SDK) generate bundle -q --overwrite --manifests --version $(OPERATOR_VERSION) $(BUNDLE_METADATA_OPTS)
```
VineethReddy02, line 208: By default, the version here is something like `0.29.0-5-gf6a55f6` but for CI level validation we need the current operator version. So adding a new variable with the name `OPERATOR_VERSION` which picks operator version from `versions.txt` 

I am not sure of any other issues with generating bundle files with the current version, But open for suggestions here @jpkrohling

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -21,7 +21,7 @@ metadata:
     containerImage: quay.io/opentelemetry/opentelemetry-operator
     createdAt: "2020-12-16T13:37:00+00:00"
     description: Provides the OpenTelemetry components, including the Collector
-    operators.operatorframework.io/builder: operator-sdk-v1.9.0
+    operators.operatorframework.io/builder: operator-sdk-v1.8.0+git
```
VineethReddy02, line 24:
```
    - uses: jpkrohling/setup-operator-sdk@v1.1.0
      with:
        operator-sdk-version: v1.9.0
```

The operator-sdk in CI Is generating the version as `v1.8.0+git`. So changed it back but open to make changes to `operator-sdk` github action to generate the version as `v1.9.0`

Makefile
```diff
@@ -43,11 +44,16 @@ endif
 KUBE_VERSION ?= 1.21
 KIND_CONFIG ?= kind-$(KUBE_VERSION).yaml
 
+ensure-generate-is-noop: generate bundle
+	cat ./bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```
jpkrohling, line 48: I'd keep this out. If anything, we can just show the diff. Otherwise, it will spit out this file every time, including when we don't need it.

Makefile
jpkrohling: I'm not quite sure. I think I would prefer to have `VERSION=x.y.z make bundle` (or `ensure-...`) as part of the CI instead. Generating a dev bundle (`make bundle` locally) should still use the notation we are currently using.

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
jpkrohling: Sounds reasonable to me. cc @owais

Makefile
VineethReddy02: This is just to show that the issue. I will definitely undo this line before merging.... Just to confirm it in CI I added this for now. :)

jpkrohling:
This is still missing passing the right version to generate the bundle via the workflow file, right?

VineethReddy02:
Passing the current operator version is specific to `ensure-generate-is-noop` target right? If the user performs locally `make ensure-generate-is-noop` it should perform the same as workflow CI, right? So I am passing version at the `Makefile` level so whoever triggers this target performs the same...

```
ensure-generate-is-noop: VERSION=$(OPERATOR_VERSION)
ensure-generate-is-noop: generate bundle
	@git diff -s --exit-code api/v1alpha1/zz_generated.*.go || (echo "Build failed: a model has been changed but the generated resources aren't up to date. Run 'make generate' and update your PR." && exit 1)
	@git diff -s --exit-code bundle config || (echo "Build failed: the bundle, config files has been changed but the generated bundle, config files aren't up to date. Run 'make bundle' and update your PR." && exit 1)
```

VineethReddy02:
I was thinking adding VERSION value in Makefile `ensure-generate-is-noop` target level performs the same action when `make ensure-generated-is-noop` is triggered locally or at CI level. So I added the current operator version in the `ensure-generate-is-noop` target as
```
ensure-generate-is-noop: VERSION=$(OPERATOR_VERSION)
ensure-generate-is-noop: generate bundle
	@git diff -s --exit-code api/v1alpha1/zz_generated.*.go || (echo "Build failed: a model has been changed but the generated resources aren't up to date. Run 'make generate' and update your PR." && exit 1)
	@git diff -s --exit-code bundle config || (echo "Build failed: the bundle, config files has been changed but the generated bundle, config files aren't up to date. Run 'make bundle' and update your PR." && exit 1)
```

Any specific reason why should this be part of the CI workflow level?

jpkrohling:
I missed that change, I think it is indeed better!

jpkrohling (approved):

Issue #339, opened 2021-07-15
Update to collector v0.30.0


jpkrohling:
Same as #323, but for 0.30.0. Changelog: https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.30.0

PR #342, opened 2021-07-15
Bumps collector version to v0.30.0

VineethReddy02:
fixes: #339

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Changes by Version
 ==================
 
+0.30.0 (2021-07-15)
+-------------------
+* Bumped OpenTelemetry Collector to v0.30.0
+* Container Security Context ([#332](https://github.com/open-telemetry/opentelemetry-operator/pull/332), [@owais](https://github.com/owais))
+
 0.29.0 (2021-06-25)
 -------------------
 * Bumped OpenTelemetry Collector to v0.29.0
--- a/README.md
+++ b/README.md
@@ -164,6 +164,7 @@ The OpenTelemetry Operator *might* work on versions outside of the given range, 
 
 | OpenTelemetry Operator | Kubernetes           |
 |------------------------|----------------------|
+| v0.30.0                | v1.19 to v1.21       |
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 Changes by Version
 ==================
 
+0.30.0 (2021-07-15)
+-------------------
+* Bumped OpenTelemetry Collector to v0.30.0
+* Add CI check for auto-generated files ([#336](https://github.com/open-telemetry/opentelemetry-operator/pull/336), [@VineethReddy02](https://github.com/VineethReddy02))
```
jpkrohling, line 7: I think this should be left out. The changelog should contain only entries relevant to our end users.

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
...
+* Container Security Context ([#332](https://github.com/open-telemetry/opentelemetry-operator/pull/332), [@owais](https://github.com/owais))
+* Small manual adjusts for docs and bundle for v0.29.0 ([#329](https://github.com/open-telemetry/opentelemetry-operator/pull/329), [@jpkrohling](https://github.com/jpkrohling))
```
jpkrohling, line 9: This should also be out.

jpkrohling (commented):
What do you think about turning this type of commit into release ones? As such, they would have to follow the first steps from the RELEASE.md

Concretely, this PR would be missing:

1. bundle changes
2. update the compat matrix

This way, we could release the operator right after this type of commit.

VineethReddy02:
@jpkrohling I haven't noticed any migration-related steps in OtelCol changelog. But I would like you to confirm this :)

jpkrohling:
Looks good, can you just confirm that you ran the `make bundle` as described in the readme, instead of just bumping the versions?

VineethReddy02:
yup, we do have checks for the bundle if any changes are left locally as we recently added `ensure-generate-is-noop` in Makefile.

jpkrohling (approved):

Issue #128, opened 2020-11-24
Sync list of receivers


jpkrohling:
The operator knows about a few receivers, but I'm not sure this list is up to date:
https://github.com/open-telemetry/opentelemetry-operator/tree/master/pkg/collector/parser

This ticket is to sync this list.

Current delta:

- [ ] awscontainermetrics
- [ ] awsxray
- [ ] dockerstats
- [ ] jmx
- [ ] k8scluster
- [ ] kubeletstats
- [ ] memcached
- [ ] nginx
- [ ] prometheusexec
- [ ] recever
- [ ] redis
- [ ] simpleprometheus
- [ ] splunk
- [ ] stanza

gramidt:
I'll take this @jpkrohling

gramidt:
## The current deltas:

### collector

Currently update to date. May need to compare settings to make sure everything is still following convention.

### collector-contrib

- awscontainermetrics
- awsxray
- dockerstats
- jmx
- k8scluster
- kubeletstats
- memcached
- nginx
- prometheusexec
- recever
- redis
- simpleprometheus
- splunk
- stanza

gramidt:
@jpkrohling - What are your thoughts about the "collector-contrib" receivers? Was your plan to rely on `receiver_generic.go` for deltas?

jpkrohling:
No, the generic is just an implementation detail, as most receivers can be managed in the same way. Basically, if the receiver has an `endpoint` property that accepts a `host:port` value, it can make use of the generic adapter.

jpkrohling:
Perhaps it was implicit in the comment above, but for now, all we care about is the `endpoint` property for the receiver, as we'll use this to automatically add the needed ports to the `Service`. So, there's no need to sync the other properties in each receiver.

gramidt:
Perfect! That's exactly what I was thinking. I'll start tackling the deltas 1-by-1 with individual PRs for each. Should we also split this into multiple issues, @jpkrohling?

jpkrohling:
This issue here has the list of receivers already, if we reference this issue in all individual PRs, it should be fine.

gramidt:
@jpkrohling - Could you add me as an assignee, so I can more easily track this work?

jpkrohling:
Done

gramidt:
Thank you!

gramidt:
[...]

PR #361, opened 2021-07-27
Add memcached receiver

VineethReddy02:
[memcached receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/memcachedreceiver)

Part of #128


VineethReddy02:
As this is a scrape receiver. we do not need parser in the operator repo.

PR #362, opened 2021-07-28
Add nginx receiver

VineethReddy02:
[nginx receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/nginxreceiver)

Part of #128


VineethReddy02:
As this is a scrape receiver. we do not need parser in the operator repo.

PR #363, opened 2021-07-28
Add docker-stats receiver

VineethReddy02:
[docker_stats receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/dockerstatsreceiver)

Part of #128


VineethReddy02:
As this is a scrape receiver. we do not need a parser in the operator repo.

Issue #366, opened 2021-07-29
Update to collector v0.31.0


VineethReddy02:
Same as #323, but for 0.31.0. Changelog: https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.31.0

Issue #359, opened 2021-07-26
Revert kustomization of the manager image


jpkrohling:
In a previous update, a customization of the manager image got committed, causing all CSV and other generated resources to use this instead of newer versions.

https://github.com/open-telemetry/opentelemetry-operator/blob/7250758d04076b0e011b196d10bdea885b84fe80/config/manager/kustomization.yaml#L5-L8

Example of a side-effect: https://github.com/k8s-operatorhub/community-operators/pull/39

VineethReddy02:
closing the issue as its been fixed #369

PR #369, opened 2021-07-29
Bump collector to v0.31.0

VineethReddy02:
Fixes: #366 #359

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,8 @@ KIND_CONFIG ?= kind-$(KUBE_VERSION).yaml
 
 ensure-generate-is-noop: VERSION=$(OPERATOR_VERSION)
 ensure-generate-is-noop: generate bundle
+	# on make bundle config/manager/kustomization.yaml includes changes, which should be ignored for the below check
+	@git restore config/manager/kustomization.yaml
 	@git diff -s --exit-code api/v1alpha1/zz_generated.*.go || (echo "Build failed: a model has been changed but the generated resources aren't up to date. Run 'make generate' and update your PR." && exit 1)
 	@git diff -s --exit-code bundle config || (echo "Build failed: the bundle, config files has been changed but the generated bundle, config files aren't up to date. Run 'make bundle' and update your PR." && exit 1)
 
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,8 @@ KUBE_VERSION ?= 1.21
 KIND_CONFIG ?= kind-$(KUBE_VERSION).yaml
 
 ensure-generate-is-noop: VERSION=$(OPERATOR_VERSION)
-ensure-generate-is-noop: generate bundle
+ensure-generate-is-noop: USER=opentelemetry
+ensure-generate-is-noop: set-image-controller generate bundle
 	# on make bundle config/manager/kustomization.yaml includes changes, which should be ignored for the below check
 	@git restore config/manager/kustomization.yaml
 	@git diff -s --exit-code api/v1alpha1/zz_generated.*.go || (echo "Build failed: a model has been changed but the generated resources aren't up to date. Run 'make generate' and update your PR." && exit 1)
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,7 +2,7 @@
 
 Steps to release a new version of the OpenTelemetry Operator:
 
-1. Change the `versions.txt`, so that it lists the target version of the OpenTelemetry Collector (operand), and the desired version for the operator. The `major.minor` should typically match, with the patch portion being possibly different.
+1. Change the `versions.txt` & `config/manager/kustomization.yaml`, so that it lists the target version of the OpenTelemetry Collector (operand), and the desired version for the operator. The `major.minor` should typically match, with the patch portion being possibly different.
 1. Run `make bundle VERSION=0.29.0`, using the version that will be released.
 1. Change the compatibility matrix in the readme file, using the OpenTelemetry Operator version to be released and the current latest Kubernetes version as the latest supported version, with N-2 being the lower bound. Make sure that the CI is currently testing the latest Kubernetes version!
 1. Add the changes to the changelog
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,7 +2,7 @@
 
 Steps to release a new version of the OpenTelemetry Operator:
 
-1. Change the `versions.txt` & `config/manager/kustomization.yaml`, so that it lists the target version of the OpenTelemetry Collector (operand), and the desired version for the operator. The `major.minor` should typically match, with the patch portion being possibly different.
+1. Change the `versions.txt`, so that it lists the target version of the OpenTelemetry Collector (operand), and the desired version for the operator. The `major.minor` should typically match, with the patch portion being possibly different.
 1. Run `make bundle VERSION=0.29.0`, using the version that will be released.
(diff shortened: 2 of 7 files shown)


config/manager/kustomization.yaml
```diff
@@ -2,7 +2,3 @@ resources:
 - manager.yaml
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
-images:
```
VineethReddy02, line 5: This part fixes #359

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -206,7 +206,7 @@ spec:
                 - --enable-leader-election
                 command:
                 - /manager
-                image: quay.io/opentelemetry/opentelemetry-operator:v0.28.0
```
jpkrohling, line 209: This one actually has to be with the version, as this is a "final" artifact (similar to the one we add to the release). I think this is the source of the problem behind #359.

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -206,7 +206,7 @@ spec:
...
+                image: controller:v0.28.0
```
jpkrohling: No, sorry, I meant that this should be the actual image. We use a tool called "kustomize", which will take templates (bases) and merge with customizatoins. This is all within the config directory. The bundle is not a template nor a customization, it's a final result, similar to the manifest we generate during the release workflow. Therefore, whatever is inside the config directory should use "controller" for the operator image name, but things outside of it should use an actual image.

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -206,7 +206,7 @@ spec:
                 - --enable-leader-election
                 command:
                 - /manager
-                image: quay.io/opentelemetry/opentelemetry-operator:v0.28.0
+                image: controller:v0.31.0
```
jpkrohling, line 209: No, sorry, I meant that this should be the actual image. We use a tool called "kustomize", which will take templates (bases) and merge with customizatoins. This is all within the config directory. The bundle is not a template nor a customization, it's a final result, similar to the manifest we generate during the release workflow. Therefore, whatever is inside the config directory should use "controller" for the operator image name, but things outside of it should use an actual image.

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
```
jpkrohling, line 5: This whole node has to go.

RELEASE.md
```diff
@@ -2,7 +2,7 @@
 
 Steps to release a new version of the OpenTelemetry Operator:
 
-1. Change the `versions.txt`, so that it lists the target version of the OpenTelemetry Collector (operand), and the desired version for the operator. The `major.minor` should typically match, with the patch portion being possibly different.
+1. Change the `versions.txt` & `config/manager/kustomization.yaml`, so that it lists the target version of the OpenTelemetry Collector (operand), and the desired version for the operator. The `major.minor` should typically match, with the patch portion being possibly different.
```
jpkrohling, line 5: This probably needs to be reverted, right?

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
jpkrohling: To be clear: in this PR, this has to be:

```
quay.io/opentelemetry/opentelemetry-operator:v0.31.0
```

The `make bundle VERSION=0.31.0` should generate the right bundle. But doing it might cause the file `config/manager/kustomization.yaml` to be changed, fixing the version on 0.31.0, which is not desired for that file.

jpkrohling: In time: looks like we never got this bundle right in this repo. I think we always ended up fixing it when sending the PR to the operator hub. Here's what we have for the v0.29.0 tag:

https://github.com/open-telemetry/opentelemetry-operator/blob/a1aacb3b9c64f74650ce4836cfc75bb0da5ad949/bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml#L209

And this for v0.28.0:
https://github.com/open-telemetry/opentelemetry-operator/blob/af398389a1936c0fef3d4e5c94685da71184b9ac/bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml#L209

v0.27.0 looks similar to this PR, but that's still wrong:
https://github.com/open-telemetry/opentelemetry-operator/blob/d4f9c2448b239379a7b2731f0ec1a9c820b10715/bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml#L197

jpkrohling (approved):

VineethReddy02:
@jpkrohling now the CI failing due to check `ensure-generate-is-noop` as this checks

```
@git diff -s --exit-code bundle config 
```

Should I drop the diff check for `config` directory as `kustomization.yaml` is part of config dir. On `make bundle` this file is undergoing changes.

jpkrohling (dismissed):
LGTM now!

jpkrohling:
I'll merge this, but we need to figure out a way to get the bundle check fixed for this kind of commit:

```
Build failed: the bundle, config files has been changed but the generated bundle, config files aren't up to date. Run 'make bundle' and update your PR.
```

jpkrohling:
> Should I drop the diff check for config directory as kustomization.yaml is part of config dir. On make bundle this file is undergoing changes.

As we discussed on slack, a better approach would be to add something like this to the affected target: `git restore config/manager/kustomization.yaml`

Makefile
```diff
@@ -45,7 +45,10 @@ KUBE_VERSION ?= 1.21
 KIND_CONFIG ?= kind-$(KUBE_VERSION).yaml
 
 ensure-generate-is-noop: VERSION=$(OPERATOR_VERSION)
-ensure-generate-is-noop: generate bundle
+ensure-generate-is-noop: USER=opentelemetry
```
VineethReddy02, line 48: ci is generating the file `bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml` with below diff as USER in CI is referred to` runner`

```
--- a/bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
@@ -206,7 +206,7 @@ spec:
                 - --enable-leader-election
                 command:
                 - /manager
-                image: quay.io/opentelemetry/opentelemetry-operator:v0.31.0
+                image: quay.io/runner/opentelemetry-operator:v0.31.0
                 name: manager
                 ports:
                 - containerPort: 9443
Build failed: the bundle, config files has been changed but the generated bundle, config files aren't up to date. Run 'make bundle' and update your PR.
make: *** [Makefile:53: ensure-generate-is-noop] Error 1
```

So setting the USER as `opentelemetry` for the target `ensure-generate-is-noop`, would definitely appreciate any other thoughts for this fix. :)

Makefile
```diff
@@ -45,7 +45,10 @@ KUBE_VERSION ?= 1.21
...
+ensure-generate-is-noop: set-image-controller generate bundle
```
VineethReddy02, line 49: Added `set-image-controller` target to this check. As it adds the `imageName` to the generated bundle files.  Else the image name will be `controller` by default and the diff check fails.

jpkrohling: Another option would be to use an env var in the github workflow file, but it wouldn't work for local runs of `make ci`. Your solution is also good, IMO.

jpkrohling (approved):

PR #380, opened 2021-08-05
Download and install kustomize binary

### LINKED ISSUE #379: Prevent the Makefile from installing kustomize
#373 has shown that installing via `go install` or `go get` isn't particularly reliable. As such, we should switch to use the same approach we have for kuttl/kubebuilder.

jpkrohling:
This PR closes #379 by creating a script that installs kustomize and changing the Makefile target to just check for kustomize's presence, similar to what we do with other tools.

﻿Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>


VineethReddy02 (approved):

VineethReddy02:
LGTM!

PR #389, opened 2021-08-10
Target Allocator implementation (Part 3 - Collector Enhancement for Prometheus Config Update) 

Raul9595:
### Description

This PR addresses issue # 60 in the wg-prometheus repo. This is part 3 of a 3 part update in which enhancements are presented to the OTel Operator to support the deployment and reconcile support for a target allocator resource. This update enables the collectors to use _http_sd_config_ option to determine their scrape targets exposed by the target allocation service. Please refer to the [Design Doc](https://docs.google.com/document/d/19yO6tR7qxU6mBmp1KNTbyblIpkKZ1efy/edit?usp=sharing&ouid=104349769849560567749&rtpof=true&sd=true), [PDF version for code readability](https://drive.google.com/file/d/174B9KFh9gVWrXwjXD5owT-06ToUZJ4sW/view?usp=sharing) for more details

It extends [Part 1](https://github.com/open-telemetry/opentelemetry-operator/pull/351) and should be merged after Part 1 is complete.

### Type of change

New feature & Enhancement

### Testing

Implemented go unit tests as part of the new logic responsible for replacing Prometheus config.

cc @alolita @Aneurysm9 @alexperez52


Raul9595:
Hi @Aneurysm9 and @jpkrohling. Please review when you have time.

pkg/collector/reconcile/config_replace.go
```diff
@@ -0,0 +1,84 @@
...
+		return "", err
+	}
+
+	var cfg Config
+	if err = yaml.UnmarshalStrict(promCfg, &cfg); err != nil {
+		return "", fmt.Errorf("error unmarshaling YAML: %w", err)
+	}
+
+	for i := range cfg.PromConfig.ScrapeConfigs {
+		cfg.PromConfig.ScrapeConfigs[i].ServiceDiscoveryConfigs = discovery.Configs{
```
dashpole, line 64: Is it possible `ServiceDiscoveryConfigs` is already set?  What would happen in that case?

pkg/collector/reconcile/config_replace.go
```diff
@@ -0,0 +1,84 @@
...
+			},
+		}
+	}
+
+	updPromCfgMap := make(map[string]interface{})
+	if err := mapstructure.Decode(cfg, &updPromCfgMap); err != nil {
+		return "", err
+	}
+
+	// type corecion checks are handled in the ConfigToPromConfig method above
```
dashpole, line 76: corecion -> correction

pkg/collector/reconcile/configmap.go
```diff
@@ -66,6 +66,10 @@ func desiredConfigMap(_ context.Context, params Params) corev1.ConfigMap {
 	name := naming.ConfigMap(params.Instance)
 	labels := collector.Labels(params.Instance)
 	labels["app.kubernetes.io/name"] = name
+	config, err := ReplaceConfig(params)
+	if err != nil {
+		params.Log.V(2).Info("failed to parse config: ", err)
```
dashpole, line 71: Is this always a failure to parse?  It looks like there are many possible errors this can return.

pkg/collector/reconcile/config_replace.go
Raul9595: Yes, `ServiceDiscoveryConfigs` can be already set. But since we have extracted all the targets in the target allocator service (Part 2) and allocated it to the Collector instances, we would not need to worry about the existing targets. They are exposed at the HTTP endpoint. Thus,  we can replace the old `ServiceDiscoveryConfigs` with one new `ServiceDiscoveryConfigs` entry which is the `http_sd_config` pointing to the HTTP endpoint.

pkg/collector/reconcile/configmap.go
Raul9595: Yes there can be multiple errors. Since the overall goal is to parse the config and replace it with the new `http_sd_config` entry, I thought the initial message can be "failed to parse config: " and then the err can talk about the actual error as to why the parsing failed.
But I think it may confuse the user. Does "failed to replace config: " make more sense? or should I return the err alone?

dashpole: Maybe a clue as to why we are replacing the config would be helpful: "failed to update prometheus config to use sharded targets"

Raul9595: Sounds good! Thank you.

alolita:
Can you please re-review @Aneurysm9 @dashpole @jpkrohling ty!

dashpole (approved):

jpkrohling:
I'll review this once the part 2 is merged.

alolita:
Hi @Aneurysm9 let's get this PR approved and merged by @jpkrohling please. Thx.

Raul9595:
Just wanted to mention that this can be reviewed independently and does not have any dependencies on Part 2.

Aneurysm9 (approved):

PR #421, opened 2021-09-07
Add aws xray receiver

VineethReddy02:
Syncing AWS XRAY receiver from [collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/awsxrayreceiver#aws-x-ray-receiver). Added the default port as `2000` as mentioned in `README.md`.

Just to be clear I am not adding any other default config parameters such as `transport` as `udp`  and other fields as we expect this to be taken care by users right?. We are only concerned with default port that needs to be exposed for this particular receiver.

WDYT @jpkrohling ?

Part of #128

### DIFF
--- a/pkg/collector/parser/receiver_aws-xray.go
+++ b/pkg/collector/parser/receiver_aws-xray.go
@@ -1,0 +1,34 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package parser
+
+import "github.com/go-logr/logr"
+
+const parserNameAWSXRAY = "__awsxray"
+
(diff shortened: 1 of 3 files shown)


jpkrohling:
> We are only concerned with default port that needs to be exposed for this particular receiver

Correct, but if the default port is on UDP, then we might want to handle that as well. Otherwise, we'll open a default port for TCP when it should be UDP.

VineethReddy02:
In AWS xray receiver the port for both UDP and TCP are the same, they differentiate this with another field called transport user should configure it as either `udp` or `tcp`. 

Aws xray [udp endpoint](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/awsxrayreceiver#endpoint-optional)
Aws xray [tcp endpoint](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/awsxrayreceiver#endpoint-optional-1)

VineethReddy02:
@jpkrohling  once #422 gets merged I will rebase this PR to resolve the conflicts.

pkg/collector/parser/receiver_aws-xray.go
```diff
@@ -0,0 +1,34 @@
...
+// limitations under the License.
+
+package parser
+
+import "github.com/go-logr/logr"
+
+const parserNameAWSXRAY = "__awsxray"
+
+// NewAwsXrayReceiverParser builds a new parser for AWS xray receivers, from the contrib repository.
+func NewAwsXrayReceiverParser(logger logr.Logger, name string, config map[interface{}]interface{}) ReceiverParser {
```
jpkrohling, line 22:
```suggestion
func NewAWSXrayReceiverParser(logger logr.Logger, name string, config map[interface{}]interface{}) ReceiverParser {
```

jpkrohling (approved):

PR #451, opened 2021-10-05
Support changing pod collector annotations

### LINKED ISSUE #426: Provide a way to add annotations to Pods
Hi,

We're using DataDog to collect metrics. I've noticed that `prometheus.io/scrape` is [added](https://github.com/open-telemetry/opentelemetry-operator/blob/ea688b527a5b1323413d18d6843ec02486c46350/pkg/collector/annotations.go#L29-L32) to Deployment. DD supports `prometheus.io/scrape` but it has to be in the Pod spec, not in the Deployment spec. Their current implementation is quite limited though and I'd prefer to use their OpenMetrics autodiscovery hook.

For that I'd need to add annotations to Pod, e.g.:
```
  annotations:
    ad.datadoghq.com/otc-container.check_names: '["openmetrics"]'
    ad.datadoghq.com/otc-container.init_configs: '[{}]'
    ad.datadoghq.com/otc-container.instances: |-
      [
        {
          "prometheus_url": "http://%%host%%:8888/metrics",
          "namespace": "opentelemetry.collector",
          "metrics": [
            {"otelcol_exporter_queue_size": "exporter.queue_size"}
          ]
        }
      ]
```

I noticed that it is possible to add `annotations` to `OpenTelemetryCollector` CRD, but then this annotation is propagated to everything (deployment, services, pods, etc). This causes error noise and I'd like to avoid it.

It would be good to be able to control Pod annotations.

indrekj:
This allows setting podAnnotations, e.g:
```
---
kind: OpenTelemetryCollector
metadata:
  name: opentelemetry
  annotations:
    foo: "this won't be applied to pods"
spec:
  mode: deployment
  podAnnotations:
    ad.datadoghq.com/otc-container.check_names: '["openmetrics"]'
    ad.datadoghq.com/otc-container.init_configs: '[{}]'
    ad.datadoghq.com/otc-container.instances: |-
      [
        {
          "prometheus_url": "http://%%host%%:8888/metrics",
          "namespace": "opentelemetry.collector",
          "metrics": [
            {"otelcol_exporter_queue_size": "exporter.queue_size"},
            {"otelcol_exporter_send_failed_spans": "exporter.send_failed_spans"},
            {"otelcol_exporter_sent_spans": "exporter.sent_spans"},
          ]
        }
      ]
  # ...
```

Fixes #426


api/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -119,6 +119,11 @@ type OpenTelemetryCollectorSpec struct {
 	// +optional
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
 	Tolerations []v1.Toleration `json:"tolerations,omitempty"`
+
+	// PodAnnotations is the set of annotations that will be attached to Collector pods
```
jpkrohling, line 123:
```suggestion
	// PodAnnotations is the set of annotations that will be attached to Collector and Target Allocator pods
```

pkg/collector/daemonset.go
```diff
@@ -46,7 +46,7 @@ func DaemonSet(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Labels:      labels,
-					Annotations: otelcol.Annotations,
```
jpkrohling, line 49: Could you also add an entry to the changelog (perhaps after this is merged?) about this? This can be considered a breaking change.

api/v1alpha1/opentelemetrycollector_types.go
indrekj: https://github.com/open-telemetry/opentelemetry-operator/compare/8697a3f18b18e850dc6b9ef9f8d73d7fdae0019c..efad7067c17550c99f65709d28ce5e68e1ee0965

pkg/collector/daemonset.go
indrekj: You mean like a separate PR after this? I can do that. I could also add it to changelog right now under `(unreleased)` version if that's ok?

jpkrohling (approved):

pkg/collector/daemonset.go
jpkrohling: That would be fine as well.

jpkrohling:
Looks good to me. It would be even better if you could add a new e2e test, which serves as an example as well. Like this: https://github.com/open-telemetry/opentelemetry-operator/tree/main/tests/e2e/smoke-statefulset

pkg/collector/daemonset.go
indrekj: Added

indrekj:
@jpkrohling Added e2e test as well.

jpkrohling (approved):

jpkrohling:
Looks great, thanks!

PR #464, opened 2021-10-20
Add initial support for auto-instrumentation

pavolloffay:
Updates #455 

This PR adds functionality to inject OpenTelemetry auto-instrumentation (at this point only javaagent) into a pod via admission controller. Note that this is the smallest PR I could submit to support at least something end-to-end. There will be follow up PRs to provide additional features.

Notable changes 
* new CR to define SDK and instrumentation configuration
* javaagent is injected and configured if `java.inst.opentelemetry.io/inject: "true"` is specified. The value van be the instrumentation CR name.
* the javaagent injection uses init container to copy javaagent jar from init container image to a volume that is mounted to the application container
* the configuration of the instrumentation and SDK is done via environment variables

#### CR
```yaml
apiVersion: opentelemetry.io/v1alpha1
kind: OpenTelemetryInst
metadata:
  name: java-instrumentation
spec:
  exporter:
    endpoint: http://otel-collector:1
  java:
    image: ghcr.io/pavolloffay/otel-javaagent:1.5.3
```

#### Follow up PRs
* Support better service name
* Inject k8s resources - namespace, cluster etc.
* add sampling configuration to CR
* add support for other languages
* add status .e.g. show which workloads are instrumented and what is the sampling configuration...
* add flag to set default javaagent image
* add e2e test

PR #475, opened 2021-10-27
Publish a docker image containing Java autoinstrumentation files.

anuraaga:
To allow auto-instrumentation from the operator, we need to package the actual instrumentation into a docker image for copying into the user's container. This may be a java agent, or a directory of node_modules, or python pip installed packages, etc.

This adds `ghcr.io/open-telemetry/autoinstrumentation-java` as an image to contain the java instrumentation and is what would be used in conjunction with the pod mutator in https://github.com/open-telemetry/opentelemetry-operator/pull/464/files#diff-8792ce6e230133d6b38ced97a2fdbb6f6282a7c8e5f939cc7783521be7823232R59

Note that I discussed with the Java maintainers, and we didn't find there to be enough value in a base image for end users to consume to publish it there, so instead this is meant to be more of an internal detail of the operator. And it would make even less sense for something like Python, where end users just use pip anyways for example (javaagent users don't have such a package management approach in general making it slightly unique).

Ran the workflow on my fork to produce https://github.com/anuraaga/opentelemetry-operator/pkgs/container/opentelemetry-operator%2Fautoinstrumentation-java


pavolloffay:
@VineethReddy02 @jpkrohling could you please take a look and move this forward?

.github/workflows/publish-autoinstrumentation-java.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - uses: actions/checkout@v2.3.5
+
+      - name: Read version
+        run: echo "VERSION=$(cat autoinstrumentation/java/version.txt)" >> $GITHUB_ENV
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v3.6.0
+        with:
+          images: ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-java
```
jpkrohling, line 26: We should publish to quay, like the other images:

https://github.com/open-telemetry/opentelemetry-operator/blob/c3f411cb6f839443354a2039a0d1190d269147bd/.github/workflows/publish-images.yaml#L31-L39

autoinstrumentation/java/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM busybox
```
jpkrohling, line 1: We seem to be using this for the operator image:

https://github.com/open-telemetry/opentelemetry-operator/blob/c3f411cb6f839443354a2039a0d1190d269147bd/Dockerfile#L31

versions.txt
```diff
@@ -9,3 +9,7 @@ operator=0.37.0
 
 # Represents the current release of the Target Allocator.
 targetallocator=0.1.0
+
+# Represents the current release of Java instrumentation.
```
jpkrohling, line 13: Based on the discussion above, is this going to be removed?

VineethReddy02 (approved):
LGTM! I think after iterating over Juraci comments. We can go ahead with the merge :)

versions.txt
anuraaga: Think we need it here as well for the operator to reference.

.github/workflows/publish-autoinstrumentation-java.yaml
anuraaga: Switched to quay. I'd recommend separately considering migrating to GHCR for everything. For an OSS contributor project like OTel I think there is a lot to like about having creds / permissions tied to the repo itself rather than a backchannel

jpkrohling: Agree. Created https://github.com/open-telemetry/opentelemetry-operator/issues/492

pavolloffay:
@anuraaga the CI failed

pavolloffay:
This is the CI failure 

```bash
=== RUN   Test0_36_0Upgrade
    v0_36_0_test.go:119: 
        	Error Trace:	v0_36_0_test.go:119
        	Error:      	Not equal: 
        	            	expected: "upgrade to v0.36.0 has changed the tls_settings field name to tls in grpc protocol of otlp/mtls receiver"
        	            	actual  : "upgrade to v0.36.0 has changed the tls_settings field name to tls in http protocol of otlp/mtls receiver"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-upgrade to v0.36.0 has changed the tls_settings field name to tls in grpc protocol of otlp/mtls receiver
        	            	+upgrade to v0.36.0 has changed the tls_settings field name to tls in http protocol of otlp/mtls receiver
        	Test:       	Test0_36_0Upgrade
    v0_36_0_test.go:120: 
        	Error Trace:	v0_36_0_test.go:120
        	Error:      	Not equal: 
        	            	expected: "upgrade to v0.36.0 has changed the tls_settings field name to tls in http protocol of otlp/mtls receiver"
        	            	actual  : "upgrade to v0.36.0 has changed the tls_settings field name to tls in grpc protocol of otlp/mtls receiver"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-upgrade to v0.36.0 has changed the tls_settings field name to tls in http protocol of otlp/mtls receiver
        	            	+upgrade to v0.36.0 has changed the tls_settings field name to tls in grpc protocol of otlp/mtls receiver
        	Test:       	Test0_36_0Upgrade
--- FAIL: Test0_36_0Upgrade (0.00s)
```

Indeed not related to this PR.

pavolloffay:
Here is fix https://github.com/open-telemetry/opentelemetry-operator/pull/494

jpkrohling:
This PR needs a rebase, to get the fix for the flaky test.

pavolloffay:
@jpkrohling CI is green now

Issue #488, opened 2021-10-30
Update collector to v0.38.0


VineethReddy02:
Same as #466, but for 0.38.0. Changelog: https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.38.0

jpkrohling:
I changed the link to the changelog, as it's more relevant to have the one for core than for releases.

jpkrohling:
This might need to be in the upgrade mechanism: https://github.com/open-telemetry/opentelemetry-collector/pull/4213

VineethReddy02:
FYI, I created the upgrade routine PR. #491 

Now, all we need is to bump the Otelcol version PR. #466 includes all the changes that need to be made to this PR.

PR #491, opened 2021-11-02
Add upgrade routine for 0.38.0

VineethReddy02:
Part of: #488

### DIFF
--- a/pkg/collector/upgrade/v0_38_0.go
+++ b/pkg/collector/upgrade/v0_38_0.go
@@ -1,0 +1,46 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


pkg/collector/upgrade/v0_38_0.go
```diff
@@ -0,0 +1,46 @@
...
+	// return if args exist
+	if len(otelcol.Spec.Args) == 0 {
+		return otelcol, nil
+	}
+
+	// Remove otelcol args --log-level, --log-profile, --log-format
+	// are deprecated in reference to https://github.com/open-telemetry/opentelemetry-collector/pull/4213
+	var foundLoggingArgs []string
+	for argKey := range otelcol.Spec.Args {
+		if argKey == "--log-level" || argKey == "--log-profile" || argKey == "--log-format" {
```
jpkrohling, line 35: Can/should we translate that into config changes, to accomplish the same? I believe most, or all of those properties are now available in the collector config itself.

jpkrohling (approved):

pkg/collector/upgrade/v0_38_0.go
VineethReddy02: config changes? I do not see an option in otel collector configuration which includes args for otel collector container. :) @jpkrohling

jpkrohling: No, I mean that the args from the collector became options in the collector config: https://github.com/open-telemetry/opentelemetry-collector/pull/4009 . Like this:

```
service:
  telemetry:
    logs:
      level: "DEBUG"
      development: true
      encoding: "json"
```

jpkrohling (approved):
LGTM.

PR #499, opened 2021-11-03
Fix typos in Readme for auto-instrumentation

pavolloffay:

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,12 +156,9 @@ EOF
 
 ### OpenTelemetry auto-instrumentation injection
 
-The operator can inject and configure OpenTelemetry auto-instrumentation libraries. At this moment, the operator can inject only
-OpenTelemetry [Java auto-instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -168,11 +168,11 @@ instrumentation.opentelemetry.io/inject-java: "true"
...
 The value can be 
-* `false` - do not inject
-* `true` - inject
-* `java-instrumentation` - name of `Instrumentation` CR instance.
+* `"false"` - do not inject
+* `"true"` - inject and `Instrumentation` resource from the namespace.
+* `"java-instrumentation"` - name of `Instrumentation` CR instance.
 
-In the addition to the annotation the following `CR` has to be created. The `Instrumentation`
+In addition to the annotation, the following `CR` has to be created. The `Instrumentation` resource
```
jpkrohling, line 175: How about the line wrapping? The new content (the whole auto instrumentation section) should follow the same style as the other parts of the doc. Either change the new parts, or the old parts, but please keep it consistent. I personally prefer to not have forced line wraps, as it's easy to set a line wrapping size in any code editor and it makes editing easier.

pavolloffay: I don't any have preference as it is rendered the same way. I didn't know that there is a soft wrap feature and so I was always wrapping lines to make it always easily readable.

jpkrohling (approved):

PR #505, opened 2021-11-04
Generate docs from CRD spec

### LINKED ISSUE #325: Generate docs from CRD spec
Currently, we introduced a [docs file ](https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/otelcol_cr_spec.md)for details on otelcol custom resource spec this needs to be manually maintained for all changes on CRD spec. So we should find an efficient way to generate the docs from CRD.

iblancasa:
Closes #325

Signed-off-by: Israel Blancas <iblancas@redhat.com>


.github/workflows/continuous-integration.yaml
```diff
@@ -62,3 +62,23 @@ jobs:
...
+      - name: Clone repo
+        uses: actions/checkout@v2.4.0
+      - name: Setup go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17
+      - name: Run api-docs
+        run: |
+          make api-docs
+      - name: Check if working tree is dirty
```
jpkrohling, line 78: On second thought, I think we need to call `make api-docs` from within the `generate` command and use the `ensure-generate-is-noop` for this check.

iblancasa:
I applied some changes. I think this is what you are suggesting, right?

.github/workflows/continuous-integration.yaml
```diff
@@ -74,7 +74,7 @@ jobs:
           go-version: 1.17
       - name: Run api-docs
         run: |
-          make api-docs
+          make ensure-generate-is-noo api-docs
```
jpkrohling, line 77: This target name looks wrong. But my point was: if you have the `ensure-generate-is-noop` also check the location where the api-docs are expected to be generated, you don't need this extra step, as we check it elsewhere already. Look at the `ci` target.

iblancasa: Oh I see. Thanks!

.github/workflows/continuous-integration.yaml
```diff
@@ -74,11 +74,4 @@ jobs:
...
         run: |
-          make ensure-generate-is-noo api-docs
-      - name: Check if working tree is dirty
-        run: |
-          if [[ $(git status --porcelain) ]]; then
-            git diff
-            echo 'run make api-docs and commit changes'
-            exit 1
-          fi
+          make api-docs
```
jpkrohling, line 77: Almost! If you add `api-docs` as dependency for the `ci` job, you don't need to change this workflow at all :-)

Makefile
```diff
@@ -52,6 +52,7 @@ ensure-generate-is-noop: set-image-controller generate bundle
 	@git restore config/manager/kustomization.yaml
 	@git diff -s --exit-code api/v1alpha1/zz_generated.*.go || (echo "Build failed: a model has been changed but the generated resources aren't up to date. Run 'make generate' and update your PR." && exit 1)
 	@git diff -s --exit-code bundle config || (echo "Build failed: the bundle, config files has been changed but the generated bundle, config files aren't up to date. Run 'make bundle' and update your PR." && git diff && exit 1)
+	@git diff -s --exit-code docs/api.md || (echo "Build failed: the api.md file has been changed but the generated api.md file isn't up to date. Run 'make api-docs' and update your PR." && git diff && exit 1)
```
jpkrohling, line 55: \o/

pavolloffay (approved):

Makefile
```diff
@@ -227,3 +228,28 @@ bundle-build:
...
+
+# Find or download gen-crd-api-reference-docs
+gen-crd-api-reference-docs:
+ifeq (, $(shell which crdoc))
+	@{ \
+	set -e ;\
+	API_REF_GEN_TMP_DIR=$$(mktemp -d) ;\
+	cd $$API_REF_GEN_TMP_DIR ;\
+	go mod init tmp ;\
+	go get fybrik.io/crdoc@latest ;\
```
roee88, line 249: Consider pinning to a specific version rather than latest to avoid getting affected by breaking changes.
Also note that gen-crd-api-reference-docs is a different tool so the comment on this target is confusing.
(crdoc maintainer)

.github/workflows/continuous-integration.yaml
iblancasa: That's true. Good catch! Thanks

Makefile
iblancasa: Thanks for your comments. I used the same as [in this PR](https://github.com/grafana-operator/grafana-operator/pull/521/files) (suggested as a good example to follow in #325). What do you think about the new comment?

jpkrohling (approved):

Issue #460, opened 2021-10-18
Restart collector on config change


pavolloffay:
At the moment operator does not restart the collector when `spec.config` part of the CR changes (e.g. configure new processor). The operator could restart the collector if the config changes. 

The config update could be implemented on the collector itself, but that seems like a long way to go https://github.com/open-telemetry/opentelemetry-collector/issues/273.

BronzeDeer:
@bhiravabhatla I just encountered the same problem as @pavolloffay with a Deployment not updating. The Annotation needs to go on the pod spec template not the Deployment/Daemonset itself, otherwise kubernetes does not trigger a rolling update

bhiravabhatla:
@BronzeDeer  -> As far as I remember, I used annotations on deployment knowingly, whenever configmap is updated, we are publishing an event.

https://github.com/open-telemetry/opentelemetry-operator/blob/dd248a328fd020dc387df79ffba4ddde3a90598c/pkg/collector/reconcile/configmap.go#L178-L180

And deployment reconciller listens for configMap events - 

https://github.com/open-telemetry/opentelemetry-operator/blob/dd248a328fd020dc387df79ffba4ddde3a90598c/controllers/opentelemetrycollector_controller.go#L172

Reconciller should be triggered whenever there is a new event for configmap. Let me debug what is happening. Will update if i find something.

bhiravabhatla:
But I see your point even if deployment annotation is change, new replicaset wont be rolledout unless pod spec is not updated. May be that is the issue, not sure how missed this. Thanks for pointing it out.

bhiravabhatla:
Yep - this was miss. would make required changes and raise another PR. Apologies for the same. I guess we should add an e2e test for this as well. 

 @jpkrohling  and @pavolloffay  had a question, I see that we are adding prometheus scrape annotations to deployment.metadata.annotations.

https://github.com/open-telemetry/opentelemetry-operator/blob/dd248a328fd020dc387df79ffba4ddde3a90598c/pkg/collector/annotations.go#L29-L32

 Shouldn't they be in deployment.spec.template.metadata.annotations as well? If I am not wrong prometheus kuberenetes-pod job (default config) looks for annotations on pods not deployments.
 
 https://github.com/prometheus-community/helm-charts/blob/936d4712cda5ab28da304516cd7894c74067b63b/charts/prometheus/values.yaml#L1673-L1681

BronzeDeer:
[...]

PR #506, opened 2021-11-04
Fix rollout when the collector config changes

bhiravabhatla:

Use pod annotations instead of deployment annotations to store SHA256 for config. This would help restart opentelemetry collector whenever config is updated.

Fixes https://github.com/open-telemetry/opentelemetry-operator/issues/460

### DIFF
--- a/pkg/collector/annotations.go
+++ b/pkg/collector/annotations.go
@@ -37,6 +37,18 @@ func Annotations(instance v1alpha1.OpenTelemetryCollector) map[string]string {
 			annotations[k] = v
 		}
 	}
+	return annotations
+}
+
+func PodAnnotations(instance v1alpha1.OpenTelemetryCollector) map[string]string {
+	// new map every time, so that we don't touch the instance's annotations
+	annotations := map[string]string{}
+
+	if nil != instance.Spec.PodAnnotations {
+		for k, v := range instance.Spec.PodAnnotations {
+			annotations[k] = v
+		}
+	}
 	// make sure sha256 for configMap is always calculated
 	annotations["opentelemetry-operator-config/sha256"] = getConfigMapSHA(instance.Spec.Config)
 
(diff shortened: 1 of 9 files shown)


jpkrohling:
The e2e tests seem to be failing. 

@pavolloffay, once this is ready, can you review it?

bhiravabhatla:
Fixed failing test @jpkrohling  @pavolloffay

pkg/collector/annotations.go
```diff
@@ -37,13 +37,28 @@ func Annotations(instance v1alpha1.OpenTelemetryCollector) map[string]string {
...
 	// make sure sha256 for configMap is always calculated
 	annotations["opentelemetry-operator-config/sha256"] = getConfigMapSHA(instance.Spec.Config)
 
 	return annotations
 }
 
 func getConfigMapSHA(config string) string {
-	h := sha256.Sum256([]byte(config))
-	return fmt.Sprintf("%x", h)
+	if config != "" {
```
pavolloffay, line 59: any reason why there is an empty check? 

what retunrs the hash func on an empty string?

pkg/collector/annotations_test.go
```diff
@@ -42,7 +42,6 @@ func TestDefaultAnnotations(t *testing.T) {
 	assert.Equal(t, "true", annotations["prometheus.io/scrape"])
 	assert.Equal(t, "8888", annotations["prometheus.io/port"])
 	assert.Equal(t, "/metrics", annotations["prometheus.io/path"])
-	assert.Equal(t, "9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08", annotations["opentelemetry-operator-config/sha256"])
```
pavolloffay, line 45: why is this removed?

pavolloffay (commented):
I would suggest adding an e2e test for this:
1. deploy collector as deployment
2. assert it is deployed
3. change config in CR - e.g. add new receiver
4. assert the collector pod restarted with the new changes - check if the new receiver port is exposed.

VineethReddy02:
@bhiravabhatla can we move this PR forward? Also, do you mind replying back to the review comments?

Thanks!

PR #507, opened 2021-11-05
Add autoinstrumentation of NodeJS

anuraaga:
@pavolloffay @jpkrohling


anuraaga:
I think this is basically code complete for a first version, I will add docs but want to confirm the approach for supporting multiple languages. I went with a language annotation as it seemed the most straight forward among a few options I thought of such as changing the language specs to pointers and apply all specs that are defined (assuming there should only be one in practice). But let me know if this makes sense or of any better ideas.

jpkrohling:
I'm really excited about this one, thank you for the PR!

There are a couple of tests failing:

```
=== RUN   TestMutatePod/missing_language
    podmutator_test.go:450: 
        	Error Trace:	podmutator_test.go:450
        	Error:      	Received unexpected error:
        	            	instrumentation.opentelemetry.io/language must be set to the desired SDK language
        	Test:       	TestMutatePod/missing_language
=== RUN   TestMutatePod/missing_annotation
=== RUN   TestMutatePod/annotation_set_to_false
=== RUN   TestMutatePod/annotation_set_to_non_existing_instance
    podmutator_test.go:453: 
        	Error Trace:	podmutator_test.go:453
        	Error:      	"instrumentation.opentelemetry.io/language must be set to the desired SDK language" does not contain "instrumentations.opentelemetry.io \"doesnotexists\" not found"
        	Test:       	TestMutatePod/annotation_set_to_non_existing_instance
--- FAIL: TestMutatePod (0.07s)
    --- PASS: TestMutatePod/javaagent_injection,_true (0.03s)
    --- PASS: TestMutatePod/nodejs_injection,_true (0.01s)
    --- FAIL: TestMutatePod/missing_language (0.00s)
    --- PASS: TestMutatePod/missing_annotation (0.01s)
    --- PASS: TestMutatePod/annotation_set_to_false (0.01s)
    --- FAIL: TestMutatePod/annotation_set_to_non_existing_instance (0.00s)
```

As the original author of the auto-instrumentation for Java, I'll let @pavolloffay review this one.

pkg/instrumentation/annotation.go
```diff
@@ -23,15 +23,16 @@ import (
 const (
 	// annotationInjectJava indicates whether java auto-instrumentation should be injected or not.
 	// Possible values are "true", "false" or "<Instrumentation>" name.
-	annotationInjectJava = "instrumentation.opentelemetry.io/inject-java"
```
pavolloffay, line 26: can you explain the intentions for changing this annotation?

pavolloffay (commented):
I would like to know why the annotations changes. Could you do this change in a separate PR?

pkg/instrumentation/annotation.go
anuraaga: Yeah I'm happy to split out the annotation change if it makes sense.

I found our current, SDK-agnostic (?) sdk.go uses this Java annotation. I need it to be able to support multiple languages and came up with this change. Any idea that could work better?

pavolloffay: Could you please explain why we need two annotations? 

We can have `"instrumentation.opentelemetry.io/inject-java"`, `"instrumentation.opentelemetry.io/inject-node"` and then `"instrumentation.opentelemetry.io/inject" - to just inject the basic configuration (users could use the operator as a control plane to manage instrumentations).

anuraaga: I don't think we want to allow multiple languages on the same pod as that shouldn't ever be a use case I think. While we could check the annotations and fail if there are multiple this split seemed to model more directly that only one language can be injected.

I made language required without thinking too much but could definitely remove that to allow the control plane only injection if this otherwise makes sense.

pavolloffay: I don't see much value of limiting it to only a single language, quite the opposite. 

What I don't like is to have two annotations from the UX perspective. 

```
	annotationInject   = "instrumentation.opentelemetry.io/inject"
	annotationLanguage = "instrumentation.opentelemetry.io/language"
```

anuraaga: I believe these sort of annotations are always copy-pasted or frameworked, so I don't think there is a usability difference between the two patterns. If we think there is a use case for multiple annotations, then it makes sense but for example having both `otel.inject=true` and `otel.inject-java=true` on the same pod would be redundant and I think that leads to some cognitive dissonance.

Great to hear everyone's ideas on the matter :)

anuraaga: Discussed with @pavolloffay offline and for now sticking to the current annotation format, but we'll need to address it soon. We can see the awkwardness here

https://github.com/open-telemetry/opentelemetry-operator/pull/507/files#diff-92f790ce871022e5e7de195a2f96235dac6ac3fb04337974f8584c7c020a67cbR48

where common config can be injected from different instrumentations in a surprising way. Having separate annotations for the instrumentation config and what languages to use in the future will hopefully solve that.

README.md
```diff
@@ -193,6 +196,42 @@ EOF
...
+  name: nodejs-instrumentation
+spec:
+  exporter:
+    endpoint: http://otel-collector:4317
+  propagators:
+    - tracecontext
+    - baggage
+    - b3
+  nodejs:
+    image: ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-nodejs:latest
```
anuraaga, line 229: @pavolloffay I didn't document this image like you did in Java since I think it's too complicated to create oneself, at least unless someone asks for that.

anuraaga:
Sorry for the churn - I filed a couple of issues for some of the challenges I run into while developing here. But this should finally be ready @pavolloffay @jpkrohling

README.md
pavolloffay: I think, that is fine. I do not expect many people providing custom builds.

I hoped that vendors could use this operator with their agent distros in the future.

README.md
```diff
@@ -193,6 +196,42 @@ EOF
...
+The value can be
+* `"false"` - do not inject
+* `"true"` - inject and `Instrumentation` resource from the namespace.
+* `"nodejs-instrumentation"` - name of `Instrumentation` CR instance.
+
+In addition to the annotation, the following `CR` has to be created. The `Instrumentation` resource provides configuration for OpenTelemetry SDK and auto-instrumentation.
+
+```yaml
+kubectl apply -f - <<EOF
+apiVersion: opentelemetry.io/v1alpha1
```
pavolloffay, line 217: I think we should have only a single instrumentation CR in the readme and then have a bulletoints/section for each language with annotations.

pkg/instrumentation/annotation.go
pavolloffay: +1 let's discuss it in a separate issue. maybe there are other ways we can mitigate the issue

tests/e2e/instrumentation-nodejs/01-install-app.yaml
```diff
@@ -0,0 +1,20 @@
...
+    metadata:
+      labels:
+        app: my-pod-with-sidecar
+      annotations:
+        sidecar.opentelemetry.io/inject: "true"
+        instrumentation.opentelemetry.io/inject-nodejs: "true"
+    spec:
+      containers:
+      - name: myapp
+        image: ghcr.io/anuraaga/express-hello-world:latest
```
pavolloffay, line 20: We should probably create an application that would call an API on our test apps (java, js) that would result in a trace. 
Then report the trace to an inmemory storage and asser it exists (e.g. what we do in the java auto-instrumentation tests).

anuraaga: Yup think we will want this

README.md
anuraaga: Tweaked the README - hoping to be able to follow up with additional doc nits since it's been a big struggle keeping up with merge conflicts.

pavolloffay: Agree I had to rebase yesterday like 100k times

README.md
```diff
@@ -188,12 +177,42 @@ spec:
     argument: "0.25"
   java:
     image: ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-java:latest # <1>
-EOF
+  nodejs:
+    image: ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-nodejs:latest
+  EOF
```
pavolloffay, line 182: there are empty spaces which should not be there

README.md
```diff
@@ -188,12 +177,42 @@ spec:
...
+#### NodeJS
+
+The operator can inject OpenTelemetry [NodeJS auto-instrumentation](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/metapackages/auto-instrumentations-node).
+The injection of the NodeJS instrumentation can be enabled by adding an annotation to the namespace, so that all pods within that namespace will get the instrumentation, or by adding the annotation to individual PodSpec objects, available as part of Deployment, Statefulset, and other resources.
+
+```bash
+instrumentation.opentelemetry.io/inject-nodejs: "true"
+```
+
+The value can be
```
pavolloffay, line 212: I think we should DRY this, the semantics for the annotation are the same for all languages, no need to repeat this.

README.md
```diff
@@ -188,12 +177,42 @@ spec:
...
     image: ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-java:latest # <1>
-EOF
+  nodejs:
+    image: ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-nodejs:latest
+  EOF
 ```
 
+The above CR can be queried by `kubectl get otelinst`.
+
 1. Container image with [OpenTelemetry Java auto-instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation). The image must contain the Java agent JAR `/javaagent.jar`, and the operator will copy it to a shared volume mounted to the application container.
```
pavolloffay, line 187: Let's delete this and consider it as impl detail for now

pavolloffay (approved):

jpkrohling (approved):
I'm approving based on @pavolloffay's approval.

PR #508, opened 2021-11-05
Add docker image containing nodejs auto instrumentation

anuraaga:
Created image: https://github.com/anuraaga/opentelemetry-operator/pkgs/container/opentelemetry-operator%2Fautoinstrumentation-nodejs

I think the autoinstrumentation wrapper could be pushed to js-contrib though we'll still need to package something with Docker. Hopefully it's OK to include it here for now to try out the autoinstrumentation experience.

@pavolloffay @jpkrohling 

/cc @dyladan 

Would be used by https://github.com/open-telemetry/opentelemetry-operator/pull/507

### DIFF
--- a/autoinstrumentation/nodejs/src/autoinstrumentation.ts
+++ b/autoinstrumentation/nodejs/src/autoinstrumentation.ts
@@ -1,0 +1,12 @@
+import { getNodeAutoInstrumentations } from '@opentelemetry/auto-instrumentations-node';
+import { OTLPTraceExporter } from '@opentelemetry/exporter-otlp-grpc';
+
+import { NodeSDK } from '@opentelemetry/sdk-node';
+
+const sdk = new NodeSDK({
+    autoDetectResources: true,
+    instrumentations: [getNodeAutoInstrumentations()],
+    traceExporter: new OTLPTraceExporter(),
+});
+
+sdk.start();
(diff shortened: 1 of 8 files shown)


autoinstrumentation/nodejs/src/autoinstrumentation.ts
```diff
@@ -0,0 +1,12 @@
+import { getNodeAutoInstrumentations } from '@opentelemetry/auto-instrumentations-node';
+import { OTLPTraceExporter } from '@opentelemetry/exporter-otlp-grpc';
```
anuraaga, line 2: From what I understand NodeJS doesn't support env vars like OTEL_TRACES_EXPORTER yet. When it does, we would remove this sort of hard-coding.

jpkrohling: Should this be an issue on this repository then?

jpkrohling (commented):
I have no idea about the nodejs parts, but the github workflow is looking good. I will wait for a review from someone who knows about the nodejs parts before merging.

pavolloffay:
@anuraaga how does the injection of nodeJS instrumentation to an application work? Could you please briefly describe here what the operator will have to do?

anuraaga:
@pavolloffay The code is in #507 :) But if it would help to also have text explanation let me know

pavolloffay (approved):

jpkrohling (approved):

PR #512, opened 2021-11-05
Fix project structure to multiple kinds

pavolloffay:
At the moment the project structure is correct in regards that we have multiple kinds and the tolling does not work properly (e.g. use `operator-sdk` to generate new kind or webhook/controller).

First changing this repo to multigroup
```bash
kubebuilder edit --multigroup=true
```

Then adding metadata to `PROJECT` for the instrumentation kind.


jpkrohling (approved):

jpkrohling:
I believe @VineethReddy02 is releasing 0.38.0 today, let's merge this right afterward so that we have enough time before the 0.39.0 release to test it.

pavolloffay:
@jpkrohling there is no functional change here. It is just for tooling to allow us to generate new webhooks, controllers and ultimately add new kinds.

PR #514, opened 2021-11-08
Add sampler configuration to instrumentation kind

pavolloffay:

Related to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration


pavolloffay:
@anuraaga would you like to review this one?

pkg/instrumentation/sdk.go
```diff
@@ -61,6 +63,24 @@ func injectCommonSDKConfig(otelinst v1alpha1.Instrumentation, pod corev1.Pod) co
 		})
 	}
 
+	idx = getIndexOfEnv(container.Env, envOTELTracesSampler)
+	// configure sampler only if it is configured in the CR
+	if idx == -1 && otelinst.Spec.Sampler.Type != "" {
+		idxSamplerArg := getIndexOfEnv(container.Env, envOTELTracesSamplerArg)
+		if idxSamplerArg == -1 {
```
anuraaga, line 70: Hmm - given we know that there is no arg for the default sampler, parent_based_always_on, I think we can be a bit more agressive here and ignore the user's SAMPLER_ARG even if it's set. Or either way, it's probably worth to log a message as it seems like a user error

anuraaga (approved):

jpkrohling:
This needs a rebase.

pkg/instrumentation/sdk.go
pavolloffay: I don't think it's worth handling this case. If a user accidentally adds an argument for the always_on it's gonna by ignored by the SDK anyways.

pavolloffay:
rebased

VineethReddy02 (approved):

PR #515, opened 2021-11-08
Add propagators configuration to the Instrumentation kind

pavolloffay:

### DIFF
--- a/apis/instrumentation/v1alpha1/instrumentation_types.go
+++ b/apis/instrumentation/v1alpha1/instrumentation_types.go
@@ -28,6 +28,11 @@ type InstrumentationSpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
 	ResourceAttributes map[string]string `json:"resourceAttributes,omitempty"`
 
+	// Propagators defines inter-process context propagation configuration.
+	// +optional
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	Propagators []Propagator `json:"propagators,omitempty"`
+
 	// Java defines configuration for java auto-instrumentation.
 	// +optional
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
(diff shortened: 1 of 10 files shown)


anuraaga (approved):

jpkrohling:
This needs a rebase.

apis/instrumentation/v1alpha1/instrumentation_types.go
```diff
@@ -28,6 +28,11 @@ type InstrumentationSpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
 	ResourceAttributes map[string]string `json:"resourceAttributes,omitempty"`
 
+	// Propagators defines inter-process context propagation configuration.
+	// +optional
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	Propagators []string `json:"propagators,omitempty"`
```
jpkrohling, line 34: Isn't there an annotation that generates some validation of the possible values? I know we have that for enums, but I wonder if there's something for a list of values?

pkg/instrumentation/sdk.go
```diff
@@ -82,9 +83,11 @@ func injectCommonSDKConfig(otelinst v1alpha1.Instrumentation, ns corev1.Namespac
 	}
 	idx = getIndexOfEnv(container.Env, envOTELPropagators)
 	if idx == -1 && len(otelinst.Spec.Propagators) > 0 {
+		specialScopes := *(*[]string)((unsafe.Pointer(&otelinst.Spec.Propagators)))
```
jpkrohling: This now has the potential to break the operator: could you please add some negative tests? Perhaps isolate this in its own func to facilitate testing.

pavolloffay: > This now has the potential to break the operator

Could you please explain why it has the potential to break the operator?

jpkrohling: I need to play around, but you have a few conversions and pointer operations happening there. If you are confident that they are safe with all possible types of inputs, then I'm fine with it.

pavolloffay: I am fine with this.

You can do the exercise and get back here if this is risky.

jpkrohling (approved):

PR #516, opened 2021-11-08
Fix port in Instrumentation exporter to 4317 to use grpc

pavolloffay:

The GRPC exporter uses 4317 port not 4318.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ metadata:
   name: java-instrumentation
 spec:
   exporter:
-    endpoint: http://otel-collector:4318
+    endpoint: http://otel-collector:4317
   java:
     image: ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-java:latest # <1>
 EOF
--- a/pkg/instrumentation/sdk_test.go
+++ b/pkg/instrumentation/sdk_test.go
@@ -37,7 +37,7 @@ func TestSDKInjection(t *testing.T) {
 			inst: v1alpha1.Instrumentation{
 				Spec: v1alpha1.InstrumentationSpec{
 					Exporter: v1alpha1.Exporter{
-						Endpoint: "https://collector:4318",
+						Endpoint: "https://collector:4317",
 					},
 				},
 			},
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -179,7 +179,7 @@ metadata:
   name: java-instrumentation
 spec:
   exporter:
-    endpoint: http://otel-collector:4318
+    endpoint: http://otel-collector:4317
```
jpkrohling, line 182: It would probably be less confusing if it's without the HTTP protocol, just `otel-collector:4317`.

pavolloffay: Unfortunatelly the spec mandates to use http (to indicate that it is insecure connection)

jpkrohling: No, it requires only for OTLP HTTP, not for OTLP gRPC:

> The endpoint SHOULD accept any form allowed by the underlying gRPC client implementation. Additionally, the endpoint MUST accept a URL with a scheme of either http or https. A scheme of https indicates a secure connection and takes precedence over the insecure configuration setting. If the gRPC client implementation does not support an endpoint with a scheme of http or https then the endpoint SHOULD be transformed to the most sensible format for that implementation.

There has been a discussion in the past (collector SIG? spec?) about gRPC clients that do not accept the protocol part (like the Go one). I'm surprised that the default by the spec still includes the protocol part, but note that the spec requires the library to parse the value and remove the protocol if the underlying gRPC library doesn't support it.

pavolloffay: I will have a look at removal of the protocol in a separate PR, I will have to retest if it works. In the past I have noticed there were differences how SDKs interpret the endpint. 

Most importantly the port 4318 does not work and this is what the PR fixes

pavolloffay: > Target to which the exporter is going to send spans or metrics. The endpoint SHOULD accept any form allowed by the underlying gRPC client implementation. Additionally, the endpoint MUST accept a URL with a scheme of either http or https. A scheme of https indicates a secure connection and takes precedence over the insecure configuration setting. If the gRPC client implementation does not support an endpoint with a scheme of http or https then the endpoint SHOULD be transformed to the most sensible format for that implementation.

From https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md

jpkrohling: That's the same that I quoted before :-) Note that it has different wording than the HTTP exporter, which *requires* the protocol part to be specified.

jpkrohling (approved):

jpkrohling:
I'm merging this as is but do test the gRPC connection without the protocol part: it has to work.

PR #517, opened 2021-11-08
Change API group for instrumentation add defaulting webhook

pavolloffay:
Notable changes:
* changes the group for instrumentation from `opentelemetry.io` to `instrumentation.opentelemetry.io`. This is required if we want to keep decoupled collector and instrumentation kinds.
* Adds defaulting webhook (empty for now)
* Changes project medatada to `go.kubebuilder.io/v3` which is the kubebuilder version used by the `operator-sdk` tool we are using. After this fix `operator-sdk` is able to correctly set labels/metadata and other fields in this project. 
* Added scorecard tests - now with the correct project metadata the bundle/generate adds scorecard annotations.

The following was used to generate the webhook. This worked after the group and kubebuilder metadata change.
```bash
operator-sdk create webhook  --version v1alpha1 --kind Instrumentation --group instrumentation  --defaulting --programmatic-validation 
```

Note that all major changes to the project should come from the `operator-sdk` scaffolding otherwise upgrades of the tooling do not work properly. 

My notes:
* k8s core groups https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/
* `kubectl api-resources -o wide`


pavolloffay:
This is an alternative solution to https://github.com/open-telemetry/opentelemetry-operator/pull/518. But I prefer https://github.com/open-telemetry/opentelemetry-operator/pull/518 bc it is simpler and does not require the webhook.

Makefile
```diff
@@ -99,6 +99,10 @@ release-artifacts: set-image-controller
 manifests: controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 
+# Generate code
+generate: controller-gen
```
pavolloffay, line 103: moving it here to be closer to the template project https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v3/memcached-operator/Makefile 

 + this allows me to see the difference between manifests and generate target

pavolloffay:
@anuraaga take a look at this PR, especially on the API group change for the instrumentation

bundle/manifests/instrumentation.opentelemetry.io_instrumentations.yaml
```diff
@@ -4,9 +4,9 @@ metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.6.0-beta.0
   creationTimestamp: null
-  name: instrumentations.opentelemetry.io
+  name: instrumentations.instrumentation.opentelemetry.io
```
jpkrohling, line 7: No, this is too verbose. It should really be `instrumentations.opentelemetry.io`.

jpkrohling:
>This is required if we want to keep decoupled collector and instrumentation kinds.

How's that? We can definitely have the same apiVersion and different kinds...

pavolloffay:
that is right kinds from the same group can be versioned separately. This https://github.com/open-telemetry/opentelemetry-operator/pull/533 fixes the problem with generating the project.

PR #518, opened 2021-11-08
Set default auto-instrumentation java image

pavolloffay:

This PRs set default auto-instrumentation java image if the image is not specified in the CR. The operator does the same for OTELcol and allocator images.

### DIFF
--- a/pkg/instrumentation/podmutator.go
+++ b/pkg/instrumentation/podmutator.go
@@ -34,8 +34,9 @@ var (
 )
 
 type instPodMutator struct {
-	Logger logr.Logger
-	Client client.Client
+	Logger                  logr.Logger
+	Client                  client.Client
+	autoInstrumentationJava string
 }
 
 type languageInstrumentations struct {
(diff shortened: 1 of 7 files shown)


config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
-  - manager.yaml
+- manager.yaml
```
VineethReddy02, line 2: These changes shouldn't be committed. AFAIK they get changes during the local build, deployment.

pavolloffay:
This is ready for review

VineethReddy02 (approved):
LGTM!
@jpkrohling can you take a final look?

pkg/instrumentation/podmutator.go
```diff
@@ -111,3 +114,10 @@ func (pm *instPodMutator) selectInstrumentationInstanceFromNamespace(ctx context
 		return otelInsts.Items[0], nil
 	}
 }
+
+func (pm *instPodMutator) applyDefaults(otelinst v1alpha1.Instrumentation) v1alpha1.Instrumentation {
+	if otelinst.Spec.Java.Image == "" {
```
jpkrohling, line 119: This should be part of the Defaulter.

https://github.com/open-telemetry/opentelemetry-operator/blob/de6219f576b1af2e88c9119452ae8ef1db67cb7a/apis/collector/v1alpha1/opentelemetrycollector_webhook.go#L39-L55

pavolloffay: Collector image is not handled in the defaulter either. 

That said I might agree with you (I need to test the defaulter, when it comes to updates).  I have this WIP PR https://github.com/open-telemetry/opentelemetry-operator/pull/517 that might change this to use the defaulter.

jpkrohling: The collector image was probably handled before we implemented the defaulter.

pavolloffay: I would like to get this merged and then change just remove this function in the #517 PR.

PR #524, opened 2021-11-10
Create Docker image for Python auto-instrumentation.

anuraaga:
Created https://github.com/anuraaga/opentelemetry-operator/pkgs/container/opentelemetry-operator%2Fautoinstrumentation-python with it

I will write up the PR for the operator change after #507 is in. The mechanism, which I verified locally, is

`PYTHONPATH=/otel-autoinstrumentation/opentelemetry/instrumentation/auto_instrumentation:$PYTHONPATH:/otel-autoinstrumentation`

The trick is that the first entry is pointing to the folder with OTel's `sitecustomize.py` which gets automatically executed.

This is the same as what the `opentelemetry-instrument` script does I think except also appends the opentelemetry libraries to pythonpath, which otherwise is usually pip installed.

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py#L105

### DIFF
--- a/autoinstrumentation/python/Dockerfile
+++ b/autoinstrumentation/python/Dockerfile
@@ -1,0 +1,11 @@
+FROM python:3.10-alpine AS build
+
+WORKDIR /operator-build
+
+ADD requirements.txt .
+
+RUN mkdir workspace && pip install --target workspace -r requirements.txt
+
+FROM busybox
+
+COPY --from=build /operator-build/workspace /autoinstrumentation
--- a/autoinstrumentation/python/requirements.txt
+++ b/autoinstrumentation/python/requirements.txt
@@ -1,0 +1,43 @@
+opentelemetry-distro==0.25b2
+# We don't use the distro[otlp] option which automatically includes exporters since gRPC is not appropriate for
+# injected auto-instrumentation, where it has a strict dependency on the OS / Python version the artifact is built for.
+opentelemetry-exporter-otlp-proto-http==1.6.2
+
+opentelemetry-instrumentation==0.25b2
+
+# Copied in from https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation
+# except for aws-lambda
+opentelemetry-instrumentation-aiohttp-client==0.25b2
+opentelemetry-instrumentation-aiopg==0.25b2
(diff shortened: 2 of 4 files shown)


anuraaga:
@pavolloffay @jpkrohling 

and @owais I guess this one you might be particularly interested in ;)

autoinstrumentation/python/requirements.txt
```diff
@@ -0,0 +1,43 @@
+opentelemetry-distro==0.25b2
+# We don't use the distro[otlp] option which automatically includes exporters since gRPC is not appropriate for
```
anuraaga, line 2: I made this decision, but not sure on it. @owais In particular perhaps you have some advice on what we should use here in our "autoinstrumentation distro". The somewhat annoying aspect is that the operator will have to specify `OTEL_TRACES_EXPORTER` for the user.

autoinstrumentation/python/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM python:3.10-alpine AS build
+
+WORKDIR /operator-build
+
+ADD requirements.txt .
+
+RUN mkdir workspace && pip install --target workspace -r requirements.txt
+
+FROM busybox
```
pavolloffay, line 9: I was wondering why we are using `busybox` as a base image. Other operator images are based on `gcr.io/distroless/static:nonroot`. If there is no reason I would prefer to be consistent.

anuraaga: The distroless image doesn't contain the `cp` command which we need.

autoinstrumentation/python/requirements.txt
owais: I think using proto http sounds reasonable if you can get away with it. I was working on a similar feature at some point and had decided to build one image per platform/architecture.

anuraaga:
e2e in #532 will fail until this is merged

anuraaga:
Don't think anything is blocking this - hopefully can merge and I'll wrap up 532

pavolloffay (approved):

jpkrohling (approved):

PR #528, opened 2021-11-11
Delete namespaces created in tests.

### LINKED ISSUE #523: Clean up namespaces after unit test run
I suspect this affects `make test` too though I'm currently having trouble with that command and have only checked with running tests in GoLand directly, including with debugger. It seems that namespaces get created and not deleted, and then tests cannot be re-ran until I manually delete the namespaces. The tests should have a cleanup hook to delete namespaces, regardless of test success or failure, to allow rerunning easily.

anuraaga:
This allows tests to be run multiple times without needing to manually delete the namespaces, or recreate the cluster, between runs.

Fixes #523

### DIFF
--- a/internal/webhookhandler/webhookhandler_test.go
+++ b/internal/webhookhandler/webhookhandler_test.go
@@ -126,6 +126,7 @@ func TestShouldInjectSidecar(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			err := k8sClient.Create(context.Background(), &tt.ns)
 			require.NoError(t, err)
+			defer k8sClient.Delete(context.Background(), &tt.ns)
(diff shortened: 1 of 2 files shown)


jpkrohling (approved):

pavolloffay:
@anuraaga lint failed 

```
run golangci-lint
  Running [/home/runner/golangci-lint-1.29.0-linux-amd64/golangci-lint run --out-format=github-actions -v] in [] ...
  Error: Error return value of `k8sClient.Delete` is not checked (errcheck)
  Error: Error return value of `k8sClient.Delete` is not checked (errcheck)
  Error: Error return value of `k8sClient.Delete` is not checked (errcheck)
```

PR #532, opened 2021-11-11
Add autoinstrumentation for Python

anuraaga:


anuraaga:
@pavolloffay @jpkrohling

pkg/instrumentation/sdk.go
```diff
@@ -58,6 +58,12 @@ func inject(logger logr.Logger, insts languageInstrumentations, ns corev1.Namesp
 		pod = injectCommonSDKConfig(otelinst, ns, pod)
 		pod = injectNodeJSSDK(logger, otelinst.Spec.NodeJS, pod)
 	}
+	if insts.Python != nil {
+		otelinst := *insts.Python
+		logger.V(1).Info("injecting instrumentation into pod", "otelinst-namespace", otelinst.Namespace, "otelinst-name", otelinst.Name)
```
pavolloffay, line 63: it would be good to include python in the log message

pkg/instrumentation/python.go
```diff
@@ -0,0 +1,82 @@
...
+	} else if idx > -1 {
+		if container.Env[idx].ValueFrom != nil {
+			// TODO add to status object or submit it as an event
+			logger.Info("Skipping Python SDK injection, the container defines PYTHONPATH env var value via ValueFrom", "container", container.Name)
+			return pod
+		}
+		container.Env[idx].Value = fmt.Sprintf("%s:%s:%s", pythonPathPrefix, container.Env[idx].Value, pythonPathSuffix)
+	}
+
+	// Set OTEL_TRACES_EXPORTER to HTTP exporter if not set by user because it is what our autoinstrumentation supports.
```
pavolloffay, line 51: python does not support OTLP GRPC?

anuraaga: It does but uses a native extension wrapping gRPC-C. This means the extension can only be loaded on the same python version as when pip install was invoked.

HTTP transport is pure python so avoids causing potential version problems.

anuraaga: By the way related comment here

https://github.com/open-telemetry/opentelemetry-operator/pull/524#discussion_r746518160

pkg/instrumentation/sdk.go
anuraaga: Thanks, fixed

pavolloffay (approved):

jpkrohling (approved):
Approving based on @pavolloffay's approval.

PR #533, opened 2021-11-11
Fix project structure to align with kubebuilder scaffolding

pavolloffay:
When working on the webhook - https://github.com/open-telemetry/opentelemetry-operator/pull/517 I could not get the generated webhook working:

```
operator-sdk create webhook  --version v1alpha1 --kind Instrumentation --group instrumentation  --defaulting --programmatic-validation 
```

The issue is that the `api` package structure. The API package follows this `apis/<group>/<version>`.  In our case the group is empty (do not confuse it with `opentelemetry.io` which is the domain!). If we had a group `collector` like the `PROJECT` file has  in the `main` branch then the CR for the  collector would be `opentelemetrycollectors.collector.opentelemetry.io` (see https://github.com/open-telemetry/opentelemetry-operator/pull/517#discussion_r747340847).

We can either stick to to aligning to kubebuilder but then code generators will not work and we will have to manually edit generated code. In my case I spend I day fighting why the generated webhook does not work. 

That said I would prefer to fix the project structure to make the tooling work as expected.


PROJECT
```diff
@@ -1,23 +1,29 @@
...
-- controller: true
-  group: collector
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: opentelemetry.io
   kind: OpenTelemetryCollector
-  path: github.com/open-telemetry/opentelemetry-operator/apis/collector/v1alpha1
+  path: github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1
```
jpkrohling, line 16: Does this mean that we can only have one version for all CRDs? If we need to bump the one for instrumentation, would we need to bump the one for the collector as well?

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -43,10 +43,16 @@ spec:
   apiservicedefinitions: {}
   customresourcedefinitions:
     owned:
-    - kind: Instrumentation
+    - description: 'Instrumentation is the spec for OpenTelemetry instrumentation.
+        nolint: maligned'
```
jpkrohling, line 47: This line shouldn't be part of the CSV :-)

PROJECT
pavolloffay: no, then we change the version in `PROJECT` for a given kind and create `apis/v1` and move the kind there.

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
pavolloffay: removed the `nolint` annotation. I had to reshuffle the struct. It looks weird but there is no functional difference. Perhaps we can disable the `maligned` globally.

jpkrohling (approved):

pavolloffay:
@jpkrohling is the CI stuck?

PR #535, opened 2021-11-11
Add defaulter for instrumentation kind and set javaagent image

pavolloffay:


main.go
```diff
@@ -173,6 +174,14 @@ func main() {
 			setupLog.Error(err, "unable to create webhook", "webhook", "OpenTelemetryCollector")
 			os.Exit(1)
 		}
+		if err = (&otelv1alpha1.Instrumentation{
+			ObjectMeta: metav1.ObjectMeta{
+				Annotations: map[string]string{otelv1alpha1.AnnotationDefaultAutoInstrumentationJava: autoInstrumentationJava},
```
pavolloffay, line 179: This is sweet, user can set defaults even in the CR via annotations.

The update procedure (in a separate PR) will use the annotation and if the value matches the image it will update it along with the image.

pavolloffay: There are a couple of approaches how to pass default values from `main` to the API package. 

1. via annotation like this PR does. 
2.  add new private field to CR and use the private field in the webhook. Note that the upgrade procedure will not have access to the private field, hence the upgrade procedure will have to be initialized with default values as well.

```go
// Instrumentation is the spec for OpenTelemetry instrumentation.
type Instrumentation struct {
	Status            InstrumentationStatus `json:"status,omitempty"`
	metav1.ObjectMeta `json:"metadata,omitempty"`
	Spec              InstrumentationSpec `json:"spec,omitempty"`
	metav1.TypeMeta   `json:",inline"`

	autoInstrumentationJava string `json:"-"`
}

func DefaultInstrumentation(autoInstrumentationJava string) Instrumentation {
	return Instrumentation{
		autoInstrumentationJava: autoInstrumentationJava,
	}
}
```

3. define new type that is composed from `Instrumentation` and use it as webhook impl. I could not make this work
```
type InstrumentationWithDefaults struct {
	Instrumentation
       AutoInstrumentationJava string
}
```

VineethReddy02: I like the option 3

VineethReddy02 (approved):
LGTM overall!

VineethReddy02:
@pavolloffay Please create an issue for update procedure as described in the comment to the track the missing part :)

main.go
pavolloffay: The option 3, unfortunately didn't work.

pavolloffay:
I will open a PR right away.

PR #545, opened 2021-11-16
Upgrade auto-instrumentation java to 1.7.2

pavolloffay:


VineethReddy02 (approved):
LGTM! Is it okay in the future to jump multiple versions? If we miss upgrading multiple versions.

anuraaga (approved):
We'll ideally release at a similar cadence as the instrumentation but I don't think it's required in general.

pavolloffay:
I am splitting the PR into two:
1. publish the image
2. upgrade the default version in the operator

Issue #526, opened 2021-11-10
Pass all scorecard pass


pavolloffay:
https://github.com/open-telemetry/opentelemetry-operator/pull/517 added basic scorecard tests. 

We should add missing tests to align with https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v3/memcached-operator/bundle/tests/scorecard/config.yaml

pavolloffay:
Now two missing tests are `olm-spec-descriptors-test` and `olm-status-descriptors-test`. 

These tests check if the description for fields used in the alm-examples are in the CSV.

The descriptions can be added via `+operator-sdk:csv:customresourcedefinitions:type=[spec,status]` on a field. 

That said, I could not make this work for the instrumentation example, I have added the annotation to the exporter (the only fields used on the example) and the test still failed. On the other hand the test passed for the otelcol example that is using only `config` field.

Either the test is not properly implemented for object fields or the rendering of the description is flaky.

pavolloffay:
Link for the descriptors explanation: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/building-your-csv.md#your-custom-resource-definitions

pavolloffay:
I am closing this for now, the set of scorecard tests that are used in this repository is enough to asses the quality of the  CSV for the OLM.

PR #555, opened 2021-11-17
Add scorecard resources test

pavolloffay:

Updates #526 

See https://sdk.operatorframework.io/docs/testing-operators/scorecard/

### DIFF
--- a/apis/v1alpha1/opentelemetrycollector_types.go
+++ b/apis/v1alpha1/opentelemetrycollector_types.go
@@ -164,6 +164,8 @@ type OpenTelemetryTargetAllocatorSpec struct {
 // +kubebuilder:printcolumn:name="Version",type="string",JSONPath=".status.version",description="OpenTelemetry Version"
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
 // +operator-sdk:csv:customresourcedefinitions:displayName="OpenTelemetry Collector"
+// This annotation provides a hint for OLM which resources are managed by OpenTelemetryCollector kind.
+// It's not mandatory to list all resources.
 // +operator-sdk:csv:customresourcedefinitions:resources={{Pod,v1},{Deployment,apps/v1},{DaemonSets,apps/v1},{StatefulSets,apps/v1},{ConfigMaps,v1},{Service,v1}}
 
 // OpenTelemetryCollector is the Schema for the opentelemetrycollectors API.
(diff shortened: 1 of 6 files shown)


apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -164,6 +164,7 @@ type OpenTelemetryTargetAllocatorSpec struct {
 // +kubebuilder:printcolumn:name="Version",type="string",JSONPath=".status.version",description="OpenTelemetry Version"
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
 // +operator-sdk:csv:customresourcedefinitions:displayName="OpenTelemetry Collector"
+// +operator-sdk:csv:customresourcedefinitions:resources={{Pod,v1},{Deployment,v1},{Service, v1}}
```
jpkrohling, line 167: What does this mean? Resources that might be created as a result of this CR? If so, this needs to include at least DaemonSets and StatefulSets, and potentially ConfigMaps (among others?)

pavolloffay: yes, this does not have to be a complete list to pass the test. It will pop up in OLM to let the user know what the Kinds does

jpkrohling: If the goal is to make the test pass, then it sounds good. But the purpose is to signal which resources types are created, then we should really try to list them all here. Or leave a comment stating the criteria.

pavolloffay: The primary goal is to make the test to pass. I have added other objects that you suggested. My gut feeling says it will get outdated anyways, but it gives a good indicated to OLM and users what resources are managed by this kind.

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -164,7 +164,7 @@ type OpenTelemetryTargetAllocatorSpec struct {
 // +kubebuilder:printcolumn:name="Version",type="string",JSONPath=".status.version",description="OpenTelemetry Version"
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
 // +operator-sdk:csv:customresourcedefinitions:displayName="OpenTelemetry Collector"
-// +operator-sdk:csv:customresourcedefinitions:resources={{Pod,v1},{Deployment,v1},{Service, v1}}
+// +operator-sdk:csv:customresourcedefinitions:resources={{Pod,v1},{Deployment,apps/v1},{DaemonSets,apps/v1},{StatefulSets,apps/v1},{ConfigMaps,v1},{Service,v1}}
```
jpkrohling, line 167: Services? Service Accounts? Double-check the objects we create.

pavolloffay: service is there. again I don't want to list everything, see my previous comment

jpkrohling: It's fine to not list everything, just state the expectation so that your future self knows when to update this list.

jpkrohling (approved):

PR #556, opened 2021-11-17
Remove unused operator-sdk annotations from kinds

pavolloffay:

The removed annotations seem to be noop. The CSV/CRD does not change after removal.

### DIFF
--- a/apis/v1alpha1/instrumentation_types.go
+++ b/apis/v1alpha1/instrumentation_types.go
@@ -22,37 +22,30 @@ import (
 type InstrumentationSpec struct {
 	// Exporter defines exporter configuration.
 	// +optional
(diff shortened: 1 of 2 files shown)


jpkrohling (commented):
Are they not required anymore? Any sources of info? It would be a shame to remove it now only to have it fail later on when sending a PR to the operator hub.

pavolloffay:
I don't have any more info. The operator hub should consume the bundle that is generated in this repo and as we can see the removal of annotations does not change the bundle. 

e.g. `OpenTelemetryTargetAllocatorSpec` does not have this annotations and publish to the hub passed ;)

jpkrohling (approved):

PR #557, opened 2021-11-17
Refactor instrumentation types

pavolloffay:

* Removing to `Spec` from language types to align with other types
* moving the langulagte types below SDK config to align what the core spec order.


pavolloffay:
@jpkrohling this is a simple PR ready for review

jpkrohling:
> Removing to Spec from language types to align with other types

Why removing `Spec` from two types, instead of adding to the missing one? Why is `Spec` wrong?

pavolloffay:
Note that the top level type is `Instrumentation` and it embeds `InstrumentationSpec`. A similar concept is in core k8s  lie Pod and PodSpec, but then PodSpec embeds direct types and not `<whatever>Spec` (e.g. there is no  `VolumeSpec` but `Volume` ).

At the end of the day it is a personal preference and this PR represents mine with alignment to k8s naming convention.

jpkrohling (approved):

jpkrohling:
> this PR represents mine with alignment to k8s naming convention

If that's the case, I agree.

PR #558, opened 2021-11-17
[FIX]: adds case to handle kubeletstats reciever to ignore endpoint

### LINKED ISSUE #527: Add a mechanism to ignore certain receivers when deriving service from config
The specific receiver configuration that sparked this issue:

```
receivers:
      kubeletstats/node:
        auth_type: serviceAccount
        collection_interval: 1m
        endpoint: ${K8S_NODE_NAME}:10250
        insecure_skip_verify: true
        k8s_api_config:
          auth_type: serviceAccount
        metric_groups:
        - node
      kubeletstats/pods:
        auth_type: serviceAccount
        collection_interval: 1m
        endpoint: ${K8S_NODE_NAME}:10250
        insecure_skip_verify: true
        k8s_api_config:
          auth_type: serviceAccount
        metric_groups:
        - pod
```

This fails, because the service derives that there are two endpoints, each with port 10250, giving this error:
```
{"level":"error","ts":1636559029.2833467,"logger":"controllers.OpenTelemetryCollector","msg":"failed to reconcile services","error":"failed to reconcile the expected services: failed to apply changes: Service \"opentelemetry-node-agent-collector\" is invalid: spec.ports[1]: Duplicate value: core.ServicePort{Name:\"\", Protocol:\"TCP\", AppProtocol:(*string)(nil), Port:10250, TargetPort:intstr.IntOrString{Type:0, IntVal:0, StrVal:\"\"}, NodePort:0}","stacktrace":"github.com/open-telemetry/opentelemetry-operator/controllers.(*OpenTelemetryCollectorReconciler).Reconcile\n\t/workspace/controllers/opentelemetrycollector_controller.go:147\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.6/pkg/internal/controller/controller.go:298\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.6/pkg/internal/controller/controller.go:253\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.6/pkg/internal/controller/controller.go:214"}
```

The problem here is that `endpoint` doesn't define an endpoint for something connecting to the collector, rather it's an endpoint that is scraped by the collector. These ports shouldn't be mapped in the service, since they are not for inbound connections.

Currently the workaround for this is to not use `endpoint` for this specific scenario and have the receiver figure it out, but I could see this being a problem with other "scraping" receivers as well.

Ideally, there'd be a way to skip gathering ports from certain receivers, or even a way to disable the port auto-detection as a whole.

mritunjaysharma394:
This PR is an attempt  to fix #527 to ignore certain kubeletstats receivers when deriving service from config 
Signed-off-by: Mritunjay Sharma <mritunjaysharma394@gmail.com>

### DIFF
--- a/pkg/collector/parser/receiver.go
+++ b/pkg/collector/parser/receiver.go
@@ -100,6 +100,11 @@ func singlePortFromConfigEndpoint(logger logr.Logger, name string, config map[in
 	case name == "tcplog" || name == "udplog":
 		endpoint = getAddressFromConfig(logger, name, listenAddressKey, config)
 
+	// in case of kubeletstats reciever, the endpoint is ignored
+	case name == "kubeletstats":
+		endpoint = nil
+		return nil
+
 	default:
 		endpoint = getAddressFromConfig(logger, name, endpointKey, config)
 	}
--- a/pkg/collector/parser/receiver.go
+++ b/pkg/collector/parser/receiver.go
@@ -100,9 +100,10 @@ func singlePortFromConfigEndpoint(logger logr.Logger, name string, config map[in
 	case name == "tcplog" || name == "udplog":
 		endpoint = getAddressFromConfig(logger, name, listenAddressKey, config)
 
-	// in case of kubeletstats reciever, the endpoint is ignored
+	// ignore kubeletstats receiver as it holds the field key endpoint, and it
+	// is a scraper, we only expose endpoint through k8s service objects for
+	// receivers that aren't scrapers.
 	case name == "kubeletstats":
-		endpoint = nil
 		return nil
 
 	default:
--- a/pkg/collector/parser/receiver_test.go
+++ b/pkg/collector/parser/receiver_test.go
@@ -18,6 +18,7 @@ import (
 	"testing"
 
 	"github.com/go-logr/logr"
+	"github.com/open-telemetry/opentelemetry-operator/pkg/collector/parser"
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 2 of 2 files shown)


pkg/collector/parser/receiver.go
```diff
@@ -100,6 +100,11 @@ func singlePortFromConfigEndpoint(logger logr.Logger, name string, config map[in
 	case name == "tcplog" || name == "udplog":
 		endpoint = getAddressFromConfig(logger, name, listenAddressKey, config)
 
+	// in case of kubeletstats reciever, the endpoint is ignored
```
jpkrohling, line 103:
```suggestion
	// in the case of kubeletstats receiver, the endpoint is ignored
```

jpkrohling:
@VineethReddy02, could you please review this one as you are familiar with this code?

pkg/collector/parser/receiver.go
```diff
@@ -100,6 +100,11 @@ func singlePortFromConfigEndpoint(logger logr.Logger, name string, config map[in
...
+	case name == "kubeletstats":
+		endpoint = nil
```
VineethReddy02, line 105: All you have to do is `return nil` here. No use of `nil` assignment. Also, add unit tests for the same.

VineethReddy02 (changes requested):

pkg/collector/parser/receiver_test.go
```diff
@@ -98,6 +98,19 @@ func TestReceiverFailsWhenPortIsntString(t *testing.T) {
 	assert.Nil(t, p)
 }
 
+func TestReceiverFailsWhenNameIsKubeletStats(t *testing.T) {
```
VineethReddy02, line 101: I would do something like this with more comments on the test case and with a `endpoint` field in the test data that usually comes along with `kubeletstats` receiver. Also, will add additional checks like

```
func TestIgnorekubeletstatsEndpoint(t *testing.T) {
	// ignore "kubeletstats" receiver endpoint field, this is special case
	// as this receiver gets parsed by generic receiver parser
	builder := parser.NewGenericReceiverParser(logger, "kubeletstats", map[interface{}]interface{}{
		"endpoint": "0.0.0.0:9000",
	})

	// test
	ports, err := builder.Ports()

	// verify
	assert.NoError(t, err)
	assert.Len(t, ports, 0)
}
```

VineethReddy02 (changes requested):

VineethReddy02:
@mritunjaysharma394 can you re-iterate on the suggested changes?

mritunjaysharma394:
Hi @VineethReddy02 , thanks a lot for the review and suggestions, I will complete the suggested changes by tomorrow, sorry for the delay due to exams!

pkg/collector/parser/receiver_test.go
mritunjaysharma394: Hi @VineethReddy02, thanks a lot for all the help and this learning, have tried to modify the code according to suggestions, kindly review it at your convenient time, thanks a lot!

VineethReddy02 (approved):

PR #569, opened 2021-11-22
Extract component dependencies to a separate function

### LINKED ISSUE #568: Refactor bootstrap of upgrade routines
happy to do it in a separate refactoring PR

_Originally posted by @pavolloffay in https://github.com/open-telemetry/opentelemetry-operator/pull/548#discussion_r753086609_

pavolloffay:
Resolves #568 

### DIFF
--- a/main.go
+++ b/main.go
@@ -109,6 +109,7 @@ func main() {
 		config.WithVersion(v),
 		config.WithCollectorImage(collectorImage),
 		config.WithTargetAllocatorImage(targetAllocatorImage),
+		config.WithAutoInstrumentationJavaImage(autoInstrumentationJava),
 		config.WithAutoDetect(ad),
 	)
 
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -207,8 +187,40 @@ func main() {
...
 
 	setupLog.Info("starting manager")
-	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
+	if err := mgr.Start(ctx); err != nil {
 		setupLog.Error(err, "problem running manager")
 		os.Exit(1)
 	}
 }
+
+func addDependencies(_ context.Context, mgr ctrl.Manager, cfg config.Config, v version.Version) error {
```
pavolloffay, line 196: First I moved this to a separate package, but there was not real benefit in it - e.g. I would be only able to test if a dependency was added to the manager which is not very useful.

pavolloffay:
@VineethReddy02 @jpkrohling could you please review this one?

main.go
```diff
@@ -109,6 +109,7 @@ func main() {
 		config.WithVersion(v),
 		config.WithCollectorImage(collectorImage),
 		config.WithTargetAllocatorImage(targetAllocatorImage),
+		config.WithAutoInstrumentationJavaImage(autoInstrumentationJava),
```
VineethReddy02, line 112: @pavolloffay Curious here, Why does Java auto-instrumentation logic is part of `main.go` but not the nodejs and python instrumentation logic? I'm assuming the auto-instrumentation is done using the sidecar so mostly the logic of injecting the container using the mutating webhook should be common. 

Note: I haven't followed the instrumentation PRs closely. I will make some time to go through the instrumentation pkg.

VineethReddy02 (approved):
LGTM!

main.go
pavolloffay: great point :)))

> but not the nodejs and python instrumentation logic?

because the nodejs and pythond does not support defaulting for the image and upgrade. Once this is merged I will submit a PR to make all auto-instrumentation the same.

pavolloffay:
@VineethReddy02 could we move this PR forward?

PR #570, opened 2021-11-22
Add some resource attributes via k8s downward API

### LINKED ISSUE #554: The pod name is empty in pod webhook
The pod name is empty/not set in the pod webhook when the pod is created from higher-level object - deployment, statefulset. This also applies for pod UID. The only time pod name is present is when the pod is created from the pod object. 

We should figure out how to get the pod name and set it in the resource attributes. One way to get the pod name is via k8s downward API, but we cannot use it out-of-the-box as there is no way to set it to resource variable - see https://github.com/open-telemetry/opentelemetry-specification/issues/2135

@jpkrohling any thoughts?

pavolloffay:
Pod name, UID and node name might not be set in pod object in webhook.

Resolves #554

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -1,2 +1,8 @@
 resources:
-  - manager.yaml
+- manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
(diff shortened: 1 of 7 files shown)


VineethReddy02 (approved):
LGTM!

PR #574, opened 2021-11-23
Add liveness probe to created container if otelcol configuration supports a health_check.

### LINKED ISSUE #571: Sidecar injection does not create container probers
When a `OpenTelemetryCollector` running in `sidecar` mode is configured with a `spec.config` containing a healthcheck extenstion, I would expect that the injected container contain at least a `LivenessProbe`. In the current state, this is not the case.

-------

With a configuration like
```yaml
---
kind: OpenTelemetryCollector
metadata:
  name: otlp-sidecar
  namespace: my-namespace
spec:
  mode: sidecar
  config: |
    receivers:
      otlp:
        protocols:
          grpc:
    processors:
      batch:
        ....
    exporters:
      otlp:
        endpoint: distributor....
    extensions:
      health_check:
    service:
      extensions: [health_check]
      pipelines:
        traces:
          receivers: [otlp]
          processors: [batch]
          exporters: [otlp]
```

I would expect the injected sidecar to contain a readiness and liveness probe.

The running collector agent does expose the healthcheck, as one would expect:
```
$ curl -v http://0.0.0.0:13133
*   Trying 0.0.0.0:13133...
....[snip]...
< HTTP/1.1 200 OK
< Content-Type: application/json
< Date: Mon, 22 Nov 2021 21:08:33 GMT
< Content-Length: 102
< 
* Connection #0 to host 0.0.0.0 left intact
{"status":"Server available","upSince":"2021-11-22T14:41:53.003085969Z","uptime":"6h26m40.625191838s"}$ 
```

However there is no `LivenessProbe` nor a `ReadinessProbe` defined in the creation of the container:
https://github.com/open-telemetry/opentelemetry-operator/blob/cb2472adcb8ca1e86091b750655c118a0c1df9e9/pkg/collector/container.go#L80-L90

And we end up with an injected container such as:
```yaml
  - args:
    - --config=/conf/collector.yaml
    env:
    - name: POD_NAME
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.name
    image: otel/opentelemetry-collector:0.38.0
    imagePullPolicy: IfNotPresent
    name: otc-container
    resources:
      limits:
        memory: 96Mi
      requests:
        cpu: 20m
        memory: 96Mi
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
    volumeMounts:
    - mountPath: /conf
      name: otc-internal
```

I'd be happy to take a look at this if this is a desired feature.

adriankostrubiak-tomtom:
Add liveness probe to created container if otelcol configuration supports a health_check.

Fixes #571

Signed-off-by: Adrian Kostrubiak <adrian.kostrubiak@tomtom.com>

### DIFF
--- a/pkg/collector/adapters/config_to_probe.go
+++ b/pkg/collector/adapters/config_to_probe.go
@@ -1,0 +1,171 @@
+package adapters
+
+import (
+	"errors"
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"strings"
+)
+
+var (
+	// ErrNoService indicates that there is no service in the configuration.
+	ErrNoService = errors.New("no service available as part of the configuration")
+	// ErrNoExtensions indicates that there are no extensions in the configuration
+	ErrNoExtensions = errors.New("no extensions available as part of the configuration")
+
+	// ErrServiceNotAMap indicates that the service property isn't a map of values.
+	ErrServiceNotAMap = errors.New("service property in the configuration doesn't contain valid services")
+	// ErrExtensionsNotAMap indicates that the extensions property isn't a map of values.
+	ErrExtensionsNotAMap = errors.New("extensions property in the configuration doesn't contain valid extensions")
+
+	// ErrNoExtensionHealthCheck indicates no health_check extension was found for the
+	ErrNoExtensionHealthCheck = errors.New("extensions property in the configuration does not contain the expected health_check extension")
+
+	// ErrNoServiceExtensions indicates the service configuration does not contain any extensions
+	ErrNoServiceExtensions = errors.New("service property in the configuration doesn't contain extensions")
+
+	// ErrServiceExtensionsNotSlice indicates the service extensions property isn't a slice as expected
+	ErrServiceExtensionsNotSlice = errors.New("service extensions property in the configuration does not contain valid extensions")
+
+	// ErrNoServiceExtensionHealthCheck indicates no health_check
+	ErrNoServiceExtensionHealthCheck = errors.New("no healthcheck extension available in service extension configuration")
+)
+
+// ConfigToContainerProbe converts the incoming configuration object into a container probe or returns an error
+func ConfigToContainerProbe(logger logr.Logger, config map[interface{}]interface{}) (*corev1.Probe, error) {
+	serviceProperty, ok := config["service"]
+	if !ok {
+		return nil, ErrNoService
+	}
+	service, ok := serviceProperty.(map[interface{}]interface{})
+	if !ok {
+		return nil, ErrServiceNotAMap
+	}
+
+	serviceExtensionsProperty, ok := service["extensions"]
+	if !ok {
+		return nil, ErrNoServiceExtensions
+	}
+
+	healthcheckForProbe := ""
+
+	serviceExtensions, ok := serviceExtensionsProperty.([]interface{})
+	if !ok {
+		return nil, ErrServiceExtensionsNotSlice
+	}
+	// in the event of multiple health_check extensions defined, we arbitrarily take the first one found
+	for _, ext := range serviceExtensions {
+		parsedExt, ok := ext.(string)
+		if ok && strings.HasPrefix(parsedExt, "health_check") {
+			healthcheckForProbe = parsedExt
+			break
+		}
+	}
+
+	if healthcheckForProbe == "" {
+		return nil, ErrNoServiceExtensionHealthCheck
+	}
+
+	extensionsProperty, ok := config["extensions"]
+	if !ok {
+		return nil, ErrNoExtensions
+	}
+	extensions, ok := extensionsProperty.(map[interface{}]interface{})
+	if !ok {
+		return nil, ErrExtensionsNotAMap
+	}
+	healthcheckExtension, ok := extensions[healthcheckForProbe]
+	if !ok {
+		return nil, ErrNoExtensionHealthCheck
+	}
+
+	return createProbeFromExtension(healthcheckExtension)
+}
+
+func createProbeFromExtension(extension interface{}) (*corev1.Probe, error) {
+	probeCfg := extractProbeConfigurationFromExtension(extension)
+	return &corev1.Probe{
+		Handler: corev1.Handler{
+			HTTPGet: &corev1.HTTPGetAction{
+				Path: probeCfg.path,
+				Port: probeCfg.port,
+				Host: probeCfg.host,
+			},
+		},
+	}, nil
+}
+
+type probeConfiguration struct {
+	path string
+	port intstr.IntOrString
+	host string
+}
+
+const (
+	defaultHealthCheckPath = "/"
+	defaultHealthCheckPort = 13133
+	defaultHealthCheckHost = "0.0.0.0"
+)
+
+func extractProbeConfigurationFromExtension(ext interface{}) probeConfiguration {
+	extensionCfg, ok := ext.(map[interface{}]interface{})
+	if !ok {
+		return defaultProbeConfiguration()
+	}
+	endpoint := extractEndpointFromExtensionConfig(extensionCfg)
+	return probeConfiguration{
+		path: extractPathFromExtensionConfig(extensionCfg),
+		port: endpoint.port,
+		host: endpoint.host,
+	}
+}
+
+func defaultProbeConfiguration() probeConfiguration {
+	return probeConfiguration{
+		path: defaultHealthCheckPath,
+		port: intstr.FromInt(defaultHealthCheckPort),
+		host: defaultHealthCheckHost,
+	}
+}
+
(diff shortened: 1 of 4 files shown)


pkg/collector/adapters/config_to_probe_test.go
```diff
@@ -0,0 +1,243 @@
...
+	config, err := adapters.ConfigFromString(configStr)
+	require.NoError(t, err)
+	require.NotEmpty(t, config)
+
+	// test
+	actualProbe, err := adapters.ConfigToContainerProbe(logger, config)
+	assert.NoError(t, err)
+	assert.Equal(t, "/", actualProbe.HTTPGet.Path)
+	assert.Equal(t, int32(13133), actualProbe.HTTPGet.Port.IntVal)
+	assert.Equal(t, "0.0.0.0", actualProbe.HTTPGet.Host)
```
jpkrohling, line 26: This is now interesting: `0.0.0.0`, from the server perspective, means "bind to all IPs". IIRC, it's not a valid IP from the client's perspective. I just tried locally, and `ping 0.0.0.0` sends the ping to `127.0.0.1`, which is compatible with my expectations, given that a service bound to `0.0.0.0` will also be available at `127.0.0.1`.

Perhaps we should default to `127.0.0.1` instead then? Or is there an RFC somewhere stating what `0.0.0.0` means from a client perspective?

pkg/collector/adapters/config_to_probe_test.go
```diff
@@ -0,0 +1,243 @@
...
+
+import (
+	"github.com/open-telemetry/opentelemetry-operator/pkg/collector/adapters"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestSimpleCase(t *testing.T) {
+	configStr := `extensions:
```
jpkrohling, line 11: I think the tests can be converted to use a table-driven test. This would make it more readable.

pkg/collector/adapters/config_to_probe_test.go
```diff
@@ -0,0 +1,243 @@
...
+	assert.NoError(t, err)
+	assert.Equal(t, "/checkit", actualProbe.HTTPGet.Path)
+	assert.Equal(t, int32(1234), actualProbe.HTTPGet.Port.IntVal)
+	assert.Equal(t, "localhost", actualProbe.HTTPGet.Host)
+}
+
+func TestShouldUseCustomEndpointAndDefaultPath(t *testing.T) {
+	configStr := `extensions:
+  health_check:
+    endpoint: localhost:1234
```
jpkrohling, line 53: One case seems to be missing: `:1234`. Can you add it there as well?

pkg/collector/adapters/config_to_probe_test.go
```diff
@@ -0,0 +1,243 @@
...
+  extensions: [health_check/1, health_check]`
+
+	// prepare
+	config, err := adapters.ConfigFromString(configStr)
+	require.NoError(t, err)
+	require.NotEmpty(t, config)
+
+	// test
+	_, err = adapters.ConfigToContainerProbe(logger, config)
+	assert.Equal(t, adapters.ErrNoExtensionHealthCheck, err)
```
jpkrohling, line 242: I would expect this to not return an error?

jpkrohling (commented):
Looks great! A few adjustments here and there and this is ready to go.

pkg/collector/adapters/config_to_probe.go
```diff
@@ -0,0 +1,171 @@
...
+	if !ok {
+		return nil, ErrServiceNotAMap
+	}
+
+	serviceExtensionsProperty, ok := service["extensions"]
+	if !ok {
+		return nil, ErrNoServiceExtensions
+	}
+
+	healthcheckForProbe := ""
```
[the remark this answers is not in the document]
adriankostrubiak-tomtom: Good call, thanks.

pkg/collector/adapters/config_to_probe.go
```diff
@@ -0,0 +1,171 @@
...
+
+	serviceExtensions, ok := serviceExtensionsProperty.([]interface{})
+	if !ok {
+		return nil, ErrServiceExtensionsNotSlice
+	}
+	// in the event of multiple health_check extensions defined, we arbitrarily take the first one found
```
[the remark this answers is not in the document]
adriankostrubiak-tomtom: Yes, the container spec only allows for one LivenessProbe.

pkg/collector/adapters/config_to_probe_test.go
adriankostrubiak-tomtom: Good idea, thanks.

pkg/collector/adapters/config_to_probe_test.go
adriankostrubiak-tomtom: Whoa, thank you for your comment. I must have been thinking about the fact that the collector healthcheck defaults to listening to 0.0.0.0:13133.  But that's totally irrelevant to the probe configuration.  Clearly I should have had some more coffee before looking at this one. 🤦‍♂️

The probe should not be configured with a host, it should just be the port and the path. Per `corev1.HTTPGetAction`, 
```golang
// Host name to connect to, defaults to the pod IP. You probably want to set "Host" in httpHeaders instead.
```

I'll get that fixed up.

pkg/collector/adapters/config_to_probe_test.go
adriankostrubiak-tomtom: Agreed, that makes more sense. Have since adjusted logic.

pkg/collector/adapters/config_to_probe.go
```diff
@@ -0,0 +1,171 @@
...
+	"errors"
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"strings"
+)
+
+var (
+	// ErrNoService indicates that there is no service in the configuration.
+	ErrNoService = errors.New("no service available as part of the configuration")
```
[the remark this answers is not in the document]
adriankostrubiak-tomtom: I've removed the comments, but have left the errors as exported, as they're used in the tests. I don't have a strong feeling about that though, so I'm happy to change it if you think that makes more sense.

pkg/collector/adapters/config_to_probe_test.go
adriankostrubiak-tomtom: good call, thanks.

adriankostrubiak-tomtom:
@jpkrohling, thanks for the feedback - I appreciated it. I pushed updates to address your feedback.

pkg/collector/adapters/config_to_probe.go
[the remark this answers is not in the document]
jpkrohling: As long as the tests are in the same package, you don't need to export the errors. Exporting makes it part of the public API for this package, which would probably not be appropriate for this.

pkg/collector/adapters/config_to_probe_test.go
```diff
@@ -0,0 +1,173 @@
+package adapters_test
```
jpkrohling, line 1: You can use the same package name as the code under test. This way, you can use the non-exported errors. Use a different package if you want to exercise only the public API.

jpkrohling (commented):
Looks good, just get the test under the same package as the code, so that the errors can be kept local. I know there's an argument about exposing the errors as part of the public API, but I don't think it's what we are currently doing in this repo at the moment.

pkg/collector/adapters/config_to_probe_test.go
adriankostrubiak-tomtom: Sounds good, have updated accordingly.

jpkrohling (approved):

pkg/collector/container.go
```diff
@@ -16,10 +16,12 @@ package collector
 
 import (
 	"fmt"
+
 	"github.com/go-logr/logr"
-	"github.com/open-telemetry/opentelemetry-operator/pkg/collector/adapters"
 	corev1 "k8s.io/api/core/v1"
 
+	"github.com/open-telemetry/opentelemetry-operator/pkg/collector/adapters"
+
```
jpkrohling, line 24: nit: this extra line shouldn't be there :-)

jpkrohling (approved):
Just a small nit about import format, but I'll merge it in the morning if you don't want to fix it for this PR.

pkg/collector/container.go
adriankostrubiak-tomtom: No worries. Thought i'd ran `goimports` on it, but I guess not? Regardless, all set now.

jpkrohling (approved):

pkg/collector/container.go
jpkrohling: You might have, but if you add an extra import section explicitly, it will keep it.

Issue #577, opened 2021-11-24
Update otel collector to 0.39.0


VineethReddy02:
As part of the upgrade, Make sure to upgrade the breaking changes on the collector config i.e.

1. `httpdreceiver` renamed to `apachereceiver` to match industry standards [(#6207)](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/6207)
2. Remove deprecated config (already no-op) `ballast_size_mib` in memorylimiterprocessor [(#4365)](https://github.com/open-telemetry/opentelemetry-collector/pull/4365)

PR #578, opened 2021-11-25
Add 0.39.0 upgrade routine

VineethReddy02:
Part of #577

### DIFF
--- a/pkg/collector/upgrade/v0_24_0_test.go
+++ b/pkg/collector/upgrade/v0_24_0_test.go
@@ -57,7 +57,7 @@ func TestHealthCheckEndpointMigration(t *testing.T) {
 
 	// verify
 	assert.Equal(t, `extensions:
-  health_check: null
+  health_check:
   health_check/1: ""
   health_check/2:
     endpoint: localhost:13133
--- a/pkg/collector/upgrade/v0_39_0.go
+++ b/pkg/collector/upgrade/v0_39_0.go
@@ -1,0 +1,119 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package upgrade
+
+import (
+	"fmt"
+	"strings"
+
+	"gopkg.in/yaml.v2"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
+	"github.com/open-telemetry/opentelemetry-operator/pkg/collector/adapters"
+)
+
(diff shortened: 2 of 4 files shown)


pkg/collector/upgrade/v0_39_0.go
```diff
@@ -0,0 +1,119 @@
...
+
+package upgrade
+
+import (
+	"fmt"
+	"strings"
+
+	"gopkg.in/yaml.v2"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
jpkrohling, line 23: The import grouping looks odd here.

pkg/collector/upgrade/v0_39_0_test.go
```diff
@@ -0,0 +1,158 @@
...
+// limitations under the License.
+
+package upgrade_test
+
+import (
+	"context"
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
```
jpkrohling, line 22: Import order looks odd here as well.

pkg/collector/upgrade/v0_39_0_test.go
```diff
@@ -0,0 +1,158 @@
...
+	assert.NoError(t, err)
+
+	assert.Equal(t, `processors:
+  memory_limiter: null
+  memory_limiter/with-settings:
+    check_interval: 5s
+    limit_mib: 4000
+    spike_limit_mib: 500
+receivers:
+  apache: null
```
jpkrohling, line 82: Why the null here? Perhaps we should instead have an empty object?

jpkrohling (commented):
Other than the import order, I'm just concerned that a `null` object might break the collector. We should probably have an empty object instead of null (or just nothing, like the input).

pkg/collector/upgrade/v0_39_0.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func updateConfig(otelcol *v1alpha1.OpenTelemetryCollector, cfg map[interface{}]interface{}) (*v1alpha1.OpenTelemetryCollector, error) {
+	res, err := yaml.Marshal(cfg)
+	if err != nil {
+		return otelcol, fmt.Errorf("couldn't upgrade to v0.39.0, failed to marshall back configuration: %w", err)
+	}
+
+	// Note: This is a hack to drop null occurrences from our config parsing above in upgrade routine
+	otelcol.Spec.Config = strings.ReplaceAll(string(res), "null", "")
```
jpkrohling, line 117: I wonder if this is going to have bad side effects. Like for someone posting traces to `mynullservice.com` :-)

VineethReddy02: Maybe we should do `   null   `? with trailing spaces...

pkg/collector/upgrade/v0_24_0_test.go
```diff
@@ -57,7 +57,7 @@ func TestHealthCheckEndpointMigration(t *testing.T) {
 
 	// verify
 	assert.Equal(t, `extensions:
-  health_check: null
```
VineethReddy02, line 60: @jpkrohling the issue of `null` existed in the past too. The other contributor ignored it by adding `null` during the validation as the parser injects null to empty value.

jpkrohling (approved):

pkg/collector/upgrade/v0_24_0_test.go
jpkrohling: Can you please create an issue to track this?

PR #586, opened 2021-11-29
Rename resource attributes in instrumentation CR

pavolloffay:

### DIFF
--- a/apis/v1alpha1/instrumentation_types.go
+++ b/apis/v1alpha1/instrumentation_types.go
@@ -55,7 +55,7 @@ type Resource struct {
 	// Attributes defines attributes that are added to the resource.
 	// For example environment: dev
 	// +optional
-	Attributes map[string]string `json:"resourceAttributes,omitempty"`
+	Attributes map[string]string `json:"attributes,omitempty"`
 
 	// AddK8sUIDAttributes defines whether K8s UID attributes should be collected (e.g. k8s.deployment.uid).
 	// +optional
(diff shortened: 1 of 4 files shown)


apis/v1alpha1/instrumentation_types.go
```diff
@@ -55,7 +55,7 @@ type Resource struct {
 	// Attributes defines attributes that are added to the resource.
 	// For example environment: dev
 	// +optional
-	Attributes map[string]string `json:"resourceAttributes,omitempty"`
+	Attributes map[string]string `json:"attributes,omitempty"`
```
pavolloffay, line 58: This is a breaking change since 0.39.0, but it should not be too bad - I do not expect there is high adoption of 0.39.0 and 0.40.0 is going out soon. 

cc) @VineethReddy02 @jpkrohling

jpkrohling: Even if just to practice, how about doing it in two steps? Deprecate this and add a new property, and in the next release you remove this deprecated property?

pavolloffay: > Even if just to practice

Then I will leave it for somebody who wants to play around.  For the reasons I mentioned above, I would do it right away.

jpkrohling: Sure. Just to confirm: you are still the owner of this code, right?

pavolloffay: I will not comment on that for some reasons, but I am the original author of the feature.

pavolloffay:
Closing 0.40.0 was released without this rename.

PR #583, opened 2021-11-26
Use cmctl to test if cert-manager is ready

pavolloffay:
The current approach is flaky and cert-manager
docs suggests to use cmctl to test if the cert-manager is
ready.

This should fix flaky CI e2e jobs:
* https://github.com/open-telemetry/opentelemetry-operator/runs/4331745148?check_suite_focus=true

Signed-off-by: Pavol Loffay <p.loffay@gmail.com>

PR #587, opened 2021-11-29
Do not set exporter endpoint if it is empty

### LINKED ISSUE #531: OTEL_EXPORTER_OTLP_ENDPOINT set to empty string if not set in spec
It should not be injected if it's not set in the instrumentation config.

pavolloffay:
Resolves #531 


pavolloffay:
the e2e test failed bc of the cert-manager not being ready - this should be fixed in #583

VineethReddy02 (approved):
LGTM!

PR #595, opened 2021-12-02
Recommend using headless service

pavolloffay:

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ kubectl apply -f - <<EOF
 apiVersion: opentelemetry.io/v1alpha1
 kind: OpenTelemetryCollector
 metadata:
-  name: simplest
+  name: otel
 spec:
   config: |
     receivers:
@@ -51,7 +51,7 @@ EOF
 **_WARNING:_** Until the OpenTelemetry Collector format is stable, changes may be required in the above example to remain
 compatible with the latest version of the OpenTelemetry Collector image being referenced.
 
-This will create an OpenTelemetry Collector instance named `simplest`, exposing a `jaeger-grpc` port to consume spans from your instrumented applications and exporting those spans via `logging`, which writes the spans to the console (`stdout`) of the OpenTelemetry Collector instance that receives the span.
+This will create an OpenTelemetry Collector instance named `otel`, exposing two services `otel-collector` and `otel-collector-headless`. The received spans will be exported via `logging` exporter and written to the console (`stdout`). The headless service is preferred if gRPC protocol is used to export data to the collector.
 
 The `config` node holds the `YAML` that should be passed down as-is to the underlying OpenTelemetry Collector instances. Refer to the [OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector) documentation for a reference of the possible entries.
 
@@ -166,7 +166,7 @@ metadata:
   name: my-instrumentation
 spec:
   exporter:
-    endpoint: http://otel-collector:4317
+    endpoint: http://otel-collector-headless:4317
   propagators:
     - tracecontext
     - baggage
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ EOF
 **_WARNING:_** Until the OpenTelemetry Collector format is stable, changes may be required in the above example to remain
 compatible with the latest version of the OpenTelemetry Collector image being referenced.
 
-This will create an OpenTelemetry Collector instance named `otel`, exposing two services `otel-collector` and `otel-collector-headless`. The received spans will be exported via `logging` exporter and written to the console (`stdout`). The headless service is preferred if gRPC protocol is used to export data to the collector.
+This will create an OpenTelemetry Collector instance named `otel`, exposing two services `otel-collector` and `otel-collector-headless`. The received spans will be exported via `logging` exporter and written to the console (`stdout`). If a gRPC protocol is used to export data to the collector use the headless service that ensures load balancing will properly work.
 
 The `config` node holds the `YAML` that should be passed down as-is to the underlying OpenTelemetry Collector instances. Refer to the [OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector) documentation for a reference of the possible entries.
 
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ kubectl apply -f - <<EOF
 apiVersion: opentelemetry.io/v1alpha1
 kind: OpenTelemetryCollector
 metadata:
-  name: otel
+  name: simplest
 spec:
   config: |
     receivers:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,7 +26,7 @@ kubectl apply -f - <<EOF
 apiVersion: opentelemetry.io/v1alpha1
 kind: OpenTelemetryCollector
 metadata:
-  name: simplest
+  name: otel
```
pavolloffay, line 29: renaming this to otel. Simplest creates `simplest-collector` and otel creates `otel-collector` service which is more idiomatic for an example.

VineethReddy02:
To understand the context here, any specific reason for recommending headless service? I mean the normal service name would do the job, right?

pavolloffay:
The load balancing does not work with gRPC for the normal service.

pavolloffay:
I will update the text to include it there

README.md
jpkrohling: That doesn't seem right to me for two reasons:

- `otel` != `otelcol`. An OpenTelemetry Collector (otelcol) CR named just "OpenTelemetry" (otel) is technically wrong. How about OpenTelemetry Instrumentations? Shouldn't they also be called `otel` then?
- naming the CR (`otel`) after one of the resources it creates (`otel-collector`) doesn't look like a good practice

Besides, as someone starting with this project, I think it's clearer to understand the effects of my actions if I use a name that isn't related to the underlying tech. If I create a CR named `simplest` and see a `simplest-collector-headless`, I know immediately that this is specific to my service. If I just named it `otel`, I could miss the fact that `otel-collector-headless` is specific to my service, as opposed to something the operator itself needs.

README.md
```diff
@@ -51,7 +51,7 @@ EOF
 **_WARNING:_** Until the OpenTelemetry Collector format is stable, changes may be required in the above example to remain
 compatible with the latest version of the OpenTelemetry Collector image being referenced.
 
-This will create an OpenTelemetry Collector instance named `simplest`, exposing a `jaeger-grpc` port to consume spans from your instrumented applications and exporting those spans via `logging`, which writes the spans to the console (`stdout`) of the OpenTelemetry Collector instance that receives the span.
+This will create an OpenTelemetry Collector instance named `otel`, exposing two services `otel-collector` and `otel-collector-headless`. The received spans will be exported via `logging` exporter and written to the console (`stdout`). If a gRPC protocol is used to export data to the collector use the headless service that ensures load balancing will properly work.
```
jpkrohling, line 54:
```suggestion
This will create an OpenTelemetry Collector instance named `otel`, exposing two services `otel-collector` and `otel-collector-headless`. The received spans will be exported via `logging` exporter and written to the console (`stdout`). If a gRPC protocol is used to export data to the collector, use the headless service to ensure that load balancing will properly work.
```

Doesn't the endpoint need to be prefixed with `dns:///` for that to happen? Like `dns:///simplest-collector-headless:14250`. Just using the headless service won't ensure the load balancing will work.

README.md
pavolloffay: `simplest` creates (and other objects named similarly) - `service/simplest-collector` and `deployment/simplest-collectors`. Now, I fear that this is not expressive enough and might collide with other objects in a namespace. Users should know what technology they are deploying, bc then it depends on what they can integrate it with or connect to it. 

> otel != otelcol.

yes, but the created objects in the cluster make total sense to me `deployment.apps/otel-collector`, `service/otel-collector`.  Now in the instrumentation or envoy I will use `otel-collector:4317` which is idiomatic enough to understand what it is instead of `simplest-collector:4317` (what is simplest? Is it Jaeger, Zipkin or my simple demo from the last kubecon?) 


That said the name in the ALM examples is also `otel`  https://github.com/open-telemetry/opentelemetry-operator/blob/main/bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml#L23.

README.md
pavolloffay: It should be enough without dns

README.md
jpkrohling: > That said the name in the ALM examples is also otel https://github.com/open-telemetry/opentelemetry-operator/blob/main/bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml#L23.

Apparently you changed it in #519 ;-)

README.md
jpkrohling: It's not. In fact, you should also have a balancer name set to `round_robin`, otherwise, it will use `pick_first` by default.

README.md
pavolloffay: yes bc the name was `opentelemetrycollector-sample` and it created objects named `opentelemetrycollector-sample-collector`

README.md
pavolloffay: The round-robin is set in the application we cannot control it here.

jpkrohling: Right. While I would still think that providing the actual string for them to use as an example would be useful, the text here is technically correct.

README.md
jpkrohling: My point remains: `otel` is a bad example name for this CR, despite the existence of an example with that name.

README.md
```diff
@@ -166,7 +166,7 @@ metadata:
   name: my-instrumentation
 spec:
   exporter:
-    endpoint: http://otel-collector:4317
+    endpoint: http://otel-collector-headless:4317
```
jpkrohling, line 169: Did you verify that this is indeed load balancing? Is this the recommended pattern for users?

README.md
pavolloffay: For instance I don't care about the instrumentation CR name because it does not create any objects. Which is not the case for `OpenTelemetryCollector` CR. If the CR creates objects I do care more about the object names and especially about service, deployment and pod names and not that much about the CR name. The service,deployment,pod names are the most important this is what people see and relate to.

README.md
pavolloffay: I did test this with OTEL javaagent. The agent does not set round_robin but that is another story for another repository. 

That said, there were people asking about load balancing in clients for other languages.

README.md
pavolloffay: I am open to suggestions :)

README.md
jpkrohling: I'm asking because if you were to set this to `dns:///otel-collector-headless:4317`, the gRPC client will get all the IPs from the DNS entry `otel-collector-headless` (which is an A record) and use them all. The default strategy is pick_first, which will just get one of those IPs. Upon failure, the next IP is used. When using a regular service, only one IP is resolved, which is the service that is backed by a Kubernetes service.

My point with all of this is that I don't understand the benefits that the headless service is bringing when using HTTP as the protocol instead of `dns:///`. I would have guessed that a regular service would work better when not using the DNS resolution, as it would be backed by a Kubernetes service that has the chance of redirecting requests in case of failures.

pavolloffay: Did you try using `dns:///` with OTEL auto-instrumentations? It might not even work.

pavolloffay: Apparently not https://github.com/open-telemetry/opentelemetry-java/blob/65e7145e9c6151416042757941d45f8970819248/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java#L154

pavolloffay: Please submit an issue to spec repo `OTEL_EXPORTER_OTLP_ENDPOINT` should support `dns:///` prefix.

README.md
jpkrohling: > I am open to suggestions :)

Don't change it ;-)

pavolloffay: I really don't like `simplest-collector` it is not idiomatic and when I see it in my cluster I literally don't know what it is!

Would you name a MongoDB instance `simple-database`?

pavolloffay: I am curious what instance name of oTEL collector are you using in grafana.

pavolloffay: nevermind I have renamed `otel` back to `simplest` even though I find it ambiguous for new users.

README.md
jpkrohling: I still don't understand why headless is better for this case if the client cannot take advantage of obtaining the backing IP addresses the service provides.

jpkrohling:
I think I spent too much energy on this PR already. @VineethReddy02, could you please take over this one? Review and merge at your discretion.

pavolloffay:
I feel exactly the same way! Trying to improve docs by changing 3 lines and I had to make good ~20 comments.

jpkrohling:
`s/Trying to improve docs/Trying to change the docs without need/` -- FTFY.

pavolloffay:
That is not very kind, nor professional. 

I am sorry that you don't see any good intentions here and fixing docs that point to jaeger ports, while jaeger is not being used in the example.

jpkrohling:
> That is not very kind, nor professional.

I'm sorry, my comment was meant to demonstrate that the contention here is now down to opinions: my opinion is that "simplest" is clearer for users beginning with the operator, and you have a different opinion. The other contention is on using the headless service when there's no apparent benefit in doing that.

Because I do realize that my opinions might not be the best ones, I've removed myself from this PR, so that another maintainer (@VineethReddy02) can share his opinions as well.

PR #600, opened 2021-12-03
Fix container image field in CSV

### LINKED ISSUE #599: Community Operators - OpenTelemetry operator failing on Kiwi test
After releasing v0.39.0 to community operators repo, I noticed that we are facing the following issue: 

`TASK [check_image_availability : Print err message] ****************************
fatal: [localhost]: FAILED! => changed=false 
  msg: |-
    Value of 'metadata.annotations.containerImage' not used in any container, please check.
    The operator is using some image which must have exactly the same value as 'metadata.annotations.containerImage'.
    Please check both values and make them the same.
`
https://github.com/k8s-operatorhub/community-operators/runs/4406065202?check_suite_focus=true

pavolloffay:

Resolves #599

### DIFF
--- a/config/manifests/bases/opentelemetry-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/opentelemetry-operator.clusterserviceversion.yaml
@@ -6,11 +6,9 @@ metadata:
     capabilities: Basic Install
     categories: Logging & Tracing
     certified: "false"
-    containerImage: quay.io/opentelemetry/opentelemetry-operator
+    containerImage: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
     createdAt: "2020-12-16T13:37:00+00:00"
     description: Provides the OpenTelemetry components, including the Collector
-    operators.operatorframework.io/builder: operator-sdk-v1.2.0
-    operators.operatorframework.io/project_layout: go
     repository: github.com/open-telemetry/opentelemetry-operator
     support: OpenTelemetry Community
   name: opentelemetry-operator.v0.0.0
--- a/bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
@@ -30,7 +30,7 @@ metadata:
     capabilities: Basic Install
     categories: Logging & Tracing
     certified: "false"
-    containerImage: quay.io/opentelemetry/opentelemetry-operator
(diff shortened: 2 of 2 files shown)


config/manifests/bases/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -6,11 +6,9 @@ metadata:
     capabilities: Basic Install
     categories: Logging & Tracing
     certified: "false"
-    containerImage: quay.io/opentelemetry/opentelemetry-operator
+    containerImage: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
     createdAt: "2020-12-16T13:37:00+00:00"
     description: Provides the OpenTelemetry components, including the Collector
-    operators.operatorframework.io/builder: operator-sdk-v1.2.0
```
pavolloffay, line 12: this is pulled in automatically from `PROJECT` file.

jpkrohling (approved):
LGTM, but don't we have a target that checks whether the generated bundle is in line with the sources?

pavolloffay:
I don't know what you mean. 

Do you mean something like that failing `kiwi` test from the community-operators? I don't know if the SDK/OLM provides those tests to be consumed in the operator projects.

jpkrohling:
We have a `make generate` and `make bundle`, and we check during the CI whether their output is noop. Should this change have been applied as part of those operations?

pavolloffay:
sorry I still don't get you. 

The only way to test the correctness of this change would be to run similar tests with the community-operators does.

jpkrohling:
Isn't this file being generated via `make bundle`?

```
bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```

pavolloffay:
yes, but I still don't know what do you want to say

jpkrohling:
If the file is auto-generated, then changing it like it's being done in this PR won't solve the problem, right?

pavolloffay:
The `config/manifests/bases/opentelemetry-operator.clusterserviceversion.yaml` serves as a base to generate the  `bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml`

I still don't know what you are trying to suggest.

jpkrohling:
> The config/manifests/bases/opentelemetry-operator.clusterserviceversion.yaml serves as a base to generate the bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml

Sorry! I missed that there are two files to this PR, one being the source, and the other being the auto-generated. Merging.

pavolloffay:
awesome it took only 4 comments! :)

PR #620, opened 2021-12-09
Add OpenTelemetryCollector.Spec.UpgradeStrategy

### LINKED ISSUE #598: Opt-out of managed instance upgrade
As a cluster admin running the `opentelemetry-operator`, I would like to avoid having the controller automatically upgrade the configuration of `OpenTelemetryCollector`s and would rather manage that process myself.

---

Automatic upgrades can be an undesirable event, and lead to confusing situations. This is especially the case when using a custom `spec.image`. 

The upgrade process assumes using the latest `otel/opentelemetry-collector` as defined by the version of the `opentelemetry-operator`.  As such, it is possible for the controller's upgrade process to upgrade a `OpenTelemetryCollector` resource's configuration to an invalid configuration given the image being run.  I believe the opportunity for this to occur is greater when upgrading over a wider version range (eg 0.24 -> 0.36 vs 0.38 -> 0.40) as the underlying configuration has more opportunity to change across the versions.

Looking into the code a bit, and validated with an experiment, this is currently accomplishable by applying a label to an `OpenTelemetryCollector` of `app.kubernetes.io/managed-by` set to anything other than `"opentelemetry-operator"`:
https://github.com/open-telemetry/opentelemetry-operator/blob/a94a13888165ec688a5f83d4e444e9fbc934fca2/pkg/collector/upgrade/upgrade.go#L35-L39

However, it is unclear if such a configuration will result in unintended consequences.

If this is indeed the appropriate approach, I would like to suggest improving the documentation around the upgrade process.

If not, I believe there should be a configuration such that the controller will not upgrade a given `OpenTelemetryCollector`.

Regardless of the answer, I'd be happy to take a crack at the resulting work :)

adriankostrubiak-tomtom:
Add OpenTelemetryCollector.Spec.UpgradeStrategy with allowable values: `automatic`, `none`

fixes #598

Signed-off-by: Adrian Kostrubiak adrian.kostrubiak@tomtom.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,25 @@ The `config` node holds the `YAML` that should be passed down as-is to the under
 
 At this point, the Operator does *not* validate the contents of the configuration file: if the configuration is invalid, the instance will still be created but the underlying OpenTelemetry Collector might crash.
 
+
+### Upgrades
+
+As noted above, the OpenTelemetry Collector format is continuing to evolve.  However, a best-effort attempt is made to upgrade all managed `OpenTelemetryCollector` resources.
+
+In certain scenarios, it may be desirable to prevent the operator from upgrading certain `OpenTelemetryCollector` resources. For example, when a resource is configured with a custom `.Spec.Image`, end users may wish to manage configuration themselves as opposed to having the operator upgrade it.  This can be configured on a resource by resource basis with the exposed property `.Spec.UpgradeStrategy`.
+
+By configuring a resource's `.Spec.UpgradeStrategy` to `none`, the operator will skip the given instance during the upgrade routine.
+
+The default and only other acceptable value for `.Spec.UpgradeStrategy` is `automatic`.
+
+
+### `opentelemetry-operator` vs `OpenTelemetryCollector` Versioning
+
+By default, the operator ensures consistent versioning between itself and the managed `OpenTelemetryCollector` resources.  That is, if the operator is based on version `0.40.0`, it will create resources with an underlying core `opentelemetry-collector` at version `0.40.0`.
+
+When a custom `Spec.Image` is used, the operator will not manage this versioning and upgrading. In this scenario, it is best practice that the operator version should match the underlying core version. Given a `OpenTelemetryCollector` resource with a `Spec.Image` configured to a custom image based on underlying core `opentelemetry-collector` at version `0.40.0`, it is recommended that the operator is kept at version `0.40.0`.
+
+
 ### Deployment modes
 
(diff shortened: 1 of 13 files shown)


adriankostrubiak-tomtom:
(I still want to look at this a little further, and double check a few things (are there enough tests?), but figured I'd throw a draft PR out there with this idea to get any feedback)

README.md
```diff
@@ -57,6 +57,18 @@ The `config` node holds the `YAML` that should be passed down as-is to the under
 
 At this point, the Operator does *not* validate the contents of the configuration file: if the configuration is invalid, the instance will still be created but the underlying OpenTelemetry Collector might crash.
 
+
+### Upgrades
+
+As noted above, the OpenTelemetry Collector format is continuing to evolve.  However, a best-effort attempt is made to upgrade all managed `OpenTelemetryCollector` resources.
+
+In certain scenarios, it may be desirable to prevent the operator from upgrading certain `OpenTelemetryCollector` resources. For example, when a resource is configured with a custom `.Spec.Image`, end users may wish to manage configuration themselves as opposed to having the operator upgrade it.  This can be configured on a resource by resource basis with the exposed property `.Spec.UpgradeStrategy`.
```
jpkrohling, line 65: It should be noted (elsewhere?) that the operator version should match the underlying core version. If your custom image is based on core 0.40.0, then it's recommended that the operator is kept at 0.40.0.

pkg/collector/upgrade/upgrade_test.go
```diff
@@ -138,3 +126,15 @@ func TestVersionsShouldNotBeChanged(t *testing.T) {
 		})
 	}
 }
+
+func makeOtc(nsn types.NamespacedName) v1alpha1.OpenTelemetryCollector {
```
jpkrohling, line 130:
```suggestion
func makeOtelcol(nsn types.NamespacedName) v1alpha1.OpenTelemetryCollector {
```

jpkrohling (commented):
This looks great!

README.md
adriankostrubiak-tomtom: Good call, added some more doc below. Let me know if you think that's clear enough.

README.md
```diff
@@ -57,6 +57,25 @@ The `config` node holds the `YAML` that should be passed down as-is to the under
...
+
+By configuring a resource's `.Spec.UpgradeStrategy` to `none`, the operator will skip the given instance during the upgrade routine.
+
+The default and only other acceptable value for `.Spec.UpgradeStrategy` is `automatic`.
+
+
+### `opentelemetry-operator` vs `OpenTelemetryCollector` Versioning
```
jpkrohling, line 72: We have a section named "### OpenTelemetry Operator vs. OpenTelemetry Collector", perhaps it could be adapted with the extra information from this part here?

I do like the actual text, though!

adriankostrubiak-tomtom: Good call, I've adjusted accordingly. Happy to tweak further if you think it still needs further clarification/etc/etc. Just let me know.

jpkrohling (approved):

PR #631, opened 2021-12-12
Update comp matrix to include Kubernetes 1.23

### LINKED ISSUE #615: Add Kubernetes 1.23 to the test matrix
Similar to #378, Kubernetes 1.23 has just been released and we need to add it to the test and compatibility matrix.

https://github.com/kubernetes/kubernetes/releases/tag/v1.23.0

shree007:
Closes #615.

This also removes 1.20, to keep the compat matrix manageable.


README.md
```diff
@@ -246,16 +246,16 @@ The OpenTelemetry Operator *might* work on versions outside of the given range,
...
-| v0.40.0                | v1.20 to v1.22       | 1.6.1                |
-| v0.39.0                | v1.20 to v1.22       | 1.6.1                |
-| v0.38.0                | v1.20 to v1.22       | 1.6.1                |
-| v0.37.1                | v1.20 to v1.22       | v1.4.0 to v1.6.1     |
-| v0.37.0                | v1.20 to v1.22       | v1.4.0 to v1.5.4     |
-| v0.36.0                | v1.20 to v1.22       | v1.4.0 to v1.5.4     |
-| v0.35.0                | v1.20 to v1.22       | v1.4.0 to v1.5.4     |
-| v0.34.0                | v1.20 to v1.22       | v1.4.0 to v1.5.4     |
-| v0.33.0                | v1.20 to v1.22       | v1.4.0 to v1.5.4     |
+| v0.41.0                | v1.21 to v1.23       | 1.6.1                |
```
jpkrohling, line 249: This is not correct. Released versions shouldn't be changed, the matrix is changed only from this point and on. So, just add a new entry like this:

```
| v0.42.0 (unreleased)               | v1.21 to v1.23       | 1.6.1                |
```

shree007: Noted.

shree007: @jpkrohling  Please check, I have updated

jpkrohling: 0.41.0 has been released already without this PR, so, it really has to be "0.42.0 (unreleased)" here.

shree007: @jpkrohling updated.

.github/workflows/e2e.yaml
```diff
@@ -14,9 +14,9 @@ jobs:
       matrix:
        kube-version:
        - "1.19"
```
VineethReddy02, line 16: @jpkrohling any specific reason, why are we dropping `1.20` but not the `1.19` ?

README.md
jpkrohling: 0.41.0 is still wrong, it shouldn't be changed at all, as it has been released already.

.github/workflows/e2e.yaml
jpkrohling: I thought we had removed 1.19 already. It should indeed be removed.

README.md
```diff
@@ -246,7 +246,8 @@ The OpenTelemetry Operator *might* work on versions outside of the given range,
 
 | OpenTelemetry Operator | Kubernetes           | Cert-Manager         |
 |------------------------|----------------------|----------------------|
-| v0.41.0                | v1.20 to v1.22       | 1.6.1                |
+| v0.42.0 (unreleased)   | v1.21 to v1.23       | 1.6.1
+| v0.41.0                | v1.21 to v1.22       | 1.6.1                |
```
jpkrohling, line 250: This should really be left untouched :-)

VineethReddy02:
closing this PR in favor to #633

PR #633, opened 2021-12-14
Update comp matrix to include Kubernetes 1.23

shree007:
… and removing the 1.19, 1.20 and adding 1.23


shree007:
@jpkrohling Please review
Thanks

PR #640, opened 2021-12-17
Add read permission to 'group' for java autoinstrumentation JAR

### LINKED ISSUE #636: Instrumentation example using java-agent fails (agent jar in container at wrong place)
When trying the operator Intrumentation example in the main readme.dm of this project, using the opentelemetry-operator.yaml suggested there and as instrumentation the following file...

```
kind: Instrumentation
metadata:
  name: my-instrumentation
spec:
  exporter:
    endpoint: http://otel-collector:4317
  propagators:
    - tracecontext
    - baggage
    - b3
  sampler:
    argument: "0.25"
  java:
    image: ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-java:latest
```

...the injection fails with this error:

$ k logs pod/xxxxxx-5fcb94b7b5-dhcft --tail=20
Picked up JAVA_TOOL_OPTIONS: -javaagent:/otel-auto-instrumentation/javaagent.jar
Error opening zip file or JAR manifest missing : /otel-auto-instrumentation/javaagent.jar
Error occurred during initialization of VM
agent library failed to init: instrument
 
It seems as in the referred image autoinstrumentation-java, the jar is in the root fs and not in /otel... dir.

$ docker run -it ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-java sh
/ # ls -l
```
total 22004
drwxr-xr-x 2 root root 12288 Nov 29 18:55 bin
drwxr-xr-x 5 root root 360 Dec 7 17:37 dev
drwxr-xr-x 1 root root 4096 Dec 7 17:37 etc
drwxr-xr-x 2 nobody nobody 4096 Nov 29 18:55 home
-rw------- 1 root root 22492882 Nov 4 07:24 javaagent.jar
dr-xr-xr-x 227 root root 0 Dec 7 17:37 proc
drwx------ 1 root root 4096 Dec 7 17:37 root
dr-xr-xr-x 11 root root 0 Dec 7 17:37 sys
drwxrwxrwt 2 root root 4096 Nov 29 18:55 tmp
drwxr-xr-x 3 root root 4096 Nov 29 18:55 usr
drwxr-xr-x 4 root root 4096 Nov 29 18:55 var
```

luigibk:
Fixes Java auto-instrumentation permissions when fsGroup is in use in the container. 

Fixes #636

### DIFF
--- a/autoinstrumentation/java/Dockerfile
+++ b/autoinstrumentation/java/Dockerfile
@@ -4,4 +4,4 @@ ARG version
 
 ADD https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/download/v$version/opentelemetry-javaagent.jar /javaagent.jar
 
-RUN chmod -R o+r /javaagent.jar
+RUN chmod -R go+r /javaagent.jar
--- a/autoinstrumentation/nodejs/Dockerfile
+++ b/autoinstrumentation/nodejs/Dockerfile
@@ -9,4 +9,4 @@ FROM busybox
 
 COPY --from=build /operator-build/build/workspace /autoinstrumentation
 
-RUN chmod -R o+r /autoinstrumentation
+RUN chmod -R go+r /autoinstrumentation
--- a/autoinstrumentation/python/Dockerfile
+++ b/autoinstrumentation/python/Dockerfile
@@ -10,4 +10,4 @@ FROM busybox
 
 COPY --from=build /operator-build/workspace /autoinstrumentation
 
-RUN chmod -R o+r /autoinstrumentation
+RUN chmod -R go+r /autoinstrumentation


autoinstrumentation/java/Dockerfile
```diff
@@ -4,4 +4,4 @@ ARG version
 
 ADD https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/download/v$version/opentelemetry-javaagent.jar /javaagent.jar
 
-RUN chmod -R o+r /javaagent.jar
+RUN chmod -R go+r /javaagent.jar
```
pavolloffay, line 7: Could you please explain why `o`/`others` is not sufficient? I think it should be sufficient. 

> Any other user who has access to a file. This person has neither created the file, nor he belongs to a usergroup who could own the file. Practically, it means everybody else. Hence, when you set the permission for others, it is also referred as set permissions for the world.

pavolloffay: from https://www.guru99.com/file-permissions.html

pavolloffay: thanks for the explanation, now I would like to do 2 things:
1. add the same to other instrumentations
2. test this in e2e tests - most likely add security context here https://github.com/open-telemetry/opentelemetry-operator/blob/main/tests/e2e/instrumentation-java/01-install-app.yaml

pavolloffay: Here are docs for the security context https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod

> In the configuration file, the runAsUser field specifies that for any Containers in the Pod, all processes run with user ID 1000. The runAsGroup field specifies the primary group ID of 3000 for all processes within any containers of the Pod. If this field is omitted, the primary group ID of the containers will be root(0). Any files created will also be owned by user 1000 and group 3000 when runAsGroup is specified. Since fsGroup field is specified, all processes of the container are also part of the supplementary group ID 2000. The owner for volume /data/demo and any files created in that volume will be Group ID 2000.

My understanding is that when the `fsGroup` is set then copied `javaagent.jar` file will inherit that group. But I still don't understand why the file cannot be read if the file has `o+r` set? cc) @jpkrohling

luigibk: Because user belongs to fsGroup and "others" means everybody else. That's what I see happening at least without group read permissions. 
I'll try to add an e2e test, but not sure I can finish that before the holidays, so it might have to wait until January. 
Thanks for getting back to me so quickly.

jpkrohling: `o` means "everyone else" (any _other_ user), not "everyone". You can try that locally, like this (if you don't have the apache user/group locally, use some other group that your account doesn't belong):

```
$ echo contents > testfile 
$ ls -lah testfile
-rw-r--r--. 1 jpkroehling wheel 9 Dec 17 11:18 testfile
$ sudo chown apache:apache testfile
$ sudo chmod 604 testfile
$ ls -lah testfile
-rw----r--. 1 apache apache 9 Dec 17 11:18 testfile
$ cat testfile
contents
$ sudo chown apache:wheel testfile
$ groups
wheel libvirt
$ cat testfile
cat: testfile: Permission denied
```

pavolloffay: I think I get it now, in the last example the `jpkroehling` is part of the `wheel` group but the file does not give permission to the group. If `jpkroehling` was removed from the `wheel` group it could read the file. 

thanks!

pavolloffay:
@luigibk please add similar permissions to the other instrumentations, I can take a look at the e2e test.

pavolloffay (approved):

PR #643, opened 2021-12-20
Test file permissions for auto-instrumentations

pavolloffay:

Related to #640

### DIFF
--- a/tests/e2e/instrumentation-java/01-install-app.yaml
+++ b/tests/e2e/instrumentation-java/01-install-app.yaml
@@ -15,6 +15,10 @@ spec:
         sidecar.opentelemetry.io/inject: "true"
         instrumentation.opentelemetry.io/inject-java: "true"
     spec:
+      securityContext:
+        runAsUser: 1000
+        runAsGroup: 3000
+        fsGroup: 2000
       containers:
       - name: myapp
         image: ghcr.io/pavolloffay/spring-petclinic:latest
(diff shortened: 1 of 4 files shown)


tests/e2e/instrumentation-java/01-install-app.yaml
```diff
@@ -15,6 +15,10 @@ spec:
         sidecar.opentelemetry.io/inject: "true"
         instrumentation.opentelemetry.io/inject-java: "true"
     spec:
+      securityContext:
```
jpkrohling, line 18: Would this work on OpenShift? My understanding is that it needs a special SCC configuration done by the cluster-admin in order to allow this config at the workload level.

Note also that OpenShift (and likely other platforms) sets a random user ID to the actual processes, so, if you need the user ID to be a specific number, your code will certainly break at runtime.

pavolloffay: I don't know if this works on OCP or other platforms. 

The previous commit shows that it fails if the image does not have `go+r`.

pavolloffay:
@jpkrohling any update from your side on this PR? 

The goal of this PR is to add tests for https://github.com/open-telemetry/opentelemetry-operator/pull/640. The tests fail without that PR being merged.

jpkrohling (approved):
It's OK for me, but if this means we have to rely on SCC to get things to work, then things might *not* work on multi-tenant Kubernetes platforms, as SCC is disabled to regular, non-admin users.

PR #651, opened 2021-12-22
Fix port derivation

### LINKED ISSUE #646: Parsing httpd endpoint in Otel-operator is broken.
I'm trying to use the httpd receiver with a declared endpoint, like : 

```
kind: OpenTelemetryCollector
metadata:
  name: simplest
spec:
  config: |
    receivers:
      otlp:
        protocols:
          grpc:
            endpoint: localhost:9090
      httpd:
        endpoint: http://localhost:8080/server-status?auto
        collection_interval: 10s
    processors:
      nop:
    exporters:
      logging:
    service:
      pipelines:
        traces:
          receivers: [otlp]
          processors: []
          exporters: [logging]
        metrics:
          receivers: [httpd]
          processors: [nop]
          exporters: [logging]
```

However, I'm getting the following error once the CR is deployed: 
```
opentelemetry-operator-controller-manager-68cf456bd-6lmbs manager {"level":"info","ts":1640033774.9538267,"logger":"controllers.OpenTelemetryCollector","msg":"couldn't parse the endpoint's port","opentelemetrycollector":"opentelemetry-operator-system/simplest","endpoint":"http://localhost:80/server-status?auto"}
opentelemetry-operator-controller-manager-68cf456bd-6lmbs manager {"level":"info","ts":1640033774.9549108,"logger":"controllers.OpenTelemetryCollector","msg":"couldn't parse the endpoint's port","opentelemetrycollector":"opentelemetry-operator-system/simplest","endpoint":"http://localhost:80/server-status?auto"}
```

yuriolisa:
- Change the approach to extract endpoint ports.
- Creating tests to test it.

Resolves #646


yuriolisa:
Fixes #646

pkg/collector/parser/receiver.go
```diff
@@ -20,6 +20,9 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"net"
```
VineethReddy02, line 23: Please fix the go imports formatting here

pkg/collector/parser/receiver.go
```diff
@@ -155,10 +158,27 @@ func portName(receiverName string, port int32) string {
...
+	host, port, err := net.SplitHostPort(u.Host)
+
+	if err != nil {
+		return 0, err
+	}
+
+	if len(host) > 0 {
+		port, err := strconv.ParseInt(port,10,32)
+		return int32(port), err
+	} else {
```
VineethReddy02, line 176: We don't need `else` here as we used return in `if`

tests/e2e/endpoint-extraction/00-assert.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: apps/v1
```
VineethReddy02, line 1: I see there is a unit test for the same `func` that you are modifying the logic i.e. `TestReceiverParsePortFromEndpoint` in `receiver_test.go` I would recommend adding an additional case in the test function rather than creating an e2e suite for a corner/edge case.

VineethReddy02 (changes requested):
Needs to some changes

pkg/collector/parser/receiver.go
```diff
@@ -155,10 +158,27 @@ func portName(receiverName string, port int32) string {
 }
 
 func portFromEndpoint(endpoint string) (int32, error) {
-	i := strings.LastIndex(endpoint, ":") + 1
-	part := endpoint[i:]
-	port, err := strconv.ParseInt(part, 10, 32)
-	return int32(port), err
+	u, err := url.Parse(endpoint)
+
```
VineethReddy02, line 162: Remove newline

pkg/collector/parser/receiver.go
```diff
@@ -155,10 +158,27 @@ func portName(receiverName string, port int32) string {
...
+	if err != nil {
+		return 0, err
+	}
+
+	host, port, err := net.SplitHostPort(u.Host)
+
```
VineethReddy02, line 168: remove newline

tests/e2e/endpoint-extraction/00-assert.yaml
VineethReddy02: @yuriolisa I don't think you need these tests as you added them into unit tests.

pkg/collector/parser/receiver.go
```diff
@@ -155,12 +157,29 @@ func portName(receiverName string, port int32) string {
...
+	var err error
+	var cport int64
+	var port int32
+
+	u, err := url.Parse(endpoint)
+	if err != nil {
+			return 0, err
+	}
+
+	host, hostport, err = net.SplitHostPort(u.Host)
```
VineethReddy02, line 170: Add `err` check

pkg/collector/parser/receiver.go
```diff
@@ -155,12 +157,29 @@ func portName(receiverName string, port int32) string {
...
 
+	if len(host) == 0 {
+			i := strings.LastIndex(endpoint, ":") + 1
+			var part = endpoint[i:]
+			cport, err = strconv.ParseInt(part, 10, 32)
```
VineethReddy02, line 175: `cport, err = strconv.ParseInt(endpoint[i:], 10, 32)`

pkg/collector/parser/receiver.go
```diff
@@ -155,9 +155,21 @@ func portName(receiverName string, port int32) string {
...
+	var err error
+	var cport string
+	var port int64
+
+	var r = regexp.MustCompile(":[0-9]+")
+
+	if r.MatchString(endpoint) {
+		cport = strings.Replace(r.FindString(endpoint), ":", "", -1)
+	}
+	port, err = strconv.ParseInt(cport, 10, 32)
```
pavolloffay, line 167: this looks weird. If there is no match the `strconv.ParseInt` is called on an empty `cport`?

pavolloffay: it is more idiomatic to directly return `0, nil`  if the match is false, if this is what you want to return

go.mod
```diff
@@ -97,12 +97,15 @@ require (
 	go.uber.org/multierr v1.6.0 // indirect
 	go.uber.org/zap v1.18.1 // indirect
 	golang.org/x/crypto v0.0.0-20210220033148-5ea612d1eb83 // indirect
-	golang.org/x/net v0.0.0-20210610132358-84b48f89b13b // indirect
+	golang.org/x/mod v0.5.1 // indirect
```
pavolloffay, line 100: Why all these `go.mod` changes? The PR is not changing any imports nor bumping dependencies. For clarity I would revert dependencies changes.

pkg/collector/parser/receiver.go
yuriolisa: @pavolloffay , I did it to get the `strconv.ParseInt` error return.

go.mod
yuriolisa: I've just reverted the changes.

pkg/collector/parser/receiver.go
```diff
@@ -155,9 +156,23 @@ func portName(receiverName string, port int32) string {
 }
 
 func portFromEndpoint(endpoint string) (int32, error) {
-	i := strings.LastIndex(endpoint, ":") + 1
-	part := endpoint[i:]
-	port, err := strconv.ParseInt(part, 10, 32)
+	var err error
+	var port int64
+
+	var r = regexp.MustCompile(":[0-9]+")
```
pavolloffay, line 162: nit: syntax `r := ....` is more used if the variable is initialized at creation

pavolloffay (approved):

pkg/collector/parser/receiver.go
yuriolisa: Just fixed.

pavolloffay (approved):

go.mod
```diff
@@ -119,4 +119,4 @@ require (
 	k8s.io/utils v0.0.0-20210819203725-bdf08cb9a70a // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.1.2 // indirect
 	sigs.k8s.io/yaml v1.2.0 // indirect
-)
+)
```
VineethReddy02, line 122: As you do not have a change in this file. I would fix the change trailing space or newline. To move this file out from list files changed.

VineethReddy02 (approved):
Needs some refactoring on error case evaluation.

pkg/collector/parser/receiver.go
```diff
@@ -155,9 +156,23 @@ func portName(receiverName string, port int32) string {
 }
 
 func portFromEndpoint(endpoint string) (int32, error) {
-	i := strings.LastIndex(endpoint, ":") + 1
-	part := endpoint[i:]
-	port, err := strconv.ParseInt(part, 10, 32)
+	var err error
```
VineethReddy02, line 159: You don't need to declare this `err`

pkg/collector/parser/receiver.go
```diff
@@ -155,9 +156,23 @@ func portName(receiverName string, port int32) string {
...
+	var port int64
+
+	r := regexp.MustCompile(":[0-9]+")
+
+	if r.MatchString(endpoint) {
+		port, err = strconv.ParseInt(strings.Replace(r.FindString(endpoint), ":", "", -1), 10, 32)
+	}
+
+	if err != nil {
```
VineethReddy02, line 168: Move error handling to inside if condition as err evaluation is inside the if loop.

VineethReddy02 (changes requested):

pkg/collector/parser/receiver.go
yuriolisa: done

VineethReddy02 (approved):

Issue #383, opened 2021-08-05
Will opentelemetry-operator support prometheus CRDs, like ServiceMonitor


ifelseend:
Hi, guys

It is hard to manager prometheus receiver configs when there are many scrape jobs.
I want to know is it a plan to support prometheus ServiceMonitor CRDs, so that we can manage prometheus receiver configs with a better way.

jpkrohling:
Perhaps a question for @alolita or @rakyll.

Aneurysm9:
This is definitely something we'd like to support.  I [added support upstream](https://github.com/prometheus-operator/prometheus-operator/pull/3972) so that we could re-use the config generation logic from the Prometheus operator.  I think the remaining work would be to cause the config reconciliation to happen when any of the *Monitor CRs are changed.  Since I'm not sure that the collector handles hot config reloading well (or at all) it might make sense to filter such changes through the recently added target allocation mechanism.  It does handle hot config reloads and separates the collector from target discovery.

jpkrohling:
>  Since I'm not sure that the collector handles hot config reloading well (or at all) it might make sense to filter such changes through the recently added target allocation mechanism. It does handle hot config reloads and separates the collector from target discovery.

The collector does not handle that. While I kinda understand why target allocator is doing that, I would rather do a regular Kubernetes rollout instead or restarting the container from the pod. This way, we protect the collector from crashing in case the config file can't be used (deprecated/new options).

tejaswiniVadlamudi:
Does OTel Operator define its own CRDs and also support migration from Prometheus Operator CRDs to OTel CRDs?
or 
Does OTel Operator support Prometheus Operator CRDs? => not inventing anything new

secustor:
Initially the second option is more likely  as we can reuse code from PrometheusOperator and their CRDs are widely used in the K8 ecosystem. 
A minimal implementation of this is right now under review. https://github.com/open-telemetry/opentelemetry-operator/pull/653

yuriolisa:
@pavolloffay, @jpkrohling, @Aneurysm9, do you have any thoughts about that issue? I've seen there are many suggestions but I would like to define the way in which we are going to support the PRom CRDs.

secustor:
I don't think this has been written down in it's entirety yet. 

The current implementation idea, is to support target discovery CRDs only ( PodMonitor, ServiceMonitor ). This part is already implemented in the TargetAllocator.    
[...]

PR #653, opened 2021-12-28
[targetallocator] PrometheusOperator CRD MVC

secustor:
Context: #383 

Currently implemented: 
- [x] fetch CRDs `ServiceMonitor` and `Podmonitor`
- [x] generate PromConfigs from monitor instances
- [x] update on creation, deletion and change of instances

TBD: 
- [x] clean up server shutdown
- [x] add flag to make PrometheusCR watcher optional
- [ ] add option to control watched namespaces
- [ ] add option to apply a label selector for monitors

Needs discussion: 
- [ ] splitting of application configuration and promconfig in the tracked config map

cc @Aneurysm9


Aneurysm9:
😻

secustor:
Splitting this here of as else the PR would get to big.

Aneurysm9 (approved):

jpkrohling:
@pavolloffay, @VineethReddy02, is one of you interested in reviewing this?

secustor:
Any chance we get this reviewed by a maintainer?      

I would like to propose that we use the operator/collector, but we heavely lean on PrometheusCRDs to configure our scrape targets.

VineethReddy02:
I took a quick pass through the PR. Looks good to me in the first pass. Can we add e2e tests for the same?

secustor:
@VineethReddy02  I have modified the KUTTL tests to use the local code and not the already published image.      
Further they now assure that the necessary arguments are set.

pavolloffay:
I would like to review, but I don't have any experience with the `otel-allocator`.

I am having a hard time understanding what `./cmd/otel-allocator` does. There is no readme and the CRD also does not have any meaningful docs https://github.com/open-telemetry/opentelemetry-operator/blob/0617cf7c75734930deebb951bf39dde1bfef48a5/apis/v1alpha1/opentelemetrycollector_types.go#L50. 

@secustor would it be possible to add/improve the docs?

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -120,11 +120,21 @@ type OpenTelemetryTargetAllocator struct {
 	// +optional
 	Enabled bool `json:"enabled,omitempty"`
 
+	// PrometheusCR defines the configuration for Prometheus custom resource retrieval.
```
pavolloffay, line 123: What CR is it exactly? could you please document the fully qualified name (group.kind)? (I am playing an ignorant here not knowing anything about p8s)

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -120,11 +120,21 @@ type OpenTelemetryTargetAllocator struct {
...
+	// +optional
+	PrometheusCR OpenTelemetryTargetAllocatorPrometheusCR `json:"prometheusCR,omitempty"`
+
 	// Image indicates the container image to use for the OpenTelemetry TargetAllocator.
 	// +optional
 	Image string `json:"image,omitempty"`
 }
 
+type OpenTelemetryTargetAllocatorPrometheusCR struct {
+	// Enabled indicates whether to use a Prometheus custom resources as target or not.
```
pavolloffay, line 133: Which p8s CR does it use? Can there be multiple in a namespace?

apis/v1alpha1/opentelemetrycollector_types.go
secustor: It's referring to the `ServiceMonitor` and `PodMonitor` CRDs in the `monitoring.coreos.com/v1` API

apis/v1alpha1/opentelemetrycollector_types.go
secustor: As this is a MVC all instances of `ServiceMonitor` and `PodMonitor` in all namespaces the service account is able to retrieve from the API server

secustor:
> I am having a hard time understanding what ./cmd/otel-allocator does. There is no readme and the CRD also does not have any meaningful docs 

The short explanation would be that it is an endpoint for the HTTP_SD of Prometheus, which shards the targets for multiple collector instances.

> would it be possible to add/improve the docs?

Sure, I can do that. 

I have been working from the links which @Aneurysm9 has provided me in this comment https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/6344#discussion_r759518685     

- https://docs.google.com/document/d/1YRFIZl-N0Igr1uQQ_MVm10JX2-9hy9OItNDU2PJ-eiE/edit#heading=h.w6m58i6u3fzh
- https://docs.google.com/document/d/1GdpCqdkz72291l_8S-jio7EhVltminobvqshsFih6vA/edit#heading=h.256cyy7pbquw

secustor:
During documenting the TargetAllocator, I have come across an issue regarding this implementation.     

Currently the Operator does not use the `/jobs` endpoint exposed  by the TargetAllocator, but instead uses only the jobs defined in the PrometheusReceiver configuration. Therefore all jobs outside this configuration are not added to the Collector configs.     
https://github.com/open-telemetry/opentelemetry-operator/blob/d965ac3127a61338418a17567de24bb971cc1a15/pkg/collector/reconcile/config_replace.go

I see following possibilities to move forward, expect dropping this feature entirely: 
1. Call the `/jobs` endpoint from Operator.           
This is not ideal as this would introduce a requirement that the TargetAllocator has to be reachable by the operator. I think this would be the most flexible solution regarding the addition of new watcher in the TargetAllocator or BYO target allocators.
2. Add a catch all job for PrometheusCR jobs if this feature is enabled.       
In that case the metrics generated would not be the same as the Prometheus instances managed by the PrometheusOperator as the `job` label would differ between the two implementations. 
3. Move the PrometheusCRD ( ServiceMonitor and PodMonitor ) watcher into the operator and only provide the resolved configuration to the TargetAllocator     
This can lead to high load on big clusters, as the operator would watch generate configs for all Collector instances
4. Create an [Observer](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/observer) for the TargetAllocator and extend the ReceiverCreator to allow looping over observer results.     
This seems to be cleanest solution IMO.

I suggest to use the first solution and to pursue the last one as a long term goal. 

WDYT?

Aneurysm9:
Would the operator update the collector configuration and recycle the collector deployment upon identifying new jobs?  I don't think that the collector currently will reload pipelines when its configuration file changes.

What about having the `prometheus` receiver  watch the `/jobs` endpoint and update its configuration as appropriate.  This could be managed by a config flag to limit impact where the target allocator is not in use and would not require creating new receivers or reloading pipelines.

secustor:
No, the Collectors would have to be redeployed when CRs would be changed.    

That would be definitely the cleanest solution. 

I agree,  I'm going to create an issue if this PR is accepted.

jpkrohling:
@pavolloffay, @VineethReddy02 you've made a couple of comments to this PR, do you think this is ready to be merged?

VineethReddy02 (approved):
LGTM

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -120,11 +120,21 @@ type OpenTelemetryTargetAllocator struct {
 	// +optional
 	Enabled bool `json:"enabled,omitempty"`
 
+	// PrometheusCR defines the configuration for the retrieval of PrometheusOperator CRDs ( servicemonitor.monitoring.coreos.com/v1 and podmonitor.monitoring.coreos.com/v1 )  retrieval.
```
pavolloffay, line 123: question as before, from which namespaces are these objects queried?

cmd/otel-allocator/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Watchers
+Watchers are responsible for the translation of external sources into Prometheus readable scrape configurations and 
+triggers updates to the DiscoveryManager
+
+### DiscoveryManager
+Watches the Prometheus service discovery for new targets and sets targets to the Allocator 
+
+### Allocator
+Shards the received targets based on the discovered Collector instances
```
pavolloffay, line 23: > based on the discovered 

Is this correct? Strictly speaking the TA does not discover collectors. The collectors should be known from the collector CR.

cmd/otel-allocator/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Collector
```
pavolloffay, line 25: Is this OTEL collector? Maybe we should find a different name if it is not.

cmd/otel-allocator/watcher/promOperator.go
```diff
@@ -0,0 +1,155 @@
+package watcher
+
+import (
+	"fmt"
+
+	"github.com/go-kit/log"
+	"github.com/go-logr/logr"
+	allocatorconfig "github.com/otel-allocator/config"
```
pavolloffay, line 8: is this import from this repository? Then it should be moved to a 3rd group. Why doesn't it have the full path?

cmd/otel-allocator/README.md
```diff
@@ -0,0 +1,82 @@
+# Target Allocator
+
+The TargetAllocator is an optional separately deployed component of an OpenTelemetry Collector setup, which is used to
+distribute targets of the PrometheusReceiver on all deployed Collector instances.
+
+# Design
+
+If the Allocator is activated, all Prometheus configurations will be transferred in a separate ConfigMap which get in
```
pavolloffay, line 8: > all Prometheus configurations

Is it OTEL p8s receiver configuration or as well p8s CRs?

cmd/otel-allocator/README.md
```diff
@@ -0,0 +1,82 @@
...
+turn mounted to the Allocator.    
+This configuration will be resolved to target configurations and then split across all OpenTelemetryCollector instances.
+
+TargetAllocators exposes the results as [HTTP_SD endpoints](https://prometheus.io/docs/prometheus/latest/http_sd/)
```
pavolloffay, line 12: How does the OTEL colletor get the final config (resolved targets)? 

Does the OTEL operator configure the collector to use the http_sd to get the targets?

pavolloffay:
Sorry for super long delay in reviewing this. I had to take some time to lear about target allocator.

cmd/otel-allocator/README.md
Aneurysm9: It does watch the associated stateful set to observe scaling events and reallocate targets as necessary when the size of the set of collectors changes.

cmd/otel-allocator/README.md
Aneurysm9: This applies to the configuration of the `prometheus` receiver in the collector configuration.  The Prometheus operator CRs are observed separately after the target allocator has been started, at which point it further modifies the service discovery configuration it is using.

cmd/otel-allocator/README.md
Aneurysm9: Yes, the OTEL operator rewrites the collector `prometheus` receiver config to use `http_sd` for all configured jobs, pointing at the target allocator instance for that collector.

There is still a gap, that @secustor has proposed to close with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/8055.  This addresses the fact that the Prometheus operator CRs can cause new jobs to be exposed at the target allocator but the collector configuration is static and does not currently have a mechanism for discovering those jobs.  The target allocator exposes a job list resource that should suffice to discover new/ended jobs.

cmd/otel-allocator/watcher/promOperator.go
secustor: Yes, this is referencing the local module. 

Why this module name has been chosen, I can not say.

cmd/otel-allocator/README.md
secustor: Yes, that's correct. The operator configures the collector to use http_sd endpoints which are provided by the TargetAllocator.

Currently the Operator defines the p8s jobs which the collector should query, therefore no p8s operator CR defined jobs are added to the collector. Later on the the collector should query the jobs directly from the TargetAllocator. 
See following issues: 
- operator: https://github.com/open-telemetry/opentelemetry-operator/issues/713
- collector: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/7944

cmd/otel-allocator/watcher/promOperator.go
Aneurysm9: I think this import path is incorrect and should be changed.  Probably best to change it in a separate PR, though.

templarfelix (approved):

codeboten (commented):
@pavolloffay it appears your comments have been addressed, can you confirm whether there are still outstanding items to address?

cmd/otel-allocator/config/config.go
```diff
@@ -2,31 +2,54 @@ package config
...
+	ClusterConfig  *rest.Config
+	// KubeConfigFilePath empty if in cluster configuration is in use
+	KubeConfigFilePath string
+	RootLogger         logr.Logger
+	PromCRWatcherConf  PrometheusCRWatcherConfig
+}
+
+func Load(file string) (Config, error) {
 	var cfg Config
 	if err := unmarshal(&cfg, file); err != nil {
```
jaronoff97, line 54: question: It seems previously the allocator would load a default config file if no file was provided, it seems that was because we would always pass in an empty string in `main.go`. Should that functionality remain in case there any consumers of this function?

jaronoff97 (approved):
Love the approach here, one non-blocking question. I see you're adding the changes to the end-to-end tests, would it be possible to add a test involving a prometheus CRD to gut check that the expected configuration gets generated?

secustor:
>  see you're adding the changes to the end-to-end tests, would it be possible to add a test involving a prometheus CRD to gut check that the expected configuration gets generated?
I don't think it is possible right now to test API responses with KUTTL.

cmd/otel-allocator/config/config.go
secustor: Are you referring to users of the TargetAllocator or Devs which import this package? 

In the first case the behavior has not changed. It is still loading the same default file as before. This is implement using pflags now. https://github.com/open-telemetry/opentelemetry-operator/pull/653/files#diff-d292758014333553bcd82dfdf75743ef82b0b32f6414e2dc345777fc342e4480R77

pavolloffay (approved):
I am approving this PR, however I am still not familiar with the target-allocator and it is not area where I focus.

I am fine merging this if there will be people maintaining it.

secustor:
@pavolloffay Can you merge this? All checks are successful.

pavolloffay:
Merging based on the support discussion in https://cloud-native.slack.com/archives/C033BJ8BASU/p1650372898413199. 

@secustor @Aneurysm9 (AWS) are willing to support the target-allocator.

PR #449, opened 2021-10-05
Add flags to override hardcoded images

pavolloffay:
Resolves #447 

Signed-off-by: Pavol Loffay <p.loffay@gmail.com>

PR #662, opened 2022-01-12
Parse flags before using them in config

rubenvp8510:
Signed-off-by: Ruben Vargas <ruben.vp8510@gmail.com>

`pflag.Parse()` Should be called before we use the arguments to initialize the config structure. If not, the variables won't contain the argument values, and the operator won't honor the argument values.

Also, removed the config flags, as this is not used anymore since this PR https://github.com/open-telemetry/opentelemetry-operator/pull/449 migrated the flags to `main.go`

### DIFF
--- a/internal/config/main.go
+++ b/internal/config/main.go
@@ -19,7 +19,6 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
-	"github.com/spf13/pflag"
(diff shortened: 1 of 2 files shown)


pavolloffay (approved):
LGTM, 

it would be great to have a test, but that would either require moving parts of the main to a separate package or writing a custom e2e test.

PR #670, opened 2022-01-19
Fix collector config update

### LINKED ISSUE #460: Restart collector on config change

mcariapas:
I added check sum to pod annotations. This will fix rollout when the collector config changes.
Yes, I know this is duplicated pull request with #506 
 but the other one has been forgotten for some time.
 
 Fixes #460


mcariasap:
@pavolloffay Can you review it?

apis/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,4 +1,3 @@
-//go:build !ignore_autogenerated
```
VineethReddy02, line 1: Why is this removed?

go.sum
```diff
@@ -27,22 +27,30 @@ cloud.google.com/go/bigquery v1.3.0/go.mod h1:PjpwJnslEMmckchkHFfq+HTD2DmtT67aNF
 cloud.google.com/go/bigquery v1.4.0/go.mod h1:S8dzgnTigyfTmLBfrtrhyYhwRxG72rYxvftPBK2Dvzc=
 cloud.google.com/go/bigquery v1.5.0/go.mod h1:snEHRnqQbz117VIFhE8bmtwIDY80NLUZUMb4Nv6dBIg=
 cloud.google.com/go/bigquery v1.7.0/go.mod h1://okPTzCYNXSlb24MZs83e2Do+h+VXtc4gLoIoXIAPc=
+cloud.google.com/go/bigquery v1.8.0 h1:PQcPefKFdaIzjQFbiyOgAqyx8q5djaE7x9Sqe712DPA=
```
VineethReddy02: I do not see any go import changes in this PR. Why do we need these changes?

apis/v1alpha1/zz_generated.deepcopy.go
mcariapas: It was generated automatically....

go.sum
mcariapas: I don't know how go mode works in background, but when I tested it I downloaded all dependencies. I can remove these changes from the PR.

apis/v1alpha1/zz_generated.deepcopy.go
mcariapas: I can remove it from the PR if it is necessary

pkg/collector/annotations.go
```diff
@@ -37,10 +37,18 @@ func Annotations(instance v1alpha1.OpenTelemetryCollector) map[string]string {
...
+
+	if nil != instance.Spec.PodAnnotations {
+		for k, v := range instance.Spec.PodAnnotations {
+			podAnnotations[k] = v
+		}
+	}
+
 	// make sure sha256 for configMap is always calculated
 	annotations["opentelemetry-operator-config/sha256"] = getConfigMapSHA(instance.Spec.Config)
+	podAnnotations["opentelemetry-operator-config/sha256"] = getConfigMapSHA(instance.Spec.Config)
```
pavolloffay, line 49: Get the hash only once and use it twice

pkg/collector/annotations.go
```diff
@@ -22,9 +22,9 @@ import (
 )
 
 // Annotations return the annotations for OpenTelemetryCollector pod.
-func Annotations(instance v1alpha1.OpenTelemetryCollector) map[string]string {
+func Annotations(instance v1alpha1.OpenTelemetryCollector) (map[string]string, map[string]string) {
```
pavolloffay, line 25: I liked more a separate function  `PodAnnotations` from the original PR. 

Please also update the godocs.

go.sum
pavolloffay: As @VineethReddy02 mentioned there should not be changes to this file as the PR does not change any dependencies.

pavolloffay (commented):
Thanks for looking into this and it looks good to me ❤️ 

I would love to see e2e test for this change (something like this)
1. deploy collector as deployment
2. assert it is deployed
3. change config in CR - e.g. add new receiver
4. assert the collector pod restarted with the new changes - check if the new receiver port is exposed.

tests/e2e/smoke-restarting-deployment/00-errors.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: restarting-collector
+spec:
+  ports:
+  - targetPort: 4317
```
pavolloffay, line 7: Does this file define that the service should not expose 4317 port?

pavolloffay (approved):

tests/e2e/smoke-restarting-deployment/00-errors.yaml
mcariapas: yes, because we use it when we change the configuration and we have to check if it is not already.

mcariapas:
Can we merge it before the next release?

pavolloffay:
yes, thanks for the PR :) 

cc) @jpkrohling @VineethReddy02 could you please merge?

PR #683, opened 2022-01-31
[WIP] auto-instrumentation: Allow per container control

### LINKED ISSUE #529: Handle case where multiple instrumentations can be applied to a pod
Currently, both `inject-java` and `inject-nodejs` can be applied to the same pod. This creates some awkwardness, especially if the instrumentation instance referenced is different, common SDK configuration will be applied from two unrelated instrumentations.

One approach that could solve this is to split the current annotations into "whether to inject" and "what language to inject"

```
auto.opentelemetry.io/inject = "true"
auto.opentelemetry.io/lanuage = "java"
```

_Originally posted by @anuraaga in https://github.com/open-telemetry/opentelemetry-operator/pull/507#r745427307_

### LINKED ISSUE #618: Auto instrumentation injects into wrong container
Currently, Autoinstrumentation seems to inject into the first container in the list, which is especially troublesome if used with istio, which injects its proxy ahead of the normal container.

[While injection into the first container](https://github.com/open-telemetry/opentelemetry-operator/blob/ae8f3a852623f465a6c1802d03ffa31651bf0888/pkg/instrumentation/javaagent.go#L31) is surely a good default, an optional annotation could be added which contains the name(s) of containers to inject into

BronzeDeer:
NOTE:  Due to the changes in injection behaviour I have only included the javaagent changes for early review. If the java changes are greenlit, I will add python and nodejs accordingly

fixes #618
fixes #529

Currently, it is only possible to inject a single autoinstrumentation
per pod and it will always be injected into the first container and
first container only. This is especially troublesome if a mutating
webhook will be injecting additional sidecars (e.g. Istio) which will
often receive the instrumentation instead of the main container.
Furthermore, for multi-container, multi-language pods only one container
can currently be auto-instrumented.

This commit introduces a new set of annotations of the form
`instrumentation.opentelemetry.io/inject-<language>-container-names`,
which tells the operator which containers should receive which language
instrumentation. The value of the annotation should be a comma-delimited
list of container-names to inject into. If the annotation is not
present, injection will use the default container-selection (currently
first container). The annotation will be ignored if the corresponding
language annotation `instrumentation.opentelemetry.io/inject-<language>`
is not present or set to enable instrumentation for `<language>`.

Note: It is now possible that instrumentation for multiple languages
needs to be prepared for the same pod. To avoid clashes, the volumes,
volume mount paths and initContainers for each language have been
postfixed with `-<language>`. Furthermore, to avoid ballooning the
injection logic complexity, the initContainer and volume for a language
will always be added to the Pod as long as injection for that language
is enabled, even if the injection into every container fails.
Previously, with only one language and one container to inject the
operator would only inject the volume and initContainer if the container
could successfully be modified.

Signed-off-by: Joel Pepper <pepper@bronze-deer.de>

### DIFF
--- a/pkg/instrumentation/javaagent.go
+++ b/pkg/instrumentation/javaagent.go
@@ -15,6 +15,9 @@
 package instrumentation
 
 import (
+	"path"
+	"strings"
+
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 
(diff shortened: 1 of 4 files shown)


BronzeDeer:
The changes around volumes and initContainers required updates to the existing tests, I took the liberty to also replace the existing string literals with the corresponding const variables to make the tests less brittle to future changes in names or paths

BronzeDeer:
On the topic of tests: It seems there is a lot of duplication in the tests of the individual <language>_test, podmutator_test and sdk_test. When I add tests for the new cases, where should I add which cases?

BronzeDeer:
@pavolloffay could you give me a quick feedback on the test question and change in injection behaviour (see above) so I can move this PR along? We've even had duplicate work start croping up, since this stuck in limbo

BronzeDeer:
Currently on hold in favor of #689 which implements the simpler special case, see discussion in https://github.com/open-telemetry/opentelemetry-operator/pull/689#issuecomment-1047566558. Reopen if need for the more general solution arises.

PR #687, opened 2022-02-02
Update nodejs instrumentation to 0.27.0

anuraaga:
And runs build on PR

### DIFF
--- a/autoinstrumentation/nodejs/tsconfig.json
+++ b/autoinstrumentation/nodejs/tsconfig.json
@@ -19,6 +19,7 @@
     "noImplicitReturns": true,
     "noUnusedLocals": true,
     "pretty": true,
+    "skipLibCheck": true,
     "sourceMap": true,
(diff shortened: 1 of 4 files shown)


autoinstrumentation/nodejs/tsconfig.json
```diff
@@ -19,6 +19,7 @@
     "noImplicitReturns": true,
     "noUnusedLocals": true,
     "pretty": true,
+    "skipLibCheck": true,
```
anuraaga, line 22: By default, TS does a deep typecheck through transitive dependencies. I don't know if it's a but in opentelemetry-js (/cc @dyladan) but after updating I got a transitive type failure on graphql library.

For our meager code, the transitive type check is definitely not needed.

https://www.typescriptlang.org/tsconfig#skipLibCheck

pavolloffay (approved):

PR #689, opened 2022-02-04
Choose target container injection with annotation

### LINKED ISSUE #618: Auto instrumentation injects into wrong container

fscellos:
This PR address the same problem of the following PR (https://github.com/open-telemetry/opentelemetry-operator/pull/683) without introducing big changes in codebase.
I just introduce a new annotation near existing auto-inject instrumenation that allow to search first for a corresponding containername before fallback to the first one by default : 
   

```
instrumentation.opentelemetry.io/container-name: autodeploy
instrumentation.opentelemetry.io/inject-java: "true"
```

This approach present the advantage to let end user to define exact injection target (as is some cases, cluster administrator won't want user to manipulate Instrumentation ressources to keep control on sampling rate)

Resolves #618


VineethReddy02:
@pavolloffay can you review these changes as they are around auto-instrumentation?

pavolloffay:
@fscellos does this PR implement the discussed approach from 
https://github.com/open-telemetry/opentelemetry-operator/issues/529#issuecomment-1017365664 ?

fscellos:
Hello. Yes i think so but in a simplest way. 
The simple case i wan't to adress is the common case of a pod that contains multiple containers (in particular istio-proxy that always get the 0-position) and i want to precise to instrumentation on which container to inject env variables.
I try it on a real application with istio enabled with success but i notice i have some other properties to update too (the container tag name for span is always set to the first container, not the one i choose)

pavolloffay:
This PR is missing e2e test, this is a major feature that deserves it.

pkg/instrumentation/sdk.go
```diff
@@ -56,39 +56,48 @@ type sdkInjector struct {
 	client client.Client
 }
 
-func (i *sdkInjector) inject(ctx context.Context, insts languageInstrumentations, ns corev1.Namespace, pod corev1.Pod) corev1.Pod {
+func (i *sdkInjector) inject(ctx context.Context, insts languageInstrumentations, ns corev1.Namespace, pod corev1.Pod, containerName string) corev1.Pod {
```
pavolloffay, line 59: `chooseServiceName(pod corev1.Pod, resources map[string]string) string {` 

has to be updated as well to correctly choose the service name.


`createResourceMap(ctx context.Context, otelinst v1alpha1.Instrumentation, ns corev1.Namespace, pod corev1.Pod) map[string]string {`

has to be as well updated to choose the right container.

fscellos: Ok thank you for your feeback. I Will take a closer on 21 Feb week and also write one kuttl test.

pavolloffay:
@anuraaga @BronzeDeer could you please take a look at the proposal in this PR? 

The proposal is to basically add an optional annotation e.g. `instrumentation.opentelemetry.io/container-name: app` to specify the container for injection. If annotation is not present injection to the first container is applied. 

The original proposal from https://github.com/open-telemetry/opentelemetry-operator/issues/529#issuecomment-1017365664 allows specifying container name per language.

BronzeDeer:
This PR solves the problem for the special case of injecting into a single container that is not the first, while the suggestion from #529 and PR #683 is the more general solution which allows to configure injection of arbitrary languages into arbitrary many containers at the cost of a slightly increased complexity (allthough not much IMO, its basically filter and map over the container list with the existing injection function, you could even argue that it serves to decouple the injection logic from pod modification logic)

Which solution is the correct one in the long-term probably comes down more to whether the idea of injecting multiple containers with potentially different languages in a single pods is a probable use case or falls under YAGNI. Making that call likely falls solely on the shoulders of the maintainers/stewards of the project

BronzeDeer:
@pavolloffay @anuraaga could you comment on the design decision between the simple and more general case? And if you decide we stick with the more general approach, could you answer my two small questions in #683, so I can get that PR ready?

pavolloffay:
I am fine to move this forward. Right now we don't have any requests to support instrumenting a pod with 2 different languages  (e.g. java and nodejs containers). If this requirements comes we could introduce more specific container name annotation per language: 

```
instrumentation.opentelemetry.io/inject-java-container-names = "spring-petclinic"
```

README.md
```diff
@@ -215,6 +215,14 @@ The possible values for the annotation can be
 * `"my-instrumentation"` - name of `Instrumentation` CR instance.
 * `"false"` - do not inject
 
+It as also possible in case of multiple-containers Pods (like Istio sidecars) to specify the name of the container in which you want to inject OTEL auto-instrumentation (by default the first container is choosen)
+
+For this add this annotation to your Pod Specification with target container name :
+
+```bash
+instrumentation.opentelemetry.io/container-name: "name_of_your_target_container"
```
pavolloffay, line 223: @fscellos could you please rename it to `instrumentation.opentelemetry.io/container-names` and support list as well?

fscellos: Hello @pavolloffay. I will try to do it at worst next week, yes.

fscellos:
Hello @pavolloffay, not now.
I keep it in mind but have to deal with one other priorities before. I'll keep you inform.

fscellos:
Hello @pavolloffay : i can begin to work again on it again. I'll keep you inform.

README.md
```diff
@@ -215,6 +215,37 @@ The possible values for the annotation can be
 * `"my-instrumentation"` - name of `Instrumentation` CR instance.
 * `"false"` - do not inject
 
+It as also possible in case of multiple-containers Pods (like Istio sidecars) to specify the name of the container in which you want to inject OTEL auto-instrumentation (by default the first container is choosen)
+
+For this add this annotation to your Pod Specification with target container name :
```
pavolloffay, line 220: > Use the following annotation on `Pod` template to specify container names for auto-instrumentation injection:

README.md
```diff
@@ -215,6 +215,37 @@ The possible values for the annotation can be
...
+
+```bash
+instrumentation.opentelemetry.io/container-names: "name_of_your_target_container"
+```
+In case of injection on multiple containers, you can indicates all your containers'name separate with comma.
```
pavolloffay, line 225: > Multiple container names can be specified, for instance `app,sidecar`.

README.md
```diff
@@ -215,6 +215,37 @@ The possible values for the annotation can be
...
+
+#### Use customized or vendor instrumentation
```
pavolloffay, line 227: This change should not be here, maybe the PR was not rebased properly?

tests/e2e/instrumentation-nodejs-multicontainer/NAR.YAML
```diff
@@ -0,0 +1,361 @@
+apiVersion: v1
```
pavolloffay, line 1: Is this file intentional here? It does not seem to follow kuttl name conventions.

pkg/instrumentation/sdk.go
```diff
@@ -56,39 +56,48 @@ type sdkInjector struct {
 	client client.Client
 }
 
-func (i *sdkInjector) inject(ctx context.Context, insts languageInstrumentations, ns corev1.Namespace, pod corev1.Pod) corev1.Pod {
+func (i *sdkInjector) inject(ctx context.Context, insts languageInstrumentations, ns corev1.Namespace, pod corev1.Pod, containerName string, init bool) corev1.Pod {
```
pavolloffay, line 59: Why do we need to pass the last `init` parameter? I don't think it is needed.

pavolloffay:
@fscellos thanks for working on this, I have left some comments. I would like to move this PR forward, let me know if you have time or I should take over.

fscellos:
Hello @pavolloffay.
I will take a closer look on it next monday.

pkg/instrumentation/sdk.go
fscellos: I do it to keep code as similar as possible with existing one. 
As i call multiple time the injectXXXAgent functions and as this functions also contains the injection of Volume that host agent and init container that copy this agent, i use this simple boolean to avoid this injection multiple time.
The creation of volume and initContainers is only do one time for the first eligible container of injection.

pavolloffay: The function takes pod that can be used to verify what needs to be ingected

tests/e2e/instrumentation-nodejs-multicontainer/NAR.YAML
fscellos: No i remove it. Sorry.
Il also corrected the second it test of python-multicontainer

pkg/instrumentation/sdk.go
fscellos: You are right i don't think to use it. 
I will check existence of init containers and volumes by adding anoter function i will call from injectXXXAgent.

fscellos:
Hello @pavolloffay.
I push asked modification and let you check if all is ok for you.

jpkrohling:
Tests started.

fscellos:
Hello @pavolloffay , @jpkrohling. I correct the two linting problems. 
Sorry i miss it on local before my commit as i didn't know i have to launch it explicitly

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
-  - manager.yaml
+- manager.yaml
```
pavolloffay, line 2: this file shouldn't be changed in this PR

README.md
pavolloffay: This section should not be added in this PR

pavolloffay:
@fscellos please clean the PR - readme and kustomization.yaml and also the lint failed. Otherwise it looks good and we an merge it.

README.md
fscellos: OK. I didn't understand you don't want any modifications in documentation. 
I remove all of this and get Readme from main branch of main repo.

config/manager/kustomization.yaml
fscellos: Yes. I don't understand what happens locally. I keep file from main branch of principal repository.
Apologize.

fscellos: It don't know why but it seems that something in prepare-e2e (or e2e) targets, something modify this file.
I didn't commit it again.

README.md
```diff
@@ -258,6 +258,8 @@ The OpenTelemetry Operator *might* work on versions outside of the given range,
 
 | OpenTelemetry Operator | Kubernetes           | Cert-Manager         |
 |------------------------|----------------------|----------------------|
+| v0.48.0                | v1.19 to v1.23       | 1.6.1                |
```
pavolloffay, line 261: There should not be changes in this support table.

pavolloffay (commented):
PR needs to be rebased and we should add the annotation to the root readme.

README.md
fscellos: I rebased. I will let you add documentation on your convenience (I already added it then remove)

pavolloffay:
@fscellos now the CI is failing :/

fscellos:
Hello @pavolloffay . Yes i saw that. I would check it but impossible to pull any image from github repo this morning (so all e2e failed on my side too). 
I will recheck it in a few hours.

fscellos:
Hello @pavolloffay . 
No chance, but i confirm you that we probably encountered a problem with github registry for previous build.
I rerun to launch e2e test and all seems to be ok now without any modifications on code; it is possible for you to relaunch e2e tests on CI ?
image

pavolloffay:
@fscellos I have restarted failed jobs. 

However the PR needs to be rebased again :/.

fscellos:
Hello @pavolloffay. Rebase done.
I launch e2e tests on kind cluster 1.19 & 1.23 locally before.
I hope it is last rebase/commit :-).. On May i will have conferences to see in valence (i see you'll have a session there) and get less time ;-)

fscellos:
Hello @pavolloffay : any news on review ?

pavolloffay (approved):
@fscellos great work! 

I think we have lost the readme documentation for this feature during the rebasing. I will merge it without the docs, but please open a follow up PR to document this.

fscellos:
Ok. I think i didin't understand one of your remarks because i explicitlty remove documention about this feature. 
I will open another PR for documentation monday 16.

Issue #690, opened 2022-02-06
Uses "ClusterFirstWithHostNet" DNS policy when `hostNetwork` is set to true


gai6948:
Per the current operator, when trying to deploy otel-collector as agent (using daemonset), setting to `hostNetwork` is required for pods running on same host to reach the agent, but then communication to otel-collector deployed as gateway inside the cluster fails because of DNS error. Upon investigation it turns out pods using `hostNetwork` should also set `dnsPolicy` to `ClusterFirstWithHostNet`. https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy

Can we make the operator aware of this logic?

My current OpenTelemetryCollector manifest:
```
apiVersion: opentelemetry.io/v1alpha1
kind: OpenTelemetryCollector
metadata:
  name: otel-agent
  namespace: opentelemetry-agent
spec:
  mode: daemonset
  hostNetwork: true
  image: otel/opentelemetry-collector-contrib:0.43.0
  resources:
    limits:
      cpu: 200m
      memory: 128Mi
    requests:
      cpu: 100m
      memory: 64Mi
  config: |
    receivers:
      otlp:
        protocols:
          grpc:
    processors:
      # The k8sattributes in the Agent is in passthrough mode
      # so that it only tags with the minimal info for the
      # collector k8sattributes to complete
      k8sattributes:
        passthrough: true
      memory_limiter:
        check_interval: 1s
        limit_percentage: 50
        spike_limit_percentage: 30
    extensions:
      memory_ballast:
        size_in_percentage: 20    
    exporters:
      otlp:
        endpoint: "otel-collector-collector.opentelemetry-collector.svc.cluster.local:4317"
        tls:
          insecure: true      
      logging:

    service:
      pipelines:
        traces:
          receivers: [otlp]
          processors: [k8sattributes, memory_limiter]
          exporters: [otlp, logging]
```

Daemonset that was actually created by the operator:
```
      restartPolicy: Always
      terminationGracePeriodSeconds: 30
      dnsPolicy: ClusterFirst
      serviceAccountName: otel-agent-collector
      serviceAccount: otel-agent-collector
      hostNetwork: true
```

slenky:
@pavolloffay using that we could reference agent address:
```
            - name: HOST_IP
              valueFrom:
                fieldRef:
                  fieldPath: status.hostIP
            - name: SPRING_ZIPKIN_BASEURL
[...]

PR #691, opened 2022-02-06
Match pod `dnsPolicy` to `hostNetwork` config

gai6948:
Resolves https://github.com/open-telemetry/opentelemetry-operator/issues/690

I added a simple check to make the pods' DNSPolicy matching the `hostNetwork` option

### DIFF
--- a/pkg/collector/utils.go
+++ b/pkg/collector/utils.go
@@ -1,0 +1,29 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package collector
+
+import (
+	corev1 "k8s.io/api/core/v1"
+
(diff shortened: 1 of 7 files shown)


pkg/collector/utils.go
```diff
@@ -0,0 +1,29 @@
...
+package collector
+
+import (
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
+)
+
+func getDnsPolicy(otelcol v1alpha1.OpenTelemetryCollector) corev1.DNSPolicy {
+	dnsPolicy := corev1.DNSClusterFirst
```
pavolloffay, line 24: Shouldn't this be rather "Default" ? 

https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy

pavolloffay (commented):
Please add test to validate this.

pkg/collector/utils.go
gai6948: Found this section on that page:
```
Note: "Default" is not the default DNS policy. If dnsPolicy is not explicitly specified, then "ClusterFirst" is used.
```

gai6948:
Actually I don't know if `hostNetwork=true` makes sense for deployment and statefulsets, if it isn't perhaps we only need to test on daemonsets and not needing to duplicate such simple tests

pavolloffay:
> Actually I don't know if hostNetwork=true makes sense for deployment and statefulsets, if it isn't perhaps we only need to test on daemonsets and not needing to duplicate such simple tests

This is why we should understand why people want to use `hostNetwork`

pavolloffay:
@gai6948 the CI is still failing? Please fix it to move this forward.

gai6948:
The tests should pass now

pavolloffay (approved):

PR #704, opened 2022-02-10
Add appProtocol for otlp and jaeger receiver parsers

binjip978:
Signed-off-by: Sergei Semenchuk <pdp.eleven11@gmail.com>

### DIFF
--- a/pkg/collector/parser/receiver_generic.go
+++ b/pkg/collector/parser/receiver_generic.go
@@ -25,11 +25,13 @@ var _ ReceiverParser = &GenericReceiver{}
 
 // GenericReceiver is a special parser for generic receivers. It doesn't self-register and should be created/used directly.
 type GenericReceiver struct {
-	logger      logr.Logger
-	name        string
-	config      map[interface{}]interface{}
-	defaultPort int32
-	parserName  string
+	logger             logr.Logger
+	name               string
+	config             map[interface{}]interface{}
+	defaultPort        int32
+	defaultProtocol    corev1.Protocol
+	defaultAppProtocol *string
+	parserName         string
 }
 
 // NOTE: Operator will sync with only receivers that aren't scrapers. Operator sync up receivers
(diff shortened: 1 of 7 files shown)


pavolloffay (commented):
LGTM, 

if you could fix as well Zipkin that would be great, it's often times used in service mesh. 

It would be great as well improve e2e test and assert the service e.g. https://github.com/open-telemetry/opentelemetry-operator/blob/main/tests/e2e/smoke-simplest/00-assert.yaml

VineethReddy02:
Few tests are failing in the CI more details: https://github.com/open-telemetry/opentelemetry-operator/runs/5137930906?check_suite_focus=true#step:7:776

binjip978:
> Few tests are failing in the CI more details: https://github.com/open-telemetry/opentelemetry-operator/runs/5137930906?check_suite_focus=true#step:7:776

yeah, it's quite tricky to run the full suite on mac, but it should be fine now.

binjip978:
> LGTM,
> 
> if you could fix as well Zipkin that would be great, it's often times used in service mesh.
> 
> It would be great as well improve e2e test and assert the service e.g. https://github.com/open-telemetry/opentelemetry-operator/blob/main/tests/e2e/smoke-simplest/00-assert.yaml

for the zipkin part, I'm not really familiar with it and looks like it use GenericReceiver, so it's possible to setup TCP if the name is zipkin, but I'm not sure about appProtocol, it should be http, http2 or grcp? Looks like it depends https://zipkin.io/pages/tracers_instrumentation

pkg/collector/parser/receiver_generic.go
```diff
@@ -53,6 +53,16 @@ func (g *GenericReceiver) Ports() ([]corev1.ServicePort, error) {
 	}
 
 	if g.defaultPort > 0 {
+		if g.name == "zipkin" {
```
pavolloffay, line 56: I am not sure about this approach. The protocol type should be rather set in https://github.com/open-telemetry/opentelemetry-operator/blob/eb549cea898a53b84017cab05c6b51b3022df9aa/pkg/collector/parser/receiver_zipkin.go#L23 and then just used to configure the service

pavolloffay (commented):
Please add the e2e test- verify the service has the protocol set.

pkg/collector/parser/receiver_generic.go
binjip978: @pavolloffay just to be clear for zipkin config we can expect something like this:
```yaml
zipkin:
---
zipkin:
  protocols:
    http:
      endpoint: "0.0.0.0:9999"
```

And in case of default config "zipkin:" is this ok to add to GenericReceiver fields like defaultProtocol and defaultAppProtocol?

pavolloffay: I think we will have to add a field there

pavolloffay (approved):

PR #734, opened 2022-02-22
Support k8s 1.19 and higher

pavolloffay:

Resolves #734 

This PR changes CI to run tests on the lowest supported k8s version 1.19 and highest 1.23 

The PR essentially ensures that the operator is supported on a broader range of Kubernetes versions. We are committed to resource the support on k8s 1.19 and higher. 

The tests are run on 1.19 and the highest version, the versions in between does not have to be tested as there is no breaking change "in between". We are taking the same approach in other k8s operators.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -13,8 +13,7 @@ jobs:
     strategy:
       matrix:
        kube-version:
-       - "1.19" # long term support for OpenShift 4.6 based on k8s 1.19
-       - "1.21"
+       - "1.19"
        - "1.23"
 
     steps:
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
        kube-version:
-       - "1.19"
+       - "1.19" # long term support for OpenShift 4.9 based on k8s 1.19
        - "1.21"
        - "1.23"
 
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
        kube-version:
-       - "1.19" # long term support for OpenShift 4.9 based on k8s 1.19
+       - "1.19" # long term support for OpenShift 4.6 based on k8s 1.19
        - "1.21"
        - "1.23"
 
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -12,6 +12,9 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
+       # The e2e tests are run on the lowest and highest supported k8s version.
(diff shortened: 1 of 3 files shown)


.github/workflows/scorecard.yaml
```diff
@@ -13,10 +13,9 @@ jobs:
     strategy:
       matrix:
         kube-version:
-          - "1.19"
-          - "1.20"
+          - "1.19" # long term support for OpenShift 4.6 based on k8s 1.19
           - "1.21"
-          - "1.22"
+          - "1.23"
```
frzifus, line 18: why 1.21 and not 1.22?

frzifus (approved):

.github/workflows/e2e.yaml
```diff
@@ -13,8 +13,8 @@ jobs:
     strategy:
       matrix:
        kube-version:
+       - "1.19" # long term support for OpenShift 4.6 based on k8s 1.19
```
VineethReddy02, line 16: This looks opinionated to the RH use case. 

WDYT @jpkrohling?

pavolloffay: I am removing this.

pavolloffay:
@VineethReddy02 @jpkrohling could you please review?

README.md
```diff
@@ -230,10 +230,6 @@ When a custom `Spec.Image` is used with an `OpenTelemetryCollector` resource, th
 
 We strive to be compatible with the widest range of Kubernetes versions as possible, but some changes to Kubernetes itself require us to break compatibility with older Kubernetes versions, be it because of code incompatibilities, or in the name of maintainability.
 
-Our promise is that we'll follow what's common practice in the Kubernetes world and support N-2 versions, based on the release date of the OpenTelemetry Operator.
```
jpkrohling, line 233: I think this has to be better discussed. Could you please open an issue to discuss this?

VineethReddy02 (approved):

README.md
jpkrohling: Thanks, I added a couple of comments there.

.github/workflows/e2e.yaml
```diff
@@ -13,8 +13,7 @@ jobs:
     strategy:
       matrix:
        kube-version:
-       - "1.21"
-       - "1.22"
+       - "1.19"
```
jpkrohling, line 16: So, no explicit test on 1.21 and 1.22? It might be good to document the reasoning.

README.md
jpkrohling: The outcome of the discussion should be documented here.  Like, "every released operator will support a specific range of Kubernetes versions, to be determined at the latest during the release."

pavolloffay: good idea I will add there to the readme.

.github/workflows/e2e.yaml
pavolloffay: added a comment to the workflow

jpkrohling (approved):

PR #737, opened 2022-02-23
Use default github token to publish the release

pavolloffay:

Right now the release process is using @jpkrohling personal token (https://github.com/open-telemetry/opentelemetry-operator/releases). It works but it confused me that he actually explicitly created the release I was working on.

Using standard token should be preferred as we don't care about e.g. expiration.

### DIFF
--- a/.ci/create-release-github.sh
+++ b/.ci/create-release-github.sh
@@ -1,7 +1,6 @@
 #!/bin/bash
 
 OPERATOR_VERSION=$(git describe --tags)
-echo "${GH_WRITE_TOKEN}" | gh auth login --with-token
 
 gh config set prompt disabled
 gh release create \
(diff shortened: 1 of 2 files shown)


frzifus (approved):

VineethReddy02 (approved):

.ci/create-release-github.sh
```diff
@@ -1,7 +1,6 @@
 #!/bin/bash
 
 OPERATOR_VERSION=$(git describe --tags)
-echo "${GH_WRITE_TOKEN}" | gh auth login --with-token
 
 gh config set prompt disabled
 gh release create \
```
anuraaga, line 6: Doesn't have to be this PR but I would remove this script since it's effectively one line (can and probably should use the value of the tag in the CI event rather than needing `git describe` for it)

anuraaga (approved):

PR #738, opened 2022-02-23
Change container image USER to UID

kraman:
Kubernetes requires USER to be UID in order to verify running as non root.
ref https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/security_context_others.go#L49[](https://github.com/kraman)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,6 +34,6 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -ldflags="-X $
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY --from=builder /workspace/manager .
-USER nonroot:nonroot
+USER 1002:1002
 
 ENTRYPOINT ["/manager"]


Dockerfile
```diff
@@ -34,6 +34,6 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -ldflags="-X $
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY --from=builder /workspace/manager .
-USER nonroot:nonroot
+USER 1002:1002
```
pavolloffay, line 37: Is there any reason to use `1002` if no I would prefer to use `65532`. It is default what operator-sdk uses.

* https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v3/memcached-operator/Dockerfile#L25
* https://github.com/jaegertracing/jaeger-operator/blob/main/Dockerfile#L38

pavolloffay:
@kraman please sign the CLA

Dockerfile
kraman: Im good with 65532, updating

pavolloffay (approved):

PR #746, opened 2022-02-26
add autoscale option to enable support for Horizontal Pod Autoscaling

### LINKED ISSUE #729: Collector replicas count and HorizontalPodAutoscaler
Hi, I want to use opentelemetry collector with Horizontal Pod Autoscaler, with help of OpenTelemetry Operator but OpenTelemetryCollectorSpec can have only fixed number of replicas allowed. And obviously both controllers does not work well together.

```bash
kubectl apply -f basic-otel-config.yaml
kubectl autoscale deployment simplest-collector --min=2 --max=3
```

and both opentelemetry collector and HPA controller will update the deployment:
```
simplest-collector-5db4bf447-dsjqx   0/1     Pending   0          0s
simplest-collector-5db4bf447-dsjqx   0/1     Pending   0          0s
simplest-collector-5db4bf447-dsjqx   0/1     ContainerCreating   0          0s
simplest-collector-5db4bf447-dsjqx   1/1     Running             0          1s
simplest-collector-5db4bf447-ds9mk   0/1     Pending             0          0s
simplest-collector-5db4bf447-ds9mk   0/1     Pending             0          0s
simplest-collector-5db4bf447-ds9mk   0/1     ContainerCreating   0          0s
simplest-collector-5db4bf447-ds9mk   0/1     Terminating         0          0s
simplest-collector-5db4bf447-ds9mk   1/1     Terminating         0          1s
simplest-collector-5db4bf447-ds9mk   0/1     Terminating         0          2s
simplest-collector-5db4bf447-ds9mk   0/1     Terminating         0          2s
simplest-collector-5db4bf447-ds9mk   0/1     Terminating         0          2s
simplest-collector-5db4bf447-hv9dg   0/1     Pending             0          0s
simplest-collector-5db4bf447-hv9dg   0/1     Pending             0          0s
simplest-collector-5db4bf447-hv9dg   0/1     ContainerCreating   0          0s
simplest-collector-5db4bf447-hv9dg   0/1     Terminating         0          0s
simplest-collector-5db4bf447-hv9dg   1/1     Terminating         0          1s
```

So currently my option are:
- don't use HPA at all and over-provision collectors with higher replica count.
- write my own HPA that will change replica count in OpenTelemetryCollector, which is probably to complicated.

May be im missing a better way off doing this, but right now I only see the way of achieving this by adding additional parameter (like replicasManagedByHPA or something like this) to the OpenTelemetryCollector spec, that reconcile loop will not trigger deployment update if the replicas has been changed and replicasManagedByHPA is true.

binjip978:
Resolves #729

### DIFF
--- a/apis/v1alpha1/opentelemetrycollector_types.go
+++ b/apis/v1alpha1/opentelemetrycollector_types.go
@@ -41,10 +41,6 @@ type OpenTelemetryCollectorSpec struct {
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
 
-	// MinReplicas sets a lower bound to the autoscaling feature.
-	// +optional
-	MinReplicas *int32 `json:"minReplicas,omitempty"`
-
 	// MaxReplicas sets an upper bound to the autoscaling feature. When autoscaling is enabled and no value is provided, a default value is used.
 	// +optional
 	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
--- a/apis/v1alpha1/opentelemetrycollector_types.go
+++ b/apis/v1alpha1/opentelemetrycollector_types.go
@@ -150,6 +150,7 @@ type OpenTelemetryCollectorStatus struct {
 	// Messages about actions performed by the operator on this resource.
 	// +optional
 	// +listType=atomic
+	// Deprecated: use Kubernetes events instead.
 	Messages []string `json:"messages,omitempty"`
 }
 
--- a/apis/v1alpha1/opentelemetrycollector_types.go
+++ b/apis/v1alpha1/opentelemetrycollector_types.go
@@ -33,9 +33,21 @@ type OpenTelemetryCollectorSpec struct {
 	// +optional
 	Args map[string]string `json:"args,omitempty"`
 
+	// Autoscale turns on/off the autoscale feature. By default, it's enabled if the Replicas field is not set.
+	// +optional
+	Autoscale *bool `json:"autoscale,omitempty"`
+
 	// Replicas is the number of pod instances for the underlying OpenTelemetry Collector
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
+
+	// MinReplicas sets a lower bound to the autoscaling feature.
+	// +optional
+	MinReplicas *int32 `json:"minReplicas,omitempty"`
+
+	// MaxReplicas sets an upper bound to the autoscaling feature. When autoscaling is enabled and no value is provided, a default value is used.
+	// +optional
+	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
 
 	// ImagePullPolicy indicates the pull policy to be used for retrieving the container image (Always, Never, IfNotPresent)
 	// +optional
--- a/apis/v1alpha1/opentelemetrycollector_types.go
+++ b/apis/v1alpha1/opentelemetrycollector_types.go
@@ -33,15 +33,11 @@ type OpenTelemetryCollectorSpec struct {
 	// +optional
 	Args map[string]string `json:"args,omitempty"`
 
-	// Autoscale turns on/off the autoscale feature. By default, it's enabled if the Replicas field is not set.
-	// +optional
-	Autoscale *bool `json:"autoscale,omitempty"`
-
 	// Replicas is the number of pod instances for the underlying OpenTelemetry Collector
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
 
-	// MaxReplicas sets an upper bound to the autoscaling feature. When autoscaling is enabled and no value is provided, a default value is used.
+	// MaxReplicas sets an upper bound to the autoscaling feature. If MaxReplicas is set autoscaling is enabled.
 	// +optional
 	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
 
--- a/apis/v1alpha1/opentelemetrycollector_webhook.go
+++ b/apis/v1alpha1/opentelemetrycollector_webhook.go
@@ -109,5 +109,20 @@ func (r *OpenTelemetryCollector) validateCRDSpec() error {
 		}
 	}
 
+	// validate autoscale with horizontal pod autoscaler
+	if r.Spec.Autoscale != nil && *r.Spec.Autoscale {
+		if r.Spec.Replicas != nil {
+			return fmt.Errorf("the OpenTelemetry Spec autoscale configuration is incorrect, replicas should be nil")
+		}
+
+		if r.Spec.MaxReplicas == nil || *r.Spec.MaxReplicas < int32(1) {
+			return fmt.Errorf("the OpenTelemetry Spec autoscale configuration is incorrect, maaxReplicas should be defined and more than one")
+		}
+
+		if r.Spec.MinReplicas != nil && *r.Spec.MinReplicas > *r.Spec.MaxReplicas {
+			return fmt.Errorf("the OpenTelemetry Spec autoscale configuration is incorrect, minReplicas must not be greater than maxReplicas")
+		}
+	}
+
 	return nil
 }
--- a/apis/v1alpha1/opentelemetrycollector_webhook.go
+++ b/apis/v1alpha1/opentelemetrycollector_webhook.go
@@ -111,7 +111,7 @@ func (r *OpenTelemetryCollector) validateCRDSpec() error {
 
 	// validate autoscale with horizontal pod autoscaler
 	if r.Spec.MaxReplicas != nil {
-		if r.Spec.MaxReplicas == nil || *r.Spec.MaxReplicas < int32(1) {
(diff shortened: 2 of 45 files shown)


pkg/collector/horizontalpodautoscaler.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/open-telemetry/opentelemetry-operator/internal/config"
+	"github.com/open-telemetry/opentelemetry-operator/pkg/naming"
+)
+
+func HorizontalPodAutoscaler(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelemetryCollector) autoscalingv1.HorizontalPodAutoscaler {
+	labels := Labels(otelcol)
+	labels["app.kubernetes.io/name"] = naming.Collector(otelcol)
+
+	annotations := Annotations(otelcol)
+	var cpuTarget int32 = 90
```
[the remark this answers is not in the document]
binjip978: yes, that make sense to do, but I initially tried to be consistent with jaeger operator https://github.com/jaegertracing/jaeger-operator/pull/856/files where it also hardcoded.

apis/v1alpha1/opentelemetrycollector_webhook.go
```diff
@@ -109,5 +109,20 @@ func (r *OpenTelemetryCollector) validateCRDSpec() error {
...
+		if r.Spec.Replicas != nil {
+			return fmt.Errorf("the OpenTelemetry Spec autoscale configuration is incorrect, replicas should be nil")
+		}
+
+		if r.Spec.MaxReplicas == nil || *r.Spec.MaxReplicas < int32(1) {
+			return fmt.Errorf("the OpenTelemetry Spec autoscale configuration is incorrect, maaxReplicas should be defined and more than one")
+		}
+
+		if r.Spec.MinReplicas != nil && *r.Spec.MinReplicas > *r.Spec.MaxReplicas {
+			return fmt.Errorf("the OpenTelemetry Spec autoscale configuration is incorrect, minReplicas could be more than maxReplicas")
```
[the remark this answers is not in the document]
binjip978: updated

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -33,10 +33,22 @@ type OpenTelemetryCollectorSpec struct {
 	// +optional
 	Args map[string]string `json:"args,omitempty"`
 
+	// Autoscale turns on/off the autoscale feature. By default, it's enabled if the Replicas field is not set.
+	// +optional
+	Autoscale *bool `json:"autoscale,omitempty"`
```
pavolloffay, line 38: Should this be set to true in the defaulting webhook if `replicas` is nil?

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -33,10 +33,22 @@ type OpenTelemetryCollectorSpec struct {
...
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
 
+	// MinReplicas sets a lower bound to the autoscaling feature.
+	// +optional
+	MinReplicas *int32 `json:"minReplicas,omitempty"`
+
+	// MaxReplicas sets an upper bound to the autoscaling feature. When autoscaling is enabled and no value is provided, a default value is used.
+	// +optional
+	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
```
pavolloffay, line 50: I have a proposal to make the config/CR simpler. 

Could we add only `MaxReplicas` field in this PR to support autostacling? 

If the max replicas is set the operator would create HPA. In the validating webhook it would check if maxReplicas is higher than replicas.

pkg/collector/horizontalpodautoscaler.go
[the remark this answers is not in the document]
pavolloffay: It's fine for the first version. 

I would recommend adding a comment and extracting the value to a constant.

apis/v1alpha1/opentelemetrycollector_types.go
pavolloffay: It seems that `MaxReplicas` can be nil (e.g. for infinite scaling) in this case we need `Autoscale` flag to enable the HPA.

pkg/collector/reconcile/horizontalpodautoscaler.go
```diff
@@ -0,0 +1,144 @@
...
+		existing := &autoscalingv1.HorizontalPodAutoscaler{}
+		nns := types.NamespacedName{Namespace: desired.Namespace, Name: desired.Name}
+		err := params.Client.Get(ctx, nns, existing)
+		if err != nil && k8serrors.IsNotFound(err) {
+			if err := params.Client.Create(ctx, &desired); err != nil {
+				return fmt.Errorf("failed to create: %w", err)
+			}
+			params.Log.V(2).Info("created", "hpa.name", desired.Name, "hpa.namespace", desired.Namespace)
+			continue
+		} else if err != nil {
```
pavolloffay, line 75: This construct can be made easier to read e.g. 

```golang
if err != nil {
 if isNotFound(err) {
 // create
 } else {
  return fmt.Errorf("failed to get %w", err)
 }
}
```

or the following if `isNotFound` can be called with `nil` (I think it can)

```golang
if isNotFound(err){
} else err != nil {
}
```

apis/v1alpha1/opentelemetrycollector_types.go
binjip978: Probably we can do It (remove autoscale), https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/horizontal-pod-autoscaler-v1/#HorizontalPodAutoscalerSpec
because maxReplicas is required:

maxReplicas (int32), required
upper limit for the number of pods that can be set by the autoscaler; cannot be smaller than MinReplicas.

apis/v1alpha1/opentelemetrycollector_types.go
binjip978: not sure, right now the default is one replica, if you don't specify replicas, I tried to not change default behaviour

pkg/collector/reconcile/horizontalpodautoscaler.go
binjip978: updated

apis/v1alpha1/opentelemetrycollector_webhook.go
```diff
@@ -109,5 +109,16 @@ func (r *OpenTelemetryCollector) validateCRDSpec() error {
 		}
 	}
 
+	// validate autoscale with horizontal pod autoscaler
+	if r.Spec.MaxReplicas != nil {
+		if r.Spec.MaxReplicas == nil || *r.Spec.MaxReplicas < int32(1) {
```
pavolloffay, line 114: `r.Spec.MaxReplicas == nil ` can be removed

apis/v1alpha1/opentelemetrycollector_webhook.go
```diff
@@ -109,5 +109,16 @@ func (r *OpenTelemetryCollector) validateCRDSpec() error {
...
+			return fmt.Errorf("the OpenTelemetry Spec autoscale configuration is incorrect, maxReplicas should be defined and more than one")
+		}
+
+		if r.Spec.Replicas != nil && *r.Spec.Replicas > *r.Spec.MaxReplicas {
```
pavolloffay, line 118: while we are at this, could you please set the `Replicas` in the defaulting webhook if it is null and remove other code that sets the replicas outside of the defaulting webhook?

pavolloffay (commented):
LGTM, just some minor comments

apis/v1alpha1/opentelemetrycollector_webhook.go
binjip978: not sure that I got this right, do you ask to set Spec.Replicas equal to 1 in Default() if r.Spec.Replicas == nil? Because I can see where we set Spec.Replicas outside of it, only in tests and here:
```go
	// if autoscale is enabled, use replicas from current Status
		if params.Instance.Spec.MaxReplicas != nil {
			currentReplicas := existing.Status.Replicas
			// if replicas (minReplicas from HPA perspective) is bigger than
			// current status use it.
			if *params.Instance.Spec.Replicas > currentReplicas {
				currentReplicas = *params.Instance.Spec.Replicas
			}
			updated.Spec.Replicas = &currentReplicas
		}
```

apis/v1alpha1/opentelemetrycollector_webhook.go
binjip978: updated

apis/v1alpha1/opentelemetrycollector_webhook.go
pavolloffay: The replicas should be set to 1 in deployment if it is nil. 

It's fine I will take a look at it in another PR.

pavolloffay (approved):

PR #771, opened 2022-03-14
Remove v prefix from the container image version/tag

pavolloffay:

All other OTEL images are published as `X.Y.Z` (without `v` prefix).  

The operator image is already published without prefix - https://github.com/open-telemetry/opentelemetry-operator/pkgs/container/opentelemetry-operator%2Fopentelemetry-operator 

@jpkrohling is there any reason why the operator image has `v` prefix?


jpkrohling:
No. Can we publish both versions, so that people won't get a bad surprise when they try to upgrade by just bumping the version? Adding an entry to the changelog and removing this support in a couple of versions should be fine.

jpkrohling (approved):

PR #778, opened 2022-03-16
Check exposed svc ports

### LINKED ISSUE #257: Service exposes ports for receivers which are not enabled
If I understand open-telemetry config correctly, a receiver is enabled only if its mentioned in  service section of config

```
Configuring a receiver does not enable it. Receivers are enabled via pipelines within the service section.
```
Reference - https://opentelemetry.io/docs/collector/configuration/

In our current implementation of service reconciliation, while fetching inferred Ports - I do not see this check. 

https://github.com/open-telemetry/opentelemetry-operator/blob/db40230d99040201ba7b85a64700788392e96b62/pkg/collector/adapters/config_to_ports.go#L35

https://github.com/open-telemetry/opentelemetry-operator/blob/db40230d99040201ba7b85a64700788392e96b62/pkg/collector/reconcile/service.go#L78-L82

This would imply that we would  have disabled receiver ports in service spec. Is this expected. Or am I missing something?

yuriolisa:
Signed-off-by: Yuri Sa [yurimsa@gmail.com](mailto:yurimsa@gmail.com)

[x] Created function `ConfigValidate` which checks if the services are present in the pipeline set.
[x] Created unit test to validate the new function.
[x] Included validation on `ConfigToReceiverPorts` function to open only ports for enabled receivers.
[x] Changed `config_to_ports_test.go`:
     -  Fixing indentation. 
     - Included service spec in tested config string.

Resolves #257


yuriolisa:
Resolves #257

pkg/collector/adapters/config_to_probe.go
```diff
@@ -23,18 +23,18 @@ import (
 )
 
 var (
-	errNoService    = errors.New("no service available as part of the configuration")
-	errNoExtensions = errors.New("no extensions available as part of the configuration")
+	ErrNoService    = errors.New("no service available as part of the configuration")
```
jpkrohling, line 26: Are you using this outside of the package? It shouldn't be exported without a good reason.

pkg/collector/adapters/config_to_ports.go
```diff
@@ -51,7 +52,11 @@ func ConfigToReceiverPorts(logger logr.Logger, config map[interface{}]interface{
 	if !ok {
 		return nil, ErrNoReceivers
 	}
-
+	recEnabled, err := ConfigValidate(logger, config)
+	if err != nil {
+		return nil, err
+	}
+	fmt.Printf("%v", recEnabled)
```
jpkrohling, line 59: Oops?

pkg/collector/adapters/config_validate.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+
+	"github.com/go-logr/logr"
+)
+
+var (
+	ErrNoPipeline = errors.New("no pipeline available as part of the configuration")
+)
+
+func ConfigValidate(logger logr.Logger, config map[interface{}]interface{}) (map[string]bool, error) {
```
jpkrohling, line 28: Document this function, especially what's the expected output.

jpkrohling: And contrary to what I mentioned during our conversation, it should probably be a map[interface{}] and not map[string], so that it's consistent with other usages. It would also allow you to skip the cast elsewhere in the code.

pkg/collector/adapters/config_to_probe.go
yuriolisa: Fixed them.

pkg/collector/adapters/config_to_ports.go
yuriolisa: Removed the Printf.

pkg/collector/adapters/config_validate.go
yuriolisa: I've documented the function and set the return as map[interface{}].

jpkrohling (approved):
LGTM

pkg/collector/adapters/config_validate.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	receivers, ok := cfgReceivers.(map[interface{}]interface{})
+	if !ok {
+		return nil, ErrReceiversNotAMap
+	}
+	availableReceivers := map[interface{}]bool{}
+
+	for recvID := range receivers {
+		//Getting all receivers present in the receivers section and setting them to false.
+		availableReceivers[recvID.(string)] = false
```
pavolloffay, line 43: This could probably panic if the config file does not have the correct format. The same issues is in multiple places in this PR.

yuriolisa: @pavolloffay, @jpkrohling

yuriolisa:
@pavolloffay, CI is fixed.

pkg/collector/adapters/config_to_probe.go
```diff
@@ -51,7 +51,7 @@ const (
 func ConfigToContainerProbe(config map[interface{}]interface{}) (*corev1.Probe, error) {
 	serviceProperty, ok := config["service"]
 	if !ok {
-		return nil, errNoService
+		return nil, ErrNoService
```
VineethReddy02, line 54: Why is this changed from `errNoService` to `ErrNoService`. I see that all ref are in the `adapters` package.

VineethReddy02 (changes requested):

jpkrohling:
> let collector deal with issues in the config like no service/pipelines

Correct, we should not fail or return errors when this happens, we should just ignore and continue. In this case, if there are no pipelines, there should be no ports being exposed.

pkg/collector/adapters/config_to_probe.go
yuriolisa: I'll check it.

yuriolisa:
@jpkrohling , @VineethReddy02 , @pavolloffay . Could you please review it?

pkg/collector/adapters/config_validate.go
```diff
@@ -27,7 +27,7 @@ var (
 
 //Following Otel Doc: Configuring a receiver does not enable it. The receivers are enabled via pipelines within the service section.
 //ConfigValidate returns all receivers, setting them as true for enabled and false for non-configured services in pipeline set.
-func ConfigValidate(logger logr.Logger, config map[interface{}]interface{}) (map[interface{}]bool, error) {
+func GetEnabledReceivers(logger logr.Logger, config map[interface{}]interface{}) (map[interface{}]bool, error) {
 	cfgReceivers, ok := config["receivers"]
 	if !ok {
 		return nil, ErrNoReceivers
```
jpkrohling, line 33: I like this, but it should still never return an error, ie, never fail. If the data is bad, then no receivers are enabled, so, just return nil.

yuriolisa: Thank you @jpkrohling. I've developed to return only the enabled receivers but forget to change the function signature and all the components related.

pkg/collector/adapters/config_validate.go
```diff
@@ -75,19 +68,19 @@ func GetEnabledReceivers(logger logr.Logger, config map[interface{}]interface{})
...
 				if !ok {
-					return nil, fmt.Errorf("pipeline was not properly configured")
+					return nil
 				}
 				for pipSpecID, pipSpecCfg := range pipelineDesc {
 					if pipSpecID.(string) == "receivers" {
 						receiversList, ok := pipSpecCfg.([]interface{})
 						if !ok {
-							return nil, fmt.Errorf("no receivers on pipeline configuration %q", receiversList...)
+							return nil
```
jpkrohling, line 83: Is this an error, though? Perhaps we should just continue with the next pipeline?

jpkrohling: The same comment applies to other places within this `for` loop.

pkg/collector/adapters/config_validate_test.go
```diff
@@ -104,8 +102,7 @@ service:
 	require.NotEmpty(t, config)
 
 	// test
-	check, err := GetEnabledReceivers(logger, config)
-	assert.NoError(t, err)
+	check := GetEnabledReceivers(logger, config)
 	require.Empty(t, check)
 	//require.NotEmpty(t, check)
```
jpkrohling, line 107: Could you also remove this line here?

jpkrohling (approved):
LGTM, but would be good to have a review from a fresh pair of eyes. @pavolloffay, @VineethReddy02?

pavolloffay (approved):

jpkrohling:
@VineethReddy02, given the outcome of your previous review, would you also be able to take a look at this?

VineethReddy02 (approved):

PR #786, opened 2022-03-17
Use golang 1.18

pavolloffay:

Replaces #784


.github/workflows/continuous-integration.yaml
```diff
@@ -35,6 +35,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v2
       with:
+        # TODO switch to Golang 1.18 once https://github.com/golangci/golangci-lint/issues/2649 is fixed
         go-version: 1.17
```
pavolloffay, line 39: I was able to fix the issue by using temporarily go 1.17 for lint. 

once this PR is merged I will create an issue in this repo to track this.

cmd/otel-allocator/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/otel-allocator
 
-go 1.17
+go 1.18
```
frzifus, line 3: Isn't this the minimum version that is required to compile? Iam asking [because of](https://github.com/open-telemetry/opentelemetry-operator/pull/784#issuecomment-1070528328):
> at RH there is probably no go 1.18 image for productization. We have to make sure the project can be built with go 1.17 as well

pavolloffay: I am able to build it with go 1.17 without issues (e.g. by changing go builder version in the dockerfile)

frzifus: oh ok, but it would allow people to use 1.18 features, right? Would then make sense to keep it at 1.17 until we have a build image?

frzifus (approved):

cmd/otel-allocator/go.mod
pavolloffay: That is a valid argument. If somebody will use 1.18 features we will have to fork :/

jpkrohling: 1.18 features make MUCH sense for Kubernetes Operators, like the support for generics. Much of the code can be simplified with that, I presume. But we are not in a hurry to change to it, though.

VineethReddy02 (approved):
LGTM!

pavolloffay:
PR rebased.

With the latest version of the golang-ci our lint job passes fine without any workaround.

jpkrohling (approved):

yuriolisa:
LGTM!

PR #789, opened 2022-03-18
Support nodeSelector field for non-sidecar collectors

### LINKED ISSUE #779: Collector CRD should support specifying a nodeSelector
In environments where users want certain workloads to run on specific node pools, we can taint nodes and define our pods with nodeSelectors and tolerations. Currently the Collector CRD allows us to specify tolerations for a non-sidecar Collector, but not a nodeSelector. This means that we _may_ end up running our collector on the intended node pool, but it isn't guaranteed.

Adding support for this configuration would be a nice improvement.

jutley:
This PR adds support for specifying a nodeSelector on collectors (unless they are sidecars). This allows users to schedule their collectors onto specific node pools. Fixes #779

### DIFF
--- a/apis/v1alpha1/opentelemetrycollector_types.go
+++ b/apis/v1alpha1/opentelemetrycollector_types.go
@@ -108,12 +108,12 @@ type OpenTelemetryCollectorSpec struct {
 	Resources v1.ResourceRequirements `json:"resources,omitempty"`
 
 	// Toleration to schedule OpenTelemetry Collector pods.
-	// This is only relevant to daemonsets, statefulsets and deployments
+	// This is only relevant to daemonset, statefulset, and deployment mode
 	// +optional
 	Tolerations []v1.Toleration `json:"tolerations,omitempty"`
 
 	// NodeSelector to schedule OpenTelemetry Collector pods.
-	// This is only relevant to daemonsets, statefulsets and deployments
+	// This is only relevant to daemonset, statefulset, and deployment mode
 	// +optional
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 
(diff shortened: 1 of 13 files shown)


config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
-  - manager.yaml
+- manager.yaml
```
VineethReddy02, line 2: This file shouldn't change :)

VineethReddy02 (changes requested):

config/manager/kustomization.yaml
jutley: Makes sense. I just patched it.

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -112,6 +112,11 @@ type OpenTelemetryCollectorSpec struct {
 	// +optional
 	Tolerations []v1.Toleration `json:"tolerations,omitempty"`
 
+	// NodeSelector to schedule OpenTelemetry Collector pods.
+	// This is only relevant to daemonsets, statefulsets and deployments
```
pavolloffay, line 116: Change to: 
> This is only relevant to daemonset, statefulset and deployment mode.

jutley: That phrase was copied directly from the field above, so I revised both.

pavolloffay:
@jutley the PR needs rebasing

jutley:
@pavolloffay Just rebased it.

jpkrohling:
Tests started.

pavolloffay (approved):

VineethReddy02 (approved):

PR #797, opened 2022-03-24
Including new label

### LINKED ISSUE #394: Missing labels in ConfigMap and Lease
There are two resources created by opentelemetry-operator that are missing recommended labels.
I would be satisfied with labels such as:

```
metadata
    labels
        opentelemetrycollectors.opentelemetry.io: v0.6.0-beta.0
```

name: .opentelemtry.io, kind: ConfigMap
image

name: .opentelemtry.io, kind: Lease
image

yuriolisa:
Signed-off-by: Yuri Sa <yurimsa@gmail.com>

Resolves #394


pkg/collector/labels.go
```diff
@@ -28,9 +30,11 @@ func Labels(instance v1alpha1.OpenTelemetryCollector) map[string]string {
 			base[k] = v
 		}
 	}
+	version := strings.Split(instance.Spec.Image, ":")
 
 	base["app.kubernetes.io/managed-by"] = "opentelemetry-operator"
 	base["app.kubernetes.io/instance"] = naming.Truncate("%s.%s", 63, instance.Namespace, instance.Name)
+	base["app.kubernetes.io/version"] = version[len(version)-1]
```
frzifus, line 37: Could this cause a panic if the length of version is empty? as well as in other places in the patch.

Aneurysm9: I'm not sure that `version` can be empty here, I think it would include at least an empty string, but it can certainly be length 1 which would not produce the expected result.  The length of the list should definitely be checked and `latest` used as a default if it is not 2.

yuriolisa: Thank you @Aneurysm9 and @frzifus for your consideration. I'll check it.

pkg/collector/reconcile/configmap.go
```diff
@@ -86,8 +93,10 @@ func desiredConfigMap(_ context.Context, params Params) corev1.ConfigMap {
 
 func desiredTAConfigMap(params Params) (corev1.ConfigMap, error) {
 	name := naming.TAConfigMap(params.Instance)
+	version := strings.Split(params.Instance.Spec.Image, ":")
 	labels := targetallocator.Labels(params.Instance)
 	labels["app.kubernetes.io/name"] = name
+	labels["app.kubernetes.io/version"] = version[len(version)-1]
```
frzifus, line 99: do we need to check the array length here as well? and would it make sense to maybe outsource this check to a small function?

yuriolisa: @frzifus, I've just included the missing part. As we are doing this only twice I guess It could be in that format.

jpkrohling (approved):

ringerc:
Yeah, this introduced a breaking regression per https://github.com/open-telemetry/opentelemetry-operator/issues/879 @yuriolisa  @jpkrohling

PR #811, opened 2022-04-06
Do not block pod creating on internal error in webhook

### LINKED ISSUE #802: Error of admission webhook prevents scaling of deployment despite `falurePolicy: Ignore`
We use auto-instrumentation via annotation with our pods. For that, there has to be an `Instrumentation` CRD in the same namespace as the pod. If this is not the case the following error occurs in the opentelemetry-operator:

```
msg":"failed to select an OpenTelemetry Instrumentation instance for this pod","namespace":"","name":"","error":"no OpenTelemetry Instrumentation instances available"
``` 

The error message is fine, but what happens is that the pod/deployment does not get scaled up because the addmission webhook denies the request with the event:

```
message: 'Error creating: admission webhook "mpod.kb.io" denied the request: no OpenTelemetry Instrumentation instances available'
```

because of  https://github.com/open-telemetry/opentelemetry-operator/blob/ecb139c9dc453961f695a186d665666cdd1dfd3c/internal/webhookhandler/webhookhandler.go#L85-L88

From my understanding, this error should be ignored and the respective deployment/pod (which has the auto-instumentation annotation) should be scaled up nevertheless, albeit without the mutation, because the admission webhook has `failurePolicy: Ignore`. Am I missing something?

The instrumentation feature is not critical for the deployment (at least in my context), so I would rather have a running pod without instrumentation than no pod at all.

pavolloffay:

Resolves #802

### DIFF
--- a/internal/webhookhandler/webhookhandler.go
+++ b/internal/webhookhandler/webhookhandler.go
@@ -29,7 +29,7 @@ import (
 	"github.com/open-telemetry/opentelemetry-operator/internal/config"
 )
 
-// +kubebuilder:webhook:path=/mutate-v1-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=mpod.kb.io,sideEffects=none,admissionReviewVersions=v1
+// +kubebuilder:webhook:path=/mutate-v1-pod,mutating=true,failurePolicy=ignore,groups="",resources=pods,verbs=create;update,versions=v1,name=mpod.kb.io,sideEffects=none,admissionReviewVersions=v1
 // +kubebuilder:rbac:groups="",resources=namespaces,verbs=list;watch
 // +kubebuilder:rbac:groups=opentelemetry.io,resources=opentelemetrycollectors,verbs=get;list;watch
 // +kubebuilder:rbac:groups=opentelemetry.io,resources=instrumentations,verbs=get;list;watch
@@ -79,7 +79,10 @@ func (p *podSidecarInjector) Handle(ctx context.Context, req admission.Request) 
 	err = p.client.Get(ctx, types.NamespacedName{Name: req.Namespace, Namespace: ""}, &ns)
 	if err != nil {
 		res := admission.Errored(http.StatusInternalServerError, err)
+		// By default, admission.Errored sets Allowed to false which blocks pod creation even though the failurePolicy=ignore.
+		// Allowed set to true makes sure failure does not block pod creation in case of an error.
 		res.Allowed = true
+		return res
 	}
 
 	for _, m := range p.podMutators {
@@ -87,6 +90,7 @@ func (p *podSidecarInjector) Handle(ctx context.Context, req admission.Request) 
 		if err != nil {
 			res := admission.Errored(http.StatusInternalServerError, err)
 			res.Allowed = true
+			return res
 		}
 	}
 
@@ -94,6 +98,7 @@ func (p *podSidecarInjector) Handle(ctx context.Context, req admission.Request) 
 	if err != nil {
 		res := admission.Errored(http.StatusInternalServerError, err)
 		res.Allowed = true
+		return res
 	}
 	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
 }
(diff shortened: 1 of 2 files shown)


pavolloffay:
@trutty would you like to review?

internal/webhookhandler/webhookhandler.go
```diff
@@ -78,19 +78,22 @@ func (p *podSidecarInjector) Handle(ctx context.Context, req admission.Request)
 	ns := corev1.Namespace{}
 	err = p.client.Get(ctx, types.NamespacedName{Name: req.Namespace, Namespace: ""}, &ns)
 	if err != nil {
-		return admission.Errored(http.StatusInternalServerError, err)
+		res := admission.Errored(http.StatusInternalServerError, err)
+		res.Allowed = true
```
frzifus, line 82: do we not miss a return here? (and on the other places)

frzifus (commented):
Could it lead to some other issues, if we return code 500 and an error message in the status result?

pavolloffay:
> Could it lead to some other issues, if we return code 500 and an error message in the status result?

I cannot think of any. Do you have anything in mind? 

The error still gets propagated to the events but it does not block pod creation.

frzifus:
> I cannot think of any. Do you have anything in mind?

No, i was just wondering. :+1:

internal/webhookhandler/webhookhandler_test.go
```diff
@@ -403,15 +403,17 @@ func TestFailOnInvalidRequest(t *testing.T) {
...
+		allowed  bool
 	}{
 		{
-			"empty payload",
-			admission.Request{},
-			http.StatusBadRequest,
+			name:     "empty payload",
+			req:      admission.Request{},
+			expected: http.StatusBadRequest,
+			allowed:  false,
```
pavolloffay, line 412: Now the webhook fails the pod creation only if it cannot decode the request, which should not happen in practice.

internal/webhookhandler/webhookhandler.go
```diff
@@ -78,19 +78,27 @@ func (p *podSidecarInjector) Handle(ctx context.Context, req admission.Request)
...
+		return res
 	}
 
 	for _, m := range p.podMutators {
 		pod, err = m.Mutate(ctx, ns, pod)
 		if err != nil {
-			return admission.Errored(http.StatusInternalServerError, err)
+			res := admission.Errored(http.StatusInternalServerError, err)
+			res.Allowed = true
+			return res
```
trutty, line 93: I am not sure whether creating an error and later overriding `allowed` is the best option - or what happens in the admission controller when `allowed = true` but `result.code != 200`.

I would have suggested to create an allowed response but add warnings to it like so:

```suggestion
			return admission.Allowed("").WithWarnings(err.Error())
```

Having said that, I am fairly new to the workings of an admission controller and admission webhooks, so maybe your solution is already best practice.

trutty (commented):
> @trutty would you like to review?

Thanks @pavolloffay for the PR.
As stated, I am new to the inner workings of admission controller/webooks so my suggestion below could have faulty reasoning. Just wondering whether it is the best practice implementation 😉

internal/webhookhandler/webhookhandler.go
pavolloffay: I will play around and try to add events explicitly to the pod.

pavolloffay: Something like this

```
p.recorder.Event(&pod, "Normal", "ConfigUpdate", fmt.Sprintf("Testing events"))
```

does not work in the webhook because the pod name is not known. It creates following event:

```
5s          Normal    ConfigUpdate        pod                                                Testing events
```

pavolloffay: Based on ^^ I am keeping my original version of this PR and I have added more comments with findings to the code.

pavolloffay:
@open-telemetry/operator-maintainers could you please review?

pavolloffay:
@jpkrohling  @VineethReddy02 could you please review?

jpkrohling (commented):
The original report is about auto-instrumentation webhooks, while this PR seems to fix this behavior for the sidecar injection. Are they both sharing the same function?

pavolloffay:
>  The original report is about auto-instrumentation webhooks, while this PR seems to fix this behavior for the sidecar injection. Are they both sharing the same function?

yes, both are sharing the same pod webhook skeleton.

jpkrohling (approved):

PR #798, opened 2022-03-25
fix panic when spec.replicas is nil

wei840222:
Resolves #803 

Since the spec.replicas in [docs](https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/api.md#opentelemetrycollectorspec) is not required, so it may be nil.
Need check whether replicas is nil before get value of it.
image

PR #814, opened 2022-04-07
Set replicas to default value

pavolloffay:

Follow up for #798

### DIFF
--- a/apis/v1alpha1/opentelemetrycollector_webhook.go
+++ b/apis/v1alpha1/opentelemetrycollector_webhook.go
@@ -93,11 +93,6 @@ func (r *OpenTelemetryCollector) validateCRDSpec() error {
 		return fmt.Errorf("the OpenTelemetry Collector mode is set to %s, which does not support the attribute 'volumeClaimTemplates'", r.Spec.Mode)
 	}
 
-	// validate replicas
-	if (r.Spec.Mode == ModeSidecar || r.Spec.Mode == ModeDaemonSet) && r.Spec.Replicas != nil {
-		return fmt.Errorf("the OpenTelemetry Collector mode is set to %s, which does not support the attribute 'replicas'", r.Spec.Mode)
-	}
-
 	// validate tolerations
 	if r.Spec.Mode == ModeSidecar && len(r.Spec.Tolerations) > 0 {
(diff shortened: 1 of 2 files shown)


apis/v1alpha1/opentelemetrycollector_webhook.go
```diff
@@ -40,6 +40,8 @@ var _ webhook.Defaulter = &OpenTelemetryCollector{}
 
 // Default implements webhook.Defaulter so a webhook will be registered for the type.
 func (r *OpenTelemetryCollector) Default() {
+	opentelemetrycollectorlog.Info("default", "name", r.Name)
```
jpkrohling, line 43: Is this a leftover from a local test?

apis/v1alpha1/opentelemetrycollector_webhook_test.go
```diff
@@ -0,0 +1,64 @@
+package v1alpha1
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestOELColDefaultingWebhook(t *testing.T) {
```
jpkrohling, line 10:
```suggestion
func TestOTELColDefaultingWebhook(t *testing.T) {
```

jpkrohling (approved):
LGTM, but I wonder if this is a new behavior brought in by one of our dependencies, or if we always had this bug and nobody ever complained, which seems unlikely, as I doubt all our users are setting the number of replicas in their CRs.

apis/v1alpha1/opentelemetrycollector_webhook.go
pavolloffay: nope just moving the logging from the end of the function

jpkrohling: Would it make sense to change it somehow? Perhaps even remove it?

frzifus (approved):

apis/v1alpha1/opentelemetrycollector_webhook.go
jpkrohling: Alright -- the level is set to "info", but it's not providing any real info. It sounds more like a debug from your description. I'll let it at your discretion though.

PR #824, opened 2022-04-18
Create serving certs for headless services on OpenShift (#818)

### LINKED ISSUE #818: Create serving certs for headless services on OpenShift
When running on OpenShift, the opentelemetry-operator should include a `service.beta.openshift.io/serving-cert-secret-name` annotation in the headless services it creates for the collector. This will cause a certificate to be generated for the service and stored in a secret that can be mounted and used in the associated receiver TLS configuration. 

Without this feature in the operator, it is necessary to either manage serving certificates manually, or add the annotation manually after deploying the collector. Until the annotation is added, the collector cannot start successfully because the secret being mounted as a volume does not yet exist.

rkukura:
resolves #818 

Signed-off-by: Robert Kukura <rkukura@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,8 @@ The `config` node holds the `YAML` that should be passed down as-is to the under
 
 At this point, the Operator does *not* validate the contents of the configuration file: if the configuration is invalid, the instance will still be created but the underlying OpenTelemetry Collector might crash.
 
+The Operator does examine the configuration file to discover configured receivers and their ports. If it finds receivers with ports, it creates a pair of kubernetes services, one headless, exposing those ports within the cluster. The headless service contains a `service.beta.openshift.io/serving-cert-secret-name` annotation that will cause OpenShift to create a secret containing a certificate and key. This secret can be mounted as a volume and the certificate and key used in those receivers' TLS configurations.
+
 
 ### Upgrades
 
--- a/pkg/collector/reconcile/service.go
+++ b/pkg/collector/reconcile/service.go
@@ -158,6 +158,16 @@ func headless(ctx context.Context, params Params) *corev1.Service {
 	}
 
 	h.Name = naming.HeadlessService(params.Instance)
+
+	// copy to avoid modifying params.Instance.Annotations
+	annotations := map[string]string{
+		"service.beta.openshift.io/serving-cert-secret-name": fmt.Sprintf("%s-tls", h.Name),
+	}
+	for k, v := range h.Annotations {
+		annotations[k] = v
+	}
+	h.Annotations = annotations
+
 	h.Spec.ClusterIP = "None"
 	return h
 }
--- a/pkg/collector/reconcile/service_test.go
+++ b/pkg/collector/reconcile/service_test.go
@@ -196,6 +196,7 @@ func TestDeleteServices(t *testing.T) {
 func TestHeadlessService(t *testing.T) {
 	t.Run("should return headless service", func(t *testing.T) {
 		actual := headless(context.Background(), params())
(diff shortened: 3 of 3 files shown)


pkg/collector/reconcile/service.go
```diff
@@ -158,6 +158,16 @@ func headless(ctx context.Context, params Params) *corev1.Service {
 	}
 
 	h.Name = naming.HeadlessService(params.Instance)
+
+	// copy to avoid modifying params.Instance.Annotations
+	annotations := map[string]string{
+		"service.beta.openshift.io/serving-cert-secret-name": fmt.Sprintf("%s-tls", h.Name),
```
pavolloffay, line 164: This annotation should be present only if the collector is deployed on the OpenShift. 

IIRC there is already some machinery in the project that recognizes the platform.

rkukura: I had considered this, but noticed in https://github.com/jaegertracing/jaeger-operator/blob/main/pkg/service/collector.go#L28 that jaeger-operator adds the annotation regardless of the platform. My understanding is that unrecognized annotations are silently ignored, so I don't think there is any harm. Also, making this conditional would require addition unit test code. But I can make it conditional if required.

pavolloffay:
@rkukura is there any documentation needed for this? Not familiar users might ask how to use those certs? Why does only the headless service have them?

rkukura:
> @rkukura is there any documentation needed for this? Not familiar users might ask how to use those certs? Why does only the headless service have them?

@pavolloffay I added some text to README.md. The headless service is what is typically used within the cluster, such as for connections from agents. The generated certificate is only trusted within the cluster.

README.md
```diff
@@ -57,6 +57,8 @@ The `config` node holds the `YAML` that should be passed down as-is to the under
 
 At this point, the Operator does *not* validate the contents of the configuration file: if the configuration is invalid, the instance will still be created but the underlying OpenTelemetry Collector might crash.
 
+The Operator does examine the configuration file to discover configured receivers and their ports. If it finds receivers with ports, it creates a pair of kubernetes services, one headless, exposing those ports within the cluster. The headless service contains a `service.beta.openshift.io/serving-cert-secret-name` annotation that will cause OpenShift to create a secret containing a certificate and key. This secret can be mounted as a volume and the certificate and key used in those receivers' TLS configurations.
```
pavolloffay, line 60: The text looks good/ it explains what the annotation does. I would it makes sense here also document why someone should be doing something like this/waht is the common use case? Users are often searching for solutions for use-cases.

rkukura: More general discussion of configuring and securing the collector is not specific to the operator, so probably doesn't belong in this repo. We could add some mention of this operator functionality in https://opentelemetry.io/docs/collector/configuration/#setting-up-certificates after this PR is merged. Adding new documentation in this repo that covers common operator use cases (including but not limited to securing OTLP between agent and gateway collecotrs) seems beyond the scope of this PR.

pavolloffay: > More general discussion of configuring and securing the collector is not specific to the operator, so probably doesn't belong in this repo. 

Our users should understand why there is a headless service, when to use it and what is the use case for using those certificates.

> We could add some mention of this operator functionality in https://opentelemetry.io/docs/collector/configuration/#setting-up-certificates

I don't think that operator documentation belongs there. You could try to work with the docs/website sig to add an operator page there. So far there are no operator docs in opentelemetry.io.

pavolloffay: Maybe I was not clear, I wanted a better docs for the certificates - what is the use-case for them etc.

pavolloffay: You can create an additional PR to add them

pavolloffay (approved):

PR #833, opened 2022-04-22
Set replicas to MaxReplicas if HPA is enabled

binjip978:
If number of replicas in current deployment status is bigger than
MaxReplicas and HPA is enabled set deployment replicas to MaxReplicas.

### DIFF
--- a/pkg/collector/reconcile/deployment.go
+++ b/pkg/collector/reconcile/deployment.go
@@ -24,6 +24,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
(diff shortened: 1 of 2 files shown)


pavolloffay:
@binjip978 was the current behavior causing any issues?

binjip978:
> @binjip978 was the current behavior causing any issues?

Yeah, during initial deployment or upgrade I saw that replicas count in status was much higher than it was specified in deployment or HPA configuration, the reason is not clear to me (maybe some issue with metrics server in our cluster)
But if current replica count in status will be bigger than MaxReplicas it will became part of deployment spec. And then two controllers will fight each other, most of time in our case it will converge to HPA values, but it takes time.

This problem is not always reproducible in our environment, but it may happen, most of time it works as expected.

pavolloffay (approved):

pavolloffay:
@jpkrohling could you review it as well? You did a bunch of work around HPA in Jaeger operator.

VineethReddy02 (approved):

jpkrohling (approved):

PR #837, opened 2022-04-25
Update sidecar example in README

erichsueh3:
This PR updates the sidecar example in the README, as it does not work. It removes the unnecessary  `otlpreceiver` and changes the `jaegerreceiver` protocol to `thrift_compact` instead of `grpc`. This is also reflected in a PR on the `opentelemetry-helm-charts` repository, [183](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/183).

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,11 +89,7 @@ spec:
     receivers:
       jaeger:
         protocols:
-          grpc:
-      otlp:
-        protocols:
-          grpc:
-          http:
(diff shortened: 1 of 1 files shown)


pavolloffay (approved):

pavolloffay:
I would rather have an example using the OTEL instrumentation and not Jaeger, but it's not in the scope of this PR.

Issue #871, opened 2022-05-13
e2e tests are failing


pavolloffay:
https://github.com/open-telemetry/opentelemetry-operator/pull/689 broke e2e tests. 

cc) @fscellos

fscellos:
Arg ! Why ? I fix it next monday.

pavolloffay:
Done in #875

PR #875, opened 2022-05-16
Fix e2e tests

fscellos:
(simple order change on merge on main between injected attributes)  #871


fscellos:
@pavolloffay : this fix for e2e tests. Not a really correction, just some modification that occurs between main and merged branch on order of injection of attributes for env variables of container

pavolloffay (approved):

PR #882, opened 2022-05-19
Fix docker multiarch build for operator

### LINKED ISSUE #874: Multiarch support seems to be incomplete
Hi,

I tried to install the opentelemetry-operator via helm chart on a microk8s running on Apple M1:
```
helm install \
  opentelemetry-operator open-telemetry/opentelemetry-operator -n  opentelemetry-operator-system \
  --set manager.image.tag=v0.50.0
```

According to github the image used should be multiarch:
https://github.com/open-telemetry/opentelemetry-operator/pkgs/container/opentelemetry-operator%2Fopentelemetry-operator/21667125?tag=v0.50.0

Immediately after starting the pod crashes with `standard_init_linux.go:228: exec user process caused: exec format error`.

Do you plan to support the M1 architecture?

Thanks for your help!
Cheers

pavolloffay:

Resolves #874

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,6 @@ COPY internal/ internal/
 COPY pkg/ pkg/
 COPY versions.txt versions.txt
 
-ARG TARGETARCH=amd64
 ARG VERSION_PKG
 ARG VERSION
 ARG VERSION_DATE
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,7 @@ ARG AUTO_INSTRUMENTATION_NODEJS_VERSION
 ARG AUTO_INSTRUMENTATION_PYTHON_VERSION
 
 # Build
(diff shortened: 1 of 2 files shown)


devops-42 (approved):

Dockerfile
```diff
@@ -17,7 +17,6 @@ COPY internal/ internal/
 COPY pkg/ pkg/
 COPY versions.txt versions.txt
 
-ARG TARGETARCH=amd64
```
frzifus, line 20:
```suggestion
ARG TARGETARCH
```
would it make sense to remove `=amd64`?

pavolloffay: Then we would have to pass the arch as argument which should not be necessary

VineethReddy02 (approved):

Dockerfile
frzifus: But without we are not able to build it for a different target architecture, right?

pavolloffay: we are via docker buildx

PR #887, opened 2022-05-23
chore(e2e): use public target-allocator img

mat-rumian:
It fixes failing tests in the `opentelemetry-helm-charts` repository [gh-action](https://github.com/open-telemetry/opentelemetry-helm-charts/runs/6555387039?check_suite_focus=true#step:6:995)

### DIFF
--- a/tests/e2e/smoke-targetallocator/02-install.yaml
+++ b/tests/e2e/smoke-targetallocator/02-install.yaml
@@ -6,7 +6,7 @@ spec:
   mode: statefulset
   targetAllocator:
     enabled: true
-    image: "local/opentelemetry-operator-targetallocator:e2e"
+    image: "ghcr.io/open-telemetry/opentelemetry-operator/target-allocator:0.1.0"
   config: |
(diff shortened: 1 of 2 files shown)


VineethReddy02 (approved):
LGTM!

VineethReddy02:
@mat-rumian But it looks like the local img is used to test the `target-allocator` code path in the operator repo. Changing this to the public image will not give visibility over the master `target-allocator`.

pavolloffay:
We have talked privately with @mat-rumian that it would be better to do not merge this change and instead change the helm chart repo to change the image there (or build it).

pavolloffay (changes requested):
We should keep using the locally built image that includes all changes. 

The issue is in the helm char repository and we should fix it there by either building the TA image or override the image there with a pinned version. The helm chart depends on a specific operator version so pinning the image is fine.

mat-rumian:
Going to close this PR and prepare solution in helm-chart repo :)

PR #905, opened 2022-06-01
Add cert manager 1.8.0

yuriolisa:
Signed-off-by: Yuri Sa [yurimsa@gmail.com]

- Modifying makefile to install cert-manager 1.8.0.
- Adding on Compatibility Matrix this cert-manager version for the last release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,27 +297,27 @@ The OpenTelemetry Operator *might* work on versions outside of the given range, 
 
 | OpenTelemetry Operator | Kubernetes           | Cert-Manager         |
 |------------------------|----------------------|----------------------|
-| v0.51.0                | v1.19 to v1.23       | 1.8.0                |
-| v0.50.0                | v1.19 to v1.23       | 1.6.1                |
-| v0.49.0                | v1.19 to v1.23       | 1.6.1                |
-| v0.48.0                | v1.19 to v1.23       | 1.6.1                |
-| v0.47.0                | v1.19 to v1.23       | 1.6.1                |
-| v0.46.0                | v1.19 to v1.23       | 1.6.1                |
-| v0.45.0                | v1.21 to v1.23       | 1.6.1                |
-| v0.44.0                | v1.21 to v1.23       | 1.6.1                |
-| v0.43.0                | v1.21 to v1.23       | 1.6.1                |
-| v0.42.0                | v1.21 to v1.23       | 1.6.1                |
-| v0.41.1                | v1.21 to v1.23       | 1.6.1                |
-| v0.41.0                | v1.20 to v1.22       | 1.6.1                |
-| v0.40.0                | v1.20 to v1.22       | 1.6.1                |
-| v0.39.0                | v1.20 to v1.22       | 1.6.1                |
-| v0.38.0                | v1.20 to v1.22       | 1.6.1                |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -325,6 +325,7 @@ Please see [CONTRIBUTING.md](CONTRIBUTING.md).
 Approvers ([@open-telemetry/operator-approvers](https://github.com/orgs/open-telemetry/teams/operator-approvers)):
 
 - [Dmitrii Anoshin](https://github.com/dmitryax), Splunk
+- [Yuri Oliveira Sa](https://github.com/yuriolisa), Red Hat
```
pavolloffay, line 328: This should not be here :)

README.md
```diff
@@ -296,7 +296,7 @@ The OpenTelemetry Operator *might* work on versions outside of the given range,
 
 | OpenTelemetry Operator | Kubernetes           | Cert-Manager         |
 |------------------------|----------------------|----------------------|
-| v0.51.0                | v1.19 to v1.23       | 1.6.1                |
+| v0.51.0                | v1.19 to v1.23       | 1.8.0                |
```
pavolloffay, line 299: This should not be changed either. The new release will start using the 1.8.0.

I want to raise a slightly different message here. We should probably put here the kind version instead of the cert-manager version because we should be compatible with the cert manger kinds. I would love to see if the oldest version of the cert-manger works with OTEL operator.

README.md
yuriolisa: Sure, Removing it.

README.md
```diff
@@ -297,27 +297,27 @@ The OpenTelemetry Operator *might* work on versions outside of the given range,
...
+| v0.41.0                | v1.20 to v1.22       | v1alpha2             |
+| v0.40.0                | v1.20 to v1.22       | v1alpha2             |
+| v0.39.0                | v1.20 to v1.22       | v1alpha2             |
+| v0.38.0                | v1.20 to v1.22       | v1alpha2             |
+| v0.37.1                | v1.20 to v1.22       | v1alpha2, v1alpha3   |
+| v0.37.0                | v1.20 to v1.22       | v1alpha2, v1alpha3   |
+| v0.36.0                | v1.20 to v1.22       | v1alpha2, v1alpha3   |
+| v0.35.0                | v1.20 to v1.22       | v1alpha2, v1alpha3   |
+| v0.34.0                | v1.20 to v1.22       | v1alpha2, v1alpha3   |
+| v0.33.0                | v1.20 to v1.22       | v1alpha2, v1alpha3   |
```
pavolloffay, line 320: did we actually tested 0.33.0 against e.g. the oldest version of cert-manager v1alpha3?

pavolloffay:
To move this forward we have to be confident that e.g. OTEL operator v0.33.0 works well with all cert-manager versions that have v1alpha2 kind (e.g. the oldest and newest). Are we confident? If no, then we should be rather explicit and keep the established approach.

yuriolisa:
@pavolloffay, yes, we are confident that v1alpha2 is compatible with v0.33.0, since tested on that release the cert-manager v1.4.3 which deploys the v1alpha2 API version. As you could see in the following link:

https://github.com/open-telemetry/opentelemetry-operator/blob/402be272e17c43f7139de14f6b19dabd7cb5c96d/Makefile#L137

pavolloffay (approved):

pavolloffay:
Let's give it a try then :)

PR #941, opened 2022-06-23
Change horizontal pod autoscaler to use otelcol scale subresource

### LINKED ISSUE #801: Change our HorizontalPodAutoScaler (HPA) integration to use otelcol /scale subresource
The https://github.com/open-telemetry/opentelemetry-operator/pull/746 introduced integration with HPA for collector deployment/~statefulset/daemonset~ objects. 

https://github.com/open-telemetry/opentelemetry-operator/pull/785 / #775 added `/scale` subresource to the collector CRD, which allows configuring HPA on the otelcol CR directly. It seems like a better/cleaner approach and we should consider switching it.

cc) @secat @binjip978

### LINKED ISSUE #906: HorizontalPodAutoscaler continuously creates and terminates pods
I have been trying to work on #801 and in simply trying to get something to scale I saw some odd behavior.

The collector does seem to scale correctly (at least eventually) but collector pods are repeatedly being created and then immediately terminated.  To reproduce this, do the following, which deploys a simple collector instance and uses *tracegen* to create traffic:

- Create/Login to an OpenShift instance.  (NOTE: so far I have not been able to get this to scale at all with minikube despite following the [official instructions](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/) and several blog posts based on those.)
- Install the otel-operator
- Download the attached files *otel-tracegen.txt* and *otel-collector-simplest.txt* and rename them to *otel-tracegen.yaml* and *otel-collector-simplest.yaml*
- Execute the following commands:
```
kubectl create namespace simple
kubectl apply --namespace simple -f ./otel-collector-simplest.yaml
kubectl apply --namespace simple -f ./otel-tracegen.yaml
```
- In separate terminal windows, execute the following commands:
```
kubectl get hpa --namespace simple --watch
kubectl get pods --namespace simple --watch
```
After 4 or 5 minutes the window with the *kubectl get hpa...* command should finally return an entry where there is a real value under **TARGET** rather than unknown.  At that point the window with the *kubectl get pods...* command will start constantly streaming *Pending/ContainerCreating/Terminating* events.

Eventually the deployment will scale, but pod *Pending/ContainerCreating/Terminating* will continue

[otel-collector-simplest.txt](https://github.com/open-telemetry/opentelemetry-operator/files/8816056/otel-collector-simplest.txt)
[otel-tracegen.txt](https://github.com/open-telemetry/opentelemetry-operator/files/8816057/otel-tracegen.txt)

### LINKED ISSUE #922: Setting maxReplicas (for hpa) results in wrong scaleTargetRef.kind when mode is statefulset
I have a config like this

```yaml
kind: OpenTelemetryCollector
metadata:
  name: gcp-gateway
  namespace: default
  labels:
    app: gcp-gateway
spec:
  mode: statefulset
...
  replicas: 2
  maxReplicas: 5
```

The HPA looks like this (points to a deployment):
```
observiq-gcp-gateway-collector   Deployment/observiq-gcp-gateway-collector   <unknown>/90%   2         5         0          35s
```

The HPA is correct except for its target. 

After re-thinking my approach, I can use a Deployment for my collector, but I think this is still a bug. The operator should create the correct HPA.

kevinearls:

This fixes #801 so that the HPA is created on an OpenTelemetryCollector rather than a Deployment.

I believe it also fixes #922 
I believe it also fixes #906 but will confirm


apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -37,6 +37,10 @@ type OpenTelemetryCollectorSpec struct {
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
 
+	// MinReplicas sets a lower bound to the autoscaling feature
+	// +optional
+	MinReplicas int32 `json:"minReplicas,omitempty"`
```
pavolloffay, line 42: What is the relationship between `MinReplicas` and `Replicas`? Which one should be used when? This is something that should be documented.

pkg/collector/horizontalpodautoscaler.go
```diff
@@ -35,18 +35,18 @@ func HorizontalPodAutoscaler(cfg config.Config, logger logr.Logger, otelcol v1al
 
 	return autoscalingv1.HorizontalPodAutoscaler{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        naming.Collector(otelcol),
+			Name:        otelcol.Name,
```
pavolloffay, line 38: What is the rationale for the name change? Could this produce an invalid name?

apis/v1alpha1/opentelemetrycollector_types.go
pavolloffay: With that we probably need to add some checks to the webhook 

https://github.com/open-telemetry/opentelemetry-operator/blob/b77c89d6ea846ba5d184a4ab8716744e190ef677/apis/v1alpha1/opentelemetrycollector_webhook.go#L120-L122

to test the minreplicas or any relationship between replicas, minreplias and maxreplicas.

pavolloffay:
I would like to see e2e test for this functionality. It can be actually two tests -  scaling up and down with the HPA.

pkg/collector/horizontalpodautoscaler.go
kevinearls: We are now scaling on the OpenTelemetryCollector rather than the deployment.  If the otelcol.Name is "simple" we append "-collector" to the deployment name.  

If you think it's a good idea I could add a function to naming which just does

```
	return DNSName(Truncate("%s", 63, otelcol.Name))
```

yuriolisa (changes requested):
Please fix the lint.

kevinearls:
@pavolloffay Please review.  I think I've covered all of the things we discussed.

@yuriolisa lint now passes

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -37,9 +48,8 @@ type OpenTelemetryCollectorSpec struct {
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
 
-	// MaxReplicas sets an upper bound to the autoscaling feature. If MaxReplicas is set autoscaling is enabled.
 	// +optional
-	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
```
pavolloffay, line 42: I am not if we can change the CRD. In this case a field is removed.

controllers/opentelemetrycollector_controller.go
```diff
@@ -21,7 +21,8 @@ import (
 
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
-	autoscalingv1 "k8s.io/api/autoscaling/v1"
+	autoscalingv2beta2 "k8s.io/api/autoscaling/v2beta2"
```
pavolloffay, line 24: I would prefer to keep using v1 if the v2beta2 is not needed.

kevinearls:
@pavolloffay Can you please review this, re-running the checks first if needed?  (The 1.24 e2e tests are failing in a place I didn't touch.)

I have reverted to using autoscaling v1 and removed the AutoScaleSpec changes (I can do both later) so I think that covers everything we've discussed.

hack/install-metrics-server.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# Install metrics-server for autoscale tests
```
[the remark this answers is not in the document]
kevinearls: No, with minikube you can just add --addons "metrics-server" to the start command.

Where should I document this?  In this script?  In the Makefile?

yuriolisa (approved):
@kevinearls, please ensure that change will not disrupt the existing instances like @pavolloffay mentioned above. LGTM.

kevinearls:
@yuriolisa @pavolloffay I have added code that takes care of the upgrade.  Please review and merge if it looks ok.

apis/v1alpha1/opentelemetrycollector_webhook.go
```diff
@@ -120,6 +120,15 @@ func (r *OpenTelemetryCollector) validateCRDSpec() error {
 		if r.Spec.Replicas != nil && *r.Spec.Replicas > *r.Spec.MaxReplicas {
 			return fmt.Errorf("the OpenTelemetry Spec autoscale configuration is incorrect, replicas must not be greater than maxReplicas")
 		}
+
+		if r.Spec.MinReplicas != nil && *r.Spec.MinReplicas > *r.Spec.MaxReplicas {
+			return fmt.Errorf("the OpenTelemetry Spec autoscale configuration is incorrect, minReplicas must not be greater than maxReplicas")
+		}
+
+		if r.Spec.MinReplicas != nil && *r.Spec.MinReplicas < int32(1) {
+			return fmt.Errorf("the OpenTelemetry Spec autoscale configuration is incorrect, minReplicas should be more than one")
```
yuriolisa, line 129: I consider MinReplicas should be at least one, correct?

bundle/manifests/opentelemetry.io_opentelemetrycollectors.yaml
```diff
@@ -223,6 +223,10 @@ spec:
                   If MaxReplicas is set autoscaling is enabled.
                 format: int32
                 type: integer
+              minReplicas:
+                description: MinReplicas sets a lower bound to the autoscaling feature.
```
yuriolisa, line 227: Could you please add here the lowest value?

docs/api.md
```diff
@@ -1455,6 +1455,15 @@ OpenTelemetryCollectorSpec defines the desired state of OpenTelemetryCollector.
             <i>Format</i>: int32<br/>
         </td>
         <td>false</td>
+      </tr><tr>
+        <td><b>minReplicas</b></td>
+        <td>integer</td>
+        <td>
+          MinReplicas sets a lower bound to the autoscaling feature.<br/>
```
yuriolisa, line 1462: Also here.

yuriolisa (changes requested):

yuriolisa (approved):
LGTM.

kevinearls:
Closing this in favor of #983 as I managed to get myself into rebase hell.

PR #951, opened 2022-06-26
Fix the issue that the number of target-allocator replicas 

CoderPoet:
Fix the issue that the number of target-allocator replicas 

- [x] Add a judgment that there is no need to modify replicas when dealing with target-allocators

### DIFF
--- a/pkg/collector/reconcile/deployment.go
+++ b/pkg/collector/reconcile/deployment.go
@@ -99,7 +99,7 @@ func expectedDeployments(ctx context.Context, params Params, expected []appsv1.D
 			updated.ObjectMeta.Labels[k] = v
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/collector/reconcile/deployment.go
```diff
@@ -99,7 +99,7 @@ func expectedDeployments(ctx context.Context, params Params, expected []appsv1.D
 			updated.ObjectMeta.Labels[k] = v
 		}
 
-		if params.Instance.Spec.MaxReplicas != nil {
+		if params.Instance.Spec.MaxReplicas != nil && desired.Labels["app.kubernetes.io/component"] != "opentelemetry-targetallocator" {
```
pavolloffay, line 102: Should we rather check if the component label equals to the collector?

Issue #940, opened 2022-06-22
Auto-instrumentation - never ending pod mutation


mat-rumian:
Environment:
* OpenTelemetry-Operator deployed 0.52.0
* some other operator which is modifying environment variables of the pod container

Error Flow:
1. OT Operator adds `environment variables`, `volume` and `init container` with auto-instrumentation
2. other operator takes the same pod and adds some other environment variables
3. OT Operator detects changes in the pod and try to update it (env vars)
4. other operator again detect changes and change env vars
5. and so on...

Final error log from k8s:
```
Failed to annotate pod with branch ENI details: Pod "test-pod-64cf7b8d5b-j8zbv" is invalid: spec: Forbidden: pod updates may not change fields other than `spec.containers[*].image`, `spec.initContainers[*].image`, `spec.activeDeadlineSeconds` or `spec.tolerations` (only additions to existing tolerations) 
```

Expected flow for OT Operator:
1. OT Operator checks (before pod mutation) if instrumentation is configured for pod
2. OT Operator adds `environment variables`, `volume` and `init container` with auto-instrumentation
3. other operator add applies his stuff
4. OT Operator doesn't make any other changes as pod is already instrumented

mat-rumian:
@pavolloffay I will be happy to help with this as already investigated the issue. I'm not sure about solution, how should we check if pod is already mutated by OT Operator:
1. check if env vars are already configured?
2. check if init container is already configured?
3. introduce additional annotation to the pod and check if exists (add annotation to the pod after instrumentation configuration part)
4. something else?

PR #953, opened 2022-06-27
Fix Pod Mutation loop

mat-rumian:
Functionality changes:
* added check if Pod is already instrumented (check if auto-instrumentation `initContainer` exists)
* added check for instrumentation volume mount to the container
* added check for instrumentation specific environment variable value

PR fixes https://github.com/open-telemetry/opentelemetry-operator/issues/940

### DIFF
--- a/internal/webhookhandler/helper.go
+++ b/internal/webhookhandler/helper.go
@@ -1,0 +1,13 @@
+package webhookhandler
+
+import corev1 "k8s.io/api/core/v1"
+
+// Checks if Pod is already instrumented by checking Instrumentation InitContainer presence
+func IsPodInstrumentationMissing(pod corev1.Pod) bool {
+	for _, cont := range pod.Spec.InitContainers {
+		if cont.Name == initContainerName {
+			return false
+		}
+	}
+	return true
+}
--- a/internal/webhookhandler/webhookhandler.go
+++ b/internal/webhookhandler/webhookhandler.go
@@ -35,6 +35,8 @@ import (
 // +kubebuilder:rbac:groups=opentelemetry.io,resources=instrumentations,verbs=get;list;watch
 // +kubebuilder:rbac:groups="apps",resources=replicasets,verbs=get;list;watch
 
+const initContainerName = "opentelemetry-auto-instrumentation"
+
 var _ WebhookHandler = (*podSidecarInjector)(nil)
 
 // WebhookHandler is a webhook handler that analyzes new pods and injects appropriate sidecars into it.
@@ -89,11 +91,13 @@ func (p *podSidecarInjector) Handle(ctx context.Context, req admission.Request) 
 	}
 
 	for _, m := range p.podMutators {
-		pod, err = m.Mutate(ctx, ns, pod)
-		if err != nil {
-			res := admission.Errored(http.StatusInternalServerError, err)
-			res.Allowed = true
-			return res
+		if IsPodInstrumentationMissing(pod) {
+			pod, err = m.Mutate(ctx, ns, pod)
+			if err != nil {
+				res := admission.Errored(http.StatusInternalServerError, err)
+				res.Allowed = true
+				return res
+			}
 		}
 	}
 
--- a/pkg/instrumentation/helper.go
+++ b/pkg/instrumentation/helper.go
@@ -14,7 +14,12 @@
 
 package instrumentation
 
-import corev1 "k8s.io/api/core/v1"
+import (
+	"strings"
+
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+)
 
 // Calculate if we already inject InitContainers.
 func IsInitContainerMissing(pod corev1.Pod) bool {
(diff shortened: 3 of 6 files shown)


internal/webhookhandler/helper.go
```diff
@@ -0,0 +1,13 @@
+package webhookhandler
```
yuriolisa, line 1: Please include the template header to fix the lint errror

internal/webhookhandler/helper.go
```diff
@@ -0,0 +1,13 @@
...
+
+import corev1 "k8s.io/api/core/v1"
+
+// Checks if Pod is already instrumented by checking Instrumentation InitContainer presence
```
yuriolisa, line 5: Again just a small one.

pkg/instrumentation/helper.go
```diff
@@ -25,3 +30,21 @@ func IsInitContainerMissing(pod corev1.Pod) bool {
 	}
 	return true
 }
+
+// Check if opentelemetry-auto-instrumentation volume is on the list.
+func IsOtAIVolumeMissing(volumeMounts []corev1.VolumeMount, logger logr.Logger) bool {
```
yuriolisa, line 35: Could you please include tests that will validate these new functions?

yuriolisa (changes requested):

pkg/instrumentation/helper.go
mat-rumian: Yeah, sure :) I forgot about it. Thanks!

internal/webhookhandler/webhookhandler.go
```diff
@@ -35,6 +35,8 @@ import (
 // +kubebuilder:rbac:groups=opentelemetry.io,resources=instrumentations,verbs=get;list;watch
 // +kubebuilder:rbac:groups="apps",resources=replicasets,verbs=get;list;watch
 
+const initContainerName = "opentelemetry-auto-instrumentation"
```
pavolloffay, line 38: This package should not know anything about the consumers (one of which is auto-instrumentation pod mutator). 

there are a couple of possible solutions 
* move the code to the auto-instrumentation pod mutator
* add a new method to the pod mutator interface
* create a check func/type and pass it to the webhook factory method

internal/webhookhandler/helper.go
```diff
@@ -0,0 +1,27 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package webhookhandler
+
+import corev1 "k8s.io/api/core/v1"
+
+// Checks if Pod is already instrumented by checking Instrumentation InitContainer presence.
+func IsPodInstrumentationMissing(pod corev1.Pod) bool {
```
pavolloffay, line 20: I don't want to be nitpicking but a better name could be used here:  e.g. `IsAutoInstrumentationInjected`

internal/webhookhandler/webhookhandler.go
mat-rumian: I've moved code to `instrumentation`.

internal/webhookhandler/helper.go
mat-rumian: Agree, it's more meaningful ;)

pkg/instrumentation/helper.go
```diff
@@ -25,3 +29,31 @@ func IsInitContainerMissing(pod corev1.Pod) bool {
 	}
 	return true
 }
+
+// Check if opentelemetry-auto-instrumentation volume is on the list.
+func IsOtAIVolumeMissing(volumeMounts []corev1.VolumeMount) bool {
```
pavolloffay, line 34: all functions in this package can be made private

pavolloffay: nit: consider renaming to `isInstVolumeInjected`.

pkg/instrumentation/helper.go
```diff
@@ -25,3 +29,31 @@ func IsInitContainerMissing(pod corev1.Pod) bool {
...
+	for _, volumeMount := range volumeMounts {
+		if volumeMount.Name == volumeName {
+			return false
+		}
+	}
+	return true
+}
+
+// Check if EnvVar value contains instrumentation string.
+func IsEnvVarValueInstrumentationMissing(envVar corev1.EnvVar, instrumentation string) bool {
```
pavolloffay, line 44: This is pretty much redundant function the contains on the env var value can be called directly

pkg/instrumentation/javaagent.go
```diff
@@ -50,12 +50,18 @@ func injectJavaagent(logger logr.Logger, javaSpec v1alpha1.Java, pod corev1.Pod,
 			logger.Info("Skipping javaagent injection, the container defines JAVA_TOOL_OPTIONS env var value via ValueFrom", "container", container.Name)
 			return pod
 		}
-		container.Env[idx].Value = container.Env[idx].Value + javaJVMArgument
+
+		if IsEnvVarValueInstrumentationMissing(container.Env[idx], javaJVMArgument) {
```
pavolloffay, line 54: Why is this check needed? 

The `IsAutoInstrumentationInjected` is not enough?

pkg/instrumentation/javaagent.go
```diff
@@ -50,12 +50,18 @@ func injectJavaagent(logger logr.Logger, javaSpec v1alpha1.Java, pod corev1.Pod,
...
+			container.Env[idx].Value = container.Env[idx].Value + javaJVMArgument
+		}
+	}
+
+	if IsOtAIVolumeMissing(container.VolumeMounts) {
```
pavolloffay, line 59: Why is this check needed? 

The `IsAutoInstrumentationInjected` is not enough?

mat-rumian: So `IsOtAIVolumeMissing` and `IsEnvVarValueInstrumentationMissing` were applied before I've added check for pod mutation. When Pod was mutated few times k8s was throwing errors like `duplicated volume name`. In case of env var it happened that e.g. `javaJVMArgument` was added few times to `JAVA_TOOL_OPTIONS`. So I left this checks for safety for other unexpected mutations.

pavolloffay: I would prefer to remove this and `IsEnvVarValueInstrumentationMissing` as well and rely only on the init container for simplicity. 

We should keep it here only if there is a known scenario which the init container does not cover

PR #955, opened 2022-06-28
chore(nodejs): update to 0.31.0

mat-rumian:
PR update OT NodeJS dependencies to  the latest  `OTLP gRPC` exporter and instrumentations


yuriolisa:
@mat-rumian, rather than switch the exporter, would be possible to create a kind of parameter that the end-user will be capable to decide?

mat-rumian:
@yuriolisa OTLP HTTP json is the most used exporter for JS community - 3x times based on the npmjs package repository statistics. I think it was also the first implemented exporter for OT JS. I also asked on the OT-JS community channel which exporter they consider as "default" one.

mat-rumian:
@pavolloffay @yuriolisa so after some talks on OT-JS channel it came out that there's a fix for `OTLP gRPC` exporter related to insecure flag - https://github.com/open-telemetry/opentelemetry-js/pull/3019

I propose to wait till new exporter will be released and then update the PR and do not change exporter.

yuriolisa:
@mat-rumian, I guess we could proceed further with this PR due to https://github.com/open-telemetry/opentelemetry-js/pull/3019 was already merged.

mat-rumian:
@yuriolisa I've updated `OTLP gRPC` exporter and instrumentations, also tested if traces are working fine.

pavolloffay (approved):

Issue #972, opened 2022-07-12
Bump - Upgrade Operator-sdk to latest/1.22.1


yuriolisa:

pavolloffay:
done in #982

PR #982, opened 2022-07-14
Upgrade operator-sdk

yuriolisa:
Signed-off-by: Yuri Sa <yurimsa@gmail.com>
Resolver #972


yuriolisa:
@pavolloffay, @jpkrohling, could you please take a look?

yuriolisa:
> The scorecard images need to be upgraded as well



> The scorecard images need to be upgraded as well

I've changed them.

Makefile
```diff
@@ -218,20 +223,29 @@ cmctl:
...
+KUSTOMIZE ?= $(LOCALBIN)/kustomize
+CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
+ENVTEST ?= $(LOCALBIN)/setup-envtest
+
+## Tool Versions
+KUSTOMIZE_VERSION ?= v3.8.7
+CONTROLLER_TOOLS_VERSION ?= v0.8.0
 
-KUSTOMIZE = $(shell pwd)/bin/kustomize
+KUSTOMIZE_INSTALL_SCRIPT ?= "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"
```
jpkrohling, line 234: use a fixed version, otherwise it will make builds non-reproducible

Makefile
```diff
@@ -218,20 +223,29 @@ cmctl:
...
 .PHONY: kustomize
-kustomize: ## Download kustomize locally if necessary.
-	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v3,v3.8.7)
+kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary.
+$(KUSTOMIZE): $(LOCALBIN)
+	curl -s $(KUSTOMIZE_INSTALL_SCRIPT) | bash -s -- $(subst v,,$(KUSTOMIZE_VERSION)) $(LOCALBIN)
```
jpkrohling, line 238: is there another way of doing this without `bash | curl`? I myself frown upon this, as I would have to inspect the script before running this make target, instead of just looking at the Makefile

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
-  - manager.yaml
\ No newline at end of file
+- manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
```
jpkrohling, line 7: should this be here?

Makefile
yuriolisa: I've followed the upgrade recommendations of operator-sdk; most likely, they are not so effective.

config/manager/kustomization.yaml
yuriolisa: I just removed it.

Makefile
yuriolisa: Right, I've fixed version 4.5.5.

Makefile
yuriolisa: I've rolled back to use the previous way with the v4 of `kustomize`.

jpkrohling (approved):

PR #1000, opened 2022-07-23
Enable instrumentation injecting only core SDK config

### LINKED ISSUE #990: Set environment variables without injection
Hey folks, it would be nice for the Instrumentation CRD to support adding OTEL environment variables to a Pod without injecting the init container / updating libraries and modifying env vars such as `NODE_OPTIONS`.

My use case is that I have many services written in Go, Rust, C++, etc. which aren't supported by the Instrumentation CRD but which I'd like to configure in the same way.

I could re-use the autoinstrumentation for another language like using an annotation `instrumentation.../language-python: "true"` on a Rust application PodSpec, but that feels a bit wrong, and I would expect this not to be supported.

Is there a way to set the generated OTEL environment variables on Pods which aren't using a supported language?

bilbof:
Closes #990.

Adds support for injecting SDK environment variables only,
without injecting language-specific libraries or config.

Usage:

```bash
instrumentation.opentelemetry.io/inject-sdk: "true"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,11 @@ DotNet:
 instrumentation.opentelemetry.io/inject-dotnet: "true"
 ```
 
+OpenTelemetry SDK environment variables only:
+```bash
+instrumentation.opentelemetry.io/inject-sdk: "true"
+```
+
 The possible values for the annotation can be
 * `"true"` - inject and `Instrumentation` resource from the namespace.
 * `"my-instrumentation"` - name of `Instrumentation` CR instance in the current namespace.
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -273,9 +278,21 @@ spec:
...
-The Dockerfiles for auto-instrumentation can be found in [autoinstrumentation directory](./autoinstrumentation). 
+The Dockerfiles for auto-instrumentation can be found in [autoinstrumentation directory](./autoinstrumentation).
 Follow the instructions in the Dockerfiles on how to build a custom container image.
 
+#### Inject OpenTelemetry SDK environment variables only
+
+You can configure the OpenTelemetry SDK for applications which can't
+currently be autoinstrumented by using `inject-sdk` in place of (e.g.)
+`inject-python` or `inject-java`. This will inject environment variables
+like `OTEL_RESOURCE_ATTRIBUTES` but will not actually provide the SDK via an
```
pavolloffay, line 289: > via an init container

This is an implementation detail and it will not be true for e.g. Golang instrumentation. therefore it should not be mentioned here.

tests/e2e/instrumentation-sdk/00-install-instrumentation.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: opentelemetry.io/v1alpha1
+kind: Instrumentation
+metadata:
+  name: sdk-only
+spec:
+  env:
+    - name: OTEL_TRACES_EXPORTER
```
pavolloffay, line 7: I would prefer to remove all this OTEL  env vars rely on the env vars derived from the `spec.`

bilbof:
@pavolloffay thanks for your feedback, I've amended the commit

README.md
```diff
@@ -273,9 +278,20 @@ spec:
     image: your-customized-auto-instrumentation-image:python
 ```
 
-The Dockerfiles for auto-instrumentation can be found in [autoinstrumentation directory](./autoinstrumentation). 
+The Dockerfiles for auto-instrumentation can be found in [autoinstrumentation directory](./autoinstrumentation).
 Follow the instructions in the Dockerfiles on how to build a custom container image.
 
+#### Inject OpenTelemetry SDK environment variables only
+
+You can configure the OpenTelemetry SDK for applications which can't
```
pavolloffay, line 286: nit we don't use line breaks in the readme.

README.md
```diff
@@ -273,9 +278,20 @@ spec:
...
+currently be autoinstrumented by using `inject-sdk` in place of (e.g.)
+`inject-python` or `inject-java`. This will inject environment variables
```
pavolloffay, line 288: We should mention as well reporter and sampling config.

pavolloffay (commented):
LGTM. just some nits around docs

bilbof:
@pavolloffay thanks for reviewing. I've updated the readme and rebased.

pavolloffay (approved):

PR #1006, opened 2022-07-26
BUG FIX: Version Label not properly splitting for Otel Images

blakeromano:
When specifying the image to `otel/opentelemetry-collector-contrib` or `otel/opentelemetry-collector` The version is getting set to latest on service which is not correct. This should fix that and properly set the version label.

### DIFF
--- a/pkg/collector/labels_test.go
+++ b/pkg/collector/labels_test.go
@@ -116,3 +116,25 @@ func TestSelectorLabels(t *testing.T) {
 	// verify
 	assert.Equal(t, expected, result)
 }
+
+func TestVersionLabelImage(t *testing.T) {
+	otelcol := v1alpha1.OpenTelemetryCollector{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "my-opentelemetry-collector", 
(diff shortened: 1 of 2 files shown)


pkg/collector/labels_test.go
```diff
@@ -116,3 +116,25 @@ func TestSelectorLabels(t *testing.T) {
...
 }
+
+func TestVersionLabelImage(t *testing.T) {
+	otelcol := v1alpha1.OpenTelemetryCollector{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "my-opentelemetry-collector", 
+			Namespace: "my-namespace",
+		},
+		Spec: v1alpha1.OpenTelemetryCollectorSpec{
+			Image: "otel/opentelemetry-collector-contrib:0.56.0",
```
frzifus, line 127: Can we add another test case with a custom image?
e.g.:
```
Image: "company/opentelemetry-collector-custom:golden-stuff",
```

From my understanding we should support that too, right?

jaronoff97:
@blakeromano I'm not sure this is still relevant, but please let me know if it is. Thank you!

PR #1027, opened 2022-08-09
Update the target allocator on any manifest change

### LINKED ISSUE #1008: [target-allocator] TA pod doesn't restart when changing config settings
Trying to debug some issues I have in my statefulset cluster, and attempted to disable `prometheusCR`, to stop discovering serviceMonitor jobs.  Ran into an issue when setting `prometheusCR: false`, where the target-allocator does not restart after applying these changes.  

In my collector config
```
spec:
  mode: {{ .Values.collector.mode }}
  image: {{ .Values.collector.image }}
  replicas: 3
  targetAllocator:
    enabled: true
    image: ghcr.io/open-telemetry/opentelemetry-operator/target-allocator:latest
    serviceAccount: {{ .Release.Name }}-collector-targetallocator
    prometheusCR:
      enabled: false
```

### What I expected
I expected the target-allocator pod to restart and the output of `curl http://lightstep-collector-targetallocator:80/jobs` to exclude any jobs that were discovered by serviceMonitors.

### What actually happened
No restart occurred of my target-allocator pod and even after deleting the target-allocator pod (`kubectl delete po target-allocator`), the output of `curl http://lightstep-collector-targetallocator:80/jobs` was not updated to reflect disabling `prometheusCR`

In fact the only way to get the update I expected was to completely delete my opentelemetry namespace (collectors, target-allocator, and otel-operator) and reinstall the pods.

In case its helpful I attached the configs of the collector and target-allocator

`k get OpenTelemetryCollector lightstep-collector -o yaml`: [bug_attachment_otc.txt](https://github.com/open-telemetry/opentelemetry-operator/files/9203124/bug_attachment_otc.txt)

`k get deployment lightstep-collector-targetallocator -o yaml`: [bug_attachment_ta.txt](https://github.com/open-telemetry/opentelemetry-operator/files/9203129/bug_attachment_ta.txt)

jaronoff97:
This PR closes #1008 by removing the special target allocator behavior for not updating the container's spec. The desired behavior is to update the TA on a spec change. This also fixes a superfluous print from the operator that clogs operator logs due to the fact that the prometheus receiver has no endpoint to expose as it is a scrape based receiver.

### Evidence of testing:

```
  targetAllocator:
    ...
    prometheusCR:
      enabled: false
```

There was no prometheusCR flag initially
Screen Shot 2022-08-09 at 4 09 04 PM

```
  targetAllocator:
    ...
    prometheusCR:
      enabled: true
```
Screen Shot 2022-08-09 at 4 09 25 PM
Screen Shot 2022-08-09 at 4 09 35 PM

### DIFF
--- a/pkg/collector/reconcile/deployment_test.go
+++ b/pkg/collector/reconcile/deployment_test.go
@@ -209,7 +209,8 @@ func TestExpectedDeployments(t *testing.T) {
 		assert.NoError(t, err)
 		assert.True(t, exists)
 		assert.Equal(t, orgUID, actual.OwnerReferences[0].UID)
-		assert.NotEqual(t, expectedTADeploy.Spec.Template.Spec.Containers[0].Image, actual.Spec.Template.Spec.Containers[0].Image)
+		assert.True(t, len(actual.Spec.Template.Spec.Containers[0].Args) == 1)
+		assert.True(t, actual.Spec.Template.Spec.Containers[0].Args[0] == "--enable-prometheus-cr-watcher")
 		assert.Equal(t, int32(1), *actual.Spec.Replicas)
 	})
 
--- a/pkg/collector/reconcile/deployment_test.go
+++ b/pkg/collector/reconcile/deployment_test.go
@@ -168,15 +168,14 @@ func TestExpectedDeployments(t *testing.T) {
 		assert.Equal(t, int32(2), *actual.Spec.Replicas)
 	})
 
-	t.Run("should not update target allocator deployment when the container image is not updated", func(t *testing.T) {
+	t.Run("should update target allocator deployment when the container image is updated", func(t *testing.T) {
 		ctx := context.Background()
 		createObjectIfNotExists(t, "test-targetallocator", &expectedTADeploy)
 		orgUID := expectedTADeploy.OwnerReferences[0].UID
 
 		updatedParam, err := newParams("test/test-img", "")
 		assert.NoError(t, err)
-		updatedDeploy := targetallocator.Deployment(updatedParam.Config, logger, param.Instance)
-		*updatedDeploy.Spec.Replicas = int32(3)
+		updatedDeploy := targetallocator.Deployment(updatedParam.Config, logger, updatedParam.Instance)
 
 		err = expectedDeployments(ctx, param, []v1.Deployment{updatedDeploy})
 		assert.NoError(t, err)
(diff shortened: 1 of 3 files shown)


pkg/collector/reconcile/deployment_test.go
```diff
@@ -209,7 +209,8 @@ func TestExpectedDeployments(t *testing.T) {
 		assert.NoError(t, err)
 		assert.True(t, exists)
 		assert.Equal(t, orgUID, actual.OwnerReferences[0].UID)
-		assert.NotEqual(t, expectedTADeploy.Spec.Template.Spec.Containers[0].Image, actual.Spec.Template.Spec.Containers[0].Image)
+		assert.True(t, len(actual.Spec.Template.Spec.Containers[0].Args) == 1)
+		assert.True(t, actual.Spec.Template.Spec.Containers[0].Args[0] == "--enable-prometheus-cr-watcher")
```
secustor, line 213:
```suggestion
		assert.ElementsMatch(t, actual.Spec.Template.Spec.Containers[0].Args, []string{"--enable-prometheus-cr-watcher"})
```

This would simplify the test setup, in the case there are more options added.

pkg/collector/reconcile/deployment_test.go
```diff
@@ -168,15 +168,14 @@ func TestExpectedDeployments(t *testing.T) {
...
-	t.Run("should not update target allocator deployment when the container image is not updated", func(t *testing.T) {
+	t.Run("should update target allocator deployment when the container image is updated", func(t *testing.T) {
 		ctx := context.Background()
 		createObjectIfNotExists(t, "test-targetallocator", &expectedTADeploy)
 		orgUID := expectedTADeploy.OwnerReferences[0].UID
 
 		updatedParam, err := newParams("test/test-img", "")
 		assert.NoError(t, err)
-		updatedDeploy := targetallocator.Deployment(updatedParam.Config, logger, param.Instance)
-		*updatedDeploy.Spec.Replicas = int32(3)
```
secustor, line 179: Why do you have removed the replica update? Would it more meaningful if we change the test expected test result to 3?

jaronoff97: this test actually didn't change at all, the git diff is just getting confused from the new name. I'm going to change the order of the new test to hopefully better reflect that.

secustor (approved):

Aneurysm9 (approved):

jpkrohling (approved):
Approving, as the code owner (@Aneurysm9) has approved.

PR #1032, opened 2022-08-11
Fix "sidecar.opentelemetry.io/injected" label value sometimes being longer than 63 characters

### LINKED ISSUE #1031: Label "sidecar.opentelemetry.io/injected" value sometimes longer than 63 characters
With an `OpenTelemetryCollector` in `sidecar` mode, the `sidecar.opentelemetry.io/injected` label gets added to the pod using the format `[otc_namespace].[otc_name]`

https://github.com/open-telemetry/opentelemetry-operator/blob/07b9f22df8a87852464bf4877d793afcddae37c3/pkg/sidecar/pod.go#L48

Under the right conditions, the label value can exceed 63 characters, which causes Kubernetes controllers to produce the following error:

```
Error creating: Pod "[name]-59ff54b4c6-9mh4g" is invalid: metadata.labels: Invalid value: "[otc_namespace].[otc_name]": must be no more than 63 characters
```

(Replace brackets with sufficiently long values.)

If the name of the `OpenTelemetryCollector` does not identify it as such, this error message is particularly cryptic as it is not obvious what is producing the invalid label. (Took me a while to dig through various controllers and mutating webhooks to find the culprit. Though arguably this is more an issue with how Kubernetes reports such errors.)

Possible solutions:

1. Move `sidecar.opentelemetry.io/injected` to annotations
    - Annotations do not have the same restrictions on values as labels
2. Allow user to disable addition of `sidecar.opentelemetry.io/injected` label
    - Presumably this would be an option in the `OpenTelemetryCollector` resource
3. Split the `[otc_namespace]` and `[otc_name]` across multiple labels
    - This is how I've seen other projects handle such cases. Since Kubernetes object names and label values have the same restrictions, you can safely put the `[otc_namespace]` in one label and the `[otc_name]` in another, guaranteeing that they will never exceed the character limit.

Would be happy to open a PR with your preferred solution. Thanks 🙂

KevinSnyderCodes:
Fixes #1031 

Use two new labels:

- `sidecar.opentelemetry.io/injected-namespace`: namespace of the `OpenTelemetryCollector`
- `sidecar.opentelemetry.io/injected-name`: name of the `OpenTelemetryCollector`

Since Kubernetes object names and label values have the same restrictions, we can safely use these two new labels with guarantee that it will not produce errors.

For backwards compatibility, we continue to set the `sidecar.opentelemetry.io/injected` label _only if_ it does not exceed 63 characters. This ensures that it still exists when valid, but does not cause errors when invalid.


pavolloffay:
I would like to understand how the operator uses `sidecar.opentelemetry.io/injected"`. The only reference to this label I found is https://github.com/open-telemetry/opentelemetry-operator/blob/928e9e4102dfbb35df737c8bb8fcce54bcfb854f/pkg/sidecar/pod.go#L48 - which just sets the label, but I could not find a reading of that label in the codebase.

@jpkrohling since you are the author, could you please take a look and explain what use-cases `sidecar.opentelemetry.io/injected` solves?

KevinSnyderCodes:
At the very least, can we make it so the label does not get created if it exceeds 63 characters? As it stands the injector is producing an invalid manifest, which causes pods to not get created.

pavolloffay:
If the label is not being used then we could just stop setting it altogether.

KevinSnyderCodes:
Sure, but we don't know how long it's going to take to get an answer to that question. A fix to the bug that would get us unblocked would be prudent IMO.

pavolloffay:
cc) @jpkrohling could you please take a look at the question above?

KevinSnyderCodes:
@pavolloffay How long are we willing to wait for a response before introducing a backwards-compatible fix? This continues to impact our Kubernetes environments.

yuriolisa:
@KevinSnyderCodes, did you have the opportunity to check it out?

jaronoff97:
@KevinSnyderCodes I just rebased your PR but was thinking you could actually just use our naming package's `Truncate` function seen [here](https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/naming/main.go#L54-L56) rather than only setting the label name. Let me know if you have time / want to work on that, if not I can take on this PR.

KevinSnyderCodes:
@jaronoff97 That sounds good to me. I no longer use this package in my work. Feel free to take over this PR 👍

jaronoff97:
I think was fiixed by this https://github.com/open-telemetry/opentelemetry-operator/pull/2250

Issue #1034, opened 2022-08-11
Allocator maps accessed outside of their mutex


kristinapathak:
https://github.com/open-telemetry/opentelemetry-operator/blob/07b9f22df8a87852464bf4877d793afcddae37c3/cmd/otel-allocator/allocation/allocator.go#L38-L48
`TargetItems` and `collectors` are normally protected by a mutex, but are being accessed in a few places without the mutex being locked first.

TargetItems:
https://github.com/open-telemetry/opentelemetry-operator/blob/07b9f22df8a87852464bf4877d793afcddae37c3/cmd/otel-allocator/main.go#L174
https://github.com/open-telemetry/opentelemetry-operator/blob/07b9f22df8a87852464bf4877d793afcddae37c3/cmd/otel-allocator/main.go#L184
https://github.com/open-telemetry/opentelemetry-operator/blob/07b9f22df8a87852464bf4877d793afcddae37c3/cmd/otel-allocator/allocation/http.go#L26

collectors:
https://github.com/open-telemetry/opentelemetry-operator/blob/07b9f22df8a87852464bf4877d793afcddae37c3/cmd/otel-allocator/allocation/http.go#L51

This is a data race and can cause issues with the endpoints providing the correct information.

I would like to fix this, by:
- modifying the mutex to be a `RWMutex`
- no longer exporting `TargetItems`
- adding two functions for the `Allocator` that each provide a copy of their respective maps - names could be something like `Collectors()` and `TargetItems()` or `GetCollectors()` and `GetTargetItems()`

A follow up question: If we shallow copy the map values, they will still be pointing to the same objects.  How deep do we need to copy everything? Is it enough to dereference the pointers?

PR #1040, opened 2022-08-17
Protect allocator maps behind mutex, create getter funcs for them

kristinapathak:
fixes https://github.com/open-telemetry/opentelemetry-operator/issues/1034

`Allocator.targetItems` and `Allocator.collectors` can no longer be accessed outside of Allocator functions.  Instead, two functions that copy the respective maps and return them for read-only use are exported.

### DIFF
--- a/cmd/otel-allocator/allocation/allocator.go
+++ b/cmd/otel-allocator/allocation/allocator.go
@@ -88,7 +88,6 @@ func (allocator *Allocator) Collectors() map[string]*collector {
 		collectorsCopy[k] = v
 	}
 	return collectorsCopy
-
 }
 
 // findNextCollector finds the next collector with less number of targets.
--- a/cmd/otel-allocator/allocation/allocator.go
+++ b/cmd/otel-allocator/allocation/allocator.go
@@ -57,7 +57,7 @@ type collector struct {
 // Users need to call SetTargets when they have new targets in their
 // clusters and call Reshard to process the new targets and reshard.
 type Allocator struct {
-	m sync.Mutex
+	m sync.RWMutex
 
 	targetsWaiting map[string]TargetItem // temp buffer to keep targets that are waiting to be processed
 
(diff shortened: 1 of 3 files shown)


cmd/otel-allocator/allocation/allocator.go
```diff
@@ -53,17 +53,39 @@ type collector struct {
 // Users need to call SetTargets when they have new targets in their
 // clusters and call Reshard to process the new targets and reshard.
 type Allocator struct {
-	m sync.Mutex
+	m sync.RWMutex
```
Aneurysm9, line 56: This should either have a comment detailing the fields it is protecting, be grouped together with the fields it is protecting, or both.  That appears to be everything but the logger at this point.

cmd/otel-allocator/allocation/allocator.go
```diff
@@ -53,17 +53,39 @@ type collector struct {
...
+	collectorsCopy := make(map[string]*collector)
+	for k, v := range allocator.collectors {
+		collectorsCopy[k] = v
+	}
+	return collectorsCopy
+}
+
 // findNextCollector finds the next collector with less number of targets.
 func (allocator *Allocator) findNextCollector() *collector {
 	var col *collector
```
Aneurysm9, line 91: This needs to take a read lock.  As do any other methods that read from protected members.

cmd/otel-allocator/allocation/allocator.go
```diff
@@ -133,26 +155,26 @@ func (allocator *Allocator) ReallocateCollectors() {
...
 	defer allocator.m.Unlock()
-	allocator.TargetItems = make(map[string]*TargetItem)
+	allocator.targetItems = make(map[string]*TargetItem)
 	allocator.processWaitingTargets()
 }
 
 // removeOutdatedTargets removes targets that are no longer available.
 func (allocator *Allocator) removeOutdatedTargets() {
-	for k := range allocator.TargetItems {
+	for k := range allocator.targetItems {
```
Aneurysm9, line 164: Maybe add a comment here indicating that this is expected to be protected under the lock from `AllocateTargets()`.

cmd/otel-allocator/allocation/allocator.go
```diff
@@ -133,26 +155,26 @@ func (allocator *Allocator) ReallocateCollectors() {
...
+			delete(allocator.targetItems, k)
 		}
 	}
 }
 
 // processWaitingTargets processes the newly set targets.
 func (allocator *Allocator) processWaitingTargets() {
 	for k, v := range allocator.targetsWaiting {
-		if _, ok := allocator.TargetItems[k]; !ok {
+		if _, ok := allocator.targetItems[k]; !ok {
```
Aneurysm9, line 175: Maybe add a comment here indicating that this is expected to be protected under the lock from `AllocateTargets()`.

cmd/otel-allocator/allocation/allocator.go
kristinapathak: `findNextCollector()` is called from within a lock.  I'll add a comment so that's clear.

Aneurysm9 (approved):

pavolloffay (approved):

pavolloffay:
PR needs to be rebased

PR #1041, opened 2022-08-18
Fix unnecessary and incorrect reallocation

### LINKED ISSUE #1038: [target-allocator] Target allocator assigns target to multiple collectors
I ran the target allocator with discovery via the config yaml (not using CRDs), and saw the TA assigning the same target to multiple collectors. This resulted in a target being scraped multiple times in a minute even though the `scrape_interval` is set to 1m. Also, I have noticed that sometimes it takes the target allocator 15 minutes to refresh and discover the desired endpoints, while other times the TA is able to discover the endpoints right away.

### What I expected:
Target allocator is able to discover all the endpoints when started. TA assigns each collector a unique set of targets.

### What actually happened:
TA discovered some endpoints initially when it was started at 19:02, but it took until the watch routine restart at 19:17 to discover the desired endpoints. Then the first period after discovery looks ok, but following the restart of the watch routine, each subsequent period had at least one target that was duplicated.

collector-0
```
// 19:18 - 19:32
2022-08-16T19:32:04.243Z -> service.instance.id: STRING(172.17.0.4:9121)

// After watch routine restart at 19:47:52
2022-08-16T19:52:04.240Z -> service.instance.id: STRING(172.17.0.4:9121)
2022-08-16T19:52:17.227Z -> service.instance.id: STRING(172.17.0.3:9121)
```

collector-1 
```
// 19:18 - 19:32
2022-08-16T19:32:35.003Z -> service.instance.id: STRING(172.17.0.3:9121)
2022-08-16T19:32:48.234Z -> service.instance.id: STRING(172.17.0.5:9121)

// After watch routine restart at 19:47:52
2022-08-16T19:52:35.020Z -> service.instance.id: STRING(172.17.0.3:9121)
2022-08-16T19:52:46.528Z -> service.instance.id: STRING(172.17.0.4:9121)
2022-08-16T19:52:48.236Z -> service.instance.id: STRING(172.17.0.5:9121)
```

TA logs
```
{"level":"info","ts":1660676571.953488,"msg":"Starting the Target Allocator"}
{"level":"info","ts":1660676572.0064685,"logger":"setup","msg":"Starting server..."}
{"level":"info","ts":1660676572.0111964,"logger":"allocator","msg":"Successfully started a collector pod watcher","component":"opentelemetry-targetallocator"}
{"level":"info","ts":1660677472.3938131,"logger":"allocator","msg":"Restarting watch routine","component":"opentelemetry-targetallocator"}
{"level":"info","ts":1660677472.400775,"logger":"allocator","msg":"Successfully started a collector pod watcher","component":"opentelemetry-targetallocator"}
{"level":"info","ts":1660678372.403873,"logger":"allocator","msg":"Restarting watch routine","component":"opentelemetry-targetallocator"}
{"level":"info","ts":1660678372.4173315,"logger":"allocator","msg":"Successfully started a collector pod watcher","component":"opentelemetry-targetallocator"}
{"level":"info","ts":1660679272.4196055,"logger":"allocator","msg":"Restarting watch routine","component":"opentelemetry-targetallocator"}
{"level":"info","ts":1660679272.4243846,"logger":"allocator","msg":"Successfully started a collector pod watcher","component":"opentelemetry-targetallocator"}
{"level":"info","ts":1660680172.4269636,"logger":"allocator","msg":"Restarting watch routine","component":"opentelemetry-targetallocator"}
{"level":"info","ts":1660680172.4400752,"logger":"allocator","msg":"Successfully started a collector pod watcher","component":"opentelemetry-targetallocator"}
```

jaronoff97:
Closes #1038, solves an issue where the allocator re-assigns collectors unnecessarily. This is accomplished by removing the unnecessary logic around allocating targets and collectors as they come in. Previously, the collectors were entirely cleared whenever the `SetCollectors()` method was called. A similar amount of confusion happened with discovered targets where the `targetsWaiting` map which was used for decisions in allocation would be cleared, allowing for a scenario where one target could be assigned to multiple collectors. 

The new logic combines the `SetWaitingTargets` and `processWaitingTargets` methods in to one to help clarify the logic. Logic is also simplified for `SetCollectors` which simply reassigns all targets discovered after the changed collector list is processed.

Before this change:

Screen Shot 2022-08-18 at 1 42 51 PM

After this change:

Screen Shot 2022-08-18 at 1 43 13 PM


cmd/otel-allocator/allocation/allocator.go
```diff
@@ -88,119 +88,129 @@ func (allocator *Allocator) Collectors() map[string]*collector {
...
 	defer allocator.m.Unlock()
 	if len(collectors) == 0 {
 		log.Info("No collector instances present")
 		return
+	} else if allocator.allCollectorsPresent(collectors) {
+		log.Info("No changes to the collectors found")
+		return
 	}
+
+	// Clear existing collectors
```
Aneurysm9, line 192: Food for future thought: it would be ideal if we didn't need to completely reset the `allocator.collectors` map here, but only remove those that are not in `collectors` and mark their targets for reallocation.  This would avoid unnecessarily moving targets away from collectors that are still running.

cmd/otel-allocator/allocation/allocator.go
```diff
@@ -88,119 +88,129 @@ func (allocator *Allocator) Collectors() map[string]*collector {
...
+	targetItem := TargetItem{
+		JobName:   target.JobName,
+		Link:      LinkJSON{fmt.Sprintf("/jobs/%s/targets", url.QueryEscape(target.JobName))},
+		TargetURL: target.TargetURL,
+		Label:     target.Label,
+		Collector: chosenCollector,
+	}
+	allocator.targetItems[targetItem.hash()] = &targetItem
+	chosenCollector.NumTargets++
+	targetsPerCollector.WithLabelValues(chosenCollector.Name).Set(float64(chosenCollector.NumTargets))
```
secustor, line 122:
```suggestion
        if _, found := allocator.targetItems[target.hash()]; found {
		target.Collector.NumTargets--
		targetsPerCollector.WithLabelValues(target.Collector.Name).Set(float64(target.Collector.NumTargets))
	}
	allocator.targetItems[targetItem.hash()] = &targetItem
	chosenCollector.NumTargets++
	targetsPerCollector.WithLabelValues(chosenCollector.Name).Set(float64(chosenCollector.NumTargets))
```
We have to update the metrics if the target has already been assigned

cmd/otel-allocator/allocation/allocator.go
```diff
@@ -88,119 +88,129 @@ func (allocator *Allocator) Collectors() map[string]*collector {
...
+
+	// Clear existing collectors
 	for k := range allocator.collectors {
 		delete(allocator.collectors, k)
 	}
 
+	// Insert the new collectors
 	for _, i := range collectors {
 		allocator.collectors[i] = &collector{Name: i, NumTargets: 0}
 	}
```
secustor, line 200:
```suggestion

	missing, surplus := allocator.getCollectorChanges(collectors)

	// Clear existing collectors
	for _, k := range surplus {
		delete(allocator.collectors, k)
	}

	// Insert the new collectors
	for _, i := range missing {
		allocator.collectors[i] = &collector{Name: i, NumTargets: 0}
	}
```

cmd/otel-allocator/allocation/allocator.go
```diff
@@ -88,119 +88,129 @@ func (allocator *Allocator) Collectors() map[string]*collector {
...
-				TargetURL: v.TargetURL,
-				Label:     v.Label,
-				Collector: col,
-			}
-			col.NumTargets++
-			targetsPerCollector.WithLabelValues(col.Name).Set(float64(col.NumTargets))
-			allocator.targetItems[v.hash()] = &targetItem
-		}
+	// Re-Allocate the existing targets
+	for _, item := range allocator.targetItems {
```
secustor, line 203:
```suggestion
	// find targets which need to be redistributed
	var redistribute []*TargetItem
	for _, item := range allocator.targetItems {
		for _, s := range surplus {
			if item.Collector.Name == s {
				redistribute = append(redistribute, item)
			}
		}
	}

	// Re-Allocate the existing targets
	for _, item := range redistribute {
```

cmd/otel-allocator/allocation/allocator.go
```diff
@@ -88,119 +88,129 @@ func (allocator *Allocator) Collectors() map[string]*collector {
...
-		} else {
-			if v.NumTargets < col.NumTargets {
-				col = v
-			}
-		}
-
-	}
-	return col
+	collectorsAllocatable.Set(float64(len(collectors)))
 }
```
secustor, line 209:
```suggestion

// getCollectorChanges returns the missing and surplus collectors respectively.
// This method is called from within SetCollectors, which acquires the needed lock.
func (allocator *Allocator) getCollectorChanges(collectors []string) ([]string, []string) {
	var missing []string
	var surplus []string
	for _, s := range collectors {
		if _, found := allocator.collectors[s]; !found {
			missing = append(missing, s)
		}
	}
	for k := range allocator.collectors {
		found := false
		for _, s := range collectors {
			if k == s {
				found = true
			}
		}
		if !found {
			surplus = append(surplus, k)
		}
	}
	return missing, surplus
}

```

cmd/otel-allocator/allocation/allocator.go
secustor: I have added suggestions for that feature.

cmd/otel-allocator/allocation/allocator.go
jaronoff97: This method assumes that the target hasn't been assigned already. This is only called after the collectors are cleared or when a new target has been found in the `tempTargetMap`. If we were to decrement the number of targets here we would be reporting incorrectly

cmd/otel-allocator/allocation/allocator.go
```diff
@@ -88,119 +88,129 @@ func (allocator *Allocator) Collectors() map[string]*collector {
...
+	tempTargetMap := make(map[string]TargetItem, len(targets))
+	for _, target := range targets {
+		tempTargetMap[target.hash()] = target
+	}
+
+	// Check for removals
+	for k, target := range allocator.targetItems {
+		// if the old target is no longer in the new list, remove it
+		if _, ok := tempTargetMap[k]; !ok {
+			allocator.collectors[target.Collector.Name].NumTargets--
```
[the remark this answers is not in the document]
jaronoff97: ah yes we totally do, thank you :D

cmd/otel-allocator/allocation/allocator.go
jaronoff97: Thank you for the suggestion! I implemented this method, and it allows me to replace the other method i made "allCollectorsPresent" because this essentially checks the same thing.

cmd/otel-allocator/allocation/allocator.go
secustor: This will be the case if you apply my suggestion.    
If you don't, then this assumption should be part of the Method documentation.

cmd/otel-allocator/allocation/allocator.go
jaronoff97: thank you both for the suggestion. I have implemented as much and added a test for this scenario.

cmd/otel-allocator/allocation/allocator.go
jaronoff97: this is only the case where targets are reassigned i believe, maybe the right solution is to just set the metric to 0 when we remove the unused collectors.

jaronoff97: i will also include that assumption in the comment.

jaronoff97: honestly, the right solution here would be an asynchronous gauge instrument that tracks this information so we don't need to be smart about updating it. cc @Aneurysm9 for when we migrate this to the otel sdk

secustor: If we keep this assumption we have to set it to  0 and let it then disappear.  Else the aggregations will show false data.

jaronoff97: Yep, i set it to 0 now. Because of how prometheus works, you can't delete the information entirely though. Does the approach I have out there now look correct?

cmd/otel-allocator/allocation/allocator.go
```diff
@@ -88,119 +88,150 @@ func (allocator *Allocator) Collectors() map[string]*collector {
...
-				Link:      LinkJSON{fmt.Sprintf("/jobs/%s/targets", url.QueryEscape(v.JobName))},
-				TargetURL: v.TargetURL,
-				Label:     v.Label,
-				Collector: col,
+	// find targets which need to be redistributed
+	var redistribute []*TargetItem
+	for _, item := range allocator.targetItems {
+		for _, s := range removedCollectors {
+			if item.Collector.Name == s {
+				redistribute = append(redistribute, item)
```
Aneurysm9, line 219:
```suggestion
				redistribute = append(redistribute, item)
				break
```

Maybe a micro-optimization, but no need to check any further collectors once you've found a match.

Aneurysm9 (approved):

cmd/otel-allocator/allocation/allocator.go
jaronoff97: yep, but good practice to do :)

cmd/otel-allocator/allocation/allocator.go
```diff
@@ -88,119 +88,151 @@ func (allocator *Allocator) Collectors() map[string]*collector {
...
 	log := allocator.log.WithValues("component", "opentelemetry-targetallocator")
+	timer := prometheus.NewTimer(timeToAssign.WithLabelValues("SetCollectors"))
+	defer timer.ObserveDuration()
 
 	allocator.m.Lock()
 	defer allocator.m.Unlock()
+	newCollectors, removedCollectors := allocator.getCollectorChanges(collectors)
 	if len(collectors) == 0 {
 		log.Info("No collector instances present")
 		return
```
kristinapathak, line 198: I think we should move this check above mutex locking - if there are no `collectors`, we shouldn't do any unnecessary work like acquiring a lock or finding the collector changes.

cmd/otel-allocator/allocation/allocator_test.go
```diff
@@ -120,6 +117,84 @@ func TestAllocationCollision(t *testing.T) {
...
+	}
+	// test that targets and collectors are added properly
+	s.SetTargets(targetList)
+
+	// verify
+	expectedTargetLen := len(initTargets)
+	targetItems := s.TargetItems()
+	assert.Len(t, targetItems, expectedTargetLen)
+
+	// assign new set of collectors with the same names
```
kristinapathak, line 181:
```suggestion
	// assign new set of collectors with the one different name
```

cmd/otel-allocator/allocation/allocator.go
secustor: What happens here if we scale the collectors to 0? If I'm not mistaken in that case the previously set collectors will still be kept.

jaronoff97: If we were to scale collectors to 0, there would be nothing querying the target allocator so nothing would observe incorrect information. As soon as that count is changed, we would observe reallocation occur.

secustor: True, but the `collectorsAllocatable` metric would still show the last number of allocatable collectors

jaronoff97: ah that's right, I'll update the metric at the beginning of this method.

secustor (approved):

jpkrohling (approved):

jpkrohling:
Approving by proxy (code owners approved), and merging.

PR #1049, opened 2022-08-20
feat(target-allocator): use jobs endpoint for job gathering

### LINKED ISSUE #383: Will opentelemetry-operator support prometheus CRDs, like ServiceMonitor

### LINKED ISSUE #713: Use `/job` endpoint instead of generated promconfig
Currently the operator uses jobs generated by the `promconfig` to extract a list of relevant jobs from the target allocator. 

This is working as long as all target updates are generated by the operator reconciliation, but this prevents the implementation of other sources in the TargetAllocator, e.g. custom implementation of the TargetAllocatorAPI or PrometheusCR. 

To resolve this following solutions have been proposed: 

>1. Call the /jobs endpoint from the Operator.
    This is not ideal as this would introduce a requirement that the TargetAllocator has to be reachable by the operator. I think this would be the most flexible solution regarding the addition of new watcher in the TargetAllocator or BYO target allocators.
>   2.  Add a catch all job for PrometheusCR jobs if this feature is enabled.
    In that case the metrics generated would not be the same as the ones by the PrometheusOperator, as the job label would differ between the two implementations.
> 3.   Move the PrometheusCRD ( ServiceMonitor and PodMonitor ) watcher into the operator and only provide the resolved configuration to the TargetAllocator
    This can lead to high load on big clusters, as the operator would watch the CRDs for all Collector instances
>  4.  Create an [Observer](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/observer) for the TargetAllocator and extend the ReceiverCreator to allow looping over observer results.
    This seems to be cleanest solution IMO.
  

Further @Aneurysm9  has proposed to add the functionality to retrieve the jobs directly in `prometheusreceiver` in this [PR](https://github.com/open-telemetry/opentelemetry-operator/pull/653#issuecomment-1024397482).  
> What about having the prometheus receiver watch the /jobs endpoint and update its configuration as appropriate. This could be managed by a config flag to limit impact where the target allocator is not in use and would not require creating new receivers or reloading pipelines.

This IMO is the cleanest solutions, as it would not add any constrains and promises to be most efficient.

### LINKED ISSUE #963: [target-allocator] `ServiceMonitor` must rely on an empty job
At present, when we use servicemonitor, we must set an empty job as a placeholder for config replace
eg:

```yaml
    receivers:
      prometheus:
        config:
          scrape_configs:
          - job_name: serviceMonitor/vke-system/harbor-server/0
```

image

servicemonitor is dynamically created, so there is no way to preset an empty job. Can we consider generating these scrape job configurations automatically?

secustor:
Closes #963
Closes #713
Closes #383

This PR enables dynamic job gathering of `prometheusreceiver` with the target allocator as source.     
Most of the logic has been implemented in the collector and make Prometheus config manipulation in the operator mostly unnecessary.


nlamirault:
Hi ! Any news on this feature ?

secustor:
This PR has been superseded and this should already work if you manually build a collector and operator image and use it.   
The next will update the default collector image to the necessary one and will include the changes.

secustor:
Solved in Release [v0.61.0](https://github.com/open-telemetry/opentelemetry-operator/releases/tag/v0.61.0)

PR #1055, opened 2022-08-24
Upgrade operator-sdk to 1.23.0

iblancasa:
Upgrade `operator-sdk` version from `1.22.1` to `1.23.0`


Makefile
```diff
@@ -234,22 +234,25 @@ ENVTEST ?= $(LOCALBIN)/setup-envtest
...
+CONTROLLER_TOOLS_VERSION ?= v0.9.2
 
 
+KUSTOMIZE_INSTALL_SCRIPT ?= "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"
 .PHONY: kustomize
-kustomize: ## Download kustomize locally if necessary.
-	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v4,$(KUSTOMIZE_VERSION))
+kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary.
+$(KUSTOMIZE): $(LOCALBIN)
+	test -s $(LOCALBIN)/kustomize || { curl -s $(KUSTOMIZE_INSTALL_SCRIPT) | bash -s -- $(subst v,,$(KUSTOMIZE_VERSION)) $(LOCALBIN); }
```
yuriolisa, line 244: Could we avoid using curl or bash to do that step?

Makefile
```diff
@@ -234,22 +234,25 @@ ENVTEST ?= $(LOCALBIN)/setup-envtest
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v4.5.5
-CONTROLLER_TOOLS_VERSION ?= v0.8.0
+CONTROLLER_TOOLS_VERSION ?= v0.9.2
 
 
+KUSTOMIZE_INSTALL_SCRIPT ?= "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"
```
yuriolisa, line 240: Is it required to use an external script to install it?

yuriolisa (changes requested):
We should improve how to install the Kustomize.

Makefile
iblancasa: I followed the recommendation from the [official operator-sdk upgrade guide](https://v1-23-x.sdk.operatorframework.io/docs/upgrading-sdk-version/v1.23.0/#gov3-fix-makefile-targets-to-not-download-binaries-when-they-are-found-already). I can recover the way `kustomize` was installed before. What do you think?

This comment applies to https://github.com/open-telemetry/opentelemetry-operator/pull/1055#discussion_r953558192 too

frzifus: Maybe installing it from source could be an option?

```bash
GOBIN=$(pwd)/ GO111MODULE=on go install sigs.k8s.io/kustomize/kustomize/v4@latest
```
See: 
https://kubectl.docs.kubernetes.io/installation/kustomize/source/#install-the-kustomize-cli-from-source-without-cloning-the-repo

iblancasa: We can just do what was done previously:
```makefile
kustomize: ## Download kustomize locally if necessary.
        $(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v4,$(KUSTOMIZE_VERSION))
```
I don't have a strong opinion about this

iblancasa: Something interesting about @frzifus suggestion approach: I'm not able to install a specific version. It only works with `latest`. For instance:
```sh
$ GOBIN=$(pwd)/bin GO111MODULE=on go install sigs.k8s.io/kustomize/kustomize/v4@v4.2.0

go: sigs.k8s.io/kustomize/kustomize/v4@v4.2.0 (in sigs.k8s.io/kustomize/kustomize/v4@v4.2.0):
	The go.mod file for the module providing named packages contains one or
	more exclude directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.

$ GOBIN=$(pwd)/bin GO111MODULE=on go install sigs.k8s.io/kustomize/kustomize/v4@latest

$ ./bin/kustomize version
{Version:unknown GitCommit:$Format:%H$ BuildDate:1970-01-01T00:00:00Z GoOs:linux GoArch:amd64}
```

Issue #1065, opened 2022-08-30
Allow configuration of target CPU Utilization for HPA


moh-osman3:
Currently the HPA [code](https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/collector/horizontalpodautoscaler.go#L40) hard codes 90% as the target cpu utilization.  I am currently running tests with HPA and it would be helpful to be able to change the target utilization to a custom value.  Opening this issue to track my change to add this field to the collector config.

ajsaclayan:
Can I piggy back off this request to scale on memory as well? Otherwise I can create a new issue.

pavolloffay:
@moh-osman3 could you please as well comment why do you need a custom value fo HPA? What is the use-case?

moh-osman3:
@pavolloffay The purpose of adding `TargetCPUUtilization` to the CRD is because we wish to tune this value and provide a recommended setting to users.  Previously with `90%` we are experiencing a lot of flapping when load testing.  If targets are uneven and one of the collectors has significantly fewer targets than other collectors, it becomes less likely that scale up will happen (average utilization will be too low, this can lead to throttling in collectors with too many targets which can cause scaling down).  Giving users the control over this value might allow them to tune the HPA to their specific needs, if the default is unhelpful.

Previously talked with @jaronoff97 about adding the option to scale on memory for workloads with fluctuating memory needs.  I can work on that in a followup PR in the next few weeks, if you don't mind creating an issue for this @ajsaclayan and specifying the use case?

moh-osman3:
Resolved by https://github.com/open-telemetry/opentelemetry-operator/pull/1066

PR #1066, opened 2022-08-30
[HPA] Add targetCPUUtilization field to collector config

moh-osman3:
Implementation for https://github.com/open-telemetry/opentelemetry-operator/issues/1065

Adding targetCPUUtilization (percentage) to the collector config instead of hard coding the value.


pavolloffay:
@moh-osman3 please sign the CLA

pkg/collector/horizontalpodautoscaler.go
```diff
@@ -37,7 +37,12 @@ func HorizontalPodAutoscaler(cfg config.Config, logger logr.Logger, otelcol v1al
 	labels["app.kubernetes.io/name"] = naming.Collector(otelcol)
 
 	annotations := Annotations(otelcol)
-	cpuTarget := defaultCPUTarget
+	var cpuTarget int32
+	if otelcol.Spec.TargetCPUUtilization != nil {
+		cpuTarget = *otelcol.Spec.TargetCPUUtilization
+	} else {
+		cpuTarget = defaultCPUTarget
```
pavolloffay, line 44: the default value of  `otelcol.Spec.TargetCPUUtilization should be set in the defaulting webhook.

apis/v1alpha1/opentelemetrycollector_webhook.go
```diff
@@ -129,6 +129,10 @@ func (r *OpenTelemetryCollector) validateCRDSpec() error {
 			return fmt.Errorf("the OpenTelemetry Spec autoscale configuration is incorrect, minReplicas should be one or more")
 		}
 
+		if r.Spec.TargetCPUUtilization != nil && (*r.Spec.TargetCPUUtilization < int32(1) || *r.Spec.TargetCPUUtilization > int32(99)) {
```
jaronoff97, line 132: We should also be setting a default in this file

tests/e2e/autoscale/00-assert.yaml
```diff
@@ -6,11 +6,17 @@ status:
...
-apiVersion: autoscaling/v1
+apiVersion: autoscaling/v2beta2
 kind: HorizontalPodAutoscaler
 metadata:
   name: simplest-collector
 spec:
   minReplicas: 1
   maxReplicas: 2
-
+  metrics:
```
jaronoff97, line 16: Should we keep a test for the default behavior?

jaronoff97 (commented):
Left a comment about where to add in the default behavior, also a question about testing

pkg/collector/horizontalpodautoscaler.go
moh-osman3: Thanks for the review! moved the setting of default to opentelemetrycollector_webhook.go

apis/v1alpha1/opentelemetrycollector_webhook.go
moh-osman3: Thanks for the review! Setting the default in opentelemetrycollector_webhook.go now

tests/e2e/autoscale/00-assert.yaml
moh-osman3: Added a new OpenTelemetryCollector that does not set `targetCPUUtilization` for the default.

apis/v1alpha1/opentelemetrycollector_webhook.go
```diff
@@ -62,6 +62,12 @@ func (r *OpenTelemetryCollector) Default() {
 		one := int32(1)
 		r.Spec.Replicas = &one
 	}
+
+	// if autoscaling is enabled then set default targetCPUUtilization
+	if r.Spec.MaxReplicas != nil && r.Spec.TargetCPUUtilization == nil {
```
pavolloffay, line 67: is this guard needed. Perhaps we could default always.

pkg/collector/reconcile/horizontalpodautoscaler.go
```diff
@@ -119,13 +120,23 @@ func setAutoscalerSpec(params Params, autoscalingVersion autodetect.AutoscalingV
 			} else {
 				updated.(*autoscalingv2beta2.HorizontalPodAutoscaler).Spec.MinReplicas = &one
 			}
+			if params.Instance.Spec.TargetCPUUtilization != nil {
```
pavolloffay, line 123: Is this check needed? The defaulting webhook should make sure the value is always set.

pkg/collector/reconcile/horizontalpodautoscaler.go
```diff
@@ -119,13 +120,23 @@ func setAutoscalerSpec(params Params, autoscalingVersion autodetect.AutoscalingV
...
+				updated.(*autoscalingv2beta2.HorizontalPodAutoscaler).Spec.Metrics[0].Resource.Target.AverageUtilization = &ninety
+			}
 		} else {
 			updated.(*autoscalingv2.HorizontalPodAutoscaler).Spec.MaxReplicas = *params.Instance.Spec.MaxReplicas
 			if params.Instance.Spec.MinReplicas != nil {
 				updated.(*autoscalingv2.HorizontalPodAutoscaler).Spec.MinReplicas = params.Instance.Spec.MinReplicas
 			} else {
 				updated.(*autoscalingv2.HorizontalPodAutoscaler).Spec.MinReplicas = &one
 			}
+			if params.Instance.Spec.TargetCPUUtilization != nil {
```
pavolloffay, line 135: the same question as above.

pavolloffay:
linting failed

apis/v1alpha1/opentelemetrycollector_webhook.go
moh-osman3: Hmm based on your comment below I think I included the guard bc I was worried the default might override a user set value (and I also think it makes it clear this setting is only used when autoscaling is enabled). Removed the guard here and the ones in your comment below.

pkg/collector/reconcile/horizontalpodautoscaler.go
moh-osman3: Removed.

pkg/collector/reconcile/horizontalpodautoscaler.go
moh-osman3: Removed.

moh-osman3:
> linting failed

I think that was because I had some spaces instead of tabs in the webhook file.  `golangci-lint run` is not throwing errors for me locally anymore. Can you run again please?

pavolloffay (approved):

pavolloffay:
@kevinearls could you please review as well?

kevinearls (approved):
LGTM

pavolloffay:
@moh-osman3 CI failed

moh-osman3:
> @moh-osman3 CI failed

@pavolloffay Sorry for delay I was on PTO last week! Yeah I noticed CI was failing due to unit test error 

```
--- FAIL: TestHPA (0.00s)
    --- FAIL: TestHPA/v2 (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1cf2364]
```

This was resolved by adding back the nil check in `pkg/collector/horizontalpodautoscaler.go`

I also noticed that after I applied https://github.com/open-telemetry/opentelemetry-operator/pull/1066/commits/e54ef2ac7a6e2d76627ff8ce298d0f64f75d3d3d my e2e tests are failing but unable to figure out the exact issue. I think I'm a little confused why the default set in the webhook isn't being picked up in `horizontalpodautoscaler.go`.  So I reverted that change - currently I think the only test failing is `Security` for CI.  Wondering if you have any guidance to get the tests passing?

apis/v1alpha1/opentelemetrycollector_webhook.go
moh-osman3: Seems this guard and the next comment's are needed otherwise the e2e tests begin to fail

pkg/collector/horizontalpodautoscaler.go
```diff
@@ -28,16 +28,20 @@ import (
...
 	labels := Labels(otelcol, cfg.LabelsFilter())
 	labels["app.kubernetes.io/name"] = naming.Collector(otelcol)
 
 	annotations := Annotations(otelcol)
-	cpuTarget := defaultCPUTarget
+	var cpuTarget int32
+	if otelcol.Spec.TargetCPUUtilization != nil {
+		cpuTarget = *otelcol.Spec.TargetCPUUtilization
+	} else {
+		cpuTarget = int32(90)
```
pavolloffay, line 42: I don't like that the default CPU target utilization is defined in two places: defaulting webhook and here. There should be only one place where defaults are defined - preferably in the defaulting webhook.

pkg/collector/reconcile/horizontalpodautoscaler.go
```diff
@@ -119,13 +120,23 @@ func setAutoscalerSpec(params Params, autoscalingVersion autodetect.AutoscalingV
 			} else {
 				updated.(*autoscalingv2beta2.HorizontalPodAutoscaler).Spec.MinReplicas = &one
 			}
+			if params.Instance.Spec.TargetCPUUtilization != nil {
+				updated.(*autoscalingv2beta2.HorizontalPodAutoscaler).Spec.Metrics[0].Resource.Target.AverageUtilization = params.Instance.Spec.TargetCPUUtilization
+			} else {
+				updated.(*autoscalingv2beta2.HorizontalPodAutoscaler).Spec.Metrics[0].Resource.Target.AverageUtilization = &ninety
```
pavolloffay, line 126: The same here - the default is defined in two places

pavolloffay:
the PR needs to be rebased

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -41,13 +41,15 @@ type OpenTelemetryCollectorSpec struct {
...
 	// Autoscaler specifies the pod autoscaling configuration to use
 	// for the OpenTelemetryCollector workload.
 	//
 	// +optional
 	Autoscaler *AutoscalerSpec `json:"autoscaler,omitempty"`
-
+	// TargetCPUUtilization sets the target average CPU used across all replicas.
+	// If average CPU exceeds this value, the HPA will scale up. Defaults to 90 percent.
+	// +optional
+	TargetCPUUtilization *int32 `json:"targetCPUUtilization,omitempty"`
```
pavolloffay, line 52: @kevinearls could you please review this PR.

I am curious if this should be moved into `.spec.autoscaler.`

pavolloffay: > In the future we should put any other metrics in spec.autoscaler too.

We should put this comment somewhere in the CRD or book a ticket.

moh-osman3: Thanks for pointing this out! I moved `TargetCPUUtilization` to the appropriate place in the spec now.  I also created https://github.com/open-telemetry/opentelemetry-operator/issues/1115 to track moving `MaxReplicas` and `MinReplicas` to `.spec.autoscaler`.  I don't mind taking this issue on in a followup PR.

pavolloffay (approved):

kevinearls (approved):
LGTM

PR #1068, opened 2022-08-30
Introduce ability to specify strategies for target allocation

jaronoff97:
## Goals:
* Introduce a new interface to allow a user of the TA to describe an allocation strategy
  * Allocation strategies are given the current state and the new state and should return what the next state for the allocator should be
  * Each strategy should be idempotent with minimal to no side effects
* Refactor the allocator's current strategy - least weighted – to be the first strategy

## Summary
This PR introduces the AllocationStrategy interface, implements it through a refactor of the current allocator, and changes the configuration of the TA to allow for a user to specify an option. The operator's CRDs are also updated to allow the strategy to be specified in the Collector CRD. New tests have been introduced to confirm the strategy works as expected. All previous tests should continue working.


cmd/otel-allocator/allocation/least_weighted.go
```diff
@@ -0,0 +1,110 @@
+package allocation
```
secustor, line 1: This should be own package along with the factory function.

pkg/targetallocator/container.go
```diff
@@ -50,6 +52,9 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
 	if otelcol.Spec.TargetAllocator.PrometheusCR.Enabled {
 		args = append(args, "--enable-prometheus-cr-watcher")
 	}
+	if len(otelcol.Spec.TargetAllocator.AllocationStrategy) > 0 {
+		args = append(args, fmt.Sprintf("--allocation-strategy=%s", otelcol.Spec.TargetAllocator.AllocationStrategy))
```
secustor, line 56: Maybe we should put this in the config file we have already 🤔.

cmd/otel-allocator/allocation/allocator.go
```diff
@@ -47,103 +58,96 @@ func (t TargetItem) hash() string {
...
+func NewState(collectors map[string]collector, targetItems map[string]TargetItem) State {
+	return State{collectors: collectors, targetItems: targetItems}
+}
+
+type changes[T any] struct {
+	additions map[string]T
+	removals  map[string]T
+}
+
+func diff[T any](current, new map[string]T) changes[T] {
```
secustor, line 83: Such functions along with its types, should be in an utility package IMHO.

cmd/otel-allocator/allocation/allocator.go
```diff
@@ -47,103 +58,96 @@ func (t TargetItem) hash() string {
...
 }
 
+type State struct {
+	// collectors is a map from a collector's name to a collector instance
+	collectors map[string]collector
+	// targetItems is a map from a target item's hash to the target items allocated state
+	targetItems map[string]TargetItem
+}
+
+func NewState(collectors map[string]collector, targetItems map[string]TargetItem) State {
```
secustor, line 74: Why should this function be exposed?

secustor (commented):
First glance of the code 😉

cmd/otel-allocator/allocation/least_weighted.go
Aneurysm9: Agreed.  I was thinking about the visibility of some of the types and methods and was wondering whether the `Allocator` and the allocation strategy implementations shouldn't live in separate packages to ensure that they can be implemented without using any internals.  This would be important to allow someone to bring their own allocation strategy and plug it in.

cmd/otel-allocator/allocation/allocator.go
Aneurysm9: It can't be used outside of this package as it is since it requires an unexported type as one of its arguments.

cmd/otel-allocator/allocation/allocator.go
```diff
@@ -47,103 +58,96 @@ func (t TargetItem) hash() string {
...
+}
+
+type AllocatorStrategy interface {
+	Allocate(currentState, newState State) State
+}
+
+func NewStrategy(kind string) (AllocatorStrategy, error) {
+	if kind == "least-weighted" {
+		return LeastWeightedStrategy{}, nil
+	}
```
Aneurysm9, line 112: How about inverting this and making a registry.  Strategies can register themselves in an `init()` function and this function should just be a map lookup.

```go
type AllocatorProvider func() AllocatorStrategy

var registry map[string]AllocatorProvider

func NewStrategy(name string) (AllocatorStrategy, error) {
  if p, ok := registry[name]; ok {
    return p(), nil
  }
  return nil, errors.New(...)
}

func RegisterStrategy(name string, provider AllocatorProvider) error {
  if _, ok := registry[name]; ok {
    return errors.New("already registered")
  }
  registry[name] = provider
  return nil
}
```

cmd/otel-allocator/main.go
```diff
@@ -47,7 +48,9 @@ func main() {
 	ctx := context.Background()
 
 	log := ctrl.Log.WithName("allocator")
-	allocator := allocation.NewAllocator(log)
+
+	strategy, _ := allocation.NewStrategy(*cliConf.AllocationStrategy)
```
Aneurysm9, line 52: Catch this error here.  The application can't start if there is no valid allocation strategy.

jaronoff97: yep, this was because of a refactor by intellij

pkg/targetallocator/container.go
jaronoff97: Which file would that be?

cmd/otel-allocator/allocation/least_weighted.go
jaronoff97: moving this to a new package means we lose the ability to use fields and types we don't exported like `State` and `collector`

Aneurysm9: `State` is exported, `collector` is not.  Would it make sense to export `Collector`?  I think so.

pkg/targetallocator/container.go
secustor: Config of the TA: 
https://github.com/open-telemetry/opentelemetry-operator/blob/main/cmd/otel-allocator/config/config.go#L33

Operator code which generates this file: https://github.com/open-telemetry/opentelemetry-operator/blob/19c186df1b89a3898169a16f91bd13fab5a6931f/pkg/collector/reconcile/configmap.go#L94

cmd/otel-allocator/discovery/discovery.go
```diff
@@ -3,15 +3,17 @@ package discovery
 import (
 	"context"
 
+	"github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/allocation/strategy"
```
Aneurysm9, line 6: This should be grouped with the other import from this repo.

cmd/otel-allocator/discovery/discovery_test.go
```diff
@@ -7,15 +7,17 @@ import (
 	"sort"
 	"testing"
 
+	"github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/allocation/strategy"
```
Aneurysm9, line 10: Group this with the other imports from this repo.

cmd/otel-allocator/main.go
```diff
@@ -126,11 +145,7 @@ type server struct {
 	server           *http.Server
 }
 
-func newServer(log logr.Logger, allocator *allocation.Allocator, discoveryManager *lbdiscovery.Manager, cliConf config.CLIConfig) (*server, error) {
-	k8sclient, err := configureFileDiscovery(log, allocator, discoveryManager, context.Background(), cliConf)
-	if err != nil {
-		return nil, err
-	}
+func newServer(log logr.Logger, allocator *allocation.Allocator, discoveryManager *lbdiscovery.Manager, k8sclient *collector.Client, listenAddr *string) (*server, error) {
```
Aneurysm9, line 148: Not clear why this change is being made.  It seems unrelated to the allocation strategy change.

jaronoff97: This needs to be made in order to load the config which contains the desired allocation strategy at the start of main, rather than in a method called in newServer. Prior to the instantiation of the allocator, we need an instance of the strategy.

cmd/otel-allocator/allocation/strategy/strategy.go
```diff
@@ -0,0 +1,130 @@
...
+
+func (s State) TargetItems() map[string]TargetItem {
+	return s.targetItems
+}
+
+func (s State) SetTargetItem(key string, value TargetItem) State {
+	next := s
+	next.targetItems[key] = value
+	return next
+}
```
Aneurysm9, line 108: I don't think this works as `next.targetItems` is [referencing the same underlying data](https://goplay.tools/snippet/kTXaPYuk6Nb) as `s.targetItems`.

jaronoff97: Yep, i changed it so that the setter just changes the internal map. Technically not immutable, but as you showed it doesn't really matter.

cmd/otel-allocator/allocation/allocator.go
```diff
@@ -33,156 +29,61 @@ var (
...
-			allocator.addTargetToTargetItems(&target)
-		}
+		tempTargetMap[target.Hash()] = target
 	}
+	newState := strategy.NewState(allocator.state.Collectors(), tempTargetMap)
+	allocator.state = allocator.strategy.Allocate(allocator.state, newState)
 }
 
-// SetCollectors sets the set of collectors with key=collectorName, value=Collector object.
+// SetCollectors sets the set of collectors with key=collectorName, value=CollectorName object.
```
kristinapathak, line 86: I don't think we need this comment change - it makes it seem like key == value.

cmd/otel-allocator/allocation/allocator.go
```diff
@@ -197,41 +98,21 @@ func (allocator *Allocator) SetCollectors(collectors []string) {
...
 }
 
-func NewAllocator(log logr.Logger) *Allocator {
+func NewAllocator(log logr.Logger, allocatorStrategy strategy.Allocator) *Allocator {
 	return &Allocator{
-		log:         log,
-		collectors:  make(map[string]*collector),
-		targetItems: make(map[string]*TargetItem),
+		log:      log,
+		state:    strategy.NewState(make(map[string]strategy.Collector), make(map[string]strategy.TargetItem)),
```
kristinapathak, line 115: If we don't need/use the maps here, shouldn't the `NewState()` function be able to handle `nil` values and make the maps itself?

cmd/otel-allocator/allocation/http.go
```diff
@@ -4,55 +4,43 @@ import (
...
-func GetAllTargetsByCollectorAndJob(collector string, job string, cMap map[string][]TargetItem, allocator *Allocator) []targetGroupJSON {
-	var tgs []targetGroupJSON
+func GetAllTargetsByCollectorAndJob(collector string, job string, cMap map[string][]strategy.TargetItem, allocator *Allocator) []strategy.TargetGroupJSON {
+	var tgs []strategy.TargetGroupJSON
 	group := make(map[string]string)
 	labelSet := make(map[string]model.LabelSet)
-	for _, col := range allocator.Collectors() {
-		if col.Name == collector {
+	for colName, _ := range allocator.Collectors() {
+		if colName == collector {
```
kristinapathak, line 40: The colName can only be in the map once - can we try to access it directly rather than traversing the whole map?

cmd/otel-allocator/allocation/strategy/strategy.go
```diff
@@ -0,0 +1,122 @@
...
+
+	// TargetsPerCollector records how many targets have been assigned to each collector
+	// It is currently the responsibility of the strategy to track this information.
+	TargetsPerCollector = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "opentelemetry_allocator_targets_per_collector",
+		Help: "The number of targets for each collector.",
+	}, []string{"collector_name"})
+)
+
+func New(name string) (Allocator, error) {
```
kristinapathak, line 26: Sorry if this is obvious or has already been discussed, but what is the benefit of the `init()` registration pattern? Couldn't we pass an interface into `New()` instead and not require global state? Any caller will always have to import their strategy (or strategies) and this package.

cmd/otel-allocator/allocation/strategy/strategy.go
```diff
@@ -0,0 +1,122 @@
...
+
+// Collector Creates a struct that holds Collector information
+// This struct will be parsed into endpoint with Collector and jobs info
+// This struct can be extended with information like annotations and labels in the future
+type Collector struct {
+	Name       string
+	NumTargets int
+}
+
+type State struct {
```
kristinapathak, line 89: All of this state stuff seems like it belongs in its own file.  I'm not sure if some of these other things should also be split up?

cmd/otel-allocator/allocation/strategy/strategy.go
```diff
@@ -0,0 +1,122 @@
...
+	// collectors is a map from a Collector's name to a Collector instance
+	collectors map[string]Collector
+	// targetItems is a map from a target item's hash to the target items allocated state
+	targetItems map[string]TargetItem
```
kristinapathak, line 93: The lock for these maps are up in `allocation.Allocator` but then they are also modified in the strategy - it seems easy to change something later and mess it up.

cmd/otel-allocator/allocation/strategy/strategy.go
Aneurysm9: Having a mapping of string identifiers to strategy implementations enables using the CRD or other configuration mechanisms to select from among multiple available strategies at runtime.  This could be implemented by the consuming system, but would likely result in duplicate implementation.

cmd/otel-allocator/allocation/strategy/strategy.go
Aneurysm9: This is why I think it is important that `State` is immutable and that strategies return a new `State` instance.  Then the locks held by `Allocator` are protecting the "active" `State` and anything modifying `State` doesn't need to worry about locking since it's immutable and always returning a new `State`.

kristinapathak: Yeah, I think only one object should be modifying the state if it's going to be modified - making it immutable would also solve the problem.

cmd/otel-allocator/allocation/strategy/strategy.go
kristinapathak: got it, thanks for explaining! 🙂

jaronoff97:
As discussed in slack, I'm going to close this PR in favor of a different approach where we instead make the allocator itself the interface instead of making the strategy a part of the allocator. See [this conversation](https://cloud-native.slack.com/archives/C033BJ8BASU/p1662490341290249?thread_ts=1662045544.594939&cid=C033BJ8BASU) for more detail

PR #1069, opened 2022-08-31
Change log message to V(2), be sure to pass strings so it doesn't panic

kevinearls:

### DIFF
--- a/internal/config/main.go
+++ b/internal/config/main.go
@@ -140,7 +140,7 @@ func (c *Config) AutoDetect() error {
 		return err
 	}
 	c.autoscalingVersion = hpaVersion
-	c.logger.V(2).Info("In Autodetect, Set HPA version to [", c.autoscalingVersion.String(), "] from [", hpaVersion.String(), "]")
+	c.logger.V(1).Info("autoscaling version detected", "autoscaling-version", c.autoscalingVersion.String())
 
 	return nil
 }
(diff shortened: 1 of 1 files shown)


pavolloffay:
@kevinearls could you please explain why the logger is changed to V(2)?

internal/config/main.go
```diff
@@ -140,7 +140,7 @@ func (c *Config) AutoDetect() error {
 		return err
 	}
 	c.autoscalingVersion = hpaVersion
-	c.logger.Info("In Autodetect, Set HPA version to [", c.autoscalingVersion, "] from [", hpaVersion, "]")
+	c.logger.V(2).Info("In Autodetect, Set HPA version to [", c.autoscalingVersion.String(), "] from [", hpaVersion.String(), "]")
```
kristinapathak, line 143: Aren't `autoscalingVersion` and `hpaVersion` the same value?

Also - the first string is a message but the remaining in the list are key value pairs, leaving the values as `] from [` and `]` - is this what we want?

I tried to fix this when I was working on an operator change and came up with:
```
c.logger.Info("In Autodetect, Set HPA version", "from", c.autoscalingVersion, "to", hpaVersion)
```

kevinearls:
@pavolloffay I changed this as it previously had been an Info message and was appearing every 30 seconds or 1 minute in the log.  It is not needed that often.

@kristinapathak I will update the message.

internal/config/main.go
pavolloffay: +1, after the log message follows a list of key-value pairs. so it should be something like `info("message line", "param1-name", param1-value)`

Regarding the `v(2)`. It makes sense to use the lowest value, however if the codebase already uses v(2) as a debug level then 2 is fine.

pavolloffay (approved):

Issue #1011, opened 2022-07-28
Add the ability to expose a port on a collector container


kristinapathak:
`collector.Container()` returns a `Container` that doesn't include `Port` information.

https://github.com/open-telemetry/opentelemetry-operator/blob/7ce297de5415486b568d68466a564fd65fe7d462/pkg/collector/container.go#L88-L98

Will this prevent the Prometheus PodMonitor from finding the collector? How can we add the port information? Are the container ports the same as `otelcol.Spec.Ports`?

https://github.com/open-telemetry/opentelemetry-operator/blob/7ce297de5415486b568d68466a564fd65fe7d462/apis/v1alpha1/opentelemetrycollector_types.go#L98

pavolloffay:
> Will this prevent the Prometheus PodMonitor from finding the collector? 

I don't know, perhaps if the pod monitor uses port names. Could you please report back if you played with this? I think it's worth fixing this.

kristinapathak:
I investigated this more - I set up a pod monitor for the collector while running the collecter as a statefulset.  The ports for the operator could be found in service discovery labels, but there weren't any ports with the collectors' labels.  I tried with `port` and `targetPort` (deprecated), but neither worked.

jpkrohling:
Stupid question, but can't you use a ServiceMonitor instead? From what I remember, we create a service explicitly meant to be used with the ServiceMonitor.

kristinapathak:
I believe this is a problem for both ServiceMonitor and PodMonitor but will try a ServiceMonitor scenario. 🙂

pavolloffay:
@kevinearls will be working on this sortly.

kevinearls:
Hi @kristinapathak Do you have an example CR I can use while working on this?  I don't have much experience working with prometheus and none with a collector instance that is using it.  Thanks.

kristinapathak:
Hi @kevinearls,

My setup is not public, but below is information on how I verified that the lack of ports is a problem.  I've attached my servicemonitor and podmonitor configurations as well as the helm chart and configuration files for prometheus.
[files.zip](https://github.com/open-telemetry/opentelemetry-operator/files/9378579/files.zip)

I am running the collector as a statefulset, using the following images.
- collector: `otel/opentelemetry-collector-contrib:0.58.0`
- operator: `ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator:0.56.0`
[...]

PR #1070, opened 2022-09-01
Expose container ports on the collector pod

kristinapathak:
fixes https://github.com/open-telemetry/opentelemetry-operator/issues/1011, adding a way to expose container ports for service and pod monitoring.

Uses functions from [this package](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation#IsValidPortName) to validate port name and number.  If the name is too long, we cut it off to fit.


pkg/collector/container.go
```diff
@@ -33,6 +35,22 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
 		image = cfg.CollectorImage()
 	}
 
+	// build port map
+	ports := getConfigContainerPorts(logger, otelcol.Spec.Config)
+	for _, p := range otelcol.Spec.ContainerPorts {
+		if len(p.Name) > 15 {
+			p.Name = p.Name[:15]
```
jaronoff97, line 42: Maybe we should log something here for the operator to make clear what is being truncated

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -81,6 +81,12 @@ type OpenTelemetryCollectorSpec struct {
 	// +optional
 	// +listType=atomic
 	Ports []v1.ServicePort `json:"ports,omitempty"`
+	// ContainerPorts allows a set of ports to be exposed by the underlying v1.Container. By default, the operator
+	// will attempt to infer the required ports by parsing the .Spec.Config property but this property can be
+	// used to open additional ports that can't be inferred by the operator, like for custom receivers.
```
jaronoff97, line 86: I believe the container ports in the spec will also override any inferred names, maybe we should add that in to this comment

pkg/collector/container.go
```diff
@@ -33,6 +35,22 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
...
+		}
+		if len(validation.IsValidPortName(p.Name)) == 0 && len(validation.IsValidPortNum(int(p.ContainerPort))) == 0 {
+			ports[p.Name] = p
+		}
+	}
+	ports["metrics"] = corev1.ContainerPort{
+		Name:          "metrics",
+		ContainerPort: 8888,
```
jaronoff97, line 50: iirc some people override the metrics port, we may not want to hardcode what the value is here

jaronoff97: We can pull the port from `spec.config.service.telemetry.metrics.address` i believe

pkg/collector/parser/receiver.go
```diff
@@ -133,6 +136,60 @@ func singlePortFromConfigEndpoint(logger logr.Logger, name string, config map[in
...
+
+	// tcplog and udplog receivers hold the endpoint
+	// value in `listen_address` field
+	case name == "tcplog" || name == "udplog":
+		endpoint = getAddressFromConfig(logger, name, listenAddressKey, config)
+
+	// ignore kubeletstats receiver as it holds the field key endpoint, and it
+	// is a scraper, we only expose endpoint through k8s service objects for
+	// receivers that aren't scrapers.
+	case name == "kubeletstats":
```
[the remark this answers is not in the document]
kristinapathak: `singleContainerPortFromConfigEndpoint()` is based on `singlePortFromConfigEndpoint()` so both functions have this logic. Would you prefer that I add configuration for this in this PR? I'm also happy to open an issue and fix this separately.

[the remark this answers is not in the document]
secustor: @pavolloffay What do you think about making this configurable?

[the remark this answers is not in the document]
jaronoff97: IMO this should be a separate issue given this logic is already in the operator. The concern is valid, but i think adding the logic to this PR is going to expand the scope of this.

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -81,6 +81,13 @@ type OpenTelemetryCollectorSpec struct {
 	// +optional
 	// +listType=atomic
 	Ports []v1.ServicePort `json:"ports,omitempty"`
+	// ContainerPorts allows a set of ports to be exposed by the underlying v1.Container. By default, the operator
+	// will attempt to infer the required ports by parsing the .Spec.Config property but this property can be
+	// used to open additional ports that can't be inferred by the operator, like for custom receivers.  Any ports
```
pavolloffay, line 86: > like for custom receivers.

I would assume that custom receiver ports should be defined in the `spec.ports` and the container ports are rather used for the ports that should not be exposed in the service. 


Do we even need to have service ports and container ports? We could use the service ports to open the ports on the container. Is there any use-case when container ports != service ports?

kristinapathak: I'm not sure - I included this for the flexibility and because the types are different.  I can change it so we parse the configured service ports into container ports if that's the direction we want to go in?

pkg/collector/parser/receiver.go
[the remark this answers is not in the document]
kristinapathak: I created an issue for this:
https://github.com/open-telemetry/opentelemetry-operator/issues/1095

apis/v1alpha1/opentelemetrycollector_types.go
kristinapathak: @pavolloffay, do you want me to change the logic to parse the container ports from the list of service ports? Or are you waiting for others' input on this?

pavolloffay: If there is no hard requirement to expose additional configuration in the CR then I would opt-in for deriving the ports automatically for now

kristinapathak:
Golang-ci is not running at all on target allocator code, so there are other linting issues that I will fix in addition to getting golang-ci to run on all modules in the repo.

pkg/collector/container.go
```diff
@@ -15,24 +15,56 @@
...
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/validation"
 
 	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
 	"github.com/open-telemetry/opentelemetry-operator/internal/config"
 	"github.com/open-telemetry/opentelemetry-operator/pkg/collector/adapters"
 	"github.com/open-telemetry/opentelemetry-operator/pkg/naming"
 )
 
+const maxPortLen = 15
```
pavolloffay, line 35: It would be good to add a link to the source that defines this limitation.

pkg/collector/container.go
```diff
@@ -15,24 +15,56 @@
...
+	for _, p := range otelcol.Spec.Ports {
+		truncName := naming.Truncate(p.Name, maxPortLen)
+		if p.Name != truncName {
+			logger.Info("truncating container port name",
+				"port.name.prev", p.Name, "port.name.new", truncName)
+		}
+		nameErrs := validation.IsValidPortName(truncName)
+		numErrs := validation.IsValidPortNum(int(p.Port))
+		if len(nameErrs) > 0 || len(numErrs) > 0 {
+			logger.Error(errInvalidPort, "dropping container port", "port.name", truncName, "port.num", p.Port,
```
pavolloffay, line 57: I wonder if we should rather emit k8s event to let the user know that the validation failed.

pavolloffay: @kevinearls you did some logging-related work recently. 

Do you think we should use k8s events here or normal logs?

pkg/collector/container.go
```diff
@@ -15,24 +15,56 @@
...
 // Container builds a container for the given collector.
 func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelemetryCollector) corev1.Container {
 	image := otelcol.Spec.Image
 	if len(image) == 0 {
 		image = cfg.CollectorImage()
 	}
 
+	// build container ports from service ports
+	ports := getConfigContainerPorts(logger, otelcol.Spec.Config)
+	for _, p := range otelcol.Spec.Ports {
```
pavolloffay, line 48: This function does a couple of different things:
1. name truncation
2. validation
3. create a container port

The truncation and validation should be done in the defaulting and validating webhook.  The otelcol spec received in this function should be valid and prepared for use.

pavolloffay: The validating webhook can as well return user-facing error if the name validation fails.

pkg/collector/container_test.go
```diff
@@ -57,6 +64,184 @@ func TestContainerWithImageOverridden(t *testing.T) {
...
+				},
+			},
+		},
+		{
+			description: "invalid port name",
+			specConfig:  goodConfig,
+			specPorts: []corev1.ServicePort{
+				{
+					// this port name contains a non alphanumeric character, which is invalid.
+					Name:     "-test🦄port",
```
pavolloffay, line 161: :P

pavolloffay (commented):
@kevinearls could you please review as well? You were looking into this in the past.

pkg/collector/container.go
kevinearls: @pavolloffay I don't really know, to be honest.  I think logging would be fine but we are not very consistent about this.

kevinearls (approved):
LGTM

jaronoff97 (commented):
I think the only change to be made is move some of the validation / mutation to the corresponding webhooks, otherwise this looks great. Awesome stuff 🥳

jaronoff97 (approved):

pkg/collector/container.go
kristinapathak: I can't add the current truncation to the defaulting webhook because it causes an import cycle:
```
package github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1
	imports github.com/open-telemetry/opentelemetry-operator/pkg/naming
	imports github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1: import cycle not allowed
```
Is it reasonable for us to error out when a Spec Port is too long (like any other port naming error)? Or should I truncate a different way?

pkg/collector/container.go
```diff
@@ -98,3 +121,93 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
...
+				logger.Info("truncating container port name",
+					"port.name.prev", p.Name, "port.name.new", truncName)
+			}
+			nameErrs := validation.IsValidPortName(truncName)
+			numErrs := validation.IsValidPortNum(int(p.Port))
+			if len(nameErrs) > 0 || len(numErrs) > 0 {
+				logger.Error(errInvalidPort, "dropping container port", "port.name", truncName, "port.num", p.Port,
+					"port.name.errs", nameErrs, "num.errs", numErrs)
+				continue
+			}
```
kristinapathak, line 148: I might be able to move this truncation+validation to inside the `adapters.ConfigToReceiverPorts()` function - what do you think, @pavolloffay?

apis/v1alpha1/opentelemetrycollector_webhook.go
```diff
@@ -120,9 +121,19 @@ func (r *OpenTelemetryCollector) validateCRDSpec() error {
...
+		if len(nameErrs) > 0 || len(numErrs) > 0 {
+			return fmt.Errorf("the OpenTelemetry Spec Ports configuration is incorrect, port name '%s' errors: %s, num '%d' errors: %s",
+				p.Name, nameErrs, p.Port, numErrs)
+		}
+	}
+
 	// validate autoscale with horizontal pod autoscaler
 	if r.Spec.MaxReplicas != nil {
-		if *r.Spec.MaxReplicas < int32(1) {
+		if *r.Spec.MaxReplicas <= int32(1) {
```
kristinapathak, line 136: I changed this to match the error string explanation (max replicase should be more than one) but I don't think there's a problem if maxReplicas is 1? I can change the error string instead of changing this expression if others agree.

secustor (approved):

apis/v1alpha1/opentelemetrycollector_webhook.go
kristinapathak: Moved: https://github.com/open-telemetry/opentelemetry-operator/pull/1136

I will remove these changes from this PR.

pkg/collector/container.go
kristinapathak: I tried to move it and it caused an import cycle. 😞 So I'm just going to leave the code as it is.

pavolloffay (approved):

PR #1075, opened 2022-09-05
Make sure we return the right version when autoscaling v2 is found

kevinearls:

### DIFF
--- a/pkg/autodetect/main.go
+++ b/pkg/autodetect/main.go
@@ -95,7 +95,7 @@ func (a *autoDetect) HPAVersion() (AutoscalingVersion, error) {
 
 			for _, version := range versions {
 				if version.Version == "v2" || version.Version == "v2beta2" {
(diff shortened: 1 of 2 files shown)


pkg/autodetect/main.go
```diff
@@ -117,10 +117,10 @@ func (v AutoscalingVersion) String() string {
 	return "unknown"
 }
 
-func toAutoscalingVersion(version string) AutoscalingVersion {
+func ToAutoScalingVersion(version string) AutoscalingVersion {
```
pavolloffay, line 120: any reason why the function is public now?

kevinearls: So it can be accessed by the test, as the test is in a different package

pavolloffay: It's a bit unfortunate/unnecessary to have the test in a different package in this case. But it was done like that from before so I will approve the PR.

pavolloffay (approved):

Issue #942, opened 2022-06-23
Create a test for horizontal pod autoscaling


kevinearls:
Write a test for the changes for #801   This will most likely require changes to the default HPA settings so that it does not take a 5-10 minutes to downscale.  

See https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/ for reference.

pavolloffay:
Do you mean e2e test?

kevinearls:
@pavolloffay Yes

yuriolisa:
@pavolloffay, I believe this issue can be closed.

PR #1077, opened 2022-09-06
Expose Horizontal Pod Autoscaler Behavior and add hpa scaledown test

kevinearls:


apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -40,6 +40,13 @@ type OpenTelemetryCollectorSpec struct {
 	// MaxReplicas sets an upper bound to the autoscaling feature. If MaxReplicas is set autoscaling is enabled.
 	// +optional
 	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
+
+	// Autoscaler specifies the pod autoscaling configuration to use
+	// for the OpenTelemetryCollector workload.
+	//
+	// +optional
+	Autoscaler *AutoscalerSpec `json:"autoscaler,omitempty"`
```
jaronoff97, line 48: question: did we consider embedding the HPA spec in here? Or at least embed the [autoscaling behavior spec](https://pkg.go.dev/k8s.io/api/autoscaling/v2beta2#HorizontalPodAutoscalerBehavior) here?

apis/v1alpha1/opentelemetrycollector_webhook.go
```diff
@@ -129,6 +129,15 @@ func (r *OpenTelemetryCollector) validateCRDSpec() error {
 			return fmt.Errorf("the OpenTelemetry Spec autoscale configuration is incorrect, minReplicas should be one or more")
 		}
 
+		if r.Spec.Autoscaler != nil {
```
jaronoff97, line 132: We should also check that min or max replicas is set in order to use this feature

iblancasa (approved):

apis/v1alpha1/opentelemetrycollector_webhook.go
kevinearls: That's inside an if statement that's checking whether maxReplicas is set.

apis/v1alpha1/opentelemetrycollector_types.go
kevinearls: I was just adding enough code to be able to get an e2e test to work within the time allocated, which means we need to scale down much quicker that the default 300 seconds.

@pavolloffay what do you think?  Do I need to add the other values of PA scaling rules here?

tests/e2e/autoscale/00-assert.yaml
```diff
@@ -13,4 +13,7 @@ metadata:
 spec:
   minReplicas: 1
   maxReplicas: 2
+# This is not neccesarily exact.  We really just want to wait until this is no longer <unknown>
```
iblancasa, line 16: This is a really good "hack"

iblancasa (approved):

pavolloffay:
Does this resolve #942 ?

kevinearls:
@pavolloffay Yes

apis/v1alpha1/opentelemetrycollector_types.go
jaronoff97: For my use case, I know that I would like to be able to specify policies and not just `StabilizationWindowSeconds`. Embedding only `StabilizationWindowSeconds` is going to make it so we need to add in each feature on request, making a code change for each one.

pavolloffay (approved):

PR #1078, opened 2022-09-06
Upgrade kind images and add testing for Kubernetes 1.25

iblancasa:

This PR:
* Upgrades the `kind` container images to the latest
* Add testing for Kubernetes 1.25

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -17,7 +17,7 @@ jobs:
        # should be compatible with them.
        kube-version:
        - "1.19"
-       - "1.24"
+       - "1.25"
 
     steps:
 
--- a/.github/workflows/scorecard.yaml
+++ b/.github/workflows/scorecard.yaml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         kube-version:
           - "1.19"
-          - "1.24"
+          - "1.25"
 
     steps:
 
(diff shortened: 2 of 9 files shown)


.github/workflows/e2e.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
        # should be compatible with them.
        kube-version:
        - "1.19"
-       - "1.24"
+       - "1.25"
```
jaronoff97, line 20: Question: why do we just use `1.19` and `1.25`? I see there's a 1.19 through 1.25

iblancasa: I can change it if you want. I don't have a strong opinion related to this

kevinearls: We started doing it on the jaeger-operator because of the time it took to run builds on all supported kubernetes versions.  PRs were backing up for a long time because of this.

We could try running against all supported versions here, but we may run into the same problem.

iblancasa: Oh, sorry. I misunderstood the @jaronoff97 comment.
As @kevinearls says, this is done to reduce the time spent in the E2E test executed by the CI. There is a comment in this file about this decision:
```yaml
       # The e2e tests are run on the lowest and highest supported k8s version.
       # All Kubernetes version in between expose the same APIs, hence the operator
       # should be compatible with them.
```

kevinearls (approved):

.github/workflows/e2e.yaml
jaronoff97: Ah thank you for the clarification!

jaronoff97 (approved):

iblancasa:
Is anything else pending for this PR?

PR #1079, opened 2022-09-07
Introduce ability to specify strategies for target allocation

jaronoff97:
Following from #1068 ... 

## Goals:
* Introduce a new interface to allow a user of the TA to describe an allocation strategy
  * Allocation strategies are given the changes to the collectors and the target items
  * Allocation strategies can decide how they want to manage state
* Refactor the allocator's current strategy - least weighted – to be the first strategy

## Summary
This PR introduces the Allocator interface, implements it through a refactor of the current allocator, and changes the configuration of the TA to allow for a user to specify an option. The operator's CRDs are also updated to allow the strategy to be specified in the Collector CRD. New tests have been introduced to confirm the strategy works as expected. All previous tests should continue working.


cmd/otel-allocator/allocation/least_weighted/allocator.go
```diff
@@ -0,0 +1,210 @@
...
+	The targets are allocated using the least connection method
+	Load balancer will need information about the collectors in order to set the URLs
+	Keep a Map of what each collector currently holds and update it based on new scrape target updates
+*/
+
+// LeastWeightedAllocator makes decisions to distribute work among
+// a number of OpenTelemetry collectors based on the number of targets.
+// Users need to call SetTargets when they have new targets in their
+// clusters and call SetCollectors when the collectors have changed.
+type LeastWeightedAllocator struct {
```
kristinapathak, line 40: I don't think we need to export this.

cmd/otel-allocator/allocation/least_weighted/allocator.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+	}
+	return col
+}
+
+// addTargetToTargetItems assigns a target to the next available collector and adds it to the allocator's targetItems
+// This method is called from within SetTargets and SetCollectors, whose caller acquires the needed lock.
+// This is only called after the collectors are cleared or when a new target has been found in the tempTargetMap
+func (allocator *LeastWeightedAllocator) addTargetToTargetItems(target *strategy.TargetItem) {
+	chosenCollector := allocator.findNextCollector()
```
kristinapathak, line 92: There is a possibility that the target allocator starts up without any collectors and the collector returned here could be nil.  I don't think we have to fix it in this pr, but I'm inclined to make an issue about better handling of the case of 0 collectors.

cmd/otel-allocator/allocation/least_weighted/allocator.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	// Insert the new collectors
+	for _, i := range diff.Additions() {
+		allocator.state.SetCollector(i.Name, &strategy.Collector{Name: i.Name, NumTargets: 0})
+	}
+
+	// find targets which need to be redistributed
+	var redistribute []*strategy.TargetItem
+	for _, item := range allocator.state.TargetItems() {
+		for _, s := range diff.Removals() {
```
kristinapathak, line 143: Instead of cycling through all the removed collectors, can we check if the collector is still in the collectors map? If not, the target needs to be redistributed.

cmd/otel-allocator/allocation/least_weighted/allocator.go
```diff
@@ -0,0 +1,210 @@
...
+
+func (allocator *LeastWeightedAllocator) handleCollectors(diff utility.Changes[*strategy.Collector]) {
+	// Clear existing collectors
+	for _, k := range diff.Removals() {
+		allocator.state.RemoveCollector(k.Name)
+		strategy.TargetsPerCollector.WithLabelValues(k.Name).Set(0)
+	}
+	// Insert the new collectors
+	for _, i := range diff.Additions() {
+		allocator.state.SetCollector(i.Name, &strategy.Collector{Name: i.Name, NumTargets: 0})
```
kristinapathak, line 137: We can use `NewCollector()` here.

cmd/otel-allocator/allocation/http.go
```diff
@@ -4,55 +4,43 @@ import (
...
-func GetAllTargetsByCollectorAndJob(collector string, job string, cMap map[string][]TargetItem, allocator *Allocator) []targetGroupJSON {
-	var tgs []targetGroupJSON
+func GetAllTargetsByCollectorAndJob(collector string, job string, cMap map[string][]strategy.TargetItem, allocator strategy.Allocator) []strategy.TargetGroupJSON {
+	var tgs []strategy.TargetGroupJSON
 	group := make(map[string]string)
 	labelSet := make(map[string]model.LabelSet)
-	for _, col := range allocator.Collectors() {
-		if col.Name == collector {
+	for colName, _ := range allocator.Collectors() {
+		if colName == collector {
```
kristinapathak, line 40: We should be able to get the collector by directly accessing `allocator.Collectors()[collector]` without  cycling through all map key values.

cmd/otel-allocator/allocation/least_weighted/allocator.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	// find targets which need to be redistributed
+	var redistribute []*strategy.TargetItem
+	for _, item := range allocator.state.TargetItems() {
+		for _, s := range diff.Removals() {
+			if item.CollectorName == s.Name {
+				redistribute = append(redistribute, item)
```
kristinapathak, line 145: Instead of making a slice of targets to redistribute, I think we could just call `allocator.addTargetToTargetItems(item)` here?

cmd/otel-allocator/allocation/least_weighted/allocator.go
jaronoff97: Definitely agreed, this was a problem before and is a case we need to handle more effectively. I don't believe we are able to access this code block with 0 collectors however, because we short circuit in the `HandleCollectors()` method

cmd/otel-allocator/allocation/least_weighted/allocator.go
```diff
@@ -0,0 +1,210 @@
...
+	return &LeastWeightedAllocator{
+		log:   log,
+		state: strategy.NewState(make(map[string]*strategy.Collector), make(map[string]*strategy.TargetItem)),
+	}
+}
+
+func init() {
+	err := strategy.Register("least-weighted", NewAllocator)
+	if err != nil {
+		os.Exit(1)
```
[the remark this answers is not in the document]
secustor: Why not return an error and log this using `setupLog`?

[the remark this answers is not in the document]
jaronoff97: I don't think we have access to setupLog at this point.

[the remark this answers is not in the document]
jaronoff97: because this is in init, we can't return to main and log it with `setupLog`

cmd/otel-allocator/allocation/strategy/state.go
```diff
@@ -0,0 +1,91 @@
+package strategy
```
kristinapathak, line 1: I think we should move the code in the `strategy` package back to `allocation`. `CollectorJSON` is only used in `allocation/http.go` and the other JSON structs are also around specifically for the http responses. It would also make naming make more sense (`allocation.New()` instead of `strategy.New()`).  It feels like we're splitting up code that we don't really need to split up.  I think that separating the code in this way makes it a little more difficult to navigate and figure out what's going on.

jaronoff97: I agree, i think a flat strategy here makes more sense for this codebase as it allows us to only export what's necessary

cmd/otel-allocator/allocation/http.go
```diff
@@ -11,58 +11,55 @@ type LinkJSON struct {
 	Link string `json:"_link"`
 }
 
-type collectorJSON struct {
+type CollectorJSON struct {
```
Aneurysm9, line 14: Do these types need to be exported?

cmd/otel-allocator/allocation/least_weighted.go
```diff
@@ -0,0 +1,195 @@
...
+
+	"github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/diff"
+
+	"github.com/go-logr/logr"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	_            Allocator = &leastWeightedAllocator{}
+	strategyName           = "least-weighted"
```
Aneurysm9, line 16: Should this be `const`?

cmd/otel-allocator/allocation/least_weighted.go
```diff
@@ -0,0 +1,195 @@
...
+		}
+	}
+	return col
+}
+
+// addTargetToTargetItems assigns a target to the next available collector and adds it to the allocator's targetItems
+// This method is called from within SetTargets and SetCollectors, which acquire the needed lock.
+// This is only called after the collectors are cleared or when a new target has been found in the tempTargetMap
+func (allocator *leastWeightedAllocator) addTargetToTargetItems(target *TargetItem) {
+	chosenCollector := allocator.findNextCollector()
```
Aneurysm9, line 83: This can be nil if `len(allocator.collectors) == 0`.  Validating and short-circuiting in this case is probably necessary to avoid nil pointer errors.

jaronoff97: Kristina brought up a similar concern for the `findNextCollector()`, this issue existed before and was covered by the fact that no collectors isn't currently a valid state. For right now, i think we need to introduce the same short circuit to verify the invariant for `SetTargets` instead of introducing error handling here. After this is merged, I'll open up an issue to solve this in a better way.

Aneurysm9 (approved):

cmd/otel-allocator/watcher/main.go
```diff
@@ -2,16 +2,17 @@ package watcher
 
 import (
 	"fmt"
+	"github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/allocation"
```
kristinapathak, line 5: I think this import is in the wrong place.  This is the first file I noticed it, but there's a similar issue in other files as well.

cmd/otel-allocator/diff/diff.go
```diff
@@ -0,0 +1,36 @@
+package diff
+
+type Changes[T any] struct {
```
kristinapathak, line 3: Should we include comments for the structs/functions in this file?

cmd/otel-allocator/allocation/least_weighted.go
```diff
@@ -0,0 +1,200 @@
...
+			col = v
+		}
+	}
+	return col
+}
+
+// addTargetToTargetItems assigns a target to the next available collector and adds it to the allocator's targetItems
+// This method is called from within SetTargets and SetCollectors, which acquire the needed lock.
+// This is only called after the collectors are cleared or when a new target has been found in the tempTargetMap
+// INVARIANT: allocator.collectors must have at least 1 collector set
```
kristinapathak, line 82: nice 🙂

cmd/otel-allocator/allocation/least_weighted.go
```diff
@@ -0,0 +1,200 @@
...
+		if _, ok := allocator.targetItems[k]; ok {
+			continue
+		} else {
+			// Add target to target pool and assign a collector
+			allocator.addTargetToTargetItems(target)
+		}
+	}
+}
+
+func (allocator *leastWeightedAllocator) handleCollectors(diff diff.Changes[*Collector]) {
```
kristinapathak, line 121: Other than `NewAllocator`, this is the only function without a comment explaining what it does.  I think it's pretty clear what's happening but not sure if you want to add an explanation just for completeness.

kristinapathak (approved):
Looks great! I left a few minor comments relating to imports and comments.

cmd/otel-allocator/collector/collector.go
```diff
@@ -2,6 +2,7 @@ package collector
 
 import (
 	"context"
+	"github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/allocation"
```
secustor, line 5: please, move this its own group

cmd/otel-allocator/discovery/discovery.go
```diff
@@ -2,16 +2,17 @@ package discovery
 
 import (
 	"context"
+	"github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/allocation"
```
secustor, line 5: This should be in the third group

cmd/otel-allocator/discovery/discovery_test.go
```diff
@@ -3,19 +3,20 @@ package discovery
 import (
 	"context"
 	"fmt"
+	"github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/allocation"
```
secustor, line 6: This should be in the third group

cmd/otel-allocator/main.go
```diff
@@ -9,18 +9,21 @@ import (
 	"os/signal"
 	"syscall"
 
+	"github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/allocation"
```
secustor, line 12: This should be in the third group

secustor (changes requested):

cmd/otel-allocator/main.go
jaronoff97: Gah! My goimports didn't work :( i'll do this manually

cmd/otel-allocator/allocation/least_weighted.go
jaronoff97: agreed, good idea to comment

pavolloffay (approved):

PR #1081, opened 2022-09-08
.NET - enable bytecode instrumentation

Kielek:
Enables bytecode instrumentations. All of them are enabled by default.
For v.0.3.1-beta.1 it extends set of supported libraries: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/v0.3.1-beta.1/docs/config.md#instrumented-traces-libraries-and-frameworks


Kielek:
@open-telemetry/dotnet-instrumentation-approvers FYI

Kielek:
@avadhut123pisal, FYI

lachmatt (approved):

pkg/instrumentation/dotnet.go
```diff
@@ -116,3 +112,25 @@ func injectDotNetSDK(logger logr.Logger, dotNetSpec v1alpha1.DotNet, pod corev1.
...
+		container.Env = append(container.Env, corev1.EnvVar{
+			Name:  envVarName,
+			Value: envVarValue,
+		})
+		return true
+	}
+
+	if container.Env[idx].ValueFrom != nil {
+		// TODO add to status object or submit it as an event
+		logger.Info("Skipping DotNet SDK injection, the container defines env var value via ValueFrom", "envVar", envVarName, "container", container.Name)
```
pellared, line 128: I think this behavior and the log message are misleading.

When `false` is returned then it interrupts processing next env vars.
It could leave the container in a "semi-instrumented" state. Is it expected?

This comment applies also to the code BEFORE the changes introduced in this PR. Therefore, it would be better to make a separate PR to address it.

pkg/instrumentation/dotnet.go
```diff
@@ -116,3 +112,25 @@ func injectDotNetSDK(logger logr.Logger, dotNetSpec v1alpha1.DotNet, pod corev1.
...
+		return false
+	}
+	if concatValues {
```
pellared, line 131: nit: add a blank line for consistency (and easier reading)

pellared (approved):

pkg/instrumentation/dotnet.go
Kielek: Fixed: 10a7e986e51fde03ae59a29e71da8fb6e25eb6d3

nrcventura (approved):

rajkumar-rangaraj (approved):

pavolloffay (approved):

pellared:
@Kielek I think it MAY be good to offer an easy way to disable byte-code instrumentation (disable attaching the .NET Profiler) as this can have negative consequences on the instrumented application's performance (maybe it was not enabled by default on purpose?)

Kielek:
> @Kielek I think it MAY be good to offer an easy way to disable byte-code instrumentation (disable attaching the .NET Profiler) as this can have negative consequences on the instrumented application's performance (maybe it was not enabled by default on purpose?)

My perspective is to offer as broad auto-instrumentation as possible.
It can be disabled by the user by setting `CORECLR_ENABLE_PROFILING` to `0`.

PR #1087, opened 2022-09-12
Added consistent hashing strategy

### LINKED ISSUE #1061: Implement leader election for the target allocator
Right now, the target allocator's allocation strategy (least connection) means that you can only run a single TA pod at a time. This proves difficult if a consumer wants a high availability option for their target allocation. In order to make this possible, we could use the built in go [leader election package](https://pkg.go.dev/k8s.io/client-go/tools/leaderelection) for the target allocator.

The rough process that a collector does looks like this right now:

1. collector starts up
2. prom receiver loads configuration
3. for each job
4. http_sd_config queries target allocator
5. list of targets and metadata is returned from TA
6. prom receiver runs relabel_configs on targets
7. prom receiver scrapes targets remaining
8. prom receiver applies metric_relabel_configs
9. collector converts prom to otel
10. moves config to processor stage ...

If the target allocator is down at step 4, the job will fail (and most likely the entire the scrape config.) Adding in support for HA would improve the reliability of the statefulset collector.

If this is something the community would like, I would be happy to implement and test it.

jaronoff97:
This PR adds the consistent hashing strategy. In doing so, I included tests to verify some of the expected behaviors of this strategy method. This PR is one solution to close #1061 as two target allocator replicas will be able to run simultaneously with this strategy because they will always provide the same answer.

```
spec:
  targetAllocator:
    allocationStrategy: "consistent-hashing"
    replicas: 2
    enabled: true
```

Screen Shot 2022-09-14 at 11 17 14 AM


cmd/otel-allocator/allocation/consistent_hashing.go
```diff
@@ -0,0 +1,189 @@
...
+
+	log logr.Logger
+}
+
+func newConsistentHashingAllocator(log logr.Logger) Allocator {
+	config := consistent.Config{
+		PartitionCount:    1061,
+		ReplicationFactor: 5,
+		Load:              1.1,
+		Hasher:            hasher{},
```
jaronoff97, line 44: question: should this be configurable

cmd/otel-allocator/allocation/consistent_hashing.go
```diff
@@ -0,0 +1,189 @@
...
+
+const consistentHashingStrategyName = "consistent-hashing"
+
+type hasher struct{}
+
+func (h hasher) Sum64(data []byte) uint64 {
+	return xxhash.Sum64(data)
+}
+
+type consistentHashingAllocator struct {
```
jaronoff97, line 24: A lot of this code looks _very_ similar to the least weighted strategy. It may be worth investigating in the future if we can combine these as I tried to do in #1068

cmd/otel-allocator/allocation/consistent_hashing.go
```diff
@@ -0,0 +1,189 @@
...
+		} else {
+			// Add target to target pool and assign a collector
+			c.addTargetToTargetItems(target)
+		}
+	}
+}
+
+// handleCollectors receives the new and removed collectors and reconciles the current state.
+// Any removals are removed from the allocator's collectors. New collectors are added to the allocator's collector map
+// Finally, any targets of removed collectors are reallocated to the next available collector.
```
kristinapathak, line 102:
```suggestion
// Finally, update all targets' collectors to match the consistent hashing.
```

kristinapathak (commented):
Actual logic looks great! It is extremely similar to the least weighted algorithm...

cmd/otel-allocator/allocation/consistent_hashing.go
```diff
@@ -0,0 +1,189 @@
...
+	colOwner := c.hasher.LocateKey([]byte(target.Hash()))
+	targetItem := &TargetItem{
+		JobName:       target.JobName,
+		Link:          LinkJSON{Link: fmt.Sprintf("/jobs/%s/targets", url.QueryEscape(target.JobName))},
+		TargetURL:     target.TargetURL,
+		Label:         target.Label,
+		CollectorName: colOwner.String(),
+	}
+	c.targetItems[targetItem.Hash()] = targetItem
+	c.collectors[colOwner.String()].NumTargets++
```
kristinapathak, line 69: Because we might be reallocating all targets, we should check if the targetItem already exists and, if so, decrement its collector's `NumTargets` count (and set that metric too).

jaronoff97: Ah good point! We may be getting occasionally saved currently by the fact we set it equal to the numTargets so as we assign it with least weighted we end up just setting everything correctly. This is a good call out though, ultimately this metric should be done with an async gauge rather than calling it on every method call. lemme think about the best way to fix this one

cmd/otel-allocator/allocation/consistent_hashing.go
```diff
@@ -0,0 +1,194 @@
...
+			TargetsPerCollector.WithLabelValues(target.CollectorName, consistentHashingStrategyName).Set(float64(col.NumTargets))
+		}
+	}
+
+	// Check for additions
+	for k, target := range diff.Additions() {
+		// Do nothing if the item is already there
+		if _, ok := c.targetItems[k]; ok {
+			continue
+		} else {
```
Aneurysm9, line 98: Is this necessary?  `addTargetToTargetItems()` has provision for when the new target has an existing collector attached to it, which I think would only happen when the target existed previously as well.

cmd/otel-allocator/allocation/consistent_hashing.go
```diff
@@ -0,0 +1,194 @@
+package allocation
+
+import (
+	"fmt"
+	"github.com/buraksezer/consistent"
+	"github.com/cespare/xxhash/v2"
+	"github.com/go-logr/logr"
+	"github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/diff"
```
pavolloffay, line 8: import order - probably lint is not running correctly

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -108,6 +108,10 @@ type OpenTelemetryCollectorSpec struct {
 
 // OpenTelemetryTargetAllocator defines the configurations for the Prometheus target allocator.
 type OpenTelemetryTargetAllocator struct {
+	// Replicas is the number of pod instances for the underlying TargetAllocator, this can only be set to values other
+	// than 1 if a strategy that allows for high availability is chosen.
```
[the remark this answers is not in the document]
jaronoff97: This is only advisory, the logic for enforcing may be simple now, but will be more complicated in the future when we have things like leader election or maybe even another method like distributed state. I also was thinking if someone wanted to plug in their own TA that the operator would only know the image of, we wouldn't have a way of changing the validated webhook to know to accept the custom TA strategy.

cmd/otel-allocator/allocation/consistent_hashing.go
```diff
@@ -0,0 +1,194 @@
...
+	targetItems map[string]*TargetItem
+
+	log logr.Logger
+}
+
+func newConsistentHashingAllocator(log logr.Logger) Allocator {
+	config := consistent.Config{
+		PartitionCount:    1061,
+		ReplicationFactor: 5,
+		Load:              1.1,
```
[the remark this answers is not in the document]
jaronoff97: I'm not sure... I think i'm going to open a follow up issue to make these configurable as it would require some refactoring of how configuration is passed down to the allocation strategies as right now it's only the string. Is that alright?

cmd/otel-allocator/allocation/consistent_hashing.go
jaronoff97: I believe this is still necessary. the provision in `addTargetToTargetItems()` is for when we reassign a collector after a new collector is added which shifts the amount of members in the hashring. In this case, we are iterating through the newly added targets and seeing if the target has already been added. Because we're using consistent hashing, the target would get unassigned and then reassigned to the same collector which seems like unnecessary work.

config/crd/bases/opentelemetry.io_opentelemetrycollectors.yaml
```diff
@@ -807,7 +807,9 @@ spec:
                 properties:
                   allocationStrategy:
                     description: AllocationStrategy determines which strategy the
-                      target allocator should use for allocation
+                      target allocator should use for allocation The current options
```
kristinapathak, line 810:
```suggestion
                      target allocator should use for allocation. The current options
```

config/crd/bases/opentelemetry.io_opentelemetrycollectors.yaml
```diff
@@ -829,6 +831,14 @@ spec:
                           custom resources as targets or not.
                         type: boolean
                     type: object
+                  replicas:
+                    description: Replicas is the number of pod instances for the underlying
+                      TargetAllocator, this should only be set to a value other than
```
kristinapathak, line 836:
```suggestion
                      TargetAllocator. This should only be set to a value other than
```

kristinapathak (approved):
looks great! a few grammar suggestions.

cmd/otel-allocator/allocation/consistent_hashing.go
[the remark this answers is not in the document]
Aneurysm9: Yeah, that's fine.  I think using the string for configuration might be fine, as long as it's well specified.  Something like URL query params style might work:

```
allocationStrategy: "consistentHashing?load=1.5&partitionCount=3000"
```

That would make it a straightforward addition while leaving plenty of flexibility.  Don't need to solve now though, can figure it out on the follow up issue.

Aneurysm9 (approved):

pavolloffay (commented):
approving based on the approvals from TA contributors.

pavolloffay (approved):

PR #1090, opened 2022-09-13
Fix lint issues in target allocator

kristinapathak:
I noticed that the target allocator files don't have a copyright at the top.  Given how uniform it was, I'm not sure if this was intentional, but I added them here in case they are missing.


Aneurysm9 (approved):
Should this repo include a `license-check` target that ensures these are present?  We do this [in the Go SDK](https://github.com/open-telemetry/opentelemetry-go/blob/main/Makefile#L179-L187).

codeboten (approved):

jaronoff97 (approved):

secustor (approved):

kristinapathak:
@Aneurysm9, I added the target you recommended - it runs as a part of basic checks:
https://github.com/open-telemetry/opentelemetry-operator/actions/runs/3046982755/jobs/4910456529

.ci/create-release-github.sh
```diff
@@ -1,5 +1,19 @@
 #!/bin/bash
 
+# Copyright The OpenTelemetry Authors
```
pavolloffay, line 3: The header is already defined in `header.txt`. The operator already uses golang-ci to verify the license header is present on the files. 

Could we configure the golang-ci to check the target allocator as well? It seems that the golang-ci is not checking the target allocator, we should configure it to check the whole repo.

pavolloffay (changes requested):

.ci/create-release-github.sh
kristinapathak: I will work on fixing golang-ci to check the target allocator files instead of using the make target.  What about `.sh` files - should we keep the target for those?

pavolloffay: I am with not adding the headers to the .sh files. Those files are only to support CI or local development.

kristinapathak:
Golang-ci is not running on target allocator code.  I will be working on getting golang-ci to run on all go modules in the repo and fixing other lint issues in the target allocator.

kristinapathak:
@pavolloffay, how was the configuration in .golangci.yaml decided?  Other opentelemetry repos have govet configured in a different way than this one, most notably for `govet`:
https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/29a46c89ab69347be7d6f2c7f4f94135cb4d0d4f/.golangci.yml#L58
https://github.com/open-telemetry/opentelemetry-collector/blob/b1329c94b3b31b9ffa4091ab5066d6f091d71f41/.golangci.yml#L58
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/d2490af3003387cc4b1baa5b55f7d748303d7a4b/.golangci.yml#L56

There are many `govet fieldalignment` issues that look like this:
```
allocation/consistent_hashing.go:40:33: fieldalignment: struct with 64 pointer bytes could be 40 (govet)
type consistentHashingAllocator struct {
```

I'm not sure the best path forward to fix these specific fieldalignment issues.

pavolloffay:
> @pavolloffay, how was the configuration in .golangci.yaml decided? 

I am not sure, probably back then I was not working in this repo. You could browse the PR history. 

I am fine with disabling `fieldalignment` and aligning the golang-ci to other projects in the organization (if they use the same config).

cmd/otel-allocator/.golangci.yaml
```diff
@@ -0,0 +1,28 @@
+run:
```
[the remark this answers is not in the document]
kristinapathak: I think we need it for the different path at `goimports.local-prefixes`, but I'll try to look into other options to set this config value specifically.

kristinapathak:
@pavolloffay, bringing the golangci.yaml into alignment with the golangci configs in opentelemetry-collector and opentelemetry-collector-contrib means even more lint fixes.  I would rather disable `fieldalignment` in this PR, merge this, and then work on the next lint changes in a new PR.  Is that okay with you?

pavolloffay:
>  Is that okay with you? 

Absoulutelly. 

I am not sure why CI job `Code standards (lining)`  didn't run

pavolloffay:
maybe try to squash the commits and rebased the PR

kristinapathak:
Further lint work: https://github.com/open-telemetry/opentelemetry-operator/issues/1122

kristinapathak:
@pavolloffay, I think there is an issue from `Code standards (linting)` being in a matrix now:
```
Continuous Integration / Code standards (linting) (.) (pull_request) 
Continuous Integration / Code standards (linting) (./cmd/otel-allocator) (pull_request) 
```

cmd/otel-allocator/.golangci.yaml
```diff
@@ -0,0 +1,26 @@
...
```
bogdandrutu, line 1: It is bad to duplicate this, if the action does not know to read the config from parent dirs, as the binary does, consider to manually run the binary as we do in all the other repos.

cmd/otel-allocator/.golangci.yaml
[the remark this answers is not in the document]
kristinapathak: I've removed this file.

cmd/otel-allocator/.golangci.yaml
kristinapathak: Okay - I've removed this file.

cmd/otel-allocator/header.txt
```diff
@@ -0,0 +1,13 @@
+Copyright The OpenTelemetry Authors
```
pavolloffay, line 1: could we remove this as well and reuse the header defined in the project root?

pavolloffay (approved):

PR #1114, opened 2022-09-20
Add default settings to README

timbuchwaldt:
As many probably copy the first example in the README, this should contain the suggested two default processors. Running without them is a bad idea, therefore I suggest this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,13 @@ spec:
           grpc:
           http:
     processors:
+      memory_limiter:
+        check_interval: 1s
+        limit_mib: 500
+        spike_limit_mib: 100
+      batch:
+        send_batch_size: 1000
+        timeout: 2s
 
     exporters:
       logging:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,6 +35,13 @@ spec:
           grpc:
           http:
     processors:
+      memory_limiter:
+        check_interval: 1s
+        limit_mib: 500
```
pavolloffay, line 40: In k8s it makes more sense to do percentage settings.

README.md
```diff
@@ -35,6 +35,13 @@ spec:
...
+        spike_limit_mib: 100
+      batch:
```
pavolloffay, line 42: Why 1000 and 2s? The default values are different https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/README.md

timbuchwaldt:
Hey @pavolloffay - thx for the feedback, changed accordingly.

pavolloffay (approved):

PR #1116, opened 2022-09-21
Update routine for migration of jaeger remote sampling in version 0.61.0

### LINKED ISSUE #1104: Add upgrade routine for jaeger remote sampling from 0.60 to 0.61.
In version 0.61.0 the [jaeger remote sampling functionality is removed](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/14163) from the receiver. It will be available as an extra extension.

frzifus:
Closes #1104


frzifus:
I have created a first draft. A quick question, does it make sense to import the collector packages to port the configurations or more to avoid the dependency? I assume avoiding is better here?

cc @kevinearls

pkg/collector/upgrade/v0_61_0.go
```diff
@@ -0,0 +1,142 @@
...
+		return otelcol, nil
+	}
+
+	otelCfg, err := adapters.ConfigFromString(otelcol.Spec.Config)
+	if err != nil {
+		return otelcol, fmt.Errorf("couldn't upgrade to v0.61.0, failed to parse configuration: %w", err)
+	}
+
+	// Search for removed Jaeger remote sampling settings. (https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/14163)
+	receiversConfig, ok := otelCfg["receivers"].(map[string]interface{})
```
frzifus, line 40: Any idea why we see here:
```golang
map[receivers:map[jaeger:map[protocols:map[grpc:<nil>] remote_sampling:map[endpoint:jaeger-collector:14250 strategy_file:/etc/strategy.json strategy_file_reload_interval:10s tls:map[insecure:true]]] jaeger/2:map[protocols:map[grpc:<nil>] remote_sampling:map[endpoint:jaeger-collector2:14250 strategy_file:/etc/strategy2.json strategy_file_reload_interval:15s tls:map[insecure:true]]]] service:map[pipelines:map[traces:map[exporters:[nop] receivers:[jaeger jaeger/2]]]]]
```

The returned data doesnt match `map[string]interface{}` since the key is different: `jaeger:map[protocols:map[grpc:<nil>`.

Here is the yaml that is provided:
```yaml
---
receivers:
  jaeger:
    protocols:
      grpc:
    remote_sampling:
      endpoint: "jaeger-collector:14250"
      tls:
        insecure: true
      strategy_file: "/etc/strategy.json"
      strategy_file_reload_interval: 10s
  jaeger/2:
    protocols:
      grpc:
    remote_sampling:
      endpoint: "jaeger-collector2:14250"
      tls:
        insecure: true
      strategy_file: "/etc/strategy2.json"
      strategy_file_reload_interval: 15s
service:
  pipelines:
    traces:
      receivers: [jaeger, jaeger/2]
      exporters: [nop]
```

pavolloffay:
> I have created a first draft. A quick question, does it make sense to import the collector packages to port the configurations or more to avoid the dependency? I assume avoiding is better here?

IIRC right now we don't import OTEL col config in the operator.

pkg/collector/upgrade/v0_61_0_test.go
```diff
@@ -0,0 +1,111 @@
...
+extensions:
+  jaegerremotesampling:
+    source:
+      remote:
+        endpoint: jaeger-collector:14250
+      tls:
+        insecure: true
+      reload_interval: 10s
+      file: /etc/strategy.json
+  jaegerremotesampling:
```
pavolloffay, line 91: Why is the extension `jaegerremotesampling` defined twice? Shouldn't the second one have name e.g. `/2`?

pkg/collector/upgrade/v0_61_0_test.go
```diff
@@ -0,0 +1,111 @@
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "otel-my-instance",
+			Namespace: "somewhere",
+		},
+		Spec: v1alpha1.OpenTelemetryCollectorSpec{
+			Config: `---
+receivers:
+  jaeger:
+    protocols:
```
pavolloffay, line 44: Can be jaeger receiver defined only with the remote sampling? If that holds then we should make sure that the port opened by the extension is exposed.

frzifus: Any suggestion how to test that?

frzifus:
i wonder if it wouldn't make more sense to let the upgrade fail if remote sampling is enabled and then refer the user to the OTEL documentation? The problem I see is that we can never be sure that the specified port is correct. Furthermore, it could become a problem if external collectors or similar are not reconfigured.

@pavolloffay @jpkrohling @yuriolisa

frzifus:
See: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/14707

kevinearls (approved):

jpkrohling (approved):
LGTM. Could you also raise an issue with the collector so that we can discuss using one port for multiple receivers? It would allow the jaeger receiver and the remote sampling endpoints to be served from the same port again. The way it is now, the jaeger collector isn't interchangeable with the otel collector.

frzifus:
@jpkrohling another then https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/14707?

pkg/collector/upgrade/v0_61_0.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+
+		// check if remote sampling settings exit
+		if _, ok := cfg["remote_sampling"]; !ok {
+			continue
+		}
+
+		const issueID = "https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/14707"
+		return nil, fmt.Errorf(
+			"jaegerremotesampling is no longer available as receiver configuration. "+
```
pavolloffay, line 60: I like that the message points to the GH issue, but the root cause is not obvious. I would suggest putting either here a sentence that the ports should be changed for receiver and extension or make it super clear in the GH issue.

pavolloffay: Second question:

This is the first time when we actually expect the upgrade to fail. How is this error propagated to the end-user?  It would be good to have this in the operator logs but as well recoded as k8s evet. 

Third question: What happens if the upgrade fails? The instance version in the status will not be upgraded/the image will not be upgraded? If the user fixes the CR by splitting the ports will the upgrade kick in again?

pavolloffay (changes requested):
I have some questions about the upgrade

pkg/collector/upgrade/v0_61_0.go
frzifus: > make it super clear in the GH issue

i tried to make it clear by saying: `The jaeger remote sampling extension collides when converting old jaeger receiver settings. The problem can be solved by manually switching to another port, but this leads to potentially invalid configurations for reporters.` What would you recommend instead? Or what error message do you have in mind?

> This is the first time when we actually expect the upgrade to fail. How is this error propagated to the end-user? It would be good to have this in the operator logs but as well recoded as k8s evet.

Good point. All actions are simply [ignored](https://github.com/open-telemetry/opentelemetry-operator/blob/v0.60.0/pkg/collector/upgrade/upgrade.go#L64). Right now there is only a single [log entry](https://github.com/open-telemetry/opentelemetry-operator/blob/v0.60.0/pkg/collector/upgrade/upgrade.go#L118).

> What happens if the upgrade fails? The instance version in the status will not be upgraded/the image will not be upgraded? If the user fixes the CR by splitting the ports will the upgrade kick in again?

Your CR status will not change. If you fix this issue manually, its all fine. There is no other upgrade routine for `0.61.0`. Everytime you restart or deploy a newer operator version, all upgrade routines are triggered again.

pavolloffay: > Your CR status will not change. If you fix this issue manually, its all fine. There is no other upgrade routine for 0.61.0. Everytime you restart or deploy a newer operator version, all upgrade routines are triggered again.

After the fix is done, will the operator upgrade the instance by bumping the OTELcol version?

frzifus: I need to double check that. But i assume when you modified your CR, yes.

pavolloffay: Please report back if the upgrade was tested when Jaeger receiver was used with enabled remote sampling. We need to make sure the instance will be upgraded after the CR is fixed (e.g. extension mapped to a different port)

frzifus: @pavolloffay i have now been able to test it locally. after a configuration update, the version is **not** automatically raised. To achieve this, either the operator must be restarted or the collector configuration must be deleted and recreated.

<details>
  <summary>Details...</summary>

Operator log message, once the config is detected.
```bash
1.665433033457368e+09	ERROR	collector-upgrade	failed to upgrade managed otelcol instances	{"name": "simplest", "namespace": "test", "error": "jaegerremotesampling is no longer available as receiver configuration. Please use the extension instead with a different remote sampling port. See: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/14707"}
```

procedure:
```
# status
$ kubectl get opentelemetrycollectors.opentelemetry.io -n test simplest        
NAME       MODE         VERSION   AGE
simplest   deployment   0.60.0    11m

# fix config
$ kubectl apply -n test -f v0_60_0_jaeger_cfg.yaml                      
opentelemetrycollector.opentelemetry.io/simplest configured

# status again
$ kubectl get opentelemetrycollectors.opentelemetry.io -n test simplest      
NAME       MODE         VERSION   AGE
simplest   deployment   0.60.0    12m

# delete
$ kubectl delete -n test -f v0_60_0_jaeger_cfg.yaml                
opentelemetrycollector.opentelemetry.io "simplest" deleted

# create
$ kubectl apply -n test -f v0_60_0_jaeger_cfg.yaml 
opentelemetrycollector.opentelemetry.io/simplest created

# status
$ kubectl get opentelemetrycollectors.opentelemetry.io -n test simplest 
NAME       MODE         VERSION   AGE
simplest   deployment   0.61.0    3s
```
</details>

Should we simply add that to the issue description, add a auto update flag or do you have other ideas?

pavolloffay: That is a good finding, we should definitely document this in the codebase that if the error is thrown during the upgrade then the version is never updated. 

There are several ways how we can proceed:
* document the breaking change and how to resolve it
* trigger the upgrade routine during the reconciliation logic

frzifus: In case of a failing update, we will now create a log entry and event. See: https://github.com/open-telemetry/opentelemetry-operator/pull/1116/commits/557a3ffd2a5f3d07f2a245adc1859c2559c15835

pkg/collector/upgrade/upgrade.go
```diff
@@ -61,7 +61,9 @@ func (u VersionUpgrade) ManagedInstances(ctx context.Context) error {
 		}
 		upgraded, err := u.ManagedInstance(ctx, original)
 		if err != nil {
-			// nothing to do at this level, just go to the next instance
+			const msg = "automated update not possible. Configuration must be checked manually and recreated."
```
pavolloffay, line 64: `Configuration must be corrected manually and CR instance must be re-created.`

pkg/collector/upgrade/v0_61_0-invalid.yaml
```diff
@@ -0,0 +1,8 @@
+---
```
pavolloffay, line 1: There is a convention in go to put files like these into `testdata` directory in the same package

pavolloffay (commented):
PR looks good, just some minor comments

pavolloffay (approved):

PR #1124, opened 2022-09-23
Add scrape configs endpoint

### LINKED ISSUE #1106: target allocator missing default value for scrape settings for prometheusCR
Repo with reproducer: https://github.com/seankhliao/testrepo0032

Target Allocator: 0.1.0 and main
Operator: 0.60.0
Collector: 0.60.0

When creating jobs/scrape targets from ServiceMonitors, the target allocator doesn't produce values for:
- `__scrape_interval__`
- `__scrape_timeout__`
- `__scheme__`
- `__metrics_path__`
- maybe also `__param_*`

When the prometheusreceiver reads the HTTP SD configs, it produces errors like the below for `__scrape_*`,
`"target": "//a.b.c.d:port", "error": "Get \"//a.b.c.d:port\": unsupported protocol scheme \"\""` for missing scheme, and 404s (or other invalid input) for missing paths:

```
2022-09-16T16:59:39.911Z	error	scrape/scrape.go:488	Creating target failed	{"kind": "receiver", "name": "prometheus", "pipeline": "metrics", "scrape_pool": "/jobs/serviceMonitor%2Fdefault%2Fsm-test%2F0/targets", "error": "instance 0 in group http://ta-test-targetallocator/jobs/serviceMonitor%2Fdefault%2Fsm-test%2F0/targets?collector_id=ta-test-collector-0:1: scrape interval cannot be 0", "errorVerbose": "scrape interval cannot be 0\ngithub.com/prometheus/prometheus/scrape.PopulateLabels\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/target.go:446\ngithub.com/prometheus/prometheus/scrape.TargetsFromGroup\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/target.go:504\ngithub.com/prometheus/prometheus/scrape.(*scrapePool).Sync\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/scrape.go:486\ngithub.com/prometheus/prometheus/scrape.(*Manager).reload.func1\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/manager.go:222\nruntime.goexit\n\truntime/asm_amd64.s:1571\ninstance 0 in group http://ta-test-targetallocator/jobs/serviceMonitor%2Fdefault%2Fsm-test%2F0/targets?collector_id=ta-test-collector-0:1\ngithub.com/prometheus/prometheus/scrape.TargetsFromGroup\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/target.go:506\ngithub.com/prometheus/prometheus/scrape.(*scrapePool).Sync\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/scrape.go:486\ngithub.com/prometheus/prometheus/scrape.(*Manager).reload.func1\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/manager.go:222\nruntime.goexit\n\truntime/asm_amd64.s:1571"}
github.com/prometheus/prometheus/scrape.(*scrapePool).Sync
	github.com/prometheus/prometheus@v0.38.0/scrape/scrape.go:488
github.com/prometheus/prometheus/scrape.(*Manager).reload.func1
	github.com/prometheus/prometheus@v0.38.0/scrape/manager.go:222
2022-09-16T16:59:39.911Z	error	scrape/scrape.go:488	Creating target failed	{"kind": "receiver", "name": "prometheus", "pipeline": "metrics", "scrape_pool": "/jobs/dummy/targets", "error": "instance 0 in group http://ta-test-targetallocator/jobs/dummy/targets?collector_id=ta-test-collector-0:0: scrape interval cannot be 0", "errorVerbose": "scrape interval cannot be 0\ngithub.com/prometheus/prometheus/scrape.PopulateLabels\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/target.go:446\ngithub.com/prometheus/prometheus/scrape.TargetsFromGroup\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/target.go:504\ngithub.com/prometheus/prometheus/scrape.(*scrapePool).Sync\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/scrape.go:486\ngithub.com/prometheus/prometheus/scrape.(*Manager).reload.func1\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/manager.go:222\nruntime.goexit\n\truntime/asm_amd64.s:1571\ninstance 0 in group http://ta-test-targetallocator/jobs/dummy/targets?collector_id=ta-test-collector-0:0\ngithub.com/prometheus/prometheus/scrape.TargetsFromGroup\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/target.go:506\ngithub.com/prometheus/prometheus/scrape.(*scrapePool).Sync\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/scrape.go:486\ngithub.com/prometheus/prometheus/scrape.(*Manager).reload.func1\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/manager.go:222\nruntime.goexit\n\truntime/asm_amd64.s:1571"}
github.com/prometheus/prometheus/scrape.(*scrapePool).Sync
	github.com/prometheus/prometheus@v0.38.0/scrape/scrape.go:488
github.com/prometheus/prometheus/scrape.(*Manager).reload.func1
	github.com/prometheus/prometheus@v0.38.0/scrape/manager.go:222
2022-09-16T16:59:39.911Z	error	scrape/scrape.go:488	Creating target failed	{"kind": "receiver", "name": "prometheus", "pipeline": "metrics", "scrape_pool": "/jobs/serviceMonitor%2Fdefault%2Fsm-test%2F0/targets", "error": "instance 0 in group http://ta-test-targetallocator/jobs/serviceMonitor%2Fdefault%2Fsm-test%2F0/targets?collector_id=ta-test-collector-0:0: scrape interval cannot be 0", "errorVerbose": "scrape interval cannot be 0\ngithub.com/prometheus/prometheus/scrape.PopulateLabels\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/target.go:446\ngithub.com/prometheus/prometheus/scrape.TargetsFromGroup\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/target.go:504\ngithub.com/prometheus/prometheus/scrape.(*scrapePool).Sync\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/scrape.go:486\ngithub.com/prometheus/prometheus/scrape.(*Manager).reload.func1\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/manager.go:222\nruntime.goexit\n\truntime/asm_amd64.s:1571\ninstance 0 in group http://ta-test-targetallocator/jobs/serviceMonitor%2Fdefault%2Fsm-test%2F0/targets?collector_id=ta-test-collector-0:0\ngithub.com/prometheus/prometheus/scrape.TargetsFromGroup\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/target.go:506\ngithub.com/prometheus/prometheus/scrape.(*scrapePool).Sync\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/scrape.go:486\ngithub.com/prometheus/prometheus/scrape.(*Manager).reload.func1\n\tgithub.com/prometheus/prometheus@v0.38.0/scrape/manager.go:222\nruntime.goexit\n\truntime/asm_amd64.s:1571"}
github.com/prometheus/prometheus/scrape.(*scrapePool).Sync
	github.com/prometheus/prometheus@v0.38.0/scrape/scrape.go:488
github.com/prometheus/prometheus/scrape.(*Manager).reload.func1
	github.com/prometheus/prometheus@v0.38.0/scrape/manager.go:222
```

Initial report in slack: https://cloud-native.slack.com/archives/C033BJ8BASU/p1663327567933649

jaronoff97:
This PR is the first one needed in order to close #1106. We add a new endpoint which exposes the scrape configs that the target allocator has access to both from its configmap and also from prometheus CRDs. 

A followup PR will be made in collector contrib to start using the endpoint.

### DIFF
--- a/cmd/otel-allocator/discovery/discovery.go
+++ b/cmd/otel-allocator/discovery/discovery.go
@@ -47,6 +47,16 @@ func NewManager(log logr.Logger, ctx context.Context, logger log.Logger, options
 	}
 }
 
+func (m *Manager) GetScrapeConfigs() map[string]*config.ScrapeConfig {
+	jobToScrapeConfig := map[string]*config.ScrapeConfig{}
+	for _, c := range m.configsMap {
+		for _, scrapeConfig := range c.ScrapeConfigs {
+			jobToScrapeConfig[scrapeConfig.JobName] = scrapeConfig
+		}
+	}
+	return jobToScrapeConfig
+}
+
 func (m *Manager) ApplyConfig(source allocatorWatcher.EventSource, cfg *config.Config) error {
 	m.configsMap[source] = cfg
 
--- a/cmd/otel-allocator/main.go
+++ b/cmd/otel-allocator/main.go
@@ -3,6 +3,8 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"github.com/ghodss/yaml"
+	yaml2 "gopkg.in/yaml.v2"
 	"net/http"
 	"net/url"
 	"os"
@@ -152,6 +154,7 @@ func newServer(log logr.Logger, allocator allocation.Allocator, discoveryManager
 	}
 	router := mux.NewRouter().UseEncodedPath()
 	router.Use(s.PrometheusMiddleware)
+	router.HandleFunc("/scrape_configs", s.ScrapeConfigsHandler).Methods("GET")
 	router.HandleFunc("/jobs", s.JobHandler).Methods("GET")
 	router.HandleFunc("/jobs/{job_id}/targets", s.TargetsHandler).Methods("GET")
 	router.Path("/metrics").Handler(promhttp.Handler())
@@ -188,6 +191,23 @@ func (s *server) Shutdown(ctx context.Context) error {
 	s.logger.Info("Shutting down server...")
 	s.k8sClient.Close()
 	return s.server.Shutdown(ctx)
+}
+
+func (s *server) ScrapeConfigsHandler(w http.ResponseWriter, r *http.Request) {
+	configs := s.discoveryManager.GetScrapeConfigs()
+	configBytes, err := yaml2.Marshal(configs)
+	if err != nil {
+		errorHandler(err, w, r)
+	}
+	jsonConfig, err := yaml.YAMLToJSON(configBytes)
+	if err != nil {
+		errorHandler(err, w, r)
+	}
+	w.Header().Set("Content-Type", "application/json")
+	_, err = w.Write(jsonConfig)
+	if err != nil {
+		errorHandler(err, w, r)
+	}
 }
 
 func (s *server) JobHandler(w http.ResponseWriter, r *http.Request) {


Makefile
```diff
@@ -175,7 +175,7 @@ set-test-image-vars:
...
+	docker buildx build --platform linux/amd64 -t ${IMG} --build-arg VERSION_PKG=${VERSION_PKG} --build-arg VERSION=${VERSION} --build-arg VERSION_DATE=${VERSION_DATE} --build-arg OTELCOL_VERSION=${OTELCOL_VERSION} --build-arg TARGETALLOCATOR_VERSION=${TARGETALLOCATOR_VERSION} --build-arg AUTO_INSTRUMENTATION_JAVA_VERSION=${AUTO_INSTRUMENTATION_JAVA_VERSION}  --build-arg AUTO_INSTRUMENTATION_NODEJS_VERSION=${AUTO_INSTRUMENTATION_NODEJS_VERSION} --build-arg AUTO_INSTRUMENTATION_PYTHON_VERSION=${AUTO_INSTRUMENTATION_PYTHON_VERSION} --build-arg AUTO_INSTRUMENTATION_DOTNET_VERSION=${AUTO_INSTRUMENTATION_DOTNET_VERSION} .
```
jaronoff97, line 178: this is so that I can build on my m1, does this work on other computers?

cmd/otel-allocator/watcher/file.go
```diff
@@ -37,6 +39,8 @@ func (f *FileWatcher) Start(upstreamEvents chan Event, upstreamErrors chan error
 	go func() {
 		for {
 			select {
+			case _ = <-f.closer:
```
jaronoff97, line 42: This fixes a bug found that was causing the TA to panic on a configmap change

pkg/collector/reconcile/service.go
```diff
@@ -235,6 +235,7 @@ func expectedServices(ctx context.Context, params Params, expected []corev1.Serv
 			updated.ObjectMeta.Labels[k] = v
 		}
 		updated.Spec.Ports = desired.Spec.Ports
+		updated.Spec.Selector = desired.Spec.Selector
```
jaronoff97, line 238: This fixes a bug that wasn't updating the version on services when a collector image would change

kristinapathak (commented):
Very excited for this. 🙂

cmd/otel-allocator/main.go
```diff
@@ -3,6 +3,8 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"github.com/ghodss/yaml"
+	yaml2 "gopkg.in/yaml.v2"
```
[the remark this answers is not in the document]
jaronoff97: Unfortunately yes. We need the first one to correctly marshal the scrape configs and the second one to convert it to JSON maintaining the string cases. Using only one yielded an invalid scrape config because one would not use the yaml hints which meant we wouldn't have a job name set. The other was unable to correctly return a JSON blob, open to suggestions if there's a better way to accomplish this. The comment I wrote on the handler in main.go was meant to explain this.

cmd/otel-allocator/discovery/discovery.go
```diff
@@ -47,6 +47,16 @@ func NewManager(log logr.Logger, ctx context.Context, logger log.Logger, options
 	}
 }
 
+func (m *Manager) GetScrapeConfigs() map[string]*config.ScrapeConfig {
+	jobToScrapeConfig := map[string]*config.ScrapeConfig{}
+	for _, c := range m.configsMap {
+		for _, scrapeConfig := range c.ScrapeConfigs {
+			jobToScrapeConfig[scrapeConfig.JobName] = scrapeConfig
```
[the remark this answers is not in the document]
jaronoff97: In this case job is the top level key because we are looking at scrape config where the unique identifier is the job name (that's what Prometheus discovery manager uses as it's map key)

Makefile
pavolloffay: I would prefer to keep plain docker build and have buildx as opt-in feature.

Makefile
```diff
@@ -184,7 +184,7 @@ container-push:
 
 .PHONY: container-target-allocator
 container-target-allocator:
-	docker build -t ${TARGETALLOCATOR_IMG} cmd/otel-allocator
+	docker buildx build --platform linux/amd64 -t ${TARGETALLOCATOR_IMG} cmd/otel-allocator
```
pavolloffay, line 187: The same here, see my comment about buildx

pavolloffay (changes requested):

pkg/collector/container.go
```diff
@@ -78,6 +78,15 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
 		},
 	})
 
+	if otelcol.Spec.TargetAllocator.Enabled {
+		// We need to add a SHARD here so the collector is able to keep targets after the hashmod operation which is
+		// added by default by the operator generator.
+		envVars = append(envVars, corev1.EnvVar{
+			Name:  "SHARD",
+			Value: "0",
+		})
```
secustor, line 87: Why not use the [environment variable substitution in config files](https://opentelemetry.io/docs/collector/configuration/#configuration-environment-variables) as it is done yet?    
What is the difference to `collector_id`/[`POD_NAME`](https://github.com/open-telemetry/opentelemetry-operator/pull/1124/files#diff-a217bf24650bbe5f454d95c3480eb2a8353d8e9762f560e5d88b23571cbbcafeR73) which is in use right now? 

That way we would lose sharding as all collectors will get Shard `0`.

cmd/otel-allocator/main.go
```diff
@@ -152,6 +154,7 @@ func newServer(log logr.Logger, allocator allocation.Allocator, discoveryManager
 	}
 	router := mux.NewRouter().UseEncodedPath()
 	router.Use(s.PrometheusMiddleware)
+	router.HandleFunc("/scrape_configs", s.ScrapeConfigsHandler).Methods("GET")
```
secustor, line 157: I'm worried about the response sizes and traffic which would be generated by this. Every collector would request all scrapes_configs on every sync tic. 

Can you post an example response? 

I think there should be parameter as it is implemented at the targets endpoint https://github.com/open-telemetry/opentelemetry-operator/pull/1124/files#diff-3ea793544213bc190a7701f8934a02c85faa78051509e02f2ec0c4ff3fcef608R237

Makefile
ekarlso: How would you opt in then? Like A if statement in the makefile?

cmd/otel-allocator/watcher/file.go
jaronoff97: Normally, I would agree but in order to actually run this new code which had configuration changes (something a user would most likely do when enabling this feature) this change was required to ensure the TA wouldn't panic. I can split it out if we think that's worth it, but I kept it in so that users could progressively upgrade without panics.

pkg/collector/container.go
jaronoff97: the difference is that we don't control the configuration that gets generated (the configuration that uses this is generated by prometheus.) Another issue here is that prometheus uses a non-otel-standard pattern for environment variables which requires us to manually fix this ([seen here](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/14464/files#diff-8172ccc732b059ffa8d9d8858643701e32c90c3d3ab8b6697fa74ca1c8c09fb5R200-R206)). This sharding is only used in the prometheus configuration via hashmod and is going to be unused which is desired because the TA is the component doing the sharding.

Makefile
jaronoff97: made a follow up issue for this one https://github.com/open-telemetry/opentelemetry-operator/issues/1133

pkg/collector/container.go
Aneurysm9: > This sharding is only used in the prometheus configuration via hashmod and is going to be unused which is desired because the TA is the component doing the sharding.

Can you detail this a bit more in a comment here?  It's not 100% clear to me how this works.

jaronoff97: Sure, when we call prometheus' config generator [here](https://github.com/open-telemetry/opentelemetry-operator/blob/main/cmd/otel-allocator/watcher/promOperator.go#L134), it adds in the [following rules](https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/promcfg.go#L1466-L1477) to the bottom of any CRD (servicemonitor, podmonitor, etc.) These rules are for the experimental sharding functionality provided by prometheus which use an envrionment variable `SHARD` to determine whether or not to scrape a target. 

Because we are using the target allocator for sharding rather than prometheus' sharding, we don't want to utilize this functionality at all. If we don't set the SHARD variable at all, however, the evaluation of this rule will always be false which means the keep directive will drop all targets. Because the default numShards is 1, we need to set the collector's SHARD variable to be 0 which will keep any targets from the relabelling. This is fine because the collector is given some subset of targets to scrape from the target allocator.

cmd/otel-allocator/main.go
secustor: K, then solve issues as they arise and not block this moving forward.

pkg/collector/container.go
secustor: I think we should start to actively document this. IMO it is fine for merge to merge this now, but it should be definitively on the short list.    
As users will be confused how this `SHARD` interacts with the sharding of TA.

secustor (approved):

pkg/collector/container.go
jaronoff97: i can add a comment to the docs :)

pkg/collector/container.go
```diff
@@ -78,6 +78,15 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
 		},
 	})
 
+	if otelcol.Spec.TargetAllocator.Enabled {
+		// We need to add a SHARD here so the collector is able to keep targets after the hashmod operation which is
+		// added by default by the operator generator.
```
Aneurysm9, line 83:
```suggestion
		// added by default by the Prometheus operator's config generator.
		// All collector instances use SHARD == 0 as they only receive targets
		// allocated to them and should not use the Prometheus hashmod-based
		// allocation.
```

A bit more detail at this spot is probably worthwhile.

Aneurysm9 (approved):

pavolloffay (approved):

Issue #1073, opened 2022-09-02
Acquiring leader lease at operator startup can take longer than 2 minutes causing test failures


kevinearls:
I sometimes seem log messages like the ones below which indicates that attempting to acquire a leader lease can take 2.5 minutes or more.  (Normally this will take far less than a second.). This most likely results in CI test failures as described in #959 

```
I0902 08:08:22.532168       1 leaderelection.go:248] attempting to acquire leader lease opentelemetry-operator-system/9f7554c3.opentelemetry.io...
{"level":"info","ts":1662106149.4242973,"logger":"opentelemetrycollector-resource","msg":"default","name":"sidecar"}
{"level":"info","ts":1662106149.4341345,"logger":"opentelemetrycollector-resource","msg":"validate create","name":"sidecar"}
{"level":"info","ts":1662106149.4573686,"logger":"instrumentation-resource","msg":"default","name":"java"}
{"level":"info","ts":1662106149.5142503,"logger":"instrumentation-resource","msg":"validate create","name":"java"}
I0902 08:11:26.563394       1 leaderelection.go:258] successfully acquired lease opentelemetry-operator-system/9f7554c3.opentelemetry.io

```

pavolloffay:
Maybe we should make the lease time configurable and use appropriate value for e2e tests.

data-dude:
I'm offline.  It's attempting to get a lease at opentelemetry.io and failing.  How to make this work offline?

turkenh:
We had the same issue on Crossplane side and this is how we fixed it: https://github.com/crossplane/crossplane/pull/4859/files

PR #1134, opened 2022-10-01
Make lease time configurable

x-504:
With this change flag `lease-time` can be set with the desired time in seconds.

If this the flag `lease-time` is not set, it will use default value.

See https://github.com/open-telemetry/opentelemetry-operator/issues/1073


main.go
```diff
@@ -169,6 +170,14 @@ func main() {
 		mgrOptions.NewCache = cache.MultiNamespacedCacheBuilder(strings.Split(watchNamespace, ","))
 	}
 
+	leaseTime, found := os.LookupEnv("LEASE_TIME")
```
pavolloffay, line 173: To be consistent with the configuration mechanism that the operator already uses I would propose to change this to a flag.

main.go
```diff
@@ -83,6 +83,7 @@ func main() {
 		autoInstrumentationDotNet string
 		labelsFilter              []string
 		webhookPort               int
+		leaseTime                 int
```
jaronoff97, line 86: This should probably just be a duration instead of an integer

jaronoff97 (changes requested):

jaronoff97:
@albertollamaso are you interested in working this still? Left a comment if so!

jaronoff97:
closes #1073

albertollamaso:
> @albertollamaso are you interested in working this still? Left a comment if so!

I'll have time to check in few weeks 🙏

atoulme:

PR #1138, opened 2022-10-04
adds deepcopy missing implementation for TargetCPUUtilization field of AutoscalerSpec

avadhut123pisal:
After running `make generate` , found some code that is not committed yet to the repo, adding the same.


pavolloffay:
I am not sure how PR without missing generated code could have been committed. There should be a CI check to avoid merging this.

pavolloffay (approved):

pavolloffay:
#1154 fixes the check

PR #1144, opened 2022-10-06
reconcile otel collector on given context

frzifus:
Saw randomly that a new context is created in the reconciler. Or is there a reason for this?


pavolloffay (approved):
Good catch!

PR #1149, opened 2022-10-07
Set OTEL_METRICS_EXPORTER to none to prevent using the default value

### LINKED ISSUE #1148: Python auto instrumentation fails with current version
With a very basic Python app, injection causes the auto instrumentation to fail with `RuntimeError: Requested component 'otlp_proto_grpc' not found in entry points for 'opentelemetry_metrics_exporter'`. See https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1361, but I'm happy to provide a repro as well.

What's happening is the default value for `OTEL_METRICS_EXPORTER` is `otlp_proto_grpc`, but that exporter is not included in the image's [requirements.txt](https://github.com/open-telemetry/opentelemetry-operator/blob/main/autoinstrumentation/python/requirements.txt) as the operator should be using HTTP.

Until we implement an OTLP/HTTP exporter for python that can be used in the operator (https://github.com/open-telemetry/opentelemetry-python/issues/2447) I think we should just disable metrics in the operator by setting `OTEL_METRICS_EXPORTER=none`.

aabmass:
Fixes #1148


srikanthccv (approved):

pavolloffay:
e2e test failed

pavolloffay (approved):

PR #1152, opened 2022-10-09
Using immutable labels as service selectors

### LINKED ISSUE #1107: Error updating container image version
I'm facing problems when updating the collector version through the CRD.
Apparently the label `app.kubernetes.io/version` is being updated in the Deployment `spec.template.metadata.labels`
while the Service `spec.selector` is not.
It causes application telemetry not to reach the collector's pod after the update rollout is complete.

### Steps to reproduce

1. Install the collector through the CRD OpenTelemetryCollector:

```shell
kubectl apply -f - <<EOF
kind: OpenTelemetryCollector
metadata:
  name: simplest
spec:
  image: otel/opentelemetry-collector:0.59.0
  config: |
    receivers:
      otlp:
        protocols:
          grpc:
          http:
    exporters:
      otlp/jaeger:
        endpoint: jaeger-collector.jaeger.svc.cluster.local:4317
        tls:
          insecure: true
    service:
      pipelines:
        traces:
          receivers: [ otlp ]
          exporters: [ otlp/jaeger ]
EOF
```

2. Update the collector's version from `0.59.0` to `0.60.0`

```shell
kubectl patch otelcol simplest --type='json' -p='[{"op": "replace", "path": "/spec/image", "value":"otel/opentelemetry-collector:0.60.0"}]'
```

3. Compare the Service selector with Pod labels

```shell
kubectl get svc simplest-collector -o jsonpath='{.spec.selector}'
kubectl get deploy simplest-collector -o jsonpath='{.spec.template.metadata.labels}'
```
or check your application logs to see something like that:
```log
[otel.javaagent 2022-09-17 18:20:41:732 +0000] [OkHttp http://simplest-collector.otel.svc.cluster.local:4317/...] ERROR io.opentelemetry.exporter.internal.grpc.OkHttpGrpcExporter - Failed to export spans. The request could not be executed. Full error message: Failed to connect to simplest-collector.otel.svc.cluster.local/10.96.60.59:4317
```

### Operator version

otelcol version 0.60.0

### Proposal

Maybe we could just use the immutable labels as service selectors ([like Deployment does](https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/collector/deployment.go#L46)).
Another solution would be to simply fix the Service selector by updating its value as well.

If it makes sense, I would be happy to contribute 🙂

angelokurtis:
fixes #1107

### DIFF
--- a/pkg/collector/reconcile/service.go
+++ b/pkg/collector/reconcile/service.go
@@ -71,9 +71,8 @@ func desiredService(ctx context.Context, params Params) *corev1.Service {
 	labels := collector.Labels(params.Instance, []string{})
 	labels["app.kubernetes.io/name"] = naming.Service(params.Instance)
 
-	// by coincidence, the selector is the same as the label, but note that the selector points to the deployment
-	// whereas 'labels' refers to the service
-	selector := labels
+	selector := collector.SelectorLabels(params.Instance)
+	selector["app.kubernetes.io/name"] = naming.Service(params.Instance)
 
 	config, err := adapters.ConfigFromString(params.Instance.Spec.Config)
 	if err != nil {
@@ -176,7 +175,7 @@ func monitoringService(ctx context.Context, params Params) *corev1.Service {
 	labels := collector.Labels(params.Instance, []string{})
 	labels["app.kubernetes.io/name"] = naming.MonitoringService(params.Instance)
 
-	selector := collector.Labels(params.Instance, []string{})
+	selector := collector.SelectorLabels(params.Instance)
 	selector["app.kubernetes.io/name"] = fmt.Sprintf("%s-collector", params.Instance.Name)
 
 	return &corev1.Service{
--- a/pkg/collector/reconcile/service.go
+++ b/pkg/collector/reconcile/service.go
@@ -71,9 +71,6 @@ func desiredService(ctx context.Context, params Params) *corev1.Service {
 	labels := collector.Labels(params.Instance, []string{})
 	labels["app.kubernetes.io/name"] = naming.Service(params.Instance)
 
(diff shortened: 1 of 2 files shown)


pavolloffay:
CI failed

pkg/collector/reconcile/service.go
```diff
@@ -71,9 +71,8 @@ func desiredService(ctx context.Context, params Params) *corev1.Service {
 	labels := collector.Labels(params.Instance, []string{})
 	labels["app.kubernetes.io/name"] = naming.Service(params.Instance)
 
-	// by coincidence, the selector is the same as the label, but note that the selector points to the deployment
-	// whereas 'labels' refers to the service
-	selector := labels
+	selector := collector.SelectorLabels(params.Instance)
+	selector["app.kubernetes.io/name"] = naming.Service(params.Instance)
```
pavolloffay, line 75: Do we need this label? there is already `"app.kubernetes.io/instance":   naming.Truncate("%s.%s", 63, instance.Namespace, instance.Name),`

pavolloffay (commented):
Could you please check the selector lebels in the e2e test ? ### 

https://github.com/open-telemetry/opentelemetry-operator/blob/main/tests/e2e/smoke-simplest/00-assert.yaml#L43 and for the headless as well

angelokurtis:
> Could you please check the selector lebels in the e2e test ?

We could add a check for all other selector labels, but I'm not sure if we could for `app.kubernetes.io/instance` because its value contains the namespace created by KUTTL, and it's known only during execution.

What do you think? Any suggestion?

VineethReddy02 (approved):

pavolloffay (approved):

PR #1178, opened 2022-10-16
Adds support of priorityClassName for otel collector

### LINKED ISSUE #1159: Add support for setting `priorityClassName`
I'd like to be able to specify the [`priorityClassName`](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/) with f.x. `system-node-critical` for my daemonset pods created by the operator, so I can ensure that there is a collector pod running on each node.

I was looking into modifying the helm charts to support this (https://github.com/open-telemetry/opentelemetry-helm-charts/issues/370), but then of course this operator needs updating and I'm not really familiar with Go/operator code. So while I could apply it to the chart for the operator deployment, I can't apply it to the underlying created pods by the operator I think.

avadhut123pisal:
Fixes #1159

### DIFF
--- a/apis/v1alpha1/opentelemetrycollector_types.go
+++ b/apis/v1alpha1/opentelemetrycollector_types.go
@@ -138,6 +138,11 @@ type OpenTelemetryCollectorSpec struct {
 	// HostNetwork indicates if the pod should run in the host networking namespace.
 	// +optional
 	HostNetwork bool `json:"hostNetwork,omitempty"`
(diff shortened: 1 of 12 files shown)


avadhut123pisal:
@pavolloffay  Can you please re-run the e2e test case execution, I cannot find the concrete reason behind the failure of test case.

pavolloffay (approved):

PR #1205, opened 2022-10-26
docs: add link to helm-chart

povilasv:
Adding some docs for installation using helm charts, let me know what you think?

CC @TylerHelmuth @dmitryax


jaronoff97 (approved):
This is a great add to the docs, thank you!

pavolloffay (approved):
LGTM, but we should restructure docs and list all possible options how to install it.

PR #1215, opened 2022-11-01
Prepare release 0.63.1

frzifus:
Are there any missing migrations? [[otelcol-contrib-v0.62.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.62.0)][[otelcol-contrib-v0.63.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.63.0)]

Please have a look on the bumped versions in `version.txt`.

Ups i didnt see https://github.com/open-telemetry/opentelemetry-operator/issues/1210#issuecomment-1295696100. Let me change it to `v0.63.1`.

Closes #1210 #1176

cc @pavolloffay

PR #1131, opened 2022-09-29
Release v0.61.0

yuriolisa:
Signed-off-by: Yuri Sa <yurimsa@gmail.com>

Resolves #1130 

Blocked by 
- [x] https://github.com/open-telemetry/opentelemetry-operator/pull/1116
- [x] https://github.com/open-telemetry/opentelemetry-operator/pull/1126
- [x] https://github.com/open-telemetry/opentelemetry-operator/issues/1143

PR #1217, opened 2022-11-03
Prepare release 0.62.1

### LINKED ISSUE #1176: Release the operator to use collector v0.62.1

frzifus:
Please have a look @Aneurysm9

Closes #1176

### DIFF
--- a/versions.txt
+++ b/versions.txt
@@ -2,17 +2,17 @@
 # by default with the OpenTelemetry Operator. This would usually be the latest
 # stable OpenTelemetry version. When you update this file, make sure to update the
 # the docs as well.
-opentelemetry-collector=0.61.0
+opentelemetry-collector=0.62.1
 
 # Represents the current release of the OpenTelemetry Operator.
-operator=0.61.0
+operator=0.62.1
 
 # Represents the current release of the Target Allocator.
 targetallocator=0.60.0
 
 # Represents the current release of Java instrumentation.
 # Should match autoinstrumentation/java/version.txt
-autoinstrumentation-java=1.11.1
+autoinstrumentation-java=1.19.1
 
 # Represents the current release of NodeJS instrumentation.
 # Should match value in autoinstrumentation/nodejs/package.json
@@ -24,4 +24,4 @@ autoinstrumentation-python=0.34b0
 
 # Represents the current release of DotNet instrumentation.
 # Should match autoinstrumentation/dotnet/version.txt
-autoinstrumentation-dotnet=0.3.1-beta.1
+autoinstrumentation-dotnet=0.4.0-beta.1
--- a/.github/workflows/publish-autoinstrumentation-dotnet.yaml
+++ b/.github/workflows/publish-autoinstrumentation-dotnet.yaml
@@ -30,6 +30,8 @@ jobs:
           images: ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-dotnet
           tags: |
             type=match,pattern=v(.*),group=1,value=v${{ env.VERSION }}
+          flavor: |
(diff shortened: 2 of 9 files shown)


versions.txt
```diff
@@ -2,17 +2,17 @@
...
 # the docs as well.
-opentelemetry-collector=0.61.0
+opentelemetry-collector=0.62.1
 
 # Represents the current release of the OpenTelemetry Operator.
-operator=0.61.0
+operator=0.62.1
 
 # Represents the current release of the Target Allocator.
 targetallocator=0.60.0
```
Aneurysm9, line 11:
```suggestion
targetallocator=0.62.1
```

I think this image version should be moving in parallel with the operator image.

frzifus: Then we might need to adapt the `0.63.1` release too!

In `0.61.0` the targetallocator was not bumped. [[diff](https://github.com/open-telemetry/opentelemetry-operator/pull/1131/files#diff-eb8281dc5868dec21668321219be72bacb9baa0acf7e43ac159d5a53e9d44abcR11)]

So i did also not touch it in `0.63.1`. Seems both passed the review. [[diff](https://github.com/open-telemetry/opentelemetry-operator/pull/1215/files#diff-eb8281dc5868dec21668321219be72bacb9baa0acf7e43ac159d5a53e9d44abcR11)]

How should we handle that? @Aneurysm9

Aneurysm9: v0v

I think the version of the target allocator was only recently changed to not be `0.1.0` with every release, so not sure there's a good precedent here.  I think we update it with `0.62.1` as you've done here and then going forward from `0.64.0`.

frzifus: But then its a upgrade in `0.62.1` and a downgrade in `0.63.1` right?
| Operator-Version  | targetallocator  |
|---|---|
| 0.60.0 | 0.60.0  |
| 0.61.0 | 0.60.0  |
| **0.62.1** | 0.62.1  |
| 0.63.1 | **0.60.0**  |
| 0.64.0 | 0.64.0  |

Aneurysm9: > But then its a upgrade in `0.62.1` and a downgrade in `0.63.1` right?
> ...
> 
> So would it make sense to keep `0.60.0` in the `0.62.1` release too?

No, I don't think so.  Make a `v0.36.1-oops-we-messed-up` release if you like, but I don't think it makes sense to ship old and mismatched versions of this component to maintain some notion of monotonicity that I don't think is necessary.

Aneurysm9 (approved):

versions.txt
pavolloffay: The TA image version is taken from https://github.com/open-telemetry/opentelemetry-operator/blob/main/versions.txt#L11 

And published only for changes is https://github.com/open-telemetry/opentelemetry-operator/blob/main/.github/workflows/publish-target-allocator.yaml#L5


but https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

> When using the push and pull_request events, you can configure a workflow to run based on what file paths are changed. Path filters are not evaluated for pushes of tags.


So it means that we can change TA version in the release PR ❤️

pavolloffay:
I think we need to disable publishing of latest tag for the operator https://github.com/docker/metadata-action/issues/161

pavolloffay:
After the PR is merged we should as well backport readme and changelog changes to the main branch

VineethReddy02 (approved):

frzifus:
@pavolloffay could you have a look please?

pavolloffay (approved):
the auto-instrumentation workflows didn't have to be changed, but LGTM

PR #1220, opened 2022-11-03
feat(otel-allocator): use type for AllocationStrategy

### LINKED ISSUE #1101: Make target allocator's allocation strategy a type
https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1alpha1/opentelemetrycollector_types.go#L127

Benefits of the type:
* documented in the CRD
* automatically validated

cc) @jaronoff97

secustor:
Fixes #1101

This PR changes the AllocationStrategy from String to a new type. 
Further it now the Allocator references now to the constants provided by the operator package. 

Based on the changes of `go.mod` and `go.sum`, the footprint of this additionally dependency seems to be minimal.


cmd/otel-allocator/allocation/consistent_hashing.go
```diff
@@ -22,13 +22,14 @@ import (
...
 
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
 	"github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/diff"
 	"github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/target"
 )
 
 var _ Allocator = &consistentHashingAllocator{}
 
-const consistentHashingStrategyName = "consistent-hashing"
+const consistentHashingStrategyName = string(v1alpha1.OpenTelemetryTargetAllocatorAllocationStrategyConsistentHashing)
```
jaronoff97, line 32: question: are we okay having the allocator import this from the operator? It means that the allocator now has a required dependency on the operator which i'm not sure we want. cc @pavolloffay

jaronoff97 (approved):
Only one question here, otherwise thanks for this!

cmd/otel-allocator/allocation/consistent_hashing.go
secustor: I think it is fine here, as mentioned in the description we are not adding any additional dependencies.  
The benefit would be that we have source of truth regarding the available options.

secustor:
I have deployed operator and allocator and tried all 3 different setting for `AllocationStrategy` and redeployed manually in between to be sure that the setting has been applied. 
The TA logs showed no errors.

cmd/otel-allocator/Dockerfile
```diff
@@ -3,12 +3,12 @@ FROM golang:1.19 as builder
 
 WORKDIR /app
 
-# Copy go mod and sum files
-COPY go.mod go.sum ./
+# copy operator and allocator code
+COPY . .
```
Aneurysm9, line 7: Can this change be restructured/avoided?  Doing this can cause code changes that do not change dependencies to invalidate the cached FS layer with dependencies already downloaded, which can make developing/testing with this `Dockerfile` annoying.

secustor: `go mod download` will fail we not copy the code first because of the [new replacement directive]( https://github.com/open-telemetry/opentelemetry-operator/pull/1220/files#diff-35d0de98f88a0bce74a9813fa595b33ed20736a2167c94fad07476411c32422aR175)

Alternatively we have to drop dependency to the operator here again.

jaronoff97: IMO we should drop the dependency on the operator, I think it's going to add more complexity than it's worth.

jaronoff97 (approved):

pavolloffay (approved):

Issue #1142, opened 2022-10-05
Add instrumentation state information to pods


noMoreCLI:
Hi

when working with auto-instrumentation I found no quick way to verify which pods auto-instrumentation has taken place,  applied successfully, ignored or failed.
I ended up describing pods and looking for init containers.

Having this information would greatly simplify operations & troubleshooting, if the operator would add state information to pods, when annotating namespace for auto-instrumentation. 

For example listing all pods with successful instrumentation:
`kubectl get pods --all-namespaces -l otel.auto-instrumentation.state=successful`

@svrnm & @pavankrish123

pavolloffay:
I totally agree.

This is related to old issue https://github.com/open-telemetry/opentelemetry-operator/issues/544

akhileshsingh85:
@noMoreCLI  is name "otel.auto-instrumentation.state" for label final?

noMoreCLI:
@akhileshsingh85 
no,  I don't think so. I used this as an illustrative example. 
As I anticipate that in the future additional state information could be added, I think a structured approach could make sense, but open for comments.

svrnm:
Not an expert on K8s, but from an end user perspective it would be good to have a few kinds of information:

* on which workloads have we tried to apply auto-instrumentation (and maybe which one)? so I can identify if one was skipped or if the wrong auto instrumentation was applied for whatever reasons 
* was the auto-instrumentation successful (YES, NO, PENDING, ...)
  * if NO, some error details if possible 
  * if YES, we are good
  * if PENDING, there seems to be something blocking, etc.

damemi:
Would it be better for some of these to use the [`InstrumentationStatus`](https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1alpha1/instrumentation_types.go#L149) field, so that the information is in one spot vs applied over (possibly many) workloads?

pavolloffay:
Correct, the information should be in the status field.

damemi:
> Correct, the information should be in the status field.

As a starting point then, what about just adding status conditions to the crd? https://sdk.operatorframework.io/docs/building-operators/golang/advanced-topics/#manage-cr-status-conditions

I see that CollectorStatus implemented its own fields, but I think at least the conditions @svrnm suggested could be expressed with the upstream type. Willing to help with this

damemi:
[...]

PR #1228, opened 2022-11-07
Add basic instrumentation status condition

damemi:
Ref #1142 

This adds a basic `status` field to the Instrumentation crd based on the upstream [`Condition`](https://github.com/kubernetes/apimachinery/blob/d4f471b82f0a17cda946aeba446770563f92114d/pkg/apis/meta/v1/types.go#L1368) type.

Right now this just defines a single `Ready` status that defaults to false when the webhook catches a new pod. As the relevant agents are processed, the "Reason" identifies the last agent (ie `Reason: NodeAgent`). If the agent is injected successfully, the status becomes true.

This produces successful output like this:

```
$ kubectl get -o yaml instrumentation/my-instrumentation
...
status:
  conditions:
  - lastTransitionTime: "2022-11-07T19:46:58Z"
    message: ""
    reason: JavaAgent
    status: "True"
    type: Ready
```

and failures like this
```
$ kubectl get -o yaml instrumentation/my-instrumentation
...
status:
  conditions:
  - lastTransitionTime: "2022-11-07T19:44:44Z"
    message: 'the container defines env var value via ValueFrom, envVar: JAVA_TOOL_OPTIONS'
    reason: JavaAgent
    status: "False"
    type: Ready
```

Note this status will be overwritten by the next pod the operator attempts to process, so errors may be missed if the CR isn't being actively monitored for changes.

This also adds fields to track which containers have succeeded or failed, by language:
```
$ kubectl get -o yaml instrumentation/my-instrumentation
...
status:
  conditions:
  ...
  failed:
    java:
    - default/javashowcase-service
  succeeded:
    java:
    - default/javashowcase-app
    nodejs:
    - default/nodeshowcase-app
    - default/nodeshowcase-service
```

Unfortunately not all requests will have a pod name/namespace, so this is limited to the container name (which has the obvious risk of duplicates). Not sure how to handle that, but I'm open to ideas. Also open to suggestions on formatting/naming/general approach to either of these.


pavolloffay:
> Note this status will be overwritten by the next pod the operator attempts to process, so errors may be missed if the CR isn't being actively monitored for changes.

If the status is overridden it does not seem useful to me. I would prefer to have an entry per injection in the status.

damemi:
> If the status is overridden it does not seem useful to me.

Yeah, ideally it would know the state of all instrumented pods and report a single persistent "Degraded" status if any are actively failing to roll out (with more info in the reason/message). But as an admission webhook, that's not really doable because the logic is focused more on the individual pod than the overall state of the CRD. Maybe with an `Instrumentation` controller we could do something like that.

> I would prefer to have an entry per injection in the status.

Do you mean the `succeeded`/`failed` fields I added?

yuriolisa:
@damemi, are you willing to continue working on it?

jaronoff97:
@damemi wanted to check in again and see if you were interested in working on this? this would be really appreciated!

Issue #1225, opened 2022-11-04
webhook tls min version to tls_v1.2


kangsheng89:
Currently the webhook on controller-manager is operate on port 9443. The webhook server of the https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/server.go#L77 is configurable

Is there any plan to support setting to use min tls v1.2 as open telemetry collector of v0.59.0? and 
TLS 1.0 and 1.1 are deprecated due to known vulnerabilities and should be avoided.

kangsheng89:
Before add this, the controller-runtime has to be updated to v0.13.1
https://github.com/kubernetes-sigs/controller-runtime/compare/v0.13.0...v0.13.1

kangsheng89:
i think we can update TLS config through this https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/manager.go#L253

first using crypto/tls lib
```
    import (
         "crypto/tls"
    ...
    )

```

```  
       #set const of minTLS
        const defaultMinTLSVersion = tls.VersionTLS12

        ...
        # array of function for config setting
	optionsTlSOptsFuncs := []func(*tls.Config){
		func(config *tls.Config) { minTlsDefault() },
	}
	mgrOptions := ctrl.Options{
		Scheme:                 scheme,
		MetricsBindAddress:     metricsAddr,
		Port:                   webhookPort,
		TLSOpts:                optionsTlSOptsFuncs,
        ....}
```
```
# function to set minTls Default
func minTlsDefault() *tls.Config {
	tlsConfig := &tls.Config{
		// Avoid fallback by default to SSL protocols < TLS1.2
		MinVersion: tls.VersionTLS12,
	}

	return tlsConfig
}
```
This is my initial idea, need to be test and verify. And i need some help to understand the build and debug of opentelemetry-operator on local.

PR #1229, opened 2022-11-08
 set minimun tls version of webhook server to tlsv1.2 ( #1225)

kangsheng89:
This to set the min TLS version through the controller runtime Config. When controller init, it will take in the TLS Config as default setting.


kangsheng89:
close this as using incorrect user id

PR #1230, opened 2022-11-08
fix min tls setting for webhook server (#1225)

kangsheng89:
Fix issue #1225

### DIFF
--- a/main.go
+++ b/main.go
@@ -52,7 +52,7 @@ import (
 )
 
 // We should avoid that users unknowingly use a vulnerable TLS version.
-// The defaults should be a safe configuration
+// The defaults should be a safe configuration.
 const defaultMinTLSVersion = tls.VersionTLS12
 
 var (
--- a/main.go
+++ b/main.go
@@ -290,5 +290,5 @@ func addDependencies(_ context.Context, mgr ctrl.Manager, cfg config.Config, v v
 }
 
 func minTlsDefault(cfg *tls.Config) {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -151,10 +156,16 @@ func main() {
...
+
+	optionsTlSOptsFuncs := []func(*tls.Config){
+		func(config *tls.Config) { minTlsDefault(config) },
+	}
+
 	mgrOptions := ctrl.Options{
 		Scheme:                 scheme,
 		MetricsBindAddress:     metricsAddr,
 		Port:                   webhookPort,
+		TLSOpts:                optionsTlSOptsFuncs,
```
pavolloffay, line 168: Does this configure the TLS version on the webhook server as well? 


```
mgrOptions := ctrl.Options{
		WebhookServer:
```

kangsheng89: yes, i tested it on my local, and i check from sslscan. Later i attached printscr here

kangsheng89:
@pavolloffay can help to approve for running workflow?

kangsheng89:
image
This is before change, where i portforward port from 9443 to localhost:10001

kangsheng89:
image
This is after fix, with portforward from 9443 to localhost:10001

As u can see, now the webhook server that run on port 9433 TLS setting enforce with v1.2 and above

kangsheng89:
let me check the comment, i think should remove it to make it pass the golangci-lint

Issue #1231, opened 2022-11-08
[target-allocator] targets get dropped when using relabel-config Filter with serviceMonitor


moh-osman3:
In https://github.com/open-telemetry/opentelemetry-operator/pull/1127 a new CRD field called `filterStrategy` was added. However, the `relabel-config` option drops all targets in the case of using a serviceMonitor that uses the `hashmod` action in the relabel_config. The resulting label will be matched with the environment variable `$(SHARD)`, but in the prehook no environment variable substitution is taking place. 

```
       {
        "action": "hashmod",
        "modulus": 1,
        "regex": "(.*)",
        "replacement": "$1",
        "separator": ";",
        "source_labels": [
          "__address__"
        ],
        "target_label": "__tmp_hash"
      },
      {
        "action": "keep",
        "regex": "$(SHARD)",
        "replacement": "$1",
        "separator": ";",
        "source_labels": [
          "__tmp_hash"
        ]
      }
```

# What I expected

`filterStrategy: relabel-config` to work properly with serviceMonitors. i.e. targets are filtered properly according to prometheus `relabel_config`

# What actually happened

I observe serviceMonitors running in my cluster but no jobs discovered when I set `filterStrategy: relabel-config` and run 
```
$ curl http://lightstep-collector-targetallocator:80
```

When checking logs I notice that all targets are dropped.

PR #1232, opened 2022-11-08
[target-allocator] Fix bug found when using relabel-config filterStrategy with serviceMonitors

moh-osman3:
Resolves https://github.com/open-telemetry/opentelemetry-operator/issues/1231

Also adds to go.mod
```
replace github.com/prometheus/prometheus => github.com/prometheus/prometheus v0.38.0
```

to use the latest version of prometheus. This should avoid other dependencies bringing in older versions of prometheus, which is missing a relabel method I'm using.


cmd/otel-allocator/prehook/relabel.go
```diff
@@ -82,6 +86,19 @@ func (tf *RelabelConfigTargetFilter) SetConfig(cfgs map[string][]*relabel.Config
...
+		tf.log.V(2).Info("Error Marshaling", "error", err)
+		return
+	}
+
+	byteArr := replaceShard([]byte(out))
+	err = yaml2.Unmarshal(byteArr, &relabelCfgCopy)
+	if err != nil {
+		tf.log.Info("Error Unmarshalling", "error", err)
+		return
+	}
```
secustor, line 101: I'm worried here regarding performance.    
Do you have investigated if it's possible to set the chart in the loop above?

cmd/otel-allocator/prehook/relabel_test.go
```diff
@@ -30,77 +30,134 @@ import (
...
+				Separator:    ";",
+				Modulus:      1,
+				TargetLabel:  "tmp",
+				Action:       "hashmod",
+				Replacement:  "$1",
+			},
+
+			{
+				SourceLabels: model.LabelNames{"tmp"},
+				Regex:        relabel.MustNewRegexp("$(SHARD)"),
```
kristinapathak, line 152: Is `$(SHARD)` only expected in the Regex or it can be in other fields as well?

cmd/otel-allocator/prehook/relabel.go
```diff
@@ -15,10 +15,14 @@
 package prehook
 
 import (
+	"bytes"
+
 	"github.com/go-logr/logr"
+
```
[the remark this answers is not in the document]
moh-osman3: Fixed!

cmd/otel-allocator/prehook/relabel.go
moh-osman3: Thanks for the review! I took your suggestion and moved this logic inside the loop.

cmd/otel-allocator/prehook/relabel_test.go
moh-osman3: Thanks for the review! AFAIK, `$(SHARD)` is used by Prometheus in the generated config when using the `hashmod` action.  In this case I've only seen it as a Regex field.  Unsure if there would ever be a case where it would show up in a label. Just in case I added it to the `SourceLabels` as another confirmation that substitution is working as expected.

cmd/otel-allocator/prehook/relabel.go
```diff
@@ -80,11 +83,30 @@ func (tf *RelabelConfigTargetFilter) Apply(targets map[string]*target.Item) map[
...
+	byteArr := replaceShard([]byte(out))
+	err = yaml2.Unmarshal(byteArr, &outCfg)
+	if err != nil {
+		tf.log.Info("Error Unmarshalling", "error", err)
+		return cfg
+	}
+
+	return outCfg
+}
+
```
jaronoff97, line 109: rather than marshaling and then unmarshaling, it's probably more efficient to just iterate through the relabel config list and do the replacement.

cmd/otel-allocator/prehook/relabel.go
```diff
@@ -93,6 +115,14 @@ func (tf *RelabelConfigTargetFilter) GetConfig() map[string][]*relabel.Config {
 	return relabelCfgCopy
 }
 
+func replaceShard(body []byte) []byte {
+	// See this thread [https://github.com/open-telemetry/opentelemetry-operator/pull/1124/files#r983145795]
+	// for why SHARD == 0 is a necessary substitution. Otherwise the keep action that uses this env variable,
+	// would not match the regex and all targets end up dropped.
+	shard := "0"
```
jaronoff97, line 122: nit: also should include that shard will always be 0 here, and it doesn't make sense to read the SHARD from the environment.

cmd/otel-allocator/prehook/relabel.go
moh-osman3: Thanks for the review! Changed according to this suggestion.

cmd/otel-allocator/prehook/relabel.go
moh-osman3: Added!

secustor (approved):

pavolloffay:
@jaronoff97 can this be merged?

jaronoff97 (approved):
@pavolloffay once the conflicts are fixed, this should be good to go! (cc @moh-osman3)

pavolloffay (approved):

PR #1233, opened 2022-11-08
Ignore reconcile errors that occur because a pod is being terminated

### LINKED ISSUE #1219: Log errors are created if we try to reconcile while a namespace is being deleted
This may only apply to kuttl tests, but sometimes we generate a lot of errors (like the one below) in the logs if the reconciler runs while a namespace is being deleted.   It would be nice to not produce these

```
{"level":"error","ts":"2022-11-03T16:14:56.603485502Z","logger":"controllers.OpenTelemetryCollector","msg":"failed to reconcile config maps","error":"failed to reconcile the expected configmaps: failed to create: configmaps \"simplest-collector\" is forbidden: unable to create new content in namespace kuttl-test-glorious-caribou because it is being terminated","stacktrace":"github.com/open-telemetry/opentelemetry-operator/controllers.(*OpenTelemetryCollectorReconciler).RunTasks\n\t/workspace/controllers/opentelemetrycollector_controller.go:171\ngithub.com/open-telemetry/opentelemetry-operator/controllers.(*OpenTelemetryCollectorReconciler).Reconcile\n\t/workspace/controllers/opentelemetrycollector_controller.go:160\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/internal/controller/controller.go:121\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/internal/controller/controller.go:320\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/internal/controller/controller.go:273\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/internal/controller/controller.go:234"}

```

kevinearls:

This fixes #1219 by not logging errors that occur because we are trying to reconcile something in a namespace that is being terminated.


controllers/opentelemetrycollector_controller.go
```diff
@@ -168,9 +169,12 @@ func (r *OpenTelemetryCollectorReconciler) Reconcile(ctx context.Context, req ct
 func (r *OpenTelemetryCollectorReconciler) RunTasks(ctx context.Context, params reconcile.Params) error {
 	for _, task := range r.tasks {
 		if err := task.Do(ctx, params); err != nil {
-			r.log.Error(err, fmt.Sprintf("failed to reconcile %s", task.Name))
```
frzifus, line 171: How often does this logline occur when a pod is terminated?

kevinearls: Many. :-).  Like for everything the reconciler touches.  To be honest, it's not that big a deal, and probably something that will only be seen when running kuttle tests.  But I found it very annoying, so I submitted the PR.

frzifus: Would it make sense to end the reconciling with by returing `nil` if the reason is a terminating pod? That way the user gets informed that something did not work and we would get rid of all the unnecessary log lines?

controllers/opentelemetrycollector_controller.go
```diff
@@ -168,6 +169,10 @@ func (r *OpenTelemetryCollectorReconciler) Reconcile(ctx context.Context, req ct
 func (r *OpenTelemetryCollectorReconciler) RunTasks(ctx context.Context, params reconcile.Params) error {
 	for _, task := range r.tasks {
 		if err := task.Do(ctx, params); err != nil {
+			// If we get an error that occur because a pod is being terminated then exit this loop
+			if apierrors.IsForbidden(err) && strings.Contains(err.Error(), "because it is being terminated") {
```
frzifus, line 173: one last thing, should we print the termination logline before? And maybe an extra hint, that we cancel the reconciliation?

kevinearls: @frzifus what log level should I print that at?

frzifus: i thought maybe:
```go
r.log.Error(err, fmt.Sprintf("failed to reconcile %s", task.Name))
```
And maybe an additional line that says that the reconciliation will be canceled?

frzifus (approved):
up to the others

controllers/opentelemetrycollector_controller.go
pavolloffay: I am a bit skeptical about checking the string `because it is being terminated"`. This could easily change in the future

damemi: You might be able to check for `v1.NamespaceTerminatingCause` in the error: https://github.com/kubernetes/apiserver/blob/v0.26.0-rc.0/pkg/admission/plugin/namespace/lifecycle/admission.go#L177

kevinearls: @damemi Thanks!

controllers/opentelemetrycollector_controller.go
```diff
@@ -168,6 +168,11 @@ func (r *OpenTelemetryCollectorReconciler) Reconcile(ctx context.Context, req ct
 func (r *OpenTelemetryCollectorReconciler) RunTasks(ctx context.Context, params reconcile.Params) error {
 	for _, task := range r.tasks {
 		if err := task.Do(ctx, params); err != nil {
+			// If we get an error that occurs because a pod is being terminated, then exit this loop
+			if apierrors.IsForbidden(err) && apierrors.HasStatusCause(err, corev1.NamespaceTerminatingCause) {
+				r.log.V(2).Info("Exiting reconcile loop because namespace", params.Instance.Namespace, "is being terminated")
```
damemi, line 173:
```suggestion
				r.log.V(2).Info("Exiting reconcile loop because namespace is being terminated", "namespace", params.Instance.Namespace)
```
(see https://pkg.go.dev/github.com/go-logr/logr#Logger.Info)

secustor (approved):
Linking issue https://github.com/open-telemetry/opentelemetry-operator/issues/1291

Issue #1028, opened 2022-08-10
[target-allocator] targets assigned to old pod after HPA scaled down


moh-osman3:
Observed an issue while load testing with the HPA created from the collector CRD. 

Context:

in my collector spec:

```
spec:
  mode: {{ .Values.collector.mode }}
  image: {{ .Values.collector.image }}
  minReplicas: 1
  maxReplicas: 20
  targetAllocator:
    enabled: true
    image: ghcr.io/open-telemetry/opentelemetry-operator/target-allocator:latest
    serviceAccount: {{ .Release.Name }}-collector-targetallocator
    prometheusCR:
      enabled: false
```

While load testing, the HPA scales the collector statefulset up to 12 pods.  After lowering metric workload for testing, HPA scales back down to a single collector pod in the statefulset.  

# What I expected:
I expect the target allocator to assign targets to only remaining pod after scale down

# What actually happened:
collector-0 does not have targets assigned in the TA, while collector-11 has the target I expected.  Collector-11 was terminated and therefore should not have any targets.  

```
$ kubectl get po -n opentelemetry
NAME                                                         READY   STATUS    RESTARTS   AGE
curl-moh                                                     1/1     Running   0          115m
lightstep-collector-collector-0                              1/1     Running   0          57m
lightstep-collector-targetallocator-b6865b5bb-dc4w5          1/1     Running   0          113m
opentelemetry-operator-controller-manager-575cdcbc57-4d24t   2/2     Running   0          11h

[root@curl-moh:/]$ curl http://lightstep-collector-targetallocator:80/jobs/serviceMonitor%2Favalanche%2Favalanche%2F0/targets?collector_id=lightstep-collector-collector-0
[]

[ root@curl-moh:/ ]$ curl http://lightstep-collector-targetallocator:80/jobs/serviceMonitor%2Favalanche%2Favalanche%2F0/targets?collector_id=lightstep-collector-collector-11
[
  {
    "targets": [
      "10.0.7.184:9001"
    ],
    "labels": {...}
  }
]
```

[...]

PR #1237, opened 2022-11-09
[target-allocator] restart pod watcher when no event is found

moh-osman3:
Resolves https://github.com/open-telemetry/opentelemetry-operator/issues/1028

Before fix:
```
{"level":"info","ts":1668017421.9970064,"msg":"Starting the Target Allocator"}
{"level":"info","ts":1668017421.9972935,"logger":"allocator","msg":"Unrecognized filter strategy; filtering disabled"}
{"level":"info","ts":1668017422.0179584,"logger":"setup","msg":"Starting server..."}
{"level":"info","ts":1668017422.0202508,"logger":"allocator","msg":"Successfully started a collector pod watcher","component":"opentelemetry-targetallocator"}
{"level":"info","ts":1665127338.1837273,"logger":"allocator","msg":"false","component":"opentelemetry-targetallocator"}
{"level":"info","ts":1665127338.1838017,"logger":"allocator","msg":"Collector pod watch event stopped no event","component":"opentelemetry-targetallocator"}
```

This meant pod watcher is no longer running and the TA assigns remaining targets to pods that were already terminated

After fix:

```
{"level":"info","ts":1668017421.9970064,"msg":"Starting the Target Allocator"}
{"level":"info","ts":1668017421.9972935,"logger":"allocator","msg":"Unrecognized filter strategy; filtering disabled"}
{"level":"info","ts":1668017422.0179584,"logger":"setup","msg":"Starting server..."}
{"level":"info","ts":1668017422.0202508,"logger":"allocator","msg":"Successfully started a collector pod watcher","component":"opentelemetry-targetallocator"}
{"level":"info","ts":1668020290.9531963,"logger":"allocator","msg":"No event found. Restarting watch routine","component":"opentelemetry-targetallocator"}
{"level":"info","ts":1668020290.9575274,"logger":"allocator","msg":"Successfully started a collector pod watcher","component":"opentelemetry-targetallocator"}
{"level":"info","ts":1668023632.289222,"logger":"allocator","msg":"No event found. Restarting watch routine","component":"opentelemetry-targetallocator"}
{"level":"info","ts":1668023632.292879,"logger":"allocator","msg":"Successfully started a collector pod watcher","component":"opentelemetry-targetallocator"}
```

pod watcher gets restarted until stabilization occurs -> confirmed TA is assigning targets to correct remaining pods

### DIFF
--- a/cmd/otel-allocator/collector/collector.go
+++ b/cmd/otel-allocator/collector/collector.go
@@ -48,6 +48,7 @@ type Client struct {
 	log       logr.Logger
 	k8sClient kubernetes.Interface
 	close     chan struct{}
+	timeoutSeconds int64
 }
 
 func NewClient(logger logr.Logger, kubeConfig *rest.Config) (*Client, error) {
@@ -60,6 +61,7 @@ func NewClient(logger logr.Logger, kubeConfig *rest.Config) (*Client, error) {
 		log:       logger,
 		k8sClient: clientset,
 		close:     make(chan struct{}),
+		timeoutSeconds: int64(watcherTimeout / time.Second),
 	}, nil
 }
 
@@ -67,8 +69,11 @@ func (k *Client) Watch(ctx context.Context, labelMap map[string]string, fn func(
 	collectorMap := map[string]*allocation.Collector{}
 	log := k.log.WithValues("component", "opentelemetry-targetallocator")
 
+	// convert watcherTimeout to an integer in seconds
+	interval := int64(watcherTimeout / time.Second)
 	opts := metav1.ListOptions{
 		LabelSelector: labels.SelectorFromSet(labelMap).String(),
+		TimeoutSeconds: &interval,
 	}
 	pods, err := k.k8sClient.CoreV1().Pods(ns).List(ctx, opts)
 	if err != nil {
@@ -126,7 +131,7 @@ func runWatch(ctx context.Context, k *Client, c <-chan watch.Event, collectorMap
 				delete(collectorMap, pod.Name)
 			}
 			fn(collectorMap)
-		case <-time.After(watcherTimeout):
+		case <-time.After(time.Duration(k.timeoutSeconds) * time.Second):
 			log.Info("Restarting watch routine")
 			return ""
 		}
--- a/cmd/otel-allocator/collector/collector.go
+++ b/cmd/otel-allocator/collector/collector.go
@@ -85,21 +85,31 @@ func (k *Client) Watch(ctx context.Context, labelMap map[string]string, fn func(
 	fn(collectorMap)
 
 	for {
-		// add timeout to the context before calling Watch
-		ctx, cancel := context.WithTimeout(ctx, watcherTimeout)
-		defer cancel()
-		watcher, err := k.k8sClient.CoreV1().Pods(ns).Watch(ctx, opts)
-		if err != nil {
-			log.Error(err, "unable to create collector pod watcher")
-			return
-		}
-		log.Info("Successfully started a collector pod watcher")
-		if msg := runWatch(ctx, k, watcher.ResultChan(), collectorMap, fn); msg != "" {
-			log.Info("Collector pod watch event stopped " + msg)
-			return
+		if !k.restartWatch(ctx, opts, collectorMap, fn) {
+			return nil
 		}
 	}
 }
+
+func (k *Client) restartWatch(ctx context.Context, opts metav1.ListOptions, collectorMap map[string]*allocation.Collector, fn func(collectors map[string]*allocation.Collector)) bool { 
+	log := k.log.WithValues("component", "opentelemetry-targetallocator")
+	// add timeout to the context before calling Watch
+	ctx, cancel := context.WithTimeout(ctx, watcherTimeout)
+	defer cancel()
+	watcher, err := k.k8sClient.CoreV1().Pods(ns).Watch(ctx, opts)
+	if err != nil {
+		log.Error(err, "unable to create collector pod watcher")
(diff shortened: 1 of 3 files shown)


cmd/otel-allocator/collector/collector.go
```diff
@@ -112,14 +111,14 @@ func runWatch(ctx context.Context, k *Client, c <-chan watch.Event, collectorMap
 			return "context done"
 		case event, ok := <-c:
 			if !ok {
```
kristinapathak, line 113: `!ok` means the channel is closed and empty so I think restarting the Watch when this happens is definitely what we want.  I saw below (L131) that we have wait for our timeout with `time.After(watcherTimeout)` but I'm pretty sure we can add a timeout `opt` on the `Watch()` call and then consider `!ok` to be the Watch ending / timing out.

moh-osman3: Thanks for the feedback! I tried taking this suggestion but for some reason `runWatch` just hangs when I try to add the timeout to the [ListOptions](https://github.com/open-telemetry/opentelemetry-operator/pull/1237/files#diff-9c164a07302845ef1a30b0d2893d22d8fb254f69f8e868e957f35c587baa462bR56), so my unit test just hangs.  Currently my unit test for timeout works only if I include this [block](https://github.com/open-telemetry/opentelemetry-operator/blob/main/cmd/otel-allocator/collector/collector.go#L132) in runWatch.  Hmm I'm wondering if this timeout is only relevant in the specific case that `k8sClient.CoreV1().Pods(ns).List(ctx, opts)`  hangs, but I also just might be using this incorrectly in my test

cmd/otel-allocator/collector/collector.go
```diff
@@ -87,6 +86,9 @@ func (k *Client) Watch(ctx context.Context, labelMap map[string]string, fn func(
 
 	go func() {
 		for {
+			// add timeout to the context before calling Watch
+			ctx, cancel := context.WithTimeout(ctx, watcherTimeout)
+			defer cancel()
```
Aneurysm9, line 91: Is this going to accumulate deferred executions for every iteration through this infinite loop?  Do we need to defer execution here?  Can it simply be not executed or can it be executed without deferral at the end of each loop iteration?

cmd/otel-allocator/collector/collector.go
kristinapathak: This is solved with the context timeout above. 👍

cmd/otel-allocator/collector/collector_test.go
```diff
@@ -165,3 +165,53 @@ func Test_runWatch(t *testing.T) {
...
+		description    string
+		isCloseChannel bool
+	}{
+		{
+			// event is triggered by channel closing.
+			description:    "close_channel",
+			isCloseChannel: true,
+		},
+		{
+			// event triggered by timeout.
```
kristinapathak, line 181: Does this test take 10s?

kristinapathak: I think you can make the timeout value set by each test case, and then send 0 and cause an automatic deadline exceeded error:
https://go.dev/play/p/voG92lUu-A3

cmd/otel-allocator/collector/collector.go
moh-osman3: Thanks for the review! Hmm I think if the operation returns before the timeout (no timeout triggered) then a call to cancel() is needed for cleanup.  I agree I shouldn't pile up a bunch of defer statements and should call `cancel()` as soon as possible.  If I add cancel() to the end of for loop and before the returns, I think there's still a chance if a panic occurs `cancel()` won't be called without `defer`.

To solve this I wrapped this whole block into its own method called `restartWatch` and that way `defer` can be called properly. Let me know if you have any thoughts on this!

cmd/otel-allocator/collector/collector_test.go
moh-osman3: Yes! I changed the timeout to 5s so the overall the unit tests in this folder takes roughly `5.638s`. Is this still too long? Should I go for 1s or 2s?

moh-osman3: Thanks for this suggestion!! I added a timeout for each test case.  Now the tests run in roughly `0.668s`

cmd/otel-allocator/collector/collector.go
```diff
@@ -87,20 +86,32 @@ func (k *Client) Watch(ctx context.Context, labelMap map[string]string, fn func(
...
-				log.Info("Collector pod watch event stopped " + msg)
+			if !k.restartWatch(ctx, opts, collectorMap, fn) {
 				return
 			}
 		}
 	}()
 }
 
+func (k *Client) restartWatch(ctx context.Context, opts metav1.ListOptions, collectorMap map[string]*allocation.Collector, fn func(collectors map[string]*allocation.Collector)) bool {
+	log := k.log.WithValues("component", "opentelemetry-targetallocator")
```
kristinapathak, line 97: Can we move this to the `NewClient()` function and save the `k.log` as this one? I don't think we need to do this so many times 😅

kristinapathak (approved):
Other than the logging change (which can also be done in a separate PR if we want), this looks awesome!

cmd/otel-allocator/collector/collector.go
moh-osman3: Thanks for pointing this out! Moved this up to `NewClient()` as you suggested!

Aneurysm9 (approved):

pavolloffay (approved):

PR #1240, opened 2022-11-10
Allow setting Collector config through ConfigMap

### LINKED ISSUE #1196: Is there a way to provide the config in Instrumentation and OpenTelemetryCollector in the form of a configmap?
Hello, 

I cannot find any way how to pass the config as a configmap rather than the inline `config`
```
config: |
    receivers:
      hostmetrics:
        collection_interval: 10s
        scrapers:
          cpu:
          load:
          memory:
...
```

Applies for both `Instrumentation` and `OpenTelemetryCollector`.

.. or am I just missing something?

Thx

damemi:
Fixes #1196

this adds a `spec.ConfigMap` field to the OpenTelemetryCollector CRD that lets users specify the name and key of a configmap that contains the collector config (rather than needing to use the string literal that currently exists). The configmap must be in the same namespace as the collector instance.

Internally, this works by the following:
1. enforce that only 1 of `spec.Config` or `spec.ConfigMap` is set at admission
2. at the top of the reconcile loop/webhooks, set the `instance.Spec.Config` string to the ConfigMap value (if it exists) or just use the string literal (if that is set instead)
3. Continue on the rest of the logic as usual

I tried a couple different approaches for this, and I think this is overall the best one. The next alternative I had tried to avoid mutating the instance in memory by passing a new param with the parsed config value to each function, and using that instead. This ended up needing a lot of changes, and looked confusing since `instance.Spec.Config` was still accessible in many cases. Ultimately, mutating the object in memory should be fine since the operator shouldn't be writing anything back to the cluster object besides its `status`.

### DIFF
--- a/apis/v1alpha1/opentelemetrycollector_webhook.go
+++ b/apis/v1alpha1/opentelemetrycollector_webhook.go
@@ -127,7 +127,9 @@ func (r *OpenTelemetryCollector) validateCRDSpec() error {
 	}
 
 	// validate Prometheus config for target allocation
-	if r.Spec.TargetAllocator.Enabled {
+	// we can only validate string-literal configs at admission. ConfigMap configs will be validated at runtime
+	// (when we have access to a client)
+	if r.Spec.TargetAllocator.Enabled && len(r.Spec.ConfigMap.Name) == 0 {
 		_, err := ta.ConfigToPromConfig(r.Spec.Config)
 		if err != nil {
 			return fmt.Errorf("the OpenTelemetry Spec Prometheus configuration is incorrect, %s", err)
@@ -183,6 +185,10 @@ func (r *OpenTelemetryCollector) validateCRDSpec() error {
 		return fmt.Errorf("the OptenTelemetry Spec Ingress configuiration is incorrect. Ingress can only be used in combination with the modes: %s, %s, %s",
 			ModeDeployment, ModeDaemonSet, ModeStatefulSet,
 		)
+	}
+
+	if len(r.Spec.Config) > 0 && len(r.Spec.ConfigMap.Name) > 0 {
+		return fmt.Errorf("the OpenTelemetry Collector must have only one of spec.config or spec.configMap.name set")
 	}
 
 	return nil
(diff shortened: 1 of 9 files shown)


apis/v1alpha1/opentelemetrycollector_webhook.go
```diff
@@ -127,7 +127,9 @@ func (r *OpenTelemetryCollector) validateCRDSpec() error {
 	}
 
 	// validate Prometheus config for target allocation
-	if r.Spec.TargetAllocator.Enabled {
+	// we can only validate string-literal configs at admission. ConfigMap configs will be validated at runtime
+	// (when we have access to a client)
+	if r.Spec.TargetAllocator.Enabled && len(r.Spec.ConfigMap.Name) == 0 {
```
damemi, line 132: from what I can tell by the target allocator code, `configToPromConfig` is called at runtime and processes its errors there. So when we write the configmap value to `spec.Config` at the top of the reconcile loop, it should get checked eventually.

pavolloffay:
I don't necessarily agree that we should have two ways to provide collector config. There are also other corner cases to solve:

* which configuration has higher precedence?
* is the instance updated on the configmap change? 
* the update, and deriving service ports and other functionality will have to consider the config map as well

damemi:
> * which configuration has higher precedence?

I took the stance that only 1 can be set at a time, otherwise it is a validation failure. Open to other options as well.

> * is the instance updated on the configmap change?
> * the update, and deriving service ports and other functionality will have to consider the config map as well

Yes to both of these, it should be the same as the current flow with the configmap string. I think I'll have to update the PR, but it should ultimately just be an additional watch for configmaps that triggers the normal reconcile flow once the configmap has been parsed.

bom-d-van:
Hi, we have an use case to support this feature (or introducing a better alternative).

For our case, we have an `otlp/elastic` exporter config in collector, and we have to include an API token for exporting traces to APM-Server. The current config approach means that we would have to include the token the kubernets yaml file or helm value file instead of using k8s secret or other mechanism (we currently keep encrypted secrets in a separate repo).

To make it harder, opentelemetry-operator forbids `--config` [argument override](https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/collector/container.go#L55-L65).

If we prefer not to have two ways for specifying collector config, should we allow config argument override to bypass the issue through volumes and volumeMounts. Or supporting init-container to allow users to make some magic pre-processing?

radsto:
Exactly same issue here and really just don't understand why you have forbidden to override --config arg, actually you can pass multiple config files as args and the map structure of it forbids also multiple times the same argument...

jaronoff97:
@damemi I think this is much easier after [this PR](https://github.com/open-telemetry/opentelemetry-operator/pull/1989/files) which simplifies configmap mounting. I think this is something we should investigate/simplify for v2 of our CRDs.

PR #1242, opened 2022-11-10
Update target allocator version

Aneurysm9:


frzifus (approved):

versions.txt
```diff
@@ -8,7 +8,7 @@ opentelemetry-collector=0.63.1
 operator=0.63.1
 
 # Represents the current release of the Target Allocator.
-targetallocator=0.60.0
+targetallocator=0.63.1
```
pavolloffay, line 11: What is the point of updating this? 

The operator 0.63.1 is already out and it uses TA 0.60.0. I would prefer to update TA image on the release PR.

Also note that this change will not publish TA image https://github.com/open-telemetry/opentelemetry-operator/blob/main/.github/workflows/publish-target-allocator.yaml#L6

Aneurysm9: First, it makes the version align with the other versions in this file such that if someone comes along and types `:%s/0.63.1/0.64.0/` the Right Thing™ happens.

Second, there have been changes to the Target Allocator since [the last change here](https://github.com/open-telemetry/opentelemetry-operator/commit/af44c528c04ab2c972a5f87208993da7089ad1f0).  It will need a new version for its next release and, as we have seen, it may be forgotten.

pavolloffay:
The version will be changed in release PRs e.g.  https://github.com/open-telemetry/opentelemetry-operator/pull/1253

PR #1244, opened 2022-11-11
add secure ciphersuites for TLS config

### LINKED ISSUE #1238: Add default cipher suites configuration for webhook server
Currently the webhook server default cipher suite from crypto/tls package.
There are 2 cipher is vulnerable on SWEET32 attack
TLS_ECDHE_RSA_WITH_3DES_EDE_CBC_SHA, TLS_RSA_WITH_3DES_EDE_CBC_SHA

Since webhook server is default to TLSv1.2, we can add configure to default disable cipher suite of the above, and allow the strong cipher suite.

image
The one in yellow is medium strength cipher suite
https://www.openssl.org/blog/blog/2016/08/24/sweet32/

kangsheng89:
to fix #1238


kangsheng89:
image
After fix, both DES and 3DES cipher no longer acceptable

main.go
```diff
@@ -292,3 +293,12 @@ func addDependencies(_ context.Context, mgr ctrl.Manager, cfg config.Config, v v
 func minTlsDefault(cfg *tls.Config) {
 	cfg.MinVersion = defaultMinTLSVersion
 }
+
+func secureCipherSuite(cfg *tls.Config) {
```
pavolloffay, line 297: Could you please add a comment on this function and describe what it does? 

I personally don't know why all ciphers have to be specified.

kangsheng89: By description from the comments, it should be taken in safe cipher. But after i run the test, seems like manager just directly using the cipher configuration by tls version only. Instead of using the ciphers from tls.CipherSuites() which is safe from vulnerability.

kangsheng89: @pavolloffay do u have any more concerned, or may b do u have any more ideas to improve on this?

kangsheng89: Ok let me update the comments on the code and it will be more clear for reader/maintainer

kangsheng89:
@pavolloffay can u help to check why docker buildx fail?

frzifus:
@kangsheng89 Should we consider the same here? https://github.com/jaegertracing/jaeger-operator/pull/2119#discussion_r1023739130

kangsheng89:
After some consideration, i make changes so that TLS setting for version and ciphersuite is configurable from cmd argument. I am using https://github.com/kubernetes/component-base/blob/master/cli/flag/ciphersuites_flag.go help to validate the TLS settings

main.go
```diff
@@ -103,6 +106,8 @@ func main() {
 	pflag.StringVar(&autoInstrumentationDotNet, "auto-instrumentation-dotnet-image", fmt.Sprintf("ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-dotnet:%s", v.AutoInstrumentationDotNet), "The default OpenTelemetry DotNet instrumentation image. This image is used when no image is specified in the CustomResource.")
 	pflag.StringArrayVar(&labelsFilter, "labels", []string{}, "Labels to filter away from propagating onto deploys")
 	pflag.IntVar(&webhookPort, "webhook-port", 9443, "The port the webhook endpoint binds to.")
+	pflag.StringVar(&tlsOpt.minVersion, "tls-min-version", "VersionTLS12", "Minimum TLS version supported. Value must match version names from https://golang.org/pkg/crypto/tls/#pkg-constants.")
```
pavolloffay, line 109: is it guaranteed that `VersionTLS12` is always present?

main.go
```diff
@@ -289,6 +294,16 @@ func addDependencies(_ context.Context, mgr ctrl.Manager, cfg config.Config, v v
 	return nil
 }
 
-func minTlsDefault(cfg *tls.Config) {
-	cfg.MinVersion = defaultMinTLSVersion
+func tlsConfigSetting(cfg *tls.Config, tlsOpt tlsConfig) {
```
pavolloffay, line 297: The comment would be appreciated why this functionality is here (not only what it does)

kangsheng89: OK i will add in!! hehe

kangsheng89: @pavolloffay added now

main.go
kangsheng89: This add the option that default will be set to tlsv1.2, and if user wanna choose other than v1.2, they can do it from the command arg. So we have the flexibility and in case for upgrading, they able to do it from command argument as well.

kangsheng89: Currently in Jaeger-operator, I m using the same method as well with suggestions from @frzifus. He been helping me on review and give me the idea on configurable tls and cipher settings

kangsheng89:
@pavolloffay any concerns or feedbacks to improve this?

pavolloffay (approved):

PR #1246, opened 2022-11-13
Only create ServiceAccounts if existing ServiceAccount is not specified

### LINKED ISSUE #1224: Do not auto-create service account if an existing one is specified
**Problem:**
When ` serviceAccount` is set on an OpenTelemetryCollector, a service account based on the collector name still gets created even though it's unused. The extra serviceAccount is just noise and should not be created. From looking at the code, it appears a targetAllocator serviceAccount will also be auto-created by the operator when an existing serviceAccount is specified.

**Expected:**
No service accounts are auto-created when an existing serviceAccount is specified

**Actual:**
Service accounts are auto-created even if an existing one is specified

**To Reproduce:**
Create a service account, in my case I made an `opentelemetry-collector` SA
Apply a collector with a serviceAccount, I used the `simplest` collector below as a test

```
kind: OpenTelemetryCollector
metadata:
  name: simplest
  namespace: opentelemetry
spec:
  mode: deployment
  config: |
    receivers:
      otlp:
        protocols:
          grpc:
          http:
    processors:
      memory_limiter:
        check_interval: 1s
        limit_percentage: 75
        spike_limit_percentage: 15
      batch:
        send_batch_size: 10000
        timeout: 10s

    exporters:
      logging:

    service:
      pipelines:
        traces:
          receivers: [otlp]
          processors: []
          exporters: [logging]
      telemetry:
        logs:
          level: info
          initial_fields:
            app: otel-collector
        metrics:
          level: detailed
          address: 0.0.0.0:8888

  resources:
    limits:
      memory: 500Mi
    requests:
      memory: 200Mi
      cpu: 200m
  serviceAccount: opentelemetry-collector
```

csquire:
Fixes #1224

### DIFF
--- a/bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
@@ -37,7 +37,7 @@ metadata:
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
     repository: github.com/open-telemetry/opentelemetry-operator
     support: OpenTelemetry Community
-  name: opentelemetry-operator.v0.63.1
+  name: opentelemetry-operator.v0.63.1-12-ge652e05
   namespace: placeholder
 spec:
   apiservicedefinitions: {}
@@ -294,7 +294,7 @@ spec:
                 - --enable-leader-election
                 - --zap-log-level=info
                 - --zap-time-encoding=rfc3339nano
-                image: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator:0.63.1
+                image: ghcr.io/craig.squire/opentelemetry-operator/opentelemetry-operator:0.63.1-12-ge652e05
                 livenessProbe:
                   httpGet:
                     path: /healthz
@@ -401,7 +401,7 @@ spec:
   maturity: alpha
   provider:
     name: OpenTelemetry Community
-  version: 0.63.1
+  version: 0.63.1-12-ge652e05
   webhookdefinitions:
   - admissionReviewVersions:
     - v1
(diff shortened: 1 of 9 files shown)


yuriolisa:
@csquire, could you please add an example on README using the service account?

pavolloffay:
@csquire could you please as well add a note to https://github.com/open-telemetry/opentelemetry-operator/blob/e8cf7e3f58300ab9b78def742431033177759e23/apis/v1alpha1/opentelemetrycollector_types.go#L87 and mention that if the SA is not provided the operator creates one.

> @csquire, could you please add an example on README using the service account?

I would argue that this might be a corner use-case not needed for the readme that more focuses on quick-start guides.

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -294,7 +294,7 @@ spec:
                 - --enable-leader-election
                 - --zap-log-level=info
                 - --zap-time-encoding=rfc3339nano
-                image: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator:0.63.1
+                image: ghcr.io/craig.squire/opentelemetry-operator/opentelemetry-operator:0.63.1-12-ge652e05
```
frzifus, line 297:
```suggestion
                image: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator:0.63.1
```

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -401,7 +401,7 @@ spec:
   maturity: alpha
   provider:
     name: OpenTelemetry Community
-  version: 0.63.1
+  version: 0.63.1-12-ge652e05
```
frzifus, line 404:
```suggestion
  version: 0.63.1
```

csquire: oops, thanks!

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
csquire: thank you!

pavolloffay (approved):

jaronoff97 (approved):

PR #1252, opened 2022-11-15
Add a smoke test that is run as part of the scorecard tests

kevinearls:


kevinearls:
@iblancasa Can you take a look at this?  Obviously I've got something wrong with the naming.

iblancasa:
The problem (I think) is that some commands are run in the `kuttl-test.yaml` file. Those commands are not available in the KUTTL image, that makes the test fail, no report is generated and the `scorecard` parser is not able to find it.

The changes introduced in #1254 should help to fix this.

atoulme:

PR #1256, opened 2022-11-16
Add new selector for pod and service monitor

### LINKED ISSUE #1255: [target-allocator] Allow the target allocator to use a label selector for Prometheus CRDs
Right now the target allocator pulls in every prometheus CRD (servicemonitor, podmonitor, etc.) in a cluster. This may not be desired for users that want to split up scraping among pools of collectors.

jaronoff97:
This PR resolves #1255 by adding in a pod and service monitor selector. I needed to change the signature of a few functions, but this is an otherwise minor change.

example:
Screen Shot 2022-11-16 at 5 22 54 PM
the dip here is when i enabled a `release: test` label selector for servicemonitors, and it comes back when i added the label back. The default behavior is the same as what it is now.


Aneurysm9 (approved):

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -177,6 +177,10 @@ type OpenTelemetryTargetAllocator struct {
 	// All CR instances which the ServiceAccount has access to will be retrieved. This includes other namespaces.
 	// +optional
 	PrometheusCR OpenTelemetryTargetAllocatorPrometheusCR `json:"prometheusCR,omitempty"`
+	// PodMonitors to be selected for target discovery.
+	PodMonitorSelector *map[string]string `json:"podMonitorSelector,omitempty"`
```
secustor, line 181: Wouldn't `OpenTelemetryTargetAllocatorPrometheusCR` be a better fit than `OpenTelemetryTargetAllocator` as the former is used to enable the CRD functionality?

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -177,6 +177,10 @@ type OpenTelemetryTargetAllocator struct {
...
+	// ServiceMonitors to be selected for target discovery.
```
secustor, line 182: We should reference what the options are for the fields e.g. the link to the API reference or better an user guide. https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#labelselector-v1-meta

cmd/otel-allocator/config/config.go
```diff
@@ -39,10 +39,12 @@ const DefaultResyncTime = 5 * time.Minute
...
-	LabelSelector      map[string]string  `yaml:"label_selector,omitempty"`
-	Config             *promconfig.Config `yaml:"config"`
-	AllocationStrategy *string            `yaml:"allocation_strategy,omitempty"`
-	FilterStrategy     *string            `yaml:"filter_strategy,omitempty"`
+	LabelSelector          map[string]string  `yaml:"label_selector,omitempty"`
+	Config                 *promconfig.Config `yaml:"config"`
+	AllocationStrategy     *string            `yaml:"allocation_strategy,omitempty"`
+	FilterStrategy         *string            `yaml:"filter_strategy,omitempty"`
+	PodMonitorSelector     *map[string]string `yaml:"pod_monitor_selector,omitempty"`
+	ServiceMonitorSelector *map[string]string `yaml:"service_monitor_selector,omitempty"`
```
kristinapathak, line 47: What's the benefit of making this a pointer to a map instead of `map[string]string`?

apis/v1alpha1/opentelemetrycollector_types.go
jaronoff97: I originally used the metav1.LabelSelector here, but it wasn't serializing & deserializing to YAML correctly, so i did what we do for the collector label selector. I can link these docs though.

cmd/otel-allocator/config/config.go
jaronoff97: I originally did this to make it so that these were optional, but that may not be necessary. Will rework :)

kristinapathak (approved):

secustor (approved):

pavolloffay (approved):

PR #1260, opened 2022-11-17
fix: delayed platform detection

### LINKED ISSUE #1258: [Platform-Autodetect] running after reconciler has been created
While working on #1206, i noticed that the platform detection runs after the otel reconciler has been created. 
```bash
1.668649074508897e+09	INFO	create new reconciler
platform: Unknown # println
1.6686490745111158e+09	DEBUG	config	platform detected	{"platform": "OpenShift"}
1.668649074512458e+09	INFO	setup	starting manager
```
I would like to add [a route reconciler](https://github.com/open-telemetry/opentelemetry-operator/blob/c7c962d187eb1d6470719e347b573c395d9eb0ef/controllers/opentelemetrycollector_controller.go#L113-L115) only if the route api exists. Currently the platform is determined by the existence of this api.

https://github.com/open-telemetry/opentelemetry-operator/blob/750166d77a81139369ed24dca86df2881b125106/pkg/autodetect/main.go#L65-L80

~Can we not reproduce the same logic at startup as below?~
```golang
var operatorPlatform = platform.Unknown

func init() {
	if err := routev1.AddToScheme(scheme); err != nil {
		operatorPlatform = platform.OpenShift
	} else if cfg := ctrl.GetConfigOrDie(); cfg != nil {
		operatorPlatform = platform.Kubernetes
	}
}
```

I mean, we could of course call `cfg.AutoDetect()` before creating the reconciler. However, in case the route api exists, i would like to add it to the route clientset anyway.

---

Auto detect probably starts later because the manager is started after the reconciler is created. 
See: `Add will set requested dependencies on the component, and cause the component to be started when Start is called.`

https://github.com/open-telemetry/opentelemetry-operator/blob/750166d77a81139369ed24dca86df2881b125106/main.go#L241-L245

frzifus:
Closes #1258

### DIFF
--- a/main.go
+++ b/main.go
@@ -145,6 +145,10 @@ func main() {
 		config.WithLabelFilters(labelsFilter),
 	)
 
+	if err := cfg.AutoDetect(); err != nil {
+		setupLog.Error(err, "failed to determine operator platform")
+	}
+
 	watchNamespace, found := os.LookupEnv("WATCH_NAMESPACE")
 	if found {
 		setupLog.Info("watching namespace(s)", "namespaces", watchNamespace)


main.go
```diff
@@ -145,6 +145,10 @@ func main() {
 		config.WithLabelFilters(labelsFilter),
 	)
 
+	if err := cfg.AutoDetect(); err != nil {
```
pavolloffay, line 148: The idea of auto-detect in the OTEL operator is that APIs can be dynamically installed in the cluster (e.g. after the operator was installed). 

Could you please elaborate why the initialization should block until the first auto detect cycle is executed?

pavolloffay:
@frzifus do you still need this PR to be merged? IIRC you have mentioned that you found a different approach by using `before` on the manager.

frzifus:
correct, let me rebase the other pr. I think then its rdy to go.

PR #1265, opened 2022-11-18
[target-allocator] Handle concurrency more gracefully 

jaronoff97:
This PR reorganizes the structure of the target allocator to better handle concurrency issues as well as make our code more testable. This is accomplished through a few tasks:
* moving the server in to its own file with accompanying tests and benchmarks
* moving the target discovery tasks to the target package.
* remove the http files in allocation in favor of server tests
* change the watchers to be created independently
* change the event loop to read events and have the watchers handle configuration reloading

Overall, this PR significantly cleans up the target allocator's codebase and allows us to more effectively isolate and manage individual components.


cmd/otel-allocator/main.go
```diff
@@ -16,306 +16,205 @@ package main
...
+	closer := make(chan bool, 1)
+	signal.Notify(interrupts, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
+
+	runGroup.Add(
+		func() error {
+			fileWatcherErr := fileWatcher.Start(events, errors)
+			setupLog.Info("File watcher exited")
+			return fileWatcherErr
+		},
+		func(err error) {
```
kristinapathak, line 124: None of these functions seem to do anything with the error; can we make them blank identifiers?
```suggestion
		func(_ error) {
```

kristinapathak: That would also allow us to use `err` inside these functions, and fix the problem of logging what I think is the incorrect error below.

cmd/otel-allocator/main.go
```diff
@@ -16,306 +16,205 @@ package main
...
-				if err := srv.Shutdown(ctx); err != nil {
-					setupLog.Error(err, "Cannot shutdown the server")
-				}
-				srv = newServer(log, allocator, discoveryManager, k8sclient, cliConf.ListenAddr)
-				go func() {
-					if err := srv.Start(); !errors.Is(err, http.ErrServerClosed) {
-						setupLog.Error(err, "Can't restart the server")
+		})
+	runGroup.Add(
+		func() error {
```
kristinapathak, line 185: The run function here seems to have multiple responsibilities, but the best summary I can think of is "channel watcher".  I'm not as familiar with the run package so I'm not sure if it's normal to have a "catch all" function or if it would be better to split this up into two functions: one for interrupt logic and one for handler logic.

cmd/otel-allocator/main.go
```diff
@@ -16,306 +16,205 @@ package main
...
+			for {
+				select {
+				case <-interrupts:
+					setupLog.Info("Received interrupt")
+					return nil
+				case <-closer:
+					setupLog.Info("Closing run loop")
```
kristinapathak, line 192: This is the only run group function that doesn't have its "Closing" log in the interrupt/error function - can we change the log in the interrupt function and only return here? (paired suggestion below)
```suggestion
```

cmd/otel-allocator/main.go
```diff
@@ -16,306 +16,205 @@ package main
...
-}
-
-func configureFileDiscovery(ctx context.Context, log logr.Logger, allocator allocation.Allocator, discoveryManager *lbdiscovery.Manager, cliConfig config.CLIConfig) (*collector.Client, error) {
-	cfg, err := config.Load(*cliConfig.ConfigFilePath)
-	if err != nil {
-		return nil, err
-	}
+		},
+		func(err error) {
+			setupLog.Info("Received error, shutting down")
```
kristinapathak, line 212:
```suggestion
			setupLog.Info("Closing run loop")
```

kristinapathak (commented):
This looks really good!

secustor (commented):
I like the direction this is going :)

cmd/otel-allocator/server/server_test.go
```diff
@@ -0,0 +1,249 @@
...
+)
+
+func colIndex(index, numCols int) int {
+	if numCols == 0 {
+		return -1
+	}
+	return index % numCols
+}
+
+func makeNNewTargets(n int, numCollectors int, startingIndex int) map[string]*target.Item {
```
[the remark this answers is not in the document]
jaronoff97: i agree, will do that now

cmd/otel-allocator/main.go
```diff
@@ -16,306 +16,205 @@ package main
...
+		if err != nil {
+			setupLog.Error(err, "Can't start the prometheus watcher")
+			os.Exit(1)
 		}
-	}()
-
-	for {
-		select {
-		case <-interrupts:
+		promWatcherErr := promWatcher.Start(events, errors)
```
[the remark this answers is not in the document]
jaronoff97: It's because of how Prometheus CRD watching is setup – each informer (podmonitor, servicemonitor) is already async so we can't make them match easily unless we just wanna throw a while loop around the promwatcher method.

cmd/otel-allocator/main.go
```diff
@@ -16,306 +16,205 @@ package main
...
-	}()
-
-	// creates a new discovery manager
-	discoveryManager := lbdiscovery.NewManager(ctx, log, gokitlog.NewNopLogger(), allocatorPrehook)
-	defer discoveryManager.Close()
-
-	discoveryManager.Watch(allocator.SetTargets)
-
-	k8sclient, err := configureFileDiscovery(context.Background(), log, allocator, discoveryManager, cliConf)
+	fileWatcher, err = allocatorWatcher.NewFileWatcher(setupLog.WithName("file-watcher"), cliConf)
```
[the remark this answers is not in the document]
jaronoff97: I think this makes sense, but I would say this deserves its own PR rather than adding it in here. I think this would make even more changes than I already am doing, and I think also needs more discussion too.

cmd/otel-allocator/server/server.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	if len(q) == 0 {
+		displayData := GetAllTargetsByJob(s.allocator, jobId)
+		s.jsonHandler(w, displayData)
+
+	} else {
+		tgs := s.allocator.GetTargetsForCollectorAndJob(q[0], jobId)
+		// Displays empty list if nothing matches
+		if len(tgs) == 0 {
```
moh-osman3, line 160: Hmm wondering if L164 would also cover this case?  Isn't tgs already an empty list in this case, so could you display it directly?

cmd/otel-allocator/target/discovery_test.go
```diff
@@ -52,17 +53,27 @@ func TestDiscovery(t *testing.T) {
 			want: []string{"prom.domain:9004", "prom.domain:9005", "promfile.domain:1001", "promfile.domain:3000"},
 		},
 	}
-	manager := NewManager(context.Background(), ctrl.Log.WithName("test"), gokitlog.NewNopLogger(), nil)
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	d := discovery.NewManager(ctx, gokitlog.NewNopLogger())
+	manager := NewDiscoverer(ctrl.Log.WithName("test"), d, nil)
 	defer close(manager.close)
+	defer cancelFunc()
```
moh-osman3, line 60: nit (and out of curiosity): does order of defers matter? and should this be called directly below L56 (could L57 and L58 ever cause a panic where the cancel function would not be called)

cmd/otel-allocator/watcher/file.go
```diff
@@ -17,56 +17,66 @@ package watcher
...
 	fileWatcher, err := fsnotify.NewWatcher()
 	if err != nil {
 		logger.Error(err, "Can't start the watcher")
-		return FileWatcher{}, err
+		return &FileWatcher{}, err
 	}
 
-	return FileWatcher{
+	return &FileWatcher{
+		logger:         logger,
```
moh-osman3, line 43: Would it be helpful to give this logger a name `logger.WithName("allocator-file-watcher")` or is this the callers responsibility?

jaronoff97: In this case its the caller's responsibility

kristinapathak (approved):

pavolloffay (approved):

PR #1269, opened 2022-11-21
adds ingressClassName field to collector spec

### LINKED ISSUE #1268: Support ingressClassName in collector spec
Ingress controller implementations use this field (`ingressClassName`) to know whether they should be serving the Ingress resource.

Before the IngressClass resource and `ingressClassName` field were added in Kubernetes 1.18, Ingress classes were specified with a `kubernetes.io/ingress.class` annotation on the Ingress. This annotation was never formally defined, but was widely supported by Ingress controllers.

The newer `ingressClassName` field on Ingresses is a replacement for that annotation

https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation

avadhut123pisal:
Fixes #1268


frzifus (approved):
thanks @avadhut123pisal, LGTM :)

pavolloffay (approved):

PR #1282, opened 2022-11-27
Remove default claims - fixes #1281

### LINKED ISSUE #1281: OTEL Operator adds un-needed default volume for StatefulSet
If you add a collector that has mode statefulset it will have a default volume added which is not-needed for a simple deployment unless you want to actively store state on a disk.

Here is the test for the functionality in question
https://github.com/open-telemetry/opentelemetry-operator/blob/57abb4df547ca6346bde7bd0cc8c1fe02e4c016e/pkg/collector/volumeclaim_test.go#L30

Here is the code actually adding a PVC if there are none defined by the user (causing deployments where you don't want it to get it)
https://github.com/open-telemetry/opentelemetry-operator/blob/57abb4df547ca6346bde7bd0cc8c1fe02e4c016e/pkg/collector/volumeclaim.go#L37

ekarlso:
Resolves https://github.com/open-telemetry/opentelemetry-operator/issues/1281

### DIFF
--- a/pkg/collector/container.go
+++ b/pkg/collector/container.go
@@ -76,11 +76,6 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
 
 	if len(otelcol.Spec.VolumeMounts) > 0 {
 		volumeMounts = append(volumeMounts, otelcol.Spec.VolumeMounts...)
-	} else if otelcol.Spec.Mode == "statefulset" {
-		volumeMounts = append(volumeMounts, corev1.VolumeMount{
-			Name:      "default-volume",
-			MountPath: "/usr/share/default-volume",
-		})
 	}
 
 	var envVars = otelcol.Spec.Env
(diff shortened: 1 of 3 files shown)


pkg/collector/container.go
```diff
@@ -76,11 +76,6 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
 
 	if len(otelcol.Spec.VolumeMounts) > 0 {
 		volumeMounts = append(volumeMounts, otelcol.Spec.VolumeMounts...)
-	} else if otelcol.Spec.Mode == "statefulset" {
```
pavolloffay, line 79: so the users will have to define the volumes explicitly in the CR? 

I don't think we have even documented this default volume so not sure somebody used it.

pavolloffay (approved):
LGTM, just one question

pkg/collector/container.go
ekarlso: If you need a volume you're need to declare one to have one yes. Basically if not you Will always end up having PVCs / vols required that in some circumstances is not needed by default

pavolloffay:
thanks, merging this

PR #1292, opened 2022-11-30
Add change handler to register callbacks

frzifus:
Description: https://github.com/open-telemetry/opentelemetry-operator/pull/1206#issuecomment-1332316819

### DIFF
--- a/internal/config/change_handler.go
+++ b/internal/config/change_handler.go
@@ -22,17 +22,17 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-// ChangeHandler is implemented by any structure that is able to register callbacks
+// changeHandler is implemented by any structure that is able to register callbacks
 // and call them using one single method.
-type ChangeHandler interface {
+type changeHandler interface {
 	// Change will call every registered callback.
 	Change() error
 	// Register this function as a callback that will be executed when Change() is called.
 	Register(f func() error)
 }
 
-// NewOnChange returns a thread-safe ChangeHandler.
-func NewOnChange() ChangeHandler {
+// newOnChange returns a thread-safe ChangeHandler.
+func newOnChange() changeHandler {
 	return &onChange{
 		logger: logf.Log.WithName("change-handler"),
 	}
--- a/internal/config/change_handler.go
+++ b/internal/config/change_handler.go
@@ -1,0 +1,63 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package config contains the operator's runtime configuration.
+package config
+
+import (
+	"sync"
+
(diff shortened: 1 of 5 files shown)


internal/config/change_handler.go
```diff
@@ -0,0 +1,64 @@
...
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// ChangeHandler is implemented by any structure that is able to register callbacks
+// and call them using one single method.
+type ChangeHandler interface {
+	// Change will call every registered callback.
+	Change() error
+	// Register the given function as a callback that will be executed on Change
+	// is called.
```
kevinearls, line 31: Nitpicking here.  Maybe "Register this function as a callback that will be executed when Change() is called"  or just "Register this function to be executed by Change()"

This might be just me, but I really prefer comments that can fit on one line to those that just barely wrap around to a second line.

internal/config/change_handler.go
```diff
@@ -0,0 +1,64 @@
...
+	Register(f func() error)
+}
+
+// NewOnChange returns thread-safe ChangeHandler.
```
kevinearls, line 35: "NewOnChange returns a thread-safe ChangeHandler"

internal/config/main.go
```diff
@@ -125,12 +125,10 @@ func (c *Config) AutoDetect() error {
...
 
 	if changed {
-		for _, callback := range c.onChange {
-			if err := callback(); err != nil {
-				// we don't fail if the callback failed, as the auto-detection itself
-				// did work
-				c.logger.Error(err, "configuration change notification failed for callback")
-			}
+		if err := c.onChange.Change(); err != nil {
+			// we don't fail if the callback failed, as the auto-detection itself
```
kevinearls, line 129: Again, nitpicking, maybe "Don't fail if the callback failed, as auto-detection itself worked"

kevinearls (approved):
LGTM except for some minor issues on comments

iblancasa (approved):

internal/config/change_handler.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/go-logr/logr"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// ChangeHandler is implemented by any structure that is able to register callbacks
+// and call them using one single method.
+type ChangeHandler interface {
+	// Change will call every registered callback.
+	Change() error
```
pavolloffay, line 29: shall we call it `OnChange`?

internal/config/options.go
```diff
@@ -84,12 +84,17 @@ func WithLogger(logger logr.Logger) Option {
 		o.logger = logger
 	}
 }
+func WithOnChangeHandler(h ChangeHandler) Option {
```
pavolloffay, line 87: Do we need to expose `ChangeHandler`? 

Would it be possible to keep `ChangeHandler` package private and just expose a method to register a callback?

frzifus: Sure it works both.

internal/config/change_handler.go
frzifus: Personally i think the interface name `changeHandler` is more descriptive then just `onChange`.

frzifus: oh i see - its about the `Change` member. :facepalm:

internal/config/main.go
```diff
@@ -44,7 +44,7 @@ type Config struct {
 	targetAllocatorConfigMapEntry  string
 	autoInstrumentationNodeJSImage string
 	autoInstrumentationJavaImage   string
-	onChange                       []func() error
+	onChange                       changeHandler
```
pavolloffay, line 47: shall we rename the `onChange` to `apiAutoDect`?

pavolloffay (approved):

PR #1293, opened 2022-11-30
Memory improvements first pass

### LINKED ISSUE #1257: [target-allocator] Investigate Target Allocator memory usage
@seankhliao reported in the otel-operator channel [in slack](https://cloud-native.slack.com/archives/C033BJ8BASU/p1666986594165909) that the TA was using around 200 gigs of memory for 4k+ pods. This is definitely excessive and makes me think that we either have a memory leak and/or poor memory usage. I'm planning on going through the TA code, running some benchmarks/profiling, and doing some refactoring to improve the performance of the service.

jaronoff97:
Closes #1257 

This PR is the first in a series of two. In this PR, we add new fields to the allocators and servers to cache the values they need to return to prevent excessive allocations on http calls. This also removes the intermediary of the targetGroupJSON and instead just uses the target item with json hints. This also removes the allocations needed in the map diffing. This also fixes a potential bug in the map which would have caused targets to not update when processed. Finally, this has the server keep track of when it needs to recompute the scrape configs response it has to return. 

NOTES!
* When this PR is merged, we will need to make two issues:
  * First one is to change how the diffs are generated to just make lists and not maps
  * The other is look in to a better way to fix the potential (but unlikely) race generated from saving the extra allocation


cmd/otel-allocator/allocation/http.go
```diff
@@ -18,61 +18,24 @@ import (
...
-				}
-			}
-		}
-	}
-	for _, v := range group {
-		tgs = append(tgs, targetGroupJSON{Targets: []string{v.TargetURL}, Labels: labelSet[v.Hash()]})
-	}
-
-	return tgs
+func GetAllTargetsByCollectorAndJob(allocator Allocator, collector string, job string) []*target.Item {
```
jaronoff97, line 39: in a future PR i'm going to remove this file altogether, for now i'm trying to keep changes to a minimum

cmd/otel-allocator/main.go
```diff
@@ -220,17 +224,32 @@ func (s *server) Shutdown(ctx context.Context) error {
...
+	hash, err := hashstructure.Hash(configs, nil)
 	if err != nil {
+		s.logger.Error(err, "failed to hash the config")
 		s.errorHandler(w, err)
+		return
 	}
-	jsonConfig, err := yaml.YAMLToJSON(configBytes)
-	if err != nil {
-		s.errorHandler(w, err)
+	// if the hashes are different, we need to recompute the scrape config
```
jaronoff97, line 234: this logic is similar to what the collector is doing on the receiving end i.e. it should only update this byte array when there's an updated scrape config list

cmd/otel-allocator/target/target.go
```diff
@@ -27,23 +27,25 @@ type LinkJSON struct {
...
+	Link          LinkJSON       `json:"-"`
+	TargetURL     []string       `json:"targets"`
+	Labels        model.LabelSet `json:"labels"`
+	CollectorName string         `json:"-"`
+	hash          string
 }
 
 func (t Item) Hash() string {
-	return t.JobName + t.TargetURL + t.Label.Fingerprint().String()
+	return t.hash
```
jaronoff97, line 39: by precomputing the hash, we're saving a lot of allocations

cmd/otel-allocator/allocation/strategy.go
```diff
@@ -92,11 +92,22 @@ func Register(name string, provider AllocatorProvider) error {
 	return nil
 }
 
+func GetRegisteredAllocatorNames() []string {
+	var names []string
+	for s := range registry {
+		if len(s) > 0 {
```
moh-osman3, line 98: Curious in what case would len(s) == 0? Wondering if this is rare enough that it might be worthwhile to preallocate `names` with the size len(registry) on line 96.

cmd/otel-allocator/allocation/strategy.go
```diff
@@ -92,11 +92,22 @@ func Register(name string, provider AllocatorProvider) error {
...
+	}
+	return names
+}
+
 type Allocator interface {
 	SetCollectors(collectors map[string]*Collector)
 	SetTargets(targets map[string]*target.Item)
 	TargetItems() map[string]*target.Item
 	Collectors() map[string]*Collector
+	GetTargetsForCollectorAndJob(collector string, job string) []*target.Item
```
moh-osman3, line 110: Will this new function be used outside of the allocation package?

secustor (commented):
Looks good to me, only one possible improvement. 

How has this been tested? 
Do you have compared the performance?

cmd/otel-allocator/allocation/strategy.go
jaronoff97: This will potentially be used by the server instead of using the http package in the future.

cmd/otel-allocator/allocation/http.go
```diff
@@ -18,61 +18,24 @@ import (
...
-					Labels:  t.Label,
-				})
-			}
-
-			displayData[j.CollectorName] = collectorJSON{Link: fmt.Sprintf("/jobs/%s/targets?collector_id=%s", url.QueryEscape(j.JobName), j.CollectorName), Jobs: targetGroupList}
-
-		}
+	for _, col := range allocator.Collectors() {
+		items := allocator.GetTargetsForCollectorAndJob(col.Name, job)
+		displayData[col.Name] = collectorJSON{Link: fmt.Sprintf("/jobs/%s/targets?collector_id=%s", url.QueryEscape(job), col.Name), Jobs: items}
```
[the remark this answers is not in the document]
jaronoff97: I'm not sure what you mean by having the query dropped? In this case, the call is only ever going to be made when someone calls the jobs endpoint without a collector set, which the collector will never do, only someone manually hitting that endpoint.

cmd/otel-allocator/allocation/strategy.go
jaronoff97: I actually found some cases in testing where there were empty strings in this list and figured this would be good defensive programming, the more i think of it, the more i think the problem was actually in how i was setting the names. I can remove this.

cmd/otel-allocator/allocation/least_weighted_test.go
```diff
@@ -292,3 +292,16 @@ func TestCollectorBalanceWhenAddingAndRemovingAtRandom(t *testing.T) {
 		assert.InDelta(t, i.NumTargets, count, math.Round(percent))
 	}
 }
+
+func Benchmark_leastWeightedAllocator_SetTargets(b *testing.B) {
```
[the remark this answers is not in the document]
jaronoff97: The benchmark is going to continuously tell us how many allocations p/s we are making to get an understanding of our performance. You can read more about b.N [here](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) but it's supplied by Go.

jaronoff97:
@secustor i've tested this in cluster and it has been running successfully for over a week now. I did some benchmarking before and after for the number of allocations made and the reduction was over 99.9%. I haven't tested this with a cluster at scale, but I can do that before we release this if that would give us more confidence?

cmd/otel-allocator/target/target.go
```diff
@@ -27,23 +27,25 @@ type LinkJSON struct {
 }
 
 type Item struct {
```
kristinapathak, line 29: I'm a big fan of computing the hash once, but I'm worried that the `Item` struct isn't easy to consume by other packages.  It seems easy to do it wrong:
- someone can create an item, filling out all possible values, and have an empty string for a hash.
- someone can create an item with `NewItem()` and later modify the fields, causing the hash to not match the values.

I'm not sure if it's worth it to change this, some things I thought of:
- don't export `Item` struct - I think this makes things more difficult in this case; this works when a struct is meant to implement an interface.
- exporting `hash` (with the `-` json tag) - then consumers can set their own or use the one provided in `NewItem()`. 
- have `Hash()` check if `hash` is empty, and, if so, calculate the hash and set it (requires a pointer receiver).

I think the easiest solution is just to add a comment documenting that an `Item` shouldn't be made directly and `JobName`, `TargetURL`, and `Labels` shouldn't change.

cmd/otel-allocator/target/target.go
```diff
@@ -27,23 +27,25 @@ type LinkJSON struct {
...
-	JobName       string
-	Link          LinkJSON
-	TargetURL     string
-	Label         model.LabelSet
-	CollectorName string
+	JobName       string         `json:"-"`
+	Link          LinkJSON       `json:"-"`
+	TargetURL     []string       `json:"targets"`
+	Labels        model.LabelSet `json:"labels"`
```
kristinapathak, line 33: Is this a breaking change?

kristinapathak (commented):
These look like great improvements!

cmd/otel-allocator/target/target.go
jaronoff97: This matches exactly what we were returning previously from the targetGroupJSON object, except we just return this rather than creating a new object.

jaronoff97: This would break someone who is importing this package, though. I will call that out in the PR description.

jaronoff97: Do you think I should keep the field name as `Label` instead or is it okay to make this change?

kristinapathak: Okay I see now that this is replacing `targetGroupJSON` (https://github.com/open-telemetry/opentelemetry-operator/pull/1293/files#diff-f7acbc836f7ddf2f8c66c36bbf55d831e7dbd492c606bde09215aa3979b000bdL31).  I think this is a good change 👍

cmd/otel-allocator/diff/diff.go
```diff
@@ -31,19 +39,20 @@ func (c Changes[T]) Removals() map[string]T {
 
 // Maps generates Changes for two maps with the same type signature by checking for any removals and then checking for
 // additions.
-func Maps[T any](current, new map[string]T) Changes[T] {
+// TODO: This doesn't need to create maps, it can return slices only. This function doesn't need to insert the values.
+func Maps[T Hasher](current, new map[string]T) Changes[T] {
 	additions := map[string]T{}
 	removals := map[string]T{}
-	// Used as a set to check for removed items
-	newMembership := map[string]bool{}
```
kristinapathak, line 38: nice 😎

kristinapathak (approved):

secustor (approved):
No, I think it is fine if you have running this changes in a while. I'm looking forward to see the second PR.

Issue #1115, opened 2022-09-20
Move MaxReplicas and MinReplicas to AutoscalerSpec in collector CRD


moh-osman3:
In https://github.com/open-telemetry/opentelemetry-operator/pull/1077, an Autoscaler field and object was added to the collector CRD.  `MaxReplicas` and `MinReplicas` are fields specific to autoscaling so that should be moved to the AutoscalerSpec, and the HPA source code should be refactored to reflect this.

pavolloffay:
cc) @kevinearls

pavolloffay:
Also note that we can't remove fields from a published CRD. If we want to move things around we would have to do deprecation and support the deprecated options as well.

PR #1302, opened 2022-12-06
[HPA] Move `maxReplicas` and `minReplicas` to `.spec.Autoscaler`

moh-osman3:
Resolves https://github.com/open-telemetry/opentelemetry-operator/issues/1115


pkg/collector/horizontalpodautoscaler.go
```diff
@@ -72,6 +72,32 @@ func HorizontalPodAutoscaler(cfg config.Config, logger logr.Logger, otelcol v1al
 		}
 		metrics = append(metrics, targetCPUUtilization)
 
+		// check if fields are provided by .Spec.Autoscaler
+		// otherwise check deprecated fields
+		maxReplicas := new(int32)
```
pavolloffay, line 77: Could we use defaulting webhook to set the new fields to avoid this logic in this file? 

Also how will the upgrade work?

apis/v1alpha1/opentelemetrycollector_webhook.go
```diff
@@ -67,11 +67,22 @@ func (r *OpenTelemetryCollector) Default() {
 		r.Spec.TargetAllocator.Replicas = &one
 	}
 
-	if r.Spec.MaxReplicas != nil {
+	if r.Spec.MaxReplicas != nil || (r.Spec.Autoscaler != nil && r.Spec.Autoscaler.MaxReplicas != nil) {
 		if r.Spec.Autoscaler == nil {
 			r.Spec.Autoscaler = &AutoscalerSpec{}
 		}
 
+		if r.Spec.Autoscaler.MaxReplicas == nil {
```
pavolloffay, line 75: please add a test case for this in the webhook unit test

tests/e2e/autoscale/00-install.yaml
```diff
@@ -6,9 +6,9 @@ kind: OpenTelemetryCollector
 metadata:
   name: simplest
 spec:
-  minReplicas: 1
```
pavolloffay, line 9: I would prefer to keep this for now to verify the old config works. Maybe add a comment to update it.

apis/v1alpha1/opentelemetrycollector_webhook.go
```diff
@@ -67,11 +67,22 @@ func (r *OpenTelemetryCollector) Default() {
...
+			r.Spec.Autoscaler.MaxReplicas = r.Spec.MaxReplicas
+		}
+		if r.Spec.Autoscaler.MinReplicas == nil {
+			if r.Spec.MinReplicas != nil {
+				r.Spec.Autoscaler.MinReplicas = r.Spec.MinReplicas
+			} else {
+				r.Spec.Autoscaler.MinReplicas = r.Spec.Replicas
```
pavolloffay, line 82: Are we sure that this is correct?

moh-osman3: Hmm I believe the logic is correct, but maybe there's something I'm missing?

`MinReplicas` is meant to be an optional argument, but it should eventually always have a value if `MaxReplicas` is set. So here it defaults to `.Spec.Replicas` which is validated by this [check](https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1alpha1/opentelemetrycollector_webhook.go#L158) in the validator. I'm setting this default earlier now instead of waiting for the creation of the autoscaler to set the default [here](https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/collector/horizontalpodautoscaler.go#L83).

Wondering your thoughts and why this might not be the desired logic?

pkg/collector/horizontalpodautoscaler.go
moh-osman3: Thanks for the review @pavolloffay! I updated with your suggestion to use the defaulting webhook.

I am generally unfamiliar with the upgrade process, but took a look at [v0_56_0.go](https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/collector/upgrade/v0_56_0.go) which upgraded HPA to change the scaling target reference. This upgrade makes sense because the underlying HPA should be altered from scaling on `Deployment` to scaling on `OpenTelemetryCollector`. 

For this PR, the deprecated fields are still supported, so I think underlying HPA will remain unchanged. I don't think it's necessary to upgrade the HPA in this case (unless I'm missing something)

apis/v1alpha1/opentelemetrycollector_webhook.go
pavolloffay: > Wondering your thoughts and why this might not be the desired logic?

Originally I thought that the idea here is to totally decouple these two.  But it's reasonable to default to the `.spec.replicas`

pavolloffay (approved):

PR #1314, opened 2022-12-12
docs: add note about sidecar injection

LinuxAd:
Adds a quick note for anyone trying to inject sidecars from namespaces other than the one the pod they are injecting into is in.


frzifus (commented):
thanks @LinuxAd, but that might change soon:
- https://github.com/open-telemetry/opentelemetry-operator/pull/1395

jaronoff97:
no longer relevant! 🎉

PR #1323, opened 2022-12-13
[target-allocator] marshal with jsoniter

kristinapathak:
https://github.com/open-telemetry/opentelemetry-operator/issues/1336

Using jsoniter to marshal for server responses results in less memory and CPU used.  This can be seen in the benchmarks below, where `std` marks the previous code and `jsoniter` shows the performance with this PR.


kristinapathak:
benchmark results: 
[benchmarks.txt](https://github.com/open-telemetry/opentelemetry-operator/files/10273599/benchmarks.txt)

cmd/otel-allocator/server/server.go
```diff
@@ -42,27 +41,49 @@ var (
...
+}
+
 type Server struct {
 	logger           logr.Logger
 	allocator        allocation.Allocator
-	discoveryManager *target.Discoverer
+	discoveryManager DiscoveryManager
 	server           *http.Server
+	yamlMarshaller   jsoniter.API // TODO: for testing, it would make a lot of sense to have a simpler interface
+	jsonMarshaller   jsoniter.API // TODO: for testing, it would make a lot of sense to have a simpler interface
```
kristinapathak, line 73: @jaronoff97, do you think I should define my own interfaces here or keep using the `jsoniter.API` interface (which is pretty big)?

jaronoff97: Is there a more generic interface these adhere to?

cmd/otel-allocator/server/server.go
```diff
@@ -42,27 +41,49 @@ var (
...
+		ObjectFieldMustBeSimpleString: true,
+	}.Froze()
+)
+
 type collectorJSON struct {
 	Link string         `json:"_link"`
 	Jobs []*target.Item `json:"targets"`
 }
 
+type DiscoveryManager interface {
```
jaronoff97, line 63: nice add here

cmd/otel-allocator/server/server_test.go
```diff
@@ -15,15 +15,16 @@
 package server
 
 import (
-	"crypto/rand"
 	"encoding/json"
 	"fmt"
 	"io"
-	"math/big"
+	"math/rand"
```
jaronoff97, line 21: I thought we had a linter that discourages the use of `math/rand` in favor of `crypto/rand`, i guess it's fine given this is just a test.

jaronoff97: oh now i see, you're disabling the linter down there, that's fine

jaronoff97 (approved):

cmd/otel-allocator/server/server.go
kristinapathak: Not really 😕

Issue #1168, opened 2022-10-13
Use python OTLP/proto metrics exporter once available


aabmass:
Blocked by https://github.com/open-telemetry/opentelemetry-python/issues/2447

Right now I have set the `OTEL_METRICS_EXPORTER` to none to fix https://github.com/open-telemetry/opentelemetry-operator/issues/1148: https://github.com/open-telemetry/opentelemetry-operator/blob/992b681dd1bb8445b2744fb4f602f0faa0128ac9/pkg/instrumentation/python.go#L69-L72

Once https://github.com/open-telemetry/opentelemetry-python/issues/2447 is done, the operator's default should be updated to use the HTTP/proto exporter.

vainikkaj:
Also by this https://github.com/open-telemetry/opentelemetry-python/pull/3027
Both are merged but new release is needed.

PR #1328, opened 2022-12-14
Update default python exporters to use OTLP

TylerHelmuth:
Closes https://github.com/open-telemetry/opentelemetry-operator/issues/1168.

Updates the default exporters for python to use `otlp` and `OTEL_EXPORTER_OTLP_[TRACES|METRICS]_PROTOCOL=http/protobuf`

Depends on https://github.com/open-telemetry/opentelemetry-operator/pull/1321

### DIFF
--- a/.chloggen/update-python-metrics-exporter.yaml
+++ b/.chloggen/update-python-metrics-exporter.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: instrumentation/python
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Updates the default exporters to `otlp` and sets `OTEL_EXPORTER_OTLP_[TRACES|METRICS]_PROTOCOL` to `http/protobuf`
+
+# One or more tracking issues related to the change
+issues: [1328]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/pkg/instrumentation/podmutator_test.go
+++ b/pkg/instrumentation/podmutator_test.go
@@ -406,15 +406,15 @@ func TestMutatePod(t *testing.T) {
 							},
 							{
 								Name:  "OTEL_TRACES_EXPORTER",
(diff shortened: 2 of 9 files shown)


frzifus (approved):

mat-rumian:
@TylerHelmuth shouldn't we use `OTEL_METRICS_EXPORTER` as `otlp` and then set `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` as `http/protobuf` for latest release to avoid [warning logs](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/_configuration/__init__.py#L154) which can be confusing for end users?
I think the same should be done for traces. What you think? :)

TylerHelmuth:
@mat-rumian sure I can make that change.

mat-rumian:
> Shall we as well change this comment in the requirements
> 
> https://github.com/open-telemetry/opentelemetry-operator/blob/2ea988c6094e4d4a6d3c60571da2a516f895c9bb/autoinstrumentation/python/requirements.txt#L2
> 
> ?
> So do I get it right that the exporter will use proto over HTTP and not gRPC?

We didn't use `gRPC` from the beginning. PR just changing the way exporter is configured. Previous configuration was setting up exporter to use `proto` and `HTTP` in single environment variable `OTEL_TRACES_EXPORTER` (like `otlp_proto_http`). 
OT Python community developed support for configuration described in the [OT Specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#configuration-options) - so `OTEL_x_EXPORTER` which defines the exporter (e.g. zipkin, jaeger, otlp) and `OTEL_EXPORTER_OTLP_x_PROTOCOL` which defines protocol (e.g. grpc, http/protobuf, http/json).

mat-rumian (approved):

pavolloffay (approved):

TylerHelmuth:
@frzifus @pavolloffay @mat-rumian the e2e tests highlighted an interesting scenario that my code wasn't handling.  Originally I was only setting the protocol when the exporter also needed set to `otlp`, but this doesn't work when the user has already set the the exporter to `otlp` themselves. 

I've updated the code to be less intelligent and always set the otlp protocol env var if it is not set.  This env var is unncessary when the exporter is set to something other than `otlp`, but it also shouldn't be hurting anything.  We could go the extra step and only set the protocol if it is not set AND the exporter is set to `otlp`, but that feels a little flimsy.  I think always setting the protocol if it is not set will be more flexible as the spec/python auto-instrumentation finalize.

Please re-review.

pavolloffay (approved):

Issue #924, opened 2022-06-10
Auto instrumentation python service - trace data doesn't arrive in Endpoint


tybalex:
I am trying out the auto-instrumentation feature of Open Telemetry with an example demo application.  I tried to applied auto-instrumentation to the 2 python services `recommendation` and `email`, as well as 2 nodejs services `payment` and `currency`. However I was only able to find trace data from nodejs services in my endpoint (Opensearch), but not python services'.

**My environment**
k3s-1.23
[example demo application](https://github.com/GoogleCloudPlatform/microservices-demo)
python 3.7 for the python services

manifest of the python service:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: recommendationservice
spec:
  selector:
    matchLabels:
      app: recommendationservice
  template:
    metadata:
      labels:
        app: recommendationservice
      annotations:
        sidecar.opentelemetry.io/inject: "true"
        instrumentation.opentelemetry.io/inject-python: "true"
    spec:
      serviceAccountName: default
      terminationGracePeriodSeconds: 5
      containers:
      - name: server
        image: tybalex/recommendationservice:dev
        ports:
        - containerPort: 8080
        readinessProbe:
          periodSeconds: 5
          exec:
            command: ["/bin/grpc_health_probe", "-addr=:8080"]
        livenessProbe:
          periodSeconds: 5
          exec:
            command: ["/bin/grpc_health_probe", "-addr=:8080"]
        env:
        - name: PORT
          value: "8080"
        - name: PRODUCT_CATALOG_SERVICE_ADDR
          value: "productcatalogservice:3550"
        resources:
          requests:
            cpu: 100m
            memory: 220Mi
          limits:
            cpu: 200m
            memory: 450Mi
---
apiVersion: v1
kind: Service
metadata:
  name: recommendationservice
spec:
  type: ClusterIP
  selector:
    app: recommendationservice
  ports:
  - name: grpc
    port: 8080
    targetPort: 8080
```

Example logs in the recommendation service pod: 
```
[...]

PR #1330, opened 2022-12-15
[chore] Update Instrumentation example with python env var

TylerHelmuth:
Related to https://github.com/open-telemetry/opentelemetry-operator/issues/924

Updates the README's Instrumentation object to explicitly set python's OTEL_EXPORTER_OTLP_ENDPOINT value to 4318.  

If we want we can also update the python injection to update any existing OTEL_EXPORTER_OTLP_ENDPOINT value to be `4318` so that this distinction isn't necessary, but I think that change needs some scrutiny to make sure users can still configure anything they want with confidence that we aren't messing with it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,9 @@ spec:
     argument: "0.25"
   python:
     env:
+      # Required if endpoint is set to 4317.
+      # Python autoinstrumentation uses http/proto by default
+      # so data must be sent to 4318 instead of 4137.
       - name: OTEL_EXPORTER_OTLP_ENDPOINT
         value: http://otel-collector:4318
 EOF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,10 @@ spec:
   sampler:
     type: parentbased_traceidratio
     argument: "0.25"
+  python:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -196,6 +196,10 @@ spec:
   sampler:
     type: parentbased_traceidratio
     argument: "0.25"
+  python:
+    env:
+      - name: OTEL_EXPORTER_OTLP_ENDPOINT
```
pavolloffay, line 201: I think this deserves a comment (for us but as well for the users) to explain why this needs to be overridden.

There are as well other references in the code related to this issue: 

https://github.com/open-telemetry/opentelemetry-operator/blob/2ea988c6094e4d4a6d3c60571da2a516f895c9bb/autoinstrumentation/python/requirements.txt#L2

https://github.com/open-telemetry/opentelemetry-operator/blob/992b681dd1bb8445b2744fb4f602f0faa0128ac9/pkg/instrumentation/python.go#L60

pavolloffay (approved):

PR #1335, opened 2022-12-20
[chore]  Add a smoke test that is run as part of the scorecard tests V2

andreasgerstmayr:
This is a continuation of PR #1252, rebased on latest main branch with the following updates:
* a new service account is created to run the scorecard tests (the default service account doesn't have permissions to perform the tasks of the kuttl tests, for example creating a new CR)
* the e2e cluster environment is set up before running the tests (the kuttl tests require the operator to be installed)


bundle/tests/scorecard/config.yaml
```diff
@@ -45,6 +46,13 @@ stages:
     storage:
       spec:
         mountPath: {}
+  - image: quay.io/operator-framework/scorecard-test-kuttl:v2.0.0
+    labels:
+      suite: kuttlsuite
+      test: kuttltest1
```
pavolloffay, line 52: How is the label used? 

Should we use a more descriptive label?

andreasgerstmayr: Afaics it's just printed in the scorecard output, i.e.

```
    {
      "kind": "Test",
      "apiVersion": "scorecard.operatorframework.io/v1alpha3",
      "spec": {
        "image": "quay.io/operator-framework/scorecard-test-kuttl:v2.0.0",
        "labels": {
          "suite": "kuttlsuite",
          "test": "kuttltest1"
        },
        "storage": {
          "spec": {
            "mountPath": {}
          }
        }
      },
      "status": {
        "results": [
          {
            "name": "smoketest",
            "state": "pass",
            "creationTimestamp": null
          }
        ]
      }
    }
```

Currently all tests of this kuttl suite (`bundle/tests/scorecard/kuttl`) are run in parallel as a single scorecard test. If we decide later that we want to split kuttl tests in multiple scorecard tests, we should add a descriptive label to each. For now I think it's fine, as in other projects we also run all kuttl tests in the same kuttl suite (i.e. `tests/e2e` is a single kuttle suite).

config/scorecard/patches/kuttl.config.yaml
```diff
@@ -0,0 +1,6 @@
+- op: add
+  path: /stages/0/tests/-
+  value:
+    image: docker.io/andreasgerstmayr/scorecard-test-kuttl:1681909083
```
pavolloffay, line 4: Why is it using a custom image?

bundle/tests/scorecard/config.yaml
```diff
@@ -45,6 +46,12 @@ stages:
     storage:
       spec:
         mountPath: {}
+  - image: docker.io/andreasgerstmayr/scorecard-test-kuttl:1681909083
```
pavolloffay, line 49: Why is it using a custom image? Could we switch to the upstream one?

andreasgerstmayr: Same as above (this file is auto-generated from the file above).

atoulme:

PR #1338, opened 2022-12-22
Fix autoinstrumentation E2E tests OpenShift

iblancasa:
After #1254, some E2E tests were not passing in OpenShift. This PR fixes the E2E tests to work in OpenShift.

The tests were not passing because the business applications cannot be deployed in OpenShift because the security context of the deployments are not using the user IDs, group IDS. and FSGroup that OpenShift accepts.

Fixed E2E tests:
* instrumentation-python
* instrumentation-dotnet
* instrumentation-java
* instrumentation-nodejs
* instrumentation-java-other-ns

As part of this PR I had to decouple the logic to create OpenShift Routes from the detection of the platform.


pavolloffay:
> As part of this PR I had to decouple the logic to create OpenShift Routes from the detection of the platform.

Why was it decoupled? @frzifus could you please review this change?

hack/set-openshift-security-context/main.go
```diff
@@ -0,0 +1,133 @@
...
+	if raw == "" {
+		fmt.Println("The annotation ", annotation, " is not present")
+		os.Exit(1)
+	}
+
+	lowBound := strings.Split(raw, "/")[0]
+	id, err := strconv.ParseInt(lowBound, 0, 64)
+	if err != nil {
+		fmt.Println("It was not possible to convert the number to int64: ", lowBound)
+		os.Exit(1)
```
pavolloffay, line 49: To keep good practices I would prefer to return an error and exit only in the main function

iblancasa:
@pavolloffay I'm using the autodetect to check if I'm using OpenShift. Since that feature relies on having the OpenShift Routes installed, the E2E tests always "are running in OpenShift" but this is not true (for instance, in the GitHub Actions Workflow, we're running the E2E tests in KiND but the operator things it is running in OpenShift).

hack/set-openshift-security-context/main.go
```diff
@@ -0,0 +1,133 @@
...
+	runningPlatform, err := ad.Platform()
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	if runningPlatform == platform.OpenShift {
+		fmt.Println("Connected to an OpenShift cluster")
+	} else {
+		fmt.Println("Nothing extra needs to be done")
```
pavolloffay, line 94: I would swap `extra` for something more explanatory (e.g. what the "script" does)

hack/set-openshift-security-context/main.go
```diff
@@ -0,0 +1,133 @@
...
+
+	deploymentsClient := client.AppsV1().Deployments(namespace.Name)
+	deployment, err := deploymentsClient.Get(
+		context.Background(),
+		deploymentName,
+		metav1.GetOptions{},
+	)
+
+	if err != nil {
+		fmt.Println("Deployment was not found")
```
pavolloffay, line 118: add deployment name

hack/set-openshift-security-context/main.go
```diff
@@ -0,0 +1,133 @@
...
+	*userId = getUserID(namespace)
+
+	var groupdId *int64 = new(int64)
+	*groupdId = getGroupID(namespace)
+
+	deployment.Spec.Template.Spec.SecurityContext.RunAsUser = userId
+	deployment.Spec.Template.Spec.SecurityContext.RunAsGroup = groupdId
+	deployment.Spec.Template.Spec.SecurityContext.FSGroup = groupdId
+
+	_, _ = deploymentsClient.Update(context.Background(), deployment, metav1.UpdateOptions{})
```
pavolloffay, line 132: shall we check the error?

iblancasa:
Ping @frzifus

frzifus:
on it

internal/config/main.go
```diff
@@ -64,6 +65,7 @@ func New(opts ...Option) Config {
 		version:                       version.Get(),
 		autoscalingVersion:            autodetect.DefaultAutoscalingVersion,
 		onPlatformChange:              newOnChange(),
+		openShiftRoutesAvailable:      autodetect.DefaultOpenShiftAvailability,
```
frzifus, line 68: Could we please add a test that verifies that `openShiftRoutesAvailable` is updated in all configuration copies? Since the detection only runs on the main configuration.

internal/config/main_test.go
```diff
@@ -99,7 +105,8 @@ func TestAutoDetectInBackground(t *testing.T) {
 var _ autodetect.AutoDetect = (*mockAutoDetect)(nil)
 
 type mockAutoDetect struct {
-	PlatformFunc func() (platform.Platform, error)
+	PlatformFunc                    func() (platform.Platform, error)
+	OpenShiftRoutesAvailabilityFunc func() (autodetect.OpenShiftRoutesAvailability, error)
```
frzifus, line 109: Is there a chance that the openshift platform is used but routes do not exist?

pkg/autodetect/main.go
```diff
@@ -71,7 +82,7 @@ func (a *autoDetect) Platform() (platform.Platform, error) {
 
 	apiGroups := apiList.Groups
 	for i := 0; i < len(apiGroups); i++ {
-		if apiGroups[i].Name == "route.openshift.io" {
+		if apiGroups[i].Name == "operator.openshift.io" {
```
frzifus, line 85: With this change, routes are now recognized under openshift but no longer under k8s, correct?

pkg/autodetect/main.go
```diff
@@ -105,6 +116,21 @@ func (a *autoDetect) HPAVersion() (AutoscalingVersion, error) {
...
 }
 
+func (a *autoDetect) OpenShiftRoutesAvailability() (OpenShiftRoutesAvailability, error) {
+	apiList, err := a.dcl.ServerGroups()
+	if err != nil {
+		return OpenShiftRoutesUnknown, err
+	}
+
+	for _, apiGroup := range apiList.Groups {
+		if apiGroup.Name == "route.openshift.io" {
```
frzifus, line 126: Why is the api group name here different?

pkg/autodetect/main.go
```diff
@@ -117,6 +143,18 @@ func (v AutoscalingVersion) String() string {
...
 }
 
+func (v OpenShiftRoutesAvailability) String() string {
+	switch v {
+	case OpenShiftRoutesAvailable:
+		return "available"
+	case OpenShiftRoutesNotAvailable:
+		return "not available"
+	case OpenShiftRoutesUnknown:
+		return "unknown"
```
frzifus, line 153: Lets remove that case, since we return unknown anyway.

pkg/autodetect/main.go
iblancasa: This code checks if the `route.openshift.io` API resource is installed in the cluster.

pkg/autodetect/main.go
iblancasa: The problem I'm trying to solve is this:
- Detecting if the platform is OpenShift is based on finding the `route.openshift.io` resource.
- The `route.openshift.io` is installed in KiND as part of the E2E tests
- The E2E tests think they are running in OpenShift but that is not true

So, I modified the code to:
- Detect if the platform is OpenShift based on finding another resource (I chose `operator.openshift.io` but can be another one except `route.openshift.io`)
- Add methods to detect if `route.openshift.io` is installed (what is true for OpenShift and Kubernetes clusters where `routes.openshift.io` is installed)

internal/config/main_test.go
iblancasa: I don't think so because is part of the core AFAIK. If you want, I can take scenario into consideration. What do you think?

internal/config/main.go
iblancasa: I'll take a look :)

iblancasa:
@frzifus thanks for your review! I answered the different comments inline.

internal/config/main.go
iblancasa: Does it make sense to test this? Because, if the `autodetect` is called just for the main configuration, why would this field change?

internal/config/main_test.go
frzifus: If its the same, why do we need `OpenShiftRoutesAvailabilityFunc` at all?

iblancasa: The explanation is here:
https://github.com/open-telemetry/opentelemetry-operator/pull/1338#discussion_r1071159192

It can be that somebody installs the OpenShift Routes in Kubernetes (as it happens in the E2E tests). The operator will think it is running in OpenShift when it is not true.

internal/config/main.go
frzifus: > The same case applies for autoscalingVersion, right?

Seems like, yes. :)

> Does it make sense to test this? Because, if the autodetect is called just for the main configuration, why would this field change?

From my point of view it does. Its exported and highly misleading if it gets called from somewhere else in the future. We had that issue with `onPlatformChange` previously. It was super annoying.

pkg/autodetect/main.go
frzifus: ok, i see. Then i would follow the same approach as the Platform detection and use the method to verify if routes are available in the reconciler.

iblancasa:
I'll close this PR and:
* Solve the HPA problems in a new PR
* Solve the OpenShift detection problems in a new PR
* Fix the E2E test in a new PR

Issue #1318, opened 2022-12-12
OpAMP Bridge resource in the operator


jaronoff97:
Per the discussion in the Operator SIG, the plan is for the Operator to take on creating a container that is able to remotely configure OpenTelemetryCollector CRDs via [OpAMP](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md). The design documents for this proposal can be read [here](https://docs.google.com/document/d/1M8VLNe_sv1MIfu5bUR5OV_vrMBnAI7IJN-7-IAr37JY/edit#). 

I'm planning on taking on this work in the coming weeks/months. I plan to create a new directory in the operator repo's `cmd` folder with a remoteconfiguration service. This container will be built in the same way as the target allocator. The service will be marked as alpha in its README.md. There's probably a few more organizational tasks related to this that will need to be done.

Todo (part 1)
- [x] Create the new directory for the remoteconfiguration service
- [x] Create a valid opamp agent in the remote configuration
- [x] Report health and status via OpAMP
- [x] Connect the opamp agent to a valid operator opamp server
- [x] Allow the remoteconfiguration service to create OpenTelemetryCollector CRDs
- [x] Create collector CRDs from received messages from opamp
- [x] Add tests and documentation 

Todo (part 2)
- [x] #1369
- [x] #1368

Todo (part 3)
- [x] #1370
- [x] #1932
- [x] #2131
- [x] #2132
- [x] #2282

jaronoff97:
I'm planning on starting on the part 1 next week

jaronoff97:
issues will be made as needed, i just added three more that can mostly be done in parallel if anyone else wants to hop in :)

jaronoff97:
All these are completed and the bridge is released! It's already being used in production ;)

PR #1339, opened 2022-12-22
[new-component] Operator OpAMP Bridge Service

jaronoff97:
Implements part 1 of #1318 

A few things before this is ready:

- [x] Add in deletion of collectors
- [x] Add in block rules
- [x] Re-add metrics in for the agent
- [x] Updated github actions to add to list of things to run tests and linting for


cmd/remote-configuration/agent/agent.go
```diff
@@ -0,0 +1,283 @@
...
+	reporter, err := metrics.NewMetricReporter(agent.logger, settings, agent.config.GetAgentType(), agent.config.GetAgentVersion(), agent.instanceId)
+	if err != nil {
+		agent.logger.Errorf("Cannot collect metrics: %v", err)
+		return
+	}
+
+	if agent.metricReporter != nil {
+		agent.metricReporter.Shutdown()
+	}
+	agent.metricReporter = reporter
```
kristinapathak, line 195: Can `initMeter()` be called concurrently? This feels like something that could be messy with concurrency and need a lock.

cmd/remote-configuration/agent/agent.go
```diff
@@ -0,0 +1,283 @@
...
+		if err != nil {
+			multiErr = multierr.Append(multiErr, err)
+			continue
+		}
+		err = agent.applier.Apply(name, namespace, file)
+		if err != nil {
+			multiErr = multierr.Append(multiErr, err)
+			continue
+		}
+		agent.appliedKeys[key] = true
```
kristinapathak, line 216: Can `applyRemoteConfig()` be called concurrently?

cmd/remote-configuration/agent/agent.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+	for collectorKey := range agent.appliedKeys {
+		name, namespace, err := agent.getNameAndNamespace(collectorKey)
+		if err != nil {
+			multiErr = multierr.Append(multiErr, err)
+			continue
+		}
+		if _, ok := config.Config.GetConfigMap()[collectorKey]; !ok {
+			err = agent.applier.Delete(name, namespace)
```
kristinapathak, line 225: I'm not familiar with the applier - Why are we deleting this right after applying it?

cmd/remote-configuration/agent/agent.go
```diff
@@ -0,0 +1,283 @@
...
+		}
+		err = agent.applier.Apply(name, namespace, file)
+		if err != nil {
+			multiErr = multierr.Append(multiErr, err)
+			continue
+		}
+		agent.appliedKeys[key] = true
+	}
+	for collectorKey := range agent.appliedKeys {
+		name, namespace, err := agent.getNameAndNamespace(collectorKey)
```
kristinapathak, line 219: I don't think getting the name and namespace is a lot of work (and it's only added to `appliedKeys` if this function works anyways), but couldn't we save them in `appliedKeys` so we don't have to keep doing this?

cmd/remote-configuration/agent/agent.go
```diff
@@ -0,0 +1,283 @@
...
+	agent.metricReporter = reporter
+}
+
+// Take the remote config, layer it over existing, done
+// INVARIANT: The caller must verify that config isn't nil _and_ the configuration has changed between calls.
+func (agent *Agent) applyRemoteConfig(config *protobufs.AgentRemoteConfig) (*protobufs.RemoteConfigStatus, error) {
+	var multiErr error
+	for key, file := range config.Config.GetConfigMap() {
+		if len(key) == 0 || len(file.Body) == 0 {
+			continue
```
kristinapathak, line 204: Why do we continue? If there's a partial success, the config hash is no longer accurate to what's being provided.  Is that okay?

cmd/remote-configuration/agent/agent.go
```diff
@@ -0,0 +1,283 @@
...
+	return &protobufs.RemoteConfigStatus{
+		LastRemoteConfigHash: config.ConfigHash,
+		Status:               protobufs.RemoteConfigStatuses_RemoteConfigStatuses_APPLIED,
+	}, nil
+}
+
+func (agent *Agent) Shutdown() {
+	agent.logger.Debugf("Agent shutting down...")
+	if agent.opampClient != nil {
+		_ = agent.opampClient.Stop(context.Background())
```
kristinapathak, line 248: I assume this is an error - why don't we return/log it?

kristinapathak (commented):
A few questions/thoughts 🙂 very excited to learn more about this.

cmd/remote-configuration/agent/agent.go
```diff
@@ -0,0 +1,254 @@
...
+
+	agent.logger.Debugf("Agent starting, id=%v, type=%s, version=%s.",
+		agent.instanceId.String(), config.GetAgentType(), config.GetAgentVersion())
+
+	return agent
+}
+
+func (agent *Agent) getHealth() *protobufs.AgentHealth {
+	return &protobufs.AgentHealth{
+		Healthy:           true,
```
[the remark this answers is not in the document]
jaronoff97: This is something to iterate on later, right now this is only called once, but in the future we could use this for better agent management.

cmd/remote-configuration/agent/agent.go
```diff
@@ -0,0 +1,283 @@
...
+func (agent *Agent) updateAgentIdentity(instanceId ulid.ULID) {
+	agent.logger.Debugf("Agent identify is being changed from id=%v to id=%v",
+		agent.instanceId.String(),
+		instanceId.String())
+	agent.instanceId = instanceId
+
+	// TODO: reinit or update meter (possibly using a single function to update all own connection settings
+}
+
+func (agent *Agent) getNameAndNamespace(key string) (string, string, error) {
```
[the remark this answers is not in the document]
jaronoff97: We can make this standalone

cmd/remote-configuration/agent/agent.go
jaronoff97: we're only deleting objects that no longer exist in the configmap we've received from the remote server. I'll update this with a comment to make that more clear.

kristinapathak: Oh that makes much more sense. 🙂 I misunderstood the if condition.

cmd/remote-configuration/agent/agent.go
jaronoff97: this was a choice by me to say that it's okay to have a partial success, i agree the configHash would not be accurate though. Should we just throw away everything in this case?

kristinapathak: I'm not sure. 😕 It seems difficult to make this function an atomic operation.

cmd/remote-configuration/agent/agent_test.go
```diff
@@ -0,0 +1,553 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockClient := &mockOpampClient{}
+			conf, err := config.Load(tt.fields.configFile)
+			assert.NoError(t, err, "should be able to load config")
+			applier := getFakeApplier(t, conf)
+			agent := NewAgent(clientLogger, applier, conf, mockClient)
+			err = agent.Start()
+			defer agent.Shutdown()
+			assert.NoError(t, err, "should be able to start agent")
```
kristinapathak, line 455: I don't think you want to continue trying to test if some of these things fail, so I would suggest using the `require` package instead of the `assert` package wherever we want to end the test upon getting an unexpected result.  Then, if a test fails, it will be more clear what went wrong.

cmd/remote-configuration/metrics/reporter.go
```diff
@@ -0,0 +1,209 @@
...
+	)
+	if resourceErr != nil {
+		return nil, resourceErr
+	}
+
+	provider := sdkmetric.NewMeterProvider(
+		sdkmetric.WithResource(resource),
+		sdkmetric.WithReader(sdkmetric.NewPeriodicReader(client)))
+
+	global.SetMeterProvider(provider)
```
Aneurysm9, line 113: Should this be set as the global provider or should it be embedded in the `MetricReporter`?

cmd/remote-configuration/metrics/reporter.go
```diff
@@ -0,0 +1,209 @@
...
+
+	reporter := &MetricReporter{
+		logger: logger,
+	}
+
+	reporter.done = make(chan struct{})
+
+	reporter.meter = global.Meter("opamp")
```
Aneurysm9, line 121:
```suggestion
	reporter.meter = provider.Meter("opamp")
```

No need to involve globals here.  We know which provider to use since we just created it.

cmd/remote-configuration/metrics/reporter.go
```diff
@@ -0,0 +1,209 @@
...
+func (reporter *MetricReporter) processMemoryPhysicalFunc(ctx context.Context) {
+	memory, err := reporter.process.MemoryInfo()
+	if err != nil {
+		reporter.logger.Errorf("Cannot get process memory information: %w", err)
+		return
+	}
+	reporter.processMemoryPhysical.Observe(ctx, int64(memory.RSS))
+}
+
+func (reporter *MetricReporter) sendMetrics() {
```
Aneurysm9, line 182: Is this intended to have further function down the line?  I don't see that it's doing anything other than counting the number of 5s periods elapsed since it was started.  Actually collecting and sending metrics is governed by the `NewPeriodicReader()` created above.

cmd/remote-configuration/agent/agent.go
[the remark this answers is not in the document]
frzifus: Does it make sense to add a `TODO` or even create a ticket for that?

cmd/remote-configuration/logger/logger.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"fmt"
+
+	"github.com/go-logr/logr"
+	"github.com/open-telemetry/opamp-go/client/types"
+)
+
+var _ types.Logger = &Logger{}
+
+type Logger struct {
```
frzifus, line 26: Whats the reason to not use `*logr.Logger` directly?

cmd/remote-configuration/main.go
```diff
@@ -0,0 +1,61 @@
...
+
+func main() {
+	cliConf, err := config.ParseCLI()
+	if err != nil {
+		os.Exit(1)
+	}
+	cfg, configLoadErr := config.Load(*cliConf.ConfigFilePath)
+	if configLoadErr != nil {
+		cliConf.RootLogger.Error(configLoadErr, "Unable to load configuration")
+		return
```
frzifus, line 36: Shouldnt we exit here with an error code too?

cmd/remote-configuration/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+RUN go mod download
+
+COPY . .
+
+# Build the Go app
+RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o main .
+
+######## Start a new stage from scratch #######
+FROM alpine:latest
```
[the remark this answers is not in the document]
iblancasa: I agree @frzifus

cmd/remote-configuration/operator/client.go
```diff
@@ -0,0 +1,200 @@
...
+	err := c.k8sClient.Get(ctx, client.ObjectKey{
+		Namespace: namespace,
+		Name:      name,
+	}, &result)
+	if errors.IsNotFound(err) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, err
+	}
```
iblancasa, line 168: Would it make sense to do this?
```suggestion
	if err != nil {
		if errors.IsNotFound(err) ́{
			return nil, nil
		}
		return nil, err
	}
```

Makefile
```diff
@@ -201,6 +206,10 @@ container-target-allocator-push:
 container-target-allocator:
 	docker buildx build  --load --platform linux/${ARCH} -t ${TARGETALLOCATOR_IMG} cmd/otel-allocator
 
+.PHONY: container-remote-configuration
+container-remote-configuration:
```
iblancasa, line 210: Missing `container-remote-configuration-push`

iblancasa (changes requested):
Just asking: should be created an E2E test?

cmd/remote-configuration/Dockerfile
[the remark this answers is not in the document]
Aneurysm9: All for non-root users in the scratch container, as well.  If we could avoid using `gcr.io/distroless` as the base that would make my life a lot easier.

cmd/remote-configuration/agent/agent.go
```diff
@@ -0,0 +1,290 @@
...
+
+	agent.logger.Debugf("OpAMP Client started.")
+
+	return nil
+}
+
+// updateAgentIdentity receives a new instanced Id from the remote server and updates the agent's instanceID field.
+// The meter will be reinitialized by the onMessage function.
+func (agent *Agent) updateAgentIdentity(instanceId ulid.ULID) {
+	agent.logger.Debugf("Agent identify is being changed from id=%v to id=%v",
```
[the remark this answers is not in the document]
jaronoff97: Unfortunately no. The logging interface is pretty rudimentary. I have an open issue with the opamp-go repo to improve this https://github.com/open-telemetry/opamp-go/issues/154

cmd/remote-configuration/metrics/reporter.go
jaronoff97: Probably embedded within the metricreporter.

cmd/remote-configuration/metrics/reporter.go
jaronoff97: This code was from the example agent code from the opamp-go example client. I think we can remove it. This metrics code was the one thing I wasn't able to test well as part of this.

cmd/remote-configuration/agent/agent.go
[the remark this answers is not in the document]
jaronoff97: I can add a TODO and then on merge i can make a ticket.

cmd/remote-configuration/logger/logger.go
jaronoff97: We're bound by the requirements of the logging interface of the opamp client unfortunately as i mentioned [here](https://github.com/open-telemetry/opentelemetry-operator/pull/1339#discussion_r1063551902).

cmd/remote-configuration/operator/client.go
jaronoff97: I usually try and not nest `if` blocks where possible, but your suggestion may read better. Thank you for the suggestion!

Makefile
jaronoff97: I'm leaving all CI/CD around pushing this container to a separate PR to make it easier to review that component. This goes with an E2E test as well as we don't have a CRD for this yet.

cmd/remote-configuration/metrics/reporter.go
jaronoff97: tested this and everything worked after i moved to use the grpc client :D metrics were flowing through. I added a TODO to add more metrics in.

cmd/remote-configuration/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+COPY . .
+
+# Build the Go app
+RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o main .
+
+######## Start a new stage from scratch #######
+FROM scratch
+
+RUN apk --no-cache add ca-certificates
```
Aneurysm9, line 19: This won't work.  `FROM scratch` is literally that, nothing exists in the container FS.  The `builder` stage could be updated to be `FROM golang:1.19-alpine` and then `apk` used to install the certs in that container and then copied over with the build artifact.  See https://github.com/aws-observability/aws-otel-collector/blob/main/cmd/awscollector/Dockerfile for an example of populating a scratch image.

jaronoff97: ah okay, thank you! will update.

cmd/operator-opamp-bridge/metrics/reporter.go
```diff
@@ -0,0 +1,174 @@
...
+	agentVersion string,
+	instanceId ulid.ULID,
+) (*MetricReporter, error) {
+
+	if dest.DestinationEndpoint == "" {
+		err := fmt.Errorf("metric destination must specify DestinationEndpoint")
+		return nil, err
+	}
+
+	// Create OTLP/grpc metric exporter.
```
tigrannajaryan, line 72: OpAMP spec says that own metrics should be reported using OTLP/HTTP: https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#own-telemetry-reporting

cmd/operator-opamp-bridge/agent/testdata/agent.yaml
```diff
@@ -0,0 +1,17 @@
+endpoint: ws://127.0.0.1:4320/v1/opamp
+protocol: wss
+capabilities:
+  - Unspecified
```
tigrannajaryan, line 4: Nit: "Unspecified" is not supposed to be specifiable by the user and it is a 0 value which doesn't set any capability bit, so it is essentially a no-op even if the user does  specify it.

cmd/operator-opamp-bridge/agent/testdata/agent.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - ReportsStatus
```
tigrannajaryan, line 5: Just an FYI: ReportsStatus [is always set](https://github.com/open-telemetry/opamp-spec/blob/79771f7fff1e1ae10ccdaffeee5b2b3689c8dd8c/proto/opamp.proto#LL522C10-L522C10) by OpAMP client, so it doesn't matter if the user specifies it here or no.
I don't think you need to do anything special about it, just wanted to make sure you are aware and avoid listing it e.g. in the operator docs in the future.

cmd/operator-opamp-bridge/agent/agent.go
```diff
@@ -0,0 +1,294 @@
...
+		status, err := agent.applyRemoteConfig(msg.RemoteConfig)
+		if err != nil {
+			agent.logger.Errorf(err.Error())
+		}
+		err = agent.opampClient.SetRemoteConfigStatus(status)
+		if err != nil {
+			agent.logger.Errorf(err.Error())
+			return
+		}
+		err = agent.opampClient.UpdateEffectiveConfig(ctx)
```
tigrannajaryan, line 274: Note that this will error if `ReportsEffectiveConfig` capability is not specified. It may result in unnecessary error message in the log, so perhaps check if the user wants the capability and only in that case call this function.

cmd/operator-opamp-bridge/metrics/reporter.go
jaronoff97: is there a reason it specifies HTTP over grpc? I tend to always use gRPC when it's available.

tigrannajaryan: 2 reasons:
1. OpAMP itself has no dependency on gRPC, so it was seen as an unnecessary additional burden for agents that don't use gRCP for any other purpose.
2. OTLP/HTTP is slightly more performant than OTLP/gRPC (particularly for small payloads). Generally, I am not aware of any use case where OTLP/gRPC is preferable over OTLP/HTTP.

tigrannajaryan:
@jaronoff97 it would be great to have your feedback on OpAMP spec and OpAMP Go as you work on this capability of the operator. Please feel free to open issues in OpAMP repos or ping me if you have any questions.

Aneurysm9 (approved):

cmd/operator-opamp-bridge/agent/agent.go
tigrannajaryan: I see this is guarded by `remoteConfigEnabled` which is populated from `AcceptsRemoteConfig` capability. However `AcceptsRemoteConfig` and `ReportsEffectiveConfig` are different capabilities and can be set independently. It is possible for ReportsEffectiveConfig=true and ReportsEffectiveConfig=false (although it is a great config for users to use and I don't know why anyone would want to use this combination of capabilities).

iblancasa (approved):

pavolloffay (approved):

PR #1345, opened 2023-01-03
Refactor the target allocator build to not run it as root

### LINKED ISSUE #1346: [target allocator] Not run the target allocator as root
Currently, the target allocator is run as root. This prevents the target allocator from being run in OpenShift without special capabilities.

When the target allocator pod is created, the following error can be seen in the `kubectl describe <pod name>` command output:
```
  Warning  Failed          12s (x2 over 12s)  kubelet            Error: container create failed: time="2023-01-03T15:56:58Z" level=error msg="runc create failed: unable to start container process: exec: \"./main\": stat ./main: permission denied"
```

## How to reproduce
1. Create an OCP cluster (version 4.11 in my case)
2. Build and push to a registry the OTEL Collector Operator and Target Allocator container images
3. `$ kubectl create namespace oteltest`
4. `$ kubectl -n oteltest create rolebinding default-view-oteltest --role=pod-view --serviceaccount=oteltest:ta`
5. `$ kubectl create -f tests/e2e/targetallocator-features/00-install.yaml -n oteltest:ta`
6. Wait until the pod creation fails (`$ kubectl get pods -n oteltest`)

> **Note**
> The step `4` is executed manually because, when applying the manifest, it will fail because the `TestStep` kind is part of KUTTL.

iblancasa:

Fixes #1346

This change makes the target allocator being run as non root. The Dockerfile follows the same structure than the one from the OTEL Collector Operator.

## Steps performed to ensure the issue is fixed
1. Create an OCP cluster (version 4.11 in my case)
2. Build and push to a registry the OTEL Collector Operator and Target Allocator container images
3. `$ kubectl create namespace oteltest`
4. `$ kubectl -n oteltest create rolebinding default-view-oteltest --role=pod-view --serviceaccount=oteltest:ta`
5. `$ kubectl create -f tests/e2e/targetallocator-features/00-install.yaml -n oteltest:ta`
6. Check that the pods are created properly (`$ kubectl get pods -n oteltest`)


.chloggen/1345-run-ta-nonroot.yaml
```diff
@@ -0,0 +1,16 @@
...
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: target allocator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Run the target allocator as non root user"
+
+# One or more tracking issues related to the change
+issues: []
```
jaronoff97, line 11: There needs to be a tracking issue for this

jaronoff97 (commented):
How was this tested?

iblancasa:
@jaronoff97 thanks for your review!

I created #1346 to track this issue. There, you can see how I found the issue and how I checked everything was working properly after applying the changes from this PR.

jaronoff97:
Thanks for making the issue! Could you add how you tested this to the description of the PR?

iblancasa:
@jaronoff97 done

jaronoff97 (approved):

cmd/otel-allocator/Dockerfile
```diff
@@ -1,26 +1,34 @@
...
+WORKDIR /workspace
+# Copy the Go Modules manifests
+COPY go.mod go.mod
+COPY go.sum go.sum
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
 RUN go mod download
 
-COPY . .
+# Copy the go source
```
frzifus, line 12: Does it make senes to keep `COPY . .`? Unused files are not copied in the next stage anyway?

frzifus (commented):
Otherwise lgtm :) Thanks @iblancasa

cmd/otel-allocator/Dockerfile
iblancasa: I'm not sure about that files you're referring. The problem with that is that other files that you don't want to be copied, will be copied too (and if those files change, the Docker cache will be invalidated at that layer).

But I don't have a strong opinion: I just tried to follow the same approach than in the `Dockerfile` from the root folder. If you think we should follow the `COPY . .` approach, maybe we should create a new task to address that.

frzifus: I don't have a strong opinion either. My thought was, since the overhead of files should normally be quite limited and it is the build stage, it should not be so bad to just copy the complete folder content. At least this way you won't forget to copy something / need to touch the Dockerfile again.
As for caching, I don't think it makes much difference. Since it is only a few text files that are copied.

Aneurysm9: since this is only the build stage I think it best to copy the entire context.  That way we don't need to remember to update this file if we add new packages or `.go` files in the root.

cmd/otel-allocator/Dockerfile
```diff
@@ -1,26 +1,34 @@
...
-######## Start a new stage from scratch #######
-FROM alpine:latest  
-
-RUN apk --no-cache add ca-certificates
-
-WORKDIR /root/
-
+# Use distroless as minimal base image to package the manager binary
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+FROM gcr.io/distroless/static:nonroot
```
Aneurysm9, line 28: Any chance this can be done `FROM scratch` instead?  Using `gcr.io/distroless` is likely going to cause security review headaches for me.  I think all we need is to copy certs and (maybe) tzdb from the bases `alpine` image.

iblancasa: @Aneurysm9 I'll take a look and test it using KiND and OpenShift :D

pavolloffay (approved):

PR #1347, opened 2023-01-04
[target-allocator] Change the github action to match the operator

### LINKED ISSUE #1298: Target Allocator versioning
Currently, the target allocator go module has no version/releases:
https://pkg.go.dev/github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator?tab=versions

The image version is `v0.66.0`, which matches the operator version, but @jaronoff97 pointed out that a new image is built with any merge to `main` (so `v0.66.0` has already been rebuilt [many times](https://github.com/open-telemetry/opentelemetry-operator/pkgs/container/opentelemetry-operator%2Ftarget-allocator/versions)).  This means someone can start using an image with changes not documented in the changelog for that version.

The [Changelog](https://github.com/open-telemetry/opentelemetry-operator/blob/main/CHANGELOG.md) is also kind of confusing, since it's unclear what changes are for the operator vs target allocator.

Is there an easier way to manage builds and versions for the target allocator? Does it make sense for it to have its own changelog?

jaronoff97:
Closes #1298

This changes the target allocator image building process to set the versions the same the operator does. This means that when a new release is created, the target allocator will be made using that version tag. This also will have a "main" release like the operator as well.

### DIFF
--- a/.chloggen/1298-change-ta-build-proc.yaml
+++ b/.chloggen/1298-change-ta-build-proc.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: target-allocator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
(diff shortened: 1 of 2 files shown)


pavolloffay (approved):

kristinapathak (approved):

jaronoff97:
You can see that this worked in my local repository! 

Screen Shot 2023-01-04 at 3 12 10 PM

PR #1360, opened 2023-01-10
[target allocator] Fix concurrent map access in discoverer

### LINKED ISSUE #1359: [target allocator] Data race when writing / reading `jobToScrapeConfig` in `Discoverer`
When running the target allocator in our test environment, it would often crash for us with the following output (truncated to most relevant part):

```
{"level":"info","ts":1672998710.9367316,"logger":"allocator","msg":"Successfully started a collector pod watcher","component":"opentelemetry-targetallocator"}
fatal error: concurrent map read and map write

goroutine 6072 [running]:
reflect.mapaccess_faststr(0x23078e8?, 0x7f8e9acbcf38?, {0xc17459e450?, 0x34fba60?})
	/usr/local/go/src/runtime/map.go:1343 +0x1e
reflect.Value.MapIndex({0x2966a20?, 0xc000c80fc0?, 0xc17aa5f4a0?}, {0x273e3c0, 0xc1fd84bc30, 0x98})
	/usr/local/go/src/reflect/value.go:1664 +0xc5
github.com/mitchellh/hashstructure.(*walker).visit(0xc190f7f790, {0x2966a20?, 0xc000c80fc0?, 0xc000291d60?}, 0x0)
	/go/pkg/mod/github.com/mitchellh/hashstructure@v1.1.0/hashstructure.go:225 +0x1685
github.com/mitchellh/hashstructure.Hash({0x2966a20?, 0xc000c80fc0}, 0x8?)
	/go/pkg/mod/github.com/mitchellh/hashstructure@v1.1.0/hashstructure.go:108 +0x213
github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/server.(*Server).ScrapeConfigsHandler(0xc000165860, {0x34f3110, 0xc3e789e460}, 0xc1fd84b5d0?)
	/app/server/server.go:93 +0x5c
net/http.HandlerFunc.ServeHTTP(0x100?, {0x34f3110?, 0xc3e789e460?}, 0x0?)
	/usr/local/go/src/net/http/server.go:2109 +0x2f
github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/server.(*Server).PrometheusMiddleware.func1({0x34f3110, 0xc3e789e460}, 0xc5d6c20c60?)
	/app/server/server.go:138 +0x13f
net/http.HandlerFunc.ServeHTTP(0xc190c4ba00?, {0x34f3110?, 0xc3e789e460?}, 0x800?)
	/usr/local/go/src/net/http/server.go:2109 +0x2f
github.com/gorilla/mux.(*Router).ServeHTTP(0xc0000f43c0, {0x34f3110, 0xc3e789e460}, 0xc190c4b900)
	/go/pkg/mod/github.com/gorilla/mux@v1.8.0/mux.go:210 +0x1cf
net/http.serverHandler.ServeHTTP({0xc5d6c20ae0?}, {0x34f3110, 0xc3e789e460}, 0xc190c4b900)
	/usr/local/go/src/net/http/server.go:2947 +0x30c
net/http.(*conn).serve(0xc0a4573360, {0x34f46c0, 0xc0004a4180})
	/usr/local/go/src/net/http/server.go:1991 +0x607
created by net/http.(*Server).Serve
	/usr/local/go/src/net/http/server.go:3102 +0x4db
```

After taking a look, it seems to me there's concurrent writing in `Discoverer`s `ApplyConfig` to the `jobToScrapeConfig` map and reading of this map via `GetScrapeConfigs()` in the scrape config handler during the hashing of the config. 

I'd suggest to protect the map with a mutex / return only a copy of the map for reading.

Somewhat related to similar issue in https://github.com/open-telemetry/opentelemetry-operator/pull/1040

matej-g:
Protect the job to scrape config map with a mutex and when returning the config create a copy of the map, to prevent concurrent access.

Resolves #1359

### DIFF
--- a/cmd/otel-allocator/target/discovery.go
+++ b/cmd/otel-allocator/target/discovery.go
@@ -15,6 +15,8 @@
 package target
 
 import (
+	"sync"
+
 	"github.com/go-logr/logr"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
@@ -34,12 +36,14 @@ var (
 )
 
 type Discoverer struct {
-	log               logr.Logger
-	manager           *discovery.Manager
-	close             chan struct{}
-	configsMap        map[allocatorWatcher.EventSource]*config.Config
+	log        logr.Logger
+	manager    *discovery.Manager
+	close      chan struct{}
+	configsMap map[allocatorWatcher.EventSource]*config.Config
+	hook       discoveryHook
+
+	configMtx         sync.RWMutex
 	jobToScrapeConfig map[string]*config.ScrapeConfig
-	hook              discoveryHook
 }
 
 type discoveryHook interface {
@@ -57,8 +61,17 @@ func NewDiscoverer(log logr.Logger, manager *discovery.Manager, hook discoveryHo
 	}
 }
 
+// GetScrapeConfigs retrieves current jobs and return a new copy of a map
+// for use outside of discoverer.
 func (m *Discoverer) GetScrapeConfigs() map[string]*config.ScrapeConfig {
-	return m.jobToScrapeConfig
+	m.configMtx.RLock()
+	defer m.configMtx.RUnlock()
+
+	jobToScrapeConfigCopy := make(map[string]*config.ScrapeConfig, len(m.jobToScrapeConfig))
+	for jobName, config := range m.jobToScrapeConfig {
+		jobToScrapeConfigCopy[jobName] = config
+	}
+	return jobToScrapeConfigCopy
 }
 
 func (m *Discoverer) ApplyConfig(source allocatorWatcher.EventSource, cfg *config.Config) error {
@@ -69,7 +82,11 @@ func (m *Discoverer) ApplyConfig(source allocatorWatcher.EventSource, cfg *confi
 
 	for _, value := range m.configsMap {
 		for _, scrapeConfig := range value.ScrapeConfigs {
+
+			m.configMtx.Lock()
 			m.jobToScrapeConfig[scrapeConfig.JobName] = scrapeConfig
+			m.configMtx.Unlock()
+
 			discoveryCfg[scrapeConfig.JobName] = scrapeConfig.ServiceDiscoveryConfigs
 			relabelCfg[scrapeConfig.JobName] = scrapeConfig.RelabelConfigs
 		}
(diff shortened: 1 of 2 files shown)


frzifus (approved):
Good catch :)

cmd/otel-allocator/target/discovery.go
```diff
@@ -57,8 +61,17 @@ func NewDiscoverer(log logr.Logger, manager *discovery.Manager, hook discoveryHo
...
 }
 
+// GetScrapeConfigs retrieves current jobs and return a new copy of a map
+// for use outside of discoverer.
 func (m *Discoverer) GetScrapeConfigs() map[string]*config.ScrapeConfig {
-	return m.jobToScrapeConfig
+	m.configMtx.RLock()
+	defer m.configMtx.RUnlock()
+
+	jobToScrapeConfigCopy := make(map[string]*config.ScrapeConfig, len(m.jobToScrapeConfig))
```
jaronoff97, line 70: We actually got rid of this knowing there is a risk of a race condition, optimizing for memory performance to avoid making a copy constantly. ([see this PR for more](https://github.com/open-telemetry/opentelemetry-operator/pull/1293#issue-1469943346)) Would it be possible to keep track of when the map changes to avoid allocating this memory on every call to the server?

jaronoff97: This call is made every 60s (configurable) from the collector, and if you have a pool of 10+ collectors the memory usage on the TA increases heavily. We need to find a better solution for this codepath overall, and I'm planning on talking about it at our next SIG meeting.

jaronoff97 (changes requested):

cmd/otel-allocator/target/discovery.go
matej-g: Got it, then it is indeed a bit trickier. I'm thinking whether we could directly return hash for comparing so would copy map only if changes were detected. Alternatively, I'm wondering whether the scrape config could be periodically checked / updated instead of on every `ScrapeConfigHandler` invocation, but I might be missing more context as to whether this is a possible solution.

Either way, happy to discuss this at SIG and find a proper solution, as this is currently blocking us from using target allocator.

kristinapathak (approved):

matej-g:
Hey folks,
After some of the comments here and chat with @jaronoff97 I created an alternative way to solve this in https://github.com/open-telemetry/opentelemetry-operator/pull/1413. I'd love to hear your thoughts, thanks!

jaronoff97:
@matej-g I think we can close this PR in favor of your new one. Please let me know if you'd like this reopened.

PR #1365, opened 2023-01-12
[chore] Fix E2E autoscale test for OpenShift

### LINKED ISSUE #1364: autoscale E2E test doesn't pass in OpenShift
The `autoscale` E2E is failing when is runnned in an OpenShift cluster.

The test creates two OpenTelemetry Collectors. Each one deploys one Horizontal Pod Autoscaler:
```bash
$ kubectl get pods
NAME                                 REFERENCE                                         TARGETS   MINPODS   MAXPODS   REPLICAS   AGE
simplest-collector                   OpenTelemetryCollector/simplest                   0%/90%    1         2         1          10m
simplest-set-utilization-collector   OpenTelemetryCollector/simplest-set-utilization   0%/50%    1         2         1          10m
```

Later, it tries to match them with the CRs from `tests/e2e/autoscale/00-assert.yaml`. Those CRs check for HPAs whose `apiVersion` are `autoscaling/v1` (`simplest-collector`) and `autoscaling/v2beta2` (`simplest-set-utilization-collector`). In OpenShift (4.11) both are created with an `apiVersion` with the value `autoscaling/v2`.

## Steps to reproduce
1. Create an OpenShift cluster (4.11 in my case)
2. Build and deploy the OpenTelemetry Collector Operator
3. Run the `autoscale` E2E test (`$ kuttl test --test autoscale`)

Also, the test can be flaky since it always waits for a fixed CPU % utilization (that cannot be reached depending on the resources of the machine where the tests are running).

iblancasa:

Fixes #1364
Also, this PR makes the test more reliable.

### DIFF
--- a/tests/e2e/autoscale/01-check-simplest-collector-hpa.yaml
+++ b/tests/e2e/autoscale/01-check-simplest-collector-hpa.yaml
@@ -1,0 +1,4 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: go run ./wait-until-hpa-ready.go --hpa simplest-collector
--- a/tests/e2e/autoscale/04-delete.yaml
+++ b/tests/e2e/autoscale/04-delete.yaml
@@ -1,0 +1,7 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+delete:
+- apiVersion: batch/v1
+  kind: Job
+  metadata:
+    name: tracegen
--- a/tests/e2e/autoscale/wait-until-hpa-ready.go
+++ b/tests/e2e/autoscale/wait-until-hpa-ready.go
@@ -1,0 +1,109 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/spf13/pflag"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/homedir"
+)
+
+func main() {
+	var hpaName string
+	var timeout int
+	var kubeconfigPath string
+
+	defaultKubeconfigPath := filepath.Join(homedir.HomeDir(), ".kube", "config")
+
+	pflag.IntVar(&timeout, "timeout", 600, "The timeout for the check.")
+	pflag.StringVar(&hpaName, "hpa", "", "HPA to check")
+	pflag.StringVar(&kubeconfigPath, "kubeconfig-path", defaultKubeconfigPath, "Absolute path to the KubeconfigPath file")
+	pflag.Parse()
+
+	if len(hpaName) == 0 {
+		fmt.Println("hpa flag is mandatory")
+		os.Exit(1)
+	}
+
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+	if err != nil {
+		fmt.Printf("Error reading the kubeconfig: %s\n", err)
+		os.Exit(1)
+	}
+
+	client, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	namespace, err := client.CoreV1().Namespaces().Get(context.Background(), os.Getenv("NAMESPACE"), metav1.GetOptions{})
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	hpaClientV2 := client.AutoscalingV2().HorizontalPodAutoscalers(namespace.Name)
+	hpaClientV1 := client.AutoscalingV1().HorizontalPodAutoscalers(namespace.Name)
+
+	pollInterval := time.Second
+
+	// Search in v2 and v1 for an HPA with the given name
+	err = wait.Poll(pollInterval, 0, func() (done bool, err error) {
+		hpav2, err := hpaClientV2.Get(
+			context.Background(),
+			hpaName,
+			metav1.GetOptions{},
+		)
+		if err != nil {
+			hpav1, err := hpaClientV1.Get(
+				context.Background(),
+				hpaName,
+				metav1.GetOptions{},
+			)
+			if err != nil {
+				fmt.Printf("HPA %s not found\n", hpaName)
+				return false, nil
+			}
+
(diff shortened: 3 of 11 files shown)


frzifus (approved):
Until when do we support `hpav1`?

iblancasa:
@iblancasa while we support K8s 1.23, if I'm not wrong

hack/wait-until-hpa-ready/main.go
```diff
@@ -0,0 +1,108 @@
...
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	fmt.Println(hpaName, "is ready!")
```
pavolloffay, line 107: personally, I haven't seen `fmt.Println` to be used this way. Consider using `fmt.Printf`

hack/wait-until-hpa-ready/main.go
```diff
@@ -0,0 +1,108 @@
...
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	hpaClientV2 := client.AutoscalingV2().HorizontalPodAutoscalers(namespace.Name)
+	hpaClientV1 := client.AutoscalingV1().HorizontalPodAutoscalers(namespace.Name)
+
+	pollInterval := time.Second
+
+	err = wait.Poll(pollInterval, 0, func() (done bool, err error) {
```
pavolloffay, line 72: a comment here would be helpful. Does the it try to find HPA v1 and v2 with the same name?

tests/e2e/autoscale/04-delete.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kuttl.dev/v1beta1
```
pavolloffay, line 1: Why is this needed? The https://github.com/open-telemetry/opentelemetry-operator/blob/0232b80b1885bf616d4cc5134c6422ed51f7199b/tests/e2e/autoscale/01-install.yaml#L4 should be cleaned automatically.

tests/e2e/autoscale/01-check-simplest-collector-hpa.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: go run ../../../hack/wait-until-hpa-ready/main.go --hpa simplest-collector
```
pavolloffay, line 4: I would maybe prefer to host the "script" in this directory along the tests since it's used only here.

pavolloffay:
From the issue

>  Later, it tries to match them with the CRs from tests/e2e/autoscale/00-assert.yaml. Those CRs check for HPAs whose apiVersion are autoscaling/v1 (simplest-collector) and autoscaling/v2beta2 (simplest-set-utilization-collector). In OpenShift (4.11) both are created with an apiVersion with the value autoscaling/v2

@iblancasa do we know the root cause why only `autoscaling/v2` was created? The OCP 4.11 is based on k8s 1.24. `autoscaling/v2beta1` was removed 1.25

hack/wait-until-hpa-ready/main.go
iblancasa: Yes. The code tries to find one HPA with the given name. This is because, depending on the version where the OTEL Operator is run, the HPA is created using V1 or V2.

tests/e2e/autoscale/04-delete.yaml
iblancasa: The problem is: depending on the cluster where you are running the test, the previous `duration` parameter provided to `tracegen` can be not enough to trigger the HPA and scale the collector.

With this approach, we ran `tracegen` until the number of replicas is increased. Later, we stop `tracegen` to reduce the metrics and make the HPA to scale down.

iblancasa:
> From the issue
> 
> > Later, it tries to match them with the CRs from tests/e2e/autoscale/00-assert.yaml. Those CRs check for HPAs whose apiVersion are autoscaling/v1 (simplest-collector) and autoscaling/v2beta2 (simplest-set-utilization-collector). In OpenShift (4.11) both are created with an apiVersion with the value autoscaling/v2
> 
> @iblancasa do we know the root cause why only `autoscaling/v2` was created? The OCP 4.11 is based on k8s 1.24. `autoscaling/v2beta1` was removed 1.25

I didn't research too much but I think it can be related to this comment in `00-install.yaml`:
```yaml
# TODO: these tests use .Spec.MaxReplicas and .Spec.MinReplicas. These fields are
# deprecated and moved to .Spec.Autoscaler. Fine to use these fields to test that old CRD is
# still supported but should eventually be updated.
```
If you want, I can create a new issue for that.

tests/e2e/autoscale/wait-until-hpa-ready.go
```diff
@@ -0,0 +1,109 @@
...
+		os.Exit(1)
+	}
+
+	hpaClientV2 := client.AutoscalingV2().HorizontalPodAutoscalers(namespace.Name)
+	hpaClientV1 := client.AutoscalingV1().HorizontalPodAutoscalers(namespace.Name)
+
+	pollInterval := time.Second
+
+	// Search in v2 and v1 for an HPA with the given name
+	err = wait.Poll(pollInterval, 0, func() (done bool, err error) {
```
pavolloffay, line 73: I am still not sure why actually v1 and v2 HPAs are created in a test. 

My understanding is that only a single HPA version should be used in a given cluster. Could you please explain why both are created?

iblancasa: I didn't write the test. I'm just trying to make it work on OpenShift. But this is what I found while checking for the purpose of this E2E test:
* The test creates 2 `OpenTelemetryCollector` instances to test 2 ways of creating HPAs. From the `00-install.yaml` file:
```yaml
# This creates two different deployments:
#   * The first one will be used to see if we scale properly
#   * The second is to check the targetCPUUtilization option
```
* This creates 2 HPAs (I wait for their creation and metrics reporting in steps 1 and 2)
* We start `tracegen` in step 3 and wait for one of the `OpenTelemetryCollector` instances to scale up to 2 replicas
* We remove the `tracegen` deployment to stop reporting traces in step 4
* Wait until the `OpenTelemetryCollector` scales down in step 5

When the HPAs are created, they will be created using `autoscaling/v1` or `autoscaling/v2beta2`. If you check how this test was written before my changes, you can see how in `00-assert.yaml` , the test tries to assert the `simplest-collector` HPA with `autoscaling/v1` and the `simplest-set-utilization-collector` HPA with `autoscaling/v2beta2`.

When I ran this in OpenShift 4.11, both of them were created using `autoscaling/v2beta2`  (as I pointed [in this comment](https://github.com/open-telemetry/opentelemetry-operator/issues/1364#issue-1530344787)). 

So, since in KUTTL there is no way to conditionally check for one resource or another, I created the `wait-until-hpa-ready.go` script to (given a name) dynamically:
* Look for an HPA in the `autoscaling/v2beta2` API. If found, check if the HPA status is different from unknown 
* If the HPA was not found in `autoscaling/v2beta2`, look for it in `autoscaling/v1`. If found, check if the HPA status is different from unknown

Another thing: why the HPAs are created using different `autoscaling` API versions (as we can see in `00-assert.yaml`) in the Kubernetes versions tested during the CI? I think this is because one is setting the `.spec.minReplicas` and `.spec.maxReplicas` values and the other is setting them in `.spec.autoscaler` (as the comment in `00-install.yaml` explains). If you want, I can do a deeper investigation about why this happens, but in a separate issue since is not related to the current PR.

pavolloffay:
@iblancasa could you please fix the CI?

iblancasa:
I broke something in my branch. Fixing...

iblancasa:
Fixed!

pavolloffay (approved):
Let's merge this and book a ticket to make sure only a single HPA version is used by the operator for a given k8s version.

PR #1373, opened 2023-01-16
[chore] Add SKIP_DEPLOYMENT make variable

### LINKED ISSUE #1372: Add a variable to makefile to skip the build and deployment of the Operator to speed up running the E2E tests
When fixing issues with the E2E tests, most of the time is not necessary to build, push and deploy the operator again. Having a variable in the Makefile to disable these steps will be really helpful in saving time and disk space.

iblancasa:

Fixes #1372

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,7 @@ GOTEST_OPTS=-race -v
 endif
 
 START_KIND_CLUSTER ?= true
+SKIP_DEPLOYMENT ?= false
 
 KUBE_VERSION ?= 1.24
 KIND_CONFIG ?= kind-$(KUBE_VERSION).yaml
@@ -122,8 +123,12 @@ set-image-controller: manifests kustomize
 # Deploy controller in the current Kubernetes context, configured in ~/.kube/config
 .PHONY: deploy
 deploy: set-image-controller
+ifeq (true,$(SKIP_DEPLOYMENT))
+	echo "Skip operator deployment since SKIP_DEPLOYMENT is true"
+else
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
 	go run hack/check-operator-ready.go 300
+endif
 
 # Undeploy controller in the current Kubernetes context, configured in ~/.kube/config
 .PHONY: undeploy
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -191,12 +196,20 @@ scorecard-tests: operator-sdk
...
+	echo "Skipping container operator build since SKIP_DEPLOYMENT is true"
+else
 	docker buildx build --load --platform linux/${ARCH} -t ${IMG} --build-arg VERSION_PKG=${VERSION_PKG} --build-arg VERSION=${VERSION} --build-arg VERSION_DATE=${VERSION_DATE} --build-arg OTELCOL_VERSION=${OTELCOL_VERSION} --build-arg TARGETALLOCATOR_VERSION=${TARGETALLOCATOR_VERSION} --build-arg AUTO_INSTRUMENTATION_JAVA_VERSION=${AUTO_INSTRUMENTATION_JAVA_VERSION}  --build-arg AUTO_INSTRUMENTATION_NODEJS_VERSION=${AUTO_INSTRUMENTATION_NODEJS_VERSION} --build-arg AUTO_INSTRUMENTATION_PYTHON_VERSION=${AUTO_INSTRUMENTATION_PYTHON_VERSION} --build-arg AUTO_INSTRUMENTATION_DOTNET_VERSION=${AUTO_INSTRUMENTATION_DOTNET_VERSION} .
+endif
 
 # Push the container image, used only for local dev purposes
 .PHONY: container-push
 container-push:
+ifeq (true,$(SKIP_DEPLOYMENT))
+	echo "Skipping container operator push since SKIP_DEPLOYMENT is true"
```
frzifus, line 209: Isnt `container-push` != `deployment` anyway?

frzifus (commented):
How will the variable be used afterwards?
```
1. SKIP_DEPLOYMENT=true make generate bundle container container-push deploy
2. make generate bundle container 
```
Would line 1 and 2 be the same?

Makefile
iblancasa: It is but: if you want to skip the deployment and the build was skipped, it doesn't make sense to push the container image to the repository, right?

frzifus: In that case the checksum would stay the same anyway and not much would happen.

Personally, when I read `SKIP_DEPLOYMENT=true make ... container-push` I expect a new build in my registry and that no rollout will be executed.

But I am more concerned that we add logic that undermines explicit parameters.

Like here:
```
SKIP_DEPLOYMENT=true make ... container-push
# is equal to
make ...
```

iblancasa: But if I didn't build the image, I would not have any image to push. That is an example if I want to run the tests about a release, for instance, where I will use the image provided by the manifest.

if the problem is the name of the variable, could you provide a better name? Or: should I create 3 variables (one to skip deployment, another to skip build and another to skip push??

iblancasa:
@frzifus any other comments? Thanks!

iblancasa:
@frzifus is there anything pending from this PR?

Makefile
frzifus: Actually i still do not see the need for it. Why can't the desired make rule be removed as an argument?

Thats what i tried to say in the comment above. Instead of `SKIP_SOMETHING_1=true make something1 something2` we could call `make something2` right?

frzifus:
https://github.com/open-telemetry/opentelemetry-operator/pull/1373#discussion_r1086341437

Makefile
iblancasa: Ok, now I see your point.
The problem is: if the `prepare-e2e` rule is modified, we need to modify the developer's workflow. While, with this, it is not needed.

frzifus: >  The problem is: if the prepare-e2e rule is modified, we need to modify the developer's workflow. While, with this, it is not needed.

But why?

---

Whats your thoughts on that @pavolloffay @rubenvp8510 ?

iblancasa: It's ok. I'll close the PR since is not as important and we're spending too much time on it.

PR #1375, opened 2023-01-18
Add basic TA readme documentation

jpkrohling:
I struggled to get the target allocator to work, mostly because of the lack of documentation. After reading the sources and talking to some folks (thanks @jaronoff97 and @Aneurysm9), I understood how to get it working and how it works behind the scenes. I tried to record my learnings in a new section as part of the readme, keeping it simple and with end-users in mind.

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,6 +313,75 @@ You can configure the OpenTelemetry SDK for applications which can't currently b
 instrumentation.opentelemetry.io/inject-sdk: "true"
 ```
 
+### Target Allocator
+
+The OpenTelemetry Operator comes with an optional component, the Target Allocator (TA). When creating an OpenTelemetryCollector Custom Resource (CR) and setting the TA as enabled, the Operator will create a new deployment and service to serve specific `http_sd_config` directives for each Collector pod as part of that CR. It will also change the Prometheus receiver configuration in the CR, so that it uses the `http_sd_config` from the TA. The following example shows how to get started with the Target Allocator:
+
+```yaml
+apiVersion: opentelemetry.io/v1alpha1
+kind: OpenTelemetryCollector
+metadata:
+  name: collector-with-ta
+spec:
+  mode: statefulset
+  targetAllocator:
+    enabled: true
+  config: |
+    receivers:
+      prometheus:
+        config:
+          scrape_configs:
+          - job_name: 'otel-collector'
+            scrape_interval: 10s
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -313,6 +313,75 @@ You can configure the OpenTelemetry SDK for applications which can't currently b
...
+
+    service:
+      pipelines:
+        traces:
+          receivers: [prometheus]
+          processors: []
+          exporters: [logging]
+```
+
+Behind the scenes, the OpenTelemetry Operator will convert the Collector’s configuration after the reconciliation into the following:
```
frzifus, line 350: The Operator also updates the [Prometheus receiver options](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/72799c31e8156923b7e080861117af84044325cc/receiver/prometheusreceiver#opentelemetry-operator) right?

---
grafik

frzifus (commented):
Thanks @jpkrohling :)

README.md
```diff
@@ -313,6 +313,75 @@ You can configure the OpenTelemetry SDK for applications which can't currently b
 instrumentation.opentelemetry.io/inject-sdk: "true"
 ```
 
+### Target Allocator
+
+The OpenTelemetry Operator comes with an optional component, the Target Allocator (TA). When creating an OpenTelemetryCollector Custom Resource (CR) and setting the TA as enabled, the Operator will create a new deployment and service to serve specific `http_sd_config` directives for each Collector pod as part of that CR. It will also change the Prometheus receiver configuration in the CR, so that it uses the `http_sd_config` from the TA. The following example shows how to get started with the Target Allocator:
```
blakeromano, line 318: It may be worth mentioning how this is a seperate service that could be deployed separately as well outside of the OpenTelemetry CRD. (It seems like docs about TA really should be more within the broader Otel docs rather than an Operator thing)

Aneurysm9: It could possibly be deployed to a k8s cluster without the operator, but it depends on the k8s APIs and information provided by the operator in its configuration to identify collector instances amongst which it should allocate targets.  That's a fairly advanced use case I'm not sure we should try to document at this time as it is likely to add more confusion.

I agree that having documentation on the otel.io site would be good, but I'd suggest we start here and then work with the comms SIG to figure out what should live there and how to present it.

frzifus (approved):

PR #1383, opened 2023-01-19
[target-allocator] Replace deprecated `gorilla/mux` dependency with `gin`

### LINKED ISSUE #1352: [target allocator] Replace deprecated `gorilla/mux` dependency
The allocator server uses the router from `gorilla/mux`, however the project has been deprecated recently (https://github.com/gorilla#gorilla-toolkit) and is no longer actively maintained. We should replace it with an alternative that's still maintained.

matej-g:
Resolves #1352.

This change aims to replace the deprecated `gorilla/mux` with `gin` while maintaining the functionality of current solution.

Note: Due to difference in convention for path variables used by both frameworks (`gorilla/mux` using curly braces, .e.g `/jobs/{job_id}/targets` vs. `gin` using colon e.g. `/jobs/:job_id/targets`), the only change will be in the reported metric label for the jobs path.

### DIFF
--- a/cmd/otel-allocator/server/server.go
+++ b/cmd/otel-allocator/server/server.go
@@ -23,8 +23,8 @@ import (
 	"time"
 
 	yaml2 "github.com/ghodss/yaml"
+	"github.com/gin-gonic/gin"
 	"github.com/go-logr/logr"
-	"github.com/gorilla/mux"
 	"github.com/mitchellh/hashstructure"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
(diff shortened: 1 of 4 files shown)


cmd/otel-allocator/server/server.go
```diff
@@ -64,12 +64,16 @@ func NewServer(log logr.Logger, allocator allocation.Allocator, discoveryManager
 		discoveryManager: discoveryManager,
 		compareHash:      uint64(0),
 	}
-	router := mux.NewRouter().UseEncodedPath()
+
+	router := gin.Default()
+	router.UseRawPath = true
+	router.UnescapePathValues = false
```
frzifus, line 70: Why is it explicit set to false?

frzifus (approved):
Thanks @matej-g for taking care :+1:  Actually i never used gin and the added context, feels a bit strage - but aside of that it looks good to me. :)

Whats your thoughts @open-telemetry/operator-ta-maintainers ?

cmd/otel-allocator/server/server.go
matej-g: It is set to `true` by default so I'm setting it to `false` explicitly here. Basically these two options should provide same functionality as we had with `gorilla/mux` and [`UseEncodedPath`](https://pkg.go.dev/github.com/gorilla/mux#Router.UseEncodedPath). 

If it helps I can add a comment, although if you consult the comments for the [`gin.Engine`](https://pkg.go.dev/github.com/gin-gonic/gin#Engine) type it should be clearer.

pavolloffay (approved):

jaronoff97:
I'm rerunning the unit tests with @kristinapathak 's recent change [here](https://github.com/open-telemetry/opentelemetry-operator/pull/1357) just to be sure we have some added confidence on changing the framework :)

jaronoff97 (approved):

kristinapathak (approved):

PR #1395, opened 2023-01-25
Inject otelcol sidecar into any namespace

### LINKED ISSUE #199: Support sidecar injection in another pod namespace different from the otel colletor instance.
If I have an OpenTelemetry collector on sidecar mode in one namespace (eg, test1) but want a pod on other namespace (test2) to be injected, this is not actually possible with the opentelemetry-operator.

Steps to reproduce:
1) Install opentelemetry-operator
2) Create namespace test1 `kubectl create namespace test1`
2) Create an opentelemetry collector instance mode=sidecar:

```
kubectl apply -n test1 -f - <<EOF
kind: OpenTelemetryCollector
metadata:
  name: simplest
spec:
  mode: sidecar
  config: |
    receivers:
      jaeger:
        protocols:
          grpc:
    processors:
      queued_retry:

    exporters:
      logging:

    service:
      pipelines:
        traces:
          receivers: [jaeger]
          processors: [queued_retry]
          exporters: [logging]
EOF
```
3) Create a pod in namespace test1
```
kubectl apply -n test1 -f - <<EOF
apiVersion: v1
kind: Pod
metadata:
  name: myapp
  labels:
    app: myapp
  annotations:
    sidecar.opentelemetry.io/inject: "true"
spec:
  containers:
  - name: myapp
    image: jaegertracing/vertx-create-span:operator-e2e-tests
    ports:
      - containerPort: 8080
        protocol: TCP
EOF
```
4) Verify that sidecar is injected 
```
kubectl get pods -n test1
NAME    READY   STATUS              RESTARTS   AGE
myapp   0/2     ContainerCreating   0          14s
```

5) Create namespace test2 `kubectl create namespace test2`
6) Create pod on namespace `test2
````
kubectl apply -n test1 -f - <<EOF
apiVersion: v1
kind: Pod
metadata:
  name: myapp
  labels:
    app: myapp
  annotations:
    sidecar.opentelemetry.io/inject: "true"
spec:
  containers:
  - name: myapp
    image: jaegertracing/vertx-create-span:operator-e2e-tests
    ports:
      - containerPort: 8080
        protocol: TCP
EOF
```
7)Sidecar is not injected
```
kubectl get pods -n test2
NAME    READY   STATUS    RESTARTS   AGE
myapp   1/1     Running   0          5s
```

pavolloffay:

Resolves #199 

* This PR adds support for injecting `otelcol` sidecar into any namespace - currently the sidecar can be injected into workloads that are in the same namespace as the `otelcol` CR.
* The injection is enabled via the same annotation - ` sidecar.opentelemetry.io/inject: "default/simplest"` (`default` is the namespace)
* The PR changes how the sidecar config is handled - now the collector config is passed an env var to to the collector container.  ---- This approach was chosen to eliminate issues with cleaning the config maps needed for the collector.

Tested on with:

```yaml
kubectl apply -f - <<EOF
kind: OpenTelemetryCollector
metadata:
  name: simplest
spec:
  mode: sidecar
  config: |
    receivers:
      jaeger:
        protocols:
          thrift_compact:
    processors:

    exporters:
      logging:

    service:
      pipelines:
        traces:
          receivers: [jaeger]
          processors: []
          exporters: [logging]
EOF
```

```yaml
kubectl apply -f - <<EOF
apiVersion: v1
kind: Pod
metadata:
  name: myapp
  annotations:
    sidecar.opentelemetry.io/inject: "default/simplest"
spec:
  containers:
  - name: myapp
    image: jaegertracing/vertx-create-span:operator-e2e-tests
    ports:
      - containerPort: 8080
        protocol: TCP
EOF
```

Result:
```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"Pod","metadata":{"annotations":{"sidecar.opentelemetry.io/inject":"default/simplest"},"name":"myapp","namespace":"ploffay"},"spec":{"containers":[{"image":"jaegertracing/vertx-create-span:operator-e2e-tests","name":"myapp","ports":[{"containerPort":8080,"protocol":"TCP"}]}]}}
    sidecar.opentelemetry.io/inject: default/simplest
  creationTimestamp: "2023-02-01T10:54:42Z"
  labels:
    sidecar.opentelemetry.io/injected: default.simplest
  name: myapp
  namespace: ploffay
  resourceVersion: "6276"
  uid: 232aea7f-3630-4255-b259-43355a404013
spec:
  containers:
  - image: jaegertracing/vertx-create-span:operator-e2e-tests
    imagePullPolicy: IfNotPresent
    name: myapp
    ports:
    - containerPort: 8080
      protocol: TCP
    resources: {}
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
    volumeMounts:
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: kube-api-access-hllwm
      readOnly: true
  - args:
    - --config=env:OTEL_CONFIG
    env:
    - name: POD_NAME
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.name
    - name: OTEL_CONFIG
      value: |
        receivers:
          jaeger:
            protocols:
              thrift_compact:
        processors:

        exporters:
          logging:

        service:
          pipelines:
            traces:
              receivers: [jaeger]
              processors: []
              exporters: [logging]
    - name: OTEL_RESOURCE_ATTRIBUTES_POD_NAME
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.name
    - name: OTEL_RESOURCE_ATTRIBUTES_POD_UID
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.uid
    - name: OTEL_RESOURCE_ATTRIBUTES_NODE_NAME
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: spec.nodeName
    - name: OTEL_RESOURCE_ATTRIBUTES
      value: k8s.namespace.name=ploffay,k8s.node.name=$(OTEL_RESOURCE_ATTRIBUTES_NODE_NAME),k8s.pod.name=$(OTEL_RESOURCE_ATTRIBUTES_POD_NAME),k8s.pod.uid=$(OTEL_RESOURCE_ATTRIBUTES_POD_UID)
    image: ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector:0.68.0
    imagePullPolicy: IfNotPresent
    name: otc-container
    ports:
    - containerPort: 6831
      name: jaeger-thrift-c
      protocol: UDP
    - containerPort: 8888
      name: metrics
      protocol: TCP
    resources: {}
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
    volumeMounts:
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: kube-api-access-hllwm
      readOnly: true
  dnsPolicy: ClusterFirst
  enableServiceLinks: true
  nodeName: minikube
  preemptionPolicy: PreemptLowerPriority
  priority: 0
  restartPolicy: Always
  schedulerName: default-scheduler
  securityContext: {}
  serviceAccount: default
  serviceAccountName: default
  terminationGracePeriodSeconds: 30
  tolerations:
  - effect: NoExecute
    key: node.kubernetes.io/not-ready
    operator: Exists
    tolerationSeconds: 300
  - effect: NoExecute
    key: node.kubernetes.io/unreachable
    operator: Exists
    tolerationSeconds: 300
  volumes:
  - name: kube-api-access-hllwm
    projected:
      defaultMode: 420
      sources:
      - serviceAccountToken:
          expirationSeconds: 3607
          path: token
      - configMap:
          items:
          - key: ca.crt
            path: ca.crt
          name: kube-root-ca.crt
      - downwardAPI:
          items:
          - fieldRef:
              apiVersion: v1
              fieldPath: metadata.namespace
            path: namespace
```


pkg/sidecar/pod.go
```diff
@@ -33,8 +35,39 @@ const (
...
+	}
+
+	otelColCfg, err := reconcile.ReplaceConfig(otelcol)
+	if err != nil {
+		return pod, err
+	}
+	otelColCfgStr := base64.StdEncoding.EncodeToString([]byte(otelColCfg))
+
+	// add the container
+	pod.Spec.InitContainers = append(pod.Spec.InitContainers, corev1.Container{
```
pavolloffay, line 57: @jaronoff97 @Aneurysm9 could you please take a look and give me 👍🏼 on the approach?

The approach was chosen to avoid copying otelcol config map to the namespace where the injection is enabled and as well avoid issues with cleaning the config map. 

My notes:
* owner reference cannot reference objects from a different namespace
* when a pod is created directly it does not have any owner references

Once we agree I will make the init container image configurable via a flag on the operator (if needed expose it in the CR as well).

pkg/sidecar/pod.go
```diff
@@ -33,8 +35,39 @@ const (
...
+		Name:    "otc-container-config-prepper",
+		Image:   "alpine:latest",
+		Command: []string{"/bin/sh"},
+		Args:    []string{"-c", "echo ${OTEL_CONFIG} | base64 -d > /conf/collector.yaml && cat /conf/collector.yaml"},
```
jaronoff97, line 61: Is the `cat` here just for debugging purposes?

pkg/sidecar/pod.go
```diff
@@ -33,8 +35,39 @@ const (
...
+		Env: []corev1.EnvVar{
+			{
+				Name:  "OTEL_CONFIG",
+				Value: otelColCfgStr,
```
jaronoff97, line 65: Do we have to worry about a maximum string length here?

jaronoff97 (commented):
I just had two questions/concerns, otherwise I think this approach makes sense.

pkg/sidecar/pod.go
pavolloffay: yes

jaronoff97 (approved):

main.go
```diff
@@ -101,6 +102,7 @@ func main() {
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
 	pflag.StringVar(&collectorImage, "collector-image", fmt.Sprintf("ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector:%s", v.OpenTelemetryCollector), "The default OpenTelemetry collector image. This image is used when no image is specified in the CustomResource.")
+	pflag.StringVar(&sidecarConfigPrepperImage, "sidecar-config-prepper-image", "alpine:latest", "\"The default container image that is used to prepare config file for the collector config. The image uses sh, echo, base64 and cat commands")
```
Aneurysm9, line 105: Can we use a more limited container image?  I've already had concerns expressed around the use of busybox for init containers for auto instrumentation, a full alpine container seems like even more surface area with very little used.

pkg/sidecar/pod.go
```diff
@@ -33,8 +35,39 @@ const (
...
+	volumeMount := corev1.VolumeMount{
+		Name:      naming.ConfigMapVolume(),
+		MountPath: "/conf",
+	}
+
+	otelColCfg, err := reconcile.ReplaceConfig(otelcol)
+	if err != nil {
+		return pod, err
+	}
+	otelColCfgStr := base64.StdEncoding.EncodeToString([]byte(otelColCfg))
```
Aneurysm9, line 54: Why the base64 encoding and decoding?  The config should be text, no?

main.go
pavolloffay: Do you have any proposals for the image?

pkg/sidecar/pod.go
pavolloffay: The issue is with new line symbols:

```yaml
  initContainers:
  - args:
    - -c
    - echo ${OTEL_CONFIG} > /conf/collector.yaml && cat /conf/collector.yaml
    command:
    - /bin/sh
    env:
    - name: OTEL_CONFIG
      value: |
        receivers:
          otlp:
            protocols:
              grpc:
              http:
        processors:
          memory_limiter:
            check_interval: 1s
            limit_percentage: 75
            spike_limit_percentage: 15
          batch:
            send_batch_size: 10000
            timeout: 10s

        exporters:
          logging:

        service:
          pipelines:
            traces:
              receivers: [otlp]
              processors: []
              exporters: [logging]

```
produces:

```bash
k logs myapp -c otc-container-config-prepper                                                                                                                                                    ploffay@fedora
receivers: otlp: protocols: grpc: http: processors: memory_limiter: check_interval: 1s limit_percentage: 75 spike_limit_percentage: 15 batch: send_batch_size: 10000 timeout: 10s exporters: logging: service: pipelines: traces: receivers: [otlp] processors: [] exporters: [logging]

```

main.go
pavolloffay: I have switched to `busybox` to align with the auto-instrumentation images.

Aneurysm9: `busybox` is better.  It's got a smaller surface and alignment with the auto-instrumentation images means only one base image to qualify.  Thanks.

pkg/sidecar/pod.go
Aneurysm9: Huh, apparently another example of zsh's superiority as its `echo` deals with this just fine.  With bash you can preserve the newlines by quoting the variable:

```sh
echo "${OTEL_CONFIG}" > /conf/collector.yaml && cat /conf/collector.yaml
```

pavolloffay: yuh, that works. Thanks 👍🏼

pkg/sidecar/pod.go
Aneurysm9: Can you expand on the concerns with using config maps?  While the proposed system appears to work, it's got more moving parts than I like.

pavolloffay: > Can you expand on the concerns with using config maps? 

Correctly implementing the config map approach for the sidecar is tricky:
* The sidecar config map would have to be created in the target namespace.
* A custom cleaning would have to be implemented.  An owner reference to `otelcol` cannot be used (it's a cross namespace) nor owner reference to a pod (a pod UID is not known in the admission webhook).


The init container approach is a cleaner approach because no cleaning is required and the exact sidecar configuration is known (via the `OTEL_CONFIG` env var and `cat config.yaml` in the `initContainer`).  In the current implementation in the `main` branch it is hard to figure out what is the already injected sidecar configuration if the `otelcol` CR changes.


>  but I would prefer a uniform solution for internal and external namespaces. Otherwise there is the possibility that the implementations drift apart.

The init container approach that this PR introduces is used for both internal and external namespaces. There are as well tests (unit+e2e) to ensure the solution works.  The deployment/statefulset/daemonset deployment modes keep using the mounted configmap  as before - which makes perfect sense bc there is only a single version of the "instance".

frzifus (approved):

pkg/sidecar/pod.go
Aneurysm9: Discussed with @pavolloffay OOB and I recommend setting the configuration in an environment variable on the collector container and using the config from there by setting the argument `--config=env:OTEL_CONFIG`.  This avoids the need for an init container simply to copy the config and leaves nothing additional to clean up.

PR #1408, opened 2023-01-31
register pprof endpoints for allocator

seankhliao:
These are generally useful for tracking down performance issues.
Allocator clients are generally trusted, these shouldn't need to be on a separate endpoint.

### DIFF
--- a/.chloggen/allocator-pprof.yaml
+++ b/.chloggen/allocator-pprof.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: target allocator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: register pprof endpoints under /debug/pprof
+
+# One or more tracking issues related to the change
+issues: [188]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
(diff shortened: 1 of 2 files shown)


cmd/otel-allocator/server/server.go
```diff
@@ -78,6 +79,7 @@ func NewServer(log logr.Logger, allocator allocation.Allocator, discoveryManager
 	router.GET("/jobs", s.JobHandler)
 	router.GET("/jobs/:job_id/targets", s.TargetsHandler)
 	router.GET("/metrics", gin.WrapH(promhttp.Handler()))
+	registerPprof(router.Group("/debug/pprof/"))
```
frzifus, line 82: Would it make sense to design the registration in a way that the end point can be switched on and off?

pavolloffay: is there any reason for it? E.g. perf penalty?

seankhliao: As far as I'm aware, just the import has no impact on performance.
When I do see flags for controlling it, it's usually because it's run on a separate port, as the main handler is exposed to untrusted clients or has auth requirements.
The allocator runs with trusted clients, metrics are on the same listener, leading me to the conclusion that there's no need to hide it behind a flag.

jaronoff97 (approved):

jaronoff97:
Thanks for adding this in :) Could you fill out a changelog for this too please?

seankhliao:
i've added one

frzifus (approved):

.chloggen/allocator-pprof.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: target allocator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: register pprof endpoints under /debug/pprof
```
pavolloffay, line 8: Could you please start the sentence with a capital `R` ?

pavolloffay (commented):
LGTM, just a nit for the changelog generator

Issue #1336, opened 2022-12-21
[target-allocator] Use json-iter to marshal json


kristinapathak:
In server handling, we currently have two use cases:
1. marshaling into yaml and then converting yaml to json.
2. marshaling straight to json using the standard library.

There are json libraries more performant than the `json` standard library package and provide configuration on the build tag, allowing us to not have to effectively marshal twice for the yaml -> json conversion.  I looked into `json-iterator`, which is imported in other open-telemetry repos and showed good [benchmarks](https://pkg.go.dev/github.com/json-iterator/go#section-readme).  To see if it was the right fit for us, I wrote benchmarks for our server handling functions to see how json-iterator compares to the standard library package:

https://github.com/open-telemetry/opentelemetry-operator/pull/1323
[benchmarks.txt](https://github.com/open-telemetry/opentelemetry-operator/files/10273573/benchmarks.txt)

In terms of both memory and cpu, json-iterator performed better in all of my tests, especially as objects grew larger.

I'm interested in hearing others' thoughts on this - is this a change people are open to making?

jaronoff97:
Given those benchmarks, I am _very_ interested in this change going through!

kristinapathak:
These changes had to be reverted: https://github.com/open-telemetry/opentelemetry-operator/pull/1351

The prometheus structs have special `MarshalYAML()` functions that a json marshaler won't find and use.  Specifically the Regex [struct](https://pkg.go.dev/github.com/grafana/regexp#Regexp) has no exported fields, so all data is lost there.

PR #1410, opened 2023-01-31
[target-allocator] use json iterator to marshal in server

kristinapathak:
fixes https://github.com/open-telemetry/opentelemetry-operator/issues/1336

This time, server functionality is covered by unit tests.  Below are the results of the benchmarks, showing using the standard json library (std) vs jsoniter.

[test_results.txt](https://github.com/open-telemetry/opentelemetry-operator/files/10541897/test_results.txt)


cmd/otel-allocator/server/server.go
```diff
@@ -174,6 +185,14 @@ func (s *Server) errorHandler(w http.ResponseWriter, err error) {
 }
 
 func (s *Server) jsonHandler(w http.ResponseWriter, data interface{}) {
+	w.Header().Set("Content-Type", "application/json")
+	err := s.jsonMarshaller.NewEncoder(w).Encode(data)
+	if err != nil {
```
jaronoff97, line 190: do our tests already cover this case? i.e. is this change just going to work with what we do in the collector? using specifically the [yaml unmarshal](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/metrics_receiver.go#L224) function?

jaronoff97 (commented):
after this change goes out, should we upstream a change to the collector?

cmd/otel-allocator/server/server.go
kristinapathak: That yaml unmarshal function is for a response from the scrape config handler, which doesn't use the `jsonHandler` function.  [Here](https://github.com/open-telemetry/opentelemetry-operator/blob/6ca223ad8354d5f37b0c29489a155de751cdd2ef/cmd/otel-allocator/server/server.go#L100) is the marshaling logic for the scrape config endpoint.

The `jsonHandler()` function is used for the job and targets handlers, both of which have test cases which ensure that this marshaling is still compatible with `json.Unmarshal()`:
- https://github.com/open-telemetry/opentelemetry-operator/blob/6ca223ad8354d5f37b0c29489a155de751cdd2ef/cmd/otel-allocator/server/server_test.go#L175
- https://github.com/open-telemetry/opentelemetry-operator/blob/6ca223ad8354d5f37b0c29489a155de751cdd2ef/cmd/otel-allocator/server/server_test.go#L532

kristinapathak:
Latest benchmarks
[Uploading test_results.txt…]()

jaronoff97 (approved):

Issue #1404, opened 2023-01-27
[target-allocator] Remove yaml to json conversion for ScrapeConfigs endpoint


kristinapathak:
As discussed in the SIG meeting on January 19th, the plan is to modify the ScrapeConfigs endpoint so it sends a yaml response instead of json. Currently, we marshal a yaml response and then convert it to json - sending json directly allows us to stop doing that work.

The client for these requests is expected to be the prometheus receiver in opentelemetry-collector-contrib.  The logic in the prometheus receiver is already calling yaml.Unmarshal [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/72d590e8e76a18bbec6d4ce8b6200520c5e94732/receiver/prometheusreceiver/metrics_receiver.go#L205-L233).

PR #1411, opened 2023-01-31
[target-allocator] Remove json conversion from /scrape_configs

kristinapathak:
fixes https://github.com/open-telemetry/opentelemetry-operator/issues/1404

This change is breaking in that the scrape configs endpoint will now response with YAML instead of JSON.  However, the prometheus receiver which sends requests to `/scrape_configs` is currently using `yam.Unmarshal`: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/72d590e8e76a18bbec6d4ce8b6200520c5e94732/receiver/prometheusreceiver/metrics_receiver.go#L205-L233

This doesn't give us a large performance gain (compared to using jsoniter), but it is noticeable in benchmarks:
[scrapeconfig_bench_results.txt](https://github.com/open-telemetry/opentelemetry-operator/files/10542012/scrapeconfig_bench_results.txt)

### DIFF
--- a/.chloggen/scrapeconfig-json-conversion.yaml
+++ b/.chloggen/scrapeconfig-json-conversion.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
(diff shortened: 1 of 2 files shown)


matej-g:
Is anything preventing this and #1410 from moving forward? Happy to help :)

PR #1413, opened 2023-01-31
[target-allocator] Update scrape config upon discovery config apply

### LINKED ISSUE #1359: [target allocator] Data race when writing / reading `jobToScrapeConfig` in `Discoverer`

matej-g:
Fixes #1359.

This is a proposed alternative to solve the data race issue in scrape config handler and at the same time now blowing up the memory on a repeated handler call.

The stored scrape config response gets updated only upon relevant changes in the `Discoverer`. When a change is detected, it calls a method to update the stored scrape config in the HTTP server response. All responses to scrape config handler are then served from this stored JSON response (or alternatively an error is returned if the latest config update failed). We also keep the hash of config within the discoverer, to prevent sending unnecessarily scrape configs over channel if no update has been detected.

Benchmarks from my machine:
```
goos: darwin
goarch: arm64
pkg: github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/server
                                    │     old.txt     │            new_sync.txt             │
                                    │     sec/op      │    sec/op     vs base               │
ScrapeConfigsHandler/0_targets-10        1.929µ ± ∞ ¹   1.880µ ± ∞ ¹   -2.54% (p=0.032 n=5)
ScrapeConfigsHandler/5_targets-10      240.840µ ± ∞ ¹   2.107µ ± ∞ ¹  -99.13% (p=0.008 n=5)
ScrapeConfigsHandler/10_targets-10     452.602µ ± ∞ ¹   2.543µ ± ∞ ¹  -99.44% (p=0.008 n=5)
ScrapeConfigsHandler/50_targets-10    2246.603µ ± ∞ ¹   3.737µ ± ∞ ¹  -99.83% (p=0.008 n=5)
ScrapeConfigsHandler/100_targets-10   4672.521µ ± ∞ ¹   6.042µ ± ∞ ¹  -99.87% (p=0.008 n=5)
ScrapeConfigsHandler/500_targets-10   23730.46µ ± ∞ ¹   18.43µ ± ∞ ¹  -99.92% (p=0.008 n=5)
geomean                                  611.7µ         4.015µ        -99.34%
¹ need >= 6 samples for confidence interval at level 0.95

                                    │     old.txt     │             new_sync.txt             │
                                    │      B/op       │     B/op       vs base               │
ScrapeConfigsHandler/0_targets-10       6.376Ki ± ∞ ¹   6.423Ki ± ∞ ¹   +0.74% (p=0.008 n=5)
ScrapeConfigsHandler/5_targets-10     206.839Ki ± ∞ ¹   8.110Ki ± ∞ ¹  -96.08% (p=0.008 n=5)
ScrapeConfigsHandler/10_targets-10     311.29Ki ± ∞ ¹   10.36Ki ± ∞ ¹  -96.67% (p=0.008 n=5)
ScrapeConfigsHandler/50_targets-10    1651.12Ki ± ∞ ¹   24.36Ki ± ∞ ¹  -98.52% (p=0.008 n=5)
ScrapeConfigsHandler/100_targets-10   4117.17Ki ± ∞ ¹   46.37Ki ± ∞ ¹  -98.87% (p=0.008 n=5)
ScrapeConfigsHandler/500_targets-10   22878.4Ki ± ∞ ¹   182.4Ki ± ∞ ¹  -99.20% (p=0.008 n=5)
geomean                                 632.2Ki         21.93Ki        -96.53%
¹ need >= 6 samples for confidence interval at level 0.95

                                    │     old.txt     │            new_sync.txt            │
                                    │    allocs/op    │  allocs/op   vs base               │
ScrapeConfigsHandler/0_targets-10         24.00 ± ∞ ¹   23.00 ± ∞ ¹   -4.17% (p=0.008 n=5)
ScrapeConfigsHandler/5_targets-10       3396.00 ± ∞ ¹   23.00 ± ∞ ¹  -99.32% (p=0.008 n=5)
ScrapeConfigsHandler/10_targets-10      6702.00 ± ∞ ¹   23.00 ± ∞ ¹  -99.66% (p=0.008 n=5)
ScrapeConfigsHandler/50_targets-10     33144.00 ± ∞ ¹   23.00 ± ∞ ¹  -99.93% (p=0.008 n=5)
ScrapeConfigsHandler/100_targets-10    66208.00 ± ∞ ¹   23.00 ± ∞ ¹  -99.97% (p=0.008 n=5)
ScrapeConfigsHandler/500_targets-10   330516.00 ± ∞ ¹   23.00 ± ∞ ¹  -99.99% (p=0.008 n=5)
geomean                                  8.570k         23.00        -99.73%
¹ need >= 6 samples for confidence interval at level 0.95
```


cmd/otel-allocator/main.go
```diff
@@ -209,6 +212,21 @@ func main() {
 			setupLog.Info("Closing watcher loop")
 			close(eventCloser)
 		})
+	runGroup.Add(
+		func() error {
+			for {
+				select {
+				case scrapeCfg := <-scrapeConfigChan:
+					srv.UpdateScrapeConfigResponse(scrapeCfg)
```
matej-g, line 220: One thing I'm wondering is how to handle update failures such as unmarshalling. For now I tried to keep the behavior we have currently (return server error in the HTTP handler), but I wonder if we should not handle it here with `setupLog.Error` and keep returning last known good scrape configuration? Thoughts?

matej-g:
I wonder now if there is actually need to do this over channel rather than doing this synchronously and just update the scrape config response directly from `Discoverer`

jaronoff97:
yeah i'm not sure we need the channel here, you can do as we've done with [interfaces like the hook](https://github.com/open-telemetry/opentelemetry-operator/blob/main/cmd/otel-allocator/target/discovery.go#L45-L47) and just pass that around. Could you also add some tests for the UpdateScrapeConfigResponse? They'll probably look similar to the ones removed in the `TestScrapeConfigsHandler_Hashing`

matej-g:
Yeah, I guess no reason to complicate it here, let me do that + add tests and come back with more commits

cmd/otel-allocator/main.go
matej-g: No longer an open question I guess

matej-g:
I have updated the PR, the description and I'm undrafting this for reviews, thanks! cc @jaronoff97 @kristinapathak

cmd/otel-allocator/server/server.go
```diff
@@ -93,38 +85,36 @@ func (s *Server) Shutdown(ctx context.Context) error {
...
+	}
+	s.scrapeConfigErr = nil
+	s.scrapeConfigResponse = jsonConfig
+}
+
+// ScrapeConfigsHandler returns the available scrape configuration discovered by the target allocator.
+func (s *Server) ScrapeConfigsHandler(c *gin.Context) {
+	response, err := s.scrapeConfigResponse, s.scrapeConfigErr
+	if err != nil {
+		s.errorHandler(c.Writer, s.scrapeConfigErr)
```
frzifus, line 112: Can async reading `s.scrapeConfigErr` here and writing it in ` func UpdateScrapeConfigResponse` lead to some issues?

.chloggen/1413-cache-scrape-config-response.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: "target allocator"
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Save the scrape config response in the HTTP server upon relevant confif change, instead of building it on every handler call. At the same time, this avoids data race when accessing the scrape configs map."
```
frzifus, line 8:
```suggestion
note: "Save the scrape config response in the HTTP server upon relevant config change, instead of building it on every handler call. At the same time, this avoids data race when accessing the scrape configs map."
```

cmd/otel-allocator/server/server.go
matej-g: Good catch, I think actually I should put `err` here instead from `s.scrapeConfigErr`. That should ensure even if we change / remove the error cached on the HTTP server in the meantime, we won't put incorrect / nil error into the response.

cmd/otel-allocator/server/server.go
```diff
@@ -48,26 +47,19 @@ type collectorJSON struct {
...
-	logger           logr.Logger
-	allocator        allocation.Allocator
-	discoveryManager DiscoveryManager
-	server           *http.Server
+	logger    logr.Logger
+	allocator allocation.Allocator
+	server    *http.Server
 
-	compareHash          uint64
 	scrapeConfigResponse []byte
```
jaronoff97, line 55: I think we need a lock around the scrape config response to prevent concurrent access from the discovery manager?

cmd/otel-allocator/server/server.go
```diff
@@ -93,38 +85,36 @@ func (s *Server) Shutdown(ctx context.Context) error {
...
-	configs := s.discoveryManager.GetScrapeConfigs()
-
-	hash, err := hashstructure.Hash(configs, nil)
+func (s *Server) UpdateScrapeConfigResponse(configs map[string]*promconfig.ScrapeConfig) {
+	var configBytes []byte
+	configBytes, err := yaml.Marshal(configs)
 	if err != nil {
-		s.logger.Error(err, "failed to hash the config")
-		s.errorHandler(c.Writer, err)
+		s.scrapeConfigErr = err
```
jaronoff97, line 95: IMO we should return an error from here rather than store it, the discovery manager's method already has error handling baked in. This also means that if we see some type of bogus configuration here, we can bubble that up to the caller while still holding on to the last good configuration. In this case we'd want to store the hash only after a successful call to `UpdateScrapeConfigResponse`.

cmd/otel-allocator/target/discovery.go
```diff
@@ -75,6 +79,17 @@ func (m *Discoverer) ApplyConfig(source allocatorWatcher.EventSource, cfg *confi
...
 	}
 
+	hash, err := hashstructure.Hash(m.jobToScrapeConfig, nil)
+	if err != nil {
+		return err
+	}
+	// If the hash has changed, updated stored hash and send the new config.
+	// Otherwise skip updating scrape configs.
+	if m.scrapeConfigsUpdater != nil && m.scrapeConfigsHash != hash {
+		m.scrapeConfigsHash = hash
```
jaronoff97, line 89: as mentioned above, this should only come after our successful call to the `UpdateScrapeConfigResponse`

jaronoff97 (changes requested):
Similar to @frzifus review, i had a few ideas around improved error handling. This is looking great though, thank you.

cmd/otel-allocator/server/server.go
matej-g: Sounds right to me, I initially wanted to keep behavior similar to what we have now, but returning last known good config is preferable to me as well.

matej-g:
Thanks for the reviews @jaronoff97 @frzifus, this should be ready for another round 🙇

jaronoff97 (approved):
Looking great! I had one ask for another test for the new error behavior.

kristinapathak (approved):
looks great! 🥳

cmd/otel-allocator/target/discovery.go
```diff
@@ -75,6 +79,21 @@ func (m *Discoverer) ApplyConfig(source allocatorWatcher.EventSource, cfg *confi
...
+	hash, err := hashstructure.Hash(m.jobToScrapeConfig, nil)
+	if err != nil {
+		return err
+	}
+	// If the hash has changed, updated stored hash and send the new config.
+	// Otherwise skip updating scrape configs.
+	if m.scrapeConfigsUpdater != nil && m.scrapeConfigsHash != hash {
+		err := m.scrapeConfigsUpdater.UpdateScrapeConfigResponse(m.jobToScrapeConfig)
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
matej-g: That's a good idea, adding 👍

jaronoff97 (approved):

PR #1414, opened 2023-01-31
[target-allocator] Use `gin` in release mode and without default logger middleware

matej-g:

This is a follow-up to #1383.

This PR changes the `gin` router configuration to be suitable for production use, namely:
- It sets `gin` to be used in `release` mode (to my surprise, the default was `debug`)
- it opts out of using the default logger middleware. The logging seems rather verbose (logging each request) and the middleware is not using logging format that is consistent with logging format of the rest of the components in `target-allocator`. 

Eventually, it would be nice if we introduced our own logging middleware with formatting consistent with other components, but for now the default one seems noisy and not too useful to me.

cc @kristinapathak

### DIFF
--- a/.chloggen/1414-adjust-gin-config.yaml
+++ b/.chloggen/1414-adjust-gin-config.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: "target allocator"
+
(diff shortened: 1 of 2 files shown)


frzifus (approved):

matej-g:
Is the issue number mandatory in changelog? 🤔

frzifus:
> Is the issue number mandatory in changelog? thinking

yes :D

matej-g:
I did not create one but I guess it still should be part of #1352, so linking it to that one. Thanks @frzifus 😊.

frzifus:
makes sense to me :)

jaronoff97 (approved):

yuriolisa (approved):

pavolloffay (approved):

PR #1442, opened 2023-02-03
[docs] Update sampler and propagation docs

TylerHelmuth:
Updates the sampler and propagation documentation to explicitly state which environment variables get set when this fields are used.  Added links to otel specification.  For propagators, added list of valid values.

I updated the docs directly, but I get the sense they are auto-generated.  I couldn't figure out what the root source of truth is, so if I've updated stuff incorrectly please point me in the right direction.

Related to https://github.com/open-telemetry/opentelemetry.io/pull/2253

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,6 +213,13 @@ spec:
 EOF
 ```
 
+The values for `propagators` are added to the `OTEL_PROPAGATORS` environment variable.
+Valid values for `propagators` are defined by the [OpenTelemetry Specification for OTEL_PROPAGATORS](https://opentelemetry.io/docs/concepts/sdk-configuration/general-sdk-configuration/#otel_propagators).
+
+The value for `sampler.type` is added to the `OTEL_TRACES_SAMPLER` envrionment variable.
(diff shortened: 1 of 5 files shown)


TylerHelmuth:
@open-telemetry/operator-approvers any advice on how to get the CI to pass?

jaronoff97:
@TylerHelmuth I added the skip changelog and an going to rerun the unit tests.

TylerHelmuth:
@jaronoff97 moved comments into API strings.  I don't see them showing up as api doc comments tho.

jaronoff97:
@TylerHelmuth i think you may need to run `make bundle` first which will generate the CRDs and then `make api-docs`, you'll want to make sure you don't include anything related to versioning however. Running locally, you'll want to ignore changes to:
```
bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
config/manager/kustomization.yaml
```

TylerHelmuth:
@jaronoff97 that did the trick

jaronoff97:
Looks like it's still failing because it wants to remove those changes... I'm going to pull this down locally and see what's happening.

TylerHelmuth:
@jaronoff97 ready for review

jaronoff97 (approved):

PR #1305, opened 2022-12-06
CRD defs for Apache HTTPD Autoinstrumentation

chrlic:
This is a followup PR in Apache HTTPD Autoinstrumentation implementation project to previous PR #1236.

It contains CRD definition, defaults, validations, and test.

Next PR will contain the autoinstrumentation logic itself and e2e tests.

PR #1444, opened 2023-02-04
Apache HTTPD Autoinstrumentation - webhook, readme, tests, etc. #1305

chrlic:
This PR is a third (and conceptually last) in the series implementing autoinstrumentation for Apache HTTPD server following on PR #1305. 
It contains the webhook with instrumentation logic, all other related code changes, readme, and e2e tests.


README.md
```diff
@@ -307,11 +313,35 @@ spec:
...
+    attrs:
+    - name: ApacheModuleOtelMaxQueueSize
+      value: "4096"
+    - name: ...
+      value: ...
+```
+List of all available attributes can be found at [otel-webserver-module](https://github.com/open-telemetry/opentelemetry-cpp-contrib/tree/main/instrumentation/otel-webserver-module)
+
+
+
```
pavolloffay, line 344: please remove redundant empty lines

pavolloffay (commented):
A couple of comments, but mostly LGTM. 

Could somebody else from CISCO review as well?

README.md
```diff
@@ -239,6 +239,12 @@ DotNet:
 instrumentation.opentelemetry.io/inject-dotnet: "true"
 ```
 
+Apache HTTPD:
+```bash
+instrumentation.opentelemetry.io/inject-apache-httpd: "true"
+instrumentation.opentelemetry.io/otel-namespace: "my-custom-namespace" # optional, if missing, OTEL_NAMESPACE will be set to pod namespace
```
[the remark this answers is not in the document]
chrlic: Yes, by default, it is set to httpd pod namespace. 
Some backends may use OTEL_NAMESPACE value in a different context than K8S namespace, like AppDynamics use it for "application name" and users may want to set it to anything they consider application name.

README.md
chrlic: will do

tests/e2e/instrumentation-apache-httpd/00-install-instrumentation.yaml
```diff
@@ -0,0 +1,18 @@
...
+  exporter:
+    endpoint: http://localhost:4317
+  propagators:
+    - jaeger
+    - b3
+  sampler:
+    type: parentbased_traceidratio
+    argument: "0.25"
+  apacheHttpd:
+    image: docker.io/chrlic/opentelemetry-webserver-agent:latest
```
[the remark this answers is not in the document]
chrlic: I'd love to host it on OTEL org repo - how would I do that?

README.md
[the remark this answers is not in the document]
pavolloffay: Is the value of the annotation exclusively used to set `OTEL_NAMESPACE` ? If so could users use the resource section in the CR?

tests/e2e/instrumentation-apache-httpd/00-install-instrumentation.yaml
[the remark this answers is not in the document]
chrlic: It relies on binary releases of https://github.com/open-telemetry/opentelemetry-cpp-contrib, which get pulled during docker image build. 

But I entirely forgot I also prepared a git action for image build earlier - so now there's an image (autoinstrumentation-apache-httpd:1.0.2 and latest) in the ghcr.io/open-telemetry/opentelemetry-operator. Also updated in the e2e tests.

README.md
chrlic: done

README.md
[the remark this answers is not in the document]
chrlic: It's more importantly used in a configuration file of httpd which get's generated as part of the instrumentation - it's one of the mandatory attributes for the webserver instrumentation library. The value from a CR would apply to all deployments of httpd using that Instrumentation CR, in practice requiring potentially a lot of those, each "application specific". 

I found it more logical to have it specified as part of the actual deployment/pod spec and I needed some mechanism to do so. 

What do you think?

tests/e2e/instrumentation-apache-multicontainer/00-install-instrumentation.yaml
```diff
@@ -0,0 +1,18 @@
...
+  exporter:
+    endpoint: http://localhost:4317
+  propagators:
+    - jaeger
+    - b3
+  sampler:
+    type: parentbased_traceidratio
+    argument: "0.25"
+  apacheHttpd:
+    image: ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-apache-httpd:1.0.2
```
pavolloffay, line 15: isn't the default image set automatically by the operator? At least this is the case for other auto-instrumentations. 

I would prefer to use the default image, as this will get forgotten if there will be new image version

README.md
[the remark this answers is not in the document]
pavolloffay: Could we move this to a separate issue/PR? 

I am concerned of adding this annotation and especially just for the HTTPD.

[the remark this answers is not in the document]
chrlic: I totally get that. 

And thinking of it - it is actually generic topic - if someone would, say, have a Java application and want to trace it via AppDynamics, where OTEL_NAMESPACE is required, it would have to be set either on CR level or in the application manifest. 

So I'll remove it for the time being and let it on the K8S namespace as a default. And let's think of a better way of doing it generically.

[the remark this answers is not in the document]
chrlic: I'll open an issue for this.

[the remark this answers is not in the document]
chrlic: annotation removed, tested, issue for discussion raised at https://github.com/open-telemetry/opentelemetry-operator/issues/1572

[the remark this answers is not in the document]
chrlic: @pavolloffay - in your opinion - is it ok like this, i.e. for the time being, using K8S namespace as default for OTEL_SERVICE_NAMESPACE? If so, can we move this forward?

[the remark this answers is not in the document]
chrlic: @pavolloffay - can we proceed with this PR?

README.md
```diff
@@ -183,7 +183,7 @@ When using sidecar mode the OpenTelemetry collector container will have the envi
 
 ### OpenTelemetry auto-instrumentation injection
 
-The operator can inject and configure OpenTelemetry auto-instrumentation libraries. Currently DotNet, Java, NodeJS and Python are supported.
+The operator can inject and configure OpenTelemetry auto-instrumentation libraries. Currently DotNet, Java, NodeJS, Python and Apache HTTPD are supported.
```
TylerHelmuth, line 186:
```suggestion
The operator can inject and configure OpenTelemetry auto-instrumentation libraries. Currently Apache HTTPD, DotNet, Java, NodeJS and Python are supported.
```
Lets keep the alphabetical order.

TylerHelmuth (commented):
@chrlic please add unit tests in `podmutator_test.go`, `sdk_test.go` and `upgrade_test.go`.

chrlic:
> @chrlic please add unit tests in `podmutator_test.go`, `sdk_test.go` and `upgrade_test.go`.

Unit tests added and tested

pkg/featuregate/featuregate.go
```diff
@@ -45,6 +45,10 @@ var (
 		"operator.autoinstrumentation.go",
 		featuregate.StageAlpha,
 		featuregate.WithRegisterDescription("controls whether the operator supports Goland auto-instrumentation"))
+	EnableApacheHTTPAutoInstrumentationSupport = featuregate.GlobalRegistry().MustRegister(
+		"operator.autoinstrumentation.apache-http",
+		featuregate.StageAlpha,
```
TylerHelmuth, line 50: If we want this instrumentation enabled by default then we need to use StageBeta.  If we don't want it enabled by default then we need to update the README file to reflect that it is disabled by default.

pkg/instrumentation/podmutator_test.go
```diff
@@ -1034,58 +1031,31 @@ func TestMutatePod(t *testing.T) {
 			},
 		},
 		{
-			name: "dotnet injection feature gate disabled",
+			name: "apache httpd injection, true",
```
TylerHelmuth, line 1034: Maybe a bad merge in this file?  Looks like some existing tests were removed instead of this test being added.

pkg/instrumentation/sdk_test.go
```diff
@@ -1101,6 +1101,141 @@ func TestInjectGo(t *testing.T) {
 	}
 }
 
+func TestInjectApacheHttpd(t *testing.T) {
```
TylerHelmuth, line 1104: nit: can you restructure this test to follow the same "name, inst, pod, expected" struct pattern that the other tests follow?

pkg/instrumentation/podmutator.go
```diff
@@ -141,14 +142,21 @@ func (pm *instPodMutator) Mutate(ctx context.Context, ns corev1.Namespace, pod c
 		pm.Recorder.Event(pod.DeepCopy(), "Warning", "InstrumentationRequestRejected", "support for Go auto instrumentation is not enabled")
 	}
 
+	if inst, err = pm.getInstrumentationInstance(ctx, ns, pod, annotationInjectApacheHttpd); err != nil {
+		// we still allow the pod to be created, but we log a message to the operator's logs
+		logger.Error(err, "failed to select an OpenTelemetry Instrumentation instance for this pod")
+		return pod, err
+	}
+	insts.ApacheHttpd = inst
```
TylerHelmuth, line 150: We should add the feature-gate check here like the other languages

pkg/instrumentation/upgrade/upgrade.go
```diff
@@ -138,5 +139,13 @@ func (u *InstrumentationUpgrade) upgrade(_ context.Context, inst v1alpha1.Instru
 			u.Recorder.Event(inst.DeepCopy(), "Warning", "InstrumentationUpgradeRejected", "support for Go auto instrumentation is not enabled")
 		}
 	}
+	autoInstApacheHttpd := inst.Annotations[v1alpha1.AnnotationDefaultAutoInstrumentationApacheHttpd]
+	if autoInstApacheHttpd != "" {
+		// upgrade the image only if the image matches the annotation
+		if inst.Spec.ApacheHttpd.Image == autoInstApacheHttpd {
```
TylerHelmuth, line 145: We should add the feature-gate check here like the other languages

pkg/featuregate/featuregate.go
chrlic: featuregate.StageBeta used

pkg/instrumentation/podmutator_test.go
chrlic: Oh sorry - bad merge indeed. Fixed.

pkg/instrumentation/sdk_test.go
chrlic: Restructured with the test struct as advised, thank you - much more readable and scalable.

pkg/instrumentation/podmutator.go
chrlic: Done.

pkg/instrumentation/upgrade/upgrade.go
chrlic: Done.

TylerHelmuth (approved):
Thanks for doing all this work!

chrlic:
Can I do anything else to have this merged?

README.md
```diff
@@ -358,13 +385,14 @@ Prefix a gate with '-' to disable support for the corresponding language.
...
-| Python   | `operator.autoinstrumentation.python` | enabled       |
-| DotNet   | `operator.autoinstrumentation.dotnet` | enabled       |
-| Go       | `operator.autoinstrumentation.go`     | disabled      |
+| Language | Gate                                        | Default Value |
+|----------|---------------------------------------------|---------------|
+| Java     | `operator.autoinstrumentation.java`         | enabled       |
+| NodeJS   | `operator.autoinstrumentation.nodejs`       | enabled       |
+| Python   | `operator.autoinstrumentation.python`       | enabled       |
+| DotNet   | `operator.autoinstrumentation.dotnet`       | enabled       |
+| DotNet   | `operator.autoinstrumentation.apache-httpd` | enabled       |
```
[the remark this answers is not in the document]
jaronoff97: should this be default disabled given it's a netnew auto instrumentation feature?

pkg/instrumentation/apachehttpd.go
```diff
@@ -0,0 +1,236 @@
...
+	apacheAgentInitContainerName  = "otel-agent-attach-apache"
+	apacheAgentCloneContainerName = "otel-agent-source-container-clone"
+	apacheAgentConfigVolume       = "otel-apache-conf-dir"
+	apacheAgentVolume             = "otel-apache-agent"
+	apacheAttributesEnvVar        = "OTEL_APACHE_AGENT_CONF"
+	apacheServiceInstanceId       = "<<SID-PLACEHOLDER>>"
+	apacheServiceInstanceIdEnvVar = "APACHE_SERVICE_INSTANCE_ID"
+)
+
+func injectApacheHttpdagent(_ logr.Logger, apacheSpec v1alpha1.ApacheHttpd, pod corev1.Pod, index int, otlpEndpoint string, resourceMap map[string]string) corev1.Pod {
```
jaronoff97, line 46: can you add a comment here about what exactly this method is doing? i'm a bit confused by the two `isApacheInitContainerMissing` checks.

pkg/instrumentation/apachehttpd.go
```diff
@@ -0,0 +1,236 @@
...
+		otelEndpoint = "http://localhost:4317/"
+	}
+	serviceName := chooseServiceName(pod, resourceMap, index)
+	serviceNamespace := pod.GetNamespace()
+	// Namespace name override TBD
+
+	versionSuffix := ""
+	if apacheSpec.Version == "2.2" {
+		versionSuffix = "22"
+	}
```
jaronoff97, line 202: How was this version decided? it looks like there's also apache 2.4 above?

pkg/instrumentation/apachehttpd_test.go
```diff
@@ -0,0 +1,132 @@
...
+		pod      corev1.Pod
+		expected corev1.Pod
+	}{
+		{
+			name:        "Clone Container not present",
+			ApacheHttpd: v1alpha1.ApacheHttpd{Image: "foo/bar:1"},
+			pod: corev1.Pod{
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{
+						{},
```
jaronoff97, line 41: I don't think this is a valid pod spec... therefore shouldn't we be unable to inject init containers?

pkg/instrumentation/apachehttpd_test.go
```diff
@@ -0,0 +1,132 @@
...
+				},
+			},
+		},
+	}
+
+	resourceMap := map[string]string{
+		string(semconv.K8SDeploymentNameKey): "apache-httpd-service-name",
+	}
+
+	for _, test := range tests {
```
jaronoff97, line 126: could you add more tests for the `isApacheInitContainerMissing` cases in your function?

pkg/featuregate/featuregate.go
```diff
@@ -45,6 +45,10 @@ var (
 		"operator.autoinstrumentation.go",
 		featuregate.StageAlpha,
 		featuregate.WithRegisterDescription("controls whether the operator supports Goland auto-instrumentation"))
+	EnableApacheHTTPAutoInstrumentationSupport = featuregate.GlobalRegistry().MustRegister(
+		"operator.autoinstrumentation.apache-http",
+		featuregate.StageBeta,
```
jaronoff97, line 50: IMO this should be `StageAlpha`

pkg/instrumentation/sdk_test.go
```diff
@@ -1101,6 +1101,155 @@ func TestInjectGo(t *testing.T) {
...
+
+	tests := []struct {
```
jaronoff97, line 1106: What happens when exporter is not set?

tests/e2e/instrumentation-apache-httpd/00-install-instrumentation.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: opentelemetry.io/v1alpha1
+kind: Instrumentation
+metadata:
+  name: apache
+spec:
+  exporter:
+    endpoint: http://localhost:4317
```
jaronoff97, line 7: given we default here, can we remove this from the install which will help confirm that expectation?

tests/e2e/instrumentation-apache-httpd/01-install-app.yaml
```diff
@@ -0,0 +1,30 @@
...
+    metadata:
+      labels:
+        app: my-apache
+      annotations:
+        sidecar.opentelemetry.io/inject: "true"
+        instrumentation.opentelemetry.io/inject-apache-httpd: "true"
+    spec:
+      containers:
+      - name: myapp
+        image: docker.io/chrlic/apache-test
```
jaronoff97, line 20: What's the origin of this image? what's the license on it?

internal/config/main.go
```diff
@@ -135,6 +137,10 @@ func (c *Config) AutoDetect() error {
 		c.logger.V(1).Info("HPA version detected", "version", hpaV)
 		c.hpaVersion.Set(hpaV)
 	}
+	if c.hpaVersion.Get() != hpaV {
+		c.logger.V(1).Info("HPA version detected", "version", hpaV)
+		c.hpaVersion.Set(hpaV)
+	}
```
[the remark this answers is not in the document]
chrlic: yes. removed

pkg/instrumentation/apachehttpd.go
```diff
@@ -0,0 +1,236 @@
...
+		cloneContainer.Name = apacheAgentCloneContainerName
+		cloneContainer.Command = []string{"/bin/sh", "-c"}
+		cloneContainer.Args = []string{"cp -r /usr/local/apache2/conf/* " + apacheAgentConfDirFull}
+		cloneContainer.VolumeMounts = append(cloneContainer.VolumeMounts, corev1.VolumeMount{
+			Name:      apacheAgentConfigVolume,
+			MountPath: apacheAgentConfDirFull,
+		})
+		// remove resource requirements since those are then reserved for the lifetime of a pod
+		// and we definitely do not need them for the init container for cp command
+		cloneContainer.Resources = corev1.ResourceRequirements{}
```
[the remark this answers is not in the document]
chrlic: thank you - resource spec taken from Instrumentation spec.

pkg/instrumentation/apachehttpd.go
chrlic: I agree - it's worth explaining, since for Apache HTTPD, it works a bit specifically. Comments included in the code.

pkg/instrumentation/apachehttpd.go
chrlic: explained in the comment in the code. 2.4 is default and the OTEL module file does not have any version suffix. Apache 2.2 has version suffix "22".

pkg/instrumentation/apachehttpd_test.go
chrlic: It's a test for instrumentation only, checking the result of instrumentation. In that case, it does not need IMHO full application pod spec specified. 
Instrumentation does not validate the application pod/container in any way, I think this is the task for K8S validators?

tests/e2e/instrumentation-apache-httpd/01-install-app.yaml
chrlic: image origin and license info included in comments.
also added image checksum

pkg/featuregate/featuregate.go
chrlic: agreed, done

pavolloffay (approved):
LGTM 🎉 

Why did we decide to disable it by default?

chrlic:
> LGTM 🎉
> 
> Why did we decide to disable it by default?

It was suggested because it's fresh new, I believe

pavolloffay:
If the instrumentation is considered stable then let's enable it by default.

chrlic:
Ok, I'll change it back to enabled by default.

jaronoff97 (approved):
Just one small change... I think a few more tests would be great, but can also be done in a followup.

jaronoff97:
absolutely :) thanks so much! merging this in now.

svrnm:
great work @chrlic! Thanks for making this happen!

chrlic:
Thank you all for your help! Really appeciated!

PR #1455, opened 2023-02-07
Adds Github action for publishing container image of Operator OpAMP Bridge

### LINKED ISSUE #1369: Add CI/CD for OpAMP Bridge container
Now that we have the OpAMP bridge in the repository, we need to set up a new gcr image to push to. This will be mostly a copy-paste job from the target-allocator.

avadhut123pisal:
Closes #1369


avadhut123pisal:
Hi @jaronoff97, do we need to make the changes in the Operator's code to access the version information of the OpAMP Bridge as part of this PR or that is to be done when CRD for OpAMP is ready ?

jaronoff97:
@avadhut123pisal let's make the necessary operator changes in a follow-up to this PR to keep the scope of this small.

versions.txt
```diff
@@ -29,3 +29,6 @@ autoinstrumentation-dotnet=0.5.0
 # Represents the current release of Apache HTTPD instrumentation.
 # Should match autoinstrumentation/apache-httpd/version.txt
 autoinstrumentation-apache-httpd=1.0.2
+
+# Represents the current release of the Operator OpAMP Bridge.
+operator-opamp-bridge=0.1.0
```
jaronoff97, line 34: Can we move this up next to the target allocator and match the version as well?

.github/workflows/publish-operator-opamp-bridge.yaml
```diff
@@ -0,0 +1,70 @@
+name: "Publish Operator OpAMP Bridge"
```
jaronoff97, line 1: Note: if we start adding more containers that get built in this repo, we should probably make a single action to publish multiple containers.

versions.txt
jaronoff97: this is why unit tests are failing as well. 
```
Error: invalid command options: 0.68.0 0.1.0 is not a valid semantic version: Invalid character(s) found in patch number "0 0.1.0"
```

jaronoff97: ah actually this may be an issue with the grep happening in the action.

jaronoff97: namely this grep in the Makefile:
```
OPERATOR_VERSION ?= "$(shell grep -v '\#' versions.txt | grep operator | awk -F= '{print $$2}')"
```

avadhut123pisal: Yes. As `operator` word is present in `operator-opamp-bridge=0.1.0` also, so it is causing the problem. 
Should we use `grep operator=` here so that would not conflict with `operator-opamp-bridge=0.1.0` ?

jaronoff97 (approved):

kristinapathak (approved):

.github/workflows/publish-images.yaml
```diff
@@ -24,6 +24,7 @@ jobs:
         run: |
           grep -v '\#' versions.txt | grep opentelemetry-collector | awk -F= '{print "OTELCOL_VERSION="$2}' >> $GITHUB_ENV
           grep -v '\#' versions.txt | grep targetallocator | awk -F= '{print "TARGETALLOCATOR_VERSION="$2}' >> $GITHUB_ENV
+          grep -v '\#' versions.txt | grep operator-opamp-bridge | awk -F= '{print "OPERATOR_OPAMP_BRIDGE="$2}' >> $GITHUB_ENV
```
jaronoff97, line 27: small nit: can we change the environment variable name here to include `_VERSION` i.e. `OPERATOR_OPAMP_BRIDGE_VERSION` like the other versions being set in this workflow?

avadhut123pisal: Done

jaronoff97: thank you!

jaronoff97:
@avadhut123pisal sorry for the inconvenience, but could you pull in the latest changes as we are now on version 0.69.0?

avadhut123pisal:
> @avadhut123pisal let's make the necessary operator changes in a follow-up to this PR to keep the scope of this small. Thank you for your contribution here.

@jaronoff97, I will raise the new PR with Operator changes soon.

Issue #1439, opened 2023-02-03
spec.autoscaler.targetMemoryUtilization does not add memory scaling to autoscaler


jsirianni:
When deploying a collector with autoscaling enabled, the HPA is created with CPU metric scaling only, despite `targetMemoryUtilization` being set in `spec.autoscaler`.

I have the following collector config:

```yaml
apiVersion: opentelemetry.io/v1alpha1
kind: OpenTelemetryCollector
metadata:
  name: collector
spec:
  autoscaler:
    minReplicas: 2
    maxReplicas: 10
    targetCPUUtilization: 60
    targetMemoryUtilization: 80
    ...
```

The collector deploys and works great, but the hpa does not have memory configured:

```yaml
apiVersion: autoscaling/v2
kind: HorizontalPodAutoscaler
...
spec:
  maxReplicas: 10
  metrics:
  - resource:
      name: cpu
      target:
        averageUtilization: 60
        type: Utilization
    type: Resource
  minReplicas: 2
  scaleTargetRef:
    apiVersion: opentelemetry.io/v1alpha1
    kind: OpenTelemetryCollector
    name: collector
...
```

Kubernetes: `1.25.5` (GKE: 1.25.5-gke.2000)
Operator: `ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator:0.68.0`

I have other working HPA's defined outside of the operator which do have memory scaling, so I am not sure it is an issue with my cluster.

```yaml
...
apiVersion: autoscaling/v2
kind: HorizontalPodAutoscaler
...
  metrics:
  - resource:
      name: memory
      target:
        averageUtilization: 60
        type: Utilization
    type: Resource
  - resource:
      name: cpu
      target:
        averageUtilization: 60
        type: Utilization
    type: Resource
...
```

moh-osman3:
Hmm this is strange, I wonder if there's an issue with updating the HPA when reapplying your new config. Have you tried deleting your namespace entirely and installing the operator and collector again? I am using the same operator version and autoscaler config and my hpa seems to run with no issues

```
% k get hpa -o yaml             
...
apiVersion: autoscaling/v2
kind: HorizontalPodAutoscaler
...
spec:
  maxReplicas: 10
  metrics:
  - resource:
      name: memory
      target:
        averageUtilization: 80
        type: Utilization
    type: Resource
  - resource:
      name: cpu
      target:
        averageUtilization: 60
        type: Utilization
    type: Resource
[...]

PR #1462, opened 2023-02-09
[HPA] Properly update HPA when an additional metric is added to the spec

moh-osman3:
Currently the HPA fails to update properly when a new metric is added. [setAutoscalerSpec](https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/collector/reconcile/horizontalpodautoscaler.go#L113) assumes the HPA spec structure will remain static when it is updated, but this is not the case when a new metric is added to the spec. Now the existing MetricSpec copies the desired MetricSpec directly so that any changes to the structure will also be copied.

Resolves https://github.com/open-telemetry/opentelemetry-operator/issues/1439


moh-osman3:
@kevinearls @pavolloffay Could you please take a look when you get a chance? This deviates from the `client.Patch` that is used to update most other objects in this repository. I have been unable to properly patch when the underlying HPA changes to add a new metric. Instead I am using `client.Update` to perform the update, and this has successfully resolved the issue reported. Any input or suggestions would be appreciated! Thank you!

kevinearls:
Hi @moh-osman3 sorry, but I'm no longer working on this project.

moh-osman3:
@kevinearls No worries, thanks for letting me know. @pavolloffay would you have some time to review this PR? It fixes an issue with updating the HPA when a second metric is added to the CRD and reapplied to the cluster. Thanks!

.chloggen/1439-fix-HPA-updates.yaml
```diff
@@ -0,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: Autoscaler
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: This PR fixes issue where HPA fails to update when an additional metric is added to the spec.
```
pavolloffay, line 8: `Fix the issue where HPA fails to update when an additional metric is added to the spec.`

pkg/collector/reconcile/horizontalpodautoscaler.go
```diff
@@ -83,64 +83,27 @@ func expectedHorizontalPodAutoscalers(ctx context.Context, params Params, expect
 			return fmt.Errorf("failed to get %w", err)
 		}
 
-		updated := existing.DeepCopyObject().(client.Object)
-		updated.SetOwnerReferences(desired.GetOwnerReferences())
-		setAutoscalerSpec(params, autoscalingVersion, updated)
-
-		annotations := updated.GetAnnotations()
```
pavolloffay, line 90: This removal seems suspicious (handling of annotations and owner references). Could we keep this code and change the copying of the spec instead?

iblancasa (changes requested):
How about adding extra steps to the `autoscaling` E2E test to verify the fix is working 100%?

pkg/collector/reconcile/horizontalpodautoscaler.go
moh-osman3: Thanks for the review! Hmm it's not clear to me why these need to be manually copied.  Shouldn't these be taken care of  when the desired object is created in this [function](https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/collector/horizontalpodautoscaler.go#L36)?  Maybe I'm just confusing myself. As another example it isn't clear to me why `setAutoscalerSpec` is necessary, when these fields are set when the HPA is created. Is there a case I'm not understanding where these fields could be overwritten or missing?

moh-osman3: I am using your suggestion for now to just do the correct update in setAutoscalerSpec , but am certainly interested in knowing whether there should be a followup PR to clean up the way this update is completed (maybe I'm overthinking this).

.chloggen/1439-fix-HPA-updates.yaml
moh-osman3: fixed!

pkg/collector/reconcile/horizontalpodautoscaler.go
moh-osman3: Hey @pavolloffay, I think I better understand why the old annotations and ownerReferences should be copied instead of relying on the new ones generated when the desired HPA is created. 

I updated the PR once again with a simplified version of copying the spec. Now I am simply assigning the desired MetricSpec to the updated MetricSpec. Can you please take a look? Thanks!

moh-osman3:
> How about adding extra steps to the `autoscaling` E2E test to verify the fix is working 100%?

@iblancasa Thanks for the review! I added to the e2e tests to now check the number of metrics in the HPA Metric Spec. Then I install an update with the same name and an additional metric.  Let me know if you have any more suggestions on how to improve the e2e test.

pkg/collector/reconcile/horizontalpodautoscaler.go
```diff
@@ -87,6 +86,15 @@ func expectedHorizontalPodAutoscalers(ctx context.Context, params Params, expect
...
 
+		// update metrics separately because the structure of the metrics can change.
+		if autoscalingVersion == autodetect.AutoscalingVersionV2Beta2 {
+			desiredMetrics := desired.(*autoscalingv2beta2.HorizontalPodAutoscaler).Spec.Metrics
+			updated.(*autoscalingv2beta2.HorizontalPodAutoscaler).Spec.Metrics = desiredMetrics
+		} else { // autoscalingv2
+			desiredMetrics := desired.(*autoscalingv2.HorizontalPodAutoscaler).Spec.Metrics
+			updated.(*autoscalingv2.HorizontalPodAutoscaler).Spec.Metrics = desiredMetrics
+		}
+
```
matej-g, line 97: Should this be part of the `setAutoscalerSpec` since we're manipulating the spec?

iblancasa (changes requested):

pkg/collector/reconcile/horizontalpodautoscaler.go
moh-osman3: Thanks for the feedback! I've now placed this logic in `setAutoscalerSpec`

tests/e2e/autoscale/wait-until-hpa-ready.go
```diff
@@ -32,13 +32,15 @@ import (
 func main() {
```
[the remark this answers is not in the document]
moh-osman3: Added a new script to verify the metrics are correct!

tests/e2e/autoscale/02-install.yaml
```diff
@@ -0,0 +1,37 @@
+# This updates an existing deployment with a new Spec.
```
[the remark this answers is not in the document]
moh-osman3: Thanks for outlining the steps for me @iblancasa - this helped me understand how the e2e test is working a lot better. The final step waiting for scale down failed in the e2e tests and I realized that I had to add a timeout (more than 5 minutes) because it seems the HPA can take a lot of time to finally scale down.  I've tried modifying the `scaleDown` behavior but sadly I have not been able to get scale down to happen faster.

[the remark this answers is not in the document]
moh-osman3: For reference this is what I tried adding to remove the 6 minute timeout, but this doesn't seem to have the intended effect.
```
    behavior:
      scaleDown:
        policies:
        - type: Pods
          value: 1
          periodSeconds: 10
        stabilizationWindowSeconds: 1
```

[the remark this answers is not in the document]
moh-osman3: Ah nevermind I now see the issue is that the behavior is not set when the HPA is updated, so I had to add this to the original 0-install.yaml. I created an[ issue](https://github.com/open-telemetry/opentelemetry-operator/issues/1516) for this and have a followup PR in progress. @iblancasa Let me know if you have any feedback on the added script and testing steps - thank you!

tests/e2e/autoscale/03-assert.yaml
```diff
@@ -6,3 +6,11 @@ metadata:
 status:
   scale:
     replicas: 2
+---
+apiVersion: opentelemetry.io/v1alpha1
+kind: OpenTelemetryCollector
+metadata:
+  name: simplest-set-utilization
```
iblancasa, line 13: Before your changes, the test was done against the `simplest` `OpenTelemetryCollector` instance. That one is just to check the old behavior still works. But it would make more sense to do the test against the `simplest-set-utilization` `OpenTelemetryCollector` instance (and will simplify your test).

iblancasa (changes requested):
Good job! Just a minor thing related to simplify the tests.

tests/e2e/autoscale/03-assert.yaml
moh-osman3: Hmm is your recommendation to remove `simplest` collector entirely? I think I'd like to keep it around for now because it also confirms HPA is created with default `targetCPUUtilization` (i.e. scale up  and scale down happens without having to explicitly set the cpu utilization because the default is used).

moh-osman3: Happy to continue on e2e tests improvements on my followup PR though, so let me know what you think!

iblancasa: > Hmm is your recommendation to remove simplest collector entirely? I think I'd like to keep it around for now because it also confirms HPA is created with default targetCPUUtilization (i.e. scale up and scale down happens without having to explicitly set the cpu utilization because the default is used).

Not removing it completely but not checking if it scales. Just checking if the new approach is the one that scales up and down and can be updated (as you did).

moh-osman3: Ah I see, sounds good to me! I now removed the tracegen and scaleup/scaledown assertions from the test for the `simplest` collector. Thanks for the suggestions @iblancasa !

iblancasa (approved):

matej-g (approved):
This looks good me now, nice job @moh-osman3 🙌

jaronoff97 (approved):

PR #1463, opened 2023-02-09
[chore] Refactor platform to be openshift routes availability

### LINKED ISSUE #1418: Replace the autodetect platform logic by something that express in a clearer way that OpenShift Routes are available instead of the Platform used
Issue created from #1338.

The detection checks if OpenShift Routes API is installed. If true, it means we're running the Operator in OpenShift. The problem is that, for the E2E tests, we're installing the OpenShift Routes in KiND (Kubernetes). So, the E2E tests think they are running in OpenShift.

iblancasa:
Closes #1418.

Replace the OpenShift platform detection with OpenShift Routes API detection.


pkg/autodetect/main.go
```diff
@@ -62,21 +60,21 @@ func New(restConfig *rest.Config) (AutoDetect, error) {
 	}, nil
 }
 
-// Platform returns the detected platform this operator is running on. Possible values: Kubernetes, OpenShift.
-func (a *autoDetect) Platform() (platform.Platform, error) {
+// OSRoutesAvailability checks if OpenShift Routes is available.
```
frzifus, line 63: I think it should be:
```suggestion
// OSRoutesAvailability checks if OpenShift Routes are available.
```

pkg/autodetect/openshiftroutes.go
```diff
@@ -0,0 +1,31 @@
...
+
+// OpenShiftRoutesAvailability holds the auto-detected OpenShift Routes availability API.
+type OpenShiftRoutesAvailability int
+
+const (
+	// OpenShiftRoutesAvailable represents the route.openshift.io API is available.
+	OpenShiftRoutesAvailable OpenShiftRoutesAvailability = iota
+
+	// OpenShiftRoutesNotAvailable represents the route.openshift.io API is not available.
+	OpenShiftRoutesNotAvailable OpenShiftRoutesAvailability = iota
```
frzifus, line 26:
```suggestion
	OpenShiftRoutesNotAvailable OpenShiftRoutesAvailability
```
Otherwise its both 0, right?

frzifus (commented):
Cool, only 2 small things ^^

pkg/autodetect/openshiftroutes.go
iblancasa: Both are valid: https://yourbasic.org/golang/iota/

But I'll simplify.

jaronoff97:
@iblancasa could you remove the [chore] in the title and add a changelog entry please? this seems to be a bit more than a minor version upgrade / test fix.

.chloggen/1418-change-platform-to-api-detection.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Change the platform detection to be API detection. Previously, the operator was checking the plaform where it was running -OpenShift or Kubernetes-. Now, we just check if the OpenShift Route API is enabled"
```
frzifus, line 8:
```suggestion
note: "Change the platform detection to be API detection. Previously, the operator was checking the plaform where it was running OpenShift or Kubernetes. Now, we just check if the OpenShift Route API is enabled"
```

frzifus (approved):

.chloggen/1418-change-platform-to-api-detection.yaml
iblancasa: I think this change lefts the sentence with a different meaning or not well written. I would prefer to keep how it was written

pavolloffay: This does not need to be in the changelog. it's an internal refactoring.

pkg/autodetect/main.go
```diff
@@ -21,15 +21,13 @@ import (
...
-
-	"github.com/open-telemetry/opentelemetry-operator/pkg/platform"
 )
 
 var _ AutoDetect = (*autoDetect)(nil)
 
 // AutoDetect provides an assortment of routines that auto-detect traits based on the runtime.
 type AutoDetect interface {
-	Platform() (platform.Platform, error)
+	OSRoutesAvailability() (OpenShiftRoutesAvailability, error)
```
pavolloffay, line 30: s/OS/OCP

pkg/autodetect/main.go
```diff
@@ -62,21 +60,21 @@ func New(restConfig *rest.Config) (AutoDetect, error) {
 	}, nil
 }
 
-// Platform returns the detected platform this operator is running on. Possible values: Kubernetes, OpenShift.
-func (a *autoDetect) Platform() (platform.Platform, error) {
+// OSRoutesAvailability checks if OpenShift Route are available.
+func (a *autoDetect) OSRoutesAvailability() (OpenShiftRoutesAvailability, error) {
```
pavolloffay, line 64: s/OS/OCP

pkg/autodetect/main.go
pavolloffay: or call it just OpenShiftRo... to match other APIs

pavolloffay (changes requested):

pavolloffay:
> @iblancasa could you remove the [chore] in the title and add a changelog entry please? this seems to be a bit more than a minor version upgrade / test fix.

@jaronoff97  do we need to add a changelog for these "internal" changes?

pavolloffay:
The code is exposed, but I think we should consider making most of the packages internal as they were not designed to be imported externally.

jaronoff97:
@pavolloffay I agree, if these were internal only, i don't think the changelog would be necessary. Given these have the potential to be used externally right now, I was thinking it would be valuable to have the entry.

pavolloffay (approved):

Issue #1316, opened 2022-12-12
Documentation Improvements


jaronoff97:
From our operator SIG meeting, we had discussed improving the documentation of the operator and the target allocator. For this to get done, there are a few things we need to do:

- [ ] Add operator examples to the otel website
- [ ] Add target allocator examples to the otel website
- [ ] Add a detailed README to `cmd/otel-allocator`
- [ ] Instrument the community demo using the operator

If anyone has other suggestions on what we should do to improve docs, please add suggestions below.

PR #1472, opened 2023-02-11
[target-allocator] Improve TargetAllocator README (#1316)

peterlindsten:
Add some documentation for what is required for the Target Allocator to work with Prometheus CRs. Assists with #1316

### DIFF
--- a/cmd/otel-allocator/README.md
+++ b/cmd/otel-allocator/README.md
@@ -5,8 +5,9 @@ distribute targets of the PrometheusReceiver on all deployed Collector instances
 operator's most recent release as well.
 
 In essence, Prometheus Receiver configs are "stolen" and replaced with a http_sd_configs directive that points to the 
-Allocator, these are then loadbalanced/sharded to the collectors. In addition, the TargetAllocator can discover targets via Prometheus CRs (currently ServiceMonitor,
-PodMonitor) which it presents on the `/jobs` endpoint.
+Allocator, these are then loadbalanced/sharded to the collectors. In addition, the TargetAllocator can discover targets
+via Prometheus CRs (currently ServiceMonitor, PodMonitor) which it presents as scrape configs and jobs on the
+`/scrape_configs` and `/jobs` endpoints respectively.
 
 # Usage
 The `targetAllocator:` controls the TargetAllocator general properties. Full API spec can be found here: [api.md#opentelemetrycollectorspectargetallocator](../../docs/api.md#opentelemetrycollectorspectargetallocator)
@@ -61,13 +62,13 @@ OpenTelemetry Collector operator specific config.
         target_allocator:
           endpoint: http://my-targetallocator-service
           interval: 30s
-          collector_id: collector-1
+          collector_id: "${POD_NAME}"
 ```
 Upstream documentation here: [Prometheusreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/prometheusreceiver#opentelemetry-operator)
 
 The TargetAllocator service is named based on the OpenTelemetryCollector CR name. `collector_id` should be unique per
-collector instance, such as the pod name. This can be passed down to the pod with env vars and replaced in the config
-with `${ENV}` notation.
+collector instance, such as the pod name. The `POD_NAME` environment variable is convenient since this is supplied
+to collector instance pods by default.
 
 The CRDs naturally also has to exist for the Allocator to pick them up. The best place to get them is from
 prometheus-operator: [Releases](https://github.com/prometheus-operator/prometheus-operator/releases). Only the CRDs for
--- a/cmd/otel-allocator/README.md
+++ b/cmd/otel-allocator/README.md
@@ -4,15 +4,16 @@ The TargetAllocator is an optional separately deployed component of an OpenTelem
 distribute targets of the PrometheusReceiver on all deployed Collector instances. The release version matches the
 operator's most recent release as well.
 
-In essence, Prometheus Receiver configs are "stolen" and replaced with a http_sd_configs directive that points to the 
-Allocator, these are then loadbalanced/sharded to the collectors. In addition, the TargetAllocator can discover targets
-via Prometheus CRs (currently ServiceMonitor, PodMonitor) which it presents as scrape configs and jobs on the
-`/scrape_configs` and `/jobs` endpoints respectively.
+In essence, Prometheus Receiver configs are overridden with a http_sd_configs directive that points to the
+Allocator, these are then loadbalanced/sharded to the collectors. The Prometheus Receiver configs that are overridden
+are what will be distributed with the same name. In addition to picking up receiver configs, the TargetAllocator
+can discover targets via Prometheus CRs (currently ServiceMonitor, PodMonitor) which it presents as scrape configs
+and jobs on the `/scrape_configs` and `/jobs` endpoints respectively.
 
 # Usage
-The `targetAllocator:` controls the TargetAllocator general properties. Full API spec can be found here: [api.md#opentelemetrycollectorspectargetallocator](../../docs/api.md#opentelemetrycollectorspectargetallocator)
+The `spec.targetAllocator:` controls the TargetAllocator general properties. Full API spec can be found here: [api.md#opentelemetrycollectorspectargetallocator](../../docs/api.md#opentelemetrycollectorspectargetallocator)
 
-A basic example that deploys, but does nothing.
+A basic example that deploys.
 ```yaml
 apiVersion: opentelemetry.io/v1alpha1
 kind: OpenTelemetryCollector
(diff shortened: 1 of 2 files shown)


cmd/otel-allocator/README.md
```diff
@@ -4,6 +4,135 @@ The TargetAllocator is an optional separately deployed component of an OpenTelem
...
+        target_allocator:
+          endpoint: http://my-targetallocator-service
+          interval: 30s
+          collector_id: collector-1
+```
+Upstream documentation here: [Prometheusreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/prometheusreceiver#opentelemetry-operator)
+
+The TargetAllocator service is named based on the OpenTelemetryCollector CR name. `collector_id` should be unique per
+collector instance, such as the pod name. This can be passed down to the pod with env vars and replaced in the config
+with `${ENV}` notation.
```
peterlindsten, line 70: I'm unsure of this detail. It seems to work, but I don't know Why. Perhaps someone with more understanding can confirm this is indeed correct?

cmd/otel-allocator/README.md
```diff
@@ -4,6 +4,135 @@ The TargetAllocator is an optional separately deployed component of an OpenTelem
...
+
+The CRDs naturally also has to exist for the Allocator to pick them up. The best place to get them is from
+prometheus-operator: [Releases](https://github.com/prometheus-operator/prometheus-operator/releases). Only the CRDs for
+CRs that the Allocator watches for need to be deployed. They can be picked out from the bundle.yaml file.
```
peterlindsten, line 74: It's a bit of a shame that this is bundled upstream. Is there a way only link the parts we look at, while staying up to date with upstream?
Or just copy because the API is stable (v1)?

cmd/otel-allocator/README.md
```diff
@@ -4,6 +4,135 @@ The TargetAllocator is an optional separately deployed component of an OpenTelem
...
+metadata:
+  name: opentelemetry-targetallocator-role
+rules:
+- apiGroups: [""]
+  resources:
+  - nodes
+  - nodes/metrics
+  - services
+  - endpoints
+  - pods
```
peterlindsten, line 107: Unsure if we actually need all of these

peterlindsten (commented):
Perhaps a more general restructuring is wanted too. Moving the #Design section to a DESIGN.md?

cmd/otel-allocator/README.md
```diff
@@ -4,6 +4,135 @@ The TargetAllocator is an optional separately deployed component of an OpenTelem
 distribute targets of the PrometheusReceiver on all deployed Collector instances. The release version matches the
 operator's most recent release as well.
 
+In essence, Prometheus Receiver configs are "stolen" and replaced with a http_sd_configs directive that points to the 
+Allocator, these are then loadbalanced/sharded to the collectors. In addition, the TargetAllocator can discover targets via Prometheus CRs (currently ServiceMonitor,
+PodMonitor) which it presents on the `/jobs` endpoint.
```
jaronoff97, line 9: The configuration is provided by the `/scrape_configs` endpoint, and the resultant jobs are on the `/jobs` endpoint

cmd/otel-allocator/README.md
jaronoff97: You could replace this to be `${POD_NAME}` (it's automatically supplied by the operator [here](https://github.com/open-telemetry/opentelemetry-operator/blob/6d3da4219a9027aa49728d14036b84d47071b347/pkg/collector/container.go#L88))

cmd/otel-allocator/README.md
jaronoff97: We mostly provide this ability for users migrating to opentelemetry from prometheus rather than someone starting out with the collector. For now, I'd recommend we may want to leave this out and open a follow up issue on deploying these CRs with the operator's CRs to make this easier at which point we can document this more effectively. What do you think @kristinapathak?

cmd/otel-allocator/README.md
jaronoff97: i think all of these are required if someone wants to use the `kubernetes_sd_config`

jaronoff97 (commented):
Thank you so much for adding these much-needed docs. I had a few recommendations/thoughts.

cmd/otel-allocator/README.md
peterlindsten: Fixed

cmd/otel-allocator/README.md
peterlindsten: Updated

cmd/otel-allocator/README.md
```diff
@@ -4,6 +4,136 @@ The TargetAllocator is an optional separately deployed component of an OpenTelem
...
+          scrape_configs:
+          - job_name: 'otel-collector'
+
+    exporters:
+      logging:
+
+    service:
+      pipelines:
+        traces:
+          receivers: [prometheus]
```
avadhut123pisal, line 38: `Prometheus` receiver should be the part of metrics pipeline. I think, same fix is needed on the project level Readme also: https://github.com/open-telemetry/opentelemetry-operator/blob/6d3da4219a9027aa49728d14036b84d47071b347/README.md?plain=1#L352

cmd/otel-allocator/README.md
```diff
@@ -4,6 +4,135 @@ The TargetAllocator is an optional separately deployed component of an OpenTelem
 distribute targets of the PrometheusReceiver on all deployed Collector instances. The release version matches the
 operator's most recent release as well.
 
+In essence, Prometheus Receiver configs are "stolen" and replaced with a http_sd_configs directive that points to the
```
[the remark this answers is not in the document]
jaronoff97: There are two things that happen here: one is the overriding of the service discovery in the collector's scrape config, and the other thing is the target allocator can add new scrape configs. I see them as two distinct actions because one can use the target allocator without the prometheus CR functionality.

cmd/otel-allocator/README.md
```diff
@@ -4,6 +4,136 @@ The TargetAllocator is an optional separately deployed component of an OpenTelem
...
 
+In essence, Prometheus Receiver configs are "stolen" and replaced with a http_sd_configs directive that points to the 
+Allocator, these are then loadbalanced/sharded to the collectors. In addition, the TargetAllocator can discover targets
+via Prometheus CRs (currently ServiceMonitor, PodMonitor) which it presents as scrape configs and jobs on the
+`/scrape_configs` and `/jobs` endpoints respectively.
+
+# Usage
+The `targetAllocator:` controls the TargetAllocator general properties. Full API spec can be found here: [api.md#opentelemetrycollectorspectargetallocator](../../docs/api.md#opentelemetrycollectorspectargetallocator)
+
+A basic example that deploys, but does nothing.
```
matej-g, line 15: Maybe instead of 'does nothing' we could add a static config for `otel-collector`, to provide the simplest example of what allocator does?

cmd/otel-allocator/README.md
```diff
@@ -4,6 +4,136 @@ The TargetAllocator is an optional separately deployed component of an OpenTelem
...
+```
+
+## PrometheusCR specifics
+TargetAllocator discovery of PrometheusCRs can be turned on with
+```yaml
+  targetAllocator:
+    enabled: true
+    prometheusCR:
+      enabled: true
+```
```
matej-g, line 50: Maybe instead of a YAML snippet we could just specify this can be enabled with `prometheusCR.enabled=true` and link to docs for additional features?

cmd/otel-allocator/README.md
```diff
@@ -4,6 +4,136 @@ The TargetAllocator is an optional separately deployed component of an OpenTelem
...
+          interval: 30s
+          collector_id: "${POD_NAME}"
+```
+Upstream documentation here: [Prometheusreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/prometheusreceiver#opentelemetry-operator)
+
+The TargetAllocator service is named based on the OpenTelemetryCollector CR name. `collector_id` should be unique per
+collector instance, such as the pod name. The `POD_NAME` environment variable is convenient since this is supplied
+to collector instance pods by default.
+
+The CRDs naturally also has to exist for the Allocator to pick them up. The best place to get them is from
```
matej-g, line 73:
```suggestion
The Prometheus CRDs also have to exist for the Allocator to pick them up. The best place to get them is from
```

Otherwise I'd agree with @jaronoff97 to perhaps leave this out for now and do it in follow up?

matej-g (commented):
Nice job! I added few small-ish suggestions.

One thing I personally would like also documented is different allocation strategies (and related number of replicas). But this can always be done in a follow up.

cmd/otel-allocator/README.md
peterlindsten: Sounds good, resolving

cmd/otel-allocator/README.md
peterlindsten: I think it should be mentioned that this is not done by automagic (which is very common in the k8s world). It can be removed later in favour of a link to a proper doc. Accepting your change.

cmd/otel-allocator/README.md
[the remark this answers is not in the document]
peterlindsten: I'll have a look at rewriting this section this weekend to clarify.

cmd/otel-allocator/README.md
peterlindsten: Blind copy-paste error. Oops

cmd/otel-allocator/README.md
peterlindsten: Sure, with a full path, I think that is OK.
The reason I did it this way is to show the the targetAllocator has to be enabled - something I can totally see someone leaving out.

cmd/otel-allocator/README.md
peterlindsten: This is what the main README.md does - which demonstrates that the allocator places itself as a http_sd_config. What is confusing about that as a newcomer is that the shown scrape_config doesn't seem to do anything, and its really hard to understand which collector will scrape. This is tied into the allocation strategies (which I do not yet understand, so cannot really write a description of). The 'does nothing' case allows for less astonishment to me, until there is a more exact description/example of what happens.

cmd/otel-allocator/README.md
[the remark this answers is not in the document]
peterlindsten: Updated

cmd/otel-allocator/README.md
peterlindsten: Fixed, and in main README too

cmd/otel-allocator/README.md
peterlindsten: Updated to be a copy of the main readme.

cmd/otel-allocator/README.md
peterlindsten: Changed to full path.

jaronoff97 (approved):

PR #1479, opened 2023-02-16
[target-allocator] Fix pprof endpoints

### LINKED ISSUE #1478: [target allocator] Pprof endpoint does not correctly handle all profile types
Recently, we added pprof endpoints in https://github.com/open-telemetry/opentelemetry-operator/pull/1408, but it seems we are not handling all types of profiles. 

For example, trying to obtain the heap profile (``) leads to `404` I response. I believe this is due to quirk in how pprof HTTP handles paths (some are under `Index` handler, others have their own handler) and how `gin` routes - we're only handling the actual index (`/`) but not other paths that belong under index handler (e.g. `/debug/pprof/heap`).

matej-g:
Fixes #1478.

Due to the fact that `gin` does not allow mixing wildcards and specified paths, I opted for using a handler with wildcard which then uses the correct, specific pprof handler _or_ falls back to the default index handler.


cmd/otel-allocator/go.mod
```diff
@@ -173,7 +173,6 @@ require (
 	google.golang.org/genproto v0.0.0-20220808204814-fd01256a5276 // indirect
 	google.golang.org/grpc v1.48.0 // indirect
 	google.golang.org/protobuf v1.28.1 // indirect
-	gopkg.in/go-playground/validator.v9 v9.29.1 // indirect
```
matej-g, line 176: Not 100% sure about this, `go mod tidy` removed the entry.

cmd/otel-allocator/server/server.go
```diff
@@ -190,10 +191,22 @@ func GetAllTargetsByJob(allocator allocation.Allocator, job string) map[string]c
 	return displayData
 }
 
+// registerPprof registers the pprof handlers and either serves the requested
+// specific profile or falls back to index handler.
 func registerPprof(g *gin.RouterGroup) {
```
kristinapathak, line 196: Is it worthwhile to have any tests for this or not really?

kristinapathak (approved):

pavolloffay (approved):

frzifus (approved):

cmd/otel-allocator/server/server.go
matej-g: Hm commonly I would not write tests for such non-core functionality. I tried the different profile manually and all works fine. I'd assume since this isn't the critical part of the allocator and I would not expect this functionality to really change, we should be fine without tests.

PR #1498, opened 2023-02-22
Feature/upgrade operator-sdk 1.27.0

### LINKED ISSUE #1501: Upgrade operator-sdk to 1.27.0
Upgrade the `operator-sdk` version to 1.27.0

yuriolisa:
Resolves #1501


bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -31,9 +31,9 @@ metadata:
     categories: Logging & Tracing
     certified: "false"
     containerImage: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
-    createdAt: "2020-12-16T13:37:00+00:00"
```
iblancasa, line 34: Since operator-sdk 1.26.0, `make bundle` changes the `createdAt` field from the bundle even if it is patched:
https://github.com/operator-framework/operator-sdk/pull/6136

I added a workaround this issue in the Jaeger Operator:
https://github.com/jaegertracing/jaeger-operator/blob/main/Makefile#L356
https://github.com/jaegertracing/jaeger-operator/blob/main/hack/ignore-createdAt-bundle.sh

iblancasa (changes requested):

iblancasa:
Update the version in the scorecar tests

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
yuriolisa: Thank you for the heads-up. I've just submitted the fix.

yuriolisa:
> Update the version in the scorecar tests

It shouldn't be updated.

pavolloffay:
> It shouldn't be updated.

@yuriolisa any reason why it should not be updated? It was always bumped as part of the SDK version bump. Also see the example operator https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v3/memcached-operator/bundle/tests/scorecard/config.yaml

pavolloffay (commented):
After the update the project structure and dependency stack should be aligned with https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v3/memcached-operator

Therefore:
* `CONTROLLER_TOOLS_VERSION`  should be bumped to 0.10.0

yuriolisa:
> > It shouldn't be updated.
> 
> @yuriolisa any reason why it should not be updated? It was always bumped as part of the SDK version bump. Also see the example operator https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v3/memcached-operator/bundle/tests/scorecard/config.yaml

Thank you for the heads-up. I've mislooked at the scorecard-test version.

pavolloffay:
@yuriolisa unit tests failed

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -31,9 +31,9 @@ metadata:
...
+    createdAt: "2023-03-02T10:14:11Z"
```
frzifus, line 34: hm.. is this information displayed somewhere? Ignoring it makes sense, otherwise our diff test would fail. But at the same time, this field will be `2023-03-02T10:14:11Z` for ever.

frzifus (approved):

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
yuriolisa: AFAIK, since the `operator-sdk` version `1.26`, the `createdAt` field is being changed.  Due to that, @iblancasa pointed to creating a workaround for it.

https://github.com/operator-framework/operator-sdk/issues/6285

iblancasa: It seems it was not updated until now. So... it seems it is not really important.

iblancasa (approved):

yuriolisa:
@pavolloffay , could you please have a look?

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
frzifus: @yuriolisa yes, that thing helps to ignore the timestamp. But is it right to ignore this one? 
- If `createdAt` describes when the operator was first created, we should not update the timestamp.
- If the field describes when the shipped version was created, we should update the field with each release.

(My understanding is that a field like `createdAt` should never change and instead a field like `updatedAt` should be changed with each release.)

On [Operatorhub](https://operatorhub.io/operator/opentelemetry-operator) i found this:
image

iblancasa: Since @frzifus reacted with `👎🏻`, I'll try to clarify my comment.

> It seems it was not updated until now. So... it seems it is not really important.

By this I meant: if it was expected to be updated, nobody paid attention to it. Sorry if I was not clear enough.
 
I think we should not update it but since it seems it is not clear what it means... I don't have a strong opinion. It can be:
* When the bundle was created
* When the operator was created

PR #1151, opened 2022-10-08
Change app image and context propagator b3 to b3multi in .Net e2e test case 

avadhut123pisal:
Current docker image of the application used in the e2e test case for .Net auto-instrumentation throws error due to some dependency related issues. This PR replaces that image with the new sample asp .Net application image.

Also, previously for e2e test cases for .Net auto-instrumentation, b3 propagator was configured. But, as per the [opentelemetry-dotnet-instrumentation release v0.3.0-beta.1](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/releases/tag/v0.3.0-beta.1), supported propagators are `b3multi, baggage, tracecontext`. So, with this PR b3 propagator is replaced with b3multi.

PR #1499, opened 2023-02-22
[chore] Use tagged images in e2e tests to not trigger docker.com rate limit

andreasgerstmayr:
Kubernetes uses imagePullPolicy: Always for images with latest or no tag, which can trigger docker.com rate limits if one runs the e2e tests too often.

### DIFF
--- a/tests/e2e/instrumentation-dotnet-multicontainer/01-install-app.yaml
+++ b/tests/e2e/instrumentation-dotnet-multicontainer/01-install-app.yaml
@@ -18,6 +18,6 @@ spec:
     spec:
       containers:
       - name: myapp
-        image: docker.io/avadhutp123/aspnetapp:latest # source code of the application: https://github.com/dotnet/dotnet-docker/tree/main/samples/aspnetapp
+        image: docker.io/avadhutp123/aspnetapp@sha256:d2e8d3415f6f12efae0369aa0a9777a58583841fb133f33e10a73adb9fb392da # source code of the application: https://github.com/dotnet/dotnet-docker/tree/main/samples/aspnetapp
       - name: myrabbit
-        image: rabbitmq
+        image: rabbitmq:3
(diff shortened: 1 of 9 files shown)


tests/e2e/instrumentation-dotnet-multicontainer/01-install-app.yaml
```diff
@@ -18,6 +18,6 @@ spec:
     spec:
       containers:
       - name: myapp
-        image: docker.io/avadhutp123/aspnetapp:latest # source code of the application: https://github.com/dotnet/dotnet-docker/tree/main/samples/aspnetapp
+        image: docker.io/avadhutp123/aspnetapp@sha256:d2e8d3415f6f12efae0369aa0a9777a58583841fb133f33e10a73adb9fb392da # source code of the application: https://github.com/dotnet/dotnet-docker/tree/main/samples/aspnetapp
```
jaronoff97, line 21: @pavolloffay do you know why we're not using the dotnet images from microsoft? i.e. `mcr.microsoft.com/dotnet/samples:aspnetapp`?

iblancasa: @jaronoff97 The reason is described in https://github.com/open-telemetry/opentelemetry-operator/pull/1151

tests/e2e/instrumentation-python-multicontainer/02-install-app.yaml
```diff
@@ -20,4 +20,4 @@ spec:
       - name: myapp
         image: ghcr.io/anuraaga/flask-hello-world:latest
```
iblancasa, line 21: I know the problem is related to the Docker Hub images but... should we apply the same philosophy to all the images?

iblancasa (approved):

pavolloffay (approved):

PR #1500, opened 2023-02-22
[chore] Fix `make` on MacOS by using `gsed`

matej-g:
Opt for using `gsed` instead of the MacOS version of `sed`, which is incompatible with certain flags that work on Linux.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,6 +72,10 @@ LOCALBIN ?= $(shell pwd)/bin
 $(LOCALBIN):
 	mkdir -p $(LOCALBIN)
 
+## On MacOS, use gsed instead of sed, to make sed behavior
+## consistent with Linux.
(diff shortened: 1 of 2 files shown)


iblancasa (changes requested):
Could you send a PR for each change? Since they are not related, I think it would make more sense.

Also, add `[chore]` to the title of your PRs to skip the changelog.

matej-g:
@iblancasa done

iblancasa (changes requested):
Could you rerun the CI?
```sh
git commit -sm "Add commit to trigger build" --allow-empty
```

matej-g:
I assume the failure is a flaky test and not related @open-telemetry/operator-approvers?

iblancasa (approved):

jaronoff97 (approved):

PR #1505, opened 2023-02-23
Fix annotations propagation

### LINKED ISSUE #900: Propagate annotations to Deployment's PodTemplate
Annotations on an `OpenTelemetryCollector` resource in `mode: deployment` are propagated to the created `Deployment` object's `metadata.annotations` but not added to the `spec.template.spec.metadata.annotations` for the pods.

As a result annotations to control the behaviour of other components that inject sidecars etc are ineffective on the operator-controlled pods. For example, setting `sidecar.istio.io/interceptionMode` to `TPROXY` has no effect.

yuriolisa:
Fixes #900

### DIFF
--- a/pkg/collector/annotations.go
+++ b/pkg/collector/annotations.go
@@ -43,22 +43,6 @@ func Annotations(instance v1alpha1.OpenTelemetryCollector) map[string]string {
 	return annotations
 }
 
-// PodAnnotations return the spec annotations for OpenTelemetryCollector pod.
-func PodAnnotations(instance v1alpha1.OpenTelemetryCollector) map[string]string {
-	// new map every time, so that we don't touch the instance's annotations
-	podAnnotations := map[string]string{}
-
-	// allow override of pod annotations
-	for k, v := range instance.Spec.PodAnnotations {
-		podAnnotations[k] = v
-	}
-
-	// make sure sha256 for configMap is always calculated
-	podAnnotations["opentelemetry-operator-config/sha256"] = getConfigMapSHA(instance.Spec.Config)
-
-	return podAnnotations
-}
-
 func getConfigMapSHA(config string) string {
 	h := sha256.Sum256([]byte(config))
 	return fmt.Sprintf("%x", h)
--- a/pkg/collector/daemonset_test.go
+++ b/pkg/collector/daemonset_test.go
@@ -55,6 +55,9 @@ func TestDaemonSetNewDefault(t *testing.T) {
 	// verify sha256 podAnnotation
 	expectedAnnotations := map[string]string{
 		"opentelemetry-operator-config/sha256": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
+		"prometheus.io/path":                   "/metrics",
+		"prometheus.io/port":                   "8888",
+		"prometheus.io/scrape":                 "true",
 	}
 	assert.Equal(t, expectedAnnotations, d.Spec.Template.Annotations)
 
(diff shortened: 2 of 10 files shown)


pavolloffay:
@yuriolisa please add changelog entry

pkg/collector/annotations.go
```diff
@@ -43,22 +43,6 @@ func Annotations(instance v1alpha1.OpenTelemetryCollector) map[string]string {
 	return annotations
 }
 
-// PodAnnotations return the spec annotations for OpenTelemetryCollector pod.
-func PodAnnotations(instance v1alpha1.OpenTelemetryCollector) map[string]string {
```
pavolloffay, line 47: I wonder why the pod annotations were separated from the others. Did you try to look it up in the history?

yuriolisa: I took a look at the history, but it seems this behavior was a try to fix a bug related to reconciling the collector once the configmap changes. However, I don't know why it wasn't simply added the sha256 onto the Deployment annotations and replicate them to the podAnnotations.

pavolloffay: https://github.com/open-telemetry/opentelemetry-operator/pull/670

LGTM to get rid of it

pkg/collector/daemonset_test.go
```diff
@@ -103,9 +106,17 @@ func TestDaemonsetHostNetwork(t *testing.T) {
 func TestDaemonsetPodAnnotations(t *testing.T) {
 	// prepare
 	testPodAnnotationValues := map[string]string{"annotation-key": "annotation-value"}
+
+	// Add sha256 podAnnotation
+	testPodAnnotationValues["opentelemetry-operator-config/sha256"] = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
```
pavolloffay, line 111: this can be moved up to the declaration of `testPodAnnotationValues`

pkg/collector/annotations.go
yuriolisa: Do you think we could have a situation in which pods should have a specific annotation not present on Deployment's annotations?

pkg/collector/deployment.go
```diff
@@ -48,7 +47,7 @@ func Deployment(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTele
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Labels:      labels,
-					Annotations: podAnnotations,
+					Annotations: annotations,
```
jaronoff97, line 50: Won't this introduce a regression where you can no longer set specific pod annotations? I rely on exactly this currently in a few places.

pavolloffay: correct , I missed this 👍🏼

yuriolisa: Okay, so, to avoid this kind of regression, I would propagate the annotations into podAnnotations. WDYT?

jaronoff97: I think it would be fine to append the annotations in to the pod annotations. How would you handle conflicts?

yuriolisa: WDYT to take the Deployment's annotations as precedence?

yuriolisa: @pavolloffay , @jaronoff97 ?

jaronoff97: @yuriolisa i think the pod annotations should win, otherwise there would be no way to override something on the deployment.

Issue #1385, opened 2023-01-20
Test on Kuberentes 1.26.0


pavolloffay:
https://kubernetes.io/releases/#release-v1-26

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-26

```
The autoscaling/v2beta2 API version of HorizontalPodAutoscaler is no longer served as of v1.26.

Migrate manifests and API clients to use the autoscaling/v2 API version, available since v1.23.
All existing persisted objects are accessible via the new API
```

PR #1506, opened 2023-02-23
Bump k8s dependencies to 1.26.x

pavolloffay:
Related to #1385


pkg/collector/reconcile/configmap_test.go
```diff
@@ -103,8 +103,10 @@ receivers:
         metrics_path: /metrics
         scheme: http
         follow_redirects: true
+        enable_http2: true
```
pavolloffay, line 106: @jaronoff97 could you please review this part? I am not sure sure why this has changed. It seems that bump of the prometheus dependency brought this here.

pkg/collector/reconcile/configmap_test.go
```diff
@@ -103,8 +103,10 @@ receivers:
...
         http_sd_configs:
         - follow_redirects: false
+          enable_http2: false
```
pavolloffay, line 109: @jaronoff97 here as well

Issue #1089, opened 2022-09-13
Opentelemetry operator able to create and mount array of config / configmap [yaml's] instead of single configmap creation


kabhanus007:
cc: @pavankrish123 @jpkrohling 
Hi Opentelemetry Operator team,

**Background Information**
1. Opentelemetry collector now supports multiple configuration https://github.com/open-telemetry/opentelemetry-collector/pull/4742
2. Currently opentelemetry operator takes single configuration yaml as spec.config for opentelemetry collector CRD
3. We would like to leverage point (1) above, and pass in configuration as multiple YAML objects.

**Proposed Solution**:
1. Instead of single `spec.config` for Opentelemetry collector CRD, we propose a new field, `configYamls` which is an array of yaml objects instead of single YAML. 
2. These YAML objects will eventually be written as multiple YAML files in the otelcollector pod managed by opentelemetry operator

**Sample Implementation**:

```
apiVersion: opentelemetry.io/v1alpha1
kind: OpenTelemetryCollector
metadata:
  name: simplest
spec:
  configYamls:
  - prometheusConf: |
     // yaml configuration inside which will be used to create prometheus configmap
  - otelBaseConf: | 
     // yaml configuration inside which will be used to create otelBaseConf configmap
```
Once the above configuration is applied, 2 files `prometheusConf.yaml` and `otelBaseConf.yaml` will be created in the pod and `./otelcorecol --config=file:prometheusConf.yaml --config=file:otelBaseConf.yaml` needs to be executed as start cmd.

We use a helm chart to manage these CRD and we would like to leverage this multiple configuration feature to inject multiple configuration from multiple teams through separate helm template files. So far we have been using helm logic to merge these configurations, but it is getting complicated. With this implementation, we can completely eliminate the merge logic from our helm chart.

Aneurysm9:
Beyond specifying multiple YAML configuration blocks, it might also be nice to be able to reference configuration sources by URI to allow loading config from other than a config map created by the operator.

kabhanus007:
> Beyond specifying multiple YAML configuration blocks, it might also be nice to be able to reference configuration sources by URI to allow loading config from other than a config map created by the operator.

@Aneurysm9 , maybe if you can add some example to elaborate your point in a much better way to all of them. But yes, 
 that's a kind of another good suggestion to add on top of this.

yuriolisa:
@kabhanus007, bringing the ideas together, I would suggest a third via, like: 

````
[...]

PR #1519, opened 2023-02-26
fix: enforce sorted args when creating a container object for consistency

### LINKED ISSUE #1460: Competing replicaset when using args
It would appear that using `spec.args` causes the operator to spin up `ReplicaSet` of all ordering/variant due to the way that the string array `args` does not enforce an order.  I have observed this in EKS v1.21, and have been able to replicate this in a local kind cluster for v1.21, v.1.22, and v1.25.  For example, if I use the following collector with the operator deployed using the official helm chart straight up:

```yaml
 kind: OpenTelemetryCollector
 metadata:
   name: otel
 spec:
   image: 'otel/opentelemetry-collector-contrib:0.69.0'
   args:
     feature-gates: +exporter.datadogexporter.metricexportnativeclient
   config: |
     receivers:
       otlp:
         protocols:
           grpc:
           http:
     processors:
 
     exporters:
       logging:
 
     service:
       pipelines:
         traces:
           receivers: [otlp]
           processors: []
           exporters: [logging]
```

then we can see that two sets of replica will spin up where one will be the dominant.  This may not happen every time as it does depend on how the array is arranged initially.  This is driven by the `Deployment` where the `revision` gets a bump whenever the operator wants to reconcile. To confirm, we can get the output via

```sh
kubectl get rs -o jsonpath='{range .items[*]}{.spec.template.spec.containers[0].args}'
# my output
➜ ["--config=/conf/collector.yaml","--feature-gates=+exporter.datadogexporter.metricexportnativeclient"]["--feature-gates=+exporter.datadogexporter.metricexportnativeclient","--config=/conf/collector.yaml"]
```
which shows that both ordering exists.

If we add the following test (to `pkg/collector/deployment_test.go`) representing  a minimal version of the config above, we can see that the test flip between pass and fail (probably bias to one side dependent for reasons beyond my understanding).

```golang
func TestDeploymentArgs(t *testing.T) {
    otelcol := v1alpha1.OpenTelemetryCollector{
          Name: "my-instance",
       },
       Spec: v1alpha1.OpenTelemetryCollectorSpec{
          Args: map[string]string{
             "feature-gates": "+random-feature",
          },
       },
    }
    cfg := config.New()
    d := Deployment(cfg, logger, otelcol)
    assert.Equal(t, "--config=/conf/collector.yaml", d.Spec.Template.Spec.Containers[0].Args[0])
    assert.Equal(t, "--feature-gates=+random-feature", d.Spec.Template.Spec.Containers[0].Args[1])
}
```

I haven't looked much further to see if there is a better way to use the crd or there is a quick fix in the operator itself.

edwintye:
Fixes #1460 where the `args` used to create `corev1.Container` is not ordered hence during reconcile the operator may decide that the existing `ReplicaSet` has drifted and create a new one.  This PR enforces a sorted args when creating a container object for consistent creation/reconcile.  Added a rudimentary test for flag `addConfig=true` where `config` is always the first item.


pkg/collector/container.go
```diff
@@ -57,23 +57,30 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
...
 
-	var args []string
+	var sortedArgs []string
 	for k, v := range argsMap {
-		args = append(args, fmt.Sprintf("--%s=%s", k, v))
+		sortedArgs = append(sortedArgs, fmt.Sprintf("--%s=%s", k, v))
 	}
+	// ensure that the args are ordered which iterating over a map does not guarantee and append to output args
+	sort.Strings(sortedArgs)
+	args = append(args, sortedArgs...)
```
jaronoff97, line 83: Why does a new array need to be introduced here? Can you not just sort the args array above?

edwintye: We absolutely can, and I guess it depends on your intended behaviour where `config` should be the first item or not.  The simplest solution can simply be a one-liner of `sort.Strings(args)` after iterating `argsMap`, which was the first thing I suggested in the issue. In that case tho, we should definitely put more info into the tests on why the ordering (in the test) is expected; maybe I should provide more info anyway 🤔

jaronoff97: yeah more info would be great – i don't _think_ the ordering matters here. @Aneurysm9 may have more info though.

Aneurysm9: I'm not sure that args being a map in the spec is correct at all.  Arguments such as `--config` and `--feature-gates` can be specified multiple times and the order (at least for `--config`) can be important as it will govern the order in which configuration sources are applied to build a final configuration.  Instead we should probably be taking a list of arguments directly and using it as-is.

edwintye: Thanks a lot for that information. Although I fully support changing `args` from a map to an array, that is going to be a breaking change and most certainly beyond this scope of the problem here.  Seems like this has already been discussed before inhttps://github.com/open-telemetry/opentelemetry-operator/issues/1089 but slightly deviates from the thought process of providing an array of `--config=<file from configmap>` into `args`.   

I will update the PR with comments and my thought process in the next couple of days and maybe a clearer picture will emerge.

Aneurysm9: I think the approach Juraci recommended [here](https://github.com/open-telemetry/opentelemetry-operator/issues/1089#issuecomment-1257956898) is probably the way to go.  Rather than changing `args` from a map to a list we add a new field for providing a list of configuration URIs.  The feature gates flag is order independent and can be provided as a comma-separated list, so I'm less worried about that.

jaronoff97 (approved):

pkg/collector/container.go
```diff
@@ -57,23 +57,38 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
...
 	}
+	// defines the output (sorted) array for final output
+	var args []string
+	// When adding a config via v1alpha1.OpenTelemetryCollectorSpec.Config, we ensure that it is always the
+	// first item in the args. At the time of writing, although multiple configs are allowed in the
+	// opentelemetry collector, the operator has yet to implement such functionality.  When multiple configs
+	// are present they should be merged in a deterministic manner using the order given, and because
+	// v1alpha1.OpenTelemetryCollectorSpec.Config is a required field we assume that it will always be the
+	// "primary" config and in the future additional configs can be appended to the container args in a simple manner.
+	// Note that ordering is only important for config and not other arguments such as feature gates or etc.
```
Aneurysm9, line 68:
```suggestion
	// Note that ordering is presently only important for config and not other arguments such as feature gates or etc.
```

Or this line can be removed entirely.  More arguments may be added in the future and they may be order dependent.

Aneurysm9 (approved):
I think this works as an intermediate solution.  A proper solution for multiple config sources will still be needed and will need to ensure that the user-provided order is respected rather than re-sorted based on URI.

pavolloffay (approved):

pkg/collector/container.go
edwintye: Thanks for the comment.  I have deleted the line.  You are right that we may have other order dependent args in the future.

frzifus (approved):

Issue #908, opened 2022-06-03
Add support for golang auto-instrumentation


pavolloffay:
The https://github.com/keyval-dev/opentelemetry-go-instrumentation could be used for golang auto-instrumentation. 

From the object management perspective it would be better if the project was part of the CNCF/OTEL community.

cc) @edeNFed

pavolloffay:
Golang auto-instrumentation has been donated https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/2

TylerHelmuth:
@pavolloffay @open-telemetry/operator-maintainers @open-telemetry/operator-approvers if the community is willing, I would like to start working on this.

edeNFed:
@TylerHelmuth I'll be happy to help if needed

TylerHelmuth:
@edeNFed sounds good! I'm sure I'll have some questions once I get started.

TylerHelmuth:
Please assign this to me.

TylerHelmuth:
Question: will we want to build our own version of the go auto-instrumentation agent like we do for the other languages or will it be acceptable to use the image published by opentelemetry-go-instrumentation?

TylerHelmuth:
As discussed in the SIG meeting today, we'll move forward with using an image published from open-telemetry/opentelemetry-go-instrumentation instead of building our own.  This decision is made based on the increased complexity of the go auto instrumentation due to eBPF and the necessary offsets json file we would have to manage if we had our own image.

PR #1555, opened 2023-03-03
Add support for Go auto-instrumentation

TylerHelmuth:
Adds support for Go auto-instrumentation injection.

Uses the [opentelemetry-go-instrumentation](https://github.com/open-telemetry/opentelemetry-go-instrumentation) to auto-instrument Go services.  Instead of building our own image, we rely on the image published by that repository.

Unlike the other languages currently injectable, Go auto-instrumentation relies on a sidecar instead of an init container.  The auto-instrumentation also depends on an environment variable, `OTEL_TARGET_EXE`, to be set with the service's executable path.  Users must either set this value in the Instrumentation resource or provide a specific annotation, with the latter taking precedence.

I also didn't add support for multi-container pods in this first iteration.  Since the Go instrumentation adds a side car, I wasn't entirely sure how multi-container support would work: do we need a sidecar per pod or could 1 agent do it all?  Since this is still so new, I figured keeping things simple and only supporting single pods was best.

Depends on:
- [x] https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/31
- [x] https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/36
- [x] https://github.com/open-telemetry/opentelemetry-operator/pull/1619
- [x] https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/66
- [x] https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/131

Closes https://github.com/open-telemetry/opentelemetry-operator/issues/908


pkg/instrumentation/podmutator.go
```diff
@@ -104,14 +111,25 @@ func (pm *instPodMutator) Mutate(ctx context.Context, ns corev1.Namespace, pod c
...
 
+	if inst, err = pm.getInstrumentationInstance(ctx, ns, pod, annotationInjectGolang); err != nil {
+		// we still allow the pod to be created, but we log a message to the operator's logs
+		logger.Error(err, "failed to select an OpenTelemetry Instrumentation instance for this pod")
+		return pod, err
+	}
+	if EnableGolangAutoInstrumentationSupport.IsEnabled() {
+		insts.Golang = inst
+	} else {
+		logger.Error(err, "support for Golang auto instrumentation is not enable")
```
jaronoff97, line 122: I'm trying to understand what the resulting behavior would be here if someone attempts to use golang auto inst if the flag hasn't been enabled yet. Would they receive a mutated pod with some of the added defaults, but not have the added sidecar from the mutator? Should we return an error here instead so someone would be informed they need to enable the feature gate before they can use this feature?

tests/e2e/instrumentation-golang/01-install-app.yaml
```diff
@@ -0,0 +1,24 @@
...
+      labels:
+        app: my-pod-with-sidecar
+      annotations:
+        sidecar.opentelemetry.io/inject: "true"
+        instrumentation.opentelemetry.io/inject-golang: "true"
+        instrumentation.opentelemetry.io/golang-target-exec: /usr/local/bin/emojivoto-voting-svc
+    spec:
+      containers:
+      - name: voting-svc
+        image: docker.l5d.io/buoyantio/emojivoto-voting-svc:v11
```
jaronoff97, line 21: how was this image chosen? is the license for the image permissive of us using it?

main.go
```diff
@@ -112,6 +113,7 @@ func main() {
...
 	pflag.StringVar(&autoInstrumentationPython, "auto-instrumentation-python-image", fmt.Sprintf("ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-python:%s", v.AutoInstrumentationPython), "The default OpenTelemetry Python instrumentation image. This image is used when no image is specified in the CustomResource.")
 	pflag.StringVar(&autoInstrumentationDotNet, "auto-instrumentation-dotnet-image", fmt.Sprintf("ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-dotnet:%s", v.AutoInstrumentationDotNet), "The default OpenTelemetry DotNet instrumentation image. This image is used when no image is specified in the CustomResource.")
+	pflag.StringVar(&autoInstrumentationGolang, "auto-instrumentation-golang-image", fmt.Sprintf("keyval/otel-go-agent:%s", v.AutoInstrumentationGolang), "The default OpenTelemetry Golang instrumentation image. This image is used when no image is specified in the CustomResource.")
```
jaronoff97, line 116: Is there not a published otel image yet?

jaronoff97: seems [this issue](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/31) has been closed

versions.txt
```diff
@@ -29,6 +29,9 @@ autoinstrumentation-python=0.36b0
 # Should match autoinstrumentation/dotnet/version.txt
 autoinstrumentation-dotnet=0.6.0
 
+# Represents the current release of Golang instrumentation.
+autoinstrumentation-golang=v0.6.4
```
jaronoff97, line 33: The image naming pattern we use for everything omits the v here, not sure it matters / will change in the future, just a thought that we may want to match it here and then add the v in the sprintf in the operator?

jaronoff97 (changes requested):

config/manager/manager.yaml
```diff
@@ -29,6 +29,7 @@ spec:
       containers:
       - args:
         - --enable-leader-election
+        - --feature-gates=+operator.autoinstrumentation.golang
```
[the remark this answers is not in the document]
TylerHelmuth: This was an attempt to enable the flag for e2e tests.  We definitely do not want to enable this by default. Where is the best place to adjust the operator that is installed for the e2e tests?

pkg/instrumentation/podmutator.go
TylerHelmuth: My intention was that if a pod adds the `annotationInjectGolang` annotation and the operator has not enabled the feature gate then their pod is not mutated at all for Goland.  I didn't return an error bc I wanted other  `annotationInject*` annotations on the pod to still be respected.

tests/e2e/instrumentation-golang/01-install-app.yaml
TylerHelmuth: It was chosen based on the existing opentelemetry-go-instrumentaion docs example, but the repo has some new e2e tests with their own sample app that we could probably use now instead.

main.go
TylerHelmuth: We got a release workflow, but not an actual release.  [This issue is tracking the actual release.](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/66).  We should definitely not merge this PR until we have an official OTel image.  Once we do, I'll update this PR and retest, but should be a small update that doesn't impact the core logic of the PR.

versions.txt
TylerHelmuth: I can do that if you want.  I chose this value initially as it represents the exact tag value of the image, which is what version values in the file represent.  This keeps the code in main.go consistent.

config/manager/manager.yaml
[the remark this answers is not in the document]
TylerHelmuth: Figure out the correct spot (I think)

README.md
```diff
@@ -183,7 +183,7 @@ When using sidecar mode the OpenTelemetry collector container will have the envi
 
 ### OpenTelemetry auto-instrumentation injection
 
-The operator can inject and configure OpenTelemetry auto-instrumentation libraries. Currently DotNet, Java, NodeJS and Python are supported.
+The operator can inject and configure OpenTelemetry auto-instrumentation libraries. Currently DotNet, Golang, Java, NodeJS and Python are supported.
```
abtris, line 186: Please use official name of language. 
```suggestion
The operator can inject and configure OpenTelemetry auto-instrumentation libraries. Currently DotNet, Go, Java, NodeJS and Python are supported.
```

TylerHelmuth: I switched all occurrences except for renamed `golang.go`.

tests/e2e/instrumentation-golang/01-install-app.yaml
TylerHelmuth: The opentelemetry-go-instrumentation repo's sample app is build during the e2e workflow and isn't currently published.  The license for https://github.com/BuoyantIO/emojivoto is [Apache License 2.0](https://github.com/BuoyantIO/emojivoto/blob/main/LICENSE) so believe we are allowed to use it.  I will add a link back to the repo like the dotnet image does.

TylerHelmuth: Actually, let me try one of the opentelemetry-demo Go images.

main.go
jaronoff97: kk, i'll merge as soon as that issue is complete.

pkg/instrumentation/podmutator.go
```diff
@@ -104,14 +111,25 @@ func (pm *instPodMutator) Mutate(ctx context.Context, ns corev1.Namespace, pod c
...
 
+	if inst, err = pm.getInstrumentationInstance(ctx, ns, pod, annotationInjectGo); err != nil {
+		// we still allow the pod to be created, but we log a message to the operator's logs
+		logger.Error(err, "failed to select an OpenTelemetry Instrumentation instance for this pod")
+		return pod, err
+	}
+	if EnableGoAutoInstrumentationSupport.IsEnabled() {
+		insts.Go = inst
+	} else {
+		logger.Error(err, "support for Go auto instrumentation is not enable")
```
jaronoff97, line 122:
```suggestion
		logger.Error(err, "support for Go auto instrumentation is not enabled")
```

jaronoff97: I think its fine to not return the error here, just expecting someone to come in to the channel and wonder why their go instrumentation isn't working without seeing this log.

pkg/instrumentation/sdk.go
```diff
@@ -132,14 +154,14 @@ func (i *sdkInjector) injectCommonEnvVar(otelinst v1alpha1.Instrumentation, pod
 	return pod
 }
 
-func (i *sdkInjector) injectCommonSDKConfig(ctx context.Context, otelinst v1alpha1.Instrumentation, ns corev1.Namespace, pod corev1.Pod, index int) corev1.Pod {
-	container := &pod.Spec.Containers[index]
-	resourceMap := i.createResourceMap(ctx, otelinst, ns, pod, index)
+func (i *sdkInjector) injectCommonSDKConfig(ctx context.Context, otelinst v1alpha1.Instrumentation, ns corev1.Namespace, pod corev1.Pod, agentIndex int, appIndex int) corev1.Pod {
```
jaronoff97, line 157: Could you add a docstring here explaining the difference between agentIndex and appIndex

tests/e2e/instrumentation-go/01-install-app.yaml
```diff
@@ -0,0 +1,21 @@
...
+      app: my-pod-with-sidecar
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: my-pod-with-sidecar
+      annotations:
+        sidecar.opentelemetry.io/inject: "true"
+        instrumentation.opentelemetry.io/inject-go: "true"
+        instrumentation.opentelemetry.io/go-target-exec: /usr/local/bin/emojivoto-voting-svc
```
jaronoff97, line 17: Should this be targeting the `productcatalogservice` binary?

jaronoff97 (commented):
A few thoughts/minor changes.

pkg/instrumentation/podmutator.go
TylerHelmuth: Absolutely.  Thats gonna be a problem for all the languages once we put them behind feature gates whether or not they are enabled/disabled by default.  I think the change I made in https://github.com/open-telemetry/opentelemetry-operator/pull/1629 to print out the feature gates in the operator's startup will help as well as this log.  But ya, the users is gonna have to go searching for the information on "why it didn't work".  Thats the downside of how the operator manages things right?  It can't give pod owners direct feedback.

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -307,6 +307,7 @@ spec:
                 - --enable-leader-election
                 - --zap-log-level=info
                 - --zap-time-encoding=rfc3339nano
+                - --feature-gates=+operator.autoinstrumentation.go
```
jaronoff97, line 310: Should we keep this disabled here initially given this is an alpha feature?

pkg/instrumentation/podmutator_test.go
```diff
@@ -32,13 +33,17 @@ func TestMutatePod(t *testing.T) {
 	mutator := NewMutator(logr.Discard(), k8sClient)
 	require.NotNil(t, mutator)
 
+	truee := true
```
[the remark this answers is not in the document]
TylerHelmuth: Implemented everywhere

pkg/instrumentation/sdk.go
TylerHelmuth: Added

pkg/instrumentation/sdk_test.go
```diff
@@ -805,6 +805,187 @@ func TestInjectDotNet(t *testing.T) {
...
+					Containers: []corev1.Container{
+						{
+							Name: "app",
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "OTEL_TARGET_EXE set by inst",
```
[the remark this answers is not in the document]
TylerHelmuth: Done

tests/e2e/instrumentation-go/01-install-app.yaml
TylerHelmuth: Good catch.  Since the e2e test doesn't actually check the output of the instrumentation inject this wasn't caught (or tested), but best to use a proper value anyways.  Updated.

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
TylerHelmuth: We definitely want Go instrumentation disabled by default.  I think this is getting changed because I had to change `config/default/manager_auth_proxy_patch.yaml` to ensure the feature flag was enabled for e2e tests.  Is there a way I can keep the Go e2e test and not mess with the manifest?  If the e2e test is using our default manifest we'll have to remove the Go e2e tests for now.

jaronoff97: I see... hmm @pavolloffay thoughts on what to do here?

pkg/instrumentation/podmutator.go
jaronoff97: yeah its a double edged sword... just calling it out as something we should be prepared for :)

tests/e2e/instrumentation-golang/01-install-app.yaml
damemi: If it helps we could probably publish an image for one of the sample apps

TylerHelmuth: This particular e2e tests only tests that the operator can add the sidecar, not that the sidecar produces meaningful telemetry.  I feel the demo image is sufficient and saves everyone extra work.

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
TylerHelmuth: Based on the SIG meeting it sounded like it is proper for the default manifest to expose all available features.  If that is still the sentiment keeping this flag is appropriate.

jaronoff97 (approved):

jaronoff97:
This looks great to me, thank you so much @TylerHelmuth

TylerHelmuth:
@pavolloffay please take a look for final review.

README.md
```diff
@@ -253,6 +253,15 @@ DotNet:
 instrumentation.opentelemetry.io/inject-dotnet: "true"
 ```
 
+Go:
+
+Go auto-instrumentation also honors an annotation that will be used to set the [OTEL_GO_AUTO_TARGET_EXE env var](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/docs/how-it-works.md).
```
pavolloffay, line 258: is this env var mandatory? 

What happens if it is not set? Is there any default value? I think we should document this in the CRD.

pavolloffay (commented):
Great work 🥳 

I would like to see a better-documented expectations for OTEL_GO_AUTO_TARGET_EXE. 

What is the behaviour if the annotation is not set? Does the injection go through anyways? If no shall we emit an event for user to let him know that the injection failed?

README.md
```diff
@@ -253,6 +253,17 @@ DotNet:
...
+This env var can also be set via the Instrumentation resource, with the annotation taking precedence.
+Since Go auot-instrumentation requires `OTEL_GO_AUTO_TARGET_EXE` to be set, you must supply a valid
```
pavolloffay, line 260: auto-instrumentation

pkg/instrumentation/golang.go
```diff
@@ -0,0 +1,94 @@
...
+		if idx == -1 {
+			goAgent.Env = append(goAgent.Env, env)
+		}
+	}
+
+	pod.Spec.Containers = append(pod.Spec.Containers, goAgent)
+	pod.Spec.Volumes = append(pod.Spec.Volumes, corev1.Volume{
+		Name: kernelDebugVolumeName,
+		VolumeSource: corev1.VolumeSource{
+			HostPath: &corev1.HostPathVolumeSource{
```
pavolloffay, line 88: Does this require any additional RBAC?

pavolloffay (commented):
LGTM just some minor comments

pkg/instrumentation/golang.go
TylerHelmuth: @pavolloffay No.  Its technically not even required, it is used for providing insight into what the ebpf probes are doing, but since the instrumentation is so immature its best practice to provide it right now.

README.md
TylerHelmuth: Done

pavolloffay (approved):

PR #1557, opened 2023-03-06
Add first feature flag, enable it for target allocator config addition

### LINKED ISSUE #1548: Allow for feature flags in the operator
There are a few features that have been discussed in the past weeks that would benefit from being behind a feature flag before they are considered released. Examples:

* #1547 
* #908 
* #1318 

We could use a similar feature flagging system the collector [uses](https://github.com/open-telemetry/opentelemetry-collector/tree/main/featuregate)

jaronoff97:
Closes #1548

This PR introduces the ability to manage feature flags for the operator. We use the collector's feature flag package to do this. This PR uses feature flagging for a highly requested feature – automatically rewrite the collector config map to contain the necessary configuration to get scrape config jobs from the target allocator.


jaronoff97:
Question @open-telemetry/operator-ta-maintainers should we only allow this feature when someone sets prometheusCR to be true?

pkg/collector/reconcile/config_replace.go
```diff
@@ -71,13 +87,16 @@ func ReplaceConfig(instance v1alpha1.OpenTelemetryCollector) (string, error) {
...
+		cfg.TargetAllocConfig = &targetAllocator{
+			Endpoint:    fmt.Sprintf("http://%s:80", naming.TAService(instance)),
+			Interval:    30 * time.Second,
+			CollectorID: "${POD_NAME}",
+		}
 	}
 
 	// type coercion checks are handled in the ConfigToPromConfig method above
-	config["receivers"].(map[interface{}]interface{})["prometheus"].(map[interface{}]interface{})["config"] = updPromCfgMap["PromConfig"]
+	config["receivers"].(map[interface{}]interface{})["prometheus"] = cfg
```
jaronoff97, line 99: This should massively simplify the code / understandability here... previously we were unnecessary mapstructure decoding of the datastructure, when everything already has yaml tags. This should make it easier to add fields here in the future. Tests and codepaths have been updated to reflect this change.

pkg/targetallocator/adapters/config_to_prom_config.go
```diff
@@ -55,15 +55,5 @@ func ConfigToPromConfig(cfg string) (map[interface{}]interface{}, error) {
...
-		return nil, errorNoComponent("prometheusConfig")
-	}
-
-	prometheusConfig, ok := prometheusConfigProperty.(map[interface{}]interface{})
-	if !ok {
-		return nil, errorNotAMap("prometheusConfig")
-	}
-
-	return prometheusConfig, nil
+	return prometheus, nil
```
jaronoff97, line 58: we want to return the full config specified, not just the prometheus `config` block, this would prevent us from propagating the `target_allocator` block.

kristinapathak (approved):

pkg/collector/reconcile/config_replace.go
```diff
@@ -71,13 +87,16 @@ func ReplaceConfig(instance v1alpha1.OpenTelemetryCollector) (string, error) {
 		}
 	}
 
-	updPromCfgMap := make(map[string]interface{})
-	if err := mapstructure.Decode(cfg, &updPromCfgMap); err != nil {
-		return "", err
+	if EnableTargetAllocatorRewrite.IsEnabled() {
```
matej-g, line 90: I wonder if we could correlate this with the flag enabling TA, to prevent misconfiguration? We could log a warning for example if we discover TA is not enabled but user tried to enable allocator rewrite.

matej-g (approved):
This looks good to me. It would be nice to update the TA readme with this new feature as well.

pkg/collector/reconcile/config_replace.go
jaronoff97: This will only work if the TA is enabled the top of the function makes a check
```
func ReplaceConfig(instance v1alpha1.OpenTelemetryCollector) (string, error) {
	if !instance.Spec.TargetAllocator.Enabled {
		return instance.Spec.Config, nil
	}
```

pkg/collector/reconcile/config_replace.go
```diff
@@ -71,13 +87,16 @@ func ReplaceConfig(instance v1alpha1.OpenTelemetryCollector) (string, error) {
...
+		cfg.TargetAllocConfig = &targetAllocator{
+			Endpoint:    fmt.Sprintf("http://%s:80", naming.TAService(instance)),
+			Interval:    30 * time.Second,
```
moh-osman3, line 93: If someone sets these values already in the Spec that is recovered from the call to `ConfigToPromConfig` will this overwrite those values?

pkg/collector/reconcile/config_replace.go
```diff
@@ -31,8 +33,24 @@ import (
...
+	// automatically be rewritten when the target allocator is enabled.
+	EnableTargetAllocatorRewrite = featuregate.GlobalRegistry().MustRegister(
+		"operator.enableTargetAllocatorRewrite",
+		featuregate.StageAlpha,
+		featuregate.WithRegisterDescription("controls whether the operator should configure the collector's targetAllocator configuration"))
+)
+
+type targetAllocator struct {
+	Endpoint    string        `yaml:"endpoint"`
+	Interval    time.Duration `yaml:"interval"`
```
moh-osman3, line 47: Curious what this interval is used for?

pkg/collector/reconcile/config_replace.go
```diff
@@ -71,13 +87,16 @@ func ReplaceConfig(instance v1alpha1.OpenTelemetryCollector) (string, error) {
 		}
 	}
 
-	updPromCfgMap := make(map[string]interface{})
-	if err := mapstructure.Decode(cfg, &updPromCfgMap); err != nil {
-		return "", err
+	if EnableTargetAllocatorRewrite.IsEnabled() {
+		cfg.TargetAllocConfig = &targetAllocator{
```
moh-osman3, line 91: General/basic question - Where will this new cfg field end up being used? Is there another file that will consume this `TargetAllocConfig` and do something with it?

moh-osman3 (commented):
Just had some basic questions about the goal of this change and how it will be used

pkg/collector/reconcile/config_replace.go
jaronoff97: This is the interval we query the target allocator at. So if you set it to 30s, every 30 seconds the collector will ask the target allocator for new scrape configs.

pkg/collector/reconcile/config_replace.go
jaronoff97: this is just used as a convenience for serialization, not really meant to be used beyond that. see [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/prometheusreceiver#opentelemetry-operator) for more details

pkg/collector/reconcile/config_replace.go
jaronoff97: yes, in this case we will overwrite whatever a user may have set.

moh-osman3: Oh so when the featuregate is disabled then the user settings is not overwritten? Or is the user not really supposed to set this part of the config?

jaronoff97: yeah when the featuregate is disabled, nothing will happen which means whatever settings are there will remain.

moh-osman3 (approved):
LGTM!

pkg/collector/reconcile/config_replace.go
Aneurysm9: How should this be handled when the feature gate becomes beta or stable and is enabled by default?  Should there be a log entry indicating the user-provided content is being replaced?

Aneurysm9 (approved):

pkg/collector/reconcile/config_replace.go
jaronoff97: yeah that seems reasonable, FWIW we already replace user content for this functionality above for the http_sd_config destination. That will be in a follow up when this is moved to beta

pkg/collector/reconcile/config_replace.go
matej-g: Yes, so if user _disables_ TA but _enables_ allocator rewrite, it will not have any effect but user won't know what's wrong. Not super vital, but nice to have.

pavolloffay:
The output of help command 

```
go run ./main.go --help                                                                                                                                                                                        ploffay@fedora
go: downloading go.opentelemetry.io/collector/featuregate v0.72.0
Usage of /tmp/go-build1421575205/b001/exe/main:
      --auto-instrumentation-dotnet-image string   The default OpenTelemetry DotNet instrumentation image. This image is used when no image is specified in the CustomResource. (default "ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-dotnet:0.0.0")
      --auto-instrumentation-java-image string     The default OpenTelemetry Java instrumentation image. This image is used when no image is specified in the CustomResource. (default "ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-java:0.0.0")
      --auto-instrumentation-nodejs-image string   The default OpenTelemetry NodeJS instrumentation image. This image is used when no image is specified in the CustomResource. (default "ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-nodejs:0.0.0")
      --auto-instrumentation-python-image string   The default OpenTelemetry Python instrumentation image. This image is used when no image is specified in the CustomResource. (default "ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-python:0.0.0")
      --collector-image string                     The default OpenTelemetry collector image. This image is used when no image is specified in the CustomResource. (default "ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector:0.0.0")
      --enable-leader-election                     Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.
      --feature-gates Flag                         Comma-delimited list of feature gate identifiers. Prefix with '-' to disable the feature. '+' or no prefix will enable the feature.
      --health-probe-addr string                   The address the probe endpoint binds to. (default ":8081")
      --kubeconfig string                          Paths to a kubeconfig. Only required if out-of-cluster.
      --labels stringArray                         Labels to filter away from propagating onto deploys
      --metrics-addr string                        The address the metric endpoint binds to. (default ":8080")
      --operator-opamp-bridge-image string         The default OpenTelemetry Operator OpAMP Bridge image. This image is used when no image is specified in the CustomResource. (default "ghcr.io/open-telemetry/opentelemetry-operator/operator-opamp-bridge:0.0.0")
      --target-allocator-image string              The default OpenTelemetry target allocator image. This image is used when no image is specified in the CustomResource. (default "ghcr.io/open-telemetry/opentelemetry-operator/target-allocator:0.0.0")
      --tls-cipher-suites strings                  Comma-separated list of cipher suites for the server. Values are from tls package constants (https://golang.org/pkg/crypto/tls/#pkg-constants). If omitted, the default Go cipher suites will be used
      --tls-min-version string                     Minimum TLS version supported. Value must match version names from https://golang.org/pkg/crypto/tls/#pkg-constants. (default "VersionTLS12")
      --webhook-port int                           The port the webhook endpoint binds to. (default 9443)
      --zap-devel                                  Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn). Production Mode defaults(encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error)
      --zap-encoder encoder                        Zap log encoding (one of 'json' or 'console')
      --zap-log-level level                        Zap Level to configure the verbosity of logging. Can be one of 'debug', 'info', 'error', or any integer value > 0 which corresponds to custom debug levels of increasing verbosity
      --zap-stacktrace-level level                 Zap Level at and above which stacktraces are captured (one of 'info', 'error', 'panic').
      --zap-time-encoding time-encoding            Zap time encoding (one of 'epoch', 'millis', 'nano', 'iso8601', 'rfc3339' or 'rfc3339nano'). Defaults to 'epoch'.
pflag: help requested
exit status 2
```

@jaronoff97 is there a way to print all supported feature flags?

main.go
```diff
@@ -24,6 +24,10 @@ import (
 	"strings"
 	"time"
 
+	"go.opentelemetry.io/collector/featuregate"
+
+	"github.com/open-telemetry/opentelemetry-operator/pkg/flags"
```
pavolloffay, line 29: please merge these imports with other groups.

pkg/flags/flags.go
```diff
@@ -0,0 +1,36 @@
...
+	"flag"
+
+	"go.opentelemetry.io/collector/featuregate"
+)
+
+const (
+	featureGatesFlag = "feature-gates"
+)
+
+func Flags() *flag.FlagSet {
```
pavolloffay, line 27: missing docs? 


What is the purpose of this package? Shall we move other operator flags from main here?

pkg/flags/flags.go
```diff
@@ -0,0 +1,36 @@
...
+	flagSet := new(flag.FlagSet)
+	flagSet.Var(featuregate.FlagValue{}, featureGatesFlag,
+		"Comma-delimited list of feature gate identifiers. Prefix with '-' to disable the feature. '+' or no prefix will enable the feature.")
+	return flagSet
+}
+
+func GetFeatureGatesFlag(flagSet *flag.FlagSet) featuregate.FlagValue {
```
pavolloffay, line 34: nit: godoc 

We should probably enforce this by linter.

pkg/flags/flags.go
jaronoff97: This is just for featureflags, i'll add some docs / rename this to show it's only for featureflags

jaronoff97:
Hey all, still interested in this PR however I'm on vacation currently and as a result it may take a bit longer to get to this.

jaronoff97:
This has been superseded by #1619 another PR should be opened for the target allocator changes.

PR #1571, opened 2023-03-10
set default resource limits for instrumentation init containers

### LINKED ISSUE #895: Support resources (Requests and Limits) for auto-instrumentation initContainers
An auto-instrumented Pod may fail to deploy or force a scale-up because no resource `requests` or `limits` are set on the injected `initContainer`, so the namespace defaults are used. These are generally grossly excessive for a trivial `initContainer`, and will force the pod to reserve more than it needs.

The `initContainer` injected by the `Instrumentation` resource by request of the `instrumentation.opentelemetry.io/inject-nodejs: true` pod annotation should default to some very low resource limits for CPU and memory. The `Instrumentation` CRD probably needs to make them configurable, since no doubt someone will find a problem with any sensible low default, somehow, somewhere.

If you get deployment delays like

```
  Warning  FailedScheduling  38s   default-scheduler   0/25 nodes are available: 22 Insufficient cpu, 3 node(s) had taint {CriticalAddonsOnly: true}, that the pod didn't tolerate.
```

or related errors, this is probably why.

I don't see a practical way to patch this in my deployments because the container spec is generated by the operator on the fly.

### LINKED ISSUE #1407: Instrumentation CRD's need to have resource settings - request/limit options
Hi Team,  we are facing an issue with auto instrumentation crd to copy the agent jar to application pod, when the pod comes up init container level resources are being taken from namespace defaults.. which is 8 cpu, 8GB memory .. causing resource issues, is there a way i can set up the limits in crd object. sample file applied.

```
kubectl apply -f - <<EOF
kind: Instrumentation
metadata:
  name: my-instrumentation
spec:
  exporter:
    endpoint: http://otel-collector:4317/
  propagators:
    - tracecontext
    - baggage
    - b3
  sampler:
    argument: `"0.25"`
  python:
    env:
      # Required if endpoint is set to 4317.
      # Python autoinstrumentation uses http/proto by default
      # so data must be sent to 4318 instead of 4137.
      - name: OTEL_EXPORTER_OTLP_ENDPOINT
        value: http://otel-collector:4318/
EOF
```

frzifus:
Since all initContainers only copy files that should not rly change in size, I think it is sufficient to define good default values.

Closes #1407
Closes #895


pkg/instrumentation/helper.go
```diff
@@ -14,7 +14,30 @@
...
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func defaultInitContainerResourceRequirements() corev1.ResourceRequirements {
+	const (
+		initContainerCPULimit = "1"
+		initContainerMEMLimit = "256Mi"
+		initContainerCPUReq   = "0.5"
+		initContainerMEMReq   = "128Mi"
+	)
```
frzifus, line 28: These values are a bit random. Are there any recommendations?

pkg/instrumentation/helper.go
```diff
@@ -14,7 +14,30 @@
...
-import corev1 "k8s.io/api/core/v1"
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func defaultInitContainerResourceRequirements() corev1.ResourceRequirements {
+	const (
+		initContainerCPULimit = "1"
+		initContainerMEMLimit = "256Mi"
```
pavolloffay, line 25: How big are the agents? I am not sure if any is bigger than 100mb.

frzifus: Iam assuming there that the agent size is almost equal to the image size.
```
autoinstrumentation-java     078769bc3266   4 days ago     34.1MB
autoinstrumentation-nodejs   1bcc3acff819   11 days ago    114MB
autoinstrumentation-python   4e612ec06244   11 days ago    19MB
autoinstrumentation-dotnet   7a0228b7eafc   11 days ago    76.4MB
```

pkg/instrumentation/helper.go
```diff
@@ -14,7 +14,30 @@
...
 package instrumentation
 
-import corev1 "k8s.io/api/core/v1"
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func defaultInitContainerResourceRequirements() corev1.ResourceRequirements {
+	const (
```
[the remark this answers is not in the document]
frzifus: it was intended to keep the values together in one place. But since they are  used exactly once, I have now removed them and inlined them instead.

I don't think it makes sense to provide the constants in the package scope, since they are not needed anywhere else.

pkg/instrumentation/helper.go
frzifus: Seems 1 is a bit to much :x: 
```
        +    message: '0/1 nodes are available: 1 Insufficient cpu.'
```

iblancasa (approved):

frzifus:
whats your thoughts on it? @open-telemetry/operator-approvers

.chloggen/default_limits_for_init_containers.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: set default resource limits for instrumentation init containers
```
pavolloffay, line 8: `Set default resource requirements for instrumentation init containers.`

pkg/instrumentation/helper.go
```diff
@@ -14,7 +14,23 @@
...
-import corev1 "k8s.io/api/core/v1"
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func defaultInitContainerResourceRequirements() corev1.ResourceRequirements {
+	return corev1.ResourceRequirements{
+		Limits: corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("500m"),
```
pavolloffay, line 25: I am curious why this is set to `500m`/0.5 CPU? Did you try lower numbers e.g. `100m`?

pavolloffay (commented):
LGTM just some nits.

pkg/instrumentation/helper.go
frzifus: y, even 50 works. Iam fine with 100 too. Its just the upper limit.

.chloggen/default_limits_for_init_containers.yaml
frzifus:
```suggestion
note: Set default resource requirements and limits for instrumentation init containers.
```

pavolloffay (approved):

svadlakonda575:
@frzifus @iblancasa @pavolloffay these changes looks but can we have a config option available on crd definition?
There could a issue with this where namespace container min requirements may be greater than the values provided in the defaults in which case init container will be blocked with error.
https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/cpu-constraint-namespace/

Error from server (Forbidden): error when creating "examples/admin/resource/cpu-constraints-pod-3.yaml":
pods "constraints-cpu-demo-3" is forbidden: minimum cpu usage per Container is 200m, but request is 100m.

svadlakonda575:
@frzifus setting up min default limits make sense. If we have an option to override them from instrumentation crd as memory limits that are imposed on namespace level could break things in which case we can override and mitigate the issue.

pavolloffay:
> . If we have an option to override them from instrumentation crd as memory limits that are imposed on namespace level could break things in which case we can override and mitigate the issue.

@svadlakonda575 do you have an example of that? Our assumption is that the operator will use the lowest possible resources for the init-containers. 

@frzifus did you try to set the resources as well to something really small? e.g. 10 Mib or 50m cpu?

jbrunodomingues:
@pavolloffay @svadlakonda575 Sorry jumping in on the thread but I do agree there is value in making it configurable. You can still have the default values hardcoded but then allow to override in the yaml.

For example where I'm working we have max and min limits setup at namespace level. Looking at the hardcoded values you have there it would be within our limits but different places might configure different limits. There might be some namespace that only runs heavy workloads and they might setup a minimum of 512Mi.

TylerHelmuth:
I would also vote for allowing this to be configurable.  I think the configuration could be at the root level of the InstrumentationResource or, if we want to be fancy, could also be set per-language.

If the value isn't supplied we would set the defaults that this PR uses.

svadlakonda575:
Yes I agree.. we need to have this configurable.. we recently set the default limits and deployed one of our namespaces where limitrange is set for container limits greater than the values given default, pod didn't come up with an error. We can have defaults but need to have an option to override the defaults if needed.

pkg/instrumentation/helper.go
```diff
@@ -14,7 +14,23 @@
...
+			corev1.ResourceMemory: resource.MustParse("128Mi"),
```
jaronoff97, line 26: Similar to the other comments in the PR... these should definitely be configurable on top of these defaults as part of the Instrumentation CRD or the annotations to set auto-instrumentation.

pkg/instrumentation/sdk_test.go
```diff
@@ -730,6 +733,7 @@ func TestInjectDotNet(t *testing.T) {
 						Name:      volumeName,
 						MountPath: "/otel-auto-instrumentation",
 					}},
+					Resources: defaultInitContainerResourceRequirements(),
```
jaronoff97, line 736: Should we also add an e2e test for these?

jaronoff97 (changes requested):

pkg/instrumentation/sdk_test.go
frzifus: sure :)

apis/v1alpha1/instrumentation_types.go
```diff
@@ -113,6 +113,10 @@ type Java struct {
 	// If the former var had been defined, then the other vars would be ignored.
 	// +optional
 	Env []corev1.EnvVar `json:"env,omitempty"`
+
+	// Resources describes the compute resource requirements.
+	// +optional
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
```
jaronoff97, line 119: Is there a reason to put these on the individual language structs rather than `InstrumentationSpec` Is the idea that different languages may have different resource requirements?

apis/v1alpha1/instrumentation_webhook.go
```diff
@@ -64,26 +65,86 @@ func (r *Instrumentation) Default() {
 			r.Spec.Java.Image = val
 		}
 	}
+	if r.Spec.Java.Resources.Limits == nil {
+		r.Spec.Java.Resources.Limits = corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("500m"),
+			corev1.ResourceMemory: resource.MustParse("64Mi"),
```
jaronoff97, line 71: How were these values chosen (same goes for the rest of the languages?)

apis/v1alpha1/instrumentation_types.go
frzifus: So what you want to see is another config on the InstrumentationSpec level that gets overwritten by the language specific one?

apis/v1alpha1/instrumentation_webhook.go
frzifus: They are based on the container image size.

apis/v1alpha1/instrumentation_types.go
TylerHelmuth: That would keep things consistent because we allow that pattern with environment variables.  Being able to set it at the root level would let some users set the resources in one spot for all languages, reducing their config.  

I recognize that there would be a conflict at the root level tho, and that it doesn't save that much config, so I am ok if we stick with language-only.  It could also always be added later in a future PR.

jaronoff97: +1 to what tyler is recommending here, would also be fine if we put that in a future PR :)

frzifus: > ...reducing their config.

I hope that the default values will ensure that in most cases no further settings are needed. ^^

Issue #1574, opened 2023-03-13
Helm tests failing with v0.71.0 operator


Allex1:
Hi all, 

Trying to upgrade the helm stack to use [operator v0.71.0](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/679) and tests are failing in a non obvious way. 

cloning the 0.71 release tag I get https://github.com/open-telemetry/opentelemetry-helm-charts/actions/runs/4396632296/jobs/7699210715#step:13:991

cloning main (as I saw https://github.com/open-telemetry/opentelemetry-operator/pull/1553 is not included in the release) I get https://github.com/open-telemetry/opentelemetry-helm-charts/actions/runs/4402557919/jobs/7710082951

I'm not exactly familiar with the operator tests so I could use some help. @moh-osman3 can you have a look ?

Thanks

moh-osman3:
I'm unsure what's going on here, strange this didn't fail for the operator tests in this repo.  From your second test (cloning main) it seems that a step is skipped (https://github.com/open-telemetry/opentelemetry-helm-charts/actions/runs/4402557919/jobs/7710082951#step:13:761). This test should be running this [chart](https://github.com/open-telemetry/opentelemetry-operator/blob/main/tests/e2e/autoscale/02-install.yaml) which should have the desired `behavior` fields. Once this is installed in step 2, it should be validated in [script in step 3 ](https://github.com/open-telemetry/opentelemetry-operator/blob/main/tests/e2e/autoscale/cmd/verify/wait-and-validate-metrics.go#L154). But since this `02-install.yaml` doesn't get installed this will cause a segfault instead.

It's strange that 0.71.0 test which doesn't include my recent changes, fail as well, it seems scaledown doesn't happen in time despite stabilization window being set to 15 seconds. I wonder if there's something obvious I'm missing here when I added to these e2e tests. (cc @iblancasa @pavolloffay)

iblancasa:
## When cloning the 0.71 release tag
1. For me, the error you're finding sounds like a timeout. It takes some time for the HPA to scale down the collector and, maybe, the value for `stabilizationWindowSeconds` is not small enough. Also, the tests are running in parallel... which is more resource-consuming and the HPA will take more time to scale down. Is it something you can reproduce always? And locally?

## When cloning `master`
2. I can see some errors related to the namespace creation/deletion. It is a problem related directly to KUTTL. Running sequentially can workaround the problem

[...]

PR #1576, opened 2023-03-14
[HPA/e2e] Add nil checks to behavior validation and reduce stabilization windows 

moh-osman3:
Some fixes to the e2e tests to help resolve helm test failure from https://github.com/open-telemetry/opentelemetry-operator/issues/1574

### DIFF
--- a/tests/e2e/autoscale/02-install.yaml
+++ b/tests/e2e/autoscale/02-install.yaml
@@ -11,13 +11,11 @@ spec:
   autoscaler:
     targetCPUUtilization: 50
     targetMemoryUtilization: 70
-    # Without this behavior the HPA will default to a scaledown stabilization
-    # window of 300 seconds. Tests should fail if this update is not successful.
     behavior:
       scaleDown:
-        stabilizationWindowSeconds: 15
+        stabilizationWindowSeconds: 1
       scaleUp:
-        stabilizationWindowSeconds: 25
+        stabilizationWindowSeconds: 1
   resources:
     limits:
       cpu: 500m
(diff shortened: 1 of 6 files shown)


tests/e2e/autoscale/02-install.yaml
```diff
@@ -11,13 +11,11 @@ spec:
...
     targetMemoryUtilization: 70
-    # Without this behavior the HPA will default to a scaledown stabilization
-    # window of 300 seconds. Tests should fail if this update is not successful.
     behavior:
       scaleDown:
-        stabilizationWindowSeconds: 15
+        stabilizationWindowSeconds: 1
       scaleUp:
-        stabilizationWindowSeconds: 25
+        stabilizationWindowSeconds: 1
```
pavolloffay, line 18: what is the motivation of setting both of these to 1s?

moh-osman3: No longer changing scaleUp, but I am trying to remove any downtime in scale down. The helm repo seems to be using `kubernetes v1.23` and for some reason the scaleDown option is being ignored (and the default of 300s is used).  I now also added a timeout to at least unblock the helm tests from running. Do you have any thoughts on best way forward for https://github.com/open-telemetry/opentelemetry-operator/issues/1574?

iblancasa: @moh-osman3 [they are using 1.25](https://github.com/open-telemetry/opentelemetry-helm-charts/actions/runs/4424164450/jobs/7757660240#step:3:117)

tests/e2e/autoscale/05-assert.yaml
```diff
@@ -1,3 +1,7 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 360
```
iblancasa, line 3: Please, don't increase the timeout

pavolloffay:
Replaces by #1582

PR #1625, opened 2023-04-04
Add metric export to nodejs auto instrumentation

svrnm:
This also updates all nodejs dependencies to their latest versions

### DIFF
--- a/autoinstrumentation/nodejs/package.json
+++ b/autoinstrumentation/nodejs/package.json
@@ -14,14 +14,16 @@
     "typescript": "^4.4.4"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


jaronoff97:
@svrnm could you include a changelog for this enhancement please?

TylerHelmuth (commented):
Is it possible to add the metrics exporter without bumping everything?  Or would it be appropriate to bump everything in a separate PR and then add the metrics exporter?

svrnm:
> Is it possible to add the metrics exporter without bumping everything? Or would it be appropriate to bump everything in a separate PR and then add the metrics exporter?

I can split it out into 2 PRs, let me do that!

PR #1626, opened 2023-04-04
Bump dependencies in Node.JS autoinstrumentation to latest

svrnm:
Replacement for #1625, this only bumps all versions in the nodejs auto instrumentation to latest.

PR #1628, opened 2023-04-04
Add metrics to nodejs

### LINKED ISSUE #1627: Add MetricExporter to Node.JS auto instrumentation
As of now the auto instrumentation for Node.JS only adds tracing, so metrics from auto instrumentation libraries or from API calls embedded in the code are not emitted.

svrnm:
Fixes #1627, replaces #1625: this PR does not update the versions in package.json

will keep this as a draft PR until #1626 is merged, since this otherwise leads to build errors.


TylerHelmuth (commented):
@svrnm it looks like something isn't quite right with how the metric reader is being provided: https://github.com/open-telemetry/opentelemetry-operator/actions/runs/4610787187/jobs/8149679155?pr=1628#step:9:325

svrnm:
> @svrnm it looks like something isn't quite right with how the metric reader is being provided: https://github.com/open-telemetry/opentelemetry-operator/actions/runs/4610787187/jobs/8149679155?pr=1628#step:9:325

Indeed I forgot to copy over one dependency when splitting the PR, looks like this is fixed now.

TylerHelmuth (approved):

jaronoff97 (approved):

Issue #1560, opened 2023-03-08
[HPA] support Pod custom metrics in autoscaler spec


moh-osman3:
From internal customers, there have been multiple requests for a way to scale collectors based on Otel collector metrics. This would require exposing custom metrics to provide a way for users to configure their desired metrics beyond the `resource metrics` already supported. An example use case would be when custom metrics are not necessarily correlated with resource metrics. (e.g. ensuring latency of app is below certain threshold as a service requirement).

PR #1561, opened 2023-03-08
[HPA] Add support for Pod custom metrics

moh-osman3:
[WIP] - not tested or ready for review

Resolves https://github.com/open-telemetry/opentelemetry-operator/issues/1560

PR #1651, opened 2023-04-13
[HPA] Add v1alpha1.MetricSpec to support for Pod custom metrics

moh-osman3:
Replaces https://github.com/open-telemetry/opentelemetry-operator/pull/1561

This PR adds a new object `v1alpha1.MetricSpec` which is a subset of the metric types in https://pkg.go.dev/k8s.io/api/autoscaling/v2#MetricSpec

Currently `v1alpha1.MetricSpec` only supports `PodsMetricsSource` but can support more metric types in the metric array in the future. For example supporting `ResourceMetricsSource` in the future would allow resources to be configured in the HPA's metric array and allow us to deprecate `targetCPUUtilization` and `targetMemoryUtilization`

Resolves https://github.com/open-telemetry/opentelemetry-operator/issues/1560


apis/v1alpha1/opentelemetrycollector_webhook.go
```diff
@@ -219,6 +220,27 @@ func (r *OpenTelemetryCollector) validateCRDSpec() error {
 		if r.Spec.Autoscaler != nil && r.Spec.Autoscaler.TargetMemoryUtilization != nil && (*r.Spec.Autoscaler.TargetMemoryUtilization < int32(1) || *r.Spec.Autoscaler.TargetMemoryUtilization > int32(99)) {
 			return fmt.Errorf("the OpenTelemetry Spec autoscale configuration is incorrect, targetMemoryUtilization should be greater than 0 and less than 100")
 		}
+
+		if r.Spec.Autoscaler != nil {
```
jaronoff97, line 224: thought: I'm wondering if we can clean this up a little bit by putting the autoscaler logic in its own method and put it in a single if condition and moving it to the bottom of the function i.e.
```
if r.Spec.Autoscaler != nil {
    return checkAutoscalerSpec(r.Spec.Autoscaler)
}
return nil
```

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,2 @@
 resources:
-- manager.yaml
+- manager.yaml
```
jaronoff97, line 2: we should keep the newline

pkg/collector/horizontalpodautoscaler.go
```diff
@@ -86,11 +86,17 @@ func HorizontalPodAutoscaler(cfg config.Config, logger logr.Logger, otelcol v1al
 			},
 		}
 
-		if otelcol.Spec.Autoscaler != nil && otelcol.Spec.Autoscaler.Behavior != nil {
+		if otelcol.Spec.Autoscaler.Behavior != nil {
```
jaronoff97, line 89: have we confirmed that Autoscaler isn't nil by this point?

pkg/collector/horizontalpodautoscaler.go
```diff
@@ -86,11 +86,17 @@ func HorizontalPodAutoscaler(cfg config.Config, logger logr.Logger, otelcol v1al
...
 			behavior := ConvertToV2beta2Behavior(*otelcol.Spec.Autoscaler.Behavior)
 			autoscaler.Spec.Behavior = &behavior
 		}
 
+		// check for custom metrics
+		if len(otelcol.Spec.Autoscaler.Metrics) > 0 {
+			metrics := ConvertToV2Beta2PodMetrics(otelcol.Spec.Autoscaler.Metrics)
```
jaronoff97, line 96: are we able to make this addition to confirm to the v2 spec to start with? Or given it's already using something similar to the v1alpha we want to keep it that way?

jaronoff97: hm maybe i'm not understanding what this should be doing. What's the type of `autoscaler.Spec.Metrics` and how does that diff from what you're adding in?

moh-osman3: `otelcol.Spec.Autoscaler.Metrics` has type `v1alpha1.MetricSpec` this supports only the `PodsMetricSource`

moh-osman3: The HPA is either expecting a `v2.MetricSpec` or a `v2beta2.MetricSpec` so this does the necessary conversion for the HPA object depending on the version. This is something similar done for `.Spec.Autoscaler.Behavior`

moh-osman3: The reason I created `v1alpha1.MetricSpec` instead of directly using a `v2.MetricSpec` is because `v2.MetricSpec` is way too large of an object for the use cases of our HPA (see https://pkg.go.dev/k8s.io/api/autoscaling/v2#MetricSpec). It supports 5 different metric types which would require a lot of validation and the deprecation of `targetCPUUtilization` and `targetMemoryUtilization`.

moh-osman3: `v1alpha1.MetricSpec` is much smaller and can be updated as more functionality is required. I'm totally open to suggestions to change the type of `otelcol.Spec.Autoscaler.Metrics` though if that's still unclear!

pkg/collector/horizontalpodautoscaler.go
moh-osman3: Hmm there are several references of using `otelcol.Spec.Autoscaler` without a nil check in this function. In fact this is the only place that nil checks. If it is unset, `otelcol.Spec.Autoscaler` is created by the defaulting webhook at https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1alpha1/opentelemetrycollector_webhook.go#L72

So it should never be nil, but now I'm wondering if there cases where the defaulting webhook is never called and this causes a segfault?

config/manager/kustomization.yaml
moh-osman3: Fixed!

pkg/collector/horizontalpodautoscaler.go
jaronoff97: thats a good q... maybe we add the guard at the start of the function?

pkg/collector/horizontalpodautoscaler.go
jaronoff97: nope, makes a ton of sense. thanks for explaining!

apis/v1alpha1/opentelemetrycollector_webhook.go
```diff
@@ -202,22 +203,12 @@ func (r *OpenTelemetryCollector) validateCRDSpec() error {
...
 
-		if r.Spec.Autoscaler != nil && r.Spec.Autoscaler.Behavior != nil {
-			if r.Spec.Autoscaler.Behavior.ScaleDown != nil && r.Spec.Autoscaler.Behavior.ScaleDown.StabilizationWindowSeconds != nil &&
-				*r.Spec.Autoscaler.Behavior.ScaleDown.StabilizationWindowSeconds < int32(1) {
-				return fmt.Errorf("the OpenTelemetry Spec autoscale configuration is incorrect, scaleDown should be one or more")
+		if r.Spec.Autoscaler != nil {
+			err := checkAutoscalerSpec(r.Spec.Autoscaler)
+			if err != nil {
+				return err
 			}
```
jaronoff97, line 210: If you move this check to the end of the function (after the ingress check), we can just return `checkAutoscalerSpec(r.Spec.Autoscaler)` which either returns nil or err

pkg/collector/horizontalpodautoscaler.go
moh-osman3: Added the nil check at the start and returning nil if it is unset!

apis/v1alpha1/opentelemetrycollector_webhook.go
moh-osman3: Using this suggestion now - thanks!

jaronoff97 (approved):

apis/v1alpha1/opentelemetrycollector_webhook.go
```diff
@@ -184,6 +185,12 @@ func (r *OpenTelemetryCollector) validateCRDSpec() error {
 		}
 	}
 
+	if r.Spec.Ingress.Type == IngressTypeNginx && r.Spec.Mode == ModeSidecar {
+		return fmt.Errorf("the OptenTelemetry Spec Ingress configuiration is incorrect. Ingress can only be used in combination with the modes: %s, %s, %s",
```
iblancasa, line 189:
```suggestion
		return fmt.Errorf("the OptenTelemetry Spec Ingress configuration is incorrect. Ingress can only be used in combination with the modes: %s, %s, %s",
```

iblancasa (approved):

apis/v1alpha1/opentelemetrycollector_webhook.go
moh-osman3: Thanks for the review! Fixed.

jaronoff97 (approved):

PR #1656, opened 2023-04-16
feat: Add lifecycle hooks and terminationGracePeriodSeconds support in collector spec

### LINKED ISSUE #1618: Add support for `lifecycle` hooks and `terminationGracePeriodSeconds`
I want to be able to add a `preStop` container lifecycle hook and modify the `terminationGracePeriodSeconds` from its default 30s for my collector pods.  Is this currently possible with the Collector spec? Are people open to me adding this support?

avadhut123pisal:
closes #1618


pkg/collector/container_test.go
```diff
@@ -458,3 +458,27 @@ service:
...
+		},
+	}
+	cfg := config.New()
+
+	// test
+	c := Container(cfg, logger, otelcol, true)
+
+	// verify
+	assert.Equal(t, []string{"sh", "sleep 100"}, c.Lifecycle.PostStart.Exec.Command)
+	assert.Equal(t, []string{"sh", "sleep 300"}, c.Lifecycle.PreStop.Exec.Command)
```
TylerHelmuth, line 483: Can we assert that `c`'s entire `Lifecycle` field is equal to an expected `Lifecycle` struct?  I think that ends up being a more precise check.

jaronoff97 (approved):
Responded to your comment, I'm not sure we _need_ to create a `Lifecycle` to test against, but if you'd feel more confident doing that, it shouldn't be a big lift.

TylerHelmuth (approved):

jaronoff97 (approved):

PR #1668, opened 2023-04-20
[chore] Ensure the autoscaling version is updated in the copies of autodetect

### LINKED ISSUE #1597: Improve the platform detection for HPA
While checking the problem described in #1418, we found the autodetection for HPA is not 100% correct.

The configuration is set as a value in the `autoDetect` struct (`internal/config/main.go`). So, if we copy an instance of the `Config`  struct (like passing it by value in a method) and update the copy (from a different Goroutine for instance), the HPA value will not be reflected in the copy. This problem was solved previously for other components from `Autodetect` (https://github.com/open-telemetry/opentelemetry-operator/pull/1206#issuecomment-1332316819).

To solve this problem, we can follow the same approach done for OpenShift Routes in #1463.

So, for instance, in the `(c *Config) AutoDetect` method, instead of:
```go
	hpaVersion, err := c.autoDetect.HPAVersion()
	if err != nil {
		return err
	}
	c.autoscalingVersion = hpaVersion
```
We should do the same as we do for the OpenShift Routes:
```go
	hpaV, err := c.autoDetect.HPAVersion()
	if err != nil {
		return err
	}
	if c.hpaVersion.Get() != hpaV {
		c.logger.V(1).Info("HPA version detected", "version", hpaV)
		c.hpaVersion.Set(hpaV)
		if err = c.onHPAVersion.Do(); err != nil {
			// Don't fail if the callback failed, as auto-detection itself worked.
			c.logger.Error(err, "configuration change notification failed for callback")
		}
	}
```

iblancasa:
Resolves #1597.


internal/config/main.go
```diff
@@ -197,6 +206,35 @@ func (c *Config) RegisterOpenShiftRoutesChangeCallback(f func() error) {
...
+		current: autodetect.AutoscalingVersionUnknown,
+	}
+}
+
+type hpaVersionWrapper struct {
+	mu      sync.Mutex
+	current autodetect.AutoscalingVersion
+}
+
+func (p *hpaVersionWrapper) Set(hpaV autodetect.AutoscalingVersion) {
```
jaronoff97, line 225: to confirm the change here is that we're passing this to copy the value and using a lock to ensure there's no contention?

internal/config/main.go
```diff
@@ -126,12 +129,18 @@ func (c *Config) AutoDetect() error {
...
 
-	hpaVersion, err := c.autoDetect.HPAVersion()
+	hpaV, err := c.autoDetect.HPAVersion()
 	if err != nil {
 		return err
 	}
-	c.autoscalingVersion = hpaVersion
-	c.logger.V(2).Info("autoscaling version detected", "autoscaling-version", c.autoscalingVersion.String())
+	if c.hpaVersion.Get() != hpaV {
+		c.logger.V(1).Info("HPA version detected", "version", hpaV)
```
jaronoff97, line 137: Why are we changing the log level here?

internal/config/main.go
```diff
@@ -61,9 +62,10 @@ func New(opts ...Option) Config {
 		targetAllocatorConfigMapEntry: defaultTargetAllocatorConfigMapEntry,
 		logger:                        logf.Log.WithName("config"),
 		openshiftRoutes:               newOpenShiftRoutesWrapper(),
+		hpaVersion:                    newHPAVersionWrapper(),
 		version:                       version.Get(),
-		autoscalingVersion:            autodetect.DefaultAutoscalingVersion,
 		onOpenShiftRoutesChange:       newOnChange(),
+		onHPAVersionChange:            newOnChange(),
```
jaronoff97, line 68: Do we need to register a callback for this function?

jaronoff97: or actually, do we need to call/use `WithOnHPAVersionChangeCallback` set up in `options.go`?

internal/config/main.go
iblancasa: I set the same value used for the OpenShift Routes:
```go
		c.logger.V(1).Info("openshift routes detected", "available", ora)
```

I can keep it, just revert it or revert it + change the value for the OpenShift Routes. What do you think?

jaronoff97:
One last thing before i merge @iblancasa Were you able to test this in a real cluster?

iblancasa:
> One last thing before i merge @iblancasa Were you able to test this in a real cluster?

Yes. I have been using the Operator built from this branch to deploy different stuff and I didn't notice any issues.
Thanks!

Issue #1682, opened 2023-04-27
Update nodejs dependencies to latest released version


mat-rumian:
Update OpenTelemetry JS dependencies to latest released versions:
- sdk 1.12.0 https://github.com/open-telemetry/opentelemetry-js/releases/tag/v1.12.0
- instrumentation 0.36.6 https://github.com/open-telemetry/opentelemetry-js-contrib/releases/tag/auto-instrumentations-node-v0.36.6

PR #1683, opened 2023-04-27
feat(nodejs): bump nodejs in version.txt

mat-rumian:
PR update nodejs autoinstrumentation in version.txt. 

Issue: https://github.com/open-telemetry/opentelemetry-operator/issues/1682

### DIFF
--- a/versions.txt
+++ b/versions.txt
@@ -19,7 +19,7 @@ autoinstrumentation-java=1.24.0
 
 # Represents the current release of NodeJS instrumentation.
 # Should match value in autoinstrumentation/nodejs/package.json
-autoinstrumentation-nodejs=0.37.0
+autoinstrumentation-nodejs=0.36.6
 
 # Represents the current release of Python instrumentation.
 # Should match value in autoinstrumentation/python/requirements.txt
--- a/versions.txt
+++ b/versions.txt
@@ -19,7 +19,7 @@ autoinstrumentation-java=1.24.0
 
 # Represents the current release of NodeJS instrumentation.
 # Should match value in autoinstrumentation/nodejs/package.json
-autoinstrumentation-nodejs=0.36.6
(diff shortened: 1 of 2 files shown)


mat-rumian:
I'm sorry needed to change version to `0.38.0` as it corresponds to `@opentelemetry/sdk-node`. I'm not sure if it's right. I was expecting this to match package `@opentelemetry/auto-instrumentations-node` which defines versions of the installed instrumentation packages.

versions.txt
```diff
@@ -19,7 +19,7 @@ autoinstrumentation-java=1.24.0
 
 # Represents the current release of NodeJS instrumentation.
 # Should match value in autoinstrumentation/nodejs/package.json
-autoinstrumentation-nodejs=0.37.0
+autoinstrumentation-nodejs=0.38.0
```
jaronoff97, line 22: Do we need to upgrade [this](https://github.com/open-telemetry/opentelemetry-operator/blob/main/autoinstrumentation/nodejs/package.json#L18) as well?

VineethReddy02 (changes requested):
@mat-rumian, I don't think the version should be updated in `versions.txt`. Normally, `versions.txt` is bumped during the release process to publish the latest auto-inst for supported languages. Instead, it should be updated in the `auto-instrumentation` directory.

Here's an example of how the last Node.js bump to the latest version was performed: [https://github.com/open-telemetry/opentelemetry-operator/pull/1626/files](https://github.com/open-telemetry/opentelemetry-operator/pull/1626/files)

Also, [here](https://github.com/open-telemetry/opentelemetry-operator/pull/1640/files) is an example of how python versions are bumped.

ankit-gautam23 (approved):
LGTM

TylerHelmuth:
This PR is no longer needed since we've upgraded past 0.38.0

PR #1619, opened 2023-03-30
Add ability to use feature-gates in the operator

TylerHelmuth:
This PR adds the ability to use feature gates in the operator, but does not add any feature gates specifically.  It is based on https://github.com/open-telemetry/opentelemetry-operator/pull/1557, but with the most recent `featuregate` changes.  My goal is to add this capability as a standalone PR so it can be used in  https://github.com/open-telemetry/opentelemetry-operator/pull/1557 and https://github.com/open-telemetry/opentelemetry-operator/pull/1555.

Closes https://github.com/open-telemetry/opentelemetry-operator/issues/1548

PR #1688, opened 2023-04-28
Add feature flag for target allocator config addition

### LINKED ISSUE #1581: Automatically add target_allocator section to prometheus receiver config if allocator is enabled
Since 0.62.0, [prometheus receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/README.md#opentelemetry-operator) allows us to directly query the target allocator for job definitions like so:

```yaml
receivers:
  prometheus:
    target_allocator:
      endpoint: http://my-targetallocator-service
      interval: 30s
      collector_id: collector-1
```

We should automatically add this configuration section to the prometheus receiver config if the allocator is enabled for a given Collector object. 

Right now we only add a `http_sd_config` with the allocator url to each [scrape job definition](https://github.com/open-telemetry/opentelemetry-operator/blob/eabe95a0abe6ae206987b51897983504f2edfb9f/pkg/collector/reconcile/config_replace.go#L65), which I think is unnecessary now that the receiver has direct support for the allocator.

swiatekm-sumo:
These are changes from #1557, rebased on main and using the feature flag machinery from #1619.

One addition I've made is clearing the scrape configs in prometheus receiver itself. This doesn't actually change the semantics, but it makes it clearer what is actually going on. The change is in a separate commit here: https://github.com/open-telemetry/opentelemetry-operator/commit/15ac0ce91e3e34144a3609f76cd0e3cb9246405d.

I've done some manual E2E testing on these changes and everything worked as expected.

Fixes #1581

### DIFF
--- a/pkg/collector/reconcile/config_replace.go
+++ b/pkg/collector/reconcile/config_replace.go
@@ -17,8 +17,8 @@ package reconcile
 import (
 	"fmt"
 	"net/url"
+	"time"
 
-	"github.com/mitchellh/mapstructure"
 	promconfig "github.com/prometheus/prometheus/config"
 	"github.com/prometheus/prometheus/discovery"
 	"github.com/prometheus/prometheus/discovery/http"
@@ -27,12 +27,20 @@ import (
 
 	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
 	"github.com/open-telemetry/opentelemetry-operator/pkg/collector/adapters"
+	"github.com/open-telemetry/opentelemetry-operator/pkg/featuregate"
 	"github.com/open-telemetry/opentelemetry-operator/pkg/naming"
 	ta "github.com/open-telemetry/opentelemetry-operator/pkg/targetallocator/adapters"
 )
 
+type targetAllocator struct {
+	Endpoint    string        `yaml:"endpoint"`
+	Interval    time.Duration `yaml:"interval"`
+	CollectorID string        `yaml:"collector_id"`
+}
+
 type Config struct {
-	PromConfig *promconfig.Config `yaml:"config"`
+	PromConfig        *promconfig.Config `yaml:"config"`
+	TargetAllocConfig *targetAllocator   `yaml:"target_allocator,omitempty"`
 }
 
 func ReplaceConfig(instance v1alpha1.OpenTelemetryCollector) (string, error) {
@@ -50,9 +58,7 @@ func ReplaceConfig(instance v1alpha1.OpenTelemetryCollector) (string, error) {
 	}
 
 	// yaml marshaling/unsmarshaling is preferred because of the problems associated with the conversion of map to a struct using mapstructure
-	promCfg, marshalErr := yaml.Marshal(map[string]interface{}{
-		"config": promCfgMap,
-	})
+	promCfg, marshalErr := yaml.Marshal(promCfgMap)
 	if marshalErr != nil {
 		return "", marshalErr
 	}
@@ -71,13 +77,16 @@ func ReplaceConfig(instance v1alpha1.OpenTelemetryCollector) (string, error) {
 		}
 	}
 
-	updPromCfgMap := make(map[string]interface{})
-	if err := mapstructure.Decode(cfg, &updPromCfgMap); err != nil {
-		return "", err
+	if featuregate.EnableTargetAllocatorRewrite.IsEnabled() {
+		cfg.TargetAllocConfig = &targetAllocator{
+			Endpoint:    fmt.Sprintf("http://%s:80", naming.TAService(instance)),
+			Interval:    30 * time.Second,
+			CollectorID: "${POD_NAME}",
+		}
 	}
 
 	// type coercion checks are handled in the ConfigToPromConfig method above
-	config["receivers"].(map[interface{}]interface{})["prometheus"].(map[interface{}]interface{})["config"] = updPromCfgMap["PromConfig"]
+	config["receivers"].(map[interface{}]interface{})["prometheus"] = cfg
 
 	out, err := yaml.Marshal(config)
 	if err != nil {
--- a/pkg/collector/reconcile/config_replace.go
+++ b/pkg/collector/reconcile/config_replace.go
@@ -83,6 +83,8 @@ func ReplaceConfig(instance v1alpha1.OpenTelemetryCollector) (string, error) {
 			Interval:    30 * time.Second,
 			CollectorID: "${POD_NAME}",
 		}
+		// we don't need the scrape configs here anymore with target allocator enabled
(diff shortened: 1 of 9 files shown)


pkg/featuregate/featuregate.go
```diff
@@ -29,6 +29,13 @@ var (
 		"operator.autoinstrumentation.dotnet",
 		featuregate.StageBeta,
 		featuregate.WithRegisterDescription("controls whether the operator supports .NET auto-instrumentation"))
+
+	// EnableTargetAllocatorRewrite is the feature gate that controls whether the collector's configuration should
+	// automatically be rewritten when the target allocator is enabled.
+	EnableTargetAllocatorRewrite = featuregate.GlobalRegistry().MustRegister(
+		"operator.enableTargetAllocatorRewrite",
```
TylerHelmuth, line 36: Sticking with the hierarchical pattern, how do you feel about `operator.targetallocator.rewrite`?

swiatekm-sumo: That makes it sound like this changes target allocator behaviour, but in reality we're changing how the operator rewrites prometheusreceiver config. So maybe something like `operator.prometheusreceiver.rewritetargetallocator`? Kind of a mouthful, but more descriptive imo.

TylerHelmuth: Since the flag ultimately affects the collector how about `operator.collector.rewritetargetallocator`.  I think that would scale well with any other flags that mess with collector configuration/capabilities.

swiatekm-sumo: Yeah, that sounds better.

swiatekm-sumo: Renamed.

jaronoff97 (changes requested):

pkg/targetallocator/adapters/config_to_prom_config.go
```diff
@@ -28,7 +28,7 @@ func errorNotAMap(component string) error {
 	return fmt.Errorf("%s property in the configuration doesn't contain valid %s", component, component)
 }
 
-// ConfigToPromConfig converts the incoming configuration object into a the Prometheus receiver config.
+// ConfigToPromConfig converts the incoming configuration object into the Prometheus receiver config.
 func ConfigToPromConfig(cfg string) (map[interface{}]interface{}, error) {
```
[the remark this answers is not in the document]
swiatekm-sumo: I think I'd rather properly fix this check, as it's currently incorrect. Right now, it requires that the `config` property exists, but in reality we should check if one of the following is true:
* `config` exists
* `target_allocator` exists and target allocator is enabled on the resource
* target allocator is enabled on the resource and the feature flag is enabled

Then this is really a fix, not a breaking change, and the actual breaking change will happen when the flag is enabled by default. How's that sound?

pkg/collector/reconcile/configmap_test.go
```diff
@@ -184,6 +187,55 @@ service:
...
+  logging: null
+processors: null
+receivers:
+  prometheus:
+    config:
+      global:
+        scrape_interval: 1m
+        scrape_timeout: 10s
+        evaluation_interval: 1m
+    target_allocator:
```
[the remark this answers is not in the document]
swiatekm-sumo: It's disabled in previous tests, but only implicitly. Think it's worth disabling it explicitly there?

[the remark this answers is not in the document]
jaronoff97: hmmm... I think it's alright then, when the flag is moved to beta we'll just do the reverse.

pkg/collector/reconcile/config_replace.go
```diff
@@ -71,13 +77,18 @@ func ReplaceConfig(instance v1alpha1.OpenTelemetryCollector) (string, error) {
...
-	if err := mapstructure.Decode(cfg, &updPromCfgMap); err != nil {
-		return "", err
+	if featuregate.EnableTargetAllocatorRewrite.IsEnabled() {
+		cfg.TargetAllocConfig = &targetAllocator{
+			Endpoint:    fmt.Sprintf("http://%s:80", naming.TAService(instance)),
+			Interval:    30 * time.Second,
+			CollectorID: "${POD_NAME}",
+		}
+		// we don't need the scrape configs here anymore with target allocator enabled
+		cfg.PromConfig.ScrapeConfigs = []*promconfig.ScrapeConfig{}
```
jaronoff97, line 87: I actually think the collector may fail to startup if it doesn't have any scrape configs set, at least that was the case a few months ago...

pkg/targetallocator/adapters/config_to_prom_config.go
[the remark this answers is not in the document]
jaronoff97: yes, technically the comment on the function doesn't match what the code is actually doing, but regardless anyone potentially using this function and expecting it to return the prometheus config blob would be broken by this change. I think i'm okay to call this a 'fix', as long as we call that out in the changelog's subtext – in case someone is using this for some reason they should be able to see the difference in the release notes.

pkg/collector/reconcile/config_replace.go
jaronoff97: oh ha, nvm i fixed this a few months ago ([link](https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/21f36f87fc1d2472adbb0a663f3bf6517701ff5a#diff-c335f316da4cfd1d6feec64b950de5600d871572b350b1fadfe411c54343e073R161-R162))

jaronoff97 (commented):
One more Q: were you able to test this with a real cluster and CRD to be sure it works as expected E2E?

pkg/featuregate/featuregate.go
```diff
@@ -29,6 +29,13 @@ var (
 		"operator.autoinstrumentation.dotnet",
 		featuregate.StageBeta,
 		featuregate.WithRegisterDescription("controls whether the operator supports .NET auto-instrumentation"))
+
+	// EnableTargetAllocatorRewrite is the feature gate that controls whether the collector's configuration should
+	// automatically be rewritten when the target allocator is enabled.
+	EnableTargetAllocatorRewrite = featuregate.GlobalRegistry().MustRegister(
```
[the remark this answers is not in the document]
swiatekm-sumo: Done.

swiatekm-sumo:
> One more Q: were you able to test this with a real cluster and CRD to be sure it works as expected E2E?

Yeah. For reference, I'm adding experimental support for replacing Prometheus with otel to our Helm Chart here: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2988, and I ran our integration tests against the changes in this PR.

jaronoff97 (approved):

Issue #1600, opened 2023-03-24
BusyBox Licensing


bryan-aguilar:
The autoinstrumentation images distributed from this repository use BusyBox as a base image. BusyBox is GPLv2 licensed. See BusyBox licensing docs and enforcement [here](https://www.busybox.net/license.html). I believe the CNCF also requires additional approval for distribution of GPLv2 assets. See CNCF licensing docs [here](https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md#approved-licenses-for-allowlist). 

Our team is working on distributing an image very similar to the operator `java-autoinstrumentation` image. The only difference being that we package our ADOT Java Agent instead of the OpenTelemetry Java Agent. Instead of depending on BusyBox we are building and packaging a rust based binary that provides the needed functionality of the `cp` command. Our team would be happy to contribute this back to the operator community if that is agreed upon. Work on the Rust `cp` tool can be seen [here](https://github.com/aws-observability/aws-otel-java-instrumentation/pull/353).

matej-g:
I'd wonder if this use is even affected by the allow list policy and whether there should be concerns over GPL affecting licensing of collector. I know of at least one other CNCF project (Prometheus - https://github.com/prometheus/busybox) using BusyBox image and haven't heard similar concerns.

bryan-aguilar:
> Thanks for raising the issue and pointing to the CNCF licensing requirements.
> 
> +1 on changing the busybox images. Before going with the AWS library, do you know if there are other alternatives we could consider?

We had weighed the decision between a `cp-utility` tool written in Go or Rust. We came to the conclusion that rust would be a safer choice due to the amount of CVEs that each language has historically had. Recently we have also discovered [ToyBox](https://github.com/landley/toybox) as an option but did not evaluate how it could be integrated into the auto-instrumentation images. 

> I'd wonder if this use is even affected by the allow list policy and whether there should be concerns over GPL affecting licensing of collector. I know of at least one other CNCF project (Prometheus - https://github.com/prometheus/busybox) using BusyBox image and haven't heard similar concerns.

I think it's a fair question but is not something we should gloss over. Prometheus library may also not be compliant. IANAL so I think if we want to consider this line of questioning an operator maintainer should bring this to the GC and/or the CNCF.

[...]

PR #1698, opened 2023-05-01
[proposal] Use rust based `cp` tool rather than `busybox` in autoinstrumentation images. 

bryan-aguilar:
Issue: #1600 

Apologies for this taking so long. As mentioned in the issue we would like to propose a replacement to the busybox base image used in the auto-instrumentation images. This PR is a demonstration on what that implementation would look like for a single auto-instrumentation images. If this proposal is accepted followup PRs would be made to adjust the remaining dockerfiles and github workflows. 

The Rust based CP tool is statically linked to MUSL, which is licensed under MIT license. This implementation handles all the cases currently required by the Operator auto-instrumentation images: regular file copy (used in the Java) and archive copy (used in Python and NodeJS).

Please feel free to ask any follow up questions. I will ensure this is brought up in the next sig meeting.


bryan-aguilar:
I have added @rapphil as a codeowner also. He was the primary author of the rust `cp-utility` in our distribution and has agreed to assist in ongoing maintenance.

.github/CODEOWNERS
```diff
@@ -20,3 +20,6 @@
 
 # Target Allocator owners
 cmd/otel-allocator @open-telemetry/operator-ta-maintainers
+
+# cp-utility tool
+tools/cp-utility @rapphil @bryan-aguilar
```
jaronoff97, line 25: we'll have to make another group in order to use CODEOWNERS here

jaronoff97 (commented):
Excited to hear more at the SIG meeting about this. Thanks for putting it together!

autoinstrumentation/java/Dockerfile
```diff
@@ -4,10 +4,43 @@
 #  - Grant the necessary access to the jar. `chmod -R go+r /javaagent.jar`
 #  - For auto-instrumentation by container injection, the Linux command cp is
 #    used and must be availabe in the image.
-FROM busybox
+
+# Stage 1: Build the cp-utility binary
+FROM rust:1.69 as builder
```
frzifus, line 9: can the rust cp impl. be built for all target platforms supported by go? What I have in mind are ibm z + power (that's what we have to support).

rapphil: yes, s390x and ppc64le are supported by the rust base image and musl.

Are you using the autoinstrumentation images that are published from this repo in those platforms? I'm asking because today they only support arm and x86 (please take a look [here](https://github.com/open-telemetry/opentelemetry-operator/blob/main/.github/workflows/publish-autoinstrumentation-java.yaml#L69))

Issue #1623, opened 2023-04-04
Creating Collector pod with labelmap action fails when TargetAllocator is enabled


anunarapureddy:
Describe the bug
Failed to create collector pod with the below error log:

```
Error: failed to get config: cannot unmarshal the configuration: 1 error(s) decoding:

* error decoding 'receivers': error reading receivers configuration for "prometheus": prometheus receiver failed to
 unmarshal yaml to prometheus config: "" is invalid 'replacement' for labelmap `action`
```

This is happening when loading a valid prometheus configuration into the opentelemetrycollector resource.

Input prometheus configuration:

```
relabel_configs:
- regex: (.*id)
  action: labelmap
```

The operator creates a configmap using the input prometheus configuration where the labelmap action is transformed into the below configuration:

```
relabel_configs:
- separator: ;
  regex: (.*id)
  replacement: $1
  action: labelmap
```

At a glance,`opentelemetry-operator`'s yaml marshal/unmarshall logic seems to be adding two extra fields "replacement" and "separator" while building the configmap. The input config is being rewritten using `DefaultRelabelConfig` which seems to be appending extra defaults.  The `DefaultRelabelConfig` lives in [prometheus codebase](https://github.com/prometheus/prometheus/blob/3923e83413075a17c5a10739696a879765a05542/model/relabel/relabel.go#L103) and is being referred by the [operator code](https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/collector/reconcile/config_replace.go#L51-L62) to build the configmap only when `TargetAllocator` is enabled.

<img width="454" alt="Screen Shot 2023-04-03 at 8 43 07 PM" src="https://user-images.githubusercontent.com/23454368/229681742-7d5ee713-edc9-465a-9472-70665bc78d3a.png">

Steps to reproduce:

1. Create an opentelemetrycollector in statefulset mode with targetAllocator enabled.
2. Configure the prometheus receiver by providing a valid scrape configuration which includes `labelmap` action under `relabel_configs` section.
3. Wait for the creation of the collector pod (fails with error which results in the pod going into a CrashLoopBackoff)
4. Describe the configmap generated by the operator (labelmap with two extra fields appended)

What did you expect to see?

I expected to see the opentelemetry-operator create a collector pod successfully.

What did you see instead?
Pod creation fails with the below error which results in the pod going into a CrashLoopBackoff

```
[...]

Issue #1622, opened 2023-04-03
Prometheus configuration incompatibility between collector with TA and NO TA


murubhas:
Below configuration in `deployment` mode with TA disabled working good. However, when I change the deployment mode to `statefulset` with `TA enabled` it fails with below errors

**TA Error**
```
{"level":"error","ts":1680547146.8650134,"logger":"setup","msg":"Unable to load configuration","error":"error unmarshaling YAML: \"__param_$$1\" is invalid 'replacement' for labelmap action","stacktrace":"main.main\n\t/app/main.go:74\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:250"}
{"level":"info","ts":1680547146.8650668,"msg":"Starting the Target Allocator"}
{"level":"info","ts":1680547146.865076,"logger":"allocator","msg":"Unrecognized filter strategy; filtering disabled"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x88 pc=0x24c429f]

goroutine 64 [running]:
github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/target.(*Discoverer).ApplyConfig(0xc0003385a0, 0x0?, 0x0)
	/app/target/discovery.go:75 +0x15f
main.main.func7()
	/app/main.go:151 +0x55
github.com/oklog/run.(*Group).Run.func1({0xc000338960?, 0xc000926180?})
	/go/pkg/mod/github.com/oklog/run@v1.1.0/group.go:38 +0x2f
created by github.com/oklog/run.(*Group).Run
	/go/pkg/mod/github.com/oklog/run@v1.1.0/group.go:37 +0x22a
```
**Collector Error**
```
2023/04/03 18:41:58 ADOT Collector version: v0.26.1
2023/04/03 18:41:58 found no extra config, skip it, err: open /opt/aws/aws-otel-collector/etc/extracfg.txt: no such file or directory
Error: invalid configuration: no receiver configuration specified in config
2023/04/03 18:41:58 application run finished with error: invalid configuration: no receiver configuration specified in config
```
**Config Deployment**

```Yaml
---
apiVersion: opentelemetry.io/v1alpha1
kind: OpenTelemetryCollector
metadata:
  name: my-metric
  namespace: adot
spec:
  mode: deployment
  serviceAccount: default
  targetAllocator:
    enabled: false
    serviceAccount: default
    filterStrategy: relabel-config
[...]

Issue #958, opened 2022-05-31
Error creating a collector pod when specifying prometheus labeldrop actions


jacobwalters:
**Describe the bug**
I am seeing this error 

```2022/05/31 17:54:37 application run finished with error: failed to get config: cannot unmarshal the configuration: error reading receivers configuration for "prometheus": prometheus receiver failed to unmarshal yaml to prometheus config: labeldrop action requires only 'regex', and no other fields```

This is happening when I am loading valid prometheus configurations into the opentelemetrycollector resource.

I am specifying the label drop actions like this.

```
metric_relabel_configs:
- regex: (.*id)
  action: labeldrop
```

The operator is creating a configmap called adot-collector where the labeldrop actions are transformed into this

```
metric_relabel_configs:
- separator: ;
  regex: (.*id)
  replacement: $1
  action: labeldrop
```

Compared to how I defined this, the operator is adding the fields "replacement", and "separator". These fields the operator is adding is causing the error in the logs above.

I was able to fix this issue by removing the label drop commands from my config and redeploying.

**Steps to reproduce**

Create a opentelemetrycollector object in statefulset mode with the targetAllocator enabled. configure the prometheus receiver to scrape a prometheus endpoint and use the metric_relabel_config field to drop a label. Observe the creation of the collector pod and the configmap generated by the operator. 

**What did you expect to see?**

I expected to see the opentelemetry-operator crate a collector pod successfully.
**What did you see instead?**
I saw that the opentelemetry-operator created the pod Adot-collecotr-0. It was in a crashback loop. I looked at the logs and saw this message
```
2022/05/31 17:54:37 application run finished with error: failed to get config: cannot unmarshal the configuration: error reading receivers configuration for "prometheus": prometheus receiver failed to unmarshal yaml to prometheus config: labeldrop action requires only 'regex', and no other fields
```
When i looked at eh configmap created by the operator. The labeldrop action had additional keys ("replacement", "seperator") that I had not added.

**What version did you use?**

Version: (e.g., `v0.4.0`, `1eb551b`, etc)
the opentelemetry helm charts. opentelemetry-collector-0.17.0

**What config did you use?**

```
---
apiVersion: opentelemetry.io/v1alpha1
kind: OpenTelemetryCollector
metadata:
  name: adot
  namespace: opentelemetry
  labels:
    app: opentelemetry-operator
spec:
  image: "aws-otel-collector:0.17.0"
  mode: statefulset
[...]

PR #1712, opened 2023-05-05
fix issues related to prometheus relabel configs when target allocator is enabled

anunarapureddy:
Related to following issues: #[1623](https://github.com/open-telemetry/opentelemetry-operator/issues/1623), #[1622](https://github.com/open-telemetry/opentelemetry-operator/issues/1622), #[958](https://github.com/open-telemetry/opentelemetry-operator/issues/958)

### Cause:

The issue manifests itself in the following scenarios.

1. First, when the configMap is built by the operator, the unmarshalling logic calls the Prometheus relabel config validation logic ([relabel.go](https://github.com/prometheus/prometheus/blob/3923e83413075a17c5a10739696a879765a05542/model/relabel/relabel.go#L103)), which includes the additional defaults, including the $1 replacement field, appended to the original Prometheus configuration. Upon initialization, the collector pod loads the configMap and interprets $1 as an environment variable, leading to its replacement with an empty string. As a result, the same Prometheus relabel config validation logic fails when the config is unmarshalled in the collector code, as the replacement key is now replaced with an empty string due to env var substitution. 

2. The second scenario occurs when the Prometheus configuration explicitly includes $$ for labelmap, labeldrop, labelkeep or other actions that undergo regex validation in the Prometheus validation code ([relabel.go](https://github.com/prometheus/prometheus/blob/3923e83413075a17c5a10739696a879765a05542/model/relabel/relabel.go#L126)). In this case, when the operator creates the collector configMap, the regex validation fails, causing the configMap to get created with an empty yaml file.

3. In the third scenario, when Prometheus configuration explicitly includes $$ for labelmap, labeldrop, labelkeep or other actions, the TA ConfigMap is created with the $$ signs. Since the TA does not support environment variable substitution, the unmarshalling logic during the ConfigMap mount process in the TA pod will perform regex validation and cause the TA pod creation to fail.

### Resolution:

Below is the guideline from OpenTelemetry documentation:

> Note: Since the collector configuration supports env variable substitution $ characters in your prometheus configuration are interpreted as environment variables. If you want to use $ characters in your prometheus configuration, you must escape them using $$.

Key points considered in order to address the issue:

- To use "$" characters in Prometheus configuration, they must be escaped using "$$".
- The collector configuration supports environment variable substitution, but the TA does not.
- The ConfigMap for the collector pod should have "$$" to prevent env var substitution, unless the user intentionally wants it to happen.
- The TA ConfigMap should have a single "$", as it does not support env var substitution.

The operator was updated to modify the YAML file directly instead of going through the Prometheus unmarshalling logic to update the SD config. The TA ConfigMap creation process was modified to replace "$$" signs with a "$" to address the third scenario. These changes ensure that both the Collector and TA pods are created correctly, even when the Prometheus configuration includes labelmap, labeldrop, labelkeep, or other actions that undergo regex validation.

cc: @alolita @jaronoff97


anunarapureddy:
@open-telemetry/operator-maintainers As a first-time contributor, I need approval from a maintainer to run workflows. Would someone be able to assist me with obtaining approval?

kristinapathak:
I've looked and thought about this but mostly have follow up questions.

For the issue with including `$1` - my understanding is this is an issue due to the TA being enabled and adding back `$1`'s which are [not supported](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/prometheusreceiver#getting-started) in the collector configuration:

>Note: Since the collector configuration supports env variable substitution $ characters in your prometheus configuration are interpreted as environment variables. If you want to use $ characters in your prometheus configuration, you must escape them using $$.

While I wish they could be supported, I don't think we should be using the operator to make `$` variables work (which I believe this PR is doing).  My recommendation is instead to modify [`DefaultRelabelConfig`](https://github.com/prometheus/prometheus/blob/3923e83413075a17c5a10739696a879765a05542/model/relabel/relabel.go#L31) to no longer set `Replacement`.  This is not ideal, but because it's a public var, we can set it in the operator code to our desired default config and the prometheus code will use the new value.  Then `$1`'s will no longer be added back into the scrape config when the TA is enabled.

The second issue I see being addressed in this PR is prometheus not validating regex with `$$`.  Is this only with TA enabled or regardless of TA enablement?  And is this due to the code change of `$`->`$$`? I can't find anything in the prometheus code that explains why this wouldn't work.

I looked through the code in the prometheus receiver to find somewhere that `$$` was checked or treated specially but couldn't find anything, so my assumption is that `$$` is considered valid in the scrape config, just doesn't pass the regex checks for some reason.  In other words, the same scrape config with `$$` passed in the operator which causes a problem could be passed in a collector config directly with no problem.  Is that true?  I plan to look into this more but wanted to be sure to provide timely feedback.

My general takeaway is that there are multiple issues here, and I would rather detangle and solve them separately, if possible.  Is there any other issue being addressed here that I've missed?

anunarapureddy:
@open-telemetry/operator-maintainers As a first-time contributor, I need approval from a maintainer to run workflows. Would someone be able to assist me with obtaining approval?

anunarapureddy:
@kristinapathak, I have made changes to the pull request and also updated the description. Can you please review the latest code and provide feedback so that we can iterate on the changes? I was brainstorming with @jaronoff97 and came up with an alternative approach, which is reflected in the updated code.

anunarapureddy:
@kristinapathak, I have made changes to the pull request and also updated the description. Can you please review the latest code and provide feedback so that we can iterate on the changes?

pkg/collector/testdata/relabel_config_expected_with_sd_config.yaml
```diff
@@ -0,0 +1,57 @@
...
+          - label4
+        metrics_path: /metrics
+        relabel_configs:
+        - action: keep
+          regex: (.*)
+          source_labels:
+          - label1
+        - action: replace
+          regex: (.*)
+          replacement: $$1_$2
```
jaronoff97, line 35: should the `$2` be escaped as `$$2`? This would be what we tell a collector user to do, no?

jaronoff97: i.e. does the first `$` escape just `$1` or does it escape the entire `$1_$2`?

pkg/collector/reconcile/configmap_test.go
```diff
@@ -94,23 +94,11 @@ receivers:
...
         scrape_interval: 10s
-        scrape_timeout: 10s
-        metrics_path: /metrics
-        scheme: http
-        follow_redirects: true
-        enable_http2: true
-        http_sd_configs:
-        - follow_redirects: false
-          enable_http2: false
-          url: http://test-targetallocator:80/jobs/otel-collector/targets?collector_id=$POD_NAME
```
jaronoff97, line 113: Why are we removing these lines?

pkg/targetallocator/adapters/config_to_prom_config.go
```diff
@@ -57,3 +68,170 @@ func ConfigToPromConfig(cfg string) (map[interface{}]interface{}, error) {
...
+			if !rcErr {
+				continue
+			}
+
+			replacement, rcErr := replacementProperty.(string)
+			if !rcErr {
+				return nil, errorNotAString("replacement")
+			}
+
+			relabelConfig["replacement"] = strings.ReplaceAll(replacement, "$$", "$")
```
jaronoff97, line 132: are we unable to just do the replace all on the string configuration we're given? Do we need to only run it on relabel_configs and metric_relabel_configs?

jaronoff97: @kristinapathak thinks we should maybe try and just regex this change as well.

tests/e2e/targetallocator-features/00-install.yaml
```diff
@@ -53,10 +53,16 @@ spec:
...
-            - job_name: 'otel-collector'
-              scrape_interval: 10s
-              static_configs:
-                - targets: [ '0.0.0.0:8888' ]
+          - job_name: 'otel-collector'
+            scrape_interval: 10s
+            static_configs:
+            - targets: [ '0.0.0.0:8888' ]
+            relabel_configs:
+            - regex: __meta_kubernetes_node_label_(.+)
```
jaronoff97, line 61: for completeness, can we also add in a labelkeep action here too?

pkg/collector/reconcile/config_replace.go
```diff
@@ -44,55 +41,67 @@ type Config struct {
...
+		}
 
-	for i := range cfg.PromConfig.ScrapeConfigs {
-		escapedJob := url.QueryEscape(cfg.PromConfig.ScrapeConfigs[i].JobName)
-		cfg.PromConfig.ScrapeConfigs[i].ServiceDiscoveryConfigs = discovery.Configs{
-			&http.SDConfig{
-				URL: fmt.Sprintf("http://%s:80/jobs/%s/targets?collector_id=$POD_NAME", naming.TAService(instance), escapedJob),
-			},
+		// yaml marshaling/unsmarshaling is preferred because of the problems associated with the conversion of map to a struct using mapstructure
+		promCfg, marshalErr := yaml.Marshal(promCfgMap)
```
jaronoff97, line 62: to confirm my understanding of this – does doing the marshal and then the unmarshal prevent the defaultrelabelconfig from being added and thus removing [this error](https://github.com/prometheus/prometheus/blob/main/model/relabel/relabel.go#LL148C49-L148C49) from being thrown?

jaronoff97: or is that going to be in a separate PR?

jaronoff97 (commented):
Overall, @kristinapathak and I definitely like this version more than the massaging you had previously. It's easier to understand and the overall design is more extendable. We had a few thoughts/questions to confirm what we had discussed. Thank you again for your contribution 🙇

jaronoff97 (commented):
Overall, @kristinapathak and I definitely like this version more than the massaging you had previously. It's easier to understand and the overall design is more extendable. We had a few thoughts/questions to confirm what we had discussed. Thank you again for your contribution 🙇

TylerHelmuth:
@anunarapureddy grab the latest changes from main to deal with the failing autoscale e2e test.

pkg/collector/reconcile/config_replace.go
anunarapureddy: @jaronoff97 I had initially planned to handle this separately, but later decided to include it in the same PR to ensure a clear and coherent thought process. I pushed the latest changes to handle this scenario as well.

pkg/collector/testdata/relabel_config_expected_with_sd_config.yaml
anunarapureddy: @jaronoff97, the code path for constructing the collector configMap does not escape any dollar characters ('$') in the user-specified configuration. Instead, it updates the HTTP SD service configuration by directly modifying the promConfigMap, without the need for unmarshalling and updating the configuration. It is expected that users will properly escape '$' characters in their Prometheus configuration. This approach allows us to support environment variable substitution as well. For instance, in the example above, if a user entered "$$1_$2" in the prometheus configuration, the config map would be created with that value. The "$$1" part will be processed by the regular expression match, while the "$2" part will be replaced by the corresponding environment variable. The fix ensures that there are no problems when the users specify "$$" in the prometheus configuration (as per the OTel guideline).

pkg/collector/reconcile/configmap_test.go
anunarapureddy: @jaronoff97 These are additional global defaults that are included in the prometheus configuration when it is unmarshalled. Since we are no longer unmarshalling the configuration, these defaults had to be removed. The [original configuration](https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/collector/reconcile/test.yaml) does not contain these defaults.

tests/e2e/targetallocator-features/00-install.yaml
anunarapureddy: @jaronoff97 Done!

pkg/collector/testdata/relabel_config_expected_with_sd_config.yaml
jaronoff97: i see, so in this case `$$1_$2` for a regex match of 'thing' would be `thing_` if $2 is unset? I guess my question is, should we also escape the $2 in our test?

pkg/targetallocator/adapters/config_to_prom_config.go
```diff
@@ -57,3 +68,208 @@ func ConfigToPromConfig(cfg string) (map[interface{}]interface{}, error) {
...
+		return nil, errorNotAMap("target_allocator")
+	}
+
+	targetAllocatorCfg["endpoint"] = fmt.Sprintf("http://%s:80", taServiceName)
+	targetAllocatorCfg["interval"] = "30s"
+	targetAllocatorCfg["collector_id"] = "${POD_NAME}"
+
+	// we don't need the scrape configs here anymore with target allocator enabled
+	// Remove the scrape_configs key from the map
+	delete(prometheusCfg, "scrape_configs")
```
kristinapathak, line 272: Can you also add an explanation for this in the function comment?

pkg/targetallocator/adapters/config_to_prom_config.go
```diff
@@ -57,3 +68,208 @@ func ConfigToPromConfig(cfg string) (map[interface{}]interface{}, error) {
 
 	return prometheus, nil
 }
+
+// UnescapeDollarSignsInPromConfig replaces "$$" with "$" in the "replacement" fields of
+// both "relabel_configs" and "metric_relabel_configs" in a Prometheus configuration file.
+func UnescapeDollarSignsInPromConfig(cfg string) (map[interface{}]interface{}, error) {
```
kristinapathak, line 74: What is the test coverage like on these new functions? It seems like there are a lot of `if` statements present but very few test cases.

kristinapathak (commented):
This looks great! Left a few very minor suggestions/questions.

jaronoff97:
@anunarapureddy we just added some more e2e tests to help us catch more of these in the future, and there's already one for labeldropping that you just need to uncomment to enable. ([PR](https://github.com/open-telemetry/opentelemetry-operator/pull/1762)) Could you rebase and uncomment that test when you get a chance as well?

pkg/targetallocator/adapters/config_to_prom_config.go
```diff
@@ -57,3 +68,208 @@ func ConfigToPromConfig(cfg string) (map[interface{}]interface{}, error) {
...
+
+	scrapeConfigs, ok := scrapeConfigsProperty.([]interface{})
+	if !ok {
+		return nil, errorNotAList("scrape_configs")
+	}
+
+	for _, config := range scrapeConfigs {
+		scrapeConfig, ok := config.(map[interface{}]interface{})
+		if !ok {
+			return nil, errorNotAMap("scrape_config")
```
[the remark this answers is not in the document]
anunarapureddy: Done.

pkg/targetallocator/adapters/config_to_prom_config.go
anunarapureddy: Done.

pkg/targetallocator/adapters/config_to_prom_config.go
anunarapureddy: Adding test cases for missing and invalid configurations as well.

pkg/targetallocator/adapters/config_to_prom_config.go
```diff
@@ -57,3 +68,208 @@ func ConfigToPromConfig(cfg string) (map[interface{}]interface{}, error) {
...
+		}
+
+		// Check for other types of service discovery configs (e.g. dns_sd_configs, file_sd_configs, etc.)
+		for key := range scrapeConfig {
+			keyStr, keyErr := key.(string)
+			if !keyErr {
+				continue
+			}
+			if sdRegex.MatchString(keyStr) {
+				delete(scrapeConfig, key)
```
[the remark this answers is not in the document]
anunarapureddy: Done.

anunarapureddy:
@jaronoff97 Uncommented the "labeldrop" e2e test case and added additional unit test cases to cover missing or invalid configurations.

jaronoff97:
Awesome, thank you! running those now and will give another review today as well once they pass 🤞

jaronoff97:
looks like a single linter bug @anunarapureddy

anunarapureddy:
@jaronoff97 Seems like the linter gave up because it took too long. I checked the code on my own machine, and I didn't spot any lint failures. Everything looks good from my end!

pkg/collector/reconcile/config_replace.go
```diff
@@ -36,9 +32,6 @@ type targetAllocator struct {
 	Endpoint    string        `yaml:"endpoint"`
 	Interval    time.Duration `yaml:"interval"`
 	CollectorID string        `yaml:"collector_id"`
-	// HTTPSDConfig is a preference that can be set for the collector's target allocator, but the operator doesn't
-	// care about what the value is set to. We just need this for validation when unmarshalling the configmap.
-	HTTPSDConfig interface{} `yaml:"http_sd_config,omitempty"`
```
jaronoff97, line 41: i think we still need this? otherwise we'll run in to the unmarshal error about an unknown field.

jaronoff97 (approved):
just one quick question, otherwise this is looking good to me!

pkg/collector/reconcile/config_replace.go
anunarapureddy: @jaronoff97 Apologies! It seems that this happened during the rebase and conflict resolution process. I have thoroughly reviewed everything now, and it appears to be in good shape.

pkg/targetallocator/adapters/config_to_prom_config.go
```diff
@@ -59,6 +78,201 @@ func ConfigToPromConfig(cfg string) (map[interface{}]interface{}, error) {
...
+
+	scrapeConfigsProperty, ok := prometheusConfig["scrape_configs"]
+	if !ok {
+		return nil, errorNoComponent("scrape_configs")
+	}
+
+	scrapeConfigs, ok := scrapeConfigsProperty.([]interface{})
+	if !ok {
+		return nil, errorNotAList("scrape_configs")
+	}
```
swiatekm-sumo, line 107: I don't think we should error if these don't exist, but rather just exit. Target allocator doesn't require Prometheus configuration.

Maybe this should be done in a separate PR though, as I believe this mirrors the current behaviour. WDYT @jaronoff97 ?

pkg/collector/reconcile/configmap.go
```diff
@@ -103,7 +103,7 @@ func desiredTAConfigMap(params Params) (corev1.ConfigMap, error) {
 		labels["app.kubernetes.io/version"] = "latest"
 	}
 
-	prometheusReceiverConfig, err := ta.ConfigToPromConfig(params.Instance.Spec.Config)
+	prometheusReceiverConfig, err := ta.UnescapeDollarSignsInPromConfig(params.Instance.Spec.Config)
```
swiatekm-sumo, line 106: I think it's worth adding a comment here as well. Without context it's not clear at all why we'd need to unescape these here.

pkg/targetallocator/adapters/config_to_prom_config.go
```diff
@@ -59,6 +78,201 @@ func ConfigToPromConfig(cfg string) (map[interface{}]interface{}, error) {
...
```
swiatekm-sumo, line 200: Should we error if the Prometheus config doesn't exist, or just return quietly? See my comment for the TA config unescaping as well.

swiatekm-sumo (commented):
Looks good to me overall. I think we're being too strict about requiring the Prometheus config to exist, but that can be relaxed in a future PR if need be. Overall, despite the verbosity (should we consider parsing permissively into structs instead?), I think this will greatly improve the reliability of the config transformations.

pkg/targetallocator/adapters/config_to_prom_config.go
anunarapureddy: @jaronoff97 @swiatekm-sumo I can create a follow up PR to relax this restriction.

pkg/collector/reconcile/configmap.go
anunarapureddy: Done. Added a comment with the reason to unescape.

anunarapureddy:
@jaronoff97 @swiatekm-sumo Addressed the review comments and I can work on creating a follow up PR to relax the restrictions on prometheus configuration.

swiatekm-sumo (approved):
LGTM

anunarapureddy:
@bryan-aguilar That's a good suggestion. I can create a follow up PR to update the TA documentation.

bryan-aguilar:
> @bryan-aguilar That's a good suggestion. I can create a follow up PR to update the TA documentation.

Do we predict further changes to this functionality after this PR? I worry about documentation drift and a possible loss of context if we don't lump documentation and functional changes in the same PR.

anunarapureddy:
> > @bryan-aguilar That's a good suggestion. I can create a follow up PR to update the TA documentation.
> 
> Do we predict further changes to this functionality after this PR? I worry about documentation drift and a possible loss of context if we don't lump documentation and functional changes in the same PR.

Could you please review the Updated the README documentation which reflects the translations being done?

README.md
```diff
@@ -429,28 +436,27 @@ spec:
...
-            scrape_timeout: 10s
-            evaluation_interval: 1m
           scrape_configs:
           - job_name: otel-collector
-            honor_timestamps: true
             scrape_interval: 10s
-            scrape_timeout: 10s
-            metrics_path: /metrics
-            scheme: http
-            follow_redirects: true
```
Aneurysm9, line 450: Shouldn't these values be preserved in the configuration given to the collector?  These are scrape-time options that are needed by the receiver.

README.md
```diff
@@ -463,7 +469,26 @@ Behind the scenes, the OpenTelemetry Operator will convert the Collector’s con
           exporters: [logging]
 ```
 
-Note how the Operator added a `global` section and a new `http_sd_configs` to the `otel-collector` scrape config, pointing to a Target Allocator instance it provisioned.
+Note how the Operator removes any existing service discovery configurations (e.g., `static_configs`, `file_sd_configs`, etc.) from the `scrape_configs` section and adds a single `http_sd_configs` configuration, pointing to a Target Allocator instance it provisioned.
```
Aneurysm9, line 472:
```suggestion
Note how the Operator removes any existing service discovery configurations (e.g., `static_configs`, `file_sd_configs`, etc.) from the `scrape_configs` section and adds an `http_sd_configs` configuration pointing to a Target Allocator instance it provisioned.
```

More than one `http_sd_configs` entry will be present if there is more than one scrape job in the incoming config.  There should be one per job.  The example only had one job, but a real config is likely to have more.

pkg/collector/reconcile/config_replace.go
```diff
@@ -65,42 +63,39 @@ func ReplaceConfig(instance v1alpha1.OpenTelemetryCollector) (string, error) {
...
+		if getCfgPromErr != nil {
+			return "", getCfgPromErr
+		}
 
-	var cfg Config
-	if marshalErr = yaml.UnmarshalStrict(promCfg, &cfg); marshalErr != nil {
-		return "", fmt.Errorf("error unmarshaling YAML: %w", marshalErr)
-	}
+		// type coercion checks are handled in the AddTAConfigToPromConfig method above
+		config["receivers"].(map[interface{}]interface{})["prometheus"] = updPromCfgMap
```
Aneurysm9, line 75: I don't believe it is safe to assume that the receiver ID is always `prometheus`.  There may be multiple prometheus receiver configurations and any or all of them may have a name in addition to the type in their identifier.  This needs to iterate over all entries in the `receivers` map and process any that have an identifier matching `/^prometheus(\/.+)?$/`.

README.md
anunarapureddy: @Aneurysm9 The replacement process occurs for each job in the incoming configuration. The code iterates through all the jobs and updates the configuration accordingly. If it would be helpful for clarity, I can include an additional job in the example.

pkg/collector/reconcile/config_replace.go
anunarapureddy: @Aneurysm9 The assignment of this logic remains unchanged as it is an existing implementation (https://github.com/open-telemetry/opentelemetry-operator/blob/d6d9f2f628a80858460cddbbee85b4ca9f13fecf/pkg/collector/reconcile/config_replace.go#L99). However, I can certainly address it as an improvement in a follow-up PR.

kristinapathak (approved):
This is awesome! Thanks for putting in so much time and effort into getting `$`s working 🙂

bryan-aguilar (approved):
Thank you for adding the documentation!

jaronoff97 (approved):

jaronoff97:
once tests pass, will merge this in :)

PR #1728, opened 2023-05-10
[chore] [target-allocator] Add filterStrategy to TargetAllocator README.md

Juliaj:
I recommend adding a paragraph to TargetAllocator README.md to raise the awareness of using `filterStrategy` setting. 

`filterStrategy` is an important setting for developers to consider when reusing existing Prometheus scrape configuration for OpenTelemetry metrics collection pipeline. This configuration would help reduce your time to debug issues that you may be seeing in terms of intermittent `out of order sample` error. These errors are typically caused when the targets get split across multiple collectors which creates duplicate time series.

cc: @alolita @Aneurysm9

### DIFF
--- a/cmd/otel-allocator/README.md
+++ b/cmd/otel-allocator/README.md
@@ -43,6 +43,12 @@ spec:
           processors: []
           exporters: [logging]
 ```
+## Set up filter_strategy
+`FilterStrategy` determines how to filter targets before allocating them among multiple collectors. `relabel-config` is the currently supported option. With `FilterStrategy`, targets will be consolidated based on Prometheus configuration "relabel_config". By default, filtering is disabled.
+
+It's important to consider setting `FilterStrategy` if you have a scrape job with a discovery role for example, `endpoints` which could produce multiple targets per endpoint. If these targets lead to the same metric after a relabeling step and they are assigned to different collectors, the outcome may be undesirable because duplicate time series are created for the metric.  
+
+For metrics collection pipeline where metrics are exported to Prometheus, you may see `out of order sample` error intermittenly if the time series produced have shifted time stamps from different collectors. 
 
 ## PrometheusCR specifics
 TargetAllocator discovery of PrometheusCRs can be turned on by setting
--- a/cmd/otel-allocator/README.md
+++ b/cmd/otel-allocator/README.md
@@ -44,7 +44,7 @@ spec:
           exporters: [logging]
 ```
 ## Set up filterStrategy
-`filterStrategy` determines how to filter targets before allocating them among multiple collectors. `relabel-config` is the currently supported option. With `filterStrategy`, targets will be consolidated based on Prometheus configuration "relabel_config". By default, filtering is disabled.
+`filterStrategy` determines how to filter targets before allocating them among multiple collectors. `relabel-config` is the currently supported option. With `filterStrategy`, targets will be consolidated based on Prometheus configuration `relabel_config`. By default, filtering is disabled.
 
 It's important to consider setting `filterStrategy` if you have a scrape job with a discovery role for example, `endpoints` which could produce multiple targets per endpoint. If these targets lead to the same metric after a relabeling step and they are assigned to different collectors, the outcome may be undesirable because duplicate time series are created for the metric. 
 
--- a/cmd/otel-allocator/README.md
+++ b/cmd/otel-allocator/README.md
@@ -43,12 +43,20 @@ spec:
           processors: []
           exporters: [logging]
 ```
-## Set up filter_strategy
-`FilterStrategy` determines how to filter targets before allocating them among multiple collectors. `relabel-config` is the currently supported option. With `FilterStrategy`, targets will be consolidated based on Prometheus configuration "relabel_config". By default, filtering is disabled.
+## Set up filterStrategy
+`filterStrategy` determines how to filter targets before allocating them among multiple collectors. `relabel-config` is the currently supported option. With `filterStrategy`, targets will be consolidated based on Prometheus configuration "relabel_config". By default, filtering is disabled.
 
-It's important to consider setting `FilterStrategy` if you have a scrape job with a discovery role for example, `endpoints` which could produce multiple targets per endpoint. If these targets lead to the same metric after a relabeling step and they are assigned to different collectors, the outcome may be undesirable because duplicate time series are created for the metric.  
+It's important to consider setting `filterStrategy` if you have a scrape job with a discovery role for example, `endpoints` which could produce multiple targets per endpoint. If these targets lead to the same metric after a relabeling step and they are assigned to different collectors, the outcome may be undesirable because duplicate time series are created for the metric. 
 
 For metrics collection pipeline where metrics are exported to Prometheus, you may see `out of order sample` error intermittenly if the time series produced have shifted time stamps from different collectors. 
+
+Here is an example of this setting,
+```yaml
+  targetAllocator:
+    enabled: true
+    allocationStrategy: consistent-hashing
+    filterStrategy: relabel-config
+```
 
 ## PrometheusCR specifics
 TargetAllocator discovery of PrometheusCRs can be turned on by setting
--- a/apis/v1alpha1/opentelemetrycollector_types.go
+++ b/apis/v1alpha1/opentelemetrycollector_types.go
@@ -197,8 +197,11 @@ type OpenTelemetryTargetAllocator struct {
 	// +optional
 	AllocationStrategy OpenTelemetryTargetAllocatorAllocationStrategy `json:"allocationStrategy,omitempty"`
 	// FilterStrategy determines how to filter targets before allocating them among the collectors.
-	// The only current option is relabel-config (drops targets based on prom relabel_config).
-	// Filtering is disabled by default.
+	// The only current option is relabel-config (drops targets based on prom relabel_config). It's important to consider
+	// setting it with relabel-config if you have a scrape job with a discovery role for example, endpoints which could
+	// produce multiple targets per endpoint. If these targets lead to the same metric after a relabeling step and they
+	// are assigned to different collectors, the outcome may be undesirable because duplicate time series are created for
+	// the metric. Filtering is disabled by default.
 	// +optional
 	FilterStrategy string `json:"filterStrategy,omitempty"`
 	// ServiceAccount indicates the name of an existing service account to use with this instance. When set,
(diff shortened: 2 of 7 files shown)


cmd/otel-allocator/README.md
```diff
@@ -43,6 +43,12 @@ spec:
           processors: []
           exporters: [logging]
 ```
+## Set up filter_strategy
+`FilterStrategy` determines how to filter targets before allocating them among multiple collectors. `relabel-config` is the currently supported option. With `FilterStrategy`, targets will be consolidated based on Prometheus configuration "relabel_config". By default, filtering is disabled.
```
pavolloffay, line 47: `relabel_config` formatting seems more appropriate.

Shall we as well update the CRD docs https://github.com/open-telemetry/opentelemetry-operator/blob/2e537f04a7bbc35b7c05f61b81ae559da0018204/apis/v1alpha1/opentelemetrycollector_types.go#L203? 


Could the operator automatically set the "correct" value?

pavolloffay (commented):
Thanks for improving our docs 🎉 

just a minor comment

cmd/otel-allocator/README.md
jaronoff97: we could do this by default if we want, but that's probably another issue :)

Juliaj: @pavolloffay, thanks a bunch for your feedback. The formatting was fixed and added content to CRD docs. This is my very first PR, let me know if I missed any files/steps. 

To your other comment on setting `relabel-config` as default, I agree with Jacob's comment, we may need get inputs first thus a separate issue is better.

cmd/otel-allocator/README.md
```diff
@@ -43,6 +43,20 @@ spec:
           processors: []
           exporters: [logging]
 ```
+## Set up filterStrategy
+`filterStrategy` determines how to filter targets before allocating them among multiple collectors. `relabel-config` is the currently supported option. With `filterStrategy`, targets will be consolidated based on Prometheus configuration `relabel_config`. By default, filtering is disabled.
+
+It's important to consider setting `filterStrategy` if you have a scrape job with a discovery role for example, `endpoints` which could produce multiple targets per endpoint. If these targets lead to the same metric after a relabeling step and they are assigned to different collectors, the outcome may be undesirable because duplicate time series are created for the metric.
```
jaronoff97, line 49: IIRC filter strategy doesn't solve this issue, it would still produce the same results as before (i.e. duplication of scrape targets). This would only be solved by changing the scrape configuration or the service/pod monitor.

cmd/otel-allocator/README.md
```diff
@@ -43,6 +43,20 @@ spec:
           processors: []
           exporters: [logging]
 ```
+## Set up filterStrategy
+`filterStrategy` determines how to filter targets before allocating them among multiple collectors. `relabel-config` is the currently supported option. With `filterStrategy`, targets will be consolidated based on Prometheus configuration `relabel_config`. By default, filtering is disabled.
+
+It's important to consider setting `filterStrategy` if you have a scrape job with a discovery role for example, `endpoints` which could produce multiple targets per endpoint. If these targets lead to the same metric after a relabeling step and they are assigned to different collectors, the outcome may be undesirable because duplicate time series are created for the metric. 
+
+For metrics collection pipeline where metrics are exported to Prometheus, you may see `out of order sample` error intermittenly if the time series produced have shifted time stamps from different collectors.
```
jaronoff97, line 51: I'm not sure I've come across this issue nor if within this filter strategy doc comment is the right place to call this out.

jaronoff97 (changes requested):
Two comments on the README. I'm not sure i've seen this duplication issue related to the filter strategy. could you explain that more?

jaronoff97 (changes requested):
Two comments on the README. I'm not sure i've seen this duplication issue related to the filter strategy. could you explain that more?

cmd/otel-allocator/README.md
Juliaj: We have a metrics pipeline with number of jobs scraping EKS related metrics, see an example job below. Our pipeline setup is TA +  multi-collectors (min 3 with autoscaler). Each collector consists of a Prometheus receiver plus Prometheus remote write exporter.  We've seen lot of out of order sample errors reported from the pipeline, intermittently however persistently.  

```
- job_name: kubernetes-service-endpoints
  honor_labels: true
  honor_timestamps: true
  scrape_interval: 1m
  scrape_timeout: 10s
  metrics_path: /metrics
  scheme: http
  follow_redirects: true
  relabel_configs:
  - source_labels: [__meta_kubernetes_service_annotation_prometheus_io_scrape]
    separator: ;
    regex: "true"
    replacement: $1
    action: keep
  - source_labels: [__meta_kubernetes_service_annotation_prometheus_io_scrape_slow]
    separator: ;
    regex: "true"
    replacement: $1
    action: drop
  - source_labels: [__meta_kubernetes_service_annotation_prometheus_io_scheme]
    separator: ;
    regex: (https?)
    target_label: __scheme__
    replacement: $1
    action: replace
  - source_labels: [__meta_kubernetes_service_annotation_prometheus_io_path]
    separator: ;
    regex: (.+)
    target_label: __metrics_path__
    replacement: $1
    action: replace
  - source_labels: [__address__, __meta_kubernetes_service_annotation_prometheus_io_port]
    separator: ;
    regex: ([^:]+)(?::\d+)?;(\d+)
    target_label: __address__
    replacement: $1:$2
    action: replace
  - source_labels: [__meta_kubernetes_namespace]
    separator: ;
    regex: (.*)
    target_label: namespace
    replacement: $1
    action: replace
  - source_labels: [__meta_kubernetes_service_name]
    separator: ;
    regex: (.*)
    target_label: service
    replacement: $1
    action: replace
  - source_labels: [__meta_kubernetes_pod_node_name]
    separator: ;
    regex: (.*)
    target_label: node
    replacement: $1
    action: replace
  kubernetes_sd_configs:
  - role: endpoints
    kubeconfig_file: ""
    follow_redirects: true
...
```

After loads of debugging,  we discovered one of the causes. Two main factors contributed to the error: the scrape config from Prometheus and allocation of the discovered targets for the same eventual metric got split across different collector instances. 

Let's look at an example. With the role: `endpoints`,  for each endpoint address, a target is discovered per port.  For a service endpoint having three unique port names,  three targets 10.244.0.4:53, 10.244.0.4:53 and 10.244.0.4:9153 were discovered. We've seen  these 3 targets got split as following

```
collector #1: 10.244.0.4:53

collector #2: 10.244.0.4:53
collector #2: 10.244.0.4:9153
```
However,  after step below,  these three targets ended up with an identical label `instance=10.244.0.4:9153` for the eventual metric.  

```
- source_labels: [__address__, __meta_kubernetes_service_annotation_prometheus_io_port]
    separator: ;
    regex: ([^:]+)(?::\d+)?;(\d+)
    target_label: __address__
    replacement: $1:$2
    action: replace
```

We observed duplicated time series for the same metric were produced by these collectors with slightly shifted time stamps. Intermittently, the shifted time stamps triggered the out of order sample error if older samples were sent to PWR. 
  
You are right, we had experimented with changing the scrape configuration to get rid of the error. Considering the number of the jobs we have, the process is tedious and we're looking for alternatives. @Aneurysm9 suggested to look into `filterStrategy`.  Since then, we've been testing the fix in our pipeline.  Results so far, 

- We've seen good error reduction rate for multiple scrape jobs, job`kubernetes-service-endpoints` is one of them.
- We've only had moderate success of using this with scrape job `kubernetes-pods` (see config below) which we are still debugging. 
- We're also seeing the `out of order sample` errors reported with `target_info`.

Previously, we had reproduced the out of order sample error in a setup of single collector (no TA).  In that case,  this setting is not applicable. 

```- job_name: kubernetes-pods
  honor_labels: true
  honor_timestamps: true
  scrape_interval: 1m
  scrape_timeout: 10s
  metrics_path: /metrics
  scheme: http
...
  follow_redirects: true
  relabel_configs:
  - source_labels: [__meta_kubernetes_pod_annotation_prometheus_io_scrape]
    separator: ;
    regex: "true"
    replacement: $1
    action: keep
  - source_labels: [__meta_kubernetes_pod_annotation_prometheus_io_scrape_slow]
    separator: ;
    regex: "true"
    replacement: $1
    action: drop
  - source_labels: [__meta_kubernetes_pod_annotation_prometheus_io_scheme]
    separator: ;
    regex: (https?)
    target_label: __scheme__
    replacement: $1
    action: replace
  - source_labels: [__meta_kubernetes_pod_annotation_prometheus_io_path]
    separator: ;
    regex: (.+)
    target_label: __metrics_path__
    replacement: $1
    action: replace
  - source_labels: [__address__, __meta_kubernetes_pod_annotation_prometheus_io_port]
    separator: ;
    regex: ([^:]+)(?::\d+)?;(\d+)
    target_label: __address__
    replacement: $1:$2
    action: replace
  - source_labels: [__meta_kubernetes_namespace]
    separator: ;
    regex: (.*)
    target_label: namespace
    replacement: $1
    action: replace
  - source_labels: [__meta_kubernetes_pod_name]
    separator: ;
    regex: (.*)
    target_label: pod
    replacement: $1
    action: replace
  - source_labels: [__meta_kubernetes_pod_phase]
    separator: ;
    regex: Pending|Succeeded|Failed|Completed
    replacement: $1
    action: drop
  kubernetes_sd_configs:
  - role: pod
    kubeconfig_file: ""
    follow_redirects: true
```

jaronoff97: Although tedious, editing your scrape config to prevent dual scrapes is the correct resolution here. You're most likely seeing out of order scrapes still because the filter strategy doesn't necessarily do the functionality you are expecting here (i.e. if any of the labels discovered for a pod / scrape config are different they will still result in two separate targets) In some cases you may get lucky that your scrape config removes the duplicates, but that's not a panacea.

Juliaj: We made changes to our scrape config to introduce additional labels as well as collector_name, results have been looking stable so far, most of the errors were gone except `up`, `scraped_*` and `target_info` which we'll continue investigate.
@jaronoff97 @pavolloffay, thank you for your feedback.  I'll continue to evaluate other viable alternatives e.g.  logic from Prometheus to handle duplicated target ([code](https://github.com/prometheus/prometheus/blob/80b7f73d267a812b3689321554aec637b75f468d/scrape/scrape.go#L539)) for this. For now, I will close this PR.

PR #1729, opened 2023-05-10
Improve config validation for prometheus receiver and target allocator

swiatekm-sumo:
In a follow-up to #1688, improve validation for prometheus receiver configuration. This turned out to be more complicated than expected, so I'd like some more eyes to verify I haven't missed anything.

The validation logic for the prometheus receiver config itself is:
* if target allocator is not enabled on the Collector CR, prometheus config under the `config` key is required
* if target allocator is enabled
  - either `config` has to exist to inject http_sd_configs, or `target_allocator` needs to exist 
  - if additionally target allocator rewrite is enabled, nothing is required

However, I also noticed that target allocator itself doesn't correctly validate its configuration. For example, it's possible to pass an empty Prometheus config, but this will cause a crash in the targetDiscoverer. So I added an additional check, requiring that either:
* there is a non-empty Prometheus config
* or PrometheusCR is enabled

I submitted both of these changes in a single PR, as they're related, but I can resubmit separately if that'll make review easier.

### DIFF
--- a/apis/v1alpha1/opentelemetrycollector_webhook.go
+++ b/apis/v1alpha1/opentelemetrycollector_webhook.go
@@ -27,7 +27,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
 	"github.com/open-telemetry/opentelemetry-operator/pkg/featuregate"
-
 	ta "github.com/open-telemetry/opentelemetry-operator/pkg/targetallocator/adapters"
 )
 
--- a/apis/v1alpha1/opentelemetrycollector_webhook.go
+++ b/apis/v1alpha1/opentelemetrycollector_webhook.go
@@ -26,6 +26,8 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
+	"github.com/open-telemetry/opentelemetry-operator/pkg/featuregate"
+
 	ta "github.com/open-telemetry/opentelemetry-operator/pkg/targetallocator/adapters"
 )
 
@@ -160,7 +162,11 @@ func (r *OpenTelemetryCollector) validateCRDSpec() error {
 
 	// validate Prometheus config for target allocation
 	if r.Spec.TargetAllocator.Enabled {
-		_, err := ta.ConfigToPromConfig(r.Spec.Config)
+		promCfg, err := ta.ConfigToPromConfig(r.Spec.Config)
+		if err != nil {
+			return fmt.Errorf("the OpenTelemetry Spec Prometheus configuration is incorrect, %w", err)
+		}
+		err = ta.ValidatePromConfig(promCfg, r.Spec.TargetAllocator.Enabled, featuregate.EnableTargetAllocatorRewrite.IsEnabled())
 		if err != nil {
 			return fmt.Errorf("the OpenTelemetry Spec Prometheus configuration is incorrect, %w", err)
 		}
--- a/cmd/otel-allocator/config/config.go
+++ b/cmd/otel-allocator/config/config.go
@@ -132,7 +132,7 @@ func ParseCLI() (CLIConfig, error) {
 
 // ValidateConfig validates the cli and file configs together.
 func ValidateConfig(config *Config, cliConfig *CLIConfig) error {
-	scrapeConfigsPresent := (config.Config != nil || len(config.Config.ScrapeConfigs) > 0)
+	scrapeConfigsPresent := (config.Config != nil && len(config.Config.ScrapeConfigs) > 0)
 	if !(*cliConfig.PromCRWatcherConf.Enabled || scrapeConfigsPresent) {
 		return fmt.Errorf("at least one scrape config must be defined, or Prometheus CR watching must be enabled")
 	}
--- a/cmd/otel-allocator/config/config.go
+++ b/cmd/otel-allocator/config/config.go
@@ -132,10 +132,9 @@ func ParseCLI() (CLIConfig, error) {
 
 // ValidateConfig validates the cli and file configs together.
 func ValidateConfig(config *Config, cliConfig *CLIConfig) error {
-	if !*cliConfig.PromCRWatcherConf.Enabled &&
-		(config.Config == nil ||
-			len(config.Config.ScrapeConfigs) == 0) {
-		return fmt.Errorf("either at least one scrape config must be defined, or Prometheus CR watching must be enabled")
+	scrapeConfigsPresent := (config.Config != nil || len(config.Config.ScrapeConfigs) > 0)
+	if !(*cliConfig.PromCRWatcherConf.Enabled || scrapeConfigsPresent) {
+		return fmt.Errorf("at least one scrape config must be defined, or Prometheus CR watching must be enabled")
 	}
 	return nil
 }
--- a/cmd/otel-allocator/config/config.go
+++ b/cmd/otel-allocator/config/config.go
@@ -129,3 +129,13 @@ func ParseCLI() (CLIConfig, error) {
 	cLIConf.ClusterConfig = clusterConfig
 	return cLIConf, nil
 }
+
+// ValidateConfig validates the cli and file configs together.
+func ValidateConfig(config *Config, cliConfig *CLIConfig) error {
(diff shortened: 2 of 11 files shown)


apis/v1alpha1/opentelemetrycollector_webhook.go
```diff
@@ -23,6 +23,8 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
+	"github.com/open-telemetry/opentelemetry-operator/pkg/featuregate"
+
```
matej-g, line 27: Nit: Should this be one import group?

cmd/otel-allocator/main.go
```diff
@@ -148,10 +152,12 @@ func main() {
 	runGroup.Add(
 		func() error {
 			// Initial loading of the config file's scrape config
-			err = targetDiscoverer.ApplyConfig(allocatorWatcher.EventSourceConfigMap, cfg.Config)
-			if err != nil {
-				setupLog.Error(err, "Unable to apply initial configuration")
-				return err
+			if cfg.Config != nil {
+				err = targetDiscoverer.ApplyConfig(allocatorWatcher.EventSourceConfigMap, cfg.Config)
```
matej-g, line 156: I wonder whether it would make more sense to check for `cfg.Config == nil` inside of `ApplyConfig` and if so, just do early return

cmd/otel-allocator/config/config.go
```diff
@@ -129,3 +129,13 @@ func ParseCLI() (CLIConfig, error) {
 	cLIConf.ClusterConfig = clusterConfig
 	return cLIConf, nil
 }
+
+// ValidateConfig validates the cli and file configs together.
+func ValidateConfig(config *Config, cliConfig *CLIConfig) error {
+	if !*cliConfig.PromCRWatcherConf.Enabled &&
+		(config.Config == nil ||
+			len(config.Config.ScrapeConfigs) == 0) {
+		return fmt.Errorf("either at least one scrape config must be defined, or Prometheus CR watching must be enabled")
```
matej-g, line 138: Small nit
```suggestion
		return fmt.Errorf("at least one scrape config must be defined, or Prometheus CR watching must be enabled")
```

matej-g (approved):
This looks like a nice addition! 💪 I added only a few nits.

cmd/otel-allocator/config/config.go
jaronoff97: i'd prefer us to actually split up this if in to three statements. Although it's more LOC we'll be able to give a more descriptive error back which I think would be better.

jaronoff97 (commented):
just a few thoughts, overall this is going to be super useful though. thank you!

cmd/otel-allocator/main.go
swiatekm-sumo: I can't reverse it if it stays where it is in `main`, as `cfg.Config` == `nil` by itself does not imply an error. Or do you mean doing this inside `ApplyConfig`?

jaronoff97: wouldn't `cfg.Config == nil` imply an error is i guess my conjecture here. i.e. if we don't have a config to apply at all we should fail fast.

cmd/otel-allocator/config/config.go
swiatekm-sumo: I wrote it this way anticipating that we might want to add more checks in the future, in which case they all need to be negative checks (as in "this error condition doesn't occur"). And I can't really rewrite it into three separate statements, as it's logically an OR.

I could make it nested instead, but I'm not sure if you can really make the error message any more useful that way.

cmd/otel-allocator/main.go
swiatekm-sumo: @matej-g makes sense, you've convinced me.

@jaronoff97 Having the raw Prometheus config be nil and Prometheus CR handling enabled looks like a valid configuration to me. That's why I added the validate method and check for this specific state.

swiatekm-sumo: I moved the check to ApplyConfig and added a test for this.

cmd/otel-allocator/config/config.go
swiatekm-sumo: I rewrote this condition to be clearer, let me know if this helps.

cmd/otel-allocator/target/discovery.go
```diff
@@ -64,6 +64,9 @@ func NewDiscoverer(log logr.Logger, manager *discovery.Manager, hook discoveryHo
 }
 
 func (m *Discoverer) ApplyConfig(source allocatorWatcher.EventSource, cfg *config.Config) error {
+	if cfg == nil {
+		return nil
```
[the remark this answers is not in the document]
swiatekm-sumo: Done. Do we have a mock logger implemented somewhere to test this?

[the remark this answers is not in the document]
jaronoff97: i don't think so... wish we did. I think checking that we get a nil error on a nil config may be enough here.

jaronoff97:
@swiatekm-sumo my colleague @kristinapathak is also going to take a look at this as well before i merge. We've been deep in to configuration related PRs for the TA as of late so this will be a good continuation.

cmd/otel-allocator/config/config.go
```diff
@@ -129,3 +129,12 @@ func ParseCLI() (CLIConfig, error) {
 	cLIConf.ClusterConfig = clusterConfig
 	return cLIConf, nil
 }
+
+// ValidateConfig validates the cli and file configs together.
+func ValidateConfig(config *Config, cliConfig *CLIConfig) error {
+	scrapeConfigsPresent := (config.Config != nil || len(config.Config.ScrapeConfigs) > 0)
```
kristinapathak, line 135: From where `ValidateConfig` is currently being called, it's not possible to have a nil value for `config` or `cliConfig`, but isn't it possible for `config.Config` to be nil?  Can we add unit tests for `ValidateConfig` and make sure to include this test case?

pkg/collector/reconcile/config_replace.go
```diff
@@ -57,6 +57,11 @@ func ReplaceConfig(instance v1alpha1.OpenTelemetryCollector) (string, error) {
 		return "", getCfgPromErr
 	}
 
+	validateCfgPromErr := ta.ValidatePromConfig(promCfgMap, instance.Spec.TargetAllocator.Enabled, featuregate.EnableTargetAllocatorRewrite.IsEnabled())
```
kristinapathak, line 60: Is `ValidatePromConfig()` meant to be called every time `ta.ConfigToPromConfig()` is called? Put another way, is there ever a situation when we want to call one function without the other? I'm wondering the benefit of two separate functions.  Is it separation of concerns? Not making a breaking change to the `adapters` pkg API?

kristinapathak (commented):
This looks great! Left two questions/suggestions.

pkg/collector/reconcile/config_replace.go
swiatekm-sumo: Not making a breaking change and keeping current semantics of `ta.ConfigToPromConfig()` is part of it, although the separation of concerns is a nice bonus from that. Though overall there isn't any strong reason behind it, if we're ok breaking `ConfigToPromConfig`, we could simply make validation a private function and have `ConfigToPromConfig` call it.

cmd/otel-allocator/config/config.go
swiatekm-sumo: Done.

cmd/otel-allocator/main.go
```diff
@@ -74,6 +74,10 @@ func main() {
 		setupLog.Error(configLoadErr, "Unable to load configuration")
 	}
 
+	if validationErr := config.ValidateConfig(&cfg, &cliConf); validationErr != nil {
+		setupLog.Error(validationErr, "Invalid configuration")
```
jaronoff97, line 78: in a future version should we fast fail here?

pkg/collector/reconcile/config_replace.go
jaronoff97: that sounds like a good followup for the future :)

jaronoff97 (approved):

PR #1731, opened 2023-05-11
[chore] Only Login on main push

jaronoff97:
This PR makes it so that we only attempt github action logins on a main push to the repo rather than on every PR which was causing failures due to missing credentials.

### DIFF
--- a/.github/workflows/publish-autoinstrumentation-dotnet.yaml
+++ b/.github/workflows/publish-autoinstrumentation-dotnet.yaml
@@ -49,12 +49,14 @@ jobs:
 
       - name: Log into Docker.io
         uses: docker/login-action@v2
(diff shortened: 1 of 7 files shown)


TylerHelmuth (commented):
Without the login wont the `Build and Push` step fail?

jaronoff97:
@TylerHelmuth we only run the push step for the same condition, and i don't think we need a login to build the docker image.

jaronoff97:
example here https://github.com/open-telemetry/opentelemetry-operator/actions/runs/4949638652/jobs/8852081644?pr=1731

TylerHelmuth:
I was looking for and `if` statement and missed the `push` condition on that job 👍

TylerHelmuth (approved):

PR #1733, opened 2023-05-11
[chore] move autoscale test to its own suite

TylerHelmuth:
Updates the e2e test workflow to execute the autoscale test in its own suite. This will provide stability for that complex test and allow the other e2e test suites to scale better.  

Adds `fail-fast: false` to the matrix so that jobs can run to completion even if another fails.

Reduces the number of current tests in the `e2e` suite from 8 to 4 to help reduce flakiness.  This slightly increases the time it takes the tests to run, but thats better than rerunning them.


jaronoff97:
@TylerHelmuth is there a reason the E2E tests havent run yet for this?

TylerHelmuth:
@jaronoff97 I believe the reason is bc I renamed the action when I added another dimension to the matrix but the repository is still looking for the original action name as part of its branch protections. So it is waiting for 2 actions that no longer exist.

TylerHelmuth:
@jaronoff97 I can't see the branch protection rules, but I'm basing my response on this discussion: https://github.com/orgs/community/discussions/26698

jaronoff97:
@TylerHelmuth this may prove difficult, because i can't merge the PR until those pass...

PR #1736, opened 2023-05-12
Add OTEL service version to sdk injection

### LINKED ISSUE #1670: Detect and inject the service.version resource attribute
The Open Telemetry specification for resource attributes ([link](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service)) contains a set of attributes, such as `service.version` marked as experimental.

This attribute is very useful in debugging and monitoring using telemetry data, because it reveals when deployments of new artifacts are performed and the version of each service is attached to its telemetry data, so comparisons can be performed. This attribute could be manually set for manual instrumentation

Are there any plans for injecting the `service.version` from the operator, for manually or auto-instrumented services?

Side-question: is the operator aware of the service using manual-instrumentation, because I am also interested for this functionality on services of artifacts implemented in go.

KitSutliff:
fixes #1670

Adds version detection and reporting to sdk injection.


pkg/instrumentation/sdk.go
```diff
@@ -171,6 +171,18 @@ func (i *sdkInjector) injectCommonSDKConfig(ctx context.Context, otelinst v1alph
 			Value: chooseServiceName(pod, resourceMap, appIndex),
 		})
 	}
+
+	idx = getIndexOfEnv(container.Env, constants.EnvOTELServiceVersion)
```
jaronoff97, line 175: Can we put this behind a featuregate? ([ex](https://github.com/open-telemetry/opentelemetry-operator/blob/87d4e103b2484fbedb7232d124c6e5cd7b2c84b7/pkg/instrumentation/podmutator.go#L89))

jaronoff97 (commented):
just one thought, also be sure to add a changelog file for this awesome feature :)

jaronoff97 (commented):
just one thought, also be sure to add a changelog file for this awesome feature :)

pkg/instrumentation/sdk.go
```diff
@@ -171,6 +171,18 @@ func (i *sdkInjector) injectCommonSDKConfig(ctx context.Context, otelinst v1alph
...
+	if idx == -1 {
+		vsn := chooseServiceVersion(pod, appIndex)
+		if vsn != "" {
+			container.Env = append(container.Env, corev1.EnvVar{
```
TylerHelmuth, line 179: @KitSutliff have you reviewed the telemetry produced by a service that had this env var and the auto-instrumentation injected?  While there is an [SDK requirement to support OTEL_SERVICE_NAME](https://opentelemetry.io/docs/specs/otel/sdk-environment-variables/#general-sdk-configuration), there isn't a requirement for `OTEL_SERVICE_VERSION`.  To get the SDK to emit `service.version` I believe the operator needs to update the `OTEL_RESOURCE_ATTRIBUTES` env var and include `service.version`.  In this function it would be the `resourceMap`.

pkg/instrumentation/sdk.go
KitSutliff: Thank you for the feedback, I'll work this into the change :)

pavolloffay: This is a great feature. Will it be enabled by default?

KitSutliff: Hey, thanks!  I'd be happy to leave this on by default.  Any objections?

jaronoff97: None from me! :bow:

KitSutliff:
I think I have a version of a fix working as outlined [here](https://github.com/open-telemetry/opentelemetry-operator/issues/1670#issuecomment-1521955532).  While poking around I noticed the [Kubernetes docs](https://kubernetes.io/docs/concepts/containers/images/#image-names) mention setting the default tag to 'latest' when it's not already specified.  Is this something we'd like to implement or are we good to go as is?

jaronoff97:
@KitSutliff in lieu of us supplying a tag, kubernetes will automatically assume latest – so no need for us to do anything here.

pkg/instrumentation/sdk.go
```diff
@@ -299,6 +313,17 @@ func chooseServiceName(pod corev1.Pod, resources map[string]string, index int) s
 	return pod.Spec.Containers[index].Name
 }
 
+// obtains version by splitting image string on ":" and extracting final element from resulting array
+func chooseServiceVersion(pod corev1.Pod, index int) string {
+	parts := strings.Split(pod.Spec.Containers[index].Image, ":")
+	tag := parts[len(parts)-1]
+	//guard statement to handle case where image name has a port number
```
jaronoff97, line 320: is there a test for this guard?

jaronoff97: I guess i'm not aware of when an image would have a port number? or how checking if `/` is present means there is a port present.

jaronoff97 (approved):
Just one question, otherwise this looks great!

pkg/instrumentation/sdk.go
KitSutliff: Was trying to cover all our bases and noticed the [doc](https://kubernetes.io/docs/concepts/containers/images/#image-names) mentions port numbers as a possibility.

"Images can also include a registry hostname; for example: fictional.registry.example/imagename, and possibly a port number as well; for example: fictional.registry.example:10443/imagename."

I can nix it if you'd like, but if we want to handle that edge case I'm equally happy to write a test for that scenario.
Do you have a preference either way?

pkg/constants/env.go
```diff
@@ -22,7 +22,8 @@ const (
 	EnvOTELTracesSampler        = "OTEL_TRACES_SAMPLER"
 	EnvOTELTracesSamplerArg     = "OTEL_TRACES_SAMPLER_ARG"
 
-	EnvPodName  = "OTEL_RESOURCE_ATTRIBUTES_POD_NAME"
-	EnvPodUID   = "OTEL_RESOURCE_ATTRIBUTES_POD_UID"
-	EnvNodeName = "OTEL_RESOURCE_ATTRIBUTES_NODE_NAME"
+	EnvPodName        = "OTEL_RESOURCE_ATTRIBUTES_POD_NAME"
+	EnvPodUID         = "OTEL_RESOURCE_ATTRIBUTES_POD_UID"
+	EnvNodeName       = "OTEL_RESOURCE_ATTRIBUTES_NODE_NAME"
+	EnvServiceVersion = "OTEL_RESOURCE_ATTRIBUTES_SERVICE_VERSION"
```
TylerHelmuth, line 28: I believe this can be removed now.

TylerHelmuth:
@KitSutliff you may need to get the latest from main to get CI to pass

pkg/instrumentation/sdk.go
jaronoff97: oh wow i've never seen that before... I think you can keep it in but yeah maybe include a test for it in `sdk-test.go`

bundle/manifests/opentelemetry.io_instrumentations.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.12.0
+    controller-gen.kubebuilder.io/version: v0.11.3
```
jaronoff97, line 5: looks like a bad merge here

pkg/instrumentation/sdk.go
KitSutliff: Added in a test that should do the trick!

TylerHelmuth (approved):

KitSutliff:
Ok, think I've got it all pretty and polished!  This project is cool and I'm excited about my first commit! ✨

pkg/instrumentation/sdk.go
pavolloffay: It would be good to handle the registry ports as well and add more tests:
* with registry port
*  specific tag
*  no tag
*  latest

pavolloffay (commented):
What version is used if no tag is set? What versions is set if no tag is set and the repository port is used? What version is used if no tag is set?

pkg/instrumentation/sdk.go
KitSutliff: These cases are already handled.

- with registry port is handled here: https://github.com/open-telemetry/opentelemetry-operator/pull/1736/files#diff-260da5a412521a337a3a193ef1e3a344d0e745696b643769367d591a91f84d16R379
- no tag is handled here: https://github.com/open-telemetry/opentelemetry-operator/pull/1736/files#diff-260da5a412521a337a3a193ef1e3a344d0e745696b643769367d591a91f84d16R379
- latest is handled here: https://github.com/open-telemetry/opentelemetry-operator/pull/1736/files#diff-260da5a412521a337a3a193ef1e3a344d0e745696b643769367d591a91f84d16R147 (we actually handle this in TWENTY test cases)

I'm not sure what you mean by 'specific' tag? Can you show me another test in this file that addresses similar concerns?

According to the Kubernetes [docs around image tag naming specifications](https://kubernetes.io/docs/concepts/containers/images/#image-names) these tests already the possible range of values.

pavolloffay (approved):

jaronoff97 (approved):

PR #1737, opened 2023-05-12
provide default resource limits for go sidecar container

### LINKED ISSUE #1732: [instrumentation] Add support for Resources on Go spec.
The other supported auto-instrumentation languages allow specifying resource limits for the init container.  We should add similar functionality for the Go auto-instrumentation sidecar.

Original PR for context: https://github.com/open-telemetry/opentelemetry-operator/pull/1571

frzifus:
Closes #1732


apis/v1alpha1/instrumentation_webhook.go
```diff
@@ -134,6 +134,18 @@ func (r *Instrumentation) Default() {
...
+	if r.Spec.Go.Resources.Limits == nil {
+		r.Spec.Go.Resources.Limits = corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("500m"),
+			corev1.ResourceMemory: resource.MustParse("32Mi"),
+		}
+	}
+	if r.Spec.Go.Resources.Requests == nil {
+		r.Spec.Go.Resources.Requests = corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("50m"),
+			corev1.ResourceMemory: resource.MustParse("32Mi"),
```
TylerHelmuth, line 146: Since the Go instrumentation sidecar is continuously involved with the instrumentation process these values may be too low for a default.  @open-telemetry/go-instrumentation-approvers what would you recommend for default resources?

MikeGoldsmith: I'll review and discuss with the other contributors and get back to you.

TylerHelmuth: Based on the discussion in the Go Instr SIG today I will run some stress tests and find some good default values.

frzifus: thanks @TylerHelmuth :)

TylerHelmuth: I did some stress testing today on my local machine in a kind cluster.  It isn't a very reliable setup, but even with docker allowed to use 8 cores and 9GB of memory I was able to get the [example-greeting-service](https://github.com/honeycombio/example-greeting-service/tree/operator-injecting-go-instr/go-auto-instrumented) to produce ~1000 traces a second.  The memory of the go instrumentation sidecar was very consistent at 10Mi of memory and CPU fluctuated around .1 utilization when set to a limit of 100m.

Overall though, I don't trust my test.  Spent a lot of time on it trying different things and could push the number of traces up reliably as my cluster would just fall over. With that in mind, these defaults are probably an ok starting place.  I am unable to produce production-like traffic on my own machine but what I could produce stayed under these numbers.

frzifus: Sounds like a fair starting point - thanks for your efforts!

jaronoff97 (approved):
Seems reasonable to me, @TylerHelmuth i'll await your review before merging.

PR #1740, opened 2023-05-15
[instrumentation] Update to latest Go Instrumentation version

TylerHelmuth:


versions.txt
```diff
@@ -30,7 +30,7 @@ autoinstrumentation-python=0.38b0
 autoinstrumentation-dotnet=0.7.0
 
 # Represents the current release of Go instrumentation.
-autoinstrumentation-go=v0.2.0-alpha
+autoinstrumentation-go=v0.2.1-alpha
```
jaronoff97, line 33: i think this should just be in the next release, right?

TylerHelmuth: Looking through the RELEASE doc I'd agree with you.  If we only bump the versions during the release and not before then this can wait.

pavolloffay: The versions can be updated in the main branch and I would also argue that it is preferred if more code changes are required. 

The release process has to make sure that the latest available versions are used.

TylerHelmuth: I can confirm there are no code changes needed with this change, only the bump.

frzifus (approved):

PR #1759, opened 2023-05-19
Fix annotation's propagation

### LINKED ISSUE #900: Propagate annotations to Deployment's PodTemplate

yuriolisa:
Fixes #900 

The first PR was closed (#1505) in favor to keep the podAnnotations.


TylerHelmuth (approved):
LGTM

Issue #1515, opened 2023-02-24
Having issues with opentelemetry operator python autoinstrumentation


lukeab:
building a fastapi example to promote opentelemetry in a client environment. 
deploy helm chart of the operator, tried at 0.21.5 and updated to 0.24 just to see if it fixed it, (it did not). 

I deploy the app with the auto inject annotation,  the pod starts up and i can hit the endpoints,  but I get the following warnings on the pod startup and no tracing is working : 

```
Configuration of configurator failed
Traceback (most recent call last):
  File "/otel-auto-instrumentation/requests/compat.py", line 11, in <module>
    import chardet
ModuleNotFoundError: No module named 'chardet'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/otel-auto-instrumentation/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py", line 105, in _load_configurators
    entry_point.load()().configure(auto_instrumentation_version=__version__)  # type: ignore
  File "/otel-auto-instrumentation/opentelemetry/sdk/_configuration/__init__.py", line 384, in configure
    self._configure(**kwargs)
  File "/otel-auto-instrumentation/opentelemetry/sdk/_configuration/__init__.py", line 400, in _configure
    _initialize_components(kwargs.get("auto_instrumentation_version"))
  File "/otel-auto-instrumentation/opentelemetry/sdk/_configuration/__init__.py", line 337, in _initialize_components
    trace_exporters, metric_exporters, log_exporters = _import_exporters(
  File "/otel-auto-instrumentation/opentelemetry/sdk/_configuration/__init__.py", line 270, in _import_exporters
    for (exporter_name, exporter_impl,) in _import_config_components(
  File "/otel-auto-instrumentation/opentelemetry/sdk/_configuration/__init__.py", line 104, in _import_config_components
    component_impl = entry_point.load()
  File "/otel-auto-instrumentation/pkg_resources/__init__.py", line 2468, in load
    return self.resolve()
  File "/otel-auto-instrumentation/pkg_resources/__init__.py", line 2474, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/otel-auto-instrumentation/opentelemetry/exporter/otlp/proto/http/trace_exporter/__init__.py", line 24, in <module>
    import requests
  File "/otel-auto-instrumentation/requests/__init__.py", line 45, in <module>
    from .exceptions import RequestsDependencyWarning
[...]

PR #1763, opened 2023-05-23
[instrumentation] Switch to python:3.11 image

TylerHelmuth:
Switch to building the python requirements on the `3.11` image instead of the `3.11-alpine` image.  This ensures that the requirements are built using all necessary python dependencies.

Closes https://github.com/open-telemetry/opentelemetry-operator/issues/1515


srikanthccv (approved):
Thank you

jaronoff97 (approved):

PR #1789, opened 2023-06-01
Added new information to CRD Status

### LINKED ISSUE #1355: Improve CRD Status for OpenTelemetryCollector
Currently, the operator returns minimal information for the status of a collector:
```
⫸ k get opentelemetrycollector/kube-otel-stack-metrics
NAME                      MODE          VERSION   AGE
kube-otel-stack-metrics   statefulset   0.60.0    54d
```

This means that when someone deploying a collector wants to observe the rollout status, there could be a potential race condition when running something like 
```
k rollout status sts/kube-otel-stack-metrics-collector
```
The race exists because the operator may not have applied the changes before someone runs this rollout command.

Ideally, someone would be able to see the rollout status like:
```
k rollout status opentelemetrycollector/kube-otel-stack-metrics
```

Or we could include more information in the status for a collector (an example, not exactly what we would want)
```
⫸ k get opentelemetrycollector/kube-otel-stack-metrics
NAME                      MODE          VERSION        READY   AGE
kube-otel-stack-metrics   statefulset   0.60.0          3/4    54d
```

How do people currently watch the rollout status of their collectors? Are there other fields in here that would be useful to add to the collector's status?

yuriolisa:
Resolves #1355 

@jaronoff97

### DIFF
--- a/apis/v1alpha1/opentelemetrycollector_types.go
+++ b/apis/v1alpha1/opentelemetrycollector_types.go
@@ -82,7 +82,7 @@ type OpenTelemetryCollectorSpec struct {
 	// Replicas is the number of pod instances for the underlying OpenTelemetry Collector. Set this if your are not using autoscaling
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
-	// MinReplicas sets a lower bound to the autoscaling feature.  Set this if your are using autoscaling. It must be at least 1
+	// MinReplicas sets a lower bound to the autoscaling feature.  Set this if you are using autoscaling. It must be at least 1
 	// +optional
 	// Deprecated: use "OpenTelemetryCollector.Spec.Autoscaler.MinReplicas" instead.
 	MinReplicas *int32 `json:"minReplicas,omitempty"`
@@ -248,6 +248,10 @@ type ScaleSubresourceStatus struct {
 	// OpenTelemetryCollector's deployment or statefulSet.
 	// +optional
 	Replicas int32 `json:"replicas,omitempty"`
+
+	// readyReplicas is the number of pods targeted by this OpenTelemetryCollector with a Ready Condition.
+	// +optional
+	ReadyReplicas int32 `json:"readyReplicas,omitempty"`
 }
 
 // OpenTelemetryCollectorStatus defines the observed state of OpenTelemetryCollector.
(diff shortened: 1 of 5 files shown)


apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -278,7 +282,10 @@ type OpenTelemetryCollectorStatus struct {
 // +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.scale.replicas,selectorpath=.status.scale.selector
 // +kubebuilder:printcolumn:name="Mode",type="string",JSONPath=".spec.mode",description="Deployment Mode"
 // +kubebuilder:printcolumn:name="Version",type="string",JSONPath=".status.version",description="OpenTelemetry Version"
+// +kubebuilder:printcolumn:name="Current",type="string",JSONPath=".status.scale.readyReplicas"
+// +kubebuilder:printcolumn:name="Desired",type="string",JSONPath=".status.scale.replicas"
```
jaronoff97, line 286: don't these fields need to be added to the status struct?

jaronoff97 (approved):
Could you also update some of the e2e tests to check the status on the CRD?

apis/v1alpha1/opentelemetrycollector_types.go
jaronoff97: I didn't see image in the status struct?

TylerHelmuth (approved):

pavolloffay (approved):

pkg/collector/reconcile/opentelemetry_test.go
```diff
@@ -18,10 +18,10 @@ import (
 	"context"
 	"testing"
 
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
```
pavolloffay, line 21: no change here

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -248,6 +248,16 @@ type ScaleSubresourceStatus struct {
...
+
+	// readyReplicas is the number of pods targeted by this OpenTelemetryCollector with a Ready Condition.
+	// +optional
+	ReadyReplicas int32 `json:"readyReplicas,omitempty"`
+
+	// statusReplicas is the number of pods targeted by this OpenTelemetryCollector
+	// with a Ready Condition and non-terminated pdos targeted by this OpenTelemetryCollector's
+	// deployment, statefulSet.
+	// +optional
+	StatusReplicas string `json:"statusReplicas,omitempty"`
```
pavolloffay, line 260: Why is it string and not a number?

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -248,6 +248,16 @@ type ScaleSubresourceStatus struct {
 	// OpenTelemetryCollector's deployment or statefulSet.
 	// +optional
 	Replicas int32 `json:"replicas,omitempty"`
+
+	// readyReplicas is the number of pods targeted by this OpenTelemetryCollector with a Ready Condition.
+	// +optional
+	ReadyReplicas int32 `json:"readyReplicas,omitempty"`
+
+	// statusReplicas is the number of pods targeted by this OpenTelemetryCollector
+	// with a Ready Condition and non-terminated pdos targeted by this OpenTelemetryCollector's
```
pavolloffay, line 257: I am not sure if I get the description. Is it ready pods plus non-terminated pods? 

s/pdos/pods

apis/v1alpha1/opentelemetrycollector_types.go
yuriolisa: Due to printcolumns (kubebuilder) constraints, we cannot concatenate two int status values to present Ready -> 1/1, instead of creating two columns Current and Desired.

PR #1794, opened 2023-06-02
Don't make unnecessary local changes in make targets

swiatekm-sumo:
Kustomize works by applying successive overlays over resources. Currently, when we need to apply changes to our base, we do it directly in our manifests, modifying files which are checked into source control. This is annoying for local work, though doesn't really matter for CI.

This change uses a separate overlay for any changes we might want to make over our defaults. The kustomization.yaml file lives in the ignored `./dist` directory, and is created and modified on demand.

I tested all the targets making use of this, and everything worked as expected, but I'd appreciate review from someone involved in writing this Makefile.

This is part of my effort to make running E2E tests locally nicer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,12 +88,10 @@ SED ?= $(shell which gsed 2>/dev/null || which sed)
 .PHONY: ensure-generate-is-noop
 ensure-generate-is-noop: VERSION=$(OPERATOR_VERSION)
 ensure-generate-is-noop: USER=open-telemetry
-ensure-generate-is-noop: bundle
-	@# on make bundle config/manager/kustomization.yaml includes changes, which should be ignored for the below check
-	@git restore config/manager/kustomization.yaml
+ensure-generate-is-noop: update-bundle
 	@git diff -s --exit-code apis/v1alpha1/zz_generated.*.go || (echo "Build failed: a model has been changed but the generated resources aren't up to date. Run 'make generate' and update your PR." && exit 1)
-	@git diff -s --exit-code bundle config || (echo "Build failed: the bundle, config files has been changed but the generated bundle, config files aren't up to date. Run 'make bundle' and update your PR." && git diff && exit 1)
-	@git diff -s --exit-code bundle.Dockerfile || (echo "Build failed: the bundle.Dockerfile file has been changed. The file should be the same as generated one. Run 'make bundle' and update your PR." && git diff && exit 1)
+	@git diff -s --exit-code bundle config || (echo "Build failed: the bundle, config files has been changed but the generated bundle, config files aren't up to date. Run 'make update-bundle' and update your PR." && git diff && exit 1)
+	@git diff -s --exit-code bundle.Dockerfile || (echo "Build failed: the bundle.Dockerfile file has been changed. The file should be the same as generated one. Run 'make update-bundle' and update your PR." && git diff && exit 1)
 	@git diff -s --exit-code docs/api.md || (echo "Build failed: the api.md file has been changed but the generated api.md file isn't up to date. Run 'make api-docs' and update your PR." && git diff && exit 1)
 
 .PHONY: all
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -124,25 +129,30 @@ uninstall: manifests kustomize
...
 release-artifacts: set-image-controller
-	mkdir -p dist
-	$(KUSTOMIZE) build config/default -o dist/opentelemetry-operator.yaml
+	$(KUSTOMIZE) build $(KUSTOMIZATION_DIR) -o dist/opentelemetry-operator.yaml
+
+.PHONY: create-kustomization
+create-kustomization:
+	mkdir -p $(KUSTOMIZATION_DIR)
+	rm -f $(KUSTOMIZATION_DIR)/kustomization.yaml
+	cd $(KUSTOMIZATION_DIR) && kustomize create --resources ../$(KUSTOMIZATION_BASE)
```
pavolloffay, line 155: `$(KUSTOMIZE)`

pavolloffay:
I think this partly solves the problem with temporary changes applied to the checked-in files. For instance the CSV is still modified.

I would assume that if no code is changed in the repository, then the developer workflow `IMG=docker.io/${USER}/opentelemetry-operator:dev-$(git rev-parse --short HEAD)-$(date +%s) make generate bundle` (documented in the contributing) will not produce any changes.

```bash
IMG=docker.io/${USER}/opentelemetry-operator:dev-$(git rev-parse --short HEAD)-$(date +%s) make generate bundle 
git status                                                                                                                                    ploffay@fedora
On branch PR1794
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
ls dist                                                                                                                          130 ↵ ploffay@fedora
kustomization.yaml  opentelemetry-operator.yaml
------------------------------------------------
```

On the main branch the changes look like: 
```
~/projects/open-telemetry/opentelemetry-operator (main*) » g s                                                                                                                                      ploffay@fedora
On branch main
Your branch is up to date with 'up/main'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
	modified:   config/manager/kustomization.yaml

no changes added to commit (use "git add" and/or "git commit -a")
```

That said, I like the improvement, but it's a small benefit for unnecessary more complicated workflow in the makefile.

swiatekm-sumo:
Yeah, I wasn't really trying to address the bundle generation, which I don't understand as well as the kustomize parts. Maybe we can generate the bundle into a temporary folder in the same way, and only check it into source control when releasing?

I'd like to do that in a separate PR though.

swiatekm-sumo:
> I find the checked-in `bundle` directory useful and that it represents the current state of the repository. The released bundles can be found in the OLM repos (see the release.md).

Why does the checked-in bundle exist? Is it just to be able to track changes to it? If so, we can change the current command to `make release-bundle`, and add a new `make local-bundle` which would ignore some fields. It's kind of a hack, but I don't see a better way.

Incidentally, https://github.com/operator-framework/operator-sdk/issues/6285 recommends to just not keep the bundle in source control and to not run `make bundle` in CI.

swiatekm-sumo:
@pavolloffay the way I see it, there are three different reasons to generate a bundle:

1. We're releasing, and we actually want to publish it
2. We want to ensure our local changes (for example in CRDs) are reflected in the checked-in bundle
3. We want to run our local changes against an OpenShift cluster, and want an actual bundle with all the local changes included.

Seems like we should have different commands for these. 

1. The current `bundle` command could become `release-bundle` and continue working the same way.
2. We could have a `update-bundle` command for this which would make sure to ignore incidental changes to metadata
3. I think 3 would require more changes to be easy, but I'm not really that familiar with OpenShift

When we tell contributors to `make bundle install run`, is `bundle` even necessary there?

jaronoff97 (approved):

jaronoff97:
@pavolloffay mind taking another look here as well?

iblancasa (approved):

jaronoff97 (approved):
I think this LGTM. Can we also update the docs to say users should run `make update-bundle` if they change any manifests things?

swiatekm-sumo:
> I think this LGTM. Can we also update the docs to say users should run `make update-bundle` if they change any manifests things?

We have a make target called `ensure-generate-is-noop` which checks this and prints an error message instructing the user to run `make update-bundle` if needed. This runs as part of `make test`, so it should catch problems eariy. 

I've also added a note to the contributing doc for completeness sake.

pavolloffay:
It's been a long time since I wanted to look at this PR, sorry for too long delay. I think we can do this without increasing the complexity of the kustomize setup in the repository. We had the same issue in the tempo-oprator and it was solved without changed the kustomize. I will take a a look and submit a similar approach here as well.

pavolloffay:
here is the alternative PR https://github.com/open-telemetry/opentelemetry-operator/pull/2421

swiatekm-sumo:
A lot of the issues this PR aimed to fix were resolved in #2421 and #2479. Closing this for now, might reevaluate in the future.

PR #1795, opened 2023-06-02
Bump all the dependencies

jaronoff97:
Another dependabot doesn't know what the hell its doing PR. This bumps all our dependencies that dependabot is trying to bump.


cmd/operator-opamp-bridge/go.mod
```diff
@@ -6,89 +6,90 @@ require (
 	github.com/go-logr/logr v1.2.4
 	github.com/oklog/ulid/v2 v2.1.0
 	github.com/open-telemetry/opamp-go v0.7.0
-	github.com/open-telemetry/opentelemetry-operator v1.51.0
```
jaronoff97, line 9: We should retract this version, it's what we have as latest in go.pkg.dev right now.

cmd/operator-opamp-bridge/metrics/reporter.go
```diff
@@ -122,16 +121,17 @@ func NewMetricReporter(
 
 	// Create some metrics that will be reported according to OpenTelemetry semantic
 	// conventions for process metrics (conventions are TBD for now).
-	reporter.processCpuTime, err = reporter.meter.Float64ObservableGauge(
+	reporter.processCpuTime, err = reporter.meter.Float64ObservableCounter(
```
TylerHelmuth, line 124: Is changing this metric type a breaking change?

cmd/operator-opamp-bridge/metrics/reporter.go
```diff
@@ -122,16 +121,17 @@ func NewMetricReporter(
...
 		"process.cpu.time",
-		instrument.WithFloat64Callback(reporter.processCpuTimeFunc),
+		metric.WithFloat64Callback(reporter.processCpuTimeFunc),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("can't create process time metric: %w", err)
 	}
-	reporter.processMemoryPhysical, err = reporter.meter.Float64ObservableCounter(
+
+	reporter.processMemoryPhysical, err = reporter.meter.Float64ObservableGauge(
```
TylerHelmuth, line 132: Is changing this metric type a breaking change?

jaronoff97: it was previously incorrect, but also no one is using this image yet other than me pretty much because we don't have a way to deploy this yet.

cmd/operator-opamp-bridge/metrics/reporter.go
jaronoff97: see below

TylerHelmuth (approved):

PR #1796, opened 2023-06-05
java autoinstrumentaion init container running as regular user [chore]

maksimstankevic:
Added regular user in java init image to be able to run as non root. This allowed us to overcome the issue with setting runAsNonRoot to "true" in Pod securityContext, before this change autoinstrumentation init container would fail to run with restricted setting as it was running as root.

### DIFF
--- a/autoinstrumentation/java/Dockerfile
+++ b/autoinstrumentation/java/Dockerfile
@@ -8,6 +8,10 @@ FROM busybox
 
 ARG version
 
-ADD https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/download/v$version/opentelemetry-javaagent.jar /javaagent.jar
+RUN addgroup otel -g 1000 && adduser -u 1000 -D -G otel otel
+
+USER 1000:1000
+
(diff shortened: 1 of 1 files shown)


autoinstrumentation/java/Dockerfile
```diff
@@ -8,6 +8,10 @@ FROM busybox
 
 ARG version
 
-ADD https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/download/v$version/opentelemetry-javaagent.jar /javaagent.jar
+RUN addgroup otel -g 1000 && adduser -u 1000 -D -G otel otel
+
+USER 1000:1000
```
frzifus, line 13: I might be wrong, but ive something in mind - Was there an issue with something like this on OpenShift?

ping @iblancasa @pavolloffay

maksimstankevic: @frzifus will this Openshift thing block the merge?

maksimstankevic: I took sometime to see how that works in Openshift and yeah - they are dealing with this on their side not to run it with root and docker image with explicit UID is introducing issues for them.

maksimstankevic:
this is not going to work for Openshift support

PR #1801, opened 2023-06-05
Add ParentBasedJaegerRemote sampler & validate argument

srikanthccv:
This adds support for `parentbased_jaeger_remote` and validates the sampler argument.


apis/v1alpha1/instrumentation_webhook.go
```diff
@@ -207,7 +238,19 @@ func (r *Instrumentation) validate() error {
...
 		}
-	case AlwaysOn, AlwaysOff, JaegerRemote, ParentBasedAlwaysOn, ParentBasedAlwaysOff, XRaySampler:
+	case JaegerRemote, ParentBasedJaegerRemote:
+		// value is a comma separated list of endpoint, pollingIntervalMs, initialSamplingRate
+		// Example: `endpoint=http://localhost:14250,pollingIntervalMs=5000,initialSamplingRate=0.25`
+		if r.Spec.Sampler.Argument != "" {
+			err := parseJaegerRemoteSamplerArgument(r.Spec.Sampler.Argument)
+
+			if err != nil {
+				return fmt.Errorf("spec.sampler.argument is not a valid JaegerRemoteSampler argument: %s", err.Error())
```
TylerHelmuth, line 248: Is only ParentBasedJaegerRemote allowed to have an arg?  Does JaegerRemote need included in this case?

TylerHelmuth (commented):
I believe you'll need to run a `make api-docs` to get the CI to pass after addressing the other general lint error.

apis/v1alpha1/instrumentation_webhook.go
srikanthccv: Both are allowed to have argument value. Both are covered in the case statement above. Can you help me understand where else it needs to be included?

TylerHelmuth: I ask because the case seems to change the behavior of `JaegerRemote`.  Before is was in a nothing case, but now if a user uses JaegerRemote and supplies arguments it will drop into the `parseJaegerRemoteSamplerArgument` function.  Is that what should happen?  Is the current behavior of doing no validation for JaegerRemote a bug?

srikanthccv: > Is that what should happen? Is the current behavior of doing no validation for JaegerRemote a bug?

Yes, I don't know if we should call it a bug. When `spec.sampler.argument` is provided for `(parentbased_)jaeger_remote`, it's good to have a  basic validation as we do for the `TraceIDRatio`.

apis/v1alpha1/instrumentation_webhook_test.go
```diff
@@ -92,6 +92,53 @@ func TestInstrumentationValidatingWebhook(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "pollingIntervalMs is not a number",
+			err:  "invalid pollingIntervalMs: abc",
+			inst: Instrumentation{
+				Spec: InstrumentationSpec{
+					Sampler: Sampler{
+						Type:     JaegerRemote,
```
TylerHelmuth, line 101: Can you add duplicate tests for ParentJaegerRemote?

apis/v1alpha1/instrumentation_webhook.go
```diff
@@ -195,6 +195,37 @@ func (r *Instrumentation) ValidateDelete() (admission.Warnings, error) {
...
+func parseJaegerRemoteSamplerArgument(argument string) error {
+	parts := strings.Split(argument, ",")
+
+	for _, part := range parts {
+		kv := strings.Split(part, "=")
+		if len(kv) != 2 {
+			return fmt.Errorf("invalid argument: %s", part)
+		}
+
+		switch kv[0] {
```
pavolloffay, line 207: Should we check that some parameters are provided (e.g. endpoint) or can the SDK default to something meaningful? 

The default values are not defined in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration

apis/v1alpha1/instrumentation_webhook_test.go
pavolloffay: or just define a separate matrix for  Jaeger samplers

srikanthccv: I can do that. I thought they would be redundant since there is no difference.

TylerHelmuth (approved):

.chloggen/parentbased-jaeger-remote.yaml
```diff
@@ -0,0 +1,16 @@
...
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: samplers
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add ParentBasedJaegerRemote sampler & validate argument
+
+# One or more tracking issues related to the change
+issues: [1801]
```
pavolloffay, line 11: 1801 is this PR not issue (the issue doe not exist).

apis/v1alpha1/instrumentation_webhook.go
```diff
@@ -195,6 +195,37 @@ func (r *Instrumentation) ValidateDelete() (admission.Warnings, error) {
 	return nil, nil
 }
 
+func validateJaegerRemoteSmaplerArgument(argument string) error {
+	parts := strings.Split(argument, ",")
+
+	for _, part := range parts {
+		kv := strings.Split(part, "=")
+		if len(kv) != 2 {
+			return fmt.Errorf("invalid argument: %s", part)
```
pavolloffay, line 204: `invalid argument: %s, the argument should be in the form of  key=value`

pavolloffay (commented):
LGTM just a minor comment

.chloggen/parentbased-jaeger-remote.yaml
TylerHelmuth: chlog requires a value for `Issue`.  If no issue is present, an PR number should be used so that the changelog has something to link to.

srikanthccv: Also, GitHub redirects to PR if there is number is not an issue (ex: https://github.com/open-telemetry/opentelemetry-operator/issues/1801).

apis/v1alpha1/instrumentation_webhook.go
```diff
@@ -195,6 +195,37 @@ func (r *Instrumentation) ValidateDelete() (admission.Warnings, error) {
 	return nil, nil
 }
 
+func validateJaegerRemoteSmaplerArgument(argument string) error {
```
jaronoff97, line 198:
```suggestion
func validateJaegerRemoteSamplerArgument(argument string) error {
```

jaronoff97 (approved):
Do we also need to run `make api-docs`?

srikanthccv:
>Do we also need to run make api-docs?

I ran it, and nothing changed.

kamsobon:
@srikanthccv @pavolloffay Sorry for rehashing this old thread, but I think this change misses adjustment in CRD validation: 
As you can see in https://github.com/srikanthccv/opentelemetry-operator/blob/main/apis/v1alpha1/samplers.go#L19 only jaeger_remote is mentioned.

I can add it there but honestly I'm not sure how to start 😅 
Should I raise issue and raise PR?

PR #1803, opened 2023-06-05
Fix version in status for OpenTelemetry Collector after upgrades

### LINKED ISSUE #1802: OpenTelemetry Collector instance is not showing properly the version after upgrade
Depending on the source version you upgrade the OpenTelemetry Operator, the OpenTelemetry Collector will upgrade its version properly or not.

## Steps to reproduce

1. Deploy OpenTelemetry Operator `0.63.1`:
    ```sh
    $ kubectl create -f https://github.com/open-telemetry/opentelemetry-operator/releases/download/v0.63.1/opentelemetry-operator.yaml
    ```
2. Deploy one OpenTelemetry Collector instance
3. Upgrade the OpenTelemetry Operator version:
    ```sh
   $ kubectl apply -f https://github.com/open-telemetry/opentelemetry-operator/releases/download/v0.78.0/opentelemetry-operator.yaml
    ```
4. Wait until the upgrade process is triggered in the operator
5. Check the OTEL Collector version:
```sh
$ kubectl get otelcol
NAME       MODE         VERSION   AGE
simplest   deployment   0.63.1    20m

# Check if the image was updated
$ kubectl get deployment simplest-collector -o yaml
...
    spec:
      containers:
      - args:
        - --config=/conf/collector.yaml
        env:
        - name: POD_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.name
        image: ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector:0.78.0
...
```

As we can see, the image version was updated but not the OTEL Collector version.

## Source of the problem
The error appears because this source code (`pkg/collector/upgrade/upgrade.go`):
```go
	if instanceV.GreaterThan(&Latest.Version) {
		u.Log.Info("skipping upgrade for OpenTelemetry Collector instance, as it's newer than our latest version", "name", otelcol.Name, "namespace", otelcol.Namespace, "version", otelcol.Status.Version, "latest", Latest.Version.String())
		return otelcol, nil
	}
```

is before this one:
```go
	otelcol.Status.Version = u.Version.OpenTelemetryCollector
```

iblancasa:
resolves #1802


pavolloffay:
It seems that some tests failed

pavolloffay:
It will also need a changelog entry

pkg/collector/upgrade/upgrade.go
```diff
@@ -108,7 +108,18 @@ func (u VersionUpgrade) ManagedInstance(ctx context.Context, otelcol v1alpha1.Op
...
+		u.Log.Info("no upgrade routines are needed for the OpenTelemetry instance", "name", otelcol.Name, "namespace", otelcol.Namespace, "version", otelcol.Status.Version, "latest", Latest.Version.String())
+
+		otelColV, err := semver.NewVersion(u.Version.OpenTelemetryCollector)
+		if err != nil {
+			return otelcol, err
+		}
+		if instanceV.LessThan(otelColV) {
+			u.Log.Info("upgraded OpenTelemetry Collector version", "name", otelcol.Name, "namespace", otelcol.Namespace, "version", otelcol.Status.Version)
+			otelcol.Status.Version = u.Version.OpenTelemetryCollector
+		}
```
TylerHelmuth, line 121: The 2 logs seem to contradict each other.

jaronoff97 (approved):

pkg/collector/upgrade/upgrade.go
TylerHelmuth: Oh I missed the `"routines`" in the first log.  

In that case I think we should add a log for when we don't change anything.  We want to keep the existing "skipping upgrade for OpenTelemetry Collector instance" log for when we change nothing.

TylerHelmuth: We only change the version if `if instanceV.LessThan(otelColV)` is true right?  If it is false, then we follow the same logic as before where we could log `"skipping upgrade for OpenTelemetry Collector instance"`.

pavolloffay (approved):

TylerHelmuth (approved):

frzifus (approved):

PR #1710, opened 2023-05-04
[target-allocator] Populate store assets (authorization information) for Prometheus CR watcher

matej-g:
Resolves #1669.

This change adds logic to populate the so-called store assets - this includes information from service and pod monitors that relates to authorization information (e.g. basic auth, bearer token, TLS) - in order to propagate these information to scrape configs. Currently, this logic is not implement, resulting in scrape configurations missing proper credentials.

This PR also adds tests for the `LoadConfig` method for the Prometheus CR watcher.

PR #1653, opened 2023-04-13
Bump kubernetes to 1.27 operator and target allocator

jaronoff97:
closes #1652 

This upgrades all kubernetes dependencies in each module to the latest version 0.27.0 for Kubernetes 1.27.0

Some extra upgrades were included, the dependencies of which have been reviewed and should all be safe upgrades.

PR #1822, opened 2023-06-12
[target allocator] Re-add default global scrape interval in Promethues CR watcher configuration.

### LINKED ISSUE #1811: TargetAllocator : error during loading configuration
Hi, i've got these logs at startup:

```
{"level":"info","ts":"2023-06-05T20:17:18Z","msg":"Starting the Target Allocator"}
{"level":"info","ts":"2023-06-05T20:17:18Z","logger":"allocator","msg":"Unrecognized filter strategy; filtering disabled"}
{"level":"info","ts":"2023-06-05T20:17:18Z","logger":"allocator","msg":"Starting server..."}
{"level":"info","ts":"2023-06-05T20:17:18Z","msg":"Waiting for caches to sync for servicemonitors\n"}
{"level":"info","ts":"2023-06-05T20:17:20Z","msg":"Caches are synced for servicemonitors\n"}
{"level":"info","ts":"2023-06-05T20:17:20Z","msg":"Waiting for caches to sync for podmonitors\n"}
{"level":"info","ts":"2023-06-05T20:17:20Z","msg":"Caches are synced for podmonitors\n"}
{"level":"info","ts":"2023-06-05T20:17:20Z","logger":"allocator","msg":"Successfully started a collector pod watcher","component":"opentelemetry-targetallocator"}
{"level":"error","ts":"2023-06-05T20:17:21Z","logger":"setup","msg":"Unable to load configuration","error":"empty duration string","stacktrace":"main.main.func13\n\t/app/main.go:198\ngithub.com/oklog/run.(*Group).Run.func1\n\t/go/pkg/mod/github.com/oklog/run@v1.1.0/group.go:38"}
{"level":"error","ts":"2023-06-05T20:17:21Z","logger":"setup","msg":"Unable to load configuration","error":"empty duration string","stacktrace":"main.main.func13\n\t/app/main.go:198\ngithub.com/oklog/run.(*Group).Run.func1\n\t/go/pkg/mod/github.com/oklog/run@v1.1.0/group.go:38"}
{"level":"error","ts":"2023-06-05T20:17:22Z","logger":"setup","msg":"Unable to load configuration","error":"empty duration string","stacktrace":"main.main.func13\n\t/app/main.go:198\ngithub.com/oklog/run.(*Group).Run.func1\n\t/go/pkg/mod/github.com/oklog/run@v1.1.0/group.go:38"}
```

with this configuration:

```yaml
kind: OpenTelemetryCollector
metadata:
  labels:
    app.kubernetes.io/instance: opentelemetry-collector
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: opentelemetry-collector
    app.kubernetes.io/part-of: opentelemetry-collector
    app.kubernetes.io/version: 1.0.0
    argocd.argoproj.io/instance: opentelemetry-collector
    helm.sh/chart: opentelemetry-collector-1.0.0
  name: metrics
  namespace: opentelemetry
spec:
  config: |
    exporters:
      logging:
        verbosity: normal
      prometheus:
        endpoint: 0.0.0.0:9090
        metric_expiration: 180m
        resource_to_telemetry_conversion:
          enabled: true
      prometheusremotewrite/mimir:
        endpoint: http://mimir-nginx.monitoring.svc.cluster.local:80/api/v1/push
    extensions:
      basicauth/grafanacloud:
        client_auth:
          password: ${GRAFANA_CLOUD_METRICS_APIKEY}
          username: ${GRAFANA_CLOUD_METRICS_ID}
      health_check: null
      memory_ballast:
        size_in_percentage: 20
      pprof:
        endpoint: :1888
      zpages:
        endpoint: :55679
    processors:
      batch:
        send_batch_max_size: 1500
        send_batch_size: 1500
        timeout: 15s
      k8sattributes:
        extract:
          metadata:
          - k8s.namespace.name
          - k8s.pod.name
          - k8s.pod.uid
          - k8s.node.name
          - k8s.pod.start_time
          - k8s.deployment.name
          - k8s.replicaset.name
          - k8s.replicaset.uid
          - k8s.daemonset.name
          - k8s.daemonset.uid
          - k8s.job.name
          - k8s.job.uid
          - k8s.cronjob.name
          - k8s.statefulset.name
          - k8s.statefulset.uid
          - container.image.tag
          - container.image.name
        passthrough: false
        pod_association:
        - sources:
          - from: resource_attribute
            name: k8s.pod.name
      memory_limiter:
        check_interval: 5s
        limit_percentage: 90
        spike_limit_percentage: 30
      resource:
        attributes:
        - action: insert
          key: collector.name
          value: ${KUBE_POD_NAME}
    receivers:
      hostmetrics:
        collection_interval: 60s
        scrapers:
          cpu: null
          disk: null
          filesystem: null
          load: null
          memory: null
          network: null
          processes: null
      prometheus:
        config:
          global:
            evaluation_interval: 60s
            scrape_interval: 60s
            scrape_timeout: 60s
        target_allocator:
          collector_id: ${POD_NAME}
          endpoint: http://metrics-targetallocator:80
          http_sd_config:
            refresh_interval: 60s
          interval: 30s
    service:
      extensions:
      - health_check
      - memory_ballast
      - pprof
      - zpages
      pipelines:
        metrics:
          exporters:
          - logging
          - prometheus
          processors:
          - batch
          - memory_limiter
          - k8sattributes
          receivers:
          - hostmetrics
          - prometheus
      telemetry:
        logs:
          encoding: json
          level: info
        metrics:
          address: 0.0.0.0:8888
          level: detailed
  env:
  - name: K8S_NODE_NAME
    valueFrom:
      fieldRef:
        fieldPath: spec.nodeName
  - name: K8S_POD_NAME
    valueFrom:
      fieldRef:
        fieldPath: metadata.name
  - name: K8S_NAMESPACE
    valueFrom:
      fieldRef:
        fieldPath: metadata.namespace
  envFrom:
  - secretRef:
      name: opentelemetry-datadog-credentials
  - secretRef:
      name: opentelemetry-lightstep-credentials
  - secretRef:
      name: opentelemetry-grafanacloud-credentials
  image: ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-contrib:0.78.0
  ingress:
    route: {}
  mode: statefulset
  ports:
  - name: metrics
    port: 8888
    protocol: TCP
    targetPort: 8888
  replicas: 1
  resources:
    limits:
      memory: 3Gi
    requests:
      cpu: "1"
      memory: 2Gi
  serviceAccount: opentelemetry-collector-metrics
  targetAllocator:
    allocationStrategy: consistent-hashing
    enabled: true
    filterStrategy: relabel-config
    image: ghcr.io/open-telemetry/opentelemetry-operator/target-allocator:0.78.0
    prometheusCR:
      enabled: true
    replicas: 1
    serviceAccount: opentelemetry-collector-metrics-targetallocator
  upgradeStrategy: automatic
```

matej-g:
Resolves #1811.

After recent changes (https://github.com/open-telemetry/opentelemetry-operator/pull/1653) to update some TA dependencies (Prometheus Operator and Prometheus), the Prometheus CR watcher's method `LoadConfig` started to fail. This has been traced to the fact that we have previously removed the default scrape interval from the config passed to the method for config generation. This change simply reinstates the default to fix unmarshalling again.

Although for now this has been tested manually, ideally we'll have this tested as part of the unit tests, but we're missing some more changes in order to introduce unit tests. This is part of work in https://github.com/open-telemetry/opentelemetry-operator/pull/1710 which I intend to finish and update soon.

### DIFF
--- a/cmd/otel-allocator/watcher/promOperator.go
+++ b/cmd/otel-allocator/watcher/promOperator.go
@@ -57,7 +57,16 @@ func NewPrometheusCRWatcher(cfg allocatorconfig.Config, cliConfig allocatorconfi
 		monitoringv1.PodMonitorName:     podMonitorInformers,
 	}
 
-	generator, err := prometheus.NewConfigGenerator(log.NewNopLogger(), &monitoringv1.Prometheus{}, true) // TODO replace Nop?
+	// TODO: We should make these durations configurable
+	prom := &monitoringv1.Prometheus{
+		Spec: monitoringv1.PrometheusSpec{
+			CommonPrometheusFields: monitoringv1.CommonPrometheusFields{
+				ScrapeInterval: monitoringv1.Duration("30s"),
+			},
+		},
+	}
+
+	generator, err := prometheus.NewConfigGenerator(log.NewNopLogger(), prom, true) // TODO replace Nop?
 	if err != nil {
 		return nil, err
 	}
--- a/.chloggen/1811-fix-empty-scrape-interval.yaml
+++ b/.chloggen/1811-fix-empty-scrape-interval.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: target allocator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix the empty global scrape interval in Prometheus CR watcher, which causes configuration unmarshalling to fail.
+
+# One or more tracking issues related to the change
+issues: [1811]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:


cmd/otel-allocator/watcher/promOperator.go
```diff
@@ -57,7 +57,16 @@ func NewPrometheusCRWatcher(cfg allocatorconfig.Config, cliConfig allocatorconfi
 		monitoringv1.PodMonitorName:     podMonitorInformers,
 	}
 
-	generator, err := prometheus.NewConfigGenerator(log.NewNopLogger(), &monitoringv1.Prometheus{}, true) // TODO replace Nop?
+	// TODO: We should make these durations configurable
```
jaronoff97, line 60: Do we have an open issue for this? We have the ability to configure the collector's interval for scraping the TA, we should probs have a field for this.

jaronoff97 (approved):
Do we not have an e2e test for the TA that would have caught this?

swiatekm-sumo:
> Do we not have an e2e test for the TA that would have caught this?

If it doesn't cause the allocator to crash, then the kuttl tests lack the tools to detect the problem. In principle we could write a test that starts TA, hits the targets endpoint with curl, and compares the result with some test file, but that sounds hacky. Might be ok if we only have one of these, but if we have more requirements like this, we should consider E2E tests written in Go using e2e_framework.

cmd/otel-allocator/watcher/promOperator.go
```diff
@@ -57,7 +57,16 @@ func NewPrometheusCRWatcher(cfg allocatorconfig.Config, cliConfig allocatorconfi
...
+	prom := &monitoringv1.Prometheus{
+		Spec: monitoringv1.PrometheusSpec{
+			CommonPrometheusFields: monitoringv1.CommonPrometheusFields{
+				ScrapeInterval: monitoringv1.Duration("30s"),
+			},
+		},
+	}
+
+	generator, err := prometheus.NewConfigGenerator(log.NewNopLogger(), prom, true) // TODO replace Nop?
```
achetronic, line 69:
```suggestion
	prometheusInstance := &monitoringv1.Prometheus{
		Spec: monitoringv1.PrometheusSpec{
			CommonPrometheusFields: monitoringv1.CommonPrometheusFields{
				ScrapeInterval: monitoringv1.Duration(cfg.Config.GlobalConfig.ScrapeInterval.String()),
				ScrapeTimeout:  monitoringv1.Duration(cfg.Config.GlobalConfig.ScrapeTimeout.String()),
			},
		},
	}

	generator, err := prometheus.NewConfigGenerator(log.NewNopLogger(), prometheusInstance, true) // TODO replace Nop?
```

I would suggest to fix several things at once. Setting these params this way we fix the issue and get them from the parsed config at the same, so make the configurable

WDYT?

matej-g:
> Do we not have an e2e test for the TA that would have caught this?

I think @swiatekm-sumo gave an exhaustive answer, but I think it would also suffice to have this covered by a unit test for the config loading, it actually would have been caught by tests in https://github.com/open-telemetry/opentelemetry-operator/pull/1710, which adds tests for Prometheus CR watcher (if it was already merged).

cmd/otel-allocator/watcher/promOperator.go
matej-g: I think this is a good suggestion @achetronic. WDYT @jaronoff97 @swiatekm-sumo? We could then directly address https://github.com/open-telemetry/opentelemetry-operator/pull/1822#discussion_r1226748692. Now I wonder why it was configurable from the beginning. And in case the intervals are empty we would simply fall back to a default.

swiatekm-sumo:
To be clear, I do think it's valid to check for this in an E2E test. My point was that it'd require significantly more work, as our current test setup doesn't lend itself to these kinds of checks.

cmd/otel-allocator/watcher/promOperator.go
achetronic: @matej-g 
I was wondering if target allocator needs to be configurable as its responsibility is only scrape ServiceMonitor/PodMonitor resources to create like the half of config related to scrape_configs, right? the rest of the file is created by the operator itself. Am I wrong?

cifuab:
Hi guys i have same issue, its possible to release fix soon?

cmd/otel-allocator/watcher/promOperator.go
jaronoff97: I'm a fan of this, however, I think we should add these as fields for the TA rather than use the prometheus configuration's global config. First, some users may not be setting a prometheus configuration for their TA (a valid setup) and it would be a bit confusing to need to set only the global config and nothing else. Second, I think given this is for the TA this would be better served as a CLI flag for the TA that we add support for in the collector's CRD. 

Given ^^ I think we should do that in a follow-up and not couple those changes in with this PR.

achetronic: @jaronoff97 Hey, I'm completely new to this project :) 

Could you describe the approach a bit better? I would like to collaborate to fix this ASAP as it is affecting our setup

jaronoff97: Hey no worries! Yeah I'd love to get this PR in today. I'm proposing that rather than using the global config which is embedded within the scrape config for the target allocator, we add new fields to the target allocator and use those directly. I prefer this approach because if someone doesn't have a scrape config set they could still define a global scrape config and scrape interval. I'd like to decouple these changes into separate PRs, however, because adding more changes to this one would increase it's scope which would require more reviewing. If you'd like to chat more or even pair on these changes, reach out in our slack channel (#otel-operator) in the CNCF slack :)

Aneurysm9 (approved):

cmd/otel-allocator/watcher/promOperator.go
achetronic: Hey, thank you for merging :) Of course i will reach some of you in slack

PR #1852, opened 2023-06-18
Nginx autoinstrumentation image

chrlic:
This is the first of several PR's to implement autoinstrumentation for Nginx
This one build the image with instrumentation libraries

next will follow
- Instrumentation crd definitions for Nginx
- Implementation of instrumentation logic itself + all kinds of tests


autoinstrumentation/nginx/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+WORKDIR /opt/opentelemetry
+
+RUN wget https://github.com/open-telemetry/opentelemetry-cpp-contrib/releases/download/webserver%2Fv$version/opentelemetry-webserver-sdk-x64-linux.tgz
+RUN mkdir agent
+RUN tar -xvf opentelemetry-webserver-sdk-x64-linux.tgz -C agent
+
+############################
+# STEP 2 build streamlined image
+############################
+FROM alpine:latest
```
pavolloffay, line 19: does this need to be alpine? Could we change it to `from scratch`?

autoinstrumentation/nginx/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+# STEP 1 download the webserver agent 
+############################
+FROM alpine:latest as agent
+
+ARG version
+
+RUN mkdir /opt/opentelemetry
+WORKDIR /opt/opentelemetry
+
+RUN wget https://github.com/open-telemetry/opentelemetry-cpp-contrib/releases/download/webserver%2Fv$version/opentelemetry-webserver-sdk-x64-linux.tgz
```
pavolloffay, line 12: The image seems to be similar to what we already have for apache HTTPd https://github.com/open-telemetry/opentelemetry-operator/blob/main/autoinstrumentation/apache-httpd/Dockerfile#L12 

Is there any reason why we should maintain another one?

autoinstrumentation/nginx/Dockerfile
chrlic: interesting idea - i think it has to have the busybox at least - let me give it a try.

chrlic: if I build an image with the instrumentation libraries and a go-based utility which does all the instrumentation logic, which runs without any dependencies in `from scratch` built image - would that be ok?

autoinstrumentation/nginx/Dockerfile
chrlic: Ok. I'm going to use the current one then and if anything prevents using the same image in the future, then I'll handle it accordingly.

I'm going to close this PR and update #1853 with a comment that the apache image is used for Nginx, too.

PR #1860, opened 2023-06-21
Only admit configurations where Target Allocator has targets

### LINKED ISSUE #1859: [target_allocator] Only admit configurations where targets can exist
Target Allocator gets configuration from two sources:

* scrape configs defined in prometheus receiver configuration in the Collector config
* Prometheus CRs

Right now, you can pass it a configuration that contains neither of these, and it will happily accept it and serve no targets.

During a recent SIG discussion, we've decided that while the Target Allocator itself should accept a configuration like this, we should reject Collector CRDs which would result in this kind of state.

swiatekm-sumo:
Fixes #1859. I've also extracted some common logic in the adapters package to a separate function.

### DIFF
--- a/apis/v1alpha1/opentelemetrycollector_webhook_test.go
+++ b/apis/v1alpha1/opentelemetrycollector_webhook_test.go
@@ -207,9 +207,9 @@ func TestOTELColValidatingWebhook(t *testing.T) {
     endpoint: "0.0.0.0:12346"
   prometheus:
     config:
-      scrape_config:
-        job_name: otel-collector
-        scrape_interval: 10s
+      scrape_configs:
+        - job_name: otel-collector
+          scrape_interval: 10s
   jaeger/custom:
     protocols:
       thrift_http:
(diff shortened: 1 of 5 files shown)


apis/v1alpha1/opentelemetrycollector_webhook_test.go
```diff
@@ -207,9 +207,9 @@ func TestOTELColValidatingWebhook(t *testing.T) {
     endpoint: "0.0.0.0:12346"
   prometheus:
     config:
-      scrape_config:
-        job_name: otel-collector
-        scrape_interval: 10s
+      scrape_configs:
```
jaronoff97, line 210: did this not work previously?

jaronoff97 (approved):
just one question

apis/v1alpha1/opentelemetrycollector_webhook_test.go
swiatekm-sumo: It was used in admission tests that (until now) didn't care about the actual contents of the Prometheus receiver config, so it did work.

VineethReddy02 (approved):

PR #1864, opened 2023-06-22
Explicitly set kind version when running tests in CI

swiatekm-sumo:
Up until now we depended on the `kind` version shipped by Github, which opened us up to unexpected breakage whenever they upgraded. This has happened recently with an upgrade from `0.19.0` to `0.20.0`.

This change:
- explicitly sets the kind version in CI
- updates Node images for kind 0.20.0
- uses kind 0.17.0 to run tests for K8s <1.21, which kind 0.20.0 doesn't support out of the box
- updates kuttl to the latest released version

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -10,6 +10,10 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
+env:
+  CURRENT_KIND_VERSION: 0.20.0
+  LEGACY_KIND_VERSION: 0.17.0  # for K8s versions latest kind doesn't support anymore  
+
 jobs:
   e2e-tests:
     name: End-to-end tests
@@ -28,11 +32,16 @@ jobs:
          - e2e-autoscale
 
     steps:
-
+  
     - name: Set up Go
       uses: actions/setup-go@v4
       with:
         go-version: "1.20"
+
+    - name: Setup kind
+      env:
+        KIND_VERSION: ${{ matrix.kube_version >= 1.21 && env.KIND_VERSION || env.LEGACY_KIND_VERSION }}
+      run: go install sigs.k8s.io/kind@v${KIND_VERSION}
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@v3
--- a/kind-1.19.yaml
+++ b/kind-1.19.yaml
@@ -2,4 +2,4 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
 - role: control-plane
-  image: kindest/node:v1.19.16@sha256:707469aac7e6805e52c3bde2a8a8050ce2b15decff60db6c5077ba9975d28b98
+  image: kindest/node:v1.19.16@sha256:476cb3269232888437b61deca013832fee41f9f074f9bed79f57e4280f7c48b7
--- a/kind-1.20.yaml
+++ b/kind-1.20.yaml
@@ -2,4 +2,4 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
 - role: control-plane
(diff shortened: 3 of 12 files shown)


kind-1.19.yaml
```diff
@@ -2,4 +2,4 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
 - role: control-plane
-  image: kindest/node:v1.19.16@sha256:707469aac7e6805e52c3bde2a8a8050ce2b15decff60db6c5077ba9975d28b98
+  image: kindest/node:v1.19.16@sha256:476cb3269232888437b61deca013832fee41f9f074f9bed79f57e4280f7c48b7
```
pavolloffay, line 5: If we are dropping testing for 1.19 and 1.20 then we can remove this file.

kind-1.20.yaml
```diff
@@ -2,4 +2,4 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
 - role: control-plane
-  image: kindest/node:v1.20.15@sha256:d67de8f84143adebe80a07672f370365ec7d23f93dc86866f0e29fa29ce026fe
+  image: kindest/node:v1.20.15@sha256:a32bf55309294120616886b5338f95dd98a2f7231519c7dedcec32ba29699394
```
pavolloffay, line 5: If we are dropping testing for 1.19 and 1.20 then we can remove this file.

kind-1.19.yaml
swiatekm-sumo: Instead of dropping them, we now use an older kind version.

TylerHelmuth (approved):

.github/workflows/e2e.yaml
```diff
@@ -10,6 +10,10 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
+env:
+  CURRENT_KIND_VERSION: 0.20.0
+  LEGACY_KIND_VERSION: 0.17.0  # for K8s versions latest kind doesn't support anymore  
+
```
jaronoff97, line 16: Instead of adding these to the environment like this, we could use the matrix.include functionality of github actions. https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations

This would let us specify exactly KIND version each kube version should be using.

jaronoff97 (commented):
Just one thought on how to make this a bit cleaner without env vars.

pavolloffay (approved):
Nice  fix for the CI 🥳 

LGTM, but @jaronoff97 has a good comment :)

.github/workflows/e2e.yaml
swiatekm-sumo: I considered doing that, but given that:
1. There would only ever be two kind versions
2. There's less duplication this way, you only change the kind version in one place if you want to upgrade

I decided to do it this way. I don't have a strong opinion on it though, the matrix approach is fine.

pavolloffay:
@swiatekm-sumo should be a similar fix applied to a local setup of kind?

swiatekm-sumo:
> @swiatekm-sumo should be a similar fix applied to a local setup of kind?

I think so, but I didn't want to hold this PR up with even more changes. We should treat kind (and kuttl as well) the same way as we treat other local tools - have a make target that installs them to `LOCALBIN`, and use that binary throughout the Makefile.

TylerHelmuth:
I view this PR as a temporary solution until we drop 1.19 (and other old versions) entirely.  I'm in favor of getting this merged to unblock other work and then quickly bumping our min k8s version.

pavolloffay:
I am merging this to fix the CI, additional fixes can be submitted in follow up PRs

pavolloffay:
> I view this PR as a temporary solution until we drop 1.19 (and other old versions) entirely

@TylerHelmuth why do you see this as a temporary solution?

TylerHelmuth:
@pavolloffay I see the permanent solution to be bumping our minimum supported k8s version which I expect to correlate with being able to use a single kind version.

PR #1850, opened 2023-06-16
[chore] Add functional test for OTLP metrics, traces collection and get the metrics,traces to supported Prometheus stack. 

IshwarKanse:
Add test case for OTEL collector collection OTLP metrics and forwarding to a supported Prometheus stack. 

```
2023/06/16 18:06:27 kutt-test config testdirs is overridden with args: [ tests/functional/ ]
=== RUN   kuttl
    harness.go:462: starting setup
    harness.go:252: running tests using configured kubeconfig.
    harness.go:275: Successful connection to cluster at: https://REDACTED:6443
    harness.go:360: running tests
    harness.go:73: going to run test suite with timeout of 150 seconds for each step
    harness.go:372: testsuite: tests/functional/ has 1 tests
=== RUN   kuttl/harness
=== RUN   kuttl/harness/otlp-metrics
=== PAUSE kuttl/harness/otlp-metrics
=== CONT  kuttl/harness/otlp-metrics
    logger.go:42: 18:06:31 | otlp-metrics | Ignoring check_metrics.sh as it does not match file name regexp: ^(\d+)-(?:[^\.]+)(?:\.yaml)?$
    logger.go:42: 18:06:31 | otlp-metrics | Ignoring check_traces.sh as it does not match file name regexp: ^(\d+)-(?:[^\.]+)(?:\.yaml)?$
    logger.go:42: 18:06:31 | otlp-metrics | Creating namespace: kuttl-test-clear-wolf
    logger.go:42: 18:06:31 | otlp-metrics/0-install-jaeger | starting test step 0-install-jaeger
    logger.go:42: 18:06:33 | otlp-metrics/0-install-jaeger | Project:/otel-metrics created
    logger.go:42: 18:06:33 | otlp-metrics/0-install-jaeger | Jaeger:otel-metrics/jaeger-production created
    logger.go:42: 18:08:16 | otlp-metrics/0-install-jaeger | test step completed 0-install-jaeger
    logger.go:42: 18:08:16 | otlp-metrics/1-workload-monitoring | starting test step 1-workload-monitoring
    logger.go:42: 18:08:18 | otlp-metrics/1-workload-monitoring | ConfigMap:openshift-monitoring/cluster-monitoring-config created
    logger.go:42: 18:08:34 | otlp-metrics/1-workload-monitoring | test step completed 1-workload-monitoring
    logger.go:42: 18:08:34 | otlp-metrics/2-otel-metrics-collector | starting test step 2-otel-metrics-collector
[...]

PR #1865, opened 2023-06-24
[chore] Add Dockerfile for OpenShift containerized test execution on Prow CI.

IshwarKanse:
We will be running end-to-end tests for OpenTelemetry Operator in our [OpenShift release](https://github.com/openshift/release) using Prow CI jobs. These jobs will encompass various cluster configurations, and the Dockerfile will containerize the tests for execution within the Prow CI.

### DIFF
--- a/tests/e2e-openshift/Dockerfile
+++ b/tests/e2e-openshift/Dockerfile
@@ -1,0 +1,22 @@
+# The Dockerfile's resulting image is purpose-built for executing OpenTelemetry Operator e2e tests within the OpenShift release (https://github.com/openshift/release) using Prow CI. 
+
+FROM golang:1.20
+
+# Copy the repository files
+COPY . /tmp/opentelemetry-operator
+
+WORKDIR /tmp
+
+# Install kuttl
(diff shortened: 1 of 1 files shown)


IshwarKanse:
@TylerHelmuth  The CI job expects the Dockerfile to be in the source repository. Will it be fine if we move the file to /tests/e2e-openshift/ once this PR gets merged https://github.com/open-telemetry/opentelemetry-operator/pull/1850

IshwarKanse:
Hi @TylerHelmuth, we're in the process of integrating CI jobs for OpenTelemetry using Prow CI. These CI jobs will execute all of our end-to-end (e2e) test cases for the project. To accomplish this, we've created a Dockerfile that builds an image with the necessary binaries required for running the kuttl-based e2e tests.

During the runtime of the CI job, the test executor image is built using the Dockerfile. To ensure better maintainability, we believe it's preferable to keep the Dockerfile in the project repository rather than adding it as a literal in the CI job. To align with this approach, I have relocated the Dockerfile to the "./tests/e2e-openshift" directory, which is more appropriate for its purpose.

pavolloffay (approved):

PR #1866, opened 2023-06-24
Skip service port for scraper receivers

srikanthccv:
A good chunk of receivers in the collector are scraper kind for which we don't need to create a service port.


pkg/collector/parser/receiver.go
```diff
@@ -77,8 +77,49 @@ func IsRegistered(name string) bool {
 var (
 	endpointKey      = "endpoint"
 	listenAddressKey = "listen_address"
+	scraperReceivers = []string{
```
TylerHelmuth, line 80: Let's make `scraperReceivers` a `map[string]bool` so the lookup is is more efficient.

pkg/collector/parser/receiver.go
```diff
@@ -77,8 +77,45 @@ func IsRegistered(name string) bool {
 var (
 	endpointKey      = "endpoint"
 	listenAddressKey = "listen_address"
+	scraperReceivers = map[string]struct{}{
```
jaronoff97, line 80: Should we not include `prometheus` in this list too?

srikanthccv: There is no top-level `endpoint` for Prometheus config. However, it is covered with a note just below the lines of this change. I don't mind moving it to this list.

jaronoff97: yeah i think we should move it to this list given it provides the same functionality as the case in the switch below.

srikanthccv: Updated

jaronoff97:
@srikanthccv looks like there's a linting failure

srikanthccv:
It should be fixed now.

TylerHelmuth (approved):

pavolloffay (approved):

PR #1870, opened 2023-06-26
Add ability to set init containers

### LINKED ISSUE #1684: Support specifying additional containers for OpenTelemetryCollector podspec
The primary motivation for this request is to add additional sidecars to the otel-collector pods that are controlled by the operator.

One use-case is for if you want to tell the collector to export to a sidecar that acts as an authenticated proxy (in case the ability to specify static headers in the otel-collector otlp export config is insufficient.) I'm sure there are other use cases that would apply in the future.

The Prometheus operator [has support for a `containers` field in the spec](https://prometheus-operator.dev/docs/operator/api/#monitoring.coreos.com/v1.Prometheus), which allows adding or even overriding base container definitions. I'm not sure we have to go as far as that, but having some sort of `additionalContainers` would be nice. For completeness, `initContainers` support may also be helpful for supporting different flavors of deployment environment.

jaronoff97:
Closes #1684 

Enables the setting of init containers for the OTC CRD for statefulsets, daemonsets, deployments _and_ sidecars!


TylerHelmuth (commented):
Is there a reasonable e2e test we could add for this?

jaronoff97:
@TylerHelmuth i was thinking about it, but i wasn't sure what to set as the init container. Any thoughts?

jaronoff97:
Decided to just do an echo, this will just confirm that the init container appears as expected and the operator accepts the spec.

tests/e2e/smoke-init-containers/00-install.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+    exporters:
+      logging:
+
+    service:
+      pipelines:
+        traces:
+          receivers: [jaeger,otlp]
+          processors: []
+          exporters: [logging]
```
TylerHelmuth, line 29:
```suggestion
          exporters: [logging]

```

TylerHelmuth (approved):

srikanthccv:
I am curious if the target allocator deployment could use this?

jaronoff97:
@srikanthccv That's a good idea, but probably better done in a separate issue w/ more explanation for its purpose for that pod.

srikanthccv:
Yes, it can be independent

srikanthccv (approved):

PR #1874, opened 2023-06-27
Create ServiceMonitors to gather metrics from the OTEL Operands

### LINKED ISSUE #1768: Support collecting metrics from instances
The collector instances deployed by the OpenTelemetry Operator expose internal metrics. We can collect those metrics to get more information about how our system is behaving.

[I wrote this proposal to implement the feature](https://docs.google.com/document/d/1bUiwgnkN3rI0Z_vm9_cAECdCVtqtNN8QQ4UJcnenKxE/edit?usp=sharing).

iblancasa:
Resolves #1768


pkg/collector/reconcile/servicemonitor.go
```diff
@@ -0,0 +1,170 @@
...
+	if err := deleteServiceMonitors(ctx, params, desired); err != nil {
+		return fmt.Errorf("failed to reconcile the service monitors to be deleted: %w", err)
+	}
+
+	return nil
+}
+
+func desiredServiceMonitors(_ context.Context, params Params) []monitoringv1.ServiceMonitor {
+	col := params.Instance
+	return []monitoringv1.ServiceMonitor{
```
TylerHelmuth, line 57: I don't love the idea of the OpenTelemetry Operator being able to create objects that depend on other CRDs being preset, it feels like an opportunity for our users to get into weird situations.

jaronoff97: this also creates a direct dependency on the prometheus operator from ours which concerns me. Right now we have a dependency in the TA which makes sense as it is the component pulling down that functionality, but I worry what happens with this added dependency.

iblancasa: > I don't love the idea of the OpenTelemetry Operator being able to create objects that depend on other CRDs being preset

From the conversations we had during the 25 May SIG meeting, I understood it was OK to implement the proposal described in #1768.

> it feels like an opportunity for our users to get into weird situations.

What situations? I would be happy to check them and apply a fix if needed.

> I agree, I think this opens up some issues if we make the assumption that every operator user wants to use service monitors. 

I don't think this will be a problem since everything is protected with a feature flag. If the feature flag is not enabled, the Service Monitors will not be created. Also, you need to set to true the feature in the OpenTelemetry Collector instance.

> This is a good topic for our next SIG meeting.

I'll not be able to attend since I'll be on vacation next week. But I'll be happy to talk about it in another moment.

pavolloffay: We have discussed this on the SIG meeting that the operator will create Service/Pod monitors for operands with the assumption that the functionality will be put behind a feature flag. The implementation has to be flexible to allow OTLP in the future (once the collector will report OTLP metrics).

This work is needed to move the OTEL operator to operator level 4 https://operatorhub.io/operator/opentelemetry-operator. There is a lot of other operators that do the same.

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -345,6 +352,26 @@ type AutoscalerSpec struct {
 	TargetMemoryUtilization *int32 `json:"targetMemoryUtilization,omitempty"`
 }
 
+// MetricsConfigSpec defines a metrics config.
+type MetricsConfigSpec struct {
+	// CreateServiceMonitors specifies if ServiceMonitors should be created for the OpenTelemetry Collector.
```
pavolloffay, line 357: Document here, how this feature is enabled

pkg/collector/reconcile/servicemonitor.go
jaronoff97: ah okay, yeah sorry checked my notes and saw this now. Proceed :) will review now.

main.go
```diff
@@ -42,6 +42,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
```
jaronoff97, line 45: this should be grouped with the above.

main.go
```diff
@@ -71,6 +73,7 @@ func init() {
 
 	utilruntime.Must(otelv1alpha1.AddToScheme(scheme))
 	utilruntime.Must(routev1.AddToScheme(scheme))
+	utilruntime.Must(monitoringv1.AddToScheme(scheme))
```
jaronoff97, line 76: Should we also put this behind the flag?

jaronoff97 (approved):
The approach here makes sense, could you also add an e2e test please?

main.go
iblancasa: When this code is executed, the flag value is unknown.

tests/e2e/prometheus-config-validation/03-promreceiver-nopromconfig.yaml
```diff
@@ -7,7 +7,7 @@ spec:
   mode: statefulset
   targetAllocator:
     enabled: true
-    image: "local/opentelemetry-operator-targetallocator:e2e"
+    image: "ghcr.io/iblancasa/opentelemetry-operator/target-allocator:v0.81.0-14-g27e2a7e"
```
jaronoff97, line 10: Shouldn't we be using the local image here?

iblancasa: Sorry. I added changes that should not be in the PR. Fixing...

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -365,6 +373,27 @@ type AutoscalerSpec struct {
...
+	// The operator.observability.prometheus feature gate must be enabled to use this feature.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Create ServiceMonitors for OpenTelemetry Collector"
+	CreateServiceMonitors bool `json:"createServiceMonitors,omitempty"`
```
TylerHelmuth, line 384: Is `CreateServiceMonitors` generic enough to handle the switch to the Collector when the time comes?  It would be nice if our Config API doesn't need any breaking changes when we can replace Prometheus with the Collector's exporting capabilities.

controllers/opentelemetrycollector_controller.go
```diff
@@ -164,6 +166,11 @@ func NewReconciler(p Params) *OpenTelemetryCollectorReconciler {
 				"ingresses",
 				true,
 			},
+			{
+				reconcile.ServiceMonitors,
+				"service monitors",
+				true,
+			},
```
TylerHelmuth, line 173: Will this task cause problems when the feature gate is `false`?

pkg/collector/reconcile/servicemonitor.go
```diff
@@ -0,0 +1,170 @@
...
+				Name:      naming.ServiceMonitor(col),
+				Labels: map[string]string{
+					"app.kubernetes.io/name":       naming.ServiceMonitor(params.Instance),
+					"app.kubernetes.io/instance":   fmt.Sprintf("%s.%s", params.Instance.Namespace, params.Instance.Name),
+					"app.kubernetes.io/managed-by": "opentelemetry-operator",
+				},
+			},
+			Spec: monitoringv1.ServiceMonitorSpec{
+				Endpoints: []monitoringv1.Endpoint{{
+					Port: "monitoring",
```
TylerHelmuth, line 70: Where is this port being defined?  Does this handle situations where someone manually configures the collector's metrics port to be something other than the default?

apis/v1alpha1/opentelemetrycollector_types.go
iblancasa: Ok I agree the name is not as generic as it should be. How about `EnableMetrics`?

controllers/opentelemetrycollector_controller.go
iblancasa: The first thing to be checked during that task is if the feature flag is enabled:
```go
if !params.Instance.Spec.Observability.Metrics.CreateServiceMonitors || !featuregate.PrometheusOperatorIsAvailable.IsEnabled() {
```

pkg/collector/reconcile/servicemonitor.go
iblancasa: It is defined here:
https://github.com/open-telemetry/opentelemetry-operator/blob/56f5de94dd3f1f83ac7dbdba3b8b4a6379317b48/pkg/collector/reconcile/service.go#L197-L200

It cannot be modified (right now). But I think it could be an interesting follow-up task: to allow the personalization of the metrics port. What do you t hink?

apis/v1alpha1/opentelemetrycollector_types.go
TylerHelmuth: I like that, and that will flex well into any metric generation solution.

pkg/collector/reconcile/servicemonitor.go
iblancasa: Ok, I see your point.
Yes, the metrics become unreachable. If we want to address that use case, I would do it as part of a different issue since it would involve changes in https://github.com/open-telemetry/opentelemetry-operator/blob/56f5de94dd3f1f83ac7dbdba3b8b4a6379317b48/pkg/collector/reconcile/service.go#L197-L200

And look for a way to get the address where the metrics are exposed. What do you think?

TylerHelmuth: Doing it as part of another issue is good.  I kinda feel like we need to solve that issue before we introduce this functionality.

iblancasa: Created #1931

jaronoff97:
@TylerHelmuth mind giving one more review?

TylerHelmuth:
@jaronoff97 I am still concerned about https://github.com/open-telemetry/opentelemetry-operator/pull/1874#discussion_r1264034952. It is bad  that users can't modify their metrics endpoint today, I am hesitant to add on more functionality dependent on the metrics endpoint without the flexibility to configure it.  While I recognize we could release this first and add support for configuration later, it feels like an implementation that drives the Operator to Level 4 should also allow the users to configure the managed object (in this instance the Collector) however they need.

pkg/featuregate/featuregate.go
```diff
@@ -56,6 +56,12 @@ var (
 		"operator.collector.rewritetargetallocator",
 		featuregate.StageAlpha,
 		featuregate.WithRegisterDescription("controls whether the operator should configure the collector's targetAllocator configuration"))
+
+	// PrometheusOperatorIsAvailable is the feature gate that enables features associated to the Prometheus Operator.
+	PrometheusOperatorIsAvailable = featuregate.GlobalRegistry().MustRegister(
+		"operator.observability.prometheus",
+		featuregate.StageAlpha,
+		featuregate.WithRegisterDescription("enables features associated to the Prometheus Operator"))
```
TylerHelmuth, line 64: Please add a `featuregate.FromVersion` that marks the Operator version this feature gate was added. (This was a rather late concept in the featuregate package and we should go back and fix our other gates to include it)

jaronoff97:
I see, in that case let's block merging this until we have #1931 closed

iblancasa:
> I see, in that case let's block merging this until we have #1931 closed

Even without merging this PR, the problem is still there because the service created by the operator will be pointing to a port that does nothing. I'll work on #1931 but, TBH, I don't think there is a real need to block this PR until #1931 is merged.

.github/workflows/e2e.yaml
```diff
@@ -28,11 +28,10 @@ jobs:
        - "1.19"
        - "1.27"
        group:
-         - e2e e2e-upgrade
+         - e2e e2e-upgrade e2e-prometheuscr
```
TylerHelmuth, line 31: Can we separate `e2e-prometheuscr` into its own section in the matrix?  `e2e e2e-upgrade` already has a noisy neighbor problem and I'd prefer not to add another.

iblancasa: Sure. Just didn't want to separate in case somebody though it was not ok to create a new section.

TylerHelmuth: Luckily we have the CI checks setup to look at the dependent job, so we can now scale this worker job as needed.

TylerHelmuth (approved):

PR #1875, opened 2023-06-28
adds envs handling to target allocator

### LINKED ISSUE #1773: Cannot inject envs to targetAllocator
We have several serviceMonitors, which are using environment variables, e.g:

```
spec:
  endpoints:
  - port: otelcol-metrics
  namespaceSelector:
    matchNames:
    - $(NAMESPACE)
  selector:
    matchLabels:
      sumologic.com/app: fluentd-metrics
      sumologic.com/scrape: "true"
```

It would be nice to add and support `env` in `targetAllocator` so we could seamlessly migrate from Prometheus to OpenTelemetry

KitSutliff:
Fixes #1773


yuriolisa (approved):

pkg/targetallocator/container.go
```diff
@@ -35,7 +35,10 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
 		MountPath: "/conf",
 	}}
 
-	envVars := []corev1.EnvVar{}
+	var envVars = otelcol.Spec.TargetAllocator.Env
+	if otelcol.Spec.TargetAllocator.Env == nil {
+		envVars = []corev1.EnvVar{}
+	}
 
 	envVars = append(envVars, corev1.EnvVar{
```
jaronoff97, line 43: Hm i'm wondering if we should check if this variable already exists before we append it. I think kubernetes will accept both, but i'm wondering if we should not override an existing one? https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#container-v1-core

jaronoff97 (approved):

pkg/targetallocator/container.go
KitSutliff: Good catch!  Dropped in an implementation and added a test for it.  Please let me know if there's anything else!

jaronoff97 (approved):

jaronoff97:
Thank you for your contribution @KitSutliff :D

PR #1888, opened 2023-07-03
Add managementState field to collector crd spec

### LINKED ISSUE #1881: Add a managementState field to the Collector CRD to disable the reconciliation
The idea of this task is to add a new field in the CRD called `managementState`. It would be placed under `.spec.managementState` for the OpenTelemetry Collector instances.

Possible values:
* `Managed` _(default value)_: the reconciliation works currently does
* `Unmanaged`: disables the reconciliation

Disabling the reconciliation can be useful in different use cases:
- Debugging
- Edit the objects created by the operator without losing the changes (because of the reconciliation)
- Keep instances without upgrading (if, for whatever reason we need to upgrade the operator and all the instances except some of them)

This feature is something we implemented for the [Tempo Operator](https://github.com/grafana/tempo-operator/blob/main/docs/operator/api.md#tempostackspec--tempo-grafana-com-v1alpha1-tempostackspec-) and it is also implemented in the  [Loki Operator](https://github.com/grafana/loki/blob/main/operator/docs/operator/api.md#lokistackspec--loki-grafana-com-v1-lokistackspec-).

avadhut123pisal:
closes #1881


pkg/collector/upgrade/upgrade.go
```diff
@@ -55,6 +55,12 @@ func (u VersionUpgrade) ManagedInstances(ctx context.Context) error {
 	for i := range list.Items {
 		original := list.Items[i]
 		itemLogger := u.Log.WithValues("name", original.Name, "namespace", original.Namespace)
+
+		if original.Spec.ManagementState == v1alpha1.ManagementStateUnmanaged {
```
pavolloffay, line 59: We should book a ticket to run the upgrade as part of the reconciliation for instances that will be switched from unmanaged to managed after the upgrade

tests/e2e/managed-reconcile/01-disable-reconciliation.yaml
```diff
@@ -0,0 +1,15 @@
...
+---
+apiVersion: opentelemetry.io/v1alpha1
+kind: OpenTelemetryCollector
+metadata:
+  name: simplest
+spec:
+  # The deployments use the simplest-collector Service Account. Changing the
+  # SA here and setting the managementState to Unmanaged will make the operator
+  # to not change the original service account name
+  serviceAccount: fake-sa
```
pavolloffay, line 14: The unmanaged feature is used differently. 

e.g. user would switch to the unmanaged state. Changed objects managed by the operator directly (without changing the fields in the CR).

pavolloffay (commented):
LGTM, apart the comment in the e2e test.

The e2e test could be improved to verify the full loop: managed->unmanaged->managed

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -21,6 +21,21 @@ import (
...
+//
+// +kubebuilder:validation:Enum=Managed;Unmanaged
+type ManagementStateType string
+
+const (
+	// ManagementStateManaged when the OpenTelemetryCollector custom resource should be
+	// reconciled by the operator.
+	ManagementStateManaged ManagementStateType = "Managed"
+
+	// ManagementStateUnmanaged when the TempoStack custom resource should not be
```
jaronoff97, line 34: Tempo stack => collector

tests/e2e/managed-reconcile/01-disable-reconciliation.yaml
avadhut123pisal: Done.

pkg/collector/upgrade/upgrade.go
avadhut123pisal: > We should book a ticket to run the upgrade as part of the reconciliation for instances that will be switched from unmanaged to managed after the upgrade

#1890

tests/e2e/managed-reconcile/01-assert.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: apps/v1
```
pavolloffay, line 1: In this assertion I would check if the collector config map (check if the jaeger is there - it was not reverted by the operatr)

tests/e2e/managed-reconcile/02-assert.yaml
```diff
@@ -0,0 +1,58 @@
+apiVersion: apps/v1
```
pavolloffay, line 1: This assertion should check the collector config map - check that the jaeger was reverted

tests/e2e/managed-reconcile/01-assert.yaml
avadhut123pisal: Done

tests/e2e/managed-reconcile/02-assert.yaml
avadhut123pisal: Done

iblancasa (approved):
Thanks for implementing it

iblancasa:
@pavolloffay @jaronoff97 is there anything pending about this PR?

pavolloffay (approved):

PR #1892, opened 2023-07-05
Fixed minimal target allocator example in readme

### LINKED ISSUE #1891: [target allocator] Minimal example from README does not work
Using the feature flag `operator.collector.rewritetargetallocator` as stated in [README](https://github.com/open-telemetry/opentelemetry-operator#target-allocator-config-rewriting) only works if prometheus receiver config is valid. A minimal example with the following (cited from readme) does not work with an error:
`admission webhook "vopentelemetrycollectorcreateupdate.kb.io" denied the request: the OpenTelemetry Spec Prometheus configuration is incorrect, prometheus property in the configuration doesn't contain valid prometheus`

``` yaml
  config: |
    receivers:
      prometheus:

    exporters:
      logging:

    service:
      pipelines:
        metrics:
          receivers: [prometheus]
          processors: []
          exporters: [logging]
```

What would be a valid minimal example when using the feature flag?

EDIT: Found the solution and updated the README in #1892

cmergenthaler:
Fixes minimal example for target allocator when using feature flag `operator.collector.rewritetargetallocator`. 

Fixes #1891

### DIFF
--- a/README.md
+++ b/README.md
@@ -549,6 +549,7 @@ spec:
   config: |
     receivers:
       prometheus:
+        config:
 
     exporters:
       logging:


swiatekm-sumo:
I consider the fact that we don't accept the stated example a bug, but in the meantime, this is a good change. :+1:

swiatekm-sumo (approved):

jaronoff97 (approved):

PR #1894, opened 2023-07-05
Install golangci-lint in the local tools dir

swiatekm-sumo:
Golangci-lint output can vary significantly between versions, and each project uses a different one. It makes for a better local dev experience if we explicitly install the version we want, just as we do with other local tools.

I've also made the CI use this for consistency's sake.

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -29,9 +29,6 @@ jobs:
   lint:
     name: Code standards (linting)
     runs-on: ubuntu-20.04
-    strategy:
-      matrix:
-        workdir: [".", "./cmd/otel-allocator", "./cmd/operator-opamp-bridge"]
     steps:
     - name: Set up Go
       uses: actions/setup-go@v4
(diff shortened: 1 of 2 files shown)


pavolloffay (approved):

.github/workflows/continuous-integration.yaml
```diff
@@ -29,9 +29,6 @@ jobs:
   lint:
     name: Code standards (linting)
     runs-on: ubuntu-20.04
-    strategy:
-      matrix:
-        workdir: [".", "./cmd/otel-allocator", "./cmd/operator-opamp-bridge"]
```
pavolloffay, line 34: why was this removed? 

IIRC this was assuming the lint runs per all subprojects

swiatekm-sumo: The lint target in our Makefile already does this, so it seems redundant unless we want to have each subproject check show up as a separate CI job.

pavolloffay:
created https://github.com/open-telemetry/community/issues/1579 to unblock merging.

jaronoff97 (approved):

Issue #1717, opened 2023-05-05
Is there a way with the OpenTelemetry Operator Auto-Instrumentation to instrument multiple languages in the same Pod


jack78901:
I have an application running in Kubernetes that has a pod that has multiple containers. Each container performs part of a task. Namely, one is the API (written in node.js), a backend transport (written in java), and a drive connector (written in C).

When attempting to add multiple annotations for instrumentation, we get the error:

44s         Warning   FailedCreate       statefulset/my-pod-with-multiple-containers   create Pod my-pod-with-multiple-containers-1 in StatefulSet my-pod-with-multiple-containers failed error: Pod “my-pod-with-multiple-containers-1” is invalid: spec.containers[1].volumeMounts[2].mountPath: Invalid value: "/otel-auto-instrumentation": must be unique

I know that with OpenTelemetry, you can annotate a pod to do auto-instrumentation, such as:

apiVersion: apps/v1
kind: Statefulset
metadata:
  name: my-deployment-with-multiple-containers
spec:
  selector:
    matchLabels:
      app: my-pod-with-multiple-containers
  replicas: 1
  template:
    metadata:
      labels:
        app: my-pod-with-multiple-containers
      annotations:
        instrumentation.opentelemetry.io/inject-java: "true"
        instrumentation.opentelemetry.io/container-names: "myapp,myapp2"
    spec:
      containers:
      - name: myapp
        image: myImage1
      - name: myapp2
        image: myImage2
      - name: myapp3
        image: myImage3
The above (taken from https://github.com/open-telemetry/opentelemetry-operator/blob/main/README.md#multi-container-pods), myapp and myapp2 would get java auto-instrumentation injected.

We tried doing:

      annotations:
        instrumentation.opentelemetry.io/inject-java: "true"
        instrumentation.opentelemetry.io/container-names: "myapp"
        instrumentation.opentelemetry.io/inject-nodejs: "true" 
        instrumentation.opentelemetry.io/container-names: "myapp2"
    spec:
      containers:
      - name: myapp
        image: myImage1
      - name: myapp2
        image: myImage2
      - name: myapp3
        image: myImage3
We were hoping this would make it so that myapp would get Java instrumentation and myapp2 would get node.js instrumentation.

I was informed to move this here from Stack Overflow: https://stackoverflow.com/questions/75943971/is-there-a-way-with-the-opentelemetry-operator-auto-instrumentation-to-instrumen

[...]

PR #1273, opened 2022-11-24
Set securityContext on injected initContainer

addreas:
Encountered the issue where the injected `initContainer` was rejected because of its securityContext. This PR simply takes the securityContext from the `container` that is targeted and uses that for the `initContainer` as well.

resolves #1084 
resolves #1058

PR #1901, opened 2023-07-07
feat: add support for multi container instrumentation pod

mat-rumian:
Issue: https://github.com/open-telemetry/opentelemetry-operator/issues/1717

**Added:**

1. New feature gate introduced `operator.autoinstrumentation.multi-instrumentation`. When flag is enabled then annotations listed below are used. 

   Functionality:
   - Multi instrumentation support is enabled
   - Old annotation `instrumentation.opentelemetry.io/container-names` is not used
   - Instrumentations can't be mixed on the container level (we prevent that looking for duplicates in provided container names)
   - When multiple instrumentations are used, all of them require to provide container names

    | Case  | Flag Enabled | Flag Disabled |
    | ------------- | ------------- | ---- |
    | single instrumentation, no containers  | :heavy_check_mark: first container instrumented | :heavy_check_mark: first container instrumented |
    | single instrumentation, with containers  | :heavy_check_mark: all listed containers instrumented | :heavy_check_mark:  all listed containers instrumented |
    | mutli-instrumentation, no containers | :heavy_minus_sign: instrumentation not applied | :heavy_minus_sign: not supported |
    | multi instrumentation, with containers | :heavy_check_mark: all listed containers instrumented  | :heavy_minus_sign: not supported |
    
    
    Legend:
    - `single instrumentation` stands for single language instrumentation
    - `multi instrumentation` stands for multiple annotations enabling specific language instrumentation
    - `no containers` stands for lack of container names annotations
    - `with containers` stands for configured container names annotations

1. New language specific containers names annotations added:
    - `instrumentation.opentelemetry.io/java-container-names`
    - `instrumentation.opentelemetry.io/nodejs-container-names`
    - `instrumentation.opentelemetry.io/python-container-names`
    - `instrumentation.opentelemetry.io/dotnet-container-names`
    - `instrumentation.opentelemetry.io/go-container-names`
    - `instrumentation.opentelemetry.io/sdk-container-names`
    - `instrumentation.opentelemetry.io/apache-httpd-container-names`

**Changed:**
- Instrumentation specific `volumes` introduced
- Instrumentation specific  `initContainer` introduced
- Instrumentation specific `volumeMount` introduced

**Fixed:**
- prevents multiple instrumentation injection (duplicated volume) when `operator.autoinstrumentation.multi-instrumentation` feature gate is `disabled`

**Testing:**
- [x] new `e2e-multi-instrumentation` tests added
- [x] old `e2e` and `integration` tests updated 

**Documentation:**
ReadMe updated.


jaronoff97:
Would it be possible to make this a non-breaking change by accepting either `container-names` OR one of the language specific ones?

mat-rumian:
> Would it be possible to make this a non-breaking change by accepting either `container-names` OR one of the language specific ones?

Hey @jaronoff97,
it is possible and it could work as it was working before but I think it will create more confusion for new users and for us to manage new issues/discussions. I believe current users have some experience with operator and annotation name change won't cause them troubles.

jaronoff97:
My worry is that someone who has auto instrumentation enabled is going to upgrade the operator without checking the release notes for a minor release and have all of their traces disappear. I'd be in favor of creating the new functionality as an alpha feature flag and then deprecating the current method. After one or two releases we can move this to beta and then stable at which point we can delete the deprecated functionality. Cc @open-telemetry/operator-maintainers @open-telemetry/operator-approvers

TylerHelmuth:
We definitely can't do this in a single breaking change.  I'd prefer to do it as a non-breaking change.  If that is not possible we should definitely use feature gates.

mat-rumian:
Ok :) Then I will add new feature gate 👍

mat-rumian:
Hey @TylerHelmuth @jaronoff97 ,
I've updated the code and PR description. PR is not introducing any breaking changes. There's only one thing that I'm not sure about `multi-instrumentation`. 

In case of a `pod` with `single container` and `multi-instrumentation enabled` current implementation will try to inject all supported SDKs and mount volumes. I think this can cause many configuration issues (e.g. Python instrumentation uses OTLP/http and all other instrumentations use OTLP/gRPC to export data). This will work when:
- user will provide very accurate configuration thru `Instrumentation` resource - switch all instrumentations from gRPC to HTTP exporter
- we will change default exporter for every instrumentation from gRPC to HTTP

Please let me know what's your thoughts :)

jaronoff97:
@mat-rumian thanks for the updates here, I don't have time to review this today, but will check it out tomorrow!

tslavik:
Copilot:all

tests/e2e/instrumentation-dotnet-multicontainer/01-assert.yaml
```diff
@@ -18,15 +18,15 @@ spec:
     - name: CORECLR_PROFILER
       value: "{918728DD-259F-4A6A-AC2B-B85E1B658318}"
     - name: CORECLR_PROFILER_PATH
-      value: /otel-auto-instrumentation/linux-x64/OpenTelemetry.AutoInstrumentation.Native.so
+      value: /otel-auto-instrumentation-dotnet/linux-x64/OpenTelemetry.AutoInstrumentation.Native.so
```
TylerHelmuth, line 21: I think we should try to implement this functionality without modifying the existing tests.  The new tests should set the feature flag and these tests should not.  I am pretty sure our current test setup is too inflexible to handle that, so we might need changes in another PR, but it would be nice.

hack/modify-test-images.sh
```diff
@@ -5,3 +5,5 @@ SED_BIN=${SED_BIN:-sed}
 ${SED_BIN} -i "s#local/opentelemetry-operator-targetallocator:e2e#${TARGETALLOCATOR_IMG}#g" tests/e2e/smoke-targetallocator/*.yaml
 ${SED_BIN} -i "s#local/opentelemetry-operator-targetallocator:e2e#${TARGETALLOCATOR_IMG}#g" tests/e2e/targetallocator-features/00-install.yaml
 ${SED_BIN} -i "s#local/opentelemetry-operator-targetallocator:e2e#${TARGETALLOCATOR_IMG}#g" tests/e2e/prometheus-config-validation/*.yaml
+
+${SED_BIN} -i "s#local/opentelemetry-operator:e2e#${OPERATOR_IMG}#g" tests/e2e-multi-instrumentation/*.yaml
```
TylerHelmuth, line 9: Can we use the normal operator image? We don't do this for the other instrumentation tests.

pkg/featuregate/featuregate.go
```diff
@@ -50,6 +50,11 @@ var (
 		featuregate.StageBeta,
 		featuregate.WithRegisterDescription("controls whether the operator supports Apache HTTPD auto-instrumentation"))
 
+	EnableMultiInstrumentationSupport = featuregate.GlobalRegistry().MustRegister(
+		"operator.autoinstrumentation.multi-instrumentation",
+		featuregate.StageAlpha,
+		featuregate.WithRegisterDescription("controls whether the operator supports multi instrumentation"))
```
TylerHelmuth, line 56: Please add a `featuregate.FromVersion` option to the feature gate.

pkg/instrumentation/golang.go
```diff
@@ -36,9 +37,17 @@ func injectGoSDK(goSpec v1alpha1.Go, pod corev1.Pod) (corev1.Pod, error) {
...
+	if featuregate.EnableMultiInstrumentationSupport.IsEnabled() {
+		containerNames, ok := pod.Annotations[annotationInjectGoContainersName]
+		if ok && len(strings.Split(containerNames, ",")) > 1 {
+			return pod, fmt.Errorf("go instrumentation cannot be injected into a pod using instrumentation.opentelemetry.io/go-container-names with more than 1 container")
+		}
+	} else {
+		containerNames, ok := pod.Annotations[annotationInjectContainerName]
+		if ok && len(strings.Split(containerNames, ",")) > 1 {
+			return pod, fmt.Errorf("go instrumentation cannot be injected into a pod using instrumentation.opentelemetry.io/container-names with more than 1 container")
+		}
```
TylerHelmuth, line 50: 👍

TylerHelmuth (commented):
I'll give this a deeper review next week but so far its looking good.  This was a very complicated change, thanks for taking it on!

tests/e2e/instrumentation-dotnet-multicontainer/01-assert.yaml
mat-rumian: I'm not sure if splitting this variables using featureflag will be valuable. Another few thousands of lines of the same tests, highly increased build time. I would like to leave it as it is now, and if there will be a need I would do that in separated PR.

pkg/featuregate/featuregate.go
mat-rumian: Should it be `featuregate.WithRegisterFromVersion(VERSION)`? Where `VERSION` is next unreleased so `0.82.0`?

tests/e2e/instrumentation-dotnet-multicontainer/01-assert.yaml
TylerHelmuth: Looking at the PR again I realize now that the things that are changing aren't hidden behind the feature gate.  While _technically_ breaking changes, these are all env vars/paths that the operator is injecting/using and the user should not be depending on.  I agree that it is ok to leave these changes outside of the featuregate.

pkg/instrumentation/helper.go
```diff
@@ -43,3 +50,116 @@ func isAutoInstrumentationInjected(pod corev1.Pod) bool {
 	}
 	return false
 }
+
+// Check if single instrumentation is configured for Pod and return which is configured.
+func isSingleInstrumentationEnabled(insts languageInstrumentations) (bool, string) {
```
TylerHelmuth, line 55: For the new functions that take an `languageInstrumentations` can we add the functions to the struct?  This will prevent the need for reflection.

pkg/instrumentation/podmutator.go
```diff
@@ -42,14 +44,102 @@ type instPodMutator struct {
...
+	count := 0
+	enabledInstrumentation := ""
+	value := reflect.ValueOf(langInsts)
+	for i := 0; i < value.NumField(); i++ {
+		field := value.Field(i).FieldByName("Instrumentation")
+		if !field.IsNil() {
+			count++
+			enabledInstrumentation = value.Type().Field(i).Name
+		}
+	}
```
TylerHelmuth, line 74: Do we have to use reflection?  Can we do `langInsts.Java.Instrumentation == nil` and repeat for each language?  Avoiding reflection will make these functions easier to read/comprehend.

mat-rumian: Sure, I've decided to use reflection to generate less lines of code, but you're right without reflection it's easier to understand what's going on there :)

pkg/instrumentation/podmutator.go
```diff
@@ -42,14 +43,146 @@ type instPodMutator struct {
...
+
+	if count == 1 {
+		return true
+	} else {
+		return false
+	}
+}
+
+// Check if specific containers are provided for configured instrumentation.
+func (langInsts languageInstrumentations) areContainerNamesConfiguredForMultipleInstrumentations() (bool, string) {
```
TylerHelmuth, line 95: Lets return proper `error` (probably using fmt.Errorf) for this function.  Then the ok case can return nil.

pkg/instrumentation/podmutator.go
```diff
@@ -159,22 +292,50 @@ func (pm *instPodMutator) Mutate(ctx context.Context, ns corev1.Namespace, pod c
...
+		insts.Go.Containers = annotationValue(ns.ObjectMeta, pod.ObjectMeta, annotationInjectGoContainersName)
+		insts.ApacheHttpd.Containers = annotationValue(ns.ObjectMeta, pod.ObjectMeta, annotationInjectApacheHttpdContainersName)
+		insts.Sdk.Containers = annotationValue(ns.ObjectMeta, pod.ObjectMeta, annotationInjectSdkContainersName)
+
+		// We check if provided annotations and instrumentations are valid
+		ok, msg := insts.areContainerNamesConfiguredForMultipleInstrumentations()
+		if !ok {
+			logger.V(1).Info("skipping instrumentation injection", "reason", msg)
+			return pod, nil
+		}
```
TylerHelmuth, line 320: I think it is ok to handle this situation as an error since the user has misconfigured something.  We want to make that easy to detect.

pkg/instrumentation/podmutator.go
```diff
@@ -159,22 +292,50 @@ func (pm *instPodMutator) Mutate(ctx context.Context, ns corev1.Namespace, pod c
...
+		insts.Java.Containers = annotationValue(ns.ObjectMeta, pod.ObjectMeta, annotationInjectJavaContainersName)
+		insts.NodeJS.Containers = annotationValue(ns.ObjectMeta, pod.ObjectMeta, annotationInjectNodeJSContainersName)
+		insts.Python.Containers = annotationValue(ns.ObjectMeta, pod.ObjectMeta, annotationInjectPythonContainersName)
+		insts.DotNet.Containers = annotationValue(ns.ObjectMeta, pod.ObjectMeta, annotationInjectDotnetContainersName)
+		insts.Go.Containers = annotationValue(ns.ObjectMeta, pod.ObjectMeta, annotationInjectGoContainersName)
+		insts.ApacheHttpd.Containers = annotationValue(ns.ObjectMeta, pod.ObjectMeta, annotationInjectApacheHttpdContainersName)
+		insts.Sdk.Containers = annotationValue(ns.ObjectMeta, pod.ObjectMeta, annotationInjectSdkContainersName)
+
+		// We check if provided annotations and instrumentations are valid
+		ok, msg := insts.areContainerNamesConfiguredForMultipleInstrumentations()
```
TylerHelmuth, line 316: I may have misunderstood the situation here.

Users should be allowed to enable the feature gate but still use `instrumentation.opentelemetry.io/inject-java: "true"` and provide no container names.

Users should not be allowed to enable the feature gate and use `instrumentation.opentelemetry.io/inject-java: "true"` with both `instrumentation.opentelemetry.io/java-container-names` and `instrumentation.opentelemetry.io/nodejs-container-names`.

If I am reading this logic correctly the first use case is being blocked, is that correct?

mat-rumian: > Users should be allowed to enable the feature gate but still use instrumentation.opentelemetry.io/inject-java: "true" and provide no container names.

User is able to inject single instrumentation without *-container-names when `multi-instrumentation` feature gate is `enabled`.

> Users should not be allowed to enable the feature gate and use instrumentation.opentelemetry.io/inject-java: "true" with both instrumentation.opentelemetry.io/java-container-names and instrumentation.opentelemetry.io/nodejs-container-names.

I believe you were thinking about `instrumentation.opentelemetry.io/java-container-names` and `instrumentation.opentelemetry.io/container-names` - so in case of disabled `multi-instrumentation` only `container-names` annotation is used. When `multi-instrumentation` is enabled then specific for language container-names is used. There're tests in pod mutator related to this scenario.

pkg/instrumentation/podmutator.go
mat-rumian: Yeah, added :)

pkg/instrumentation/podmutator.go
```diff
@@ -42,14 +43,142 @@ type instPodMutator struct {
...
+	instrWithContainers := 0
+	var allContainers []string
+
+	// Check for instrumentations with and without containers.
+	if langInsts.Java.Instrumentation != nil && langInsts.Java.Containers == "" {
+		instrWithoutContainers++
+	} else if langInsts.Java.Instrumentation != nil && langInsts.Java.Containers != "" {
+		allContainers = append(allContainers, langInsts.Java.Containers)
+		instrWithContainers++
+	}
```
TylerHelmuth, line 102: We can use nested conditions so that `langInsts.Java.Instrumentation != nil` is only checked once.  We can probably also extract some duplicate logic from each language's check into a helper function.

pkg/instrumentation/helper.go
```diff
@@ -43,3 +48,34 @@ func isAutoInstrumentationInjected(pod corev1.Pod) bool {
 	}
 	return false
 }
+
+// Look for duplicates in the provided containers.
+func findDuplicatedContainers(ctrs []string) []string {
```
TylerHelmuth, line 53: Let's update this function to return an error.

pkg/instrumentation/podmutator.go
```diff
@@ -42,14 +43,142 @@ type instPodMutator struct {
...
+
+	// Look for mixed multiple instrumentations with and without container names.
+	if instrWithoutContainers > 0 && instrWithContainers > 0 {
+		return false, fmt.Errorf("incorrect instrumentation configuration - please provide container names for all instrumentations")
+		// Look for multiple instrumentations without container names.
+	} else if instrWithoutContainers > 1 && instrWithContainers == 0 {
+		return false, fmt.Errorf("incorrect instrumentation configuration - please provide container names for all instrumentations")
+	} else if instrWithoutContainers == 0 && instrWithContainers == 0 {
+		return false, fmt.Errorf("instrumentation configuration not provided")
```
TylerHelmuth, line 153: Since the conditions always return we don't need `else`.

pkg/instrumentation/podmutator.go
```diff
@@ -42,14 +43,142 @@ type instPodMutator struct {
...
+	if containerDuplicates != nil {
+		return false, fmt.Errorf("duplicated container names detected: %s", containerDuplicates)
+	}
+
+	// Look for mixed multiple instrumentations with and without container names.
+	if instrWithoutContainers > 0 && instrWithContainers > 0 {
+		return false, fmt.Errorf("incorrect instrumentation configuration - please provide container names for all instrumentations")
+		// Look for multiple instrumentations without container names.
+	} else if instrWithoutContainers > 1 && instrWithContainers == 0 {
+		return false, fmt.Errorf("incorrect instrumentation configuration - please provide container names for all instrumentations")
```
TylerHelmuth, line 151: Is this actually an error?  Isn't this the situation where someone has enabled the feature gate and has a pod with a single container to instrument, such as:

```yaml
apiVersion: apps/v1
kind: Deployment
spec:
  ...
  template:
    metadata:
      annotations:
```

The above situation should be allowed.

mat-rumian: Conditions are right. It's `instrWithoutContainers > 1` not `instrWithoutContainers == 1` :) 
The case is allowed - you can see the tests for that: 
- e2e - https://github.com/mat-rumian/opentelemetry-operator/tree/support-multi-cont-instr/tests/e2e-multi-instrumentation/instrumentation-single-instr-first-container
- integration - https://github.com/mat-rumian/opentelemetry-operator/blob/support-multi-cont-instr/pkg/instrumentation/podmutator_test.go#L4378

pkg/instrumentation/podmutator.go
```diff
@@ -42,14 +43,139 @@ type instPodMutator struct {
...
+	var instrWithoutContainers int
+	var instrWithContainers int
+	var allContainers []string
+
+	// Check for instrumentations with and without containers.
+	if langInsts.Java.Instrumentation != nil {
+		instrWithContainers += isInstrWithContainers(langInsts.Java)
+		instrWithoutContainers += isInstrWithoutContainers(langInsts.Java)
+		allContainers = append(allContainers, langInsts.Java.Containers)
+	}
```
TylerHelmuth, line 101: I like your idea here, I find this much easier to comprehend.

pkg/instrumentation/podmutator.go
TylerHelmuth: Alright now I'm following, thank you!

pkg/instrumentation/golang.go
TylerHelmuth, line 50: We can move `containerNames` and `ok` outside the condition, use the if-else conditions to set the values, and then do a single `if ok and len` check.

pkg/instrumentation/sdk.go
```diff
@@ -51,107 +51,154 @@ type sdkInjector struct {
...
-			// Ensure that after all the env var coalescing we have a value for OTEL_GO_AUTO_TARGET_EXE
-			idx := getIndexOfEnv(pod.Spec.Containers[len(pod.Spec.Containers)-1].Env, envOtelTargetExe)
-			if idx == -1 {
-				i.logger.Info("Skipping Go SDK injection", "reason", "OTEL_GO_AUTO_TARGET_EXE not set", "container", pod.Spec.Containers[index].Name)
-				pod = origPod
+
+		goContainers := insts.Go.Containers
+
+		// Go instrumentation supports only single container instrumentation. Loop here should be replaced/improved.
+		for _, container := range strings.Split(goContainers, ",") {
```
TylerHelmuth, line 140: Can this section work off the assumption that `goContainer` is exactly 1 container? We could take the value as is.

mat-rumian: Yeah, you're right :)

TylerHelmuth (approved):
@mat-rumian thanks for taking this, it was a big undertaking!

jaronoff97 (approved):
WOW! This is incredible. Great work here. :bow:

jaronoff97:
My main thought in reading through this code is that we have a LOT of duplication and I wonder if there's any way we can simplify this code path to make all of this easier... Regardless, this is a great add. What's the plan for migration, i.e. how many versions before we move this to beta and deprecate the old way of doing this?

TylerHelmuth:
@jaronoff97 if we followed the Collector's deprecation policy:

0.82.0: feature is release for the first time, feature gate is alpha
0.83.0: feature gate is switched to beta.  Users can disable it if they want.
0.84.0: Feature gate is stable - old code can now be removed.

We can increase any of those timelines as needed.

TylerHelmuth:
> I wonder if there's any way we can simplify this code path to make all of this easier

Yes, sdk.go and podmutator.go are pretty messy.  Once the featuregate is stable and we've removed the old code it is worth looking at the patterns we're relying on here and seeing if there is something to improve.

jaronoff97:
```
case.go:366: resource Pod:kuttl-test-internal-parrot/: .spec.initContainers.securityContext: key is missing from map
```

is one potential cause?

mat-rumian:
@jaronoff97 
It's more complex. In PR https://github.com/open-telemetry/opentelemetry-operator/pull/1273 `securityContext` settings are copied from `container` to `instrumentation initcontainer` but it looks it will not work properly for multiple containers. (BTW it's not working for Go and Apache as they use different initContainerName) In case of multiple containers `securityContext` for `Instrumentation InitContainer` will be overwritten for every container. Current implementation will work properly only for single container instrumentation.

This functionality require multiple `intrumentation initcontainers` of `the same language type` differ by `securityContext` copying instrumentation dependencies to corresponding container.

TylerHelmuth (approved):

.chloggen/1901-support-multi-cont-instr.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
pavolloffay, line 2: Is this a breaking change or enhancement?

README.md
```diff
@@ -337,6 +337,90 @@ In the above case, `myapp` and `myapp2` containers will be instrumented, `myapp3
...
+```bash
+instrumentation.opentelemetry.io/go-container-names: "go1"
+```
+
+ApacheHttpD:
+```bash
+instrumentation.opentelemetry.io/apache-httpd-container-names: "apache1,apache2"
+```
+
+Sdk:
```
pavolloffay, line 376: SDK

README.md
```diff
@@ -337,6 +337,90 @@ In the above case, `myapp` and `myapp2` containers will be instrumented, `myapp3
...
+  selector:
+    matchLabels:
+      app: my-pod-with-multi-containers-multi-instrumentations
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: my-pod-with-multi-containers-multi-instrumentations
+      annotations:
+        instrumentation.opentelemetry.io/inject-java: "true"
```
pavolloffay, line 402: Is `instrumentation.opentelemetry.io/inject-java` required? It seems redundant if `instrumentation.opentelemetry.io/java-container-names` is applied

README.md
```diff
@@ -337,6 +337,90 @@ In the above case, `myapp` and `myapp2` containers will be instrumented, `myapp3
...
+        image: myImage1
+      - name: myapp2
+        image: myImage2
+      - name: myapp3
+        image: myImage3
+```
+
+In the above case, `myapp` and `myapp2` containers will be instrumented using Java and `myapp3` using Python instrumentation.
+
+**NOTE**: Go auto-instrumentation **does not** support multicontainer pods. When injecting Go auto-instrumentation the first pod should be the only pod you want instrumented.
```
pavolloffay, line 418: If the go does not support multiple containers then why we even have `instrumentation.opentelemetry.io/go-container-names`  ?

README.md
```diff
@@ -337,6 +337,90 @@ In the above case, `myapp` and `myapp2` containers will be instrumented, `myapp3
 
 > 🚨 **NOTE**: Go auto-instrumentation **does not** support multicontainer pods. When injecting Go auto-instrumentation the first pod should be the only pod you want instrumented.
 
+#### Multi-container pods with multiple instrumentations
+
+Works only when `operator.autoinstrumentation.multi-instrumentation` feature is `enabled`.
```
pavolloffay, line 342: Is the feature flag implemented because it can be a breaking change?

pavolloffay: If the feature flag breaks existing deployments we should make it clear here, or how to migrate

README.md
jaronoff97: I think to follow the naming conventions.

mat-rumian: Yes, just to follow naming conventions

README.md
mat-rumian: Please see notes below under `annotations` - I believe all needed information related to configuration is provided.

README.md
mat-rumian: It's not redundant, in case of lack of `instrumentation.opentelemetry.io/inject-*` instrumentation is not going to be applied.

README.md
pavolloffay: Also there is a typo  `the first pod should be the only pod you want instrumented.` - >  `the first container should be the only pod you want instrumented.` 

Could we remove the annotation if it is noop?

README.md
pavolloffay: > If language instrumentation specific container names are not specified, instrumentation is performed on the first container available in the pod spec (only if single instrumentation injection is configured).

> only if single instrumentation injection is configured 

Do you mean then the feature flag is disabled? Or how is the single instrumentation configured?

.chloggen/1901-support-multi-cont-instr.yaml
TylerHelmuth: @pavolloffay when we switch the feature flag to Beta it will be listed as a breaking change.

pavolloffay:
@mat-rumian could you please resolve the remaining comments? Let's get this merged before the nginx instrumentation lads.

mat-rumian:
Hey @pavolloffay ,
I will try to finish it this week. I just get back from holidays and have a bunch of things to take care :)

README.md
mat-rumian: I think we can remove it but I would do that in separated PR as it `container-names` was used in golang instrumentation code.

README.md
mat-rumian: Ahh yes, this could be confusing. Single instrumentation is enabled by default and multi instrumentation flag is disabled by default. Also multi instrumentation means "multiple languages instrumentation for single or multiple containers" when single instrumentation stands for "single language instrumentation for single or multiple containers"

jaronoff97:
Looks like some more conflicts :( @mat-rumian

pkg/instrumentation/helper.go
```diff
@@ -63,6 +69,59 @@ func isAutoInstrumentationInjected(pod corev1.Pod) bool {
...
+		if str == "" {
+			continue
+		}
+
+		if count > 1 {
+			duplicates = append(duplicates, str)
+		}
+	}
+
+	sort.Strings(duplicates)
```
pavolloffay, line 98: why sort and then checking for the `nil` ?

pkg/instrumentation/helper.go
```diff
@@ -63,6 +69,59 @@ func isAutoInstrumentationInjected(pod corev1.Pod) bool {
 	return false
 }
 
+// Look for duplicates in the provided containers.
+func findDuplicatedContainers(ctrs []string) error {
+	// Merge is needed because of multiple containers can be provided for single instrumentation.
+	mergedContainers := strings.Join(ctrs, ",")
```
pavolloffay, line 75: I am sure if I understand the logic of this func. 

To find a duplicate sorting and then checking for two consecutive equal items should be enough.

mat-rumian: > To find a duplicate sorting and then checking for two consecutive equal items should be enough.

In this case it's not enough. It's possible that `ctrs` will look like this `[]string{"app1,app2", "app1", "app1,app3,app4", "app4"},` - concatenated slice of multiple strings coming from language specific container names annotations. We need to merge and then split container names to obtain slice of all provided names one by one.

pavolloffay: got it, thanks. Splitting each array item individually would work as well here.

pavolloffay:
@mat-rumian it's not clear to me whether we need a feature flag for this functionality. After reading the readme it's not clear if the FF is enabled the "old annotations" are disabled. I would prefer to do not introduce a FF for this.

mat-rumian:
> @mat-rumian it's not clear to me whether we need a feature flag for this functionality. After reading the readme it's not clear if the FF is enabled the "old annotations" are disabled. I would prefer to do not introduce a FF for this.

Hey @pavolloffay,
I've introduced FF on request from @TylerHelmuth https://github.com/open-telemetry/opentelemetry-operator/pull/1901#issuecomment-1629265264. Maybe it's something worth to discuss on the SIG?

pavolloffay:
There is overlap with `instrumentation.opentelemetry.io/container-names`  and `instrumentation.opentelemetry.io/<language>-container-names`, so it makes sense to have FF.

pavolloffay (approved):

jaronoff97:
Thank you so much @mat-rumian for this!!! 🙇

diegocejasprieto:
How do you use it when you have multiple containers of Go apps where you also need to specify the executable path of each one?

Issue #1907, opened 2023-06-01
Unable to configure servicemonitorselector under prometheusCR under targetallocator section


krimeshshah:
### Component(s)

receiver/prometheus

### Describe the issue you're reporting

Since scraping prometheus CR servicemonitor endpoints without any selection consume too much memory and drops the data , we decide to only scrape selected servicemonitor using serviceMonitorSelector option provided by opentelemtry crd from opentelemetry operator which we are using here in kueb-otel-stack - https://github.com/lightstep/otel-collector-charts/blob/main/charts/kube-otel-stack/values.yaml#L146

I read and found out that we can specify the and limit the number of servicemonitor that we want to scrape for metrics using "serviceMonitorSelector" (https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/crds/crd-opentelemetrycollector.yaml#L2039) which can reduce the memory consumption i believe as it will reduce the no of scraping endpoints.
As described in opentelemetrycollector crd, this is parameter is of type string but at the same time it is written that it is key value pair. And as per the kube-prometheus-stack i should configue like - https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L2975
But when the same way when i try to configure multiple servicemonitor with matchLabel and its key value pair it kube-otel-stack deployment fails.
Hence i tried to configure as below ## Default collector for metrics (includes infrastructure metrics) metricsCollector: name: metrics clusterName: dev image: otel/opentelemetry-collector-contrib:0.73.0 enabled: true mode: statefulset replicas: 6 targetallocator: enabled: true allocationStrategy: "consistent-hashing" filterStrategy: relabel-config replicas: 2 image: ghcr.io/open-telemetry/opentelemetry-operator/target-allocator:0.73.0 prometheusCR: enabled: true serviceMonitorSelector: matchLabels: "lightstep-metrics-collector"

But the things is , i can only specify one servicemonitor this way, how can i configure multiple servicemonitor?

jaronoff97:
[...]

PR #1923, opened 2023-07-13
[target allocator] Make scrape interval configurable

swiatekm-sumo:


cmd/otel-allocator/config/config.go
```diff
@@ -96,6 +99,12 @@ func unmarshal(cfg *Config, configFile string) error {
 	return nil
 }
 
+func createDefaultConfig() Config {
+	return Config{
+		CRScrapeInterval: DefaultCRScrapeInterval,
```
jaronoff97, line 104: Are there any other defaults that would be good to have here?

jaronoff97: i.e. would it be better for us to just embed something like `CommonPrometheusFields` in the config?

swiatekm-sumo: Having had a quick look through that (https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#commonprometheusfields), I think there's way too many that wouldn't apply to TA, or which would apply, but we'd rather set them once at the Collector level and inherit. So I think we'll need to cherry pick the ones we want, unfortunately.

swiatekm-sumo: Ah, you mean just for TA config, not for the spec in the CR? I guess we could do that. It'd still be a bit messy to have settable values that don't do anything, but if they're just exposed in the TA config and not the CR, it's less of a problem.

jaronoff97: holy crap there are a LOT of fields in there. I think it's okay to piece meal it on second thought.

swiatekm-sumo: I can open an issue to decide what we want from there, exactly. Should that block this change, though?

cmd/otel-allocator/config/config.go
```diff
@@ -38,12 +39,14 @@ import (
...
 const DefaultResyncTime = 5 * time.Minute
 const DefaultConfigFilePath string = "/conf/targetallocator.yaml"
+const DefaultCRScrapeInterval model.Duration = model.Duration(time.Second * 30)
 
 type Config struct {
 	LabelSelector          map[string]string  `yaml:"label_selector,omitempty"`
 	Config                 *promconfig.Config `yaml:"config"`
 	AllocationStrategy     *string            `yaml:"allocation_strategy,omitempty"`
 	FilterStrategy         *string            `yaml:"filter_strategy,omitempty"`
+	CRScrapeInterval       model.Duration     `yaml:"cr_scrape_interval,omitempty"`
```
jaronoff97, line 49: I thought some more about this over the weekend, and I think this may be better if we make an embedded CR config resource and that's where we'll be able to then move the incorrect Pod and Service monitor selectors to. Does that seem reasonable to you? ([this issue](https://github.com/open-telemetry/opentelemetry-operator/issues/1907))

swiatekm-sumo: Sure, what should we name it? I'm ok with `PrometheusCR` for consistency, but no strong opinions. Also, should we allow enabling the functionality via the config file (currently it can only be done via a CLI flag)?

Also, do you want to move the current selector attributes, as opposed to creating new ones with the right semantics?

jaronoff97: PrometheusCR makes sense. And I'm not sure I have an opinion on where the enabling should happen... i trust your judgement. And let's keep the current selector attributes where they are now, and then in a follow up make the new ones in the `PrometheusCR`, deprecating the current ones.

swiatekm-sumo: Done. I left the CLIConf alone for now, it's not entirely clear what the logic should be there, and I'd rather do it in a separate PR without these other changes in the way.

swiatekm-sumo:
FYI, the E2E upgrade test passes on _my machine_ :tm: .

jaronoff97 (approved):
Looks great, rerunning the tests and will merge when they pass!

PR #1924, opened 2023-07-14
[target allocator] Fix Target Allocator config reload

swiatekm-sumo:
Fix a critical issue in #1889. We were reading the configuration string from the wrong ConfigMap field and not checking the result, and therefore always adding the hash of the empty string.

You might wonder how it is that the E2E test passes even though the hash doesn't change. In this specific case, the spec is modified by enabling PrometheusCR, which actually adds a command line argument to the target allocator, and thus causes it to be restarted regardless. I've updated the test to account for this.

I've made the unit test for this more specific as well, so we check if we have the hash of the right value, not just that it's syntactically correct.

Finally, I've actually tested it manually as well, and it does work.

### DIFF
--- a/pkg/targetallocator/annotations.go
+++ b/pkg/targetallocator/annotations.go
@@ -35,7 +35,10 @@ func Annotations(instance v1alpha1.OpenTelemetryCollector) map[string]string {
 
 	configMap, err := ConfigMap(instance)
 	if err == nil {
-		annotations[configMapHashAnnotationKey] = getConfigMapSHA(configMap)
+		cmHash := getConfigMapSHA(configMap)
+		if cmHash != "" {
+			annotations[configMapHashAnnotationKey] = getConfigMapSHA(configMap)
+		}
 	}
 
 	return annotations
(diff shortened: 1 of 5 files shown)


jaronoff97 (approved):

pkg/targetallocator/annotations.go
```diff
@@ -35,15 +35,21 @@ func Annotations(instance v1alpha1.OpenTelemetryCollector) map[string]string {
...
+			annotations[configMapHashAnnotationKey] = getConfigMapSHA(configMap)
+		}
 	}
 
 	return annotations
 }
 
 // getConfigMapSHA returns the hash of the content of the TA ConfigMap.
 func getConfigMapSHA(configMap v1.ConfigMap) string {
-	configBytes := configMap.BinaryData[targetAllocatorFilename]
```
jaronoff97, line 46: this is the crux of the change, yeah?

swiatekm-sumo: Yes. In my mind, I think I thought BinaryData was an alias for Data with a different type. I blame spending too many years writing Python.

jaronoff97 (approved):

PR #1928, opened 2023-07-17
feat: add imagePullSecrets to opentelemetry spec

Jaxwood:
Add for image pull secrets for deployment, statefulsets and daemonsets


jpkrohling:
This has been requested a few times in the past. So far, the decision was that this doesn't bring any benefits compared to the current solution, which is to create a dedicated service account with the appropriate pull secrets and use that service account for the CR.

If this is still the decision, it would be good to document it somewhere.

Jaxwood:
> This has been requested a few times in the past. So far, the decision was that this doesn't bring any benefits compared to the current solution, which is to create a dedicated service account with the appropriate pull secrets and use that service account for the CR.
> 
> If this is still the decision, it would be good to document it somewhere.

Alright - thanks. Will close this PR then 👍

PR #1953, opened 2023-07-21
Expose the Prometheus exporter port when specified in the config

### LINKED ISSUE #1689: Expose prometheusexporter port in collector pod
Use case: I want to use https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/prometheusexporter to export pipeline's metrics  to prometheus by using the `podmonitor`. 
The `podmonitor` has config to set the port as string https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#endpoint.

When `prometheusexporter`  is enabled the port is not exposed as container port on the collector 

```yaml
    exporters:
      prometheus:
        endpoint: 0.0.0.0:8889
```

```yaml
    name: otc-container
    ports:
    - containerPort: 8888
      name: metrics
      protocol: TCP
    - containerPort: 4317
      name: otlp-grpc
      protocol: TCP
    - containerPort: 4318
      name: otlp-http
      protocol: TCP
    - containerPort: 55681
      name: otlp-http-legac
      protocol: TCP
```

iblancasa:
Resolves #1689


TylerHelmuth (approved):

pkg/collector/container.go
```diff
@@ -223,6 +235,36 @@ func getMetricsPort(c map[interface{}]interface{}) (int32, error) {
...
+	type cfg struct {
+		Exporters exportersCfg
+	}
+	var cOut cfg
+	err := mapstructure.Decode(c, &cOut)
+	if err != nil {
+		return 0, nil
+	}
+
+	_, port, err := net.SplitHostPort(cOut.Exporters.Prometheus.Endpoint)
```
frzifus, line 256: Does this also work when there are two prometheus exporters configured?

iblancasa: Thanks for the feedback!

pkg/naming/ports.go
```diff
@@ -0,0 +1,42 @@
...
+			if i == 0 || i == utf8.RuneCountInString(name)-1 {
+				d = append(d, 'a')
+			} else {
+				d = append(d, '-')
+			}
+		}
+	}
+
+	return string(d)
+}
```
frzifus, line 42: Or using string replace instead of a loop with nested if/else?

```suggestion
func PortName(name string) string {
    // Mapping function to replace invalid characters with "-"
    mapper := func(r rune) rune {
        if unicode.IsLetter(r) || unicode.IsDigit(r) {
            return r
        }
        return '-'
    }

    // Map the name using the mapper function and replace "-" at the beginning or end with "a"
    result := strings.Map(mapper, name)
    if strings.HasPrefix(result, "-") {
        result = "a" + result[1:]
    }
    if strings.HasSuffix(result, "-") {
        result = result[:len(result)-1] + "a"
    }

    return result
}
```

pkg/naming/ports.go
```diff
@@ -0,0 +1,42 @@
...
+
+package naming
+
+import (
+	"strings"
+	"unicode/utf8"
+)
+
+// PortName returns a dns-safe string for the given name.
+// Any char that is not [a-z0-9] is replaced by "-" or "a".
```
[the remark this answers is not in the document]
iblancasa: Why?

PR #1957, opened 2023-07-24
Refactor target allocator config handling

### LINKED ISSUE #2129: Allow target allocator to be fully configured via the configuration file
Right now, some target allocator configuration values, like the listen address, can only be specified via command line arguments. There's no reason for this not to be possible via the config file.

At the same time, we should use a single struct for the entirety of the configuration. Right now, it's somewhat arbitrarily split between CLI and file-based config.

swiatekm-sumo:
TargetAllocator config is internally (rather arbitrarily) split into config from cli, contained in the `CLIConf` struct, and config loaded from the config file, contained in the `Config` struct. This change moves all configuration attributes to the latter, and explicitly gives the CLI arguments higher priority.

I've also moved the flag arguments into an explicit flag set and added tests for it.

Resolves #2129.

### DIFF
--- a/cmd/otel-allocator/config/config.go
+++ b/cmd/otel-allocator/config/config.go
@@ -71,12 +71,33 @@ func (c Config) GetTargetsFilterStrategy() string {
 	return ""
 }
 
-func Load(file string) (Config, error) {
-	cfg := createDefaultConfig()
-	if err := unmarshal(&cfg, file); err != nil {
-		return Config{}, err
+func LoadFromFile(file string, target *Config) error {
+	return unmarshal(target, file)
+}
+
+func LoadFromCLI(target *Config) error {
+	// set the rest of the config attributes based on command-line flag values
+	target.RootLogger = zap.New(zap.UseFlagOptions(&zapCmdLineOpts))
+	klog.SetLogger(target.RootLogger)
+	ctrl.SetLogger(target.RootLogger)
+
+	target.KubeConfigFilePath = *kubeConfigPathFlag
+	clusterConfig, err := clientcmd.BuildConfigFromFlags("", *kubeConfigPathFlag)
+	if err != nil {
+		pathError := &fs.PathError{}
+		if ok := errors.As(err, &pathError); !ok {
+			return err
+		}
+		clusterConfig, err = rest.InClusterConfig()
+		if err != nil {
+			return err
+		}
 	}
-	return cfg, nil
+	target.ListenAddr = *listenAddrFlag
+	target.PrometheusCR.Enabled = *prometheusCREnabledFlag
+	target.ClusterConfig = clusterConfig
+
+	return nil
 }
 
 func unmarshal(cfg *Config, configFile string) error {
@@ -91,7 +112,7 @@ func unmarshal(cfg *Config, configFile string) error {
 	return nil
 }
 
-func createDefaultConfig() Config {
+func CreateDefaultConfig() Config {
 	return Config{
 		PrometheusCR: PrometheusCRConfig{
 			ScrapeInterval: DefaultCRScrapeInterval,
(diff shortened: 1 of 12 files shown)


cmd/otel-allocator/config/config.go
```diff
@@ -42,8 +39,12 @@ const DefaultConfigFilePath string = "/conf/targetallocator.yaml"
 const DefaultCRScrapeInterval model.Duration = model.Duration(time.Second * 30)
 
 type Config struct {
+	ListenAddr             *string
+	KubeConfigFilePath     string
+	ClusterConfig          *rest.Config
+	RootLogger             logr.Logger
```
kristinapathak, line 45: Is it intentional to make these able to be unmarshaled from the config file? If not, I think we want this tag for each line: 
```
`yaml:"-"`
```

swiatekm-sumo: Good catch. I think the answer is yes for the basic types, and no for the more complex ones.

cmd/otel-allocator/config/config.go
```diff
@@ -103,52 +112,36 @@ func unmarshal(cfg *Config, configFile string) error {
...
-	ctrl.SetLogger(cLIConf.RootLogger)
+	// load the config from the config file
+	err := LoadFromFile(*configFilePathFlag, &config)
+	if err != nil {
+		return nil, "", err
+	}
 
-	clusterConfig, err := clientcmd.BuildConfigFromFlags("", *kubeconfigPath)
-	cLIConf.KubeConfigFilePath = *kubeconfigPath
+	err = LoadFromCLI(&config)
```
jaronoff97, line 133: ah neat so this lets us override the file with args, great :)

cmd/otel-allocator/config/flags.go
```diff
@@ -0,0 +1,45 @@
...
+// Flag definitions.
+var (
+	configFilePathFlag      *string
+	listenAddrFlag          *string
+	prometheusCREnabledFlag *bool
+	kubeConfigPathFlag      *string
+	zapCmdLineOpts          zap.Options
+)
+
+func initFlags() {
```
jaronoff97, line 35: Is there a reason to have both `initFlags` and `init`?

cmd/otel-allocator/watcher/file.go
```diff
@@ -43,19 +43,20 @@ func NewFileWatcher(logger logr.Logger, config config.CLIConfig) (*FileWatcher,
...
+		configFilePath: configFilePath,
 		watcher:        fileWatcher,
 		closer:         make(chan bool),
 	}, nil
 }
 
 func (f *FileWatcher) LoadConfig(_ context.Context) (*promconfig.Config, error) {
-	cfg, err := config.Load(f.configFilePath)
+	cfg := config.CreateDefaultConfig()
+	err := config.LoadFromFile(f.configFilePath, &cfg)
```
jaronoff97, line 54: Would this result in us overriding the flags someone sets? Maybe no because I don't think they can set anything for the promConfig?

cmd/otel-allocator/watcher/promOperator.go
```diff
@@ -37,13 +37,13 @@ import (
 	allocatorconfig "github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/config"
 )
 
-func NewPrometheusCRWatcher(logger logr.Logger, cfg allocatorconfig.Config, cliConfig allocatorconfig.CLIConfig) (*PrometheusCRWatcher, error) {
-	mClient, err := monitoringclient.NewForConfig(cliConfig.ClusterConfig)
+func NewPrometheusCRWatcher(logger logr.Logger, cfg *allocatorconfig.Config) (*PrometheusCRWatcher, error) {
```
jaronoff97, line 40: Should we check for nil here?

jaronoff97 (commented):
A few questions/thoughts

cmd/otel-allocator/watcher/file.go
swiatekm-sumo: I'd like to deprecate as many flags as I can in a later step because of ambiguities like this. Maybe even get rid of this hot reload functionality, as it's a bit confusing in the context of K8s. It's an antipattern to let changes to a ConfigMap affect a Pod without restarting it - prometheus-operator mostly does it because restarting Prometheus is somewhat expensive, but that's not really a problem for TA.

cmd/otel-allocator/config/flags.go
swiatekm-sumo: Just code organization. I'm currently planning to explicitly make a flagset and manually pass `os.Args` to it - that way it's actually testable and avoids the need for this init magic.

cmd/otel-allocator/watcher/promOperator.go
swiatekm-sumo: I think I'd prefer to just pass it by value. I don't think that's more expensive in a meaningful way, and this is clearly not supposed to be nil.

cmd/otel-allocator/watcher/file.go
```diff
@@ -43,19 +43,20 @@ func NewFileWatcher(logger logr.Logger, config config.CLIConfig) (*FileWatcher,
...
 	return &FileWatcher{
 		logger:         logger,
-		configFilePath: *config.ConfigFilePath,
+		configFilePath: configFilePath,
 		watcher:        fileWatcher,
 		closer:         make(chan bool),
 	}, nil
 }
 
 func (f *FileWatcher) LoadConfig(_ context.Context) (*promconfig.Config, error) {
```
jaronoff97, line 52: the plan is to deprecate this whole file right? if so, can we add a deprecation notice?

jaronoff97 (approved):
Overall super happy with this approach, thanks for doing this!

cmd/otel-allocator/watcher/file.go
swiatekm-sumo: I'd rather do it in a separate PR and clearly call it out in the changelog. This one already has a enough changes in it imo.

PR #1959, opened 2023-07-24
Refactor reconciliation to be generic based on resource

### LINKED ISSUE #1876: Isolate basic reconcile functionality into a reusable package
I would propose to open a separate PR to isolate basic reconcile functionality e.g. methods like `expected...`/`delete...` into a reusable package. Or even better refactor the code and use mutate func for all object types e.g. https://github.com/grafana/tempo-operator/blob/main/internal/manifests/mutate.go#L27  with https://github.com/grafana/tempo-operator/blob/1a49fbb282ff1596fd874887341b80192eecffc1/controllers/tempo/tempostack_controller.go#L250

_Originally posted by @pavolloffay in https://github.com/open-telemetry/opentelemetry-operator/issues/1559#issuecomment-1573957199_

jaronoff97:
closes #1876


TylerHelmuth:
image 👀

jaronoff97:
i know :( not much to avoid this kind of diff unfortunately. I'm going to be marking the old packages deprecated for 2 versions and then delete them after 2 more releases. The actual reconciliation functionality is behind an FF which controls which group of packages is being used.

swiatekm-sumo:
> i know :( not much to avoid this kind of diff unfortunately. I'm going to be marking the old packages deprecated for 2 versions and then delete them after 2 more releases. The actual reconciliation functionality is behind an FF which controls which group of packages is being used.

It looks like some of the packages are just copied verbatim, or at least I haven't spotted any difference in the targetallocator one. Keeping changes synced between these is going to be a massive pain. For packages whose content doesn't actually change (but we still want to deprecate and move), would it be possible to keep them as-is, and then move them at the same time as they're removed after deprecation?

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -110,6 +110,10 @@ spec:
           - ""
           resources:
           - configmaps
+          - pods
+          - secrets
+          - serviceaccounts
+          - services
```
jaronoff97, line 116: I know this looks scary, but there shouldn't be a real diff here. I just changed the kubebuilder stuff to be grouped together correctly.

jaronoff97:
@pavolloffay @TylerHelmuth I think i already know the answer, but I'm thinking I should split this up in to two PRs. One for moving all the manifest building in to its own package and then another for changing how reconciliation happens. I think that's the cleanest / safest way of doing this type of large change. Does that make sense, or is there a better way to break this up?

jaronoff97:
Closing this in favor of a new, smaller PR

PR #1965, opened 2023-07-26
[chore] Refactor Manifest Generation to internal package

jaronoff97:
This PR precedes our ability to do #1876. Before we can change how we handle reconciliation we have to isolate the functionality that generates our resources. This is done by de-coupling reconciliation from how those manifests get created. This results in a much simpler and smaller scoped package – one focused on creating a resource, the other focused on reconciling a resource. 

Open Questions:
* Do we want the manifests to be in the internal package
* Should we combine manifests for similar things? i.e. target allocator, collector, etc.
* Should we move anything for sidecar generation in there?


internal/manifests/targetallocator/annotations.go
```diff
@@ -26,15 +26,14 @@ import (
 const configMapHashAnnotationKey = "opentelemetry-targetallocator-config/hash"
 
 // Annotations returns the annotations for the TargetAllocator Pod.
-func Annotations(instance v1alpha1.OpenTelemetryCollector) map[string]string {
+func Annotations(instance v1alpha1.OpenTelemetryCollector, configMap *v1.ConfigMap) map[string]string {
```
jaronoff97, line 29: This was one of the more significant changes, rather than creating the config map in this function its passed down and optionally nil. cc @swiatekm-sumo who may have thoughts here.

internal/reconcileutil/util.go
```diff
@@ -0,0 +1,42 @@
...
+	return func(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelemetryCollector) (client.Object, error) {
+		return f(cfg, logger, otelcol), nil
+	}
+}
+
+func Creator[T client.Object](f Manifest[T]) ObjectCreator {
+	return func(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelemetryCollector) (client.Object, error) {
+		return f(cfg, logger, otelcol)
+	}
+}
```
jaronoff97, line 42: I'm doing some type shenanigans here so that we can have each of the manifest functions return their actual type.

internal/manifests/targetallocator/deployment.go
```diff
@@ -22,16 +22,22 @@ import (
...
+func Deployment(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelemetryCollector) *appsv1.Deployment {
 	name := naming.TargetAllocator(otelcol)
 	labels := Labels(otelcol, name)
-	annotations := Annotations(otelcol)
 
-	return appsv1.Deployment{
+	configMap, err := ConfigMap(cfg, logger, otelcol)
+	if err != nil {
+		logger.Info("failed to construct target allocator config map for annotations")
+		configMap = nil
```
swiatekm-sumo, line 36: Should we return the Deployment here knowing the Pods will fail to be created (because the ConfigMap will be absent)? Philosophically, we should do one of the following imo:

1. Manifest generation is purely syntactic, and it's the reconciler's job to abort if one of them creates an error.
2. Manifest generation should return errors if it knows the output will fail semantically.

I think I'm ok with either, but it should be clear that the Deployment we return here may not actually work depending on other reconciliation results. In that case, returning it even though the ConfigMap fails is basically simplifying the API.

jaronoff97: I think we should error and set that on the status for the otelcol, preventing the bad thing from being rolled out and giving users the reason why. I think for now we should keep this the same as currently this change is a no-op (just shifting files around.) The next PR is going to change how we do mutation/reconciliation and is probably a better place for things like that IMO.

internal/manifests/collector/container.go
```diff
@@ -117,9 +118,13 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
...
 	var livenessProbe *corev1.Probe
-	if config, err := adapters.ConfigFromString(otelcol.Spec.Config); err == nil {
-		if probe, err := getLivenessProbe(config, otelcol.Spec.LivenessProbe); err == nil {
+	if configFromString, err := adapters.ConfigFromString(otelcol.Spec.Config); err == nil {
+		if probe, err := getLivenessProbe(configFromString, otelcol.Spec.LivenessProbe); err == nil {
 			livenessProbe = probe
+		} else if errors.Is(err, adapters.ErrNoServiceExtensions) {
+			logger.Info("Extensions not configured, skipping liveness probe creation")
+		} else if errors.Is(err, adapters.ErrNoServiceExtensionHealthCheck) {
+			logger.Info("Healthcheck extension not configured, skipping liveness probe creation")
```
jaronoff97, line 127: This is one minor change that was bugging me while debugging: rather than a noisy log we simply let people know why we are skipping liveness probe creation.

internal/manifests/targetallocator/annotations.go
swiatekm-sumo: I'm fine with this, it's basically just moving the evaluation a level up.

TylerHelmuth: Agreed

internal/manifests/collector/ingress.go
```diff
@@ -0,0 +1,127 @@
...
+						},
+					},
+				},
+			},
+			IngressClassName: otelcol.Spec.Ingress.IngressClassName,
+		},
+	}
+}
+
+// TODO: what's the correct error state here? Should we error out if we can't extract the configuration?
```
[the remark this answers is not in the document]
jaronoff97: okay, i'm going to make some follow up issues for these and link them.

TylerHelmuth (approved):
I have spent almost all my time in the Operator focused on Auto Instrumentation so I have not looked at this code much in the past. I don't see anything else I want to comment on, but still take my approval with a grain of salt bc this is a big refactor and I didn't spend time it the original version of the code.

internal/manifests/collector/collector.go
```diff
@@ -0,0 +1,64 @@
...
+		reconcileutil.CreatorWithoutError(ConfigMap),
+		reconcileutil.CreatorWithoutError(HorizontalPodAutoscaler),
+		reconcileutil.CreatorWithoutError(ServiceAccount),
+		reconcileutil.CreatorWithoutError(Service),
+		reconcileutil.CreatorWithoutError(HeadlessService),
+		reconcileutil.CreatorWithoutError(MonitoringService),
+		reconcileutil.CreatorWithoutError(Ingress),
+	}...)
+	if params.Instance.Spec.Observability.Metrics.EnableMetrics && featuregate.PrometheusOperatorIsAvailable.IsEnabled() {
+		objects = append(objects, reconcileutil.Creator(ServiceMonitor))
```
swiatekm-sumo, line 48: See my comments on the targetallocator part.

swiatekm-sumo (commented):
I looked through the relatively boring parts and didn't see anything concerning. Paid a bit more attention to the interesting parts and left some comments, mostly about naming.

internal/manifests/targetallocator/targetallocator.go
```diff
@@ -0,0 +1,43 @@
...
+		reconcileutil.Creator(ConfigMap),
+		reconcileutil.CreatorWithoutError(Deployment),
+		reconcileutil.CreatorWithoutError(ServiceAccount),
+		reconcileutil.CreatorWithoutError(Service),
+	}
+	for _, object := range objects {
+		res, err := object(params.Config, params.Log, params.Instance)
+		if err != nil {
+			return nil, err
+		} else if res != nil && res.DeepCopyObject() != nil {
```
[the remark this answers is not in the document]
jaronoff97: honestly this really confused me, but my understanding is that because res is an interface, res should never be nil because it's going to be `client.Object(nil)` when what we want to ask is if it is actually nil which is why we do the deepcopy. Open to other ideas here.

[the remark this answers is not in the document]
jaronoff97: https://stackoverflow.com/questions/13476349/check-for-nil-and-nil-interface-in-go for more info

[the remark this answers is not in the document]
swiatekm-sumo: Ah, it's actually a check for "typed nil"? I'd leave a comment explaining this here then, as it's not very obvious.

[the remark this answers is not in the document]
jaronoff97: can do! Do you think i should just use reflect here? Or is the deep copy sufficient?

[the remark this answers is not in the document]
swiatekm-sumo: I'm not a Go expert by any measure, but here's how I see the possibilities:

1. Stay with the current behaviour. Most of the time we're going to be doing unnecessary deep copies, which is far from ideal.
2. Use reflect to check if the value is a nil pointer. Effective, but a bit hacky.
3. Only check for plain nil, and allow the output to contain typed nil values. The callers downstream who will actually cast this to a concrete type (`*v1.Deployment` for example) can check for the remaining nil safely on their own.
4. Maybe there's an interface in apimachinery we could cast this to that would let us do a cheap check?

From reading the linked mailing list thread, I think 3 is supposed to be the correct solution to this problem, and personally I'm fine with it. The additional check downstream is realistically something we need to do anyway.

internal/manifests/collector/collector.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
+	"github.com/open-telemetry/opentelemetry-operator/internal/reconcileutil"
+	"github.com/open-telemetry/opentelemetry-operator/pkg/featuregate"
+)
+
+func Build(params reconcileutil.Params) ([]client.Object, error) {
```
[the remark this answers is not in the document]
frzifus: Lets put a comment on top and link this comment or an issue for it.

internal/manifests/collector/horizontalpodautoscaler.go
```diff
@@ -48,6 +46,7 @@ func HorizontalPodAutoscaler(cfg config.Config, logger logr.Logger, otelcol v1al
 		logger.Info("Autoscaler field is unset in Spec, skipping")
 		return nil
 	}
+	autoscalingVersion := cfg.AutoscalingVersion()
```
frzifus, line 49: nit, but why did we move this here? I did expect to keep it where it is, or moving the part where its actually used. ^^

internal/manifests/collector/ingress.go
```diff
@@ -0,0 +1,127 @@
...
+	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
+	"github.com/open-telemetry/opentelemetry-operator/internal/config"
+	"github.com/open-telemetry/opentelemetry-operator/internal/reconcileutil/naming"
+)
+
+func Ingress(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelemetryCollector) *networkingv1.Ingress {
```
frzifus, line 32: I know that @pavolloffay is working on some Ingress/Route related stuff. Just that you are aware, that this might be a conflict that needs to be resolved on one of both ends. :dagger: Maybe its easier to move those files instead of coping and deleting.

frzifus (commented):
Its touching quite a few places - but overall LGTM :1st_place_medal:

internal/manifests/collector/ingress.go
jaronoff97: Yeah this is basically a move, but because its decoupling the desired ingress from the reconciliation logic it marks it as a new file

internal/reconcileutil/util.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"github.com/go-logr/logr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
+	"github.com/open-telemetry/opentelemetry-operator/internal/config"
+)
+
+type Builder func(params Params) ([]client.Object, error)
```
pavolloffay, line 25: This does not seem to belong to the "util" package. I assume this will be the major API implemented by component that need reconciliation

pavolloffay:
> Do we want the manifests to be in the internal package

I would say majority of the codebase should be in the `internal` package. Probably just installation of the controllers and webhook could be exposed

internal/reconcileutil/util.go
jaronoff97: Should this just be in the internal manifests package then?

jaronoff97: i was thinking this was the right place because everything is working off of the params object.

jaronoff97: moved to internal/manifests/main.go :)

internal/manifests/main.go
pavolloffay, line 1: Could we rename this file to be more explicit? 

There are bunch of `main` files in the codebase, but it's rather an antipatter which was introduced when the project was bootstrapped.

internal/manifests/main.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"github.com/go-logr/logr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
+	"github.com/open-telemetry/opentelemetry-operator/internal/config"
+	"github.com/open-telemetry/opentelemetry-operator/internal/reconcileutil"
+)
+
+type Builder func(params reconcileutil.Params) ([]client.Object, error)
```
pavolloffay, line 26: Could we move here as well the `Params` struct? Having these two APIs in different packages does not make sense - IIRC params is a parameter to the builder

pavolloffay (approved):

PR #1970, opened 2023-07-28
Make sure port name in ingress and route match service

### LINKED ISSUE #1954: The "otlp-http-legacy" port breaks ingresses
I'm trying to use the otel operator to manage a collector ingress automatically. Unfortunately, the ingress does not work. At least, it does not work when backed by ALB controller in AWS. Here's a relevant part of my `otelcol` resource:

``` yaml
kind: OpenTelemetryCollector
metadata: ...
spec:
  ingress:
    hostname: gateway.redacted.com
    type: ingress
  config: |
    receivers:
      otlp:
        protocols:
          grpc:
          http:
            cors:
              allowed_origins:
                - "https://*"
```

And here are snippets of the `service` and `ingress` resources created by the operator.

``` yaml
apiVersion: v1
kind: Service
metadata: ...
spec:
  ...
  ports:
  - appProtocol: grpc
    name: otlp-grpc
    port: 4317
    protocol: TCP
    targetPort: 4317
  - appProtocol: http
    name: otlp-http
    port: 4318
    protocol: TCP
    targetPort: 4318
  - appProtocol: http
    name: otlp-http-legacy
    port: 55681
    protocol: TCP
    targetPort: 4318
```

``` yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata: ...
spec:
  rules:
  - host: gateway.redacted.com
    http:
      paths:
      - backend:
          service:
            name: opentelemetry-deployment-collector
            port:
              name: otlp-grpc
        path: /otlp-grpc
        pathType: Prefix
      - backend:
          service:
            name: opentelemetry-deployment-collector
            port:
              name: otlp-http
        path: /otlp-http
        pathType: Prefix
      - backend:
          service:
            name: opentelemetry-deployment-collector
            port:
              name: otlp-http-legac
        path: /otlp-http-legacy
        pathType: Prefix
```

Notice that the service's `.spec.ports[2].name` is `otlp-http-legacy` while the ingress' `.spec.rules[0].http.paths[2].backend.service.port.name` is `otlp-http-legac`. In other words, the name is truncated in the ingress, missing the final `y` character.

I believe that I found the source of this issue in [collector/reconcile/ingress.go](https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/collector/reconcile/ingress.go#L62), which limits the name to 15 characters. However, the [service](https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/collector/parser/receiver.go#L181) is parsed and limited to 63 characters.

Is there a reason that the ingress string is limited to 15 characters? Can we simply change it to 63 characters? Alternatively, can we limit the string in the service resource to the same size?

pavolloffay:
Port name in service/ingress/route can be longer than 15 characters. The port name 15 characters is enforced in ports on pods.

Updates #1954


pkg/naming/main.go
```diff
@@ -15,18 +15,14 @@
...
 package naming
 
-import (
-	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
-)
-
 // ConfigMap builds the name for the config map used in the OpenTelemetryCollector containers.
-func ConfigMap(otelcol v1alpha1.OpenTelemetryCollector) string {
-	return DNSName(Truncate("%s-collector", 63, otelcol.Name))
+func ConfigMap(otelcol string) string {
```
pavolloffay, line 19: Changing this to string to avoid circular imports.

pkg/naming/port.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+var (
+	// DNS_LABEL constraints: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names
+	dnsLabelValidation = regexp.MustCompile("^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$")
+)
+
+// PortName defines the port name used in services, ingresses and routes.
+// The port name in pods is validated differently it has to be 15 characters long.
+func PortName(receiverName string, port int32) string {
```
pavolloffay, line 16: This was moved from `parser`  package. It is being called from multiple places now

pkg/naming/main.go
jaronoff97: This is great :)

jaronoff97 (approved):

pkg/collector/reconcile/route.go
```diff
@@ -75,11 +74,10 @@ func desiredRoutes(_ context.Context, params Params) []routev1.Route {
 				Path: "/" + p.Name,
 				To: routev1.RouteTargetReference{
 					Kind: "Service",
-					Name: naming.Service(params.Instance),
+					Name: naming.Service(params.Instance.Name),
 				},
 				Port: &routev1.RoutePort{
-					// Valid names must be non-empty and no more than 15 characters long.
-					TargetPort: intstr.FromString(naming.Truncate(p.Name, 15)),
+					TargetPort: intstr.FromString(naming.PortName(p.Name, p.Port)),
```
TylerHelmuth, line 80: I might be missing something, but it seems the new code doesn't enforce a 15 character limit anymore - I didn't see anything in `naming.PortName` that is truncating this value to 15 characters.

pkg/collector/reconcile/ingress.go
```diff
@@ -56,10 +56,9 @@ func desiredIngresses(_ context.Context, params Params) *networkingv1.Ingress {
 			PathType: &pathType,
 			Backend: networkingv1.IngressBackend{
 				Service: &networkingv1.IngressServiceBackend{
-					Name: naming.Service(params.Instance),
+					Name: naming.Service(params.Instance.Name),
 					Port: networkingv1.ServiceBackendPort{
-						// Valid names must be non-empty and no more than 15 characters long.
-						Name: naming.Truncate(p.Name, 15),
+						Name: naming.PortName(p.Name, p.Port),
```
TylerHelmuth, line 61: I might be missing something, but it seems the new code doesn't enforce a 15 character limit anymore - I didn't see anything in `naming.PortName` that is truncating this value to 15 characters.

pkg/collector/reconcile/route.go
pavolloffay: correct, the 15 character limit should not be applied to service/route/ingress objects. It is only required for port names in the pod spec. 

I have documented it in the `naming` package.

iblancasa (approved):

PR #1979, opened 2023-08-02
Add dependabot grouping for common things

### LINKED ISSUE #1949: Use Dependabot grouping
Dependabot now supports grouping! https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

We should use this, as its tiring to have to manually group them

jaronoff97:
Closes #1949


swiatekm-sumo (approved):
Do we normally add changelog entries for these kinds of changes? This one seems internal and of little interest to users.

jaronoff97:
that's fair... I included it because it changes some repo processes. If you think it's not needed I can remove it :)

swiatekm-sumo:
I don't feel very strongly about it, but as a rule of thumb I tend to treat changelogs as a way of communicating changes to users, and therefore try to minimize noise in them. I don't actually know if we have any kind of policy for this project.

.github/dependabot.yml
```diff
@@ -4,14 +4,54 @@ updates:
     directory: /
     schedule:
       interval: daily
+    # Create a group of dependencies to be updated together in one pull request
+    groups:
```
bryan-aguilar, line 8: Should a wildcard (*) be added in some group? I think this means you will miss grouped updates for any dependency does not match one of the patterns below. 

Is that intentional? The [docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups) aren't clear how the grouped updates will behave if a dependency matches a pattern in multiple groups. Or I missed it, which happens.

bryan-aguilar: Thinking about this more. You could probably get a little complicated here by having a wildcard group that has an exclude pattern for all the other patterns listed earlier.

jaronoff97: I'm not sure what you mean... here's an example of what sarama does https://github.com/IBM/sarama/blob/c10bd1e5709a7b47729b445bc98f2e41bc7cc0a8/.github/dependabot.yml#L27

bryan-aguilar: https://github.com/open-telemetry/opentelemetry-operator/pull/1807

I think what I am getting at is dependencies such as `github.com/stretchr/testify` and `github.com/gin-gonic/gin`  will not be matched into any group right? So this means we will still see one of PRs raised for these. That could be completely intentional, but the goal wasn't clear.

bryan-aguilar: I guess the PR title says `common things` and not `everything`. So looks good to me! Just adding some food for thought.

bryan-aguilar (approved):

.github/dependabot.yml
jaronoff97: oh yeah very intentional. I took a list of all the things I most frequently have to update and just did some grouping on those

PR #1980, opened 2023-08-02
Adding Additional Containers to the Collector spec

nikore:
Adds `AdditionalContainers` field to the Collector spec to allow configuring sidecars to run along side the collector container in the pod. 

This is required for a use case I have where the company I'm shipping metrics too requires using their agent to ship metrics into their cloud. 

This solves #1987

### DIFF
--- a/.chloggen/1987-adding-support-for-sidecar-containers.yaml
+++ b/.chloggen/1987-adding-support-for-sidecar-containers.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `AdditionalContainers` to the collector spec allowing to configure sidecar containers. This only applies to Deployment/StatefulSet/DeamonSet deployments of the collector
+
+# One or more tracking issues related to the change
+issues: [1987]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/.chloggen/1987-adding-support-for-sidecar-containers.yaml
+++ b/.chloggen/1987-adding-support-for-sidecar-containers.yaml
@@ -5,7 +5,7 @@ change_type: enhancement
 component: operator
 
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Add `AdditionalContainers` to the collector spec allowing to configure sidecar containers. This only applies to Deployment/StatefulSet/DeamonSet deployments of the collector
+note: Add `AdditionalContainers` to the collector spec allowing to configure sidecar containers. This only applies to Deployment/StatefulSet/DeamonSet deployment modes of the collector.
 
 # One or more tracking issues related to the change
 issues: [1987]
--- a/apis/v1alpha1/opentelemetrycollector_types.go
+++ b/apis/v1alpha1/opentelemetrycollector_types.go
@@ -214,10 +214,10 @@ type OpenTelemetryCollectorSpec struct {
 	InitContainers []v1.Container `json:"initContainers,omitempty"`
 
 	// AdditionalContainers allows injecting additional containers into the Collector's pod definition.
-	// These sidecar containers can be used for authentication proxies, logging shipping sidecars, agents for shipping
+	// These sidecar containers can be used for authentication proxies, log shipping sidecars, agents for shipping
 	// metrics to their cloud, or in general sidecars that do not support automatic injection. This option only
-	// applies to Deployment, DaemonSet, and StatefulSet deployments of the collector. It does not apply to the sidecar
-	// method. More info:
+	// applies to Deployment, DaemonSet, and StatefulSet deployment modes of the collector. It does not apply to the sidecar
+	// deployment mode. More info about sidecars:
 	// https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/
 	//
 	// Container names managed by the operator:
--- a/apis/v1alpha1/opentelemetrycollector_types.go
+++ b/apis/v1alpha1/opentelemetrycollector_types.go
@@ -213,6 +213,9 @@ type OpenTelemetryCollectorSpec struct {
 	// +optional
 	InitContainers []v1.Container `json:"initContainers,omitempty"`
 
+	// AdditionalContainers allows injecting containers to run as sidecars in the pod definition.
+	AdditionalContainers []v1.Container `json:"additionalContainers,omitempty"`
+
 	// ObservabilitySpec defines how telemetry data gets handled.
 	//
 	// +optional
--- a/apis/v1alpha1/opentelemetrycollector_types.go
+++ b/apis/v1alpha1/opentelemetrycollector_types.go
@@ -213,7 +213,20 @@ type OpenTelemetryCollectorSpec struct {
 	// +optional
 	InitContainers []v1.Container `json:"initContainers,omitempty"`
 
-	// AdditionalContainers allows injecting containers to run as sidecars in the pod definition.
+	// AdditionalContainers allows injecting additional containers into the Collector's pod definition.
+	// These sidecar containers can be used to authentication proxies, logging shipping sidecars, agents for shipping
+	// to metrics to their cloud, or in general sidecars that do not support automatic injection. This option only
+	// applies to Deployment, DaemonSet, and StatefulSet deployments of the collector. It does not apply to the sidecar
+	// method. More info:
+	// https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/
+	//
+	// Container names managed by the operator:
+	// * `otc-container`
+	//
+	// Overriding containers managed by the operator is outside the scope of what the maintainers will support and by
+	// doing so, you wil accept the risk of it breaking things.
+	//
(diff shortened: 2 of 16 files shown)


TylerHelmuth:
Is there precedence for this kind of option in other operators?

bryan-aguilar:
This seems like something that could be discussed further in the SIG as it's a new feature request. 

An initial question I would have though, why does this fall under the responsibility of the operator? We would probably need more information about your use case and the sidecar agents requirements.

nikore:
> This seems like something that could be discussed further in the SIG as it's a new feature request.
> 
> An initial question I would have though, why does this fall under the responsibility of the operator? We would probably need more information about your use case and the sidecar agents requirements.

I created a ticket #1987 if you want to discuss it under there instead of in the PR. In the ticket I added some use cases that are common to having additional containers set. 

I also linked to some examples of other operators doing this.

To directly answer your question: If not the operator than who else? I am using the operator to provision a cluster of otel collectors that will receive metrics and tracing via OTLP. Using sidecar injection will not help my use case as I have no problem getting the metrics/traces to the otel collectors. 

My specific issue is I need to (dual) ship these metrics to a cloud provider that I cannot name. They require the use of their agent to ship all metrics, they want to scrap the OTel collectors and send the metrics from there. They do not currently support sidecar injection. So my two options are:

1) I manually deploy these 3rd party collectors and try to inject the otel collector. I don't like this as the 3rd party collector is temporary and we want the otel collector to be the primary container in the pod as they will be persistent. 
2) Adjust the otel collector operator to include this container in its deployment. 

I have been running this for a bout a week and its solved our specific problem and it would be great to not have to keep patching the operator manually, and it seems like other operators support this out of the box.

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -213,6 +213,9 @@ type OpenTelemetryCollectorSpec struct {
 	// +optional
 	InitContainers []v1.Container `json:"initContainers,omitempty"`
 
+	// AdditionalContainers allows injecting containers to run as sidecars in the pod definition.
+	AdditionalContainers []v1.Container `json:"additionalContainers,omitempty"`
```
pavolloffay, line 217: If we proceed with this PR we should clarify how this is supported e.g. https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/v1/prometheus_types.go#L337

We also should clarify if it is supported for all deployment modes (e.g. deployment, sidecar...)

jaronoff97 (approved):
Agreed with pavol, also if you could add a changelog entry that would be great. Thank you!

apis/v1alpha1/opentelemetrycollector_types.go
nikore: I updated the godoc comment, I added in an explanation, link to additional information about sidecars, and I added a warning about overriding the container that the operator generates. Let me know if this is good enough or if I need to improve it more.

jaronoff97 (approved):

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -213,6 +213,22 @@ type OpenTelemetryCollectorSpec struct {
 	// +optional
 	InitContainers []v1.Container `json:"initContainers,omitempty"`
 
+	// AdditionalContainers allows injecting additional containers into the Collector's pod definition.
+	// These sidecar containers can be used to authentication proxies, logging shipping sidecars, agents for shipping
```
TylerHelmuth, line 217:
```suggestion
	// These sidecar containers can be used for authentication proxies, logging shipping sidecars, agents for shipping
```

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -213,6 +213,22 @@ type OpenTelemetryCollectorSpec struct {
...
+	// to metrics to their cloud, or in general sidecars that do not support automatic injection. This option only
```
TylerHelmuth, line 218:
```suggestion
	// metrics to their cloud, or in general sidecars that do not support automatic injection. This option only
```

internal/manifests/collector/daemonset_test.go
```diff
@@ -303,3 +303,30 @@ func TestDaemonSetInitContainer(t *testing.T) {
...
+	cfg := config.New()
+
+	// test
+	d := DaemonSet(cfg, logger, otelcol)
+	assert.Equal(t, "my-instance-collector", d.Name)
+	assert.Equal(t, "my-instance-collector", d.Labels["app.kubernetes.io/name"])
+	assert.Equal(t, "true", d.Annotations["prometheus.io/scrape"])
+	assert.Equal(t, "8888", d.Annotations["prometheus.io/port"])
+	assert.Equal(t, "/metrics", d.Annotations["prometheus.io/path"])
+	assert.Len(t, d.Spec.Template.Spec.Containers, 2)
```
TylerHelmuth, line 331: It is kinda a nit, but in the test can we confirm that the `AdditionalContainers` is completely the same as the input?  It would be nice if the test can check that the operator isn't messing with the `AdditionalContainers` configuration in any way.

nikore: OK I added an assertion to confirm that the input to additional container equals the output.

TylerHelmuth (approved):

.chloggen/1987-adding-support-for-sidecar-containers.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `AdditionalContainers` to the collector spec allowing to configure sidecar containers. This only applies to Deployment/StatefulSet/DeamonSet deployments of the collector
```
pavolloffay, line 8: s/deployments of the collector/deployment modes of the collector.

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -213,6 +213,22 @@ type OpenTelemetryCollectorSpec struct {
 	// +optional
 	InitContainers []v1.Container `json:"initContainers,omitempty"`
 
+	// AdditionalContainers allows injecting additional containers into the Collector's pod definition.
+	// These sidecar containers can be used for authentication proxies, logging shipping sidecars, agents for shipping
```
pavolloffay, line 217: perhaps `s/logging shipping/log shipping` I am not a native speaker so not 100 sure :)

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -213,6 +213,22 @@ type OpenTelemetryCollectorSpec struct {
...
+	// metrics to their cloud, or in general sidecars that do not support automatic injection. This option only
+	// applies to Deployment, DaemonSet, and StatefulSet deployments of the collector. It does not apply to the sidecar
```
pavolloffay, line 219: deployment modes

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -213,6 +213,22 @@ type OpenTelemetryCollectorSpec struct {
...
+	// method. More info:
```
pavolloffay, line 220: deployment mode

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -213,6 +213,22 @@ type OpenTelemetryCollectorSpec struct {
...
+	// https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/
```
pavolloffay, line 221: Why there is this link?

internal/manifests/collector/daemonset.go
```diff
@@ -51,7 +51,7 @@ func DaemonSet(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
 				Spec: corev1.PodSpec{
 					ServiceAccountName: ServiceAccountName(otelcol),
 					InitContainers:     otelcol.Spec.InitContainers,
-					Containers:         []corev1.Container{Container(cfg, logger, otelcol, true)},
+					Containers:         append(otelcol.Spec.AdditionalContainers, Container(cfg, logger, otelcol, true)),
```
pavolloffay, line 54: Does the order matter? 

Should be the collector container first?

apis/v1alpha1/opentelemetrycollector_types.go
pavolloffay: I think we should add validation to the validating webhook and return an error to user if the mode is sidecar and there are defined additional containers

pavolloffay (commented):
LGTM: Just some minor comments

apis/v1alpha1/opentelemetrycollector_types.go
nikore: I though it would be helpful to link to more information about sidecars. I can remove it if we don't feel its useful.

jaronoff97: I think it's worth keeping, I appreciate doclinks

internal/manifests/collector/daemonset.go
nikore: The only time container order matters is if any container sets a lifecycle/postStart setting. That will block containers in the list after the postStart until its been executed. 

Outside of that all containers are started concurrently and ordering does not matter. Generally container inter-dependencies should to be handled by sidecars. To give a real example, with istio sidecar injection it, it inserts its container image first but it can take longer to start than the app container.

pavolloffay (approved):

jaronoff97:
Thank you SO much for this contribution!!!

Issue #1969, opened 2023-07-28
Route / Ingress for OTLP gRPC does not work 


pavolloffay:
I was not able to send data to OpenShift route via OTLP gRPC

```yaml
kubectl apply -f - <<EOF
kind: OpenTelemetryCollector
apiVersion: opentelemetry.io/v1alpha1
metadata:
  name: otel
spec:
  mode: deployment
  ingress: 
    type: route
    hostname: apps-crc.testing
    route:
      termination: insecure
  config: |

    receivers:
      otlp:
        protocols:
          grpc:

    exporters:
      logging:

    service:
      pipelines:
        traces:
          receivers: [otlp]
          exporters: [logging]
EOF
```

It creates following route

```yaml
kind: Route
apiVersion: route.openshift.io/v1
metadata:
  name: otlp-grpc-otel-route
  namespace: ploffay
  uid: 78f468b1-d83c-41d5-ac51-a6279895a08b
  resourceVersion: '549420'
  creationTimestamp: '2023-07-28T08:07:01Z'
  labels:
    app.kubernetes.io/instance: ploffay.otel
    app.kubernetes.io/managed-by: opentelemetry-operator
    app.kubernetes.io/name: otlp-grpc-otel-route
  ownerReferences:
    - apiVersion: opentelemetry.io/v1alpha1
      kind: OpenTelemetryCollector
      name: otel
      uid: 06ff039d-36e9-44b6-8074-2d2c563fa202
      controller: true
      blockOwnerDeletion: true
spec:
  host: otlp-grpc.apps-crc.testing
  path: /otlp-grpc
  to:
    kind: Service
    name: otel-collector
    weight: 100
  port:
    targetPort: otlp-grpc
  wildcardPolicy: None
status:
  ingress:
    - host: otlp-grpc.apps-crc.testing
      routerName: default
      conditions:
        - type: Admitted
          status: 'True'
          lastTransitionTime: '2023-07-28T08:07:01Z'
      wildcardPolicy: None
      routerCanonicalHostname: router-default.apps-crc.testing
```


Collector to report data to the route 

```yaml
docker run --rm -it --net=host -v ${PWD}:/tmp ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector:0.81.0 --config /tmp/config.yaml

receivers:
  otlp:
    protocols:
      grpc:
      http:

processors:

exporters:
  otlp:
[...]

PR #1981, opened 2023-08-03
Make sure OTLP exporter can report data to OTLP ingress/route without additional configuration

### LINKED ISSUE #1967: Route / Ingress for OTLP HTTP does not work
The following collector CR 

```yaml
kubectl apply -f - <<EOF
kind: OpenTelemetryCollector
metadata:
  name: otel
spec:
  mode: deployment
  ingress: 
    type: route
    hostname: apps-crc.testing
    route:
      termination: insecure
  config: |

    receivers:
      otlp:
        protocols:
          http:
          grpc:

    exporters:
      logging:

    service:
      pipelines:
        traces:
          receivers: [otlp]
          exporters: [logging]
EOF
```

creates this route:

```yaml
apiVersion: route.openshift.io/v1
kind: Route
metadata:
  creationTimestamp: "2023-07-27T16:41:27Z"
  labels:
    app.kubernetes.io/instance: ploffay.otel
    app.kubernetes.io/name: otlp-http-otel-route
  name: otlp-http-otel-route
  namespace: ploffay
  ownerReferences:
  - apiVersion: opentelemetry.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: OpenTelemetryCollector
    name: otel
    uid: 036cc218-dc9e-4802-9403-c9b0985d193c
  resourceVersion: "313141"
  uid: d400e226-d756-45d3-97fa-7997e15ac262
spec:
  host: otlp-http.apps-crc.testing
  path: /otlp-http
  port:
    targetPort: otlp-http
  to:
    kind: Service
    name: otel-collector
    weight: 100
  wildcardPolicy: None
```

The `path: /otlp-http` is causing issues for setting `OTEL_EXPORTER_OTLP_ENDPOINT` (see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md)

For instance:
```bash
docker run --rm -it --net=host ghcr.io/open-telemetry/opentelemetry-collector-contrib/telemetrygen:v0.74.0 traces --otlp-http --otlp-endpoint=otlp-http.apps-crc.testing:80/otlp-http --otlp-insecure --duration 10s --rate 4                                                                                                   2 ↵ ploffay@fedora

2023-07-27T16:49:34.599Z	INFO	traces/traces.go:68	starting HTTP exporter
2023-07-27T16:49:34.599Z	INFO	traces/traces.go:126	generation of traces is limited	{"per-second": 4}
2023/07/27 16:49:35 traces export: parse "http://otlp-http.apps-crc.testing:80%2Fotlp-http/v1/traces": invalid port ":80%2Fotlp-http" after host
2023/07/27 16:49:36 traces export: parse "http://otlp-http.apps-crc.testing:80%2Fotlp-http/v1/traces": invalid port ":80%2Fotlp-http" after host
```

Removing the path from route and the `--otlp-endpoint` fixed the issue.

pavolloffay:
Resolves #1967 
Updates #1969 - additional PR will be required to set the appProtocol to h2c (http2 over plaintext) or set ingress specific annotation.


apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -48,9 +63,15 @@ const (
...
 	// Type default value is: ""
-	// Supported types are: ingress
+	// Supported types are: ingress, route
 	Type IngressType `json:"type,omitempty"`
 
+	// RuleType defines how Ingress exposes collector receivers.
+	// Either multiple hosts with subdomain per receiver port are created or
+	// a single host with multiple paths.
+	// Default is path.
+	RuleType IngressRuleType `json:"ruleType,omitempty"`
```
pavolloffay, line 73: better naming ideas are appreciated :)

pavolloffay:
@frzifus could you please review?

internal/manifests/collector/ingress.go
```diff
@@ -75,22 +66,76 @@ func Ingress(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelemet
...
 
+func createSubdomainIngressRules(otelcol string, hostname string, ports []corev1.ServicePort) []networkingv1.IngressRule {
+	var rules []networkingv1.IngressRule
+	pathType := networkingv1.PathTypeExact
+	for _, port := range ports {
+		portName := naming.PortName(port.Name, port.Port)
+
+		host := fmt.Sprintf("%s.%s", portName, hostname)
+		if hostname == "" || hostname == "*" {
+			host = portName
```
jaronoff97, line 112: wouldnt this make the host something like "otlp" instead of "otlp.example.com"? should we just error out here?

internal/manifests/collector/ingress.go
```diff
@@ -75,22 +66,76 @@ func Ingress(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelemet
...
+			HTTP: &networkingv1.HTTPIngressRuleValue{
+				Paths: paths,
+			},
 		},
 	}
 }
 
+func createSubdomainIngressRules(otelcol string, hostname string, ports []corev1.ServicePort) []networkingv1.IngressRule {
+	var rules []networkingv1.IngressRule
+	pathType := networkingv1.PathTypeExact
```
jaronoff97, line 106: do we want prefix here? otherwise I think we're only going to match the root.

kind-1.19.yaml
```diff
@@ -3,3 +3,16 @@ apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
 - role: control-plane
   image: kindest/node:v1.19.16@sha256:476cb3269232888437b61deca013832fee41f9f074f9bed79f57e4280f7c48b7
+  kubeadmConfigPatches:
+    - |
+      kind: InitConfiguration
+      nodeRegistration:
+       kubeletExtraArgs:
+         node-labels: "ingress-ready=true"
+  extraPortMappings:
```
jaronoff97, line 12: What're these for?

tests/e2e/ingress/00-install.yaml
```diff
@@ -7,7 +7,8 @@ spec:
   mode: "deployment"
   ingress:
     type: ingress
-    hostname: "example.com"
+    ruleType: subdomain
```
jaronoff97, line 10: can we make this a different ingress test to verify no breaking changes for users?

kind-1.19.yaml
pavolloffay: it is required for ingress

tests/e2e/ingress/00-install.yaml
pavolloffay: done I have created a new test for subdomains

config/crd/bases/opentelemetry.io_opentelemetrycollectors.yaml
```diff
@@ -1370,6 +1370,14 @@ spec:
                         - reencrypt
                         type: string
                     type: object
+                  ruleType:
+                    description: RuleType defines how Ingress exposes collector receivers.
+                      Either multiple hosts with subdomain per receiver port are created
+                      or a single host with multiple paths. Default is path.
```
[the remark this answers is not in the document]
pavolloffay: I have changed the comment, let me know if it is better now

internal/manifests/collector/ingress.go
pavolloffay: it makes sense to error. I will add a validation to the webhook for it

pavolloffay:
@jaronoff97 could you please re-review? The PR is updated based on your comments.

apis/v1alpha1/opentelemetrycollector_webhook_test.go
```diff
@@ -605,6 +605,17 @@ func TestOTELColValidatingWebhook(t *testing.T) {
 			},
 			expectedErr: "the OpenTelemetry Spec LivenessProbe TerminationGracePeriodSeconds configuration is incorrect",
 		},
+		{
+			name: "invalid TerminationGracePeriodSeconds",
```
jaronoff97, line 609:
```suggestion
			name: "invalid ingress hostname",
```

internal/manifests/collector/ingress.go
```diff
@@ -45,24 +45,15 @@ func Ingress(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelemet
...
-		}
+	var rules []networkingv1.IngressRule
+	switch otelcol.Spec.Ingress.RuleType {
+	case v1alpha1.IngressRuleTypePath, "":
+		rules = []networkingv1.IngressRule{createPathIngressRules(otelcol.Name, otelcol.Spec.Ingress.Hostname, ports)}
+	case v1alpha1.IngressRuleTypeSubdomain:
+		rules = createSubdomainIngressRules(otelcol.Name, otelcol.Spec.Ingress.Hostname, ports)
 	}
 
+	createSubdomainIngressRules(otelcol.Name, otelcol.Spec.Ingress.Hostname, ports)
```
jaronoff97, line 56: this is a no-op right now, is that intentional?

jaronoff97 (commented):
A few more questions to confirm my understanding before we're good to go.

apis/v1alpha1/opentelemetrycollector_types.go
frzifus: What about `exposeMethod` instead of `ruleType`?

kind-1.19.yaml
frzifus: Would it still be able to run multiple kind clusters with this change at the same time?

frzifus (commented):
LGTM

kind-1.19.yaml
pavolloffay: it probably opens a port on the hosts, so no

internal/manifests/collector/ingress.go
pavolloffay: removed

internal/manifests/collector/route.go
```diff
@@ -77,15 +78,14 @@ func Routes(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelemetr
...
+				Host: fmt.Sprintf("%s.%s", portName, otelcol.Spec.Ingress.Hostname),
+				Path: "/",
 				To: routev1.RouteTargetReference{
 					Kind: "Service",
 					Name: naming.Service(otelcol.Name),
 				},
 				Port: &routev1.RoutePort{
-					// Valid names must be non-empty and no more than 15 characters long.
-					TargetPort: intstr.FromString(naming.Truncate(p.Name, 15)),
+					TargetPort: intstr.FromString(portName),
```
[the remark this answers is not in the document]
pavolloffay: good catch, this should use the port function from the `naming` package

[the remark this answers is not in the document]
pavolloffay: `portName` is already in the correct form see the line above `portName := naming.PortName(p.Name, p.Port)`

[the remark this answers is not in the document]
jaronoff97: yes, agreed. I meant to confirm that this is changing the naming scheme by moving from only `naming.Truncate` to `naming.PortName`

apis/v1alpha1/opentelemetrycollector_types.go
pavolloffay: Actually, it might be better to keep the name aligned with the upstream ingress settings. 

It seems that users don't like that operators re-invent existing concepts https://github.com/open-telemetry/opentelemetry-operator/issues/1987

pavolloffay:
@jaronoff97 are there any outstanding tasks for this PR?

jaronoff97:
is this going to break existing users because of the truncation change? Is the default still going to be path?

jaronoff97:
kk, sounds good. Could you add that in the subtext in the changelog to make clear that current users with custom parts larger than 15 characters may experience a disruption?

pavolloffay:
I have added it to the release notes

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: ghcr.io/matt/opentelemetry-operator/opentelemetry-operator
-  newTag: 0.82.0-9-g6f4246c
+  newName: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
```
jaronoff97, line 7: should we remove this?

pavolloffay: yes, also the `hcr.io/matt/opentelemetry-operator/opentelemetry-operator` should be removed it got accidentially merged in  https://github.com/open-telemetry/opentelemetry-operator/pull/1980

PR #1983, opened 2023-08-03
Create Service Monitors for the Prometheus exporters

### LINKED ISSUE #1963: Create ServiceMonitors for Prometheus exporters
When the [Prometheus Exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/prometheusexporter) is used, [the operator exposes the port associated to that exporter](https://github.com/open-telemetry/opentelemetry-operator/issues/1931). But, in order to gather those metrics, the user needs to create a `ServiceMonitor` or `PodMonitor` resource.

It would be interesting if the OpenTelemetry Operator could create those resources for the user automatically. To ensure the feature is enabled only when the Prometheus CR is there, all the logic will be protected by the `operator.observability.prometheus` feature gate (implemented as part of #1768).

This will need the work done as part of #1953.

iblancasa:
Resolves #1963

Changes:
- Move the receiver parsers to their own package
- Create an exporter parser package
- Add a parser for the Prometheus exporter
- Refactor the exporting of the Prometheus exporter ports in the Services and Deployments to use the Prometheus Exporter parser
- Create the Service Monitors using the Prometheus Exporter parser
- Added E2E to ensure the ServiceMonitors are created and deleted. Also, check the correct ports are exposed.


config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
```
pavolloffay, line 3: this should  be removed

internal/manifests/collector/adapters/config_to_ports.go
```diff
@@ -24,17 +24,91 @@ import (
...
 	ErrReceiversNotAMap = errors.New("receivers property in the configuration doesn't contain valid receivers")
+
+	// ErrExportersNotAMap indicates that the receivers property isn't a map of values.
+	ErrExportersNotAMap = errors.New("exporters property in the configuration doesn't contain valid exporters")
 )
 
+// ConfigToExporterPorts converts the incoming configuration object into a set of service ports required by the exporters.
+func ConfigToExporterPorts(logger logr.Logger, config map[interface{}]interface{}) ([]corev1.ServicePort, error) {
+	// now, we gather which ports we might need to open
+	// for that, we get all the receivers and check their `endpoint` properties,
```
pavolloffay, line 48: receivers? 

It seems the following lines need to be changed as well from receiver to exporter

internal/manifests/collector/container_test.go
```diff
@@ -81,18 +81,18 @@ service:
...
-			specPorts:     nil,
-			expectedPorts: []corev1.ContainerPort{},
-		},
-		{
-			description:   "couldn't build ports from spec config",
-			specConfig:    "",
-			specPorts:     nil,
-			expectedPorts: []corev1.ContainerPort{metricContainerPort},
-		},
+		// {
```
pavolloffay, line 84: uncoment or remove?

tests/e2e-prometheuscr/create-sm-prometheus-exporters/01-assert.yaml
```diff
@@ -0,0 +1,63 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app.kubernetes.io/instance: create-sm-prometheus.simplest
+    app.kubernetes.io/managed-by: opentelemetry-operator
+    app.kubernetes.io/name: prometheus-dev-collector
+  name: prometheus-dev-collector
+  namespace: create-sm-prometheus
+spec:
```
pavolloffay, line 10: The prometheus adds additional labels during scraping e.g. namespace, pod, container, instance, endpoint see https://github.com/openshift/openshift-docs/pull/60953/files#diff-bd0dc510dd646c60e63a7daf0b9d22eb7525adc8467753a6bdb37948345992ddR9

these labels will identify the OTEL collector. 

Similar labels k8s_{namespace, pod} etc. will be collected as well by OTEL SDKS (or collector) shall we offer the option to remove the prometheus labels?

swiatekm-sumo: FYI, if you use the prometheus receiver instead, you'll get these labels (though using the Otel semantic convention) even if you delete them in the ServiceMonitor. This is because prometheus receiver uses the parsing context directly, see: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/internal/prom_to_otlp.go.

The only way to get rid of them is to remove them in a processor after the fact. I don't think this is a problem in this case, as they should be identical to the attributes from instrumentation. Maybe we should make this consistent and just rename the Prometheus-provided attributes to the Otel schema here?

iblancasa: I tried to rename the labels to the OTEL Schema but... the original labels keep appearing.
Should I remove labels as per the link provided by @pavolloffay? Even when they will be "recovered" if using the Prometheus receiver. Or should we ignore them?

swiatekm-sumo: It would be very strange if you could remove, but not rename them. In the interest of not holding up this PR, the former is ok with me.

iblancasa: @swiatekm-sumo maybe I'm not using the correct rules or something. I tried with `labelmap` and `replace` actions. Both of them created a new label but not removed the old one.
I was not able to find a better solution.

swiatekm-sumo: Looking at the changes, you did the following:

```yaml
- action: replace
  regex: container
  replacement: k8s.container.name
```

but what this action does is match the regex against label *values*, and then copies the *value* of that label to the target. You need to drop the original separately. Try this:

```yaml
- source_labels: [pod]
  action: replace
  target_label: k8s.pod.name
  regex: (.+)
- action: labeldrop
  regex: ^pod$
```

iblancasa: I was talking to @pavolloffay offline and I'll create a new task to address this as part of a separate issue because we're not sure if we want to suggest removing or relabeling.
Thanks for your comment (I'm curious and I'll try it to learn)

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -83,9 +83,9 @@ spec:
       - description: Metrics defines the metrics configuration for operands.
         displayName: Metrics Config
         path: observability.metrics
-      - description: EnableMetrics specifies if ServiceMonitors should be created
-          for the OpenTelemetry Collector. The operator.observability.prometheus feature
-          gate must be enabled to use this feature.
+      - description: EnableMetrics specifies if ServiceMonitor should be created for
```
jaronoff97, line 86:
```suggestion
      - description: EnableMetrics specifies if ServiceMonitors should be created for
```

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -404,7 +404,7 @@ type AutoscalerSpec struct {
 
 // MetricsConfigSpec defines a metrics config.
 type MetricsConfigSpec struct {
-	// EnableMetrics specifies if ServiceMonitors should be created for the OpenTelemetry Collector.
+	// EnableMetrics specifies if ServiceMonitor should be created for the OpenTelemetry Collector and Prometheus Exporters.
```
jaronoff97, line 407:
```suggestion
	// EnableMetrics specifies if ServiceMonitors should be created for the OpenTelemetry Collector and Prometheus Exporters.
```

config/manager/kustomization.yaml
jaronoff97: @iblancasa this is back

internal/manifests/collector/container_test.go
```diff
@@ -181,14 +181,19 @@ service:
...
+            exporters: [prometheus]
+`,
+
 			specPorts: []corev1.ServicePort{},
 			expectedPorts: []corev1.ContainerPort{
 				metricContainerPort,
 				{
 					Name:          "prometheus",
 					ContainerPort: 9090,
-					Protocol:      corev1.ProtocolTCP,
```
jaronoff97, line 191: do we not need the protocol to be set anymore?

internal/manifests/collector/parser/exporter/exporter.go
```diff
@@ -0,0 +1,139 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/open-telemetry/opentelemetry-operator/internal/naming"
+)
+
+// ExporterParser is an interface that should be implemented by all exporter parsers.
```
jaronoff97, line 31: Should we not just make a ComponentPortParser interface that both receivers and exporters implement? Is there anything special about this vs receiver?

internal/manifests/collector/servicemonitor.go
```diff
@@ -51,5 +59,40 @@ func ServiceMonitor(cfg config.Config, logger logr.Logger, otelcol v1alpha1.Open
...
+	exporterPorts, err := adapters.ConfigToExporterPorts(logger, c)
+	if err != nil {
+		logger.Error(err, "couldn't build service monitors from configuration")
+		return []monitoringv1.Endpoint{}
+	}
+
+	endpoints := []monitoringv1.Endpoint{}
+
+	for _, port := range exporterPorts {
+		if strings.Contains(port.Name, "prometheus") {
```
jaronoff97, line 90: shouldn't we only be receiving known exporters by this point? do we need this check?

jaronoff97 (commented):
a few thoughts/questions. Great stuff though :D

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
iblancasa: I modified this because only one ServiceMonitor will be created. I'm not sure if it should be reverted back to ServiceMonitors or not.

internal/manifests/collector/container_test.go
iblancasa: It's TCP by default.

internal/manifests/collector/servicemonitor.go
iblancasa: If somebody adds support for new exporter receivers, they will not be filtered here and we will be creating ServiceMonitors for non-desired exporters.

internal/manifests/collector/parser/exporter/exporter.go
iblancasa: I like this. I started working on it an I think it can be a little bit tricky. Just to avoid making this PR really big... can I work on that as part of a separate issue? Otherwise, I'm fearing we can end up with a big and unmanageable PR.

jaronoff97: yeah! that's totally fine. let's keep this one as is, it's big enough.

internal/manifests/collector/servicemonitor.go
jaronoff97: maybe when you refactor the interface, we should also have a map that parsers can register themselves with for valid servicemonitor ports, though maybe that's overengineering haha

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
jaronoff97: ah i see that now. in that case we should say "a ServiceMonitor"

apis/v1alpha1/opentelemetrycollector_types.go
jaronoff97: same as above :)

jaronoff97:
looks like there's one conflict and a few comments, i think when those are done we're good to merge!

iblancasa:
@jaronoff97 I fixed the conflict :)

tests/e2e-prometheuscr/create-sm-prometheus-exporters/02-assert.yaml
```diff
@@ -0,0 +1,41 @@
...
+spec:
+  endpoints:
+  - port: monitoring
+  - port: prometheus-prod
+  namespaceSelector:
+    matchNames:
+    - create-sm-prometheus
+  selector:
+    matchLabels:
+      app.kubernetes.io/managed-by: opentelemetry-operator
```
pavolloffay, line 19: It seems that this will match all OTEL CRs in a namespace. I think it should be more specific.

iblancasa: Good catch! I added some changes to ensure this not happens

pavolloffay:
@iblancasa could you please rebase and re-generate bundle, CRD and api-docs?

iblancasa:
> @iblancasa could you please rebase and re-generate bundle, CRD and api-docs?

I think now everything is there.

pavolloffay (commented):
LGTM just one nit

jaronoff97:
looks like a conflict now :/

iblancasa:
> looks like a conflict now :/

Solved :) And added some extra E2E testing to ensure no more issues :D

pavolloffay (approved):

PR #1988, opened 2023-08-04
Align management state constant with other constants

pavolloffay:


apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -23,17 +23,17 @@ import (
...
 //
-// +kubebuilder:validation:Enum=Managed;Unmanaged
+// +kubebuilder:validation:Enum=managed;unmanaged
 type ManagementStateType string
 
 const (
 	// ManagementStateManaged when the OpenTelemetryCollector custom resource should be
 	// reconciled by the operator.
-	ManagementStateManaged ManagementStateType = "Managed"
+	ManagementStateManaged ManagementStateType = "managed"
```
pavolloffay, line 32: We don't use constants with capital letters in the codebase.

swiatekm-sumo:
Isn't this a breaking change we should bump the apiVersion for?

iblancasa:
> Isn't this a breaking change we should bump the apiVersion for?

It was added a couple of days ago #1888. It should not be needed, AFAIK

docs/api.md
```diff
@@ -3708,7 +3708,7 @@ OpenTelemetryCollectorSpec defines the desired state of OpenTelemetryCollector.
         <td>
           ManagementState defines if the CR should be managed by the operator or not. Default is managed.<br/>
           <br/>
-            <i>Enum</i>: Managed, Unmanaged<br/>
+            <i>Enum</i>: managed, unmanaged<br/>
             <i>Default</i>: Managed<br/>
```
iblancasa, line 3712:
```suggestion
            <i>Default</i>: managed<br/>
```

iblancasa (commented):
The E2E need to be updated too

pavolloffay:
> Isn't this a breaking change we should bump the apiVersion for?

It's not, the feature was not released

jaronoff97 (approved):

PR #1989, opened 2023-08-04
Enabling collector to mount multiple configMaps

### LINKED ISSUE #1819: Allow the collector CRD to specify a list of configmaps to mount
Right now to mount a configmap, a user needs to specify the volumes they want and mount them manually, it would be great if we could just specify a list of configmaps and maybe a desired fs path to mount them. Prometheus operator supplies it like [this](https://github.com/prometheus-operator/prometheus-operator/blob/e13cdad9305110b3a257ed58369cc8e468280528/pkg/apis/monitoring/v1/prometheus_types.go#L218-L222)

yuriolisa:
Resolves #1819

### DIFF
--- a/apis/v1alpha1/opentelemetrycollector_types.go
+++ b/apis/v1alpha1/opentelemetrycollector_types.go
@@ -227,6 +227,11 @@ type OpenTelemetryCollectorSpec struct {
 	// This is only relevant to statefulset, and deployment mode
 	// +optional
 	TopologySpreadConstraints []v1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+
+	// ConfigMaps is a list of ConfigMaps in the same namespace as the OpenTelemetry
+	// object, which shall be mounted into the Collector Pods.
+	// Each ConfigMap will be added to the Collector's Deployments as a volume named `configmap-<configmap-name>`.
+	ConfigMaps []string `json:"configMaps,omitempty"`
 }
 
 // OpenTelemetryTargetAllocator defines the configurations for the Prometheus target allocator.
--- a/internal/manifests/collector/volume.go
+++ b/internal/manifests/collector/volume.go
@@ -42,5 +42,20 @@ func Volumes(cfg config.Config, otelcol v1alpha1.OpenTelemetryCollector) []corev
 		volumes = append(volumes, otelcol.Spec.Volumes...)
 	}
 
+	if len(otelcol.Spec.ConfigMaps) > 0 {
+		for _, cfgMap := range otelcol.Spec.ConfigMaps {
+			volumes = append(volumes, corev1.Volume{
+				Name: "configmap_"+cfgMap,
+				VolumeSource: corev1.VolumeSource{
+					ConfigMap: &corev1.ConfigMapVolumeSource{
+						LocalObjectReference: corev1.LocalObjectReference{
+							Name: cfgMap,
+						},
+					},
+				},
+			})
+		}
+	}
+
 	return volumes
 }
--- a/internal/manifests/collector/volume_test.go
+++ b/internal/manifests/collector/volume_test.go
@@ -61,3 +61,22 @@ func TestVolumeAllowsMoreToBeAdded(t *testing.T) {
 	// check that it's the otc-internal volume, with the config map
 	assert.Equal(t, "my-volume", volumes[1].Name)
 }
+
+func TestVolumeWithMoreConfigMaps(t *testing.T) {
+	// prepare
+	otelcol := v1alpha1.OpenTelemetryCollector{
+		Spec: v1alpha1.OpenTelemetryCollectorSpec{
+				ConfigMaps: []string{"configmap-test"},
+			},
+		}
+	cfg := config.New()
+
+	// test
+	volumes := Volumes(cfg, otelcol)
+
+	// verify
+	assert.Len(t, volumes, 2)
+
+	// check if the volume with the configmap prefix is mounted after defining the config map.
+	assert.Equal(t, "configmap_configmap-test", volumes[1].Name)
+}


apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -227,6 +227,11 @@ type OpenTelemetryCollectorSpec struct {
 	// This is only relevant to statefulset, and deployment mode
 	// +optional
 	TopologySpreadConstraints []v1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+
+	// ConfigMaps is a list of ConfigMaps in the same namespace as the OpenTelemetry
```
TylerHelmuth, line 231:
```suggestion
	// ConfigMaps is a list of ConfigMaps in the same namespace as the OpenTelemetryCollector
```

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -227,6 +227,11 @@ type OpenTelemetryCollectorSpec struct {
...
+	// object, which shall be mounted into the Collector Pods.
+	// Each ConfigMap will be added to the Collector's Deployments as a volume named `configmap-<configmap-name>`.
+	ConfigMaps []string `json:"configMaps,omitempty"`
```
jaronoff97, line 234: we should also specify where the mount is in the pod as well here

internal/manifests/collector/volume.go
```diff
@@ -42,5 +42,20 @@ func Volumes(cfg config.Config, otelcol v1alpha1.OpenTelemetryCollector) []corev
 		volumes = append(volumes, otelcol.Spec.Volumes...)
 	}
 
+	if len(otelcol.Spec.ConfigMaps) > 0 {
+		for _, cfgMap := range otelcol.Spec.ConfigMaps {
```
jaronoff97, line 46: I think we also need to update the container mounts as well. Otherwise users will need to still manually specify what they want mounted.

jaronoff97: [ex](https://github.com/prometheus-operator/prometheus-operator/blob/9cc6a9dbde4b7c1e8a24df651f36e81a99c342ab/pkg/prometheus/statefulset.go#L349-L363)

jaronoff97 (changes requested):

internal/manifests/collector/container.go
```diff
@@ -107,6 +107,15 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
 		},
 	})
 
+	if len(otelcol.Spec.ConfigMaps) > 0 {
+		for keyCfgMap := range otelcol.Spec.ConfigMaps {
+			volumeMounts = append(volumeMounts, corev1.VolumeMount{
+				Name:      otelcol.Spec.ConfigMaps[keyCfgMap].Name,
+				MountPath: otelcol.Spec.ConfigMaps[keyCfgMap].MountPath + "/" + otelcol.Spec.ConfigMaps[keyCfgMap].Name,
```
TylerHelmuth, line 114: Do we need to append the configmap name to the mountpath?

swiatekm-sumo: The user specifies this explicitly, so I would say no. It's up to them to do the right thing here.

yuriolisa: Just removed the Name's append; I also think it makes sense thinking about the user perspective.

internal/manifests/collector/volume_test.go
```diff
@@ -61,3 +61,30 @@ func TestVolumeAllowsMoreToBeAdded(t *testing.T) {
...
+	otelcol := v1alpha1.OpenTelemetryCollector{
+		Spec: v1alpha1.OpenTelemetryCollectorSpec{
+			ConfigMaps: []v1alpha1.ConfigMapsSpec{{
+				Name:      "configmap-test",
+				MountPath: "/var/conf",
+			},
+				{
+					Name:      "configmap-test2",
+					MountPath: "/var/conf",
+				}},
```
jaronoff97, line 76: formatting on this is a bit odd.

internal/manifests/collector/container_test.go
```diff
@@ -276,6 +276,32 @@ func TestContainerCustomVolumes(t *testing.T) {
...
+	otelcol := v1alpha1.OpenTelemetryCollector{
+		Spec: v1alpha1.OpenTelemetryCollectorSpec{
+			ConfigMaps: []v1alpha1.ConfigMapsSpec{{
+				Name:      "configmap-test",
+				MountPath: "/var/conf",
+			},
+				{
+					Name:      "configmap-test2",
+					MountPath: "/var/conf/dir",
+				},
```
jaronoff97, line 290: also formatting here

jaronoff97 (commented):
one more change, maybe more people (@TylerHelmuth @swiatekm-sumo) have thoughts, but I think we should just choose a path for mounting rather than letting a user choose. Also some comments around formatting. Last ask: can we add a test in each of the deployment modes to verify this change does what we expect?

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -227,6 +227,11 @@ type OpenTelemetryCollectorSpec struct {
 	// This is only relevant to statefulset, and deployment mode
 	// +optional
 	TopologySpreadConstraints []v1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+
+	// ConfigMaps is a list of ConfigMaps in the same namespace as the OpenTelemetryCollector
+	// object, which shall be mounted into the Collector Pods.
+	// Each ConfigMap will be added to the Collector's Deployments as a volume named `configmap-<configmap-name>`.
+	ConfigMaps []ConfigMapsSpec `json:"configmaps,omitempty"`
```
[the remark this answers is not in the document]
TylerHelmuth: If Prometheus sets a precedence for not letting this be configurable then I am ok following that.  The most robust solution though would be to allow it to be configurable and fallback to a default value if no path is provided.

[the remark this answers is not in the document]
swiatekm-sumo: I don't like the prometheus-operator solution. I'd argue it makes things more complex for both the user and the operator code. The user, because they need to both read the documentation and then ensure they're setting the value correctly wherever they need the file name. And for the operator, because it now has to take care of generating the mount path, and the how the path is generated is now part of our public API.

As opposed to the user just setting whatever they want and the operator applying it without any additional checks. Arguably this can lead to footguns where the mount overrides something important, but I think it's the user's responsibility to make sure this doesn't happen - and this may even be a feature in some circumstances.

All that said, if we're following prometheus-operator conventions everywhere else, we should probably do it here as well for consistency if nothing else.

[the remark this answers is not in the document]
jaronoff97: I like this argument :) I think making it configurable with a default if not set seems like a good middle ground here. Does that work for you @swiatekm-sumo?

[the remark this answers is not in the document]
swiatekm-sumo: Making it configurable with a default comes with the worst of both worlds for us, but the best for the users, so I'm fine with it.

lsolovey:
What would be the behavior when the internal content of these specified ConfigMaps is updated? Will OTEL Operator be watching for updates on these ConfigMaps and restart Collectors in order to pickup new changes?

This question is coming from the issue #2070.

Thanks.

internal/manifests/collector/container.go
jaronoff97: Don't we not want the configmap name here?

internal/manifests/collector/container.go
```diff
@@ -104,6 +104,14 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
 		},
 	})
 
+	if len(otelcol.Spec.ConfigMaps) > 0 {
+		for keyCfgMap := range otelcol.Spec.ConfigMaps {
+			volumeMounts = append(volumeMounts, corev1.VolumeMount{
```
jaronoff97, line 109: I think we need to give this a name which matches the volume name

jaronoff97 (changes requested):
I think we should an e2e test for this. Also it looks like we're not using a default mount path which is fine, but just wanted to call it out

internal/manifests/collector/container.go
```diff
@@ -107,6 +107,7 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
 	if len(otelcol.Spec.ConfigMaps) > 0 {
 		for keyCfgMap := range otelcol.Spec.ConfigMaps {
 			volumeMounts = append(volumeMounts, corev1.VolumeMount{
+				Name:      "configmap-" + otelcol.Spec.ConfigMaps[keyCfgMap].Name,
```
jaronoff97, line 110: should we add this to the naming package?

yuriolisa: As we are using to name the default ConfigMaps, I would say to put it together.

jaronoff97 (approved):
This LGTM! I want a review from @swiatekm-sumo or @TylerHelmuth before i merge for some extra eyes on this.

apis/v1alpha1/opentelemetrycollector_types.go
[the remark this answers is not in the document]
TylerHelmuth: What was the outcome of this discussion?  I don't see us using a default value for the path if it is not configured.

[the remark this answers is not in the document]
yuriolisa: I believe the suggestion to make the configMaps' mountPaths configurable is the solution we found better for the users. So, I have changed the PR accordingly.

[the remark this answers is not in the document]
TylerHelmuth: If the user doesn't configure the mountPath or name what happens? Do we need a validating case for that?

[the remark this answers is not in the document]
yuriolisa: I don't think so, I would say both are required to mount extra configMaps.

internal/manifests/collector/container.go
```diff
@@ -104,6 +104,15 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
 		},
 	})
 
+	if len(otelcol.Spec.ConfigMaps) > 0 {
+		for keyCfgMap := range otelcol.Spec.ConfigMaps {
+			volumeMounts = append(volumeMounts, corev1.VolumeMount{
+				Name:      naming.ConfigMapExtra() + otelcol.Spec.ConfigMaps[keyCfgMap].Name,
```
swiatekm-sumo, line 110: `naming.ConfigMapExtra() + otelcol.Spec.ConfigMaps[keyCfgMap].Name` is used both here and in the volume spec, so it should be a separate function imo. Maybe add a parameter to `naming.ConfigMapExtra` and calculate it there?

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -506,6 +511,12 @@ type MetricSpec struct {
 	Pods *autoscalingv2.PodsMetricSource `json:"pods,omitempty"`
 }
 
+type ConfigMapsSpec struct {
+	// Configmap defines name and path where the configMaps should be mounted.
+	Name      string `json:"name,omitempty"`
+	MountPath string `json:"mountpath,omitempty"`
```
swiatekm-sumo, line 517: Both of these should be required and nonempty. Looks like MountPath should be at least `/`, and I'd use some minimum character count for name as well.

internal/manifests/collector/container.go
```diff
@@ -104,6 +104,15 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
...
+				MountPath: otelcol.Spec.ConfigMaps[keyCfgMap].MountPath + "/" + otelcol.Spec.ConfigMaps[keyCfgMap].Name,
```
swiatekm-sumo, line 111: What if MountPath is just `/`? That sounds valid and would break in this case. We should probably use https://pkg.go.dev/path#Join here.

swiatekm-sumo (commented):
Some comments about validation, but everything else looks good.

apis/v1alpha1/opentelemetrycollector_types.go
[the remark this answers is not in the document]
TylerHelmuth: If they are required we should validate that requirement.

internal/manifests/collector/container.go
yuriolisa: Thank you @swiatekm-sumo for pointing out all these topics. I've just changed the PR accordingly. Could you please have a look?

apis/v1alpha1/opentelemetrycollector_types.go
yuriolisa: Regarding the name's length, I've used the `DNSName` function to validate its length.

internal/manifests/collector/container.go
yuriolisa: Following your suggestion, I've created a parameter to validate the name and return it for both places.

swiatekm-sumo (approved):

apis/v1alpha1/opentelemetrycollector_types.go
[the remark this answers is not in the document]
yuriolisa: @TylerHelmuth, I have changed the requirements and added some validations to the configmaps's definition.
When you get a chance could you please have a look?

TylerHelmuth (approved):

PR #1995, opened 2023-08-07
Refactor Reconciliation Flow

### LINKED ISSUE #1876: Isolate basic reconcile functionality into a reusable package

jaronoff97:
This PR closes #1876

## Summary
This PR aims to centralize and de-duplicate reconciliation logic that plagued the codebase. We accomplish this by simply using the manifest generation set up by #1965. I use that logic in mutate.go through a mutation function that's used by controller runtime to handle reconciliation flow. By offloading the responsibility of deletion to the garbage control of the cluster and by centralizing this logic, we will be able to manage CRDs much more effectively with less code in the future.

## Why not feature flag this
After a lot of back and forth between the maintainers, we decided to not feature flag this PR but rather put it in its own release. Reasons:
* This is an internal functionality change that we wouldn't expect a user to enable via a flag
* Keeping the repo in a dual state for reconciliation could lead to massive user pain AND massive developer pain
* The change is 100% opaque to users and we believe our integration and e2e tests are comprehensive enough to catch any real issues

## What should you review
Please review the functionality in the controller, mutate.go, and the new tests to ensure they all make sense. Please let me know what should be documented more.

*THANK YOU!*


internal/status/handle.go
```diff
@@ -0,0 +1,57 @@
...
+	eventTypeNormal  = "Normal"
+	eventTypeWarning = "Warning"
+
+	reasonError         = "Error"
+	reasonStatusFailure = "StatusFailure"
+	reasonInfo          = "Info"
+)
+
+// HandleReconcileStatus handles updating the status of the CRDs managed by the operator.
+// TODO: make the status more useful https://github.com/open-telemetry/opentelemetry-operator/issues/1972
```
jaronoff97, line 38: I have a todo here because I think we should refactor this logic to make it simpler and more extendable. Will probably do after this as a solo improvement

pkg/featuregate/featuregate.go
```diff
@@ -78,6 +78,15 @@ var (
 		featuregate.WithRegisterDescription("enables features associated to the Prometheus Operator"),
 		featuregate.WithRegisterFromVersion("v0.82.0"),
 	)
+
+	// EnableTargetAllocatorRewrite is the feature gate that controls whether the collector's configuration should
+	// automatically be rewritten when the target allocator is enabled.
+	UseManifestReconciliation = featuregate.GlobalRegistry().MustRegister(
+		"operator.manifeststrategy",
+		featuregate.StageStable,
```
jaronoff97, line 86: Obviously I'm going to change this, but i wanted to see if all of our e2e tests would pass with this change first so I didn't need to duplicate anything

internal/manifests/collector/suite_test.go
```diff
@@ -15,170 +15,31 @@
...
-	err         error
-	cfg         *rest.Config
 )
 
 const (
 	defaultCollectorImage    = "default-collector"
 	defaultTaAllocationImage = "default-ta-allocator"
 )
 
-func TestMain(m *testing.M) {
```
jaronoff97, line 72: this was actually not needed at all and was slowing down tests

internal/manifests/mutate.go
```diff
@@ -0,0 +1,349 @@
...
+// - Deployment
+// - DaemonSet
+// - StatefulSet
+// - ServiceMonitor
+// - Ingress
+// - HorizontalPodAutoscaler
+// - Route
+// - Secret
+// In order for the operator to reconcile other types, they must be added here.
+func MutateFuncFor(existing, desired client.Object) controllerutil.MutateFn {
```
jaronoff97, line 59: This handles all reconciliation now

internal/manifests/collector/collector.go
```diff
@@ -53,7 +55,7 @@ func Build(params manifests.Params) ([]client.Object, error) {
 		res, err := factory(params.Config, params.Log, params.Instance)
 		if err != nil {
 			return nil, err
-		} else if res != nil {
+		} else if res != nil && !reflect.ValueOf(res).IsNil() {
```
jaronoff97, line 58: Is there a better way to do this? I don't think so....

jaronoff97 (commented):
While the PR may look big and scary, most of it is tests or deleting dead code. The real stuff to look at:
* the doCrud method
* the mutate.go file
* the reflection in collector.go

internal/manifests/collector/collector.go
frzifus: Should a factory ever being nil? especially, if the `err` is not nil? :no_good: 

That could be an alternative with the same outcome:
https://go.dev/play/p/7t-9SOo9FpX

internal/manifests/collector/suite_test.go
frzifus: How is the reconcile loop then tested?

jaronoff97: entirely within the new reconcile_test. This block was inside of the collector's manifest which is all essentially statically generated

internal/manifests/collector/collector.go
jaronoff97: Ill link the comment from @swiatekm-sumo later, am afk... but IIRC In this case its because the result is an interface pointer to a nil structure so we need to check that the place it points to is actually nil.

jaronoff97: I think i'll do your suggestion though

jaronoff97: https://github.com/open-telemetry/opentelemetry-operator/pull/1965#discussion_r1281705719

jaronoff97: unfortunately this didn't work for this case :/
Screenshot 2023-09-13 at 11 56 56 AM

jaronoff97: I'm going to create this as a re-usable function that we can resolve in the future.

controllers/opentelemetrycollector_controller.go
```diff
@@ -126,63 +135,22 @@ func NewReconciler(p Params) *OpenTelemetryCollectorReconciler {
...
-			},
-		}
+		// TODO: put this in line with the rest of how we generate manifests
+		// https://github.com/open-telemetry/opentelemetry-operator/issues/2108
 		r.config.RegisterOpenShiftRoutesChangeCallback(r.onOpenShiftRoutesChange)
 	}
 	return r
 }
 
+// +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
```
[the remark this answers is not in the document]
jaronoff97: sure, i was avoiding this because it causes the kubebuilder resource to be modified. but i can do this

internal/manifests/mutate.go
```diff
@@ -0,0 +1,349 @@
...
+func mutateService(existing, desired *corev1.Service) error {
+	existing.Spec.Ports = desired.Spec.Ports
+	if err := mergeWithOverride(&existing.Spec.Selector, desired.Spec.Selector); err != nil {
+		return err
+	}
+	return nil
+}
+
+func mutateDaemonset(existing, desired *appsv1.DaemonSet) error {
+	if !existing.CreationTimestamp.IsZero() && !apiequality.Semantic.DeepEqual(desired.Spec.Selector, existing.Spec.Selector) {
```
swiatekm-sumo, line 251: In general, do we not have a better way of checking if this actually exists than the creation timestamp?

swiatekm-sumo (commented):
I've not read the full thing yet, but thus far it looks very reasonable. Comments thus far are mostly nitpicks.

controllers/opentelemetrycollector_controller.go
```diff
@@ -246,10 +202,144 @@ func (r *OpenTelemetryCollectorReconciler) RunTasks(ctx context.Context, params
...
+			return nil, err
+		}
+		resources = append(resources, objs...)
+	}
+	return resources, nil
+}
+
+func (r *OpenTelemetryCollectorReconciler) doCRUD(ctx context.Context, params manifests.Params) error {
+	// Collect all objects owned by the operator, to be able to prune objects
+	// which exist in the cluster but are not managed by the operator anymore.
```
frzifus, line 238: Can we move this commend on top of the method? (Then it gets displayed in the IDE popup :rabbit: )

internal/manifests/builder.go
```diff
@@ -39,3 +41,10 @@ func Factory[T client.Object](f ManifestFactory[T]) K8sManifestFactory {
 		return f(cfg, logger, otelcol)
 	}
 }
+
+// ShouldCreateObject ensures that we only create an object IFF it isn't nil
+// and it's concrete type isn't nil either. This works around the Go type system
+// by using reflection to verify its concrete type isn't nil.
+func ShouldCreateObject(obj client.Object) bool {
```
frzifus, line 48: While the name `ShouldCreateObject` makes sense everywhere in the caller context used here, individually I feel it is a bit misleading.

What about `IsNil` or `ObjectIsNil` (alternatively, the negative form).

frzifus (commented):
There is no diff in the `mutate.go`, did you just move it to the new location?

Inspected it a bit locally. Overall it looks good! But it might be better if a few more eyes look at it. :+1:

internal/manifests/mutate.go
jaronoff97: doing it this way is the standard method for operators IIRC.

jaronoff97: [tempo example](https://github.com/grafana/tempo-operator/blob/main/internal/manifests/mutate.go#L207-L211)

internal/status/handle.go
```diff
@@ -0,0 +1,72 @@
...
+		Version:  version.Get(),
+		Client:   params.Client,
+		Recorder: params.Recorder,
+	}
+	upgraded, upgradeErr := up.ManagedInstance(ctx, *changed)
+	if upgradeErr != nil {
+		// don't fail to allow setting the status
+		params.Log.Error(upgradeErr, "failed to upgrade the OpenTelemetry CR")
+	}
+	changed = &upgraded
```
jaronoff97, line 60: @iblancasa this is where i added the logic from your recent PR... In a followup to this we should probably change the upgrade logic to use pointers rather than the way we do it now to keep it in line with the rest of our signatures.

controllers/opentelemetrycollector_controller.go
```diff
@@ -114,6 +123,124 @@ func (r *OpenTelemetryCollectorReconciler) removeRouteTask(ora autodetect.OpenSh
...
+		if isNamespaceScoped(obj) {
+			if setErr := ctrl.SetControllerReference(&params.Instance, obj, params.Scheme); setErr != nil {
+				l.Error(setErr, "failed to set controller owner reference to resource")
+				errs = append(errs, setErr)
+				continue
+			}
+		}
+
+		desired := obj.DeepCopyObject().(client.Object)
+		mutateFn := manifests.MutateFuncFor(obj, desired)
```
swiatekm-sumo, line 153: After rereading this, I think the naming is confusing. `managedObjects` is really the *desired* objects, and inside the loop we copy the loop variable and name that *desired*. The values we pass to `MutateFuncFor` are identical anyway, so this ends up not mattering, but it makes understanding this code unnecessarily difficult imo. How about:

```suggestion
	desiredObjects, err := r.BuildAll(params)
	if err != nil {
		return err
	}
	var errs []error
	for _, desired := range desiredObjects {
		l := r.log.WithValues(
			"object_name", obj.GetName(),
			"object_kind", obj.GetObjectKind(),
		)

		if isNamespaceScoped(obj) {
			if setErr := ctrl.SetControllerReference(&params.Instance, obj, params.Scheme); setErr != nil {
				l.Error(setErr, "failed to set controller owner reference to resource")
				errs = append(errs, setErr)
				continue
			}
		}
                
                // existing is an object the controller runtime will hydrate for us
                // we obtain it by deep copying the desired object because it's the most convenient way
		existing := obj.DeepCopyObject().(client.Object)
		mutateFn := manifests.MutateFuncFor(existing, desired)
```

controllers/opentelemetrycollector_controller.go
```diff
@@ -114,6 +123,124 @@ func (r *OpenTelemetryCollectorReconciler) removeRouteTask(ora autodetect.OpenSh
...
+		routesList := &routev1.RouteList{}
+		err = r.List(ctx, routesList, listOps)
+		if err != nil {
+			return nil, fmt.Errorf("error listing routes: %w", err)
+		}
+		for i := range routesList.Items {
+			ownedObjects[routesList.Items[i].GetUID()] = &routesList.Items[i]
+		}
+	}
+	return ownedObjects, nil
```
swiatekm-sumo, line 230: Am I misunderstanding something, or does this function only return Ingresses and Routes? Shouldn't it return all owned objects? Is there a test for this?

swiatekm-sumo (commented):
Reviewed the critical path again and I've got 2 questions/suggestions. Everything else looks good to me.

controllers/opentelemetrycollector_controller.go
jaronoff97: yeah i like that! adding...

controllers/opentelemetrycollector_controller.go
jaronoff97: so we do this only for objects that can be conditionally created by the operator based on a flag. This was pulled from [the tempo-operator](https://github.com/grafana/tempo-operator/blob/main/controllers/tempo/tempostack_create_or_update.go#L223) but on second look, im also not sure we need to do this. Let me run some tests.

swiatekm-sumo: And with this change, that happens in the `ctrl.SetControllerReference` call in the reconciliation loop, right? In that case, we shouldn't need to do anything special for routes and ingressses.

jaronoff97: yeah i agree, im going to test removing this logic now :) if only my kind cluster liked me

controllers/opentelemetrycollector_controller.go
jaronoff97: so i had to change this a bit, because the mutate you had actually reversed the parameters. What i did was just renamed desiredObjects to expectedObjects which is more accurate.

swiatekm-sumo (approved):

pavolloffay (approved):

internal/status/handle.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	changed := params.Instance.DeepCopy()
+
+	up := &collectorupgrade.VersionUpgrade{
+		Log:      params.Log,
+		Version:  version.Get(),
+		Client:   params.Client,
+		Recorder: params.Recorder,
+	}
+	upgraded, upgradeErr := up.ManagedInstance(ctx, *changed)
```
jigarkb, line 55: @jaronoff97 This change has made the `upgradeStrategy` setting ineffective.

PR #1998, opened 2023-08-07
Expose the pprof server in the operator

### LINKED ISSUE #1997: Expose the pprof server in the operator
The [`controller-runtime`](https://github.com/kubernetes-sigs/controller-runtime/pull/1943) dependency added a `pprof` server flag that can be enabled to get information from the operator.

This issue is to add support to the operator to expose that information. The feature should be enabled just for debugging purposes.

iblancasa:
Resolves #1997


main.go
```diff
@@ -107,6 +108,7 @@ func main() {
 
 	pflag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	pflag.StringVar(&probeAddr, "health-probe-addr", ":8081", "The address the probe endpoint binds to.")
+	pflag.StringVar(&pprofAddr, "pprof-addr", "", "The address to expose the pprof server.")
```
jaronoff97, line 111: should we set a default here?

iblancasa: The default is an empty string. When the empty string is provided to the manager, the server is not started (the default).

jaronoff97: can we add that to the docstring here? something like "default is empty string which disables the pprof server"

.chloggen/1997-expose-pprof-server.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add support to expose the pprof server in the operator
```
frzifus, line 8:
```suggestion
note: Add flag to enable support for the pprof server in the operator.
```

main.go
iblancasa: Sure!

PR #2009, opened 2023-08-10
Use golang 1.21

pavolloffay:


jaronoff97 (approved):

cmd/otel-allocator/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator
 
-go 1.20
+go 1.21
```
Aneurysm9, line 3: Is it necessary to increase the versions here?  This will prevent building with a Go toolchain less than 1.21.  I don't think we're taking any dependencies on 1.21 here.

frzifus (approved):

swiatekm-sumo (commented):
I agree with @Aneurysm9 that we shouldn't bump the required version in go.mod files. Other than that LGTM.

cmd/otel-allocator/go.mod
pavolloffay: +1

PR #2010, opened 2023-08-10
Use distroless base for target allocator and opamp bridge images

### LINKED ISSUE #2011: Use the same base image for all of our Docker images
Currently we use distroless for the manager, but scratch for the aforementioned. We should use the same base everywhere, if only to be consistent. I vote for distroless because it leads to simpler Dockerfiles, but I can live with scratch as well.

swiatekm-sumo:
Resolves #2011

### DIFF
--- a/.chloggen/chore_opamp_bridge_base.yaml
+++ b/.chloggen/chore_opamp_bridge_base.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: operator
+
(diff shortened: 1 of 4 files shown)


bryan-aguilar:
Oops, commented on the issue rather than this PR. copy/pasting what I put into the issue here. 

> I'd vote for scratch just for the sake of reducing the security surface area even further. Is there a reason to prefer distroless over scratch other than consistency?

jaronoff97 (approved):

iblancasa (approved):

bryan-aguilar (changes requested):
I would prefer to see scratch image used instead of distroless. Scratch reduces the security surface area. Unless there is a strong reason to use distroless I believe that is a better path forward.

swiatekm-sumo:
Closing in favor of #2014

PR #2013, opened 2023-08-11
get servicename from daemonset if possible

angelopoerio:
Daemonsets can be instrumented so the generated servicename should use their name for better discoverability.

Related issue:  https://github.com/open-telemetry/opentelemetry-operator/issues/2015


jaronoff97:
@angelopoerio thanks for your PR, could you add a changelog entry when you get chance please?

angelopoerio:
@jaronoff97 changelog is added

jaronoff97 (approved):

PR #2014, opened 2023-08-11
Use scratch as the base image for operator

### LINKED ISSUE #2011: Use the same base image for all of our Docker images

swiatekm-sumo:
Fixes #2011, an alternative to #2010.

### DIFF
--- a/.chloggen/chore_dockerfile-base-scratch.yaml
+++ b/.chloggen/chore_dockerfile-base-scratch.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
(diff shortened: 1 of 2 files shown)


bryan-aguilar (approved):
👍 🎊

jaronoff97 (approved):

TylerHelmuth (approved):

jvoravong (approved):

jaronoff97 (approved):

jaronoff97:
I'm going to hold off on merge until we decide whether or not to do a patch or 0.82.0 because i don't want a dockerfile change in that patch.

PR #2017, opened 2023-08-13
Fix getPrometheusExporterPorts incorrectly matching to prometheusremotewrite exporter (#2016)

hzhhong:
fix #2016

PR #2029, opened 2023-08-17
fix reconciliation error when prometheusremotewrite exporter is enabled

lauri-paypay:
https://github.com/open-telemetry/opentelemetry-operator/pull/1953 introduced the following error when prometheusremotewrite exporter is enabled

	failed to reconcile the expected deployments: failed to apply changes: Deployment.apps "cluster-collector" is invalid: spec.template.spec.containers[0].ports[3].containerPort: Required value"

required configuration is:

	exporters:
	  prometheusremotewrite:
	    endpoint: "https://example.com/api/v1/write"

getPrometheusExporterPorts() tries to parse this endpoint into a host:port pair, because the exporter name contains the string "prometheus", but prometheusremotewriteexporter's expected `endpoint` configuration is actually a URL, so that fails. additionally, due to a missing `continue` if the port parsing fails, an erroneous ContainerPort entry with port set to 0 is added to the configuration, causing the error.

### DIFF
--- a/internal/manifests/collector/container.go
+++ b/internal/manifests/collector/container.go
@@ -220,7 +220,7 @@ func getPrometheusExporterPorts(c map[interface{}]interface{}) ([]corev1.Contain
 	if exporters, ok := c["exporters"]; ok && exporters != nil {
 		for e, exporterConfig := range exporters.(map[interface{}]interface{}) {
 			exporterName := e.(string)
-			if strings.Contains(exporterName, "prometheus") {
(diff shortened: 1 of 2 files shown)


swiatekm-sumo:
Looks like a duplicate of #2017, can you have a look and confirm @lauri-paypay ?

lauri-paypay:
@swiatekm-sumo you're right, it's duplicate. closing this one.

PR #2031, opened 2023-08-17
Bump k8s dependencies to 0.28.0

swiatekm-sumo:
Bumping all the k8s dependencies together. This can't be done by dependabot, as these cause additions to our CRDs, where we embed apimachinery structures, like container specs in this case.

Should be bump the CRD version here? There's one field removal, which is technically a breaking change even if it comes from upstream K8s.

This replaces #2021 #2022 #2023 and #2024.


TylerHelmuth (approved):

swiatekm-sumo:
Apparently the removed field is a part of PersistentVolumeClaim which containts a readonly status. So this isn't a breaking change in the CRD, as you couldn't set this in the first place. I think this is mostly a quirk of how this field works for StatefulSets.

jaronoff97 (approved):

PR #2045, opened 2023-08-23
Limit auto-instrumentation volume size

### LINKED ISSUE #2044: Volume "opentelemetry-auto-instrumentation" of type emptyDir of init container does not have sizeLimit
Hi all,
When using _instrumentation.opentelemetry.io/inject-java: "true"_ in Kubernetes deployment yaml the used volume has no sizeLimit:
```
volumes:
...
  - emptyDir: {}
    name: opentelemetry-auto-instrumentation
...
```
A common kyverno rule (require-emptydir-requests-and-limits) checks for a sizeLimit for emptyDir volumes and therefore this kyverno rule reports a policy violation.
Would it make sense to add sizeLimit to this volume. 
Thanks in advance,
Kai

pavolloffay:
Resolves #2044


pkg/instrumentation/sdk.go
```diff
@@ -44,6 +45,8 @@ const (
 	sideCarName       = "opentelemetry-auto-instrumentation"
 )
 
+var volumeLimitSize = resource.MustParse("150Mi")
```
pavolloffay, line 48: The python images is the largest one ~100mb. I am using a larger number to get some buffer

jaronoff97: Should we make this configurable? I'm wondering what may happen for people building their own images that are larger than the otel defaults

pavolloffay: good point, I will make it configurable

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -31,13 +31,13 @@ metadata:
...
     containerImage: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
-    createdAt: "2023-08-18T14:30:49Z"
+    createdAt: "2023-08-25T10:42:03Z"
     description: Provides the OpenTelemetry components, including the Collector
     operators.operatorframework.io/builder: operator-sdk-v1.29.0
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
     repository: github.com/open-telemetry/opentelemetry-operator
     support: OpenTelemetry Community
-  name: opentelemetry-operator.v0.83.0
+  name: opentelemetry-operator.v0.83.0-5-ge28255d
```
jaronoff97: looks like this bundle got committed

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: docker.io/pavolloffay/opentelemetry-operator
```
jaronoff97: should remove

pkg/instrumentation/helper.go
```diff
@@ -43,3 +44,11 @@ func isAutoInstrumentationInjected(pod corev1.Pod) bool {
 	}
 	return false
 }
+
+func volumeSize(quantity *resource.Quantity) *resource.Quantity {
+	defaultSize := resource.MustParse("150Mi")
```
jaronoff97, line 49: Should this be a constant or in a vars section at the top?

pavolloffay: done

.chloggen/limit-volume-inst.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
```
jaronoff97, line 2: I think we should consider this an enhancement :)

jaronoff97 (approved):

.chloggen/limit-volume-inst.yaml
pavolloffay: +1 changed

PR #2052, opened 2023-08-24
[chore] Fix ci runner

jaronoff97:
Looks like we accidentally combined e2e and upgrade which meant upgrade never ran and e2e ran twice

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -28,7 +28,8 @@ jobs:
        - "1.19"
        - "1.27"
        group:
-         - e2e e2e-upgrade
+         - e2e
+         - e2e-upgrade
          - e2e-prometheuscr
          - e2e-autoscale
 


TylerHelmuth (commented):
If this passes I'm all for it, but during my testing I found these two suites have some sort of relationship that I couldn't figure out. They are both being run because this entry in the matrix becomes the args for a make command later in the test.

jaronoff97:
interesting... i don't think the upgrade test ever ran (see this run of [mine](https://github.com/open-telemetry/opentelemetry-operator/actions/runs/5929868046/job/16082585990?pr=1995)) whereas when i ran locally, this is the expected output:
```
--- PASS: kuttl (26.79s)
    --- PASS: kuttl/harness (0.00s)
        --- PASS: kuttl/harness/upgrade-test (21.25s)
```

TylerHelmuth:
They do run, they just need the `e2e` make target to pass first: https://github.com/open-telemetry/opentelemetry-operator/actions/runs/5953179869/job/16146835304.

But since this PR is passing with them separated that is way better

PR #2060, opened 2023-08-28
Fixes reconciling Internal Traffic Policy changes and adds test

### LINKED ISSUE #2061: Changing deploymentmode for otel-collector from daemonset to something else breaks service
When changing deployment Mode of the otel-collector from daemonset to e.g. statefulset, the Internal Traffic Policy for the services should change. Unfortunately, the operator does not reconcile changes on the ITP, so the service ends up having `internalTrafficPolicy: Local`. Accessing the service from other nodes therefore fails.

cmergenthaler:
This PR fixes #2061 and adds a unit test for that case.


TylerHelmuth (approved):

jaronoff97 (approved):

jaronoff97:
Thank you for your contribution :D

PR #2069, opened 2023-08-29
Allow the Apache HTTP instrumentation image to be used for non-root users

### LINKED ISSUE #2068: Apache HTTPD autoinstrumentation cannot be run with non-root user because of the permissions in the instrumentation image
For the Apache HTTPD autoinstrumentation, we copy some stuff from one of the init containers. When using an user different to root, the files cannot be copied (because the user will not have permissions to that).

iblancasa:
Resolves #2068.


autoinstrumentation/apache-httpd/Dockerfile
```diff
@@ -20,6 +20,7 @@ FROM alpine:latest
 
 COPY --from=agent /opt/opentelemetry/agent/opentelemetry-webserver-sdk /opt/opentelemetry
 
+RUN chmod 775 -R /opt/opentelemetry/
```
TylerHelmuth, line 23: @chrlic can you confirm this change is safe.

iblancasa: If this fix is not fine, I'll be happy implementing another approach but... I'll need some hints 😅

TylerHelmuth: Reading through the issue, I think you're right that this is totally fine.

TylerHelmuth (approved):

autoinstrumentation/apache-httpd/Dockerfile
chrlic: Yes this is safe. I'd do it the same way.

pavolloffay (approved):

PR #2059, opened 2023-08-25
Use openshift base domain if ingress hostname is missing

pavolloffay:
Docker image: docker.io/pavolloffay/opentelemetry-operator:dev-756ef72-1693233303 

Updates: https://github.com/open-telemetry/opentelemetry-operator/issues/1967 
Updates: https://github.com/open-telemetry/opentelemetry-operator/issues/1969

PR #2074, opened 2023-08-31
Make OpenShift routes work with missing hostname

pavolloffay:
Supersedes https://github.com/open-telemetry/opentelemetry-operator/pull/2059 

This is much simpler than ^^. The hostname and path in route are optional, the route controller is responsible for setting them (it uses as well basedomain and DNS) and we should rely on that rather than having that additional code in this operator.

image: docker.io/pavolloffay/opentelemetry-operator:dev-b04cb0f-1693500951

Updates: https://github.com/open-telemetry/opentelemetry-operator/issues/1967
Updates: https://github.com/open-telemetry/opentelemetry-operator/issues/1969

Example routes created with this patch (`otel4` is the CR name)
```
NAME                    HOST/PORT                                        PATH   SERVICES          PORT        TERMINATION   WILDCARD
otlp-grpc-otel4-route   otlp-grpc-otel4-route-ploffay.apps-crc.testing          otel4-collector   otlp-grpc   passthrough   None
otlp-http-otel4-route   otlp-http-otel4-route-ploffay.apps-crc.testing          otel4-collector   otlp-http   passthrough   None

```


frzifus (commented):
Not sure if fully got it. Please correct me:

> by removing the hostname entry from the route, we automatically pick the default hostname?

If thats the case, do we still support setting custom hostnames?

pavolloffay:
Good catch, the custom hostnames are still supported. I have added some code back and added a test.

iblancasa (approved):

.chloggen/route-use-defaults.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
```
frzifus, line 2: Since setting a hostname is still supported, its an enhancement, right?

.chloggen/route-use-defaults.yaml
```diff
@@ -13,4 +13,4 @@ issues: [2074]
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
-subtext: The OpenShift route hostnames change to `<port-name>-<otel-cr-name>-route-<otel-cr-namespace>-basedomain` e.g. from `otlp-http.apps-crc.testing` to `otlp-http-otel-route-ploffay.apps-crc.testing`.
+subtext: If the ingeress hostname is not specified OpenShift route hostname is set to `<port-name>-<otel-cr-name>-route-<otel-cr-namespace>-basedomain`.
```
iblancasa, line 16:
```suggestion
subtext: If the Ingress hostname is not specified OpenShift route hostname is set to `<port-name>-<otel-cr-name>-route-<otel-cr-namespace>-basedomain`.
```

frzifus (approved):

Issue #2073, opened 2023-08-31
Use business applications hosted in this repository for the instrumentation E2E tests


iblancasa:
Currently, we use some applications to test the autoinstrumentation injection. Those applications are hosted by others.

When we find a problem in those applications or their images, we depend on a third one to fix the image and make the test pass.

Also, some images are really big. Doing this can do the CI a little bit faster.

Finally, it would be awesome if we could move the instrumentation E2E tests to their own folder. It makes sense from the logical organization of the E2E tests and can help to speedup the CI since the run will be done in parallel.

TylerHelmuth:
I would also support using the otel-demo components for our tests instead of building our own test images.

iblancasa:
> I would also support using the otel-demo components for our tests instead of building our own test images.

Good idea. I'll try and, if not, use something custom.
Thanks for the suggestion!

pavolloffay:
> I would also support using the otel-demo components for our tests instead of building our own test images.

LGTM, if the image size is not large. 

The goal here is to improve maintenance and as well speed up CI (e.g. image size).

Internally we will also start testing on different architectures and we will need the test apps to run on those archs as well.

PR #2082, opened 2023-09-04
[chore] Build and publish application hosted in the repository for the DotNet instrumentation E2E test

iblancasa:
Related to https://github.com/open-telemetry/opentelemetry-operator/issues/2073.

The test migration will be done in a follow up PR after publishing the image.


TylerHelmuth (commented):
From the issue, @pavolloffay mentions one of the goals is to `improve maintenance`, and I don't think introducing our own test app is going to help with that.  If we introduce a test app for each auto-instrumentation language we'll end up with a repo with many different languages to maintain and then the burden of maintaining those test apps. I'd much rather depend on pre-existing, small, stable test apps.

pavolloffay:
>  @pavolloffay do you think there's value in publishing multiple app images we only maintain for E2E tests?

Is there a better alternative? We need a small test app that can run in isolation and the image is published for (amd64, arm64, s390x, ppc64). 



This PR can be made much simpler. Only a single controller is needed, we don't need to include JS and HTLM code.

iblancasa:
> This PR can be made much simpler. Only a single controller is needed, we don't need to include JS and HTLM code.

I just wanted to keep the application as it was created by the cli. I'll remove the "extra" files.

swiatekm-sumo:
> > @pavolloffay do you think there's value in publishing multiple app images we only maintain for E2E tests?
> 
> Is there a better alternative? We need a small test app that can run in isolation and the image is published for (amd64, arm64, s390x, ppc64).
> 
> This PR can be made much simpler. Only a single controller is needed, we don't need to include JS and HTLM code.

Why do we need it to be published? So it's possible to run tests on different archs than Github Runners support, outside of the project's normal CI process?

pavolloffay:
> Why do we need it to be published? 

I don't have a strong opinion, but those tests app won't change much (at least they have never changed in our e2e tests). If they are published it will speed up CI. If they are published we could also use them for debugging.

swiatekm-sumo:
> > Why do we need it to be published?
> 
> I don't have a strong opinion, but those tests app won't change much (at least they have never changed in our e2e tests). If they are published it will speed up CI. If they are published we could also use them for debugging.

I'm less concerned about the maintenance burden of the app itself, and more about the burden of maintaining these artifacts. Every additional CI workflow that publishes something to GHCR is another opportunity for issues that we'd then need to spend time fixing.

In some respect, the fact that the workflow in this PR only runs when the app is changed helps with this, but then we'll get all the issues at once when we make a change a year from now. Maybe this is ok, though... @iblancasa do you have a sense of how many of these apps we're going to need, this one included?

iblancasa:
> do you have a sense of how many of these apps we're going to need, this one included?

@swiatekm-sumo I would say these ones:
- [Apache HTTPD](https://github.com/open-telemetry/opentelemetry-operator/pull/2085)
- [Java](https://github.com/open-telemetry/opentelemetry-operator/pull/2094)
- [DotNet](https://github.com/open-telemetry/opentelemetry-operator/pull/2082)
- [Python](https://github.com/open-telemetry/opentelemetry-operator/pull/2095)
- Go (I'm on it)

swiatekm-sumo:
I'd like to discuss this during the SIG meeting tomorrow to make sure we have consensus among maintainers. Sorry for holding this PR up with this @iblancasa, the actual content of it looks good to me.

swiatekm-sumo (approved):
LGTM

iblancasa:
Can we merge this?

jaronoff97:
Is the plan for the other jobs to just extend this one?

iblancasa:
> Is the plan for the other jobs to just extend this one?

Sorry. I just saw this comment.

I did a reusable workflow to make everything easier.

PR #2085, opened 2023-09-05
[chore] Build and publish the Apache HTTPD image for the E2E tests

iblancasa:
Related to #2073.

Blocked by  #2084.


swiatekm-sumo (commented):
This seems a bit overkill. If the only reason we're using this image is that it can run as non-root, then how about https://hub.docker.com/r/bitnami/apache instead? Also non-root, binds to port 8080 by default, and the port can be configured via an env variable.

iblancasa:
> This seems a bit overkill. If the only reason we're using this image is that it can run as non-root, then how about https://hub.docker.com/r/bitnami/apache instead? Also non-root, binds to port 8080 by default, and the port can be configured via an env variable.

That image only supports amd64 and arm64 architectures while the one this PR creates support arm64, amd64, s390x, and ppc64le.

swiatekm-sumo (approved):

swiatekm-sumo:
> > This seems a bit overkill. If the only reason we're using this image is that it can run as non-root, then how about https://hub.docker.com/r/bitnami/apache instead? Also non-root, binds to port 8080 by default, and the port can be configured via an env variable.
> 
> That image only supports amd64 and arm64 architectures while the one this PR creates support arm64, amd64, s390x, and ppc64le.

This should probably be documented somewhere, either a comment in the Dockerfile or a README.md in `test/instrumentation-e2e-apps`.

tests/instrumentation-e2e-apps/apache-httpd/Dockerfile
```diff
@@ -0,0 +1,5 @@
+
+FROM httpd:2.4-alpine
+
+RUN sed -i "s#Listen 80#Listen 8080#g" /usr/local/apache2/conf/httpd.conf
+RUN chmod 777 -R /usr/local/apache2/
```
pavolloffay, line 5: I have an idea, is there a way to apply these changes at runtime (e.g. change the cmd command / use init container) ? 

The idea is to use just the upstream image.

iblancasa: The problem is that, since we want to use a user different from than root, the user will not have permission to modify the folder permissions or the files.

pavolloffay:
PR needs to be rebased

PR #2086, opened 2023-09-05
Run the upgrade of a collector instance after the status changes from unmanaged to managed

### LINKED ISSUE #1890: Run the upgrade as part of the reconciliation for instances that will be switched from unmanaged to managed after the upgrade
Run the upgrade as part of the reconciliation for instances that will be switched from unmanaged to managed after the upgrade.

_Originally posted by @pavolloffay in https://github.com/open-telemetry/opentelemetry-operator/pull/1888#discussion_r1251607491_

iblancasa:
Resolves #1890


pkg/collector/reconcile/opentelemetry.go
```diff
@@ -38,6 +41,17 @@ import (
 func Self(ctx context.Context, params manifests.Params) error {
 	changed := params.Instance
 
+	up := &collectorupgrade.VersionUpgrade{
+		Log:      params.Log,
+		Version:  version.Get(),
+		Client:   params.Client,
+		Recorder: record.NewFakeRecorder(collectorupgrade.RecordBufferSize),
```
jaronoff97, line 48: I believe we should have a real recorder through params at this point, we should use that here

iblancasa: So: should I add the recorder to the Params struct to propagate it here? It makes sense. I'll do it.

Thanks for the suggestion!

pkg/collector/reconcile/opentelemetry.go
```diff
@@ -28,6 +28,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	collectorupgrade "github.com/open-telemetry/opentelemetry-operator/pkg/collector/upgrade"
```
jaronoff97, line 31: nit: this import should be grouped with the below

pkg/collector/reconcile/opentelemetry.go
```diff
@@ -23,13 +23,12 @@ import (
 	"github.com/open-telemetry/opentelemetry-operator/internal/manifests"
 	"github.com/open-telemetry/opentelemetry-operator/internal/manifests/collector"
 	"github.com/open-telemetry/opentelemetry-operator/internal/naming"
+	collectorupgrade "github.com/open-telemetry/opentelemetry-operator/pkg/collector/upgrade"
```
jaronoff97, line 26: gah the imports are also wrong here, these should be in group 3

jaronoff97 (approved):

pkg/collector/reconcile/opentelemetry.go
```diff
@@ -38,6 +39,17 @@ import (
...
+	up := &collectorupgrade.VersionUpgrade{
+		Log:      params.Log,
+		Version:  version.Get(),
+		Client:   params.Client,
+		Recorder: params.Recorder,
+	}
+	changed, err := up.ManagedInstance(ctx, changed)
+	if err != nil {
+		return fmt.Errorf("failed to upgrade the OpenTelemetry CR: %w", err)
+	}
```
jaronoff97, line 51: on second look [here](https://github.com/iblancasa/opentelemetry-operator/blob/17e681d1df853be639c6351ad528677d8c33fba0/pkg/collector/upgrade/upgrade.go#L104-L108), it looks like this _may_ break users using custom images that don't adhere to semver. on failure to parse the instance, we would fail to call `updateScaleSubResourceStatus` which would mean any HPA would fail as well.

iblancasa: From what I see, that problem will happen also with a regular upgrade, right?
If I'm rigth, I think it makes sense to fix that as part of another PR.

jaronoff97: yes and no, the current way we use this is as a [manager runnable task](https://github.com/open-telemetry/opentelemetry-operator/blob/98b2f3748afc43c79ce6336bb94925db6560c250/main.go#L303-L312) which wouldn't block setting the status.

pkg/collector/reconcile/opentelemetry.go
```diff
@@ -38,6 +39,18 @@ import (
...
+	up := &collectorupgrade.VersionUpgrade{
+		Log:      params.Log,
+		Version:  version.Get(),
+		Client:   params.Client,
+		Recorder: params.Recorder,
+	}
+	changed, err := up.ManagedInstance(ctx, changed)
+	if err != nil {
+		// don't fail to allow setting the status
+		params.Log.Error(err, "failed to upgrade the OpenTelemetry CR")
```
jaronoff97, line 51: we could even add this to the recorder event if we wanted to, but this is fine for now. Thank you!

jaronoff97 (approved):

jaronoff97:
Great stuff, thank you 🙇

PR #2089, opened 2023-09-07
feat: add Spec.Java.Command support

lanxenet:
custom the InitContainers copy command,eg with config file


swiatekm-sumo:
What use case does this enable?

lanxenet:
use case:

lanxenet:
use case eg: 
```yaml
kind: Instrumentation
metadata:
  name: java-instrumentation
spec:
  exporter:
    endpoint: http://otlp.monitoring:4317
  propagators:
  - tracecontext
  - baggage
  - b3
  sampler:
    argument: '1'
  java:
    image: registry.cn-shenzhen.aliyuncs.com/inspireso/autoinstrumentation-java:easeagent-v2.2.6.3
    # copy javaagent.jar and javaagent.properties or other.
    command: ["/bin/sh", "-c", "ls -l /javaagent/ && cp /javaagent/* /otel-auto-instrumentation/"]
    env:
    - name: OTEL_APP_NAMESAPCE
      valueFrom:
        fieldRef:
          fieldPath: metadata.namespace
    - name: OTEL_APP_NAME
      valueFrom:
        fieldRef:
          fieldPath: metadata.labels['app']
```

>

lanxenet:
> What use case does this enable?

```yaml
kind: Instrumentation
metadata:
  name: java-instrumentation
spec:
  exporter:
    endpoint: http://otlp.monitoring:4317
  propagators:
  - tracecontext
  - baggage
  - b3
  sampler:
    argument: '1'
  java:
    image: registry.cn-shenzhen.aliyuncs.com/inspireso/autoinstrumentation-java:easeagent-v2.2.6.3
    # copy javaagent.jar and javaagent.properties or other.
    command: ["/bin/sh", "-c", "ls -l /javaagent/ && cp /javaagent/* /otel-auto-instrumentation/"]
    env:
    - name: OTEL_APP_NAMESAPCE
      valueFrom:
        fieldRef:
          fieldPath: metadata.namespace
    - name: OTEL_APP_NAME
      valueFrom:
        fieldRef:
          fieldPath: metadata.labels['app']
```

swiatekm-sumo:
If I understand correctly, you want to be able to customize this because you'd like to use a custom instrumentation image and want to potentially copy more files than we do by default. That seems reasonable to me. Only concern I have at the API level is whether we should make clear that it's the copy command we're overriding here - maybe the field should be called `copyCommand`? WDYT @jaronoff97 ?

jaronoff97:
I think we should open up an issue to discuss this more. I'm not sure how i feel about letting people customize this step because it feels like it could be rife with potential failures and the actual loop for development isn't very easy. if we don't propagate enough params or just want a custom path to copy, id propose we only expose what we need to here. cc @TylerHelmuth who has more experience with instrumentation things.

lanxenet:
>

lanxenet:
> I think we should open up an issue to discuss this more. I'm not sure how i feel about letting people customize this step because it feels like it could be rife with potential failures and the actual loop for development isn't very easy. if we don't propagate enough params or just want a custom path to copy, id propose we only expose what we need to here. cc @TylerHelmuth who has more experience with instrumentation things.

For users, we provide a default initContainer command, which is the original ("cp", "/javaagent.jar", "/otel-auto-instrumentation/javaagent.jar"), without adding any additional complexity for users. However, for those who have specific customization requirements, they can customize their own commands for debugging scenarios or using other Java agents, and so on.

lanxenet:
> If I understand correctly, you want to be able to customize this because you'd like to use a custom instrumentation image and want to potentially copy more files than we do by default. That seems reasonable to me. Only concern I have at the API level is whether we should make clear that it's the copy command we're overriding here - maybe the field should be called `copyCommand`? WDYT @jaronoff97 ?

Here, we use the same configuration format as Kubernetes' initContainer, making it easy for users to understand and use.

TylerHelmuth:
Since we can enable this feature with a default value such that it is a non-breaking change I think it is ok. We should include in the godoc comment that this is an advanced feature that should only be used when providing a custom image.

On a similar note to @swiatekm-sumo, we could make it clear that this field allows overriding the command of the initContainer to do anything you want.  It _should_ perform a copy to `"/otel-auto-instrumentation/javaagent.jar"`, but it technically doesn't have to.  Maybe a name like `initContainerCommand`.

Also, I'd like to hold off merging this for a reasonable amount of time to give https://github.com/open-telemetry/opentelemetry-operator/pull/1901 a chance to merge first since https://github.com/open-telemetry/opentelemetry-operator/pull/1901 affects the new default value.

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
```
pavolloffay, line 3: no change here

pavolloffay (changes requested):

PR #2095, opened 2023-09-07
[chore] Build and publish application hosted in the repository for the Python instrumentation E2E test

iblancasa:
Related to https://github.com/open-telemetry/opentelemetry-operator/issues/2073.

The test migration will be done in a follow up PR after publishing the image.


tests/instrumentation-e2e-apps/python/app.py
```diff
@@ -0,0 +1,10 @@
+from flask import Flask
+
+app = Flask(__name__)
+
+@app.route('/')
+def index():
+    return "Hi"
+
+if __name__ == "__main__":
+    app.run(host='0.0.0.0')
```
swiatekm-sumo, line 10: How about using the standard library http server instead:
```python
import http.server
import socketserver
from http import HTTPStatus

PORT = 8000

class Handler(http.server.SimpleHTTPRequestHandler):
    def do_GET(self):
        self.send_response(HTTPStatus.OK)
        self.end_headers()
        self.wfile.write(b'Hi')

with socketserver.TCPServer(("", PORT), Handler) as httpd:
    print("serving at port", PORT)
    httpd.serve_forever()
```

A bit more code, but we lose the Flask dependency and don't need to use `pip` or have a requirements file at all.

It's also possible to start a similar server without any code at all, but then it just serves files from a directory, which may not serve our purposes as well.

iblancasa: > How about using the standard library http server instead:
> ...
> 
> A bit more code, but we lose the Flask dependency and don't need to use `pip` or have a requirements file at all.

OK, I agree it is a better idea! Thanks for the suggestion.

swiatekm-sumo: I guess Flask is the simpest we can do, then.

swiatekm-sumo (approved):

pavolloffay:
https://github.com/open-telemetry/opentelemetry-operator/pull/2097 was merged please rebase and resue the GHA config to publish the image

jaronoff97 (approved):
LGTM, im rerunning the tests before i merge.

jaronoff97 (approved):

PR #2103, opened 2023-09-11
[dotnet-auto] Add support for musl based images

### LINKED ISSUE #1849: [.NET auto-instrumentation] Add support for alpine-based images
Reference: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2658

For now, opentelemetry-operator is able to inject auto-instrumentation only for glibc based linux images.
Alpine is using musl library so it cannot be supported.

.NET auto-instrumentation is supporting Alpine. There is a need to inject different set of libraries.

Kielek:
Fixes #1849


pkg/instrumentation/annotation.go
```diff
@@ -27,6 +27,7 @@ const (
 	annotationInjectNodeJS        = "instrumentation.opentelemetry.io/inject-nodejs"
 	annotationInjectPython        = "instrumentation.opentelemetry.io/inject-python"
 	annotationInjectDotNet        = "instrumentation.opentelemetry.io/inject-dotnet"
+	annotationDotNetRuntime       = "instrumentation.opentelemetry.io/otel-dotnet-auto-runtime"
```
Kielek, line 30: Annotation name is different than discussed in #1849
I have tried to make it compatible with `nnotationGoExecPath`

README.md
```diff
@@ -254,9 +254,13 @@ Python:
 instrumentation.opentelemetry.io/inject-python: "true"
 ```
 
-DotNet:
+.NET:
+.NET auto-instrumentation also honors an annotation that will be used to set the .NET runtime versions. There are two versions supported `linux-glibc` and `linux-musl`.
```
pellared, line 258:
```suggestion
.NET auto-instrumentation also honors an annotation that will be used to set the .NET runtime version. There are two versions supported `linux-glibc` and `linux-musl`.
```

pkg/instrumentation/dotnet.go
```diff
@@ -62,6 +65,18 @@ func injectDotNetSDK(dotNetSpec v1alpha1.DotNet, pod corev1.Pod, index int) (cor
 		return pod, errors.New("OTEL_DOTNET_AUTO_HOME environment variable is already set in the .NET instrumentation spec")
 	}
 
+	runtime := pod.Annotations[annotationDotNetRuntime]
+	coreClrProfilerPath := ""
+
```
pellared, line 70:
```suggestion
```

pkg/instrumentation/dotnet.go
```diff
@@ -33,11 +33,14 @@ const (
...
 	dotNetCoreClrProfilerID             = "{918728DD-259F-4A6A-AC2B-B85E1B658318}"
-	dotNetCoreClrProfilerPath           = "/otel-auto-instrumentation/linux-x64/OpenTelemetry.AutoInstrumentation.Native.so"
+	dotNetCoreClrProfilerGlibcPath      = "/otel-auto-instrumentation/linux-x64/OpenTelemetry.AutoInstrumentation.Native.so"
+	dotNetCoreClrProfilerMuslPath       = "/otel-auto-instrumentation/linux-musl-x64/OpenTelemetry.AutoInstrumentation.Native.so"
 	dotNetAdditionalDepsPath            = "/otel-auto-instrumentation/AdditionalDeps"
 	dotNetOTelAutoHomePath              = "/otel-auto-instrumentation"
 	dotNetSharedStorePath               = "/otel-auto-instrumentation/store"
 	dotNetStartupHookPath               = "/otel-auto-instrumentation/net/OpenTelemetry.AutoInstrumentation.StartupHook.dll"
+	dotNetGlibcLinuxRuntime             = "linux-glibc"
+	dotNetMuslLinuxRuntime              = "linux-musl"
```
pellared, line 43: I propose to rename:
-  `dotNetGlibcLinuxRuntime` to `dotNetRuntimeLinuxGlibc`
-  `dotNetMuslLinuxRuntime` to `dotNetRuntimeLinuxMusl`

so that they have the same prefix which is the usual pattern for defining enumerations in Go.

Also the constants could be defined in a seperate `const ( )` block with a comment above the block saying that these are supported values for dotnet-runtime annoation. 

Example: https://cs.opensource.google/go/go/+/refs/tags/go1.21.1:src/net/http/status.go;l=7-9

pellared (approved):

pkg/instrumentation/dotnet.go
```diff
@@ -62,6 +69,17 @@ func injectDotNetSDK(dotNetSpec v1alpha1.DotNet, pod corev1.Pod, index int) (cor
...
 
+	runtime := pod.Annotations[annotationDotNetRuntime]
+	coreClrProfilerPath := ""
+	switch runtime {
+	case "", dotNetRuntimeLinuxGlibc:
+		coreClrProfilerPath = dotNetCoreClrProfilerGlibcPath
+	case dotNetRuntimeLinuxMusl:
+		coreClrProfilerPath = dotNetCoreClrProfilerMuslPath
+	default:
+		return pod, errors.New("provided instrumentation.opentelemetry.io/dotnet-runtime annotation value is not supported")
```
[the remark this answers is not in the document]
Kielek: e0a8ccf167f8429d975d185ff02a13728643d72a

tests/e2e-instrumentation/instrumentation-dotnet-musl/00-install-instrumentation.yaml
```diff
@@ -0,0 +1,26 @@
...
+    - name: SPLUNK_TRACE_RESPONSE_HEADER_ENABLED
+      value: "true"
+  exporter:
+    endpoint: http://localhost:4317
+  propagators:
+    - jaeger
+    - b3multi
+  sampler:
+    type: parentbased_traceidratio
+    argument: "0.25"
```
swiatekm-sumo, line 26: I think we can just put this in the `00-install-collector.yaml` file and rename it to just `00-install.yaml.`. I don't see much benefit in making them separate.

swiatekm-sumo (approved):
LGTM, minor nitpick about test organization.

README.md
```diff
@@ -254,9 +254,13 @@ Python:
 instrumentation.opentelemetry.io/inject-python: "true"
 ```
 
-DotNet:
+.NET:
+.NET auto-instrumentation also honors an annotation that will be used to set the .NET runtime version. There are two versions supported `linux-glibc` and `linux-musl`.
```
pjanotti, line 258: I know that we think about the standard lib that we are linking against, but, for a .NET user I think it makes more sense to use the [Runtime Identifiers](https://learn.microsoft.com/en-us/dotnet/core/rid-catalog). Two reasons: first it is future proof, second it is the term and values used in .NET itself.
```suggestion
.NET auto-instrumentation also honors an annotation that will be used to set the .NET [Runtime Identifiers](https://learn.microsoft.com/en-us/dotnet/core/rid-catalog)(RIDs). Currently, only two RIDs are supported: `linux-x64` and `linux-musl-x64`.
```

README.md
```diff
@@ -254,9 +254,13 @@ Python:
...
+By default `linux-glibc` is used.
```
pjanotti, line 259:
```suggestion
By default `linux-x64` is used.
```

pjanotti: I will stop making this rename suggestion here, but, paths and other places should also follow that.

.chloggen/1849-dotnet-auto-musl-support.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: autoinstrumentation
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: .NET Automatic Instrumentation support for alpine-based images
```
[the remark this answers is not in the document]
Kielek: d19aafd7aee570336af8e727878940ada43d9656

README.md
Kielek: 176c908030b328b8025ce9afb1b1027bb39c71e0

README.md
Kielek: 176c908030b328b8025ce9afb1b1027bb39c71e0

tests/e2e-instrumentation/instrumentation-dotnet-musl/00-install-instrumentation.yaml
Kielek: Files for this tests are copied from `instrumentation-dotnet`. What is more other e2e follows same pattern.

BTW I like it as a separate step - it ilustrate what you should to execute otel-operator.
I would consider to create separate issue to fix all this files.

pavolloffay (approved):

tests/e2e-instrumentation/instrumentation-dotnet-musl/01-install-app.yaml
```diff
@@ -0,0 +1,29 @@
...
+        instrumentation.opentelemetry.io/inject-dotnet: "true"
+        instrumentation.opentelemetry.io/otel-dotnet-auto-runtime: "linux-musl-x64"
+    spec:
+      securityContext:
+        runAsUser: 1000
+        runAsGroup: 3000
+        fsGroup: 2000
+      containers:
+      - name: myapp
+        image: docker.io/avadhutp123/aspnetapp@sha256:d2e8d3415f6f12efae0369aa0a9777a58583841fb133f33e10a73adb9fb392da  # source code of the application: https://github.com/dotnet/dotnet-docker/tree/main/samples/aspnetapp
```
pavolloffay, line 25: Is there an official image available? 

We are discussing about using official apps that are well maintained or hosting the test apps in this repo.

cc) @iblancasa

pavolloffay: Also if we decide to use an image it should be published for multiple arch: arm, amd64, IBM Z/P

pavolloffay (commented):
See my comment about the test app

tests/e2e-instrumentation/instrumentation-dotnet-musl/01-install-app.yaml
pavolloffay:
```
avadhutp123/aspnetapp          <none>              8606880cf2e0   11 months ago   220MB
```

Kielek: In e2e tests in fact you are not testing applications. It can be even empty image. E2E tests are checking only if appropriate env. variables are injected.

For now, .NET auto supports linux-x64 (glibc/amd64) and linux-musl-x64 (musl/amd64). ARM (it can be supported), IBM Z/P (I am not sure if it even can be supported) are not supported by auto-instrumentation

iblancasa: > In e2e tests in fact you are not testing applications. It can be even empty image. E2E tests are checking only if appropriate env. variables are injected.

This is totally true. I think we may need to revisit the current approach to use an echo server or something similar.

> For now, .NET auto supports linux-x64 (glibc/amd64) and linux-musl-x64 (musl/amd64). ARM (it can be supported), IBM Z/P (I am not sure if it even can be supported) are not supported by auto-instrumentation

Yes, they cannot be supported.

> EDIT: I am not introducing here anything new, it was used also by glibc test.

Yes, this change is fine.

iblancasa (approved):

tests/e2e-instrumentation/instrumentation-dotnet-musl/01-install-app.yaml
Kielek: > > In e2e tests in fact you are not testing applications. It can be even empty image. E2E tests are checking only if appropriate env. variables are injected.
> 
> This is totally true. I think we may need to revisit the current approach to use an echo server or something similar.
> 
To have full e2e coverage, tt will be great to check if some auto-instrumented (for .NET - using bytecode) traces/metrics/logs are generated by the application.
.NET auto-instrumentation should silently continue in case of non-working correctly. The same approach is recommended for other OTel-auto versions.

For sure another PR.

iblancasa: Ok so, I have some PRs to minimize the applications, add support to multiple architectures and hos them in this repository. Adding a full E2E test can be a follow up of that.

TylerHelmuth (approved):

pjanotti (approved):

Issue #2001, opened 2023-08-09
Publishing the bundle image failed for 0.82.0


swiatekm-sumo:
See: https://github.com/open-telemetry/opentelemetry-operator/actions/runs/5807500252/job/15742453581

```
ERROR: failed to solve: failed to push otel/opentelemetry-operator-bundle:0.82.0: push access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
```

It looks like we're using the same credentials as for other images, so I'd assume this is a problem with permissions in Dockerhub. 

Pinging @iblancasa as the author of #1880.

iblancasa:
Who has permissions in the DockerHub organization?

swiatekm-sumo:
I see @pavolloffay already opened a maintenance request in the community repo: https://github.com/open-telemetry/community/issues/1635

pavolloffay:
Also the publish for 0.83.0 failed

pavolloffay:
Publishing to dockerhub seems to be fixed, but the same issue is with the GHCR

pavolloffay:
The publishing is still failing

swiatekm-sumo:
I tried this on a fork and it worked (albeit after getting rid of publishing to Docker): https://github.com/swiatekm-sumo/opentelemetry-operator/actions/runs/6260451251/job/16998511279. So the failure most likely has something to do with our repo settings themselves.

pavolloffay:
The publishing of the bundle image is still failing. I have created https://github.com/docker/build-push-action/issues/981

What is interesting is that the e2e test images were published https://github.com/orgs/open-telemetry/packages?repo_name=opentelemetry-operator
However sometimes the first image publish fails on a similar error - https://github.com/open-telemetry/opentelemetry-operator/actions/runs/6394622279/job/17356427547
```
ERROR: failed to solve: failed to push ghcr.io/open-telemetry/opentelemetry-operator/e2e-test-app-apache-httpd:main: failed commit on ref "manifest-sha256:c33c301c43c5636fc91147e12740f65f2b2339e606f8710b3bbc6210e87fcc0e": unexpected status from PUT request to https://ghcr.io/v2/open-telemetry/opentelemetry-operator/e2e-test-app-apache-httpd/manifests/sha256:c33c301c43c5636fc91147e12740f65f2b2339e606f8710b3bbc6210e87fcc0e: 403 Forbidden
[...]

PR #2138, opened 2023-09-20
Fix publishing of bundle image

pavolloffay:
Updates https://github.com/open-telemetry/opentelemetry-operator/issues/2001

### DIFF
--- a/.github/workflows/publish-operator-bundle.yaml
+++ b/.github/workflows/publish-operator-bundle.yaml
@@ -12,6 +12,10 @@ on:
       - 'v*'
   workflow_dispatch:
 
+permissions:
+  contents: read
+  packages: write
+
 jobs:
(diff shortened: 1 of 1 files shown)


swiatekm-sumo:
So the idea is basically that this isn't working because it doesn't have permission to *create* the package, and this change will grant that? I suppose we'll be able to later drop this permission once the package is created?

pavolloffay:
> So the idea is basically that this isn't working because it doesn't have permission to create the package, and this change will grant that? I suppose we'll be able to later drop this permission once the package is created?

We can try dropping it later

pavolloffay:
it didn't help https://github.com/open-telemetry/opentelemetry-operator/actions/runs/6251580120/job/16972995833#step:9:198

Issue #2136, opened 2023-09-20
Add support for PodDisruptionBudgets


JorTurFer:
Although `sidecar` and `daemonset` don't allow pdb, `deployment` and `statefulset` modes are workload types where pdbs make sense totally, as they are centralized workloads. 
Currently, we are using the collector helm chart because it supports pdbs but we'd like to use the operator for using `Target Allocator`, but the pdbs are a requirement for us.

I think that adding support to pdbs is a good addition to the operator and I'm willing to contribute with the feature if you also agree with having it

pavolloffay:
thanks for the request @JorTurFer 

Do you want to use `PodDisruptionBudgets` for target allocator as well?

JorTurFer:
I haven't tested properly the target allocator yet, but if it's something that requires HA, yes. I mean, I know that we can use affinities to prevent some scenarios, but we have to ensure that we have at least X instances of the collector working. That's why we needed the pdbs

JorTurFer:
Do you see the request useful? I'm willing to implement (but obviously, I prefer to ask before rather than drafting a PR that never is merged xD)

pavolloffay:
I am not well familiar with the concept perhaps @jaronoff97 ?

I found this 

> From version 1.15 PDBs support custom controllers where the [scale subresource](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource) is enabled.

https://kubernetes.io/docs/tasks/run-application/configure-pdb/#identify-an-application-to-protect

otelcol CRD implements the scale subresource

JorTurFer:
No no, I want to deploy a pdb for the generated deployment / statefulset. I mean, with this manifest:
```yaml
apiVersion: opentelemetry.io/v1alpha1
kind: OpenTelemetryCollector
metadata:
  name: sidecar-for-my-app
spec:
  mode: deployment
```
The operator generates a deployment for the collector. I want to generate also a PDB for that generated deployment, not for the CRD itself

jaronoff97:
Hey @JorTurFer thanks for getting this going here :) Is there a reason we would do this in the operator and not just let users apply this themselves (that's what i do currently)? What pavol mentions is accurate in that if we were to do this in the operator we would want to do it the same way as how the hpa created by the operator works i.e. by targeting the scale subresource.

jaronoff97:
[...]

PR #2141, opened 2023-09-21
feat: Add support for PDBs on deployment and statefulset

JorTurFer:
Fixes: https://github.com/open-telemetry/opentelemetry-operator/issues/2136


JorTurFer:
Sorry, I missed the bundle at the first try. I have updated the PR

JorTurFer:
awesome, a missing change in the stage 🤦

controllers/opentelemetrycollector_controller.go
```diff
@@ -305,11 +308,18 @@ func (r *OpenTelemetryCollectorReconciler) SetupWithManager(mgr ctrl.Manager) er
 	}
 
 	autoscalingVersion := r.config.AutoscalingVersion()
-	if autoscalingVersion == autodetect.AutoscalingVersionV2 {
+	if autoscalingVersion == autodetect.VersionV2 {
```
jaronoff97, line 311: Why is this changing?

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -123,6 +124,11 @@ type OpenTelemetryCollectorSpec struct {
 	//
 	// +optional
 	Autoscaler *AutoscalerSpec `json:"autoscaler,omitempty"`
+	// PodDisruptionBudget specifies the pod disruption budget configuration to use
+	// for the OpenTelemetryCollector workload.
+	//
+	// +optional
+	PodDisruptionBudget *PodDisruptionBudgetSpec `json:"podDisruptionBudget,omitempty"`
```
jaronoff97, line 131: note for other reviewers, this is a paired down version of the main PDB spec because we don't want to allow the configuration of the selector

internal/manifests/collector/poddisruptionbudget.go
```diff
@@ -0,0 +1,78 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
+	"github.com/open-telemetry/opentelemetry-operator/internal/config"
+	"github.com/open-telemetry/opentelemetry-operator/internal/naming"
+	"github.com/open-telemetry/opentelemetry-operator/pkg/autodetect"
+)
+
+func PodDisruptionBudget(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelemetryCollector) client.Object {
```
jaronoff97, line 30: can we add a comment here linking to the deprecation docs so we know when we can remove the dual state? https://kubernetes.io/docs/reference/using-api/deprecation-guide/#poddisruptionbudget-v125

jaronoff97 (changes requested):
Overall this looks good! Thank you for updating the logic for version detection as well to be more generic. Could you also update one of the e2e to set the PDB and ensure it gets reconciled correctly and the selector works as expected?

JorTurFer:
Thanks for the quick review, tomorrow I'll address al the comments 😄

JorTurFer:
I have addressed all the comments. I'm not 100% if I have done the e2e test correctly because it's the 1st time I have used kuttl but at least it works in local 🤣

internal/manifests/collector/poddisruptionbudget.go
pavolloffay: I am fine bumping the lowest support k8s version to 1.21. 

It seems that the oldest supported k8s version by a cloud provider out there is 1.23. I will start a GH issue to discuss it.

JorTurFer: If the lowest version is 1.21, we don't need to autodetect versions for pdb, and if the lowest version would be 1.23, the same applies for autoscaling (autoscaling/v2 was introduced on 1.23).
I don't have problems to update my PR if you have consensus about the lowest version change

JorTurFer:
It looks like I have to review the e2e test to work on multiple cluster versions. Let me do it before triggering again the CI because the check will fail again. I'm on it

JorTurFer:
@pavolloffay ,
What should I do? I mean, should I wait until https://github.com/open-telemetry/opentelemetry-operator/issues/2145 is decided? Will it require too much time? I'm not in rush with the pdb, during next month is enough for use, but if the discussion goes longes I'd prefer to merge it with support to both

pavolloffay:
> What should I do?

Let's see how quickly goes the conversation in https://github.com/open-telemetry/opentelemetry-operator/issues/2145 I am happy to open a PR next week to bump the version.

JorTurFer:
I have already rebased main branch and updated this PR. PTAL when you have some time @pavolloffay

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -444,6 +450,23 @@ type AutoscalerSpec struct {
...
 }
 
+// PodDisruptionBudgetSpec defines the OpenTelemetryCollector's pod disruption budget specification.
+type PodDisruptionBudgetSpec struct {
+	// An eviction is allowed if at least "minAvailable" pods selected by
+	// "selector" will still be available after the eviction, i.e. even in the
+	// absence of the evicted pod.  So for example you can prevent all voluntary
+	// evictions by specifying "100%".
+	// +optional
+	MinAvailable *intstr.IntOrString `json:"minAvailable,omitempty"`
```
jaronoff97, line 460: does one of these need to be set in order to be accepted?

jaronoff97 (commented):
I think we may want to add something in the validating webhook to be sure that we don't attempt a bad config

pavolloffay:
`End-to-end tests / End-to-end tests (1.23, e2e-pdb) (pull_request) ` - failed

apis/v1alpha1/opentelemetrycollector_types.go
JorTurFer: What do you mean?

JorTurFer:
I'm spinning up a cluster with that version to check it

JorTurFer:
@pavolloffay , I've tested it locally with k8s 1.23 and (at least locally) the feature works:
image

The problem I have seen is that the deletion is producing timeouts

JorTurFer:
I've noticed that if I use my own namespace for the tests, the k8s GC removes the namespaces more quickly and I can also validate the generated labels, so I've updated my test to use it's own namespace, and now 30 seconds of timeout is more than enough (the test needs a few seconds).

I say this because I see other e2e tests with > 150 seconds of timeouts and maybe the root cause it's something like this

pavolloffay:
The PR looks good to me. But I am also interested in @jaronoff97 question whether we could do some validation of the PDB fields that we added to our CR.

JorTurFer:
Which kind of validation? I mean, I added a check in webhooks to validate if both have been set, and the other validations are already done by the openapi spec (but I can be missing any important thing at some point)

apis/v1alpha1/opentelemetrycollector_types.go
jaronoff97: does someone who wants to use a PDB need to set either min or max available for the kube api to accept the object?

JorTurFer: No no, the must set only one of them, they are mutually exclusive. If they set both, the admission webhook will reject the CR: https://github.com/open-telemetry/opentelemetry-operator/pull/2141/files#diff-f7d2d1348f61f254e8b086ada3e9ad6ccc705b9cbbd171065911fd75b38bf495R251-R255.

jaronoff97: ah i missed it in my initial review, thank you! This LGTM.

jaronoff97 (approved):

apis/v1alpha1/opentelemetrycollector_types.go
JorTurFer: no worries, it wasn't in your initial review, I added it due to your feedback 😄

pavolloffay (approved):

jaronoff97:
@JorTurFer thank you so much for your contribution 🎊🎉

Issue #1417, opened 2023-02-01
Pods stuck in terminating state because of sdk injection


cdaguerre:
The `instrumentation.opentelemetry.io/inject-sdk` annotation keeps job pods from being terminated.
As soon as I remove it, the pods disappear. 
The pods have the following finalizer:
```yaml
  finalizers:
  - batch.kubernetes.io/job-tracking
```
I couldn't find any related issue here and nothing helpful from the controller manager logs...

edude03:
I have the same issue - I did see a related issue but looking for it at the moment. Basically it's because the operator is trying to modify the pod after it's been deployed already. Seems to have started since google upgrade us from 1.23.x -> 1.24.x

man0s:
I've stumbled into the same issue.

WesleyKlop:
What I see happening as well is that when I just try to remove the finalizer, the controller complains that I'm trying to edit immutable fields. These fields being the OTEL_* environment variables, which I think break because a pod that is being removed won't have a name/namespace but those are injected into the environment variable.

fyuan1316:
I believe [#2584](https://github.com/open-telemetry/opentelemetry-operator/pull/2584) has addressed this issue.

You can follow the steps below to verify that the issue will not happen again.
Create a task with a run time of 1 minute and delete the pod while the task is running. the pod will not get stuck in a terminating state.

```yaml
# Apply Instrumentation
...

# annotate ns
kubectl annotate namespace otel-ext-1 instrumentation.opentelemetry.io/inject-java="true"

# Create Job
kubectl apply -f - <<EOF
apiVersion: batch/v1
kind: Job
metadata:
  name: pause-and-sleep-job
  namespace: otel-ext-1
spec:
  template:
    spec:
      containers:
      - name: pause-and-sleep
        image: busybox
        command: ["/bin/sh", "-c", "echo 'Pausing for 20 seconds...'; sleep 60; echo 'Task completed.'"]
      restartPolicy: Never

EOF
```

pavolloffay:
Thanks for the update, please re-open if the issue is still present.

PR #2163, opened 2023-09-26
Fix bug where Pods get stuck Terminating because of SDK Injection modifications

WesleyKlop:
Hello!

I think this solves #1417 by not adding/changing instrumentation on Pods that have a Deletion timestamp. I don't know for certain if this is the most clean solution but I think it should work!

I'm missing tests but I don't know how to best test this.

### DIFF
--- a/pkg/instrumentation/podmutator.go
+++ b/pkg/instrumentation/podmutator.go
@@ -214,6 +214,13 @@ func (pm *instPodMutator) Mutate(ctx context.Context, ns corev1.Namespace, pod c
 		return pod, nil
 	}
 
+	// If the Pod is marked for deletion we skip adding instrumentation to it because we change immutable fields (env).
+	// see https://github.com/open-telemetry/opentelemetry-operator/issues/1417.
+	if pod.DeletionTimestamp != nil {
+		logger.Info("Skipping pod instrumentation - marked for deletion")
+		return pod, nil
+	}
+
 	var inst *v1alpha1.Instrumentation
 	var err error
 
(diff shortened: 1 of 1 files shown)


pkg/instrumentation/podmutator.go
```diff
@@ -75,6 +75,13 @@ func (pm *instPodMutator) Mutate(ctx context.Context, ns corev1.Namespace, pod c
 		return pod, nil
 	}
 
+	// If the Pod is marked for deletion we skip adding instrumentation to it because we change immutable fields (env).
+	// see https://github.com/open-telemetry/opentelemetry-operator/issues/1417.
+	if pod.DeletionTimestamp != nil {
+		logger.Info("Skipping pod instrumentation - marked for deletion")
+		return pod, nil
+	}
```
TylerHelmuth, line 83: Whether or not this fixes the specific issue, I believe this is proper behavior for the operator.  Let's move this above the `isAutoInstrumentationInjected` check.

WesleyKlop: Sure!

swiatekm-sumo (commented):
Can you add a unit test for this behaviour? 

It would be nice to have an E2E test as well, but the condition looks difficult to reproduce consistently in an actual cluster.

jaronoff97:
@WesleyKlop checking in here, would you be able to add a unit test?

pavolloffay:
Done in https://github.com/open-telemetry/opentelemetry-operator/pull/2584

PR #2166, opened 2023-09-27
Run an initContainer on an enabled flag

### LINKED ISSUE #1771: Run the collector's `validate` command prior to rollout
Now that the collector has a validate command as part of its binary, we should run it prior to rollout to ensure that the supplied configuration is valid and won't cause as a crash.

jaronoff97:
closes #1771 

This PR adds a new field to the collector CRD which will automatically add an init container that runs the collector's validate command. This will ensure that no pods roll out with potentially invalid config.

### DIFF
--- a/internal/manifests/collector/init_containers.go
+++ b/internal/manifests/collector/init_containers.go
@@ -1,0 +1,41 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package collector
+
+import (
+	"fmt"
+
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
+	"github.com/open-telemetry/opentelemetry-operator/internal/config"
+)
+
+func InitContainers(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelemetryCollector) []corev1.Container {
+	if !otelcol.Spec.RunValidation {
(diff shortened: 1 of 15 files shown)


internal/manifests/collector/init_containers.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
+	"github.com/open-telemetry/opentelemetry-operator/internal/config"
+)
+
+func InitContainers(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelemetryCollector) []corev1.Container {
+	if !otelcol.Spec.RunValidation {
+		return otelcol.Spec.InitContainers
+	}
+	c := Container(cfg, logger, otelcol, true)
```
iblancasa, line 31: Do we maybe need to set the securitycontext?

jaronoff97: We are copying the security context set on the main container, do we need to set anything more than that?

iblancasa: Ok that should be fine

jaronoff97:
hmm, after some testing i realize that this will only be helpful for a deployment where it's one up one down, for daemonset and statefulset this won't do much more than the current functionality as you have to delete the existing pod first. Should I instead attempt my initial approach where I run a job before applying the deployment? cc @iblancasa @pavolloffay

swiatekm-sumo:
Can you explain in more detail what this approach accomplishes vs just letting the main container crash? We get a failure in the init container instead of the main container, but in terms of a rollout the effect is the same, no?

pavolloffay:
> Should I instead attempt my initial approach where I run a job before applying the deployment?

We could explore this path, but I am not sure if it is feasible to block the validating webhook for so long.

iblancasa:
> Should I instead attempt my initial approach where I run a job before applying the deployment? cc @iblancasa @pavolloffay

Interesting. Do we know about other operators doing something similar?

iblancasa:
> Can you explain in more detail what this approach accomplishes vs just letting the main container crash? We get a failure in the init container instead of the main container, but in terms of a rollout the effect is the same, no?

I think the idea is to get the error at some point and make it part of the events of the pod.

frzifus:
@jaronoff97 Wasnt the initial idea we discussed last time starting such a pod from the webhook? That way, a invalid config would never be applied?

jaronoff97:
Yeah that's what I'm going to try out now, reworking the PR currently :) initially I'm going to test that the idea works from the reconcile loop and then I can run it from the webhook.

Issue #1926, opened 2023-07-14
TargetAllocator hash computation returning false negative


rashmichandrashekar:
I am seeing hash computation issue for regex field where it returns a false negative, when there is only change in the regex field here - https://github.com/open-telemetry/opentelemetry-operator/blob/56f5de94dd3f1f83ac7dbdba3b8b4a6379317b48/cmd/otel-allocator/target/discovery.go#L85

For example, when there are 2 scrape configs with just the regex field changed between the 2, the hash returned is the same.

    scrape_configs:
      - job_name: kube-proxy
        scrape_interval: 30s
        label_limit: 63
        label_name_length_limit: 511
        label_value_length_limit: 1023
        kubernetes_sd_configs:
        - role: pod
        relabel_configs:
        - action: keep
          source_labels:
          - __meta_kubernetes_namespace
          - __meta_kubernetes_pod_name
          separator: "/"
          regex: kube-system/kube-proxy.+
        - source_labels:
          - __address__
          action: replace
          target_label: __address__
          regex: "regex1"
          replacement: "$$1:10249"

    scrape_configs:
      - job_name: kube-proxy
        scrape_interval: 30s
        label_limit: 63
        label_name_length_limit: 511
        label_value_length_limit: 1023
        kubernetes_sd_configs:
        - role: pod
        relabel_configs:
        - action: keep
          source_labels:
          - __meta_kubernetes_namespace
          - __meta_kubernetes_pod_name
          separator: "/"
          regex: kube-system/kube-proxy.+
        - source_labels:
          - __address__
          action: replace
          target_label: __address__
          regex: "regex2"
          replacement: "$$1:10249"

These 2 configs return the same hash

rashmichandrashekar:
Similar to issue - #1358

swiatekm-sumo:
Fixed by #2171

rashmichandrashekar:
@swiatekm-sumo - Sorry for the delay in responding to this issue as there seems to be a fix made with this - #2171 
[...]

Issue #1544, opened 2023-03-02
[target allocator] Scrape configuration hashing is resource intense


matej-g:
I'm trying out the target collector in environment with roughly 600 service monitors. What I'm constantly seeing that the target allocator container is consuming unexpectedly large amount of resources. I'm in process of collecting more data, but even checking via `kubectl top` shows lot of resource consumption:
```
<target_allocator_pod_name>          3137m        6924Mi
```

When consulting the profile, it seems a lot of this is coming from the hashing in the scrape configuration handler:
![Screenshot 2023-03-01 at 18 11 36](https://user-images.githubusercontent.com/38492574/222389170-c2a93a73-1ee8-489d-bf0f-4ca9fb91d47c.png)

I assume this could be because of the large number of elements in the scrape config and the amount of walking / reflection through the structure it takes to construct a hash.

I'm considering whether hashing the configuration before it is marshalled could be more performant here, since:
1) We could use a fast hashing algorithm (like https://github.com/zeebo/xxh3) directly on the configuration as slice of bytes
2) We'd avoid having to marshal the configuration upfront, if we already can compare hashes of the byte slices

I'll provide a draft PR with suggested changes as well, but wanted to collect thoughts.

matej-g:
cc @open-telemetry/operator-ta-maintainers

rashmichandrashekar:
The PR with the fix(https://github.com/open-telemetry/opentelemetry-operator/pull/1545), fixes this issue as well - #1926 
@matej-g - Are you planning to merge the PR sometime soon?

matej-g:
Hey @rashmichandrashekar,
I unfortunately did not have enough time to compare if my solution in the end is more performant than existing hashing. But if it could solve your issue as well, I can take a look again and try to refresh PR. I'll try to take a look this week,

swiatekm-sumo:
This is particularly strange given that this code should only run on changes to the scrape configs, which mean either reloading the config file or Prometheus CRs changing. And that really shouldn't happen very often.

If it's a problem in some configurations, I'd prefer the same solution Prometheus' discovery manager uses, which is rate limiting notifications to 1 per 5 seconds.

matej-g:
Hey @swiatekm-sumo, thanks for keeping an eye on this. Unfortunately I never got to move with this issue myself.

Regarding the code, I guess it's true it should only run if config change is detected. I can't recall now the exact conditions that were causing the spike in resource usage. Still:
[...]

PR #2171, opened 2023-09-28
Rebuild targets on scrape config regex-only changes

### LINKED ISSUE #1358: [target-allocator] Hashing in ScrapeConfigsHandler results in false negative
When adding unit tests, I found that the hashing algorithm in [ScrapeConfigsHandler()](https://github.com/open-telemetry/opentelemetry-operator/blob/ba24a19c01ee95b60dd0c62f519524a1046e95f7/cmd/otel-allocator/server/server.go#L90) for the target allocator results in the same hash if the only thing that has changed is a regex value.  The regex struct used in prometheus has no exported fields, so the hashing algorithm creates the same hash for both scrape configs.  This results in a false negative, and the old object returned in the http response.

Unit test: https://github.com/open-telemetry/opentelemetry-operator/blob/782b582dfa333d775b5b40681266adc7a2037b46/cmd/otel-allocator/server/server_test.go#L622

previous scrape configs:
```
map[string]*promconfig.ScrapeConfig{
	"serviceMonitor/testapp/testapp/1": {
		JobName:         "serviceMonitor/testapp/testapp/1",
		HonorTimestamps: false,
		ScrapeTimeout:   model.Duration(30 * time.Second),
		MetricsPath:     "/metrics",
		Scheme:          "http",
		HTTPClientConfig: config.HTTPClientConfig{
			FollowRedirects: true,
		},
		RelabelConfigs: []*relabel.Config{
			{
				SourceLabels: model.LabelNames{model.LabelName("job")},
				Separator:    ";",
				Regex:        relabel.MustNewRegexp("(.*)"),
				TargetLabel:  "__tmp_prometheus_job_name",
				Replacement:  "$$1",
				Action:       relabel.Replace,
			},
		},
	},
}
```

new scrape configs:
```
map[string]*promconfig.ScrapeConfig{
	"serviceMonitor/testapp/testapp/1": {
		JobName:         "serviceMonitor/testapp/testapp/1",
		HonorTimestamps: false,
		ScrapeTimeout:   model.Duration(30 * time.Second),
		MetricsPath:     "/metrics",
		Scheme:          "http",
		HTTPClientConfig: config.HTTPClientConfig{
			FollowRedirects: true,
		},
		RelabelConfigs: []*relabel.Config{
			{
				SourceLabels: model.LabelNames{model.LabelName("job")},
				Separator:    ";",
				Regex:        relabel.MustNewRegexp("something else"),
				TargetLabel:  "__tmp_prometheus_job_name",
				Replacement:  "$$1",
				Action:       relabel.Replace,
			},
		},
	},
}
```

swiatekm-sumo:
Fixes #1358 and #1926. The fix simply serializes the scrape configs to yaml before calculating the hash, instead of using `hashstructure`. This is slower by about 40%, but still a little over 1ms for 10000 configs.


swiatekm-sumo:
@matej-g noted high resource consumption from this in #1544, which this change could potentially exacerbate, but that seems strange in general considering this code only runs on changes to ServiceMonitors and the TA config file.

matej-g:
Thanks @swiatekm-sumo, I responded in https://github.com/open-telemetry/opentelemetry-operator/issues/1544#issuecomment-1743204613, so TL;DR if there weren't performance issues reported related to this code by someone else than me, I would not worry.

If this does side-step the unexported JSON fields issue, I think it's a good way to finally address this 👍

jaronoff97:
I'd be interested in understanding more about why we would re-calculate this so often, i would expect it only once every 15mins which is the timeout for the watcher i think. As a result, i'm less concerned with this being a problem and more concerned with the source of why we calculate so often.

jaronoff97 (approved):

jaronoff97:
yep, but I wouldn't expect servicemonitors or podmonitors to be changed as frequently as we see for some of these reports.

rashmichandrashekar:
@swiatekm-sumo - Sorry for the delay in responding to this PR as this seems to fix the issue I created -
But, the fix is only partial because, if the same fix isn't made to the otel collector's prometheusreceiver, even though the hash computation here updates the scrape config, the collector fails to get the updated config and will not work e2e. The same fix needs to be made [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/1d886eacd7806128d2528bda512e3915a1fa6a25/receiver/prometheusreceiver/metrics_receiver.go#L142C85-L142C85)

Issue #1707, opened 2023-05-03
[collector] Specify the collector's configuration with structure yaml instead of a string


TylerHelmuth:
Currently the only way to define the collector's config in the `OpentlemetryCollector` resource is via a string in the `Spec.Config` field.  Although it works, it can be frustrating to write, especially for complex collector configurations, since indentation is important.  For operator contributors and maintainers working with the string internally is also challenging.

I believe it would be beneficial to allow the config to be expressed as explicit yaml, mirroring the [collector's internal config structure](https://pkg.go.dev/go.opentelemetry.io/collector@v0.76.1/otelcol#Config).  This will allow users a better configuration experience when managing `OpentlemetryCollector` resources and will allow operator contributors and maintainers and typed struct to manipulate internally.

Since changing `Spec.Config` would be a significant breaking change, I propose the following process:

1.  Update the operator's internals to convert the string to a structured representation of the collector config and use the struct for all operations.  This is a non-breaking change to users and prepares us for the next steps.
2. Introduce a new field such as `ConfigSpec` that is a structured representation of the collector config.  If this field is set, use it it for all the operations.  To avoid the need to coalesce, require exactly one of `Config` and `ConfigSpec` to be set.
3. Following the deprecation process, deprecate `Config`.
4. (Optional) Follow a rename process to rename `ConfigSpec` to `Config`.

jaronoff97:
I am a big fan of this idea! I think the scope of it is great as well.

As we discussed during our last SIG meeting, I would like to keep the config spec such that it cannot be layered or separated from the CRD at this time. A few ideas were proposed to create an OpenTelemetryConfig CRD that a collector could pull in, but this brings up a myriad of concerns around discoverability and understanding that worry me. The main question I imagine is someone asking "what configuration is my collector actually running" at which point we would need to shrug and pull together the various layers in just the right order. I believe the onus of reusable collector configuration blocks is on the cluster manager and not on the operator.

swiatekm-sumo:
Sounds like a good plan to me as well.

For the prospective OpenTelemetryConfig CRD, I don't think we need to add it now, but we should think about it while we're here. The way I imagined it, is that we'd keep the ability to specify the config directly in the Collector CR, but create a Config CR in the background to represent it, and also have an "effective config" computed field on the Collector. This would mostly be transparent for users just wanting the simple case, and would free us to add more complex config resolution logic in the Config CR without disturbing those simpler use cases.

[...]

PR #2182, opened 2023-09-29
Specify the OpenTelemetry Collector config as YAML

iblancasa:
Relates to #1707.

This PR introduces the following changes:
* Add the `.spec.configSpec` field to the OpenTelemetry Collector CRD

TODO:
* Fix the unit tests
* Fix the E2E tests
* Add an E2E test to ensure the old behavior continues working
* Mark the `.spec.config` field as deprecated
* Use the new structure across all the operator


iblancasa:
Some notes:
* I tried the different approaches described in the different edits from [this comment](https://github.com/open-telemetry/opentelemetry-operator/issues/1707#issuecomment-1721231339) but `controller-gen` always complains
* I tried adding a field to the `OpenTelemetryCollector` type (it would be the ideal thing) in order to store the representation of the information we want. I end up adding it to the `Params` structure (maybe not the best option)
* Using `runtime.RawExtension` following some recommendations from people in the `kubebuilder` channel from the Kubernetes Slack workspace for similar situations

pavolloffay:
We have discussed this at the SIG meeting. We might want to bump the CRD API version and just use the typed config

iblancasa:
> We have discussed this at the SIG meeting. We might want to bump the CRD API version and just use the typed config

The meetings are no longer recorded, right? Can you, at least, add something to #1707?

apis/v1alpha1/config.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"encoding/json"
+
+	"gopkg.in/yaml.v2"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func MustParseConfigSpec(in string) *ConfigSpec {
+	cfg := &ConfigSpec{}
+	yaml.Unmarshal([]byte(in), &cfg)
```
frzifus, line 12:
```suggestion
	if err := yaml.Unmarshal([]byte(in), &cfg); err != nil {
		panic(err)
	}
```

internal/manifests/params.go
```diff
@@ -24,8 +28,75 @@ import (
...
+	configStr string
+	configSpec v1alpha1.ConfigSpec
+}
+
+func (c *OtelConfig) String() string {
+	if c.configStr != "" {
+		return c.configStr
+	}
+	return c.configSpec.String()
+}
```
frzifus, line 95: Why is this otel config needed? Doesn't it duplicate the data from the OpenTelemetryCollector CR?

internal/manifests/params.go
```diff
@@ -24,8 +28,75 @@ import (
...
 
+func NewOtelConfig(instance *v1alpha1.OpenTelemetryCollector) (OtelConfig, error){
+	var cfg OtelConfig
+	if instance.Spec.Config != "" {
+		cfg.configStr = instance.Spec.Config
+		err:=yaml.Unmarshal([]byte(instance.Spec.Config), &cfg)
+		if err!= nil {
+			return cfg, fmt.Errorf("there was a problem parsing the config: %w", err)
+		}
+	} else {
```
frzifus, line 39: Does it make things simpler when we use an early return instead?

controllers/opentelemetrycollector_controller.go
```diff
@@ -175,7 +175,13 @@ func isNamespaceScoped(obj client.Object) bool {
 }
 
 func (r *OpenTelemetryCollectorReconciler) getParams(instance v1alpha1.OpenTelemetryCollector) manifests.Params {
+	otelCfg, err:= manifests.NewOtelConfig(&instance)
+	if err!= nil {
+		r.log.Error(err, "there was a problem while parsing the config")
+	}
+
```
frzifus, line 182: Should we somehow cancel the reconciliation in case of an error?

frzifus (commented):
> I tried the different approaches described in the different edits from https://github.com/open-telemetry/opentelemetry-operator/issues/1707#issuecomment-1721231339 but controller-gen always complains

What did it complain about?

>  I tried adding a field to the OpenTelemetryCollector type (it would be the ideal thing) in order to store the representation of the information we want. I end up adding it to the Params structure (maybe not the best option)

Why is that needed?

> > We have discussed this at the SIG meeting. We might want to bump the CRD API version and just use the typed config

> The meetings are no longer recorded, right? Can you, at least, add something to https://github.com/open-telemetry/opentelemetry-operator/issues/1707?

@iblancasa you can find the recording on [this list](https://docs.google.com/spreadsheets/d/1SYKfjYhZdm2Wh2Cl6KVQalKg_m4NhTPZqq-8SzEVO6s). Here are the [meeting notes](https://docs.google.com/document/d/1Unbs2qp_j5kp8FfL_lRH-ld7i5EOQpsq0I4djkOOSL4/edit#heading=h.okck2w75mwlk).

My internet connection was not so stable in the last meeting. But from what I have gathered, the idea is to raise api version for breaking change of config field and [provide automatic CR  conversion](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-conversion).

PR #2185, opened 2023-10-02
[chore] Change instance to otelcol for reconciliation

jaronoff97:
This is an internal-only PR as a follow up to #1995 and #1559. This simply pulls out the logic to change the signature for building to be more generic to resources beyond a collector, this also splits out some common reconciliation logic


avadhut123pisal (approved):
LGTM :)

frzifus (approved):

swiatekm-sumo (approved):

PR #2186, opened 2023-10-02
Fix .NET Automatic Instrumentation configured by namespace annotation (alpine based images)

### LINKED ISSUE #2179: Annotation for .NET auto-instrumentation runtime ignored when used in namespace
Hello,

I have been playing around with .NET auto-instrumentation as explained here: https://github.com/open-telemetry/opentelemetry-operator/blob/v0.85.0/README.md#opentelemetry-auto-instrumentation-injection

It is mentioned that the annotation to enable auto-instrumentation can be added to a namespace: https://github.com/open-telemetry/opentelemetry-operator/blob/ac63f917cf0a356cc9acf712a688b7b021a963e5/README.md?plain=1#L238-L240
And below there are the annotations to select the .NET runtime. So as I was using a namespace with deployments all referencing images based on `mcr.microsoft.com/dotnet/aspnet:7.0-alpine`, I added the annotations to the namespace like this:
```yaml
instrumentation.opentelemetry.io/inject-dotnet: vsbs-insurance-instrumentation
instrumentation.opentelemetry.io/otel-dotnet-auto-runtime: linux-musl-x64
```

Instrumentation seems to be injected correctly but the environment variable `CORECLR_PROFILER_PATH` points to `/otel-auto-instrumentation/linux-x64/OpenTelemetry.AutoInstrumentation.Native.so` instead of using the path with `linux-musl-x64`. As a result traces are not appearing.

After having a look at the code, it seems that only the pod annotation is considered, while the namespace annotation is not in the scope of the code that selects the runtime: https://github.com/open-telemetry/opentelemetry-operator/blob/bd9f8043994b8a79dd3cf811aae1a6eabc1f2917/pkg/instrumentation/dotnet.go#L72

A workaround is adding the variable `CORECLR_PROFILER_PATH` with the correct path to the env variables of the containers and that makes traces work as expected, but it prevents using only namespace annotations if for example we cannot modify the deployment manifests.

Could someone have a look to check if that's a bug in the code mentioned or I'm missing something?
Thanks in advance

Kielek:
Fixes #2179 

Please let me know if you see better option than `AdditionalAnnotations` to pass annotations from namespaces to injecting steps.

IMO it should affect also `instrumentation.opentelemetry.io/otel-go-auto-target-exe` in current state. I am not sure if it is worth to fix it, but we should have _generic_ option.

### DIFF
--- a/pkg/instrumentation/podmutator.go
+++ b/pkg/instrumentation/podmutator.go
@@ -196,7 +196,6 @@ func NewMutator(logger logr.Logger, client client.Client, recorder record.EventR
 func (pm *instPodMutator) Mutate(ctx context.Context, ns corev1.Namespace, pod corev1.Pod) (corev1.Pod, error) {
 	logger := pm.Logger.WithValues("namespace", pod.Namespace, "name", pod.Name)
 
-	// logger.Info("namespace Name: %s, NamespaceL %s", ns.Name, ns.Namespace)
 	// We check if Pod is already instrumented.
 	if isAutoInstrumentationInjected(pod) {
 		logger.Info("Skipping pod instrumentation - already instrumented")


pkg/instrumentation/podmutator.go
```diff
@@ -195,6 +196,7 @@ func NewMutator(logger logr.Logger, client client.Client, recorder record.EventR
 func (pm *instPodMutator) Mutate(ctx context.Context, ns corev1.Namespace, pod corev1.Pod) (corev1.Pod, error) {
 	logger := pm.Logger.WithValues("namespace", pod.Namespace, "name", pod.Name)
 
+	// logger.Info("namespace Name: %s, NamespaceL %s", ns.Name, ns.Namespace)
```
jaronoff97, line 199: should weleave this in?

jaronoff97 (approved):

pkg/instrumentation/podmutator.go
Kielek: Leftover from my testing. 549cef63daebc58ca74ac9221e78cafcc880c2b9

TylerHelmuth (approved):

Kielek:
@vsbs-david, FYI

vsbs-david:
Thanks a lot!

PR #2189, opened 2023-10-03
Add rate limiting for scrape config updates

swiatekm-sumo:
Limit the rate on updates to scrape configs. The idea and code are very similar to what Prometheus' Service Discovery manager does: https://github.com/prometheus/prometheus/blob/79be1b835789d7c3fde2a907003a8799c308733f/discovery/manager.go#L341. The end result is that we emit events about changes to Prometheus CRs at most every 5 seconds.

This should help with #1544

### DIFF
--- a/cmd/otel-allocator/watcher/promOperator.go
+++ b/cmd/otel-allocator/watcher/promOperator.go
@@ -38,7 +38,7 @@ import (
 	allocatorconfig "github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/config"
 )
 
-const MinEventInterval = time.Second * 5
+const minEventInterval = time.Second * 5
 
 func NewPrometheusCRWatcher(logger logr.Logger, cfg allocatorconfig.Config) (*PrometheusCRWatcher, error) {
 	mClient, err := monitoringclient.NewForConfig(cfg.ClusterConfig)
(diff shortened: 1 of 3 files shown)


pavolloffay (approved):

jaronoff97 (approved):

jaronoff97:
@swiatekm-sumo were you able to test this in a cluster? I'd love to see some metrics about how this change affected usage.

swiatekm-sumo:
> @swiatekm-sumo were you able to test this in a cluster? I'd love to see some metrics about how this change affected usage.

I did a basic smoke test, but I didn't try any synthetic stress test where I'd constantly update a bunch of ServiceMonitors and look at TA CPU usage. I can do that and post some numbers if you're interested.

jaronoff97:
yeah i'd appreciate that if you don't mind :)

swiatekm-sumo:
@jaronoff97 Did a very simple benchmark where I updated the labels on a particular ServiceMonitor as fast as I could via kubectl. Pre-change this used 400m worth of CPU, post change, less than 2m. See attached Prometheus graph:
Screenshot_20231015_133847

cmd/otel-allocator/watcher/promOperator.go
```diff
@@ -37,6 +38,8 @@ import (
 	allocatorconfig "github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/config"
 )
 
+const MinEventInterval = time.Second * 5
```
frzifus, line 41:
```suggestion
const minEventInterval = time.Second * 5
```

cmd/otel-allocator/watcher/promOperator.go
```diff
@@ -79,6 +82,7 @@ func NewPrometheusCRWatcher(logger logr.Logger, cfg allocatorconfig.Config) (*Pr
 		k8sClient:              clientset,
 		informers:              monitoringInformers,
 		stopChannel:            make(chan struct{}),
+		eventInterval:          MinEventInterval,
```
frzifus, line 85:
```suggestion
		eventInterval:          minEventInterval,
```

PR #2193, opened 2023-10-04
Enhanced Reconciliation Errors

### LINKED ISSUE #1972: Put reconciliation errors in CRD events
A user shouldn't need to look at operator logs for CRD reconciliation problems. We should add them proactively on the CRD's events to make clear why something couldn't be reconciled. This will be easier with the completion of https://github.com/open-telemetry/opentelemetry-operator/issues/1876

taniyourstruly:
Closes #1972 

Added reconciliation errors for CRD events in the operator


controllers/opentelemetrycollector_controller_test.go
```diff
@@ -17,240 +17,24 @@ package controllers_test
...
 )
 
 var logger = logf.Log.WithName("unit-tests")
-var mockAutoDetector = &mockAutoDetect{
-	OpenShiftRoutesAvailabilityFunc: func() (autodetect.OpenShiftRoutesAvailability, error) {
-		return autodetect.OpenShiftRoutesAvailable, nil
-	},
-}
-
-func TestNewObjectsOnReconciliation(t *testing.T) {
```
jaronoff97, line 52: this delete is okay IMO, when we went through this together I realized that all of these tests were actually superseded by the tests in reconcile_test.go

internal/manifests/collector/adapters/config_to_ports.go
```diff
@@ -175,7 +176,7 @@ func ConfigToReceiverPorts(logger logr.Logger, config map[interface{}]interface{
 	return ports, nil
 }
 
-func ConfigToPorts(logger logr.Logger, config map[interface{}]interface{}) []corev1.ServicePort {
+func ConfigToPorts(logger logr.Logger, config map[interface{}]interface{}) ([]corev1.ServicePort, error) {
 	ports, err := ConfigToReceiverPorts(logger, config)
 	if err != nil {
 		logger.Error(err, "there was a problem while getting the ports from the receivers")
```
jaronoff97, line 182: suggestion: we should probs propagate here (and below) as well

internal/manifests/collector/configmap.go
```diff
@@ -22,13 +22,14 @@ import (
...
 
-func ConfigMap(params manifests.Params) *corev1.ConfigMap {
+func ConfigMap(params manifests.Params) (*corev1.ConfigMap, error) {
 	name := naming.ConfigMap(params.OtelCol.Name)
 	labels := Labels(params.OtelCol, name, []string{})
 
 	replacedConf, err := ReplaceConfig(params.OtelCol)
 	if err != nil {
 		params.Log.V(2).Info("failed to update prometheus config to use sharded targets: ", "err", err)
+		return nil, err
```
jaronoff97, line 32: suggestion: could you add a test to see this failure?

internal/manifests/collector/container.go
```diff
@@ -165,7 +165,7 @@ func getConfigContainerPorts(logger logr.Logger, cfg string) map[string]corev1.C
 		logger.Error(err, "couldn't extract the configuration")
 		return ports
 	}
-	ps := adapters.ConfigToPorts(logger, c)
+	ps, _ := adapters.ConfigToPorts(logger, c)
```
jaronoff97, line 168: suggestion: we should also propagate the error up here if we can

internal/manifests/collector/container.go
```diff
@@ -165,7 +165,7 @@ func getConfigContainerPorts(logger logr.Logger, cfg string) map[string]corev1.C
 		logger.Error(err, "couldn't extract the configuration")
```
jaronoff97, line 165: suggestion: also here :)

internal/manifests/collector/horizontalpodautoscaler.go
```diff
@@ -25,7 +25,7 @@ import (
 	"github.com/open-telemetry/opentelemetry-operator/internal/naming"
 )
 
-func HorizontalPodAutoscaler(params manifests.Params) client.Object {
+func HorizontalPodAutoscaler(params manifests.Params) (client.Object, error) {
```
jaronoff97, line 28: idea: given this doesn't really return an error, we can remove the error from the signature for now.

internal/manifests/collector/ingress.go
```diff
@@ -28,21 +28,22 @@ import (
...
 
-func Ingress(params manifests.Params) *networkingv1.Ingress {
+func Ingress(params manifests.Params) (*networkingv1.Ingress, error) {
 	if params.OtelCol.Spec.Ingress.Type != v1alpha1.IngressTypeNginx {
-		return nil
+		return nil, nil
 	}
 
-	ports := servicePortsFromCfg(params.Log, params.OtelCol)
+	ports, err := servicePortsFromCfg(params.Log, params.OtelCol)
```
jaronoff97, line 36: this is awesome :D very happy about this change-

internal/manifests/collector/ingress.go
```diff
@@ -28,21 +28,22 @@ import (
...
 
 	// if we have no ports, we don't need a ingress entry
-	if len(ports) == 0 {
+	if len(ports) == 0 || err != nil {
 		params.Log.V(1).Info(
+
```
jaronoff97, line 41: nit: can remove the newline here.

internal/manifests/collector/ingress.go
```diff
@@ -53,6 +54,7 @@ func Ingress(params manifests.Params) *networkingv1.Ingress {
 		rules = createSubdomainIngressRules(params.OtelCol.Name, params.OtelCol.Spec.Ingress.Hostname, ports)
 	}
 
+	//dis is da manifest to create da ingress object
```
jaronoff97, line 57: nit: probs remove :D

internal/manifests/collector/ingress_test.go
```diff
@@ -64,8 +65,9 @@ func TestDesiredIngresses(t *testing.T) {
 			},
 		}
 
-		actual := Ingress(params)
+		actual, err := Ingress(params)
 		assert.Nil(t, actual)
+		assert.ErrorContains(t, err, "couldn't parse the opentelemetry-collector configuration")
```
jaronoff97, line 70: 🎉

internal/manifests/collector/route.go
```diff
@@ -50,10 +50,10 @@ func Routes(params manifests.Params) []*routev1.Route {
 		return nil
 	}
 
-	ports := servicePortsFromCfg(params.Log, params.OtelCol)
+	ports, err := servicePortsFromCfg(params.Log, params.OtelCol)
 
 	// if we have no ports, we don't need a ingress entry
-	if len(ports) == 0 {
+	if len(ports) == 0 || err != nil {
```
jaronoff97, line 56: Can we return error here too?

internal/manifests/collector/service.go
```diff
@@ -53,24 +53,23 @@ func HeadlessService(params manifests.Params) *corev1.Service {
...
 
 	c, err := adapters.ConfigFromString(params.OtelCol.Spec.Config)
 	// TODO: Update this to properly return an error https://github.com/open-telemetry/opentelemetry-operator/issues/1972
 	if err != nil {
 		params.Log.Error(err, "couldn't extract the configuration")
-		return nil
+		return nil, err
 	}
 
 	metricsPort, err := adapters.ConfigToMetricsPort(params.Log, c)
```
jaronoff97, line 70: note to other reviewers, we no longer return an error if we use the default metrics port.

internal/manifests/collector/service.go
```diff
@@ -155,7 +158,7 @@ func Service(params manifests.Params) *corev1.Service {
 			ClusterIP:             "",
 			Ports:                 ports,
 		},
-	}
+	}, err
```
jaronoff97, line 161: nit: we can return nil here if we add err !=nil to the if above.

internal/manifests/collector/serviceaccount.go
```diff
@@ -33,7 +33,7 @@ func ServiceAccountName(instance v1alpha1.OpenTelemetryCollector) string {
 }
 
 // ServiceAccount returns the service account for the given instance.
-func ServiceAccount(params manifests.Params) *corev1.ServiceAccount {
+func ServiceAccount(params manifests.Params) (*corev1.ServiceAccount, error) {
```
jaronoff97, line 36: nit: no need for error here either

internal/manifests/targetallocator/deployment.go
```diff
@@ -24,7 +24,7 @@ import (
 )
 
 // Deployment builds the deployment for the given instance.
-func Deployment(params manifests.Params) *appsv1.Deployment {
+func Deployment(params manifests.Params) (*appsv1.Deployment, error) {
 	name := naming.TargetAllocator(params.OtelCol.Name)
 	labels := Labels(params.OtelCol, name)
```
jaronoff97, line 29: @swiatekm-sumo do you think we should still construct the deployment if we fail to make the configmap here? Right now we pass through, but I'm wondering if we should just reject it?

internal/manifests/targetallocator/deployment_test.go
```diff
@@ -61,7 +61,7 @@ func TestDeploymentNewDefault(t *testing.T) {
 	}
 
 	// test
-	d := Deployment(params)
+	d, _ := Deployment(params)
```
jaronoff97, line 64: suggestion: let's check the error here too... depending on the conversation above we may actually just remove this.

internal/manifests/targetallocator/service.go
```diff
@@ -23,7 +23,7 @@ import (
 	"github.com/open-telemetry/opentelemetry-operator/internal/naming"
 )
 
-func Service(params manifests.Params) *corev1.Service {
+func Service(params manifests.Params) (*corev1.Service, error) {
```
jaronoff97, line 26: nit: no need for error here.

internal/manifests/targetallocator/serviceaccount.go
```diff
@@ -34,7 +34,7 @@ func ServiceAccountName(instance v1alpha1.OpenTelemetryCollector) string {
 }
 
 // ServiceAccount returns the service account for the given instance.
-func ServiceAccount(params manifests.Params) *corev1.ServiceAccount {
+func ServiceAccount(params manifests.Params) (*corev1.ServiceAccount, error) {
```
jaronoff97, line 37: nit: nor here

jaronoff97 (commented):
Overall really excited about this change, a few thoughts here and there.

internal/manifests/targetallocator/deployment.go
swiatekm-sumo: I'd return an error. If we failed here, then we also failed to actually create the ConfigMap, in which case the target allocator won't start anyway. In that case, it's better to do nothing and bubble the error up than create faulty resources.

The best we can do here, of course, is to reject this at admission, but that's beyond the scope of this PR.

controllers/opentelemetrycollector_controller_test.go
frzifus: However, it looks scary :cow2:

internal/manifests/collector/ingress.go
frzifus, line 57:
```suggestion
```

pavolloffay:
@taniyourstruly I don't think that the PR is putting the error from building the objects into the event associated with the collector or into the collector status. See https://github.com/open-telemetry/opentelemetry-operator/blob/c9c3ea37cb500de0fcf232aacbc899815cc81d2b/controllers/opentelemetrycollector_controller.go#L190

internal/config/main_test.go
```diff
@@ -80,7 +80,7 @@ func TestAutoDetectInBackground(t *testing.T) {
 	}
 	cfg := config.New(
 		config.WithAutoDetect(mock),
-		config.WithAutoDetectFrequency(500*time.Second),
```
jaronoff97, line 83: idk why it was like this lol

internal/manifests/collector/adapters/config_to_ports.go
```diff
@@ -217,7 +221,9 @@ func ConfigToMetricsPort(logger logr.Logger, config map[interface{}]interface{})
 	}
 
 	_, port, err := net.SplitHostPort(cOut.Service.Telemetry.Metrics.Address)
-	if err != nil {
+	if err != nil && strings.Contains(err.Error(), "missing port in address") {
```
jaronoff97, line 224: this mean we no longer propagate a bad error

internal/manifests/collector/collector.go
```diff
@@ -62,7 +62,7 @@ func Build(params manifests.Params) ([]client.Object, error) {
 			resourceManifests = append(resourceManifests, res)
 		}
 	}
-	routes := Routes(params)
+	routes, _ := Routes(params)
```
jaronoff97, line 65: should we check the error here?

internal/manifests/collector/container.go
```diff
@@ -43,7 +43,8 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
 	}
 
 	// build container ports from service ports
-	ports := getConfigContainerPorts(logger, otelcol.Spec.Config)
+	ports, err := getConfigContainerPorts(logger, otelcol.Spec.Config)
+	logger.Error(err, "container ports config")
```
jaronoff97, line 47: should we propagate this up from the container? I think logging here for now is fine, maybe we propagate it in a follow up.

internal/manifests/collector/route_test.go
```diff
@@ -45,7 +45,7 @@ func TestDesiredRoutes(t *testing.T) {
 			},
 		}
 
-		actual := Routes(params)
+		actual, _ := Routes(params)
```
jaronoff97, line 48: can we assert no error here? (and for the rest of the changes in this file)

internal/manifests/collector/service.go
```diff
@@ -54,24 +54,23 @@ func HeadlessService(params manifests.Params) *corev1.Service {
...
+	return h, nil
 }
 
-func MonitoringService(params manifests.Params) *corev1.Service {
+func MonitoringService(params manifests.Params) (*corev1.Service, error) {
 	name := naming.MonitoringService(params.OtelCol.Name)
 	labels := manifestutils.Labels(params.OtelCol.ObjectMeta, name, params.OtelCol.Spec.Image, ComponentOpenTelemetryCollector, []string{})
 
 	c, err := adapters.ConfigFromString(params.OtelCol.Spec.Config)
 	// TODO: Update this to properly return an error https://github.com/open-telemetry/opentelemetry-operator/issues/1972
```
jaronoff97, line 65: we can remove this TODO :)

internal/manifests/collector/service.go
```diff
@@ -133,9 +135,10 @@ func Service(params manifests.Params) *corev1.Service {
 	}
 
 	// if we have no ports, we don't need a service
-	if len(ports) == 0 {
+	if len(ports) == 0 || err != nil {
```
jaronoff97, line 138: we've already done the error check at this point, so we can just revert this line

internal/manifests/targetallocator/deployment_test.go
jaronoff97: bump on this

internal/manifests/collector/container_test.go
```diff
@@ -70,12 +95,14 @@ func TestContainerPorts(t *testing.T) {
 	var goodConfig = `receivers:
   examplereceiver:
     endpoint: "0.0.0.0:12345"
+exporters:
+  logging:
```
jaronoff97, line 99: this should be called the debug exporter, docs [here](https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/debugexporter)

internal/manifests/collector/container_test.go
```diff
@@ -184,15 +222,32 @@ service:
 			specConfig: `exporters:
     prometheus:
         endpoint: "0.0.0.0:9090"
+exporters:
+	logging:
```
jaronoff97, line 226: here too

internal/manifests/collector/container_test.go
```diff
@@ -206,12 +261,29 @@ service:
         endpoint: "0.0.0.0:9090"
     prometheus/dev:
         endpoint: "0.0.0.0:9091"
+exporters:
+	logging:
```
jaronoff97, line 265: here as well :)

internal/manifests/targetallocator/deployment_test.go
```diff
@@ -109,7 +111,7 @@ func TestDeploymentPodAnnotations(t *testing.T) {
 	}
 
 	// test
-	ds := Deployment(params)
+	ds, _ := Deployment(params)
```
jaronoff97, line 114: assert no error for all these too

pkg/collector/reconcile/route.go
```diff
@@ -46,7 +46,7 @@ func Routes(ctx context.Context, params manifests.Params) error {
 
 	var desired []*routev1.Route
 	if isSupportedMode {
-		if r := collector.Routes(params); r != nil {
+		if r, _ := collector.Routes(params); r != nil {
```
jaronoff97, line 49: we should propagate this error as well

pkg/collector/reconcile/route_test.go
```diff
@@ -40,7 +40,7 @@ func TestExpectedRoutes(t *testing.T) {
 		params.OtelCol.Spec.Ingress.Type = v1alpha1.IngressTypeRoute
 		params.OtelCol.Spec.Ingress.Route.Termination = v1alpha1.TLSRouteTerminationTypeInsecure
 
-		routes := collector.Routes(params)
+		routes, _ := collector.Routes(params)
```
jaronoff97, line 43: lets also assert no error here

jaronoff97 (approved):

internal/config/main_test.go
frzifus: Timings in test are.. Maybe to prevent it from some flakiness?

internal/manifests/collector/adapters/config_to_ports.go
frzifus: Does it make sense to write a warning in that case? e.g. "missing port in address, fall back to default %d".

frzifus (commented):
Overall lgtm, added some notes.

internal/manifests/collector/adapters/config_to_ports.go
swiatekm-sumo: I recall this log line being a bit spammy, maybe it's better off as a debug?

swiatekm-sumo: Also, we should make this into its own error type instead of using `string.Contains`, which is fragile.

controllers/builder_test.go
```diff
@@ -55,6 +55,8 @@ func TestBuildCollector(t *testing.T) {
 	var goodConfig = `receivers:
   examplereceiver:
     endpoint: "0.0.0.0:12345"
+exporters:
+  logging:
```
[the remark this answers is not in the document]
taniyourstruly: It is to explicitly declare the exporters as they don't always use the default exporters in the tests.

internal/config/main_test.go
jaronoff97: yeah no idea, it was just breaking things locally for me too

internal/manifests/collector/adapters/config_to_ports.go
jaronoff97: yeah very spammy, this removes that :)

.chloggen/target-allocator-delayed-collector-instances.yaml
```diff
@@ -13,4 +13,4 @@ issues: [2350]
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
-subtext:
+subtext:
```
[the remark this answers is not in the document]
jaronoff97: Yeah this is just from a formatter probably

controllers/opentelemetrycollector_controller_test.go
jaronoff97: Yeah, these were particularly bad tests because they actually had no config as part of them, it only checked that things like ingress resources would be generated based on the config. Now that we have enhanced checks for valid configuration all of these tests would fail. We could fix these by supplying real config, BUT all of these checks are handled in the reconcile_test.go file and builder_test.go which check for the existence of expected resource.

internal/manifests/collector/adapters/config_to_ports.go
taniyourstruly: image
I tried for 30 mins with Jacob and others to do the type assertion on the error but the net package does not expose the error in a friendly way to make that possible. The Go compiler was complaining but I couldn't make it work and this was the closest I could get it. I am open to suggestions on how to fix this.

swiatekm-sumo: I think the idiomatic way to do this is the following:

```go
	var addrErr *net.AddrError
	if err != nil {
		if errors.As(err, &addrErr) &&
			addrErr.Addr == cOut.Service.Telemetry.Metrics.Address &&
			addrErr.Err == "missing port in address" {
			return 8888, nil
		}
		return 0, err
	}
```
Somewhat confusingly, the second argument to errors.As should be a pointer to a pointer. But I don't think this actually makes things any more safe, so we can keep it as is.

swiatekm-sumo (approved):
LGTM

jaronoff97:
Thank you for your contribution @taniyourstruly :D

Issue #1368, opened 2023-01-13
Create RemoteConfiguration CRD in the operator


jaronoff97:

avadhut123pisal:
Hi @jaronoff97 / @pavolloffay , I can start working on this if it is not already in progress.

jaronoff97:
@avadhut123pisal before this is possible, #1369 will need to be done so that there is an image that the operator can use. Would you be interested in setting up #1369 ?

avadhut123pisal:
> @avadhut123pisal before this is possible, #1369 will need to be done so that there is an image that the operator can use. Would you be interested in setting up #1369 ?

Yes. Please assign that one to me.

jaronoff97:
@avadhut123pisal i can only assign issues to those who are members of the open-telemetry org i believe. You can work on it regardless, though :)

avadhut123pisal:
Hi @jaronoff97, as we have container image for OpAMP Bridge published, can we start working on this ?

jaronoff97:
yep! should be good to add this in now :) Thank you!

avadhut123pisal:
> yep! should be good to add this in now :) Thank you! Let me know if you need a hand with defining the fields required for the CRD.

As per the current implementation of the OpAMP Bridge, I can see these fields are required:

```
type Config struct {
	Endpoint     string   `yaml:"endpoint"`
	Protocol     string   `yaml:"protocol"`
	Capabilities []string `yaml:"capabilities"`

	// ComponentsAllowed is a list of allowed OpenTelemetry components for each pipeline type (receiver, processor, etc.)
	ComponentsAllowed map[string][]string `yaml:"components_allowed,omitempty"`
}
```

Do we need any other fields ?

jaronoff97:
This is closed by #1559

PR #2196, opened 2023-10-05
Refactor OpAMP bridge logging and configuration

jaronoff97:
As a side quest from #1368, this PR changes how config is loaded in to the bridge as well as the logging structure.

This has some breaking changes around the config. This is okay because currently no one is running the bridge (AFAIK!)

`components_allowed` => `componentsAllowed`
:x: `protocol` which is now inferred from endpoint
capabilities `[]string` => `map[Capability]bool` for enhanced configuration validation


srikanthccv (approved):
LGTM

avadhut123pisal:
@jaronoff97 What is your view on https://github.com/open-telemetry/opentelemetry-operator/pull/1559#discussion_r1347470831 ? Because, that would require change in the OpAMP bridge code as well as in the Operator.

jaronoff97:
@avadhut123pisal i like the recommendation, ill incorporate it here

srikanthccv:
My reasoning behind using typed struct is as the bridge implementation improves, we will find ourselves checking if a capability exists before performing certain actions (for instance, AcceptConnectionSettings). 

I find it better to do some like `if b.Capabilities.AcceptsRemoteConfig` than enumerating the options and comparing.

jaronoff97:
We are able to do that with a map check with an enum as well. I'm trying to avoid needing to do an if on every struct field for validation or defaulting.

srikanthccv:
I am more biased towards the end user's POV/experience than what needs to be done in our codebase. The enums/strings are prone to typos. The struct makes it clear about the fixed set of capabilities that the user can enable or disable and works nicely with the IDE yaml schema generation and validation. I don't have a strong opinion. I am fine with either of the approaches.

jaronoff97:
Using a map is the same user experience while also being more ergonomic for use in go by creating constants for each capability.

avadhut123pisal (approved):
LGTM :)

cmd/operator-opamp-bridge/agent/agent.go
```diff
@@ -64,7 +66,7 @@ func NewAgent(logger types.Logger, applier operator.ConfigApplier, config config
 		opampClient:         opampClient,
 	}
 
-	agent.logger.Debugf("Agent created, id=%v, type=%s, version=%s.",
+	agent.logger.V(3).Info("Agent created, id=%v, type=%s, version=%s.",
```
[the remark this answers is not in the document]
jaronoff97: ah yes, i should do this. this was copied from the example code and previously that type of kv thing wasn't supported :)

cmd/operator-opamp-bridge/main.go
```diff
@@ -18,37 +18,38 @@ import (
...
 	l := config.GetLogger()
-	cliConf, err := config.ParseCLI(l.WithName("cli-config"))
+
+	flagSet := config.GetFlagSet(pflag.ExitOnError)
+	err := flagSet.Parse(os.Args)
 	if err != nil {
-		l.Error(err, "unable to load ")
-		os.Exit(1)
+		l.Error(err, "Unable to load flags")
+		return
```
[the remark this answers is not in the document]
jaronoff97: yeah, we should still return 1. fixing.

cmd/operator-opamp-bridge/config/config.go
```diff
@@ -16,43 +16,99 @@ package config
...
+	AcceptsRestartCommand          Capability = "AcceptsRestartCommand"
+	ReportsHealth                  Capability = "ReportsHealth"
+	ReportsRemoteConfig            Capability = "ReportsRemoteConfig"
 )
 
 type Config struct {
-	Endpoint     string   `yaml:"endpoint"`
-	Protocol     string   `yaml:"protocol"`
-	Capabilities []string `yaml:"capabilities"`
+	// KubeConfigFilePath empty if in cluster configuration is in use
```
gdfast, line 87: Something about this wording is a little confusing, maybe you have an idea of how to clarify. Or maybe there's an extra "in" in here

cmd/operator-opamp-bridge/config/config.go
```diff
@@ -16,43 +16,99 @@ package config
...
+	return nil
+}
+
+func GetLogger() logr.Logger {
+	return zap.New(zap.UseFlagOptions(&zapCmdLineOpts))
+}
+
+type Capability string
+
+const (
```
gdfast, line 69: nit: is there an approach to the order of these consts? I think alpha-ordering might be easier to look at and review

gdfast (approved):
LGTM!

cmd/operator-opamp-bridge/config/config.go
jaronoff97: this came from the upstream opamp-go ordering, but agreed... it's a bit odd.

cmd/operator-opamp-bridge/config/config.go
jaronoff97: yeah i can clear that up

gdfast (approved):

.chloggen/refactor-bridge-config.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+# One or more tracking issues related to the change
+issues: [1368]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  `components_allowed` => `componentsAllowed`
+  :x: `protocol` which is now inferred from endpoint
```
pavolloffay, line 18: what does `:x:` mean?  Does it mean that is has been removed?

pavolloffay (approved):

.chloggen/refactor-bridge-config.yaml
jaronoff97: Yes

PR #2204, opened 2023-10-10
Remove autoscaling v2beta2

pavolloffay:


jaronoff97 (approved):

swiatekm-sumo:
This looks like it should merit a changelog entry (even though it's technically a noop with our support starting at K8s 1.23).

pavolloffay:
Isn't this an internal detail? The old API version is not created by the operator anymore.

swiatekm-sumo:
> Isn't this an internal detail? The old API version is not created by the operator anymore.

Ah, makes sense. I misunderstood the scope of this change.

swiatekm-sumo (approved):

PR #2210, opened 2023-10-11
[chore] Refactor Webhooks to their own packages

jaronoff97:
**Description:** 
This PR updates how webhooks are created and managed by the operator by moving them out of types and into an internal package. In the future this will let us add more complex logic to the webhooks, but for now this mostly is a no-op. One benefit for instrumentation webhooks is that we no longer need to add the annotations to the resource for defaulting because we can add a config field to the instrumentation webhook. All tests have been moved over and replaced with the actual webhooks.

**Link to tracking Issue:** n/a

**Testing:** tested locally.

**Documentation:** n/a


config/webhook/manifests.yaml
jaronoff97, line 1: The order of these is reversing, but is a noop AFAICT

pkg/instrumentation/upgrade/upgrade.go
```diff
@@ -74,97 +88,54 @@ func (u *InstrumentationUpgrade) ManagedInstances(ctx context.Context) error {
...
-		}
-	}
-	autoInstNginx := inst.Annotations[v1alpha1.AnnotationDefaultAutoInstrumentationNginx]
-	if autoInstNginx != "" {
-		if featuregate.EnableNginxAutoInstrumentationSupport.IsEnabled() {
-			// upgrade the image only if the image matches the annotation
-			if inst.Spec.Nginx.Image == autoInstNginx {
-				inst.Spec.Nginx.Image = u.DefaultAutoInstNginx
-				inst.Annotations[v1alpha1.AnnotationDefaultAutoInstrumentationNginx] = u.DefaultAutoInstNginx
+func (u *InstrumentationUpgrade) upgrade(_ context.Context, inst v1alpha1.Instrumentation) *v1alpha1.Instrumentation {
```
jaronoff97, line 91: the previous way was really annoying me. If anyone has problems with this, i am happy to revert it.

internal/webhook/instrumentation/webhook.go
```diff
@@ -188,29 +205,85 @@ func (r *Instrumentation) Default() {
...
+	if r.Annotations == nil {
+		r.Annotations = map[string]string{}
+	}
+	r.Annotations[constants.AnnotationDefaultAutoInstrumentationJava] = w.cfg.AutoInstrumentationJavaImage()
+	r.Annotations[constants.AnnotationDefaultAutoInstrumentationNodeJS] = w.cfg.AutoInstrumentationNodeJSImage()
+	r.Annotations[constants.AnnotationDefaultAutoInstrumentationPython] = w.cfg.AutoInstrumentationPythonImage()
+	r.Annotations[constants.AnnotationDefaultAutoInstrumentationDotNet] = w.cfg.AutoInstrumentationDotNetImage()
+	r.Annotations[constants.AnnotationDefaultAutoInstrumentationGo] = w.cfg.AutoInstrumentationGoImage()
+	r.Annotations[constants.AnnotationDefaultAutoInstrumentationApacheHttpd] = w.cfg.AutoInstrumentationApacheHttpdImage()
+	r.Annotations[constants.AnnotationDefaultAutoInstrumentationNginx] = w.cfg.AutoInstrumentationNginxImage()
```
TylerHelmuth, line 218: Where were we setting these values before?

cmd/operator-opamp-bridge/agent/agent_test.go
```diff
@@ -195,7 +195,7 @@ func TestAgent_onMessage(t *testing.T) {
 					},
 				},
 				status: &protobufs.RemoteConfigStatus{
-					LastRemoteConfigHash: []byte("good/testnamespace405"),
+					LastRemoteConfigHash: []byte("good/testnamespace401"),
```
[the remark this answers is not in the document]
jaronoff97: this was a bug that got pushed when i merged my config PR somehow, not sure how it got through.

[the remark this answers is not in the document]
jaronoff97: i could push a separate PR with this if you want, i just wanted my tests to pass haha

TylerHelmuth (approved):
LGTM

swiatekm-sumo (approved):
LGTM

config/webhook/manifests.yaml
```diff
@@ -10,9 +10,9 @@ webhooks:
     service:
       name: webhook-service
       namespace: system
-      path: /mutate-opentelemetry-io-v1alpha1-instrumentation
+      path: /mutate-opentelemetry-io-v1alpha1-opentelemetrycollector
```
pavolloffay, line 13: why are these changes here? 

I am concerned that this will cause issues for the upgrade (also the upgrade test was changed )

jaronoff97:
@pavolloffay should be good to go now!

internal/webhook/sidecar/webhookhandler.go
```diff
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package webhookhandler contains the webhook that injects sidecars into pods.
-package webhookhandler
+// Package sidecar contains the webhook that injects sidecars into pods.
+package sidecar
```
pavolloffay, line 16: I don't think we should call this package `sidecar` 

The package exposes a mutator API that allows mutating pod objects. It is used for instrumentation and sidecar

pavolloffay (changes requested):

pavolloffay (approved):

PR #2215, opened 2023-10-12
Publish operator images for IBM P/Z

pavolloffay:
The collector is published for IBM P/Z https://github.com/open-telemetry/opentelemetry-collector-releases/pkgs/container/opentelemetry-collector-releases%2Fopentelemetry-collector/136233520?tag=0.87.0


swiatekm-sumo:
Is it actually possible to run Kubernetes on these architectures? If so, is anyone actually doing it and interested in running the otel operator? I'm asking because we don't have any straightforward way of validating these images.

pavolloffay:
Yes, internally we run tests for these architectures

swiatekm-sumo (approved):

TylerHelmuth (approved):

TylerHelmuth:
1 think we might want to note is that the Go auto-instrumentation image is not built for those 2 architectures.

frzifus (approved):

iblancasa (approved):

pavolloffay:
> 1 think we might want to note is that the Go auto-instrumentation image is not built for those 2 architectures.

@iblancasa is working on publishing the auto-instrumentation libs for different archs. Other langs might have a similar limitation (e.g. .net).

PR #2220, opened 2023-10-13
Update the OpenTelemetry Java agent version to 1.31.0

opentelemetrybot:
Update the OpenTelemetry Java agent version to `1.31.0`.


TylerHelmuth (approved):

TylerHelmuth:
@open-telemetry/operator-approvers we've started requesting changelogs for these types for changes so if we want to continue using this action going forward we should update it to generate a changelog.

pavolloffay:
> we've started requesting changelogs for these types for changes

Has this been documented anywhere? 

@TylerHelmuth the auto-instrumentation versions are also listed in the component section in the release. Therefore users can see which versions have changed.

TylerHelmuth:
> Has this been documented anywhere?

@pavolloffay no, its an inconsitent behavior I've noticed.  I do think we should take a strong stance one way or the other.

pavolloffay (approved):

PR #2222, opened 2023-10-13
[chore] Replace the images for the E2E tests with the ones built in this repository

### LINKED ISSUE #2073: Use business applications hosted in this repository for the instrumentation E2E tests

iblancasa:
Resolves #2073


config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: ghcr.io/iblancasa/opentelemetry-operator/opentelemetry-operator
+  newTag: 0.85.0-39-g7e232dd
```
frzifus, line 8:
```suggestion
```

pavolloffay (approved):

jaronoff97 (approved):

Issue #2208, opened 2023-10-11
Move Target Allocator Rewrite Feature Gate to Beta


jaronoff97:
https://github.com/open-telemetry/opentelemetry-operator/blob/282c169f0499694dd0b780cf225bc8680c05071d/pkg/featuregate/featuregate.go#L77-L84

This feature gate has been in alpha for a few versions, I think it's time we move it to beta! We recently got a [stackoverflow post](https://stackoverflow.com/questions/77269471/opentelemetry-target-allocator-does-not-allocate-discovered-servicemonitors) about this where having it on would have fixed the user problem.

jaronoff97:
Closed by #2231

PR #2231, opened 2023-10-15
Enable Target Allocator Rewrite by default

swiatekm-sumo:
**Description:**
Move the `operator.collector.rewritetargetallocator` flag to `Beta`, enabling it by default.

**Link to tracking Issue:** #2208

**Testing:**
Converted existing tests for both the flag-enabled and flag-disabled cases. Added some unit tests to check the basic rewrite case. Also added a check for the content of the created ConfigMap in E2e tests.

**Documentation:**
Reflected the fact that this is now enabled by default in the main README.

### DIFF
--- a/.chloggen/ta-reload-beta.yaml
+++ b/.chloggen/ta-reload-beta.yaml
@@ -1,0 +1,18 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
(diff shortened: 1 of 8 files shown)


jaronoff97 (approved):
This LGTM. Can you also add a test for if someone has a `target_allocator` config set already to show the override behavior? I don't think we have one of those already and now that this is default it would be good to show we won't break anyone who is already setting this.

swiatekm-sumo:
@jaronoff97 done

jaronoff97 (approved):

Issue #2207, opened 2023-10-11
Release the operator v0.87.0


dmitryax:

pavolloffay:
@TylerHelmuth you are next on the list :) https://github.com/open-telemetry/opentelemetry-operator/blob/main/RELEASE.md#release-schedule

TylerHelmuth:
@pavolloffay do you want to go right into this one or wait a week since we just released 0.86.0?

pavolloffay:
I don't have any preference. It depends if other folks want to get in some functionality cc) @open-telemetry/operator-approvers @open-telemetry/operator-maintainers

pavolloffay:
I would like to get this one in https://github.com/open-telemetry/opentelemetry-operator/pull/2215

mx-psi:
We have some guidance here around patch releases: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/release.md#bugfix-release-criteria

> We aim to provide a release that fixes security-related issues in at most 30 days since they are publicly announced; with the current release schedule this means security issues will typically not warrant a bugfix release. An exception is critical vulnerabilities (CVSSv3 score >= 9.0), which will warrant a release within five business days.

Do we consider this to be a critical vulnerability? The score seems to be 5.3 (moderate) on Github's tracker, but I am not sure if this is what we should be looking at (the CVE talks about Swift specifically).

pavolloffay:
Thanks for the info. This is probably a better tracker https://github.com/advisories/GHSA-4374-p667-p6c8

mx-psi:
Some more trackers:
- https://access.redhat.com/security/cve/cve-2023-39325
- https://ubuntu.com/security/CVE-2023-39325
- https://www.suse.com/security/cve/CVE-2023-39325.htmll

The CVSS v3 Base Score is 7.5 and is classified as 'Important' (in between Moderate and Critical) or of 'Medium' priority.

TylerHelmuth:
I am prepping the release PR, but I believe we need to merge https://github.com/open-telemetry/opentelemetry-operator/pull/2239 before the next release.

PR #2242, opened 2023-10-17
Prep 0.87.0 release

TylerHelmuth:
closes https://github.com/open-telemetry/opentelemetry-operator/issues/2207


config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
-- manager.yaml
\ No newline at end of file
+- manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
```
jaronoff97, line 7: This should be removed

TylerHelmuth: Thanks

swiatekm-sumo:
We've had a couple more changes merged since you opened this PR. I don't mind if they have to wait for the next release, just be careful when tagging not to include them. But it's probably simpler if you rebase on main and regenerate the changelog.

PR #2250, opened 2023-10-19
fix(operator): truncate sidecar pod injected label to 63 chars

### LINKED ISSUE #1031: Label "sidecar.opentelemetry.io/injected" value sometimes longer than 63 characters

RoVernekar:
**Description:**

Truncate `sidecar.opentelemetry.io/injected` sidecar pod label to 63 characters since Kubernetes labels must be 63 characters or less.

**Link to tracking Issue:**

resolves #1031

**Testing:**

Tested with a unit test in `pod_test.go`.

**Documentation:**

### DIFF
--- a/pkg/sidecar/pod.go
+++ b/pkg/sidecar/pod.go
@@ -28,8 +28,8 @@ import (
 )
 
 const (
-	label      = "sidecar.opentelemetry.io/injected"
-	confEnvVar = "OTEL_CONFIG"
+	injectedLabel = "sidecar.opentelemetry.io/injected"
+	confEnvVar    = "OTEL_CONFIG"
 )
 
 // add a new sidecar container to the given pod, based on the given OpenTelemetryCollector.
@@ -53,7 +53,7 @@ func add(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelemetryCo
 	if pod.Labels == nil {
 		pod.Labels = map[string]string{}
 	}
-	pod.Labels[label] = fmt.Sprintf("%s.%s", otelcol.Namespace, otelcol.Name)
+	pod.Labels[injectedLabel] = naming.Truncate("%s.%s", 63, otelcol.Namespace, otelcol.Name)
 
 	return pod, nil
 }
--- a/.chloggen/truncate-pod-label.yaml
+++ b/.chloggen/truncate-pod-label.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Truncate `sidecar.opentelemetry.io/injected` sidecar pod label to 63 characters
+
+# One or more tracking issues related to the change
+issues: [1031]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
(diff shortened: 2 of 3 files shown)


pkg/sidecar/pod.go
```diff
@@ -53,7 +53,7 @@ func add(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelemetryCo
 	if pod.Labels == nil {
 		pod.Labels = map[string]string{}
 	}
-	pod.Labels[label] = fmt.Sprintf("%s.%s", otelcol.Namespace, otelcol.Name)
+	pod.Labels[label] = naming.Truncate("%s.%s", 63, otelcol.Namespace, otelcol.Name)
```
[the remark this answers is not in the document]
RoVernekar: How about `naming.PodInstanceLabel`? The same logic is used in a few other places so I figured we could reuse this function but couldn't think of a great name.

Also I renamed `label` to `injectedLabel`. How's that?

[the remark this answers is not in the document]
jaronoff97: perfect :)

jaronoff97:
@RoVernekar youll need to fill out the CLA before we can merge

RoVernekar:
> @RoVernekar youll need to fill out the CLA before we can merge

Sorry for the delay, should be good now.

internal/naming/main.go
```diff
@@ -124,3 +124,8 @@ func ServiceMonitor(otelcol string) string {
 func TargetAllocatorServiceAccount(otelcol string) string {
 	return DNSName(Truncate("%s-targetallocator", 63, otelcol))
 }
+
+// PodInstanceLabel returns a label value containing the namespace and instance name.
+func PodInstanceLabel(namespace string, otelcol string) string {
```
pavolloffay, line 129: is `Pod` the right prefix we want to use? Shouldn't this be just `InstanceLabel` ?

pavolloffay: I would revert other changes in this PR . The naming/main.go is used for k8s object names. 

Or expose this function in the labels.go

RoVernekar: Okay I went back to the original changes. Let me know if you would prefer I expose this in `labels.go` instead.

RoVernekar: @pavolloffay can you take another look at this when you get a chance?

yuriolisa (approved):

swiatekm-sumo (approved):

Issue #1921, opened 2023-07-13
add service.instance.id into resource env


Duncan-dsf:
Why not add service.instance.id into resources? Is there any consideration?

https://github.com/open-telemetry/opentelemetry-operator/blob/02b4d30b5786dc1f0ec03ccf63a91013c229d515/pkg/instrumentation/sdk.go#L324

<img width="1162" alt="image" src="https://github.com/open-telemetry/opentelemetry-operator/assets/29371215/d51cd42d-7b92-44a5-84b7-4a8af66f5fc2">

TylerHelmuth:
We could.  Is that something you'd like to contribute.

yuriolisa:
I can take this on, ok?

PR #2259, opened 2023-10-23
Added service.instance.id into resource env

yuriolisa:

Added the service.instance.id as os.Hostname into the traces resource Env.

#1921 

Changed the unit tests to include the `service.instance.id` information.


pkg/instrumentation/sdk.go
```diff
@@ -424,7 +425,7 @@ func (i *sdkInjector) createResourceMap(ctx context.Context, otelinst v1alpha1.I
 			res[k] = v
 		}
 	}
-
+	serviceInstanceId, _ := os.Hostname()
```
TylerHelmuth, line 428: If an error is returned we can ignore it, but I don't think we should set the serviceInstanceId anymore.

yuriolisa: Do you have any suggestion for values to be set as service.instance.id?

TylerHelmuth: I propose we only set if it is retrievable.  If `os.Hostname()` returns an error we should not set it.

srikanthccv: Does `os.Hostname()` guarantee the requirement?

>_service.namespace,service.name,service.instance.id triplet MUST be globally unique_

Collector uses uuid following the semconv guideline 

>_If the service has no inherent unique ID that can be used as the value of this attribute it is recommended to generate a random Version 1 or Version 4 RFC 4122 UUID (services aiming for reproducible UUIDs may also use Version 5, see RFC 4122 for more recommendations)_

yuriolisa: ok, so I basically created a new UUID and implemented it as a service.instance.id.

.chloggen/new-service-instance-id.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: Operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Added the service.instance.id as os.Hostname into the traces resource Env.
```
pavolloffay, line 8: Is this accurate? I don't see how the hostname is used for the service ID


> Added the service.instance.id as os.Hostname into the resource attributes.

pkg/instrumentation/sdk.go
pavolloffay: I think we can do better than generating a new UUID. We should consider using the pod UID which should be unique.

.chloggen/new-service-instance-id.yaml
yuriolisa: It was not accurate since I changed to UUID in the meantime. Anyway, I've put the pod.UID to be the service.instance.id.

pkg/instrumentation/sdk.go
yuriolisa: Replaced to the pod.UID.

TylerHelmuth (approved):

pkg/instrumentation/sdk.go
```diff
@@ -433,6 +433,7 @@ func (i *sdkInjector) createResourceMap(ctx context.Context, otelinst v1alpha1.I
 	k8sResources[semconv.K8SPodNameKey] = pod.Name
 	k8sResources[semconv.K8SPodUIDKey] = string(pod.UID)
 	k8sResources[semconv.K8SNodeNameKey] = pod.Spec.NodeName
+	k8sResources[semconv.ServiceInstanceIDKey] = string(pod.UID)
```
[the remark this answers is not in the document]
yuriolisa: Just changed the service.instance.id to follow the conventions defined by the link you mentioned above. Could you please have a look?

frzifus (approved):

jaronoff97 (approved):

Issue #2262, opened 2023-10-24
Prometheus Receiver with TargetAllocator watching ServiceMonitors only adds Scrape Jobs but not actually scraping


changexd:
### Component(s)

collector, target allocator

## Description
Following the tutorial for setting up TA with Prometheus serviceMonitor in [target allocator](https://github.com/open-telemetry/opentelemetry-operator/blob/main/cmd/otel-allocator/README.md)

1. the collector displays "Scrape job Added" along with serviceMonitor name, but it doesn't scrape anything except for its own static scrape config
2. TargetAllocator endpoint "/scrape_configs" outputs the converted serviceMonitor, but not displaying jobs in "/jobs"

## Steps to Reproduce
1. kubectl apply -f https://github.com/open-telemetry/opentelemetry-operator/releases/latest/download/opentelemetry-operator.yaml
2. apply resource from tutotrial with custom service-monitor
```
apiVersion: opentelemetry.io/v1alpha1
kind: OpenTelemetryCollector
metadata:
  name: prometheus-collector
  namespace: monitoring
spec:
  mode: statefulset
  targetAllocator:
    enabled: true
    prometheusCR:
      scrapeInterval: 5s
      enabled: true
  config: |
    receivers:
      prometheus:
        config:
          scrape_configs:
          - job_name: 'otel-collector'
            scrape_interval: 10s
            static_configs:
            - targets: [ '0.0.0.0:8888' ]
        target_allocator:
          endpoint: http://prometheus-collector-targetallocator
          interval: 5s
          collector_id: "${POD_NAME}"
    exporters:
      debug:

    service:
      pipelines:
        metrics:
          receivers: [prometheus]
          processors: []
          exporters: [debug]
```

## Expected Result
1. collector should be scraping jobs allocated by the targetAllocator and output to the console
2. target allocator endpoints "/jobs" should be displaying jobs from the "/scrape_configs" endpoint 

## Actual Result
1. collector only collects static target but did add scrape jobs
```
2023-10-24T06:54:23.610Z    info    service@v0.87.0/telemetry.go:84    Setting up own telemetry...
2023-10-24T06:54:23.610Z    info    service@v0.87.0/telemetry.go:201    Serving Prometheus metrics    {"address": ":8888", "level": "Basic"}
[...]

PR #2273, opened 2023-10-25
reset kubeconfig to empty string when using in-cluster config

changexd:
**Description:** 
reset kubeconfig to empty string when using in-cluster config

**Link to tracking Issue:**
 #2262

according to the discussion in #2262, it seems like we haven't added integration tests currently  to check if prometheusCR and target allocator setup actually checks metrics. So I tested on my own cluster making sure the bug is fixed.

As this is my very first pull request and with lack of experience, I simply made changes with the solution provided in the issue, but if a helping hand is needed in adding integration tests, I would love to be part of it too!


swiatekm-sumo (approved):

I'd like to hold off on merging this until we have tests demonstrating it fixes #2262. I should be done with those today.

swiatekm-sumo:
We're good to merge, see #2275

jaronoff97 (approved):

PR #2275, opened 2023-10-25
Add E2E test for metrics collection using Prometheus CRs

swiatekm-sumo:
**Description:**

Adds an E2E test verifying collecting metrics defined by a ServiceMonitor using the target allocator.

This is done by:
* adding a ServiceMonitor covering the collector's own metrics port
* scraping the metrics with Prometheus receive,  target allocator enabled
* exporting the metrics on a different port via Prometheus exporter
* running a Job which curls the above endpoint and checks that a particular metric is present

This is a little bit convoluted, but I haven't figured out how to make it any simpler. The Job approach actually ends up working well with kuttl, as we end up just needing to check the status of a K8s resource.

**Link to tracking Issue:** #2262 

Note: This will fail before #2273 is merged. The intent of this test is to catch issues like #2262 in the future.

### DIFF
--- a/tests/e2e/targetallocator-prometheuscr/00-install.yaml
+++ b/tests/e2e/targetallocator-prometheuscr/00-install.yaml
@@ -1,0 +1,137 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: ta
+automountServiceAccountToken: true
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: collector
+automountServiceAccountToken: true
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: ta
+rules:
+- apiGroups: [""]
+  resources:
+    - pods
+    - nodes
+    - services
+    - endpoints
+    - configmaps
+    - secrets
+    - namespaces
+  verbs:
+    - get
+    - watch
+    - list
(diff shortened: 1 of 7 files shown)


tests/e2e/targetallocator-prometheuscr/00-install.yaml
```diff
@@ -0,0 +1,137 @@
...
+kind: OpenTelemetryCollector
+metadata:
+  name: prometheus-cr
+spec:
+  mode: statefulset
+  serviceAccount: collector
+  targetAllocator:
+    enabled: true
+    serviceAccount: ta
+    image: "ghcr.io/sumo/opentelemetry-operator/target-allocator:v0.87.0-9-g7070b99"
```
jaronoff97, line 107: This should be the local image right?

tests/e2e/targetallocator-prometheuscr/01-install.yaml
jaronoff97, line 1: Can we also add a Check that the job appears in the TA's scrape configs endpoint and jobs endpoint?

swiatekm-sumo: I can easily check if the job name appears in the output of either of these endpoints. Actually parsing them I think might be a bit complex to do in bash.

jaronoff97 (approved):

Issue #2131, opened 2023-09-18
Modify bridge to report on status of any annotated collectors


jaronoff97:
[Per the design of the bridge](https://docs.google.com/document/d/1M8VLNe_sv1MIfu5bUR5OV_vrMBnAI7IJN-7-IAr37JY/edit#heading=h.wmpqbyhl460g) it should use the annotation `opentelemetry.io/opamp-reporting: true` to report on the status of any annotated collector pool through the OpAMP's effective configuration in the report status message.

jaronoff97:
closed by #2277

PR #2277, opened 2023-10-25
Update OpAMP Bridge to comply with its design

jaronoff97:
**Description:** 
Right now, the bridge doesn't entirely conform to the design, it didn't support an annotation for reporting on managed instances. The naming structure also didn't match the design here:

> As part of the Bridge’s AgentToServer message, the effective configuration reported will be a map of a collector’s namespaced name `(<namespace>/<otelcol-name>)` to that collector’s CRD resource. This will let the server know what should exist in the cluster. This data can be used to enrich information received from the collector’s OpAMP extension by joining on the collector’s pool name.

Further, we also weren't allowing servers to send full CRDs, only specs (another inaccuracy)

**Link to tracking Issue:** #2131 

**Testing:** Tests were updated to reflect the changes

### DIFF
--- a/cmd/operator-opamp-bridge/agent/agent_test.go
+++ b/cmd/operator-opamp-bridge/agent/agent_test.go
@@ -216,7 +216,7 @@ func TestAgent_onMessage(t *testing.T) {
 				status: &protobufs.RemoteConfigStatus{
 					LastRemoteConfigHash: []byte("testnamespace/bad513"),
 					Status:               protobufs.RemoteConfigStatuses_RemoteConfigStatuses_FAILED,
-					ErrorMessage:         "yaml: line 20: could not find expected ':'",
+					ErrorMessage:         "error converting YAML to JSON: yaml: line 21: could not find expected ':'",
 				},
 			},
 		},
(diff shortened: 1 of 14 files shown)


cmd/operator-opamp-bridge/agent/agent_test.go
```diff
@@ -167,7 +167,7 @@ func TestAgent_onMessage(t *testing.T) {
 					},
 				},
 				status: &protobufs.RemoteConfigStatus{
-					LastRemoteConfigHash: []byte("good/testnamespace401"),
+					LastRemoteConfigHash: []byte("testnamespace/good512"),
```
swiatekm-sumo, line 170: I'd consider not hardcoding these values, but instead computing them in the test body. This looks like it's going to be annoying to maintain, and at a glance it's also difficult to know what this value is actually supposed to be - I had to read through the whole test setup to figure out that `512` is the length of the config file.

swiatekm-sumo (approved):
LGTM

cmd/operator-opamp-bridge/agent/agent_test.go
jaronoff97: true. this has definitely caused some annoyances. I'll re-work this and push that up.

Issue #2183, opened 2023-10-02
Bump the OpenTelemetry Operator CRD version


iblancasa:
On the last SIG meeting there was agreement to bump CRD API version and in the new version drop the string config and expose only the typed config.

_Originally posted by @pavolloffay in https://github.com/open-telemetry/opentelemetry-operator/issues/1707#issuecomment-1742595238_
            

26Oct2023. Update [after SIG meeting](https://docs.google.com/document/d/1Unbs2qp_j5kp8FfL_lRH-ld7i5EOQpsq0I4djkOOSL4/edit):
1. Create the CRD types, no manifests or anything
2. Have the operator reconcile it behind a feature flag 
3. After some time, put the reconciliation feature flag in beta
4. Webhook Conversion behind a new feature flag 
5. After some time, move the Webhook Conversion feature flag to beta
6. Move the feature flags to stable
7. Deprecate the old types

iblancasa:
I started working on this.

frzifus:
Should we not first figure out how the new ConfigSpec is implemented? Since thats the main reason to bump the version right now?

iblancasa:
> Should we not first figure out how the new ConfigSpec is implemented? Since thats the main reason to bump the version right now?

Where do you want to implement it? In the old CRD?

bryan-aguilar:
How long will the previous CRD version be supported for?

iblancasa:
@bryan-aguilar as soon as the new one is implemented, we can take a decision about that.

iblancasa:
> Another follow up question, will improvements to the CRD be made to both versions while the previous version is supported? Or will the older version no longer receive updates?

The idea is to use this: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-conversion

So, the instances using the old CRD will be converted to the new one.

bryan-aguilar:
Awesome, thanks for the info!

pavolloffay:
Done in other issues/PRs

PR #2283, opened 2023-10-26
[chore] Add v1alpha2 types

iblancasa:
**Description:**: add the types from the new version.
Note no manifests are generated and no reconciliation is run for these types.
No changelog is generated because users will not feel any change.

**Link to tracking Issue:**: #2183


swiatekm-sumo (approved):

yuriolisa (approved):

frzifus (approved):

jaronoff97:
As part of this change, these new CRDs are where we will be modifying the specs for a simpler and enhanced experience, right? i.e. we expect to modify these types before actually using them and generating them?

iblancasa:
> As part of this change, these new CRDs are where we will be modifying the specs for a simpler and enhanced experience, right? i.e. we expect to modify these types before actually using them and generating them?

During a SIG call, I was asked to send a PR like this one. After merging it, we can generate/add the reconciliation logic as part of other PR and protect that logic behind a feature flag. And we can start doing modifications to that code to include all the new features we want. What do you think?

iblancasa:
Can somebody rerun and merge?

Issue #2008, opened 2023-08-10
Create a single interface for the Receiver and Exporter parsers


iblancasa:
Should we not just make a ComponentPortParser interface that both receivers and exporters implement? Is there anything special about this vs receiver?

_Originally posted by @jaronoff97 in https://github.com/open-telemetry/opentelemetry-operator/pull/1983#discussion_r1289177384_

iblancasa:
I'll work on it

PR #2287, opened 2023-10-27
[chore] Create a single interface for the Receiver and Exporter parsers

iblancasa:
**Description:** Create a single interface for `Receiver` and `Exporter` parsers

**Link to tracking Issue:** #2008


swiatekm-sumo (approved):
LGTM.

Can we change `componentsName` from `string` to an enum (as much of an enum as Go lets us make, anyway)? It'd be less fragile if we could write `Receivers` instead of `"receivers"`.

iblancasa:
> LGTM.
> 
> Can we change `componentsName` from `string` to an enum (as much of an enum as Go lets us make, anyway)? It'd be less fragile if we could write `Receivers` instead of `"receivers"`.

Great idea! I'll do. Thanks! :)

frzifus (commented):
LGTM - May I ask:
1. does it make sense to use the singular for the const names?
1. Since it is now an int, does it make sense to introduce an unknown type to avoid unwanted defaults?
1. Does it make sense to give the constant the `ComponentType` prefix to avoid future name conflicts and make it a bit more expressive?

internal/manifests/collector/adapters/config_to_ports.go
```diff
@@ -52,119 +50,71 @@ func ConfigToExporterPorts(logger logr.Logger, config map[interface{}]interface{
...
-			logger.V(2).Info("no parser found for '%s'", exprtName)
-			continue
+		cmptName := key.(string)
+		var cmptParser parser.ComponentPortParser
+		var err error
+		switch cType {
+		case ComponentTypeExporter:
+			cmptParser, err = exporterParser.For(logger, cmptName, exporter)
+		case ComponentTypeReceiver:
+			cmptParser, err = receiverParser.For(logger, cmptName, exporter)
```
jaronoff97, line 102: Once config is no longer a string we should refactor this section as well IMO. I.e. once config has a defined `receivers` and `exporters` type it would be great if we didn't have to do this type of switching and instead put the methods on the structs instead.

jaronoff97 (changes requested):
Just one change that I think would make this a bit cleaner, otherwise 👍 

Overall, I desperately want real types to work with here. `configFromString` is particularly problematic here as it's preventing us from just putting these parsers on a type (receiver, exporter, etc.) which would mean we don't have to do any of the swapping on the desired component type.

jaronoff97:
looks like a linting failure, will merge when the tests are ✅

internal/webhook/podmutation/webhookhandler_test.go
```diff
@@ -123,6 +123,7 @@ func TestShouldInjectSidecar(t *testing.T) {
 			},
 		},
 	} {
+		tt := tt
```
jaronoff97, line 126: why is this necessary?

iblancasa: I didn't detect the previous lint error because of the golang-ci version. I upgraded to the latest and now I'm getting a lot of `Implicit memory aliasing in for loop` errors.
I fixed them because we will have to do that in the future anyway.

jaronoff97: I thought that would only happen if we were using a reference to the test's memory address? Is this a known go issue for the testing library?

swiatekm-sumo: In general, we do need this assignment. It doesn't matter unless we enable `t.Parallel()` for the test, but it's a good practice and probably why linters are complaining about it. Fundamentally, we're using the `tt` variable in a closure via `t.Run`, so if we ran the cases in parallel without reassigning, we'd get the same value in all of them. See https://go.dev/blog/loopvar-preview for a clearer and more detailed explanation.

jaronoff97 (approved):

internal/webhook/podmutation/webhookhandler_test.go
swiatekm-sumo: Once we upgrade the minimum Go version to 1.22 in our go.mod, yes. Just compiling with Go 1.22 isn't enough according to the blog post.

PR #2288, opened 2023-10-28
[operator-opamp-bridge][chore]: Refactor OpAMPBridge Webhook

avadhut123pisal:
**Description:** 
This PR updates the OpAMPBridge Webhook code to make it consistent with the changes that are made in Collector webhook as a part of #2210. 

**Link to tracking Issue:** NA

**Testing:** Tested locally. 

**Documentation:** NA

### DIFF
--- a/apis/v1alpha1/opampbridge_webhook.go
+++ b/apis/v1alpha1/opampbridge_webhook.go
@@ -15,33 +15,69 @@
 package v1alpha1
 
 import (
+	"context"
 	"fmt"
 	"strings"
+
+	"github.com/go-logr/logr"
+
+	"github.com/open-telemetry/opentelemetry-operator/internal/config"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation"
 	ctrl "sigs.k8s.io/controller-runtime"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
-	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
-// log is for logging in this package.
-var opampbridgelog = logf.Log.WithName("opampbridge-resource")
-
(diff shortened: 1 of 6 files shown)


config/webhook/manifests.yaml
```diff
@@ -137,7 +137,7 @@ webhooks:
       namespace: system
       path: /validate-opentelemetry-io-v1alpha1-opampbridge
   failurePolicy: Fail
-  name: vopampbridge.kb.io
+  name: vopampbridgecreateupdate.kb.io
```
jaronoff97, line 140: why did this name change?

avadhut123pisal: https://github.com/avadhut123pisal/opentelemetry-operator/blob/c3088a72c7bd0af9f7d3283875cfdb860cc309f2/apis/v1alpha1/opampbridge_webhook.go#L37C1-L37C1

The change has been done considering the validation webhook configurations of Collector resource. 
One webhook is for create & update operations (with failure policy as Fail) and another one is for delete operation (with failure policy as Ignore). Previously, by default there was a single webhook for all the operations.

apis/v1alpha1/opampbridge_webhook.go
```diff
@@ -15,33 +15,69 @@
...
 
 import (
+	"context"
 	"fmt"
 	"strings"
 
+	"github.com/go-logr/logr"
+
+	"github.com/open-telemetry/opentelemetry-operator/internal/config"
+
```
jaronoff97, line 25: the structure of these imports is wrong, though idk why our linter isn't catching it...

jaronoff97 (approved):

apis/v1alpha1/opampbridge_webhook.go
jaronoff97: im going to leave this in for now, and then we can figure out why the linter isn't catching this in a follow up

avadhut123pisal: > the structure of these imports is wrong, though idk why our linter isn't catching it...

Yeah. Sometimes, it throws error for correct import structure too.

PR #2289, opened 2023-10-28
Build manager on the host

swiatekm-sumo:
**Description:**

Cross-compile the manager binary on the host and copy the appropriate binary into the container during the build. This makes image building *much* faster, as we don't need to rely on emulation via QEMU.

I've also removed our local toolchain's dependency on buildx while I was at it, as it's not strictly necessary.

Looks like we could build the binaries in a matrix job, but given that they take around 15 minutes without cache, I don't think it's worth it to mess around with artifacts.

**Issue**: #2295 

**Testing:**
Temporarily made the build happen for this PR, and it worked fine.

### DIFF
--- a/.github/workflows/publish-images.yaml
+++ b/.github/workflows/publish-images.yaml
@@ -7,12 +7,19 @@ on:
 
   workflow_dispatch:
 
+env:
+  PLATFORMS: linux/amd64,linux/arm64,linux/s390x,linux/ppc64le
+
 jobs:
   publish:
     name: Publish container images
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '~1.21.3'
 
       - name: Unshallow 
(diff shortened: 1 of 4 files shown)


.github/workflows/publish-images.yaml
```diff
@@ -86,19 +101,5 @@ jobs:
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.docker_meta.outputs.tags }}
           labels: ${{ steps.docker_meta.outputs.labels }}
-          build-args: |
```
jaronoff97, line 89: why do we not need these args anymore?

swiatekm-sumo: They're set while building the binary, straight from the environment. They're not needed for anything else.

jaronoff97 (approved):
I think this LGTM, but we should wait on @pavolloffay 's review here too

pavolloffay (approved):

jaronoff97 (approved):

Issue #2280, opened 2023-10-26
Scrape targets can be reassigned even when collector count doesn't change


swiatekm-sumo:
### Component(s)

target allocator

## Description

Running target allocator in reasonably sized clusters, I noticed that targets were being reassigned to different collector Pods even if the Pod count didn't change. We don't actually have metrics to measure this, but I noticed a second-order effect of receiving Prometheus staleness markers, which are emitted when a scrape target legitimately disappears.

This shouldn't really happen, and looking at the allocation strategy code, there doesn't appear to be a way for it to happen, *assuming the hash doesn't change for an existing target*. However, we do calculate the hash based on all the target labels, some of which are mutable. Even putting aside whether this is the cause of the problem I've experienced, there's no reason for us to include all the labels in the hash - the target address should suffice, and this is what prometheus-operator does for sharding as well. 

## Steps to Reproduce

Don't have a clean reproduction yet, but showing the behaviour of targets moving if one of the labels changes is easy in unit tests.

## Expected Result

Scrape targets should never be moved between collectors if the collector count doesn't change.

### Kubernetes Version

1.24, latest EKS

### Operator version

0.82.0

### Collector version

0.82.0

## Environment
OS: (e.g., "Ubuntu 20.04")
Compiler(if manually compiled): (e.g., "go 14.2")

jaronoff97:
Yeah I think this would happen if the labels on a collector are patched, the update could cause the assignment to move based on the hash ring.

jaronoff97:
The reason we included all of the labels is because sometimes a target will appear more than once and if we didn't use the entire label it would cause a single target to be assigned to multiple collectors IIRC. Let me find the PR where I changed that.

jaronoff97:
Yeah here was [that PR](https://github.com/open-telemetry/opentelemetry-operator/pull/1039)

swiatekm-sumo:
Wouldn't it be easier to just deduplicate the targets in that case? Plus, that sounds like a bug in Prometheus Service Discovery, so maybe it was fixed in the meantime?

jaronoff97:
it's a bit different actually, the scrape config's service discovery may generate multiple entries for the same target url but have different labels (think an application that serves metrics on two paths, not two ports). If you don't use the entire label set you would end up having a condition where you sometimes scrape the first path and sometimes scrape the second because of a race in discovery. That PR resolved it by treating it the way prometheus does (as two targets) 

[...]

PR #2290, opened 2023-10-30
Use only target address for allocation in consistent-hashing strategy

swiatekm-sumo:
**Description:**
For the consistent-hashing strategy, base target allocation decisions only on the target url. This way, for a given collector count, we always assign targets with the same url to the same collector. Before, this decision was based on all target labels, and targets could therefore be moved between collectors even if collector count stayed the same.

**Link to tracking Issue:** #2280 

**Testing:**

Tested this in a ~500 Node production cluster in addition to the unit tests. It behaved as expected and over time resulted in an average drop of around 10% in CPU and memory usage with around 4000 scrape targets.

### DIFF
--- a/.chloggen/ta_really-consistent-hashing.yaml
+++ b/.chloggen/ta_really-consistent-hashing.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: target allocator
+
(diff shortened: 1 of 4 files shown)


jaronoff97:
What did the balance of targets per collector look like after you tested this?

swiatekm-sumo:
> What did the balance of targets per collector look like after you tested this?

The relative standard deviation of it stayed roughly the same, which for this particular environment with consistent hashing is around 0.2.

jaronoff97 (approved):

jaronoff97:
@swiatekm-sumo looks like a unit test failure?

swiatekm-sumo:
> @swiatekm-sumo looks like a unit test failure?

Fixed, needed to make all the test data generators assign different target urls to their generated targets.

PR #2317, opened 2023-11-07
Use a single module for all binaries

swiatekm-sumo:
Use a single Go module for all binaries. This simplifies our project structure and tooling. Benefits include:

* It's now possible to work on any of our components without needing to set up Go workspaces, improving the new contributor experience
* Testing and linting, go mod tidy, and so on, can now be be done in a single command
* 3x less dependabot PRs
* Caching in the CI is now simpler and more robust

The only downside is that dependencies cannot be different between the components, but even that's partially a benefit.


jaronoff97 (approved):

jaronoff97:
this LGTM. were you able to build the docker containers locally and they worked totally fine? I don't think there would be any problems, but just for extra safety.

swiatekm-sumo:
> this LGTM. were you able to build the docker containers locally and they worked totally fine? I don't think there would be any problems, but just for extra safety.

I have. And e2e tests do this anyway, so we would've seen some breakage there.

Issue #2309, opened 2023-11-04
Ignore invalid CRDs and pick up valid CRDs with TargetAllocator


rashmichandrashekar:
### Component(s)

target allocator

## Description
When a podmonitor/svcmonitor CRD has an erraneous field for regex in the relabeling/metric relabeling fields, the crd creation succeeds since it can be a valid string but not a valid regex( ex- .*( ). This causes any subsequent valid CRDs being created from getting picked up.

## Steps to Reproduce
Create a podmonitor/svc monitor CRD with invalid regex (ex - .*()
## Expected Result
[This](https://github.com/open-telemetry/opentelemetry-operator/blob/4b310f5cc94d73d5b29e598a5e2217a7ef62f36f/cmd/otel-allocator/watcher/promOperator.go#L241) only generates the config without actually validating the pod/service monitors selected. It should follow a similar approach to prometheus-operator's [resource selector](https://github.com/prometheus-operator/prometheus-operator/blob/45185c8d5c0e97027f7e99beeb9114c5642aeec3/pkg/prometheus/resource_selector.go#L67) where all the monitors are looped through and validated and the invalid ones are rejected.
## Actual Result
All CRDs fail to be applied when one crd is error prone

### Kubernetes Version

1.25.11

### Operator version

v0.85.0

### Collector version

v0.85.0

### Log output

{"level":"error","ts":"2023-11-04T20:33:52Z","logger":"setup","msg":"Unable to load configuration","error":"error parsing regexp: missing closing ): `^(?:.*()$`","stacktrace":"main.main.func13\n\t/app/main.go:228\ngithub.com/oklog/run.(*Group).Run.func1\n\t/go/pkg/mod/github.com/oklog/run@v1.1.0/group.go:38"}

jaronoff97:
closed by #2328 thank you again @rashmichandrashekar !!

PR #2328, opened 2023-11-09
Add validation for pod/service monitors for TargetAllocator and skip invalid ones

rashmichandrashekar:
**Description:** 
Fixing a bug - When a podmonitor/svcmonitor CRD (with Targetallocator enabled)has an erroneous field for regex in the relabeling/metric relabeling fields, the crd creation succeeds since it can be a valid string but not a valid regex( ex- .*( ). This causes any subsequent valid CRDs being created from getting picked up.

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-operator/issues/2309

**Testing:** Tested this out with otel operator setup

### DIFF
--- a/cmd/otel-allocator/watcher/promOperator.go
+++ b/cmd/otel-allocator/watcher/promOperator.go
@@ -17,9 +17,12 @@ package watcher
 import (
 	"context"
 	"fmt"
+	"os"
+	"regexp"
 	"time"
 
 	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
 	"github.com/go-logr/logr"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	promv1alpha1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1alpha1"
@@ -27,8 +30,10 @@ import (
 	monitoringclient "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned"
 	"github.com/prometheus-operator/prometheus-operator/pkg/informers"
 	"github.com/prometheus-operator/prometheus-operator/pkg/prometheus"
+	"github.com/prometheus/common/model"
 	promconfig "github.com/prometheus/prometheus/config"
 	kubeDiscovery "github.com/prometheus/prometheus/discovery/kubernetes"
+	"github.com/prometheus/prometheus/model/relabel"
 	"gopkg.in/yaml.v2"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -220,8 +225,12 @@ func (w *PrometheusCRWatcher) LoadConfig(ctx context.Context) (*promconfig.Confi
 	smRetrieveErr := w.informers[monitoringv1.ServiceMonitorName].ListAll(w.serviceMonitorSelector, func(sm interface{}) {
 		monitor := sm.(*monitoringv1.ServiceMonitor)
 		key, _ := cache.DeletionHandlingMetaNamespaceKeyFunc(monitor)
-		w.addStoreAssetsForServiceMonitor(ctx, monitor.Name, monitor.Namespace, monitor.Spec.Endpoints, store)
-		serviceMonitorInstances[key] = monitor
+		validateError := w.addStoreAssetsForServiceMonitor(ctx, monitor.Name, monitor.Namespace, monitor.Spec.Endpoints, store)
+		if validateError != nil {
+			w.logger.Error(validateError, "Failed validating ServiceMonitor, skipping", "ServiceMonitor:", monitor.Name, "in namespace", monitor.Namespace)
+		} else {
+			serviceMonitorInstances[key] = monitor
+		}
 	})
 	if smRetrieveErr != nil {
 		return nil, smRetrieveErr
@@ -231,8 +240,12 @@ func (w *PrometheusCRWatcher) LoadConfig(ctx context.Context) (*promconfig.Confi
 	pmRetrieveErr := w.informers[monitoringv1.PodMonitorName].ListAll(w.podMonitorSelector, func(pm interface{}) {
 		monitor := pm.(*monitoringv1.PodMonitor)
 		key, _ := cache.DeletionHandlingMetaNamespaceKeyFunc(monitor)
-		w.addStoreAssetsForPodMonitor(ctx, monitor.Name, monitor.Namespace, monitor.Spec.PodMetricsEndpoints, store)
-		podMonitorInstances[key] = monitor
+		validateError := w.addStoreAssetsForPodMonitor(ctx, monitor.Name, monitor.Namespace, monitor.Spec.PodMetricsEndpoints, store)
+		if validateError != nil {
+			w.logger.Error(validateError, "Failed validating PodMonitor, skipping", "PodMonitor:", monitor.Name, "in namespace", monitor.Namespace)
+		} else {
+			podMonitorInstances[key] = monitor
+		}
 	})
 	if pmRetrieveErr != nil {
 		return nil, pmRetrieveErr
@@ -288,8 +301,9 @@ func (w *PrometheusCRWatcher) addStoreAssetsForServiceMonitor(
 	smName, smNamespace string,
 	endps []monitoringv1.Endpoint,
 	store *assets.Store,
-) {
+) error {
 	var err error
+	var validateErr error
 	for i, endp := range endps {
 		objKey := fmt.Sprintf("serviceMonitor/%s/%s/%d", smNamespace, smName, i)
 
@@ -315,11 +329,33 @@ func (w *PrometheusCRWatcher) addStoreAssetsForServiceMonitor(
 		if err = store.AddSafeAuthorizationCredentials(ctx, smNamespace, endp.Authorization, smAuthKey); err != nil {
 			break
 		}
+
+		for _, rl := range endp.RelabelConfigs {
+			if rl.Action != "" {
+				if validateErr = validateRelabelConfig(*rl); validateErr != nil {
+					break
+				}
+			}
+		}
+
+		for _, rl := range endp.MetricRelabelConfigs {
+			if rl.Action != "" {
+				if validateErr = validateRelabelConfig(*rl); validateErr != nil {
+					break
+				}
+			}
+		}
 	}
 
 	if err != nil {
 		w.logger.Error(err, "Failed to obtain credentials for a ServiceMonitor", "serviceMonitor", smName)
 	}
+
+	if validateErr != nil {
+		return validateErr
+	}
+
+	return nil
 }
 
 // addStoreAssetsForServiceMonitor adds authentication / authorization related information to the assets store,
@@ -331,7 +367,7 @@ func (w *PrometheusCRWatcher) addStoreAssetsForPodMonitor(
 	pmName, pmNamespace string,
 	podMetricsEndps []monitoringv1.PodMetricsEndpoint,
 	store *assets.Store,
-) {
+) error {
 	var err error
 	for i, endp := range podMetricsEndps {
 		objKey := fmt.Sprintf("podMonitor/%s/%s/%d", pmNamespace, pmName, i)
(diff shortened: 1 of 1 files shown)


cmd/otel-allocator/watcher/promOperator.go
```diff
@@ -358,9 +392,96 @@ func (w *PrometheusCRWatcher) addStoreAssetsForPodMonitor(
...
+	}
+
+	return nil
+}
+
+// validateRelabelConfig validates relabel config for service and pod monitor,
+// based on the service monitor and pod metrics endpoints specs.
+// This code borrows from
+// https://github.com/prometheus-operator/prometheus-operator/blob/ba536405154d18f3a6f312818283d671182af6f3/pkg/prometheus/resource_selector.go#L237
+func validateRelabelConfig(rc monitoringv1.RelabelConfig) error {
```
jaronoff97, line 428: Would it be useful for us to reach out to the prometheus operator folks to see if they would expose this method?

jaronoff97 (commented):
Could you add some tests in to the promOperator_test.go file to ensure this works how we expect?

cmd/otel-allocator/watcher/promOperator.go
rashmichandrashekar: Thanks @jaronoff97. I have reached out to them, waiting for a response,(https://github.com/prometheus-operator/prometheus-operator/pull/6057#issuecomment-1804829543)  based on that I can add the tests.

rashmichandrashekar: Also created an issue here - https://github.com/prometheus-operator/prometheus-operator/issues/6076

rashmichandrashekar: @jaronoff97 - There is a question about breaking changes. I have answered it from what I understand and what I think the way forward would be. Could you please respond to provide your perspective and correct me if i am wrong?
https://github.com/prometheus-operator/prometheus-operator/issues/6076#issuecomment-1808719873

jaronoff97: yep, i responded there :) thanks for making the issue. Let's wait to hear from them before merging this – i would love to just use their method rather than having to write it ourselves.

jaronoff97:
@rashmichandrashekar could you re-implement this PR to use the interfaces recommended by the prometheus operator?

rashmichandrashekar:
> @rashmichandrashekar could you re-implement this PR to use the interfaces recommended by the prometheus operator?

@jaronoff97 - For sure, I have the changes all ready. Just cleaning some things up and I will create a PR today. Thanks for helping with this!

rashmichandrashekar:
> > @rashmichandrashekar could you re-implement this PR to use the interfaces recommended by the prometheus operator?
> 
> @jaronoff97 - For sure, I have the changes all ready. Just cleaning some things up and I will create a PR today. Thanks for helping with this!

@jaronoff97 - I have reimplemented this and added relevant tests. Please take a look and lmk if you have any feedback. Also a couple tests seem to be timing out, please requeue them if possible. Thanks!

jaronoff97:
Thank you! I'll try to take look today or monday.

cmd/otel-allocator/watcher/promOperator.go
```diff
@@ -28,21 +28,29 @@ import (
...
-	"k8s.io/apimachinery/pkg/labels"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 
 	allocatorconfig "github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/config"
 )
 
 const minEventInterval = time.Second * 5
+const (
```
jaronoff97, line 49: can you move the above constant in with this block?

jaronoff97 (commented):
A few nits/ questions. Does this change how our existing label selector works for the target allocator? I'm wondering if we should also write a new e2e or two to verify this works as expected. cc @swiatekm-sumo

cmd/otel-allocator/watcher/promOperator.go
```diff
@@ -139,6 +194,49 @@ func (w *PrometheusCRWatcher) Watch(upstreamEvents chan Event, upstreamErrors ch
...
+			if old.ResourceVersion == cur.ResourceVersion {
+				return
+			}
+
+			for name, selector := range map[string]*metav1.LabelSelector{
+				"PodMonitorNamespaceSelector":     w.podMonitorNamespaceSelector,
+				"ServiceMonitorNamespaceSelector": w.serviceMonitorNamespaceSelector,
+			} {
+
+				sync, err := k8sutil.LabelSelectionHasChanged(old.Labels, cur.Labels, selector)
```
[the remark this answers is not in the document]
rashmichandrashekar: This event handler watches for any namespace changes and then calls this method - [LabelSelectionHasChanged](https://github.com/prometheus-operator/prometheus-operator/blob/28c536d33998e9c74e3e280db0b9c810c83ca8fc/pkg/k8sutil/labels.go#L27) to basically make sure that the TA's selector yields the same results with old and new labels, if it does then an event is not sent and the new labels lead to a change in behavior then an event is sent.
For ex- If the cfg.podMonitorNamespaceSelector was set to match namespaces with labels "testns" : "testns" and the initial set of namespace labels on for example - namespace1 were "testns" : "testns" and later if the labels were removed, then an event would be sent. But if later, the labels on the namespace were updated to 
{
"testns" : "testns",
"testns1": "testns1",
}
it wouldnt send an event.

rashmichandrashekar:
> A few nits/ questions. Does this change how our existing label selector works for the target allocator? I'm wondering if we should also write a new e2e or two to verify this works as expected. cc @swiatekm-sumo

Sure @jaronoff97. This doesnt change the existing behavior of the selectors. This just adds a couple other selectors just like in prometheus-operator. I did add unit tests for all these selectors to make sure that the selectors result in the right configuration and select the right pod and svc monitors as well as selectors on namespaces result in picking up pod and svc monitors in the matching namespaces. please lmk if e2e are also necessary.

cmd/otel-allocator/config/config.go
```diff
@@ -39,18 +39,20 @@ const DefaultConfigFilePath string = "/conf/targetallocator.yaml"
...
+	ReloadConfig                    bool               `yaml:"-"`
+	LabelSelector                   map[string]string  `yaml:"label_selector,omitempty"`
+	PromConfig                      *promconfig.Config `yaml:"config"`
+	AllocationStrategy              *string            `yaml:"allocation_strategy,omitempty"`
+	FilterStrategy                  *string            `yaml:"filter_strategy,omitempty"`
+	PrometheusCR                    PrometheusCRConfig `yaml:"prometheus_cr,omitempty"`
+	PodMonitorSelector              map[string]string  `yaml:"pod_monitor_selector,omitempty"`
+	ServiceMonitorSelector          map[string]string  `yaml:"service_monitor_selector,omitempty"`
+	ServiceMonitorNamespaceSelector map[string]string  `yaml:"service_monitor_namespace_selector,omitempty"`
+	PodMonitorNamespaceSelector     map[string]string  `yaml:"pod_monitor_namespace_selector,omitempty"`
```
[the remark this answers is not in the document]
rashmichandrashekar: Thanks @jaronoff97. Since the common [prometheus fields](https://github.com/open-telemetry/opentelemetry-operator/pull/2328/files#:~:text=%7D%2C-,ServiceMonitorNamespaceSelector%3A%20%26metav1.LabelSelector%7B,%7D%2C,-%7D%2C) have the ServiceMonitorNamespaceSelector and PodMonitorNamespaceSelector already initialized, even if the targetallocator config is not setting these selectors (or have nil selectors), the default behavior will be to match all namespaces which is the current behavior. But I can add defaults, in case you think otherwise.  Please let me know.

cmd/otel-allocator/watcher/promOperator.go
```diff
@@ -76,43 +96,78 @@ func NewPrometheusCRWatcher(logger logr.Logger, cfg allocatorconfig.Config) (*Pr
...
-	servMonSelector := getSelector(cfg.ServiceMonitorSelector)
+	store := assets.NewStore(clientset.CoreV1(), clientset.CoreV1())
+	promRegisterer := prometheusgoclient.NewRegistry()
+	operatorMetrics := operator.NewMetrics(promRegisterer)
 
-	podMonSelector := getSelector(cfg.PodMonitorSelector)
+	nsMonInf, err := getNamespaceInformer(ctx, map[string]struct{}{v1.NamespaceAll: {}}, promOperatorLogger, clientset, operatorMetrics)
+	if err != nil {
+		return nil, err
+	}
```
swiatekm-sumo, line 106: It looks like we error if we don't have a permission to view namespaces. Could we instead log a warning here? Otherwise this will be a breaking change that will require all Target Allocator users to update their roles. I'd rather avoid that. This is how the allocator has behaved thus far as well - if it didn't have permission to view some resource, it'd log a warning and continue.

swiatekm-sumo (changes requested):
Thanks for submitting this change, I really like the watcher cleanup. I do have a concern about this being a breaking change in its current state though - see my comment on the namespace informer.

cmd/otel-allocator/watcher/promOperator.go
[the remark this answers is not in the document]
jaronoff97: got it, thanks for the great explanation! :bow:

cmd/otel-allocator/watcher/promOperator.go
jaronoff97: in said note we should be explicit on the changes the user needs to make i.e. role -> clusterrole & add namespace.

rashmichandrashekar: Thanks @swiatekm-sumo and @jaronoff97.
I did update the PR to handle the case where there are no namespace list/watch permissions. Please let me know if this looks good. With this change, the TA will continue to run but will not load any config since the resourceselector doesnt get initialized.

rashmichandrashekar: Following up. @swiatekm-sumo @jaronoff97 - Pls let me know if you would like any other changes. Thanks!

jaronoff97:
@rashmichandrashekar we just talked about this at the operator SIG and want to attempt to get this in prior to our next release. In order to do so, I opened up #2426 to see if we can ensure that we are not going to break users by checking they have the new permission for the TA to function. We're going to pause merging this until we have an answer there.

cmd/otel-allocator/watcher/promOperator.go
rashmichandrashekar: @swiatekm-sumo - Sorry I didnt see this in time as I am in PST. But I agree that we should ensure we don't break customers with this change.

rashmichandrashekar:
> @rashmichandrashekar we just talked about this at the operator SIG and want to attempt to get this in prior to our next release. In order to do so, I opened up #2426 to see if we can ensure that we are not going to break users by checking they have the new permission for the TA to function. We're going to pause merging this until we have an answer there.



> @rashmichandrashekar we just talked about this at the operator SIG and want to attempt to get this in prior to our next release. In order to do so, I opened up #2426 to see if we can ensure that we are not going to break users by checking they have the new permission for the TA to function. We're going to pause merging this until we have an answer there.

Thanks @jaronoff97. That sounds good. I think this will be useful for any permission related breaking changes.

jaronoff97:
nope, this is a different SIG :) you can find in slack at #otel-operator and you can see the other sigs [here](https://github.com/open-telemetry/community/?tab=readme-ov-file#implementation-sigs).

rashmichandrashekar:
[like]  Rashmi Chandrashekar reacted to your message:
________________________________

jaronoff97:
@rashmichandrashekar i've merged in my PR so now you should be able to fix the conflicts and update the necessary TA permissions :)

rashmichandrashekar:
> @rashmichandrashekar i've merged in my PR so now you should be able to fix the conflicts and update the necessary TA permissions :)

Thanks so much @jaronoff97 for unblocking this PR. I will get the latest and test it out and update this PR.

rashmichandrashekar:
> > @rashmichandrashekar i've merged in my PR so now you should be able to fix the conflicts and update the necessary TA permissions :)
> 
> Thanks so much @jaronoff97 for unblocking this PR. I will get the latest and test it out and update this PR.

@jaronoff97 - I have updated the PR by fixing conflicts and adding permissions check for namespaces in the webhook. Please take a look. Thanks!

.chloggen/target-allocator-use-recommended-prometheus-operator-interfaces.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
swiatekm-sumo, line 2: This should be a breaking change.

.chloggen/target-allocator-use-recommended-prometheus-operator-interfaces.yaml
```diff
@@ -0,0 +1,16 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Use recommended interfaces(resource selector) by the prometheus-operator for watching CRs.
+
+# One or more tracking issues related to the change
+issues: [2309]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
swiatekm-sumo, line 16: We need to call out that the target allocator now requires get/list/watch permissions for namespaces.

```suggestion
subtext: The target allocator now requires get/list/watch permissions for namespaces. Update your RBAC permissions for the attached role, if necessary.
```

cmd/otel-allocator/README.md
```diff
@@ -139,6 +139,10 @@ rules:
   - podmonitors
   verbs:
   - '*'
+- apiGroups: [""]
+  resources:
+  - namespaces
+  verbs: ["list", "watch"]
```
swiatekm-sumo, line 145: Don't we also need a `get` here?

cmd/otel-allocator/watcher/promOperator.go
```diff
@@ -139,6 +196,48 @@ func (w *PrometheusCRWatcher) Watch(upstreamEvents chan Event, upstreamErrors ch
...
 	notifyEvents := make(chan struct{}, 1)
 
+	if w.nsInformer != nil {
+		go w.nsInformer.Run(w.stopChannel)
+		if ok := cache.WaitForNamedCacheSync("namespace", w.stopChannel, w.nsInformer.HasSynced); !ok {
+			success = false
+		}
+
+		_, _ = w.nsInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+			UpdateFunc: func(oldObj, newObj interface{}) {
```
swiatekm-sumo, line 206: I don't see a test for this logic. Can we add one?

swiatekm-sumo (commented):
This looks good to me overall, modulo some relatively minor problems.

cmd/otel-allocator/watcher/promOperator.go
```diff
@@ -76,43 +97,81 @@ func NewPrometheusCRWatcher(logger logr.Logger, cfg allocatorconfig.Config) (*Pr
 		return nil, err
 	}
 
-	servMonSelector := getSelector(cfg.ServiceMonitorSelector)
+	store := assets.NewStore(clientset.CoreV1(), clientset.CoreV1())
+	promRegisterer := prometheusgoclient.NewRegistry()
+	operatorMetrics := operator.NewMetrics(promRegisterer)
+	// eventRecorderFactory := operator.NewEventRecorderFactory(false)
```
swiatekm-sumo, line 103: This should be deleted.

cmd/otel-allocator/watcher/promOperator_test.go
```diff
@@ -724,6 +724,68 @@ func TestLoadConfig(t *testing.T) {
 				},
 			},
 		},
+		// {
```
swiatekm-sumo, line 727: Is this supposed to be commented out?

rashmichandrashekar: @swiatekm-sumo - Sorry, this is still WIP. I accidentally pushed it :)

cmd/otel-allocator/watcher/promOperator.go
rashmichandrashekar: @swiatekm-sumo - Added this too. Please take a look.

cmd/otel-allocator/watcher/promOperator_test.go
rashmichandrashekar: I have addressed the comments now, its ready to review.

cmd/otel-allocator/config/config.go
```diff
@@ -44,17 +44,19 @@ const (
...
+	RootLogger                      logr.Logger           `yaml:"-"`
+	CollectorSelector               *metav1.LabelSelector `yaml:"collector_selector,omitempty"`
+	PromConfig                      *promconfig.Config    `yaml:"config"`
+	AllocationStrategy              string                `yaml:"allocation_strategy,omitempty"`
+	FilterStrategy                  string                `yaml:"filter_strategy,omitempty"`
+	PrometheusCR                    PrometheusCRConfig    `yaml:"prometheus_cr,omitempty"`
+	PodMonitorSelector              map[string]string     `yaml:"pod_monitor_selector,omitempty"`
+	ServiceMonitorSelector          map[string]string     `yaml:"service_monitor_selector,omitempty"`
+	ServiceMonitorNamespaceSelector map[string]string     `yaml:"service_monitor_namespace_selector,omitempty"`
+	PodMonitorNamespaceSelector     map[string]string     `yaml:"pod_monitor_namespace_selector,omitempty"`
```
swiatekm-sumo, line 59: Could we make the new fields into a `metav1.LabelSelector`, just like `CollectorSelector`? We want every selector to use that format eventually.

swiatekm-sumo (commented):
LGTM, but I'd like to use LabelSelectors directly in configuration instead of repeating the mistake we made with existing ones. @jaronoff97 can you also have a final look?

cmd/otel-allocator/config/config.go
rashmichandrashekar: @swiatekm-sumo - Updated. Please take a look.

swiatekm-sumo (approved):
:+1: thanks a lot for doing this, and your patience in accepting our feedback.

jaronoff97 (approved):
absolutely fantastic work here. Thank you so much for your patience and contribution. :tada: the added tests are also great to see.

Issue #2350, opened 2023-11-14
Target Allocator loses targets discovered when collector instances take time to come up


rashmichandrashekar:
### Component(s)

target allocator

## Description
In a kubernetes cluster, have scrape jobs configured using TargetAllocator. Due to timing issue(which is very common with kubernetes clusters, where no ordering can be guaranteed), the instances are not discovered until after the first set of targets are discovered. The configuration applied until the instances are discovered are lost.

## Steps to Reproduce
Delay the collector instance creation (hard to reproduce since it happens due to timing issue)

## Expected Result
The targets discovered should be retained so that they can be assigned whenever the collector instances come up

## Actual Result
The targets discovered before the collector instances come up are lost.

### Kubernetes Version

1.25.11

### Operator version

v0.85.0

### Collector version

v0.85.0

## Environment
OS: (e.g., "Ubuntu 20.04")
Compiler(if manually compiled): (e.g., "go 14.2")

### Log output

```shell
No collector instances present, cannot set targets
```

### Additional context

Have created a PR to fix this - https://github.com/open-telemetry/opentelemetry-operator/pull/2351

jaronoff97:
closed by #2351

PR #2351, opened 2023-11-14
Fix for Target Allocator not saving targets when collector instances take time to come up

rashmichandrashekar:
**Description:** 
Fixing a bug - In a kubernetes cluster, have scrape jobs configured using TargetAllocator. Due to timing issue(which is very common with kubernetes clusters, where no ordering can be guaranteed), the instances are not discovered until after the first set of targets are discovered. The configuration applied until the instances are discovered are lost.
This PR fixes that issue where it saves the targets until the collector instances come up

https://github.com/open-telemetry/opentelemetry-operator/issues/2350

**Testing:** 
Tested that in such scenarios, the targets are not lost and assigned to collector instances when they come up


swiatekm-sumo:
Can you add a unit test for this case?

Also, does the same problem not apply to the other allocation strategy as well?

rashmichandrashekar:
> Can you add a unit test for this case?
> 
> Also, does the same problem not apply to the other allocation strategy as well?

Thanks @swiatekm-sumo. I have made the changed for the other hashing algo as well and added a couple unit tests. please lmk if anything else is needed.

swiatekm-sumo:
Can you also check the case where you add targets to a strategy with 0 collectors, and later add collectors to see if targets get allocated to them correctly?

There's also a lint error you need to fix.

rashmichandrashekar:
> Can you also check the case where you add targets to a strategy with 0 collectors, and later add collectors to see if targets get allocated to them correctly?
> 
> There's also a lint error you need to fix.

Thanks @swiatekm-sumo . That's a good test case, I have added that. Please take a look.

rashmichandrashekar:
@swiatekm-sumo - I see an e2e test failure due to timeouts during [CR creation](https://github.com/open-telemetry/opentelemetry-operator/actions/runs/6894688215/job/18757056049#step:6:359). Could you please rerun the tests? Thanks!

rashmichandrashekar:
@swiatekm-sumo - All the checks seem to have passed, could you pls help get this merged? Thanks!

swiatekm-sumo (approved):
I feel like there should be a way to do this without all the special cases, but we can refactor it later now that we have tests for them. I don't think there's any point in keeping this PR held up on DRY issues, so I'm approving it.

rashmichandrashekar:
Thanks @swiatekm-sumo! I dont have permissions to merge, could you pls merge this?
or maybe @jaronoff97 can help merge this?

jaronoff97 (approved):

PR #2352, opened 2023-11-14
Run the target allocator as non-root

dgl:
**Description:**

Add a `USER` directive to the Dockerfile so target allocator runs as non-root. This satisfies the [restricted](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted) pod security standard.

**Link to tracking Issue:**

Not sure one is needed, similar was done in #738.

**Testing:**

Built and ran the target allocator image.

**Documentation:**

n/a


cmd/operator-opamp-bridge/Dockerfile
```diff
@@ -16,4 +16,7 @@ COPY --from=certificates /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-ce
 # Copy binary built on the host
 COPY bin/opampbridge_${TARGETARCH} ./main
 
+# "nonroot"
+USER 65532:65532
```
dgl, line 20: Note I used 65532 as it's the distroless nonroot user, which the main Dockerfile uses.

swiatekm-sumo (approved):
LGTM. The changelog tool will complain unless you put an issue number in the changelog entry yaml, you can just use the issue you linked.

Issue #2032, opened 2023-08-17
Remove configuration hot reloading from the target allocator


swiatekm-sumo:
Target Allocator currently reloads configuration from the config file when it changes. This is an anti-pattern in Kubernetes - Pods should be immutable, and configuration changes should be explicitly rolled out by recreating Pods. We already restart the Target Allocator Deployment when its configuration changes since #1882. As such, the hot reload just makes the semantics more confusing without much benefit.

The plan for removing this would be:
* Add a feature flag to the target allocator to enable this, disabled by default (breaking change)
* Two versions later, remove the feature flag and the rest of the hot reload code

PR #2354, opened 2023-11-14
Disable configuration hot reload in target allocator

swiatekm-sumo:
**Description:** 
Added a flag to control configuration hot reloading in target allocator. The flag is disabled by default, and the functionality is hence considered deprecated and slated for removal in an upcoming release.

I could've done an official feature flag, but the target allocator doesn't have support for feature flags, and I don't think it's necessary to add it just for this.

**Link to tracking Issue:** #2032

**Testing:**
Added a unit test for the flag.


frzifus (commented):
LGTM, I dont think thats something that would belong behind a featuregate.

jaronoff97 (approved):

PR #2355, opened 2023-11-15
implement livenessProbe endpoint to target allocator

changexd:
**Description:** <Describe what has changed.>

Add livenessProbe endpoint for target allocator, another PR will be created for operator following this 

**Link to tracking Issue:** <Issue number if applicable>
#2258

**Testing:** <Describe what testing was performed and which tests were added.>
I wasn't sure if this should be added to the tests, but I still added a simple test just in case.

Issue #2258, opened 2023-10-23
Add readinessProbe and livenessProbe support for target allocator 


luckyxiaoqiang:
### Component(s)

target allocator

### Is your feature request related to a problem? Please describe.

To acheive HA, we needs readinessProbe and livenessProbe configuration. 

### Describe the solution you'd like

Add readinessProbe and livenessProbe properties to `spec.targetAllocator`, and pass it to target allocator `Deployment`.

changexd:
@jaronoff97 Hi I'm interested in this issue, I would like to help this out if this is still available, after some testing and a bit of research, I have couple of question for specification I would like to know before making changes that may not suit the need.

Target Allocator:
1. Should liveness and readiness be checked separately or simply a HTTPGET to "/health" endpoint for both, what should be returned when health checked(currently I'm only writing 200 to the header)
2. Is there any specific condition should be met before checking or simply an endpoint checking the server is running or not?

Operator:
1. Is additional field needed for target allocator spec for customizing probe parameters? if so, what parameters should be inside the field, if not, is it okay to simply add a default probe parameters to target allocator deployment when generating.

If there's already an idea/direction how this should be implemented, please feel free to let me know, thank you : )

jaronoff97:
@changexd for your questions:

* liveness and readiness should be separate IMO
  * liveness can be a very simple check that an endpoint just returns a 200
  * readiness can be defined as after the TA has finished initializing its configuration IMO (cc @swiatekm-sumo who may have more thoughts here)
  * this [blog post](https://developers.redhat.com/blog/2020/11/10/you-probably-need-liveness-and-readiness-probes) will have some more info 
* Hm... I think for the probe we can set a default for now (this should be done in a follow up PR to implementing the endpoints)
  * I want to avoid adding these in to be customizable for now solely because we are in the midst of working on our v2 types for these resources and this is going to add more fields to the target allocator, which I think should be avoided for now.

changexd:
Sorry, I think I mistakenly created separate PRs for readiness and liveness, should I create another PR for readiness or just edit these PRs?

swiatekm-sumo:
As a rule, smaller PRs are better, so I'm fine with what you've done.

rashmichandrashekar:
> Agreed, we should add the probes, but only make them configurable in the separate TargetAllocator CRD once that lands.
> 
[...]

PR #2356, opened 2023-11-15
Add livenessProbe to target allocator deployment for operator

changexd:
**Description:** 
This follows #2355, this PR adds livenessProbe when creating TA deployment

#2355 #2258

I've also added additional fields to both unit tests and e2e tests just in case.


swiatekm-sumo:
If the liveness probe is not implemented in this PR, then why do e2e tests pass? Do they pass before they liveness probe can restart the container?

changexd:
That's likely the case, because after 3 consecutive failures and restarts, the pod does go into crashloop state , I'll look into this to see what can be done, sorry for not noticing this.

swiatekm-sumo:
I don't think this is that big of a problem, but if we want to merge this independently of #2355, then it would be nice to assert that the target allocator is failing liveness probes somewhere in e2e tests.

changexd:
I've tried different ways to test this, as I wasn't able to come up with a better solution than just checking the container restart count after a period, since the liveness probe checks 3 time * 10 period seconds by default before restarting container, I added a test step sleeps for 35 seconds and check if the status of the pod matches expected 0 restarts. 

Additionally I also added a brief timeout to prevent waiting for another default 150 seconds after a 35 second sleep time if the test fails.

If there's a better approach, please feel free to let me know :)

swiatekm-sumo (approved):
LGTM

We should only merge after #2355 lands.

pavolloffay:
e2e tests failed, I am restarting it

PR #2368, opened 2023-11-17
implement readinessProbe endpoint to target allocator

changexd:
#2258

**Testing:** 
I've added readiness tests to see if it works when `Server.scrapeConfigResponse` is nil, I didn't call `Server.UpdateScrapeConfigResponse` when the scrape config is nil during the test because I found out even the config is nil, the eventual output after marshalling is not nil, as I wasn't sure if this is the expected behavior so I skipped this make sure it works as expected.

**Side Notes**
The readiness probe will fail if both scrape config in ta configs and service monitors are not found, I wasn't sure if this is the expected behavior either.

### DIFF
--- a/cmd/otel-allocator/server/server.go
+++ b/cmd/otel-allocator/server/server.go
@@ -89,6 +89,7 @@ func NewServer(log logr.Logger, allocator allocation.Allocator, listenAddr strin
 	router.GET("/jobs/:job_id/targets", s.TargetsHandler)
 	router.GET("/metrics", gin.WrapH(promhttp.Handler()))
 	router.GET("/livez", s.LivenessProbeHandler)
+	router.GET("/readyz", s.ReadinessProbeHandler)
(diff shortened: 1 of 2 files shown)


swiatekm-sumo (approved):
LGTM

jaronoff97 (approved):
I am almost positive the answer is yes, but what should happen who isn't doing servicemonitor and podmonitor discovery is that the scrape config response is whatever the TA has in its config file right?

changexd:
Yes

swiatekm-sumo:
If you have both a static config and Prometheus CR enabled, you could get a scrape config here before Prometheus CRs are actually processed, but I think that's fine. We can think about introducing enough synchronicity to ensure all the providers are ready before succeeding the readiness check, but I'm not sure if the benefit is worth the added complexity. At the very least, having a readiness probe at all is a massive improvement over the status quo.

PR #2369, opened 2023-11-17
Added matchLabels for PodMonitorSelector and ServiceMonitorSelector

### LINKED ISSUE #1907: Unable to configure servicemonitorselector under prometheusCR under targetallocator section

yuriolisa:
Added matchLabels for PodMonitorSelector and ServiceMonitorSelector

Resolves #1907 
smoke-e2e-test added with the new format for matchLabels.

### DIFF
--- a/.chloggen/add-matchLabels-service-pod-monitors.yaml
+++ b/.chloggen/add-matchLabels-service-pod-monitors.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: target allocator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Changed template of podmonitorSelector and serviceMonitorSelector to the metav1.LabelSelector.
+
+# One or more tracking issues related to the change
+issues: [1907]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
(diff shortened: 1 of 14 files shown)


swiatekm-sumo (commented):
This is a breaking change, if we want to do it this way, we should do it in v1alpha2 and provide a migration.

yuriolisa:
> This is a breaking change, if we want to do it this way, we should do it in v1alpha2 and provide a migration.

Ok, so let me switch to `v1alpha2` and provide the migration as well.

yuriolisa:
I'll close this PR and open a fresh one with the proper updates to v1alpha2.

Issue #2372, opened 2023-11-17
Service account is created even when one is specified in collector spec


bryan-aguilar:
### Component(s)

operator

## Description

According to serviceAccount API docs `ServiceAccount indicates the name of an existing service account to use with this instance. When set, the operator will not automatically create a ServiceAccount for the collector.`

## Steps to Reproduce
```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: test-sa
  namespace: collector-deployment
  labels:
    app.kubernetes.io/name: test-sa
---
apiVersion: opentelemetry.io/v1alpha1
kind: OpenTelemetryCollector
metadata:
  name: test
  namespace: collector-deployment
spec:
  mode: deployment
  replicas: 1
  resources: 
    requests:
      cpu: 200m
      memory: 300Mi
    limits:
      memory: 1000Mi

  serviceAccount: test-sa
  config: |
    receivers:
      otlp:
        protocols:
          http:    
    processors:
      batch:
    exporters:
      awsxray:
        region: us-west-2
    service:
      pipelines:
        traces/xray:
          receivers:
            - otlp
          processors:
            - batch
          exporters:
            - awsxray
```
## Expected Result
`test-collector` service account does not get created since `test-sa` is defnied in spec. 
## Actual Result
```
bryaag@147ddac12c15 eksTemplates % kubectl get serviceaccounts -n collector-deployment 
NAME             SECRETS   AGE
default          0         10m
test-collector   0         6s
test-sa          0         6s
```

### Kubernetes Version

1.26

### Operator version

v0.88.0

### Collector version

v0.88.0

## Environment
EKS, AL2 AMI

changexd:
It seems like operator didn't check if `OtelCol.Spec.ServiceAccount == ""`  when building manifest for collectors, so the service account will still be created even though other resources already changed to the specified service account, @jaronoff97 could you please assign this issue to me?

dlorent:
Not entirely sure if it's a configuration issue or if its related. I got the same issue as @bryan-aguilar
but additionally, it seems to ignore the service account specified, if the deployment of Application-X has one specified too.


v1.28.3-gke.1118000 Container-Optimized OS from Google

Tested with

Operator 0.87,88,89
Collector-contrib 0.87,88,89

```

apiVersion: apps/v1
kind: Deployment
metadata:
 ...............
spec:
  ...................
  template:
    metadata:
      annotations:
[...]

PR #2378, opened 2023-11-21
Fix service account is created even when one is specified in collector spec 

changexd:
Fix service account is created even when one is specified in collector spec 

#2372


swiatekm-sumo (commented):
Could you add a test verifying that the ServiceAccount *doeesn't* get created if the value is set? LGTM otherwise.

internal/manifests/collector/collector.go
```diff
@@ -45,12 +45,14 @@ func Build(params manifests.Params) ([]client.Object, error) {
 	manifestFactories = append(manifestFactories, []manifests.K8sManifestFactory{
 		manifests.FactoryWithoutError(ConfigMap),
 		manifests.FactoryWithoutError(HorizontalPodAutoscaler),
-		manifests.FactoryWithoutError(ServiceAccount),
```
jaronoff97, line 48: Could we instead just check this in the service account manifest? In the builder we would then return nil if service account is set

jaronoff97 (changes requested):

changexd:
I wasn't sure if the test should be placed in service account manifest, so I simply added test to builder since we would like to check if the builder build service account manifest when specified.

jaronoff97 (approved):

pavolloffay (approved):

Issue #2393, opened 2023-11-28
Create the permissions requested by the resourcedetection processor from the operator


iblancasa:
### Component(s)

operator

### Is your feature request related to a problem? Please describe.

The `resourcedetection` processor needs some extra permissions when enabled. For instance, for OpenShift, those permissions are:
```yaml
kind: ClusterRole
metadata:
  name: otel-collector
rules:
- apiGroups: ["config.openshift.io"]
  resources: ["infrastructures", "infrastructures/status"]
  verbs: ["get", "watch", "list"]
```

### Describe the solution you'd like

It would be great if we can create them automatically from the operator.

PR #2394, opened 2023-11-28
Automate the creation of the permissions needed by resourcedetection

iblancasa:
- Add a methods to create the ClusterRoles and ClusterRoleBindings needed
- Add the structs and methods to parse the processors
- Add support for the resourcedetection processor

**Link to tracking Issue:** #2393


internal/manifests/collector/rbac.go
```diff
@@ -0,0 +1,91 @@
...
+	rules := adapters.ConfigToRBAC(params.Log, configFromString)
+
+	if len(rules) == 0 {
+		return nil
+	}
+
+	return &rbacv1.ClusterRole{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        naming.ClusterRole(params.OtelCol.Name, params.OtelCol.Namespace),
+			Annotations: params.OtelCol.Spec.Ingress.Annotations,
```
swiatekm-sumo, line 43: Why the ingress annotations?

internal/manifests/collector/rbac.go
```diff
@@ -0,0 +1,91 @@
...
+	return &rbacv1.ClusterRoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        naming.ClusterRoleBinding(params.OtelCol.Name),
+			Annotations: params.OtelCol.Spec.Ingress.Annotations,
+			Labels: map[string]string{
+				"app.kubernetes.io/name":       naming.ClusterRoleBinding(params.OtelCol.Name),
+				"app.kubernetes.io/instance":   fmt.Sprintf("%s.%s", params.OtelCol.Namespace, params.OtelCol.Name),
+				"app.kubernetes.io/managed-by": "opentelemetry-operator",
+				"app.kubernetes.io/component":  "opentelemetry-collector",
+			},
```
swiatekm-sumo, line 76: See comments above.

internal/naming/main.go
```diff
@@ -125,6 +125,16 @@ func Route(otelcol string, prefix string) string {
...
+// ClusterRole builds the cluster role name based on the instance.
+func ClusterRole(otelcol string, namespace string) string {
+	return DNSName(Truncate("%s-%s-cluster-role", 63, otelcol, namespace))
+}
+
+// ClusterRoleBinding builds the cluster role binding name based on the instance.
+func ClusterRoleBinding(otelcol string) string {
+	return DNSName(Truncate("%s-cluster-role-binding", 63, otelcol))
+}
+
```
swiatekm-sumo, line 137: I think just `%s-collector` would be better here. That's how we name most other resources, and I don't see any reason to embed the type in the name here.

internal/manifests/collector/adapters/config_to_ports.go
```diff
@@ -93,6 +94,8 @@ func ConfigToComponentPorts(logger logr.Logger, cType ComponentType, config map[
 			cmptParser, err = exporterParser.For(logger, cmptName, exporter)
 		case ComponentTypeReceiver:
 			cmptParser, err = receiverParser.For(logger, cmptName, exporter)
+		case ComponentTypeProcessor:
+			logger.V(2).Info("processors don't provide a way to enable associated ports", "name", key)
```
jaronoff97, line 98: I think this should be a debug log (`V(4)`), otherwise this is going to get spammy

internal/manifests/collector/adapters/config_to_rbac.go
```diff
@@ -0,0 +1,62 @@
...
+		return nil
+	}
+
+	processors, ok := processorsRaw.(map[interface{}]interface{})
+	if !ok {
+		logger.V(2).Info("processors doesn't contain valid components")
+		return nil
+	}
+
+	enabledExporters := getEnabledComponents(config, ComponentTypeProcessor)
```
jaronoff97, line 37:
```suggestion
	enabledProcessors := getEnabledComponents(config, ComponentTypeProcessor)
```

internal/manifests/collector/parser/processor/processor_resourcedetection.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	detectors, ok := detectorsCfg.([]interface{})
+	if !ok {
+		return prs
+	}
+	for _, d := range detectors {
+		detectorName := fmt.Sprint(d)
+		switch detectorName {
+		case "openshift":
```
jaronoff97, line 66: while we're here, we should also do the kubernetes node one too [link here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/README.md#k8s-node-metadata)

jaronoff97 (commented):
I love the spirit of this PR, and think it would be great if we continued this for other processors like the k8sattr processor

internal/manifests/collector/rbac.go
iblancasa: Copy/paste error. Thanks!

internal/manifests/collector/rbac.go
```diff
@@ -0,0 +1,91 @@
...
+
+	return &rbacv1.ClusterRole{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        naming.ClusterRole(params.OtelCol.Name, params.OtelCol.Namespace),
+			Annotations: params.OtelCol.Spec.Ingress.Annotations,
+			Labels: map[string]string{
+				"app.kubernetes.io/name":       naming.ClusterRole(params.OtelCol.Name, params.OtelCol.Namespace),
+				"app.kubernetes.io/instance":   fmt.Sprintf("%s.%s", params.OtelCol.Namespace, params.OtelCol.Name),
+				"app.kubernetes.io/managed-by": "opentelemetry-operator",
+				"app.kubernetes.io/component":  "opentelemetry-collector",
```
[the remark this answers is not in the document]
iblancasa: You're right

internal/manifests/collector/adapters/config_to_rbac.go
iblancasa: Good catch! Thanks

pavolloffay (approved):

pavolloffay:
@iblancasa CI failed

swiatekm-sumo (commented):
LGTM, but I'd like to a e2e test as well.

pavolloffay:
@iblancasa the PR needs to be rebased, also @swiatekm-sumo suggested adding e2e test

pkg/featuregate/featuregate.go
```diff
@@ -90,6 +90,14 @@ var (
...
 		featuregate.WithRegisterFromVersion("v0.82.0"),
 	)
+
+	// CreateRBACPermissions is the feature gate that enables the automatic creation of RBAC permissions depending on the used processors.
+	CreateRBACPermissions = featuregate.GlobalRegistry().MustRegister(
+		"operator.collector.createrbacpremissions",
+		featuregate.StageAlpha,
+		featuregate.WithRegisterDescription("enables the automatic creation of RBAC permissions depending on the used processors"),
+		featuregate.WithRegisterFromVersion("v0.91.0"),
+	)
```
swiatekm-sumo, line 100: I'm wondering if we shouldn't make this an operator-level configuration option instead of a feature flag. A feature flag implies we eventually want to make this behavior impossible to disable, and I can imagine users wanting to disable it and handle RBAC on their own.

iblancasa: Right now I see this as a way to enable/disable the feature. Not as something to be "mandatory" in the future. If this is not the correct way of doing it, I can change it. Would it be a better idea to have a CLI argument or how do you propose to implement this?

pavolloffay: Did we explore to make it work without the feature flag? The operator could issue SAR request to figure out its permissions and act accordingly or just log warning (send warning in webhook) if it cannot create RBAC.

controllers/opentelemetrycollector_controller.go
```diff
@@ -138,7 +141,11 @@ func (r *OpenTelemetryCollectorReconciler) SetupWithManager(mgr ctrl.Manager) er
 		Owns(&appsv1.DaemonSet{}).
 		Owns(&appsv1.StatefulSet{}).
 		Owns(&autoscalingv2.HorizontalPodAutoscaler{}).
-		Owns(&policyV1.PodDisruptionBudget{})
+		Owns(&policyV1.PodDisruptionBudget{}).
+		Owns(&autoscalingv2.HorizontalPodAutoscaler{}).
```
pavolloffay, line 145: This is duplicated

controllers/opentelemetrycollector_controller.go
```diff
@@ -138,7 +141,11 @@ func (r *OpenTelemetryCollectorReconciler) SetupWithManager(mgr ctrl.Manager) er
...
+		Owns(&policyV1.PodDisruptionBudget{}).
```
pavolloffay, line 146: This is duplicated

main.go
```diff
@@ -124,6 +125,7 @@ func main() {
 	pflag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	pflag.BoolVar(&createRBACPermissions, "create-rbac-permissions", false, "Create automatically RBAC permissions needed by the processors")
```
jaronoff97, line 128:
```suggestion
	pflag.BoolVar(&createRBACPermissions, "create-rbac-permissions", false, "Automatically create RBAC permissions needed by the collector's processors")
```

main.go
```diff
@@ -176,6 +178,7 @@ func main() {
 		config.WithLogger(ctrl.Log.WithName("config")),
 		config.WithVersion(v),
 		config.WithCollectorImage(collectorImage),
+		config.WithCreateRBACPermissions(createRBACPermissions),
```
jaronoff97, line 181: suggestion (take it or leave it): we could incorporate the new permissions checking capability to see if the service account for the operator has access to create RBAC permissions. I'm unsure if that would go in the operator logs, collector webhook, etc. What do you think?

internal/manifests/collector/collector.go
```diff
@@ -51,13 +51,22 @@ func Build(params manifests.Params) ([]client.Object, error) {
...
+
 	if params.OtelCol.Spec.Observability.Metrics.EnableMetrics && featuregate.PrometheusOperatorIsAvailable.IsEnabled() {
 		if params.OtelCol.Spec.Mode == v1alpha1.ModeSidecar {
 			manifestFactories = append(manifestFactories, manifests.Factory(PodMonitor))
 		} else {
 			manifestFactories = append(manifestFactories, manifests.Factory(ServiceMonitor))
 		}
 	}
+
+	if !params.Config.CreateRBACPermissions() {
```
jaronoff97, line 63: shouldn't this be the reverse?

controllers/opentelemetrycollector_controller.go
```diff
@@ -83,10 +84,12 @@ func NewReconciler(p Params) *OpenTelemetryCollectorReconciler {
 // +kubebuilder:rbac:groups=apps,resources=daemonsets;deployments;statefulsets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=autoscaling,resources=horizontalpodautoscalers,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=policy,resources=poddisruptionbudgets,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterrolebindings;clusterroles,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get;list;create;update
 // +kubebuilder:rbac:groups=monitoring.coreos.com,resources=servicemonitors;podmonitors,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=networking.k8s.io,resources=ingresses,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=route.openshift.io,resources=routes;routes/custom-host,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=config.openshift.io,resources=infrastructures;infrastructures/status,verbs=get;list;watch
```
jaronoff97, line 92: why do we need this now? I thought this is the permission we're adding for the collector?

iblancasa: It is needed for the `resourcedetection` processor. Without this, I cannot create th RBAC.

main.go
iblancasa: But that check will be always `true` since we are adding the permissions as part of the bundle, right?

jaronoff97: not really. This is my confusion as to why we're adding this permission to the bundle in the first place. Aren't we adding this RBAC permission to the serviceaccount of the collector not the operator? We shouldn't need the operator to have this RBAC permission because the operator isn't doing any operations on the config.openshift.io group, right?

iblancasa: You cannot assign permissions over one service account if the service account has no permission over the resource. In other words: the operator cannot give permissions to the collector service account for the config.openshift.io group if the service account of the operator has no permissions over config.openshift.io group.

jaronoff97: ahhh i see, thank you very much for explaining! Now i understand.

jaronoff97 (approved):

internal/manifests/collector/adapters/config_to_rbac_test.go
```diff
@@ -0,0 +1,98 @@
...
+					APIGroups: []string{"config.openshift.io"},
+					Resources: []string{"infrastructures", "infrastructures/status"},
+					Verbs:     []string{"get", "watch", "list"},
+				},
+			},
+		},
+	}
+
+	var logger = logf.Log.WithName("collector-unit-tests")
+	for _, test := range tests {
```
pavolloffay, line 87: It's good that the test table is used but the tests should be run in a separate execution `t.Run`

internal/manifests/collector/adapters/config_to_rbac.go
```diff
@@ -0,0 +1,62 @@
...
+package adapters
+
+import (
+	"github.com/go-logr/logr"
+	rbacv1 "k8s.io/api/rbac/v1"
+
+	"github.com/open-telemetry/opentelemetry-operator/internal/manifests/collector/parser/processor"
+)
+
+func ConfigToRBAC(logger logr.Logger, config map[interface{}]interface{}) []rbacv1.PolicyRule {
```
pavolloffay, line 24: nit: it's worth adding a comment to a public API

internal/manifests/collector/parser/processor/processor.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/go-logr/logr"
+	rbacv1 "k8s.io/api/rbac/v1"
+)
+
+type ProcessorParser interface {
```
pavolloffay, line 26: nit: document public APIs

internal/manifests/collector/parser/processor/processor.go
```diff
@@ -0,0 +1,71 @@
...
+	ParserName() string
+	GetRBACRules() []rbacv1.PolicyRule
+}
+
+type Builder func(logr.Logger, string, map[interface{}]interface{}) ProcessorParser
```
pavolloffay, line 31: nit: document public APIs

main.go
```diff
@@ -124,6 +125,7 @@ func main() {
 	pflag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	pflag.BoolVar(&createRBACPermissions, "create-rbac-permissions", false, "Automatically create RBAC permissions needed by the processors")
```
pavolloffay, line 128: did we discuss whether it should be disabled by default? 

I would prefer to ship the OSS distro with as much enabled functionality.

pavolloffay (commented):
LGTM 👍🏼 

just few nit comments before merning.

pavolloffay (approved):

PR #2398, opened 2023-11-28
Remove old autodetection for openshift routes, cleanup

### LINKED ISSUE #2108: Migrate OpenShift route reconciliation to use manifests
Currently, reconciliation for routes is handled in a bit of an odd way, we should centralize this to be done via manifest / only once https://github.com/open-telemetry/opentelemetry-operator/blob/main/controllers/opentelemetrycollector_controller.go#L72-L92

jaronoff97:
**Description:** 
Removes autodetection running on a frequency, removes old route task code. For the future we can also use this autodetection for seeing that someone has the latest CRDs installed.

**Link to tracking Issue:** closes #2108 

**Testing:** Unit tests were added

**Documentation:** N/A


controllers/opentelemetrycollector_controller_test.go
```diff
@@ -229,60 +251,130 @@ func TestNewStatefulSetObjectsOnReconciliation(t *testing.T) {
...
+	// cleanup the deployment deliberately, otherwise a local tester will always fail as there is no gc event.
+	list := &appsv1.DeploymentList{}
+	err = k8sClient.List(context.Background(), list, opts...)
+	assert.NoError(t, err)
+	assert.Len(t, list.Items, 1)
+	require.NoError(t, k8sClient.Delete(context.Background(), list.Items[0].DeepCopy()))
 }
 
-func TestContinueOnRecoverableFailure(t *testing.T) {
+func TestNewStatefulSetObjectsOnReconciliation(t *testing.T) {
```
jaronoff97, line 292: This is a bad diff from GitHub, I only added a test but it's not recognizing it as an add

internal/config/main.go
```diff
@@ -99,20 +93,7 @@ func New(opts ...Option) Config {
 // StartAutoDetect attempts to automatically detect relevant information for this operator. This will block until the first
 // run is executed and will schedule periodic updates.
 func (c *Config) StartAutoDetect() error {
```
jaronoff97, line 95: I need to remove this method, it's not used anymore

main.go
```diff
@@ -186,6 +186,10 @@ func main() {
 		config.WithAutoDetect(ad),
 		config.WithLabelFilters(labelsFilter),
 	)
+	err = cfg.AutoDetect()
```
jaronoff97, line 189: In the future we could autodetect more things like if certain CRDs are available. We honestly should do this for Prometheus too.

yuriolisa (approved):
LGTM

pkg/collector/reconcile/route.go
```diff
@@ -1,156 +0,0 @@
...
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/types"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-
-	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
-)
-
-// Routes reconciles the route(s) required for the instance in the current context.
-// TODO: This functionality should be put with the rest of reconciliation logic in the mutate.go
```
pavolloffay, line 34: Why did we keep this? The mutate is already handling routes

jaronoff97: This was because the old task logic would dynamically add this to its jobs. either way, i think it's probably wholly unused.

pavolloffay (approved):

PR #2404, opened 2023-11-30
Added PodMonitor when using sidecar mode

### LINKED ISSUE #2306: Monitors for Prometheus exporter in sidecar mode
### Component(s)

operator

Relates to #1983 

Generating a ServiceMonitor for a collector using  `observability.metrics.enableMetrics: true` works fine using the `deployment` mode, but when generating for the `sidecar` mode, the ServiceMonitor will not scrape metrics from the sidecar.

I don't believe this can be solved using the current implementation with ServiceMonitors, as the operator won't have visibility of the services generated for the injected applications.

Could we use a `PodMonitor` instead in the sidecar case? Perhaps matching on the `sidecar.opentelemetry.io/injected` label. Also won't want to restrict to the namespace the collector is defined in, as it could feasibly be used from any.

_No response_

### Additional context

_No response_

yuriolisa:

Adding PodMonitor when using a combination of sidecar mode and Prometheus exporters.
Resolves #2306 
Created e2e test which will deploy a collector using sidecar mode and tested if we get the expected PodMonitor created.


internal/manifests/collector/collector.go
```diff
@@ -51,9 +51,12 @@ func Build(params manifests.Params) ([]client.Object, error) {
 		manifests.Factory(MonitoringService),
 		manifests.Factory(Ingress),
 	}...)
-	if params.OtelCol.Spec.Observability.Metrics.EnableMetrics && featuregate.PrometheusOperatorIsAvailable.IsEnabled() {
+	if params.OtelCol.Spec.Observability.Metrics.EnableMetrics && featuregate.PrometheusOperatorIsAvailable.IsEnabled() && params.OtelCol.Spec.Mode != v1alpha1.ModeSidecar {
 		manifestFactories = append(manifestFactories, manifests.Factory(ServiceMonitor))
 	}
+	if params.OtelCol.Spec.Observability.Metrics.EnableMetrics && featuregate.PrometheusOperatorIsAvailable.IsEnabled() && params.OtelCol.Spec.Mode == v1alpha1.ModeSidecar {
+		manifestFactories = append(manifestFactories, manifests.Factory(PodMonitor))
+	}
```
swiatekm-sumo, line 59: Can we change this to a nested conditional? Like:

```go
if params.OtelCol.Spec.Observability.Metrics.EnableMetrics && featuregate.PrometheusOperatorIsAvailable.IsEnabled() {
  if params.OtelCol.Spec.Mode == v1alpha1.ModeSidecar {
    manifestFactories = append(manifestFactories, manifests.Factory(PodMonitor))
  } else {
    manifestFactories = append(manifestFactories, manifests.Factory(ServiceMonitor))
  }
}
```

This is clearer imo.

yuriolisa: Thx, good catch!

internal/manifests/collector/podmonitor.go
```diff
@@ -0,0 +1,98 @@
...
+	config, err := adapters.ConfigFromString(otelcol.Spec.Config)
+	if err != nil {
+		logger.V(2).Error(err, "Error while parsing the configuration")
+		return []monitoringv1.PodMetricsEndpoint{}
+	}
+	exporterPorts, err := adapters.ConfigToComponentPorts(logger, adapters.ComponentTypeExporter, config)
+	if err != nil {
+		logger.Error(err, "couldn't build endpoints to podMonitors from configuration")
+		return []monitoringv1.PodMetricsEndpoint{}
+	}
```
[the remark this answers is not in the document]
yuriolisa: I didn't get the idea. Could you please clarify it?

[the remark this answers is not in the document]
swiatekm-sumo: I'm wondering if we should return the error and fail manifest generation instead of just returning an empty slice here. Is there ever a situation where this fails but the overall configuration is still valid? Looking at the code for `ConfigToComponentPorts`, this should never happen.

But on the other hand, failing reconciliation just because we can't create a PodMonitor would be a bit extreme. So this is most likely better.

swiatekm-sumo (approved):
:+1:

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -378,7 +380,7 @@ spec:
                 - --enable-leader-election
                 - --zap-log-level=info
                 - --zap-time-encoding=rfc3339nano
-                - --feature-gates=+operator.autoinstrumentation.go,+operator.autoinstrumentation.nginx
+                - --feature-gates=+operator.autoinstrumentation.go,+operator.observability.prometheus,+operator.autoinstrumentation.nginx
```
swiatekm-sumo: Is this an intended change?

yuriolisa: No, I was debugging why the `podmonitors` were not being created and by accident, I pushed this change. I've just fixed and the e2e should be fine now.

jaronoff97 (approved):

PR #2416, opened 2023-12-05
Create ServiceMonitor to monitor target allocator

### LINKED ISSUE #2311: [target allocator] Create ServiceMonitor to monitor target allocator
### Component(s)

target allocator

With the support for more target allocator specific [metrics](https://github.com/open-telemetry/opentelemetry-operator/pull/2169), it would be nice to have them scraped automatically.

Operator creates ServiceMonitor CR for target allocator, similar to the [one for otel-collectors](https://github.com/open-telemetry/opentelemetry-operator/pull/1874).
Using a config `targetAllocator.observability.metrics.enable` would be helpful to enable/disable exposing and scraping TA metrics.

_No response_

### Additional context

_No response_

changexd:
**Description:** 
this resolves #2311

#2311

I've added checks to make sure collector gets the target-allocator jobs

**side note**
I basically followed implementation from #1874, however I've encountered couple issues I wasn't sure if these are expected or not, if these needed to be fixed, I'll open issues and am willing to fix it.

1. The observability params in collector is not validated, so I did not add it either, should we add validation to these params?

2. The label `app.kubernetes.io/instance` from default service of target allocator is `$namespace.$collectorName` instead of `$namespace.$targetAllocatorName`, should we change this?

3. The service port name from TA is specifically map from “targetallocation” to container port “http”, but the collector service itself maps  “prometheus” to container port “prometheus”, is this intentional?, if not, should this be fixed?

4. After disabling observability, the manager doesn't remove the servicemonitor, only after deletion of otelcol resource, it is then removed, this happens to both collector and target allocator.


.chloggen/2311-ta-servicemonitor.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: operator
```
pavolloffay, line 5: target allocator

internal/manifests/targetallocator/servicemonitor.go
```diff
@@ -0,0 +1,70 @@
...
+			Name:      naming.TargetAllocator(params.OtelCol.Name),
+			Labels: map[string]string{
+				"app.kubernetes.io/name":       naming.TargetAllocator(params.OtelCol.Name),
+				"app.kubernetes.io/instance":   fmt.Sprintf("%s.%s", params.OtelCol.Namespace, params.OtelCol.Name),
+				"app.kubernetes.io/managed-by": "opentelemetry-operator",
+				"app.kubernetes.io/component":  "opentelemetry-targetallocator",
+			},
+		},
+		Spec: monitoringv1.ServiceMonitorSpec{
+			Endpoints: []monitoringv1.Endpoint{},
```
yuriolisa, line 49: Do we have only the targetallocation port? If so, it should be on sm definition.

internal/manifests/targetallocator/servicemonitor.go
```diff
@@ -0,0 +1,70 @@
...
+// ServiceMonitor returns the service monitor for the given instance.
+func ServiceMonitor(params manifests.Params) (*monitoringv1.ServiceMonitor, error) {
+	if !params.OtelCol.Spec.TargetAllocator.Observability.Metrics.EnableMetrics {
+		params.Log.V(2).Info("Metrics disabled for this OTEL Collector",
+			"params.OtelCol.name", params.OtelCol.Name,
+			"params.OtelCol.namespace", params.OtelCol.Namespace,
+		)
+		return nil, nil
+	}
+
```
yuriolisa, line 36: Please add a condition to avoid creating sm when running in sidecar mode.

yuriolisa (changes requested):

internal/manifests/targetallocator/servicemonitor.go
changexd: Will target allocator support sidecar mode? It seems like setting `mode != statefulset` with target allocator enabled will be denied by the admission webhook and disabling target allocator will not generate any ta manifest either.

swiatekm-sumo: We already check for this during validation, we shouldn't reproduce this check here as well imo. If anything, I'd move the check for `EnableMetrics` here from `targetallocator.go`.

tests/e2e/targetallocator-prometheuscr/00-install.yaml
```diff
@@ -107,6 +107,9 @@ spec:
     image: "local/opentelemetry-operator-targetallocator:e2e"
     prometheusCR:
       enabled: true
+    observability:
+      metrics:
+        enableMetrics: true
```
pavolloffay, line 112: shall we assert that the service monitor is created?

pavolloffay: @changexd since we have e2e test it would be good to add assertion on the created service monitor.

bundle/manifests/opentelemetry.io_opentelemetrycollectors.yaml
```diff
@@ -4966,6 +4966,22 @@ spec:
...
+                    properties:
+                      metrics:
+                        description: Metrics defines the metrics configuration for
+                          operands.
+                        properties:
+                          enableMetrics:
+                            description: EnableMetrics specifies if ServiceMonitor
+                              or PodMonitor(for sidecar mode) should be created for
+                              the OpenTelemetry Collector and Prometheus Exporters.
+                              The operator.observability.
```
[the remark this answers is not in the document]
changexd: is this okay?
~~for  the OpenTelemetry Collector and Prometheus Exporters.~~ 
-> for the service managed by the OpenTelemetry Operator.

internal/manifests/targetallocator/servicemonitor.go
changexd: It seems like I also added checks in `targetallocator.go` , so I'll remove it from the `servicemonitor.go`, however I would like to add tests for building targetallocator and it seems like `builder_test.go`  only covers collector and OpAMPBridge, should there be dedicated test for targetallocator, or should I just append test case to the collector tests?

jaronoff97 (approved):
I think this LGTM. @changexd are all of the open questions resolved?

yuriolisa (approved):

changexd:
Not yet, there are two questions waiting for confirmation, should I make the changes first and wait for reviews?

internal/manifests/targetallocator/servicemonitor.go
```diff
@@ -0,0 +1,62 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/open-telemetry/opentelemetry-operator/internal/manifests"
+	"github.com/open-telemetry/opentelemetry-operator/internal/naming"
+)
+
+// ServiceMonitor returns the service monitor for the given instance.
+func ServiceMonitor(params manifests.Params) *monitoringv1.ServiceMonitor {
+
+	sm := monitoringv1.ServiceMonitor{
```
pavolloffay, line 30: it can be returned directly, also the empty line above can be removed

kuttl-test.yaml
```diff
@@ -1,6 +1,9 @@
 apiVersion: kuttl.dev/v1beta1
 kind: TestSuite
 artifactsDir: ./tests/_build/artifacts/
+commands:
+  - command: make undeploy
+  - command: make enable-prometheus-feature-flag deploy
```
pavolloffay, line 6: Could be the test rather placed into `e2e-prometheuscr`. That suite contains the tests for prometheus integration

internal/manifests/targetallocator/servicemonitor.go
```diff
@@ -0,0 +1,59 @@
...
+	return &monitoringv1.ServiceMonitor{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: params.OtelCol.Namespace,
+			Name:      naming.TargetAllocator(params.OtelCol.Name),
+			Labels: map[string]string{
+				"app.kubernetes.io/name":       naming.TargetAllocator(params.OtelCol.Name),
+				"app.kubernetes.io/instance":   fmt.Sprintf("%s.%s", params.OtelCol.Namespace, params.OtelCol.Name),
+				"app.kubernetes.io/managed-by": "opentelemetry-operator",
+				"app.kubernetes.io/component":  "opentelemetry-targetallocator",
+			},
```
swiatekm-sumo, line 38: There's a `Labels` function that generates these, can you use it here?

internal/manifests/targetallocator/servicemonitor.go
```diff
@@ -0,0 +1,59 @@
...
+
+			NamespaceSelector: monitoringv1.NamespaceSelector{
+				MatchNames: []string{params.OtelCol.Namespace},
+			},
+			Selector: metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"app.kubernetes.io/managed-by": "opentelemetry-operator",
+					"app.kubernetes.io/instance":   fmt.Sprintf("%s.%s", params.OtelCol.Namespace, params.OtelCol.Name),
+					"app.kubernetes.io/component":  "opentelemetry-targetallocator",
```
swiatekm-sumo, line 54: Why isn't this identical to the labels defined above?

internal/manifests/targetallocator/targetallocator.go
```diff
@@ -32,6 +33,11 @@ func Build(params manifests.Params) ([]client.Object, error) {
 		manifests.FactoryWithoutError(ServiceAccount),
 		manifests.FactoryWithoutError(Service),
 	}
+
+	if params.OtelCol.Spec.TargetAllocator.Observability.Metrics.EnableMetrics && featuregate.PrometheusOperatorIsAvailable.IsEnabled() {
+		resourceFactories = append(resourceFactories, manifests.FactoryWithoutError(ServiceMonitor))
+	}
```
swiatekm-sumo, line 39: Should we create an explicit convention for where we want these checks to live @jaronoff97 @pavolloffay? My impression was that they should be in the manifest generation functions, which should return nil if there's nothing to generate. But I looked at the same logic for the collector, and there the check is also done in the builder.

I don't think this question necessarily needs to block the PR, we can change it later. But I'd like to have clarity here.

internal/manifests/targetallocator/servicemonitor.go
swiatekm-sumo: If you want to add those tests, please do so in a separate test function.

swiatekm-sumo (commented):
LGTM overall, the label generation should use existing functions. I'd also like to clarify where we should do the top-level checks, but that shouldn't block this PR imo.

internal/manifests/targetallocator/targetallocator.go
pavolloffay: +1 on being consistent and putting this knowledge into the manifest functions.

The manifest generation functions gets `manifests.Params` and can make that decision and encapsulate when the objects should be generated.

pavolloffay (approved):

jaronoff97:
@changexd looks like approvals are looking good, when you get a chance could you resolve the conflicts so we can get this merged before the next release?

internal/manifests/targetallocator/targetallocator.go
jaronoff97: I was originally thinking the check would be done in the resource's builder (like where it is now) so it's easy to understand what should be appearing. This would make individual unit testing easier for each of the manifests as they don't need to check on featuregates, however checking if a feature is enabled to make the manifest should indeed happen within the manifest function. I think I would then recommend the following:
* featurgates belong in the builder for the resource
* fields of the resource controlling whether a manifest is created (hpa, ingress, etc.) should be handled in the manifest itself

I think regardless of what we choose, we should do that in a followup.

iblancasa (approved):

changexd:
@jaronoff97 I've resolved to conflicts in code, as for `api.md`, I use `make api-docs`, is this okay or should I resolve it manually?

swiatekm-sumo (approved):

jaronoff97:
@changexd that's actually fine – it looks like we missed some api-docs changes in a recent version upgrade. I ran it off of main and there was a diff, so this resolves that 😅

jaronoff97:
thank you for your contribution :D :tada:

PR #2420, opened 2023-12-06
fix error logging in collector container creation

changexd:
**Description:**
this fix the bug that prints out error logs without checking when creating collector container resource


swiatekm-sumo (approved):
👍 Please add a changelog entry and we'll be good to go.

yuriolisa (approved):

PR #2430, opened 2023-12-08
[target-allocator] Introduce "per node" allocation strategy to target allocator

### LINKED ISSUE #1828: TargetAllocator for daemonset
Ideally, when deployed as a daemonset, collectors would scrape the pods running on the node
local to the daemonset instance. It's not obvious from the docs or configs as to if that is achievable
with the current targetAllocator strategies.

Am I missing something? Would you accept a PR that implements this?

matej-g:
**Description:**
Resolves #1828.

Introduces new allocation strategy "per node" that is intended to distribute targets to collectors on the basis of the node on which the targets reside, i.e. each collector will be instructed to scrape targers running on the collector's node.

This strategy is intended to be only used with a collector in `deamonset` mode (running as agent on each node). It's not suitable for other modes.

**Link to tracking Issue:** #1828 

**Testing:**
- Added unit tests
- Also did manual testing

**Documentation:** Updated API documentation.


cmd/otel-allocator/allocation/per_node.go
```diff
@@ -0,0 +1,295 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/model"
+)
+
+var _ Allocator = &perNodeAllocator{}
+
+const (
+	perNodeStrategyName = "per-node"
+
+	podNodeNameLabel model.LabelName = "__meta_kubernetes_pod_node_name"
```
matej-g, line 32: One doubt I'm having is how to make the label matching more robust, I'm mainly thinking about:
- Which service discoveries should we be targeting here? For service / pod monitors, we should be able depend on `__meta_kubernetes_pod_node_name`, but I wonder if there might be cases where this label is not available (perhaps due to using different role for Kubernetes SD or using different SD? Has anyone encountered people using other SDs?)
- How to handle targets that we might not match against any node (such as static configs), currently I'm skipping those. I'm thinking if that's fine (maybe should be documented?) or we want to have some type of fallback to distribute such targets, if that makes sense?

swiatekm-sumo: For the labels, we are constrained by the fact that the value needs to match the collector's Node field. So we can't do much better than Kubernetes Service Discovery, or at least not without significant additional complexity. So what I'd do there is just check all the Node fields from SD:

* `__meta_kubernetes_node_name`
* `__meta_kubernetes_pod_node_name`
* `__meta_kubernetes_endpoint_node_name`

The problem of assigning targets without a Node doesn't have an obvious solution. I can see the following options:

* Ignore these targets
* Assign them randomly
* Assign them according to some deterministic scheme (could be what the consistent hashing strategy uses)

We should discuss this in the issue, rather than here imo.

swiatekm-sumo:
> I also noticed we are reusing a lot of code between allocation strategies. I was thinking about adjusting the abstraction (perhaps allocator does not need to be an interface, we could maybe have more specific interface that would implement only methods that differ across strategies). But I did not want to introduce too many changes at once.

Yeah, we can do that afterwards in a separate PR.

cmd/otel-allocator/allocation/per_node.go
```diff
@@ -0,0 +1,321 @@
...
+// a number of OpenTelemetry collectors based on the node on which
+// the collector is running. This allocator should be used only when
+// collectors are running as daemon set (agent) on each node.
+// Users need to call SetTargets when they have new targets in their
+// clusters and call SetCollectors when the collectors have changed.
+type perNodeAllocator struct {
+	// m protects collectors and targetItems for concurrent use.
+	m sync.RWMutex
+	// collectors is a map from a Collector's name to a Collector instance
+	collectors map[string]*Collector
```
jaronoff97, line 42: thought: i think it may be more efficient if we were to store collectors as a map from nodename -> collector rather than the collector name. I think this would make a few operations like removing a collector a bit less efficient, but it would make target assignment much more efficient. This is an acceptable tradeoff as i would imagine targets change more often than collectors

jaronoff97 (approved):
Overall I really like the improvement here. I have one suggestion which may help with making it more efficient, interested in hearing your thoughts. Thank you!

swiatekm-sumo:
> Thanks for the pointer @swiatekm-sumo, I have added an E2E test and also reorganized the tests a bit so that TA E2E test can be run separately under separate target (hopefully you don't mind 🙂).

Is there a reason to do this? If you only want the target allocator tests, you can use kuttl's `--test` flag to select them.

And if you really want to, you should make sure they're run in the CI as well.

cmd/otel-allocator/allocation/per_node.go
matej-g: Makes sense to me and it's a simple optimization, made the change

matej-g:
> Is there a reason to do this? If you only want the target allocator tests, you can use kuttl's `--test` flag to select them.
> 
> And if you really want to, you should make sure they're run in the CI as well.

I think it makes sense with respect to how other test cases are organized, since we have usually different test directories for different components (OpAMP bridge, auto-instrumentation). But you're right, it was missing from the list in the action config, I fixed that now.

tests/e2e-targetallocator/targetallocator-kubernetessd/00-install.yaml
```diff
@@ -0,0 +1,133 @@
...
+  name: prometheus-kubernetessd
+spec:
+  mode: daemonset
+  serviceAccount: collector
+  targetAllocator:
+    enabled: true
+    serviceAccount: ta
+    image: "local/opentelemetry-operator-targetallocator:e2e"
+    prometheusCR:
+      enabled: false
```
jaronoff97, line 102: I may be missing something but don't we want to test the new strategy here? Wouldn't our final teststep in 01-install.yaml fail without installing the TA?

jaronoff97: wouldn't we expect [this](https://github.com/open-telemetry/opentelemetry-operator/pull/2430/files#diff-969909de4d71fcaae234fd2a1ce3bd7c03b726126e2443bb7dc04c8a5145b97eR49) to fail then? And also for [this target allocator](https://github.com/open-telemetry/opentelemetry-operator/pull/2430/files#diff-fbf022ca14c8b039493f038feac0716db643ac051a41ccb76bac47cb3bcefeb8R35-R38) section to be empty?

jaronoff97: should we require that daemonset mode has to run with the new strategy? right now we require the new strategy has daemonset but not the reverse. I don't think anything could go horribly wrong if you run with least-weighted and daemonset, but i would worry that too much node cycling would cause poorly balanced targets.

matej-g: I added a validation for this as well

matej-g:
Hey folks,
This should hopefully be ready for final reviews.

I was testing this a bit more and got a report about issue when collector pods are changed (e.g. added / removed nodes or daemonset being re-created), whereas after such change TA was not allocating target anymore. After a deep dive I finally figured this is because the node name in the pod spec is not always available when we get the pod data from watcher. To solve this, I added a condition to restart the watcher, until we receiver node name on collector pods. Let me know if this makes sense to you. Thanks.

cmd/otel-allocator/allocation/per_node.go
```diff
@@ -0,0 +1,296 @@
...
+			collectorAssigned := allocator.addTargetToTargetItems(item)
+			if !collectorAssigned {
+				unassignedTargetsForJobs[item.JobName] = struct{}{}
+			}
+		}
+	}
+
+	// Check for unassigned targets
+	if len(unassignedTargetsForJobs) > 0 {
+		jobs := make([]string, 0, len(unassignedTargetsForJobs))
```
jaronoff97, line 193: I think we should just make this a gauge metric that counts the length of the unassigned map or even just count the unassigned as an int and record that.

jaronoff97 (commented):
one comment, one suggestion. Let me know your thoughts on the suggestion

apis/v1alpha1/collector_webhook.go
```diff
@@ -229,10 +229,18 @@ func (c CollectorWebhook) validate(ctx context.Context, r *OpenTelemetryCollecto
...
-	if r.Spec.TargetAllocator.Enabled && r.Spec.Mode != ModeStatefulSet {
+	if r.Spec.TargetAllocator.Enabled && (r.Spec.Mode != ModeStatefulSet && r.Spec.Mode != ModeDaemonSet) {
 		return warnings, fmt.Errorf("the OpenTelemetry Collector mode is set to %s, which does not support the target allocation deployment", r.Spec.Mode)
 	}
 
+	if r.Spec.TargetAllocator.Enabled && (r.Spec.Mode == ModeDaemonSet && r.Spec.TargetAllocator.AllocationStrategy != OpenTelemetryTargetAllocatorAllocationStrategyPerNode) {
+		return warnings, fmt.Errorf("the OpenTelemetry Collector mode is set to %s, which must be used with target allocation strategy %s ", r.Spec.Mode, OpenTelemetryTargetAllocatorAllocationStrategyPerNode)
+	}
+
+	if r.Spec.TargetAllocator.AllocationStrategy == OpenTelemetryTargetAllocatorAllocationStrategyPerNode && r.Spec.Mode != ModeDaemonSet {
```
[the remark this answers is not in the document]
matej-g: I think that's a good idea idea at this point, I'll put TA validations into a separate method.

cmd/otel-allocator/allocation/per_node.go
matej-g: You mean in addition to logs or no logs, just metrics? I don't have strong opinion honestly, either solution works me (as long as it's not too noisy).

jaronoff97: I think maybe a debug log and metric of just the count of unassigned jobs. Otherwise it may get expensive to iterate through the entire list

pavolloffay:
@matej-g what is the status of this PR? It needs to be rebased.

matej-g:
> @matej-g what is the status of this PR? It needs to be rebased.

Hey, planning to address remaining feedback by Monday, after that hopefully @jaronoff97 can give final seal of approval

apis/v1alpha1/collector_webhook.go
```diff
@@ -228,40 +228,14 @@ func (c CollectorWebhook) validate(ctx context.Context, r *OpenTelemetryCollecto
...
 	if r.Spec.TargetAllocator.Enabled {
-		promCfg, err := ta.ConfigToPromConfig(r.Spec.Config)
-		if err != nil {
-			return warnings, fmt.Errorf("the OpenTelemetry Spec Prometheus configuration is incorrect, %w", err)
-		}
-		err = ta.ValidatePromConfig(promCfg, r.Spec.TargetAllocator.Enabled, featuregate.EnableTargetAllocatorRewrite.IsEnabled())
+		taWarnings, err := c.validateTargetAllocatorConfig(ctx, r)
 		if err != nil {
-			return warnings, fmt.Errorf("the OpenTelemetry Spec Prometheus configuration is incorrect, %w", err)
+			return warnings, err
```
jaronoff97, line 235: we should append the warnings from the above call before returning

cmd/otel-allocator/allocation/per_node.go
```diff
@@ -190,12 +190,8 @@ func (allocator *perNodeAllocator) handleTargets(diff diff.Changes[*target.Item]
 
 	// Check for unassigned targets
 	if len(unassignedTargetsForJobs) > 0 {
```
jaronoff97, line 192: this can just become an integer that's incremented on line 186, i don't think there's a need for a map, unless we care about deduping (which should happen already in the filter strategy)

apis/v1alpha1/collector_webhook.go
matej-g: D'oh... thanks, fixing it

cmd/otel-allocator/allocation/per_node.go
matej-g: I was wondering if there can be a scenario where we have multiple targets per job, that cannot be assigned. But I see the point, I simplified this to just count the number of targets, that should hopefully give enough info when debugging.

cmd/otel-allocator/allocation/per_node_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/prometheus/common/model"
+	"github.com/stretchr/testify/assert"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var loggerPerNode = logf.Log.WithName("unit-tests")
+
+// Tests that two targets with the same target url and job name but different label set are both added.
+func TestAllocationPerNode(t *testing.T) {
```
swiatekm-sumo, line 30: I'd like to see a few more tests for edge cases. No collectors, no targets, those kinds of conditions. You can probably copy those from consistent-hashing. Ideally we'd have a strategy-agnostic set of tests, but I'm not going to ask that you add those to this (already large and complex) PR.

swiatekm-sumo (commented):
Looks good overall. I've left some comments around trying to simplify the bookeeping code in the strategy implementation, but I don't think any of it is necessarily blocking - if you'd like to clean it up in a following PR, I'm ok with that as well.

matej-g:
Hey @swiatekm-sumo, thanks for the detailed review. I'm planning to address all of it, not sure how fast I will get to it. I'm happy to do it in this PR or the next one, whatever will be easier for maintainers to review. 

I'm also in no rush to merge this (I'm using a fork for my needs), but I'm not sure if there are other users waiting for this feature.

matej-g:
Hey folks,
Finally managed to get back to this, I tried to address all the feedback:
- I simplified the code as suggested where it was unnecessarily complicated from the methods taken over from other strategies
- Corrected the documentation where it referenced incorrect information
- Made the unassigned metric a gauge instead of counter
- Added unit test for edge cases (analogous from consistent hashing)

Hopefully this is ready for the final review, thank you @swiatekm-sumo @jaronoff97 🙇

cmd/otel-allocator/allocation/per_node.go
```diff
@@ -0,0 +1,292 @@
...
+// Any net-new additions are assigned to the collector on the same node as the target.
+func (allocator *perNodeAllocator) handleTargets(diff diff.Changes[*target.Item]) {
+	// Check for removals
+	for k, item := range allocator.targetItems {
+		// if the current item is in the removals list
+		if _, ok := diff.Removals()[k]; ok {
+			c, ok := allocator.collectors[item.GetNodeName()]
+			if !ok {
+				continue
+			}
```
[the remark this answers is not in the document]
matej-g: Fixed as suggested

cmd/otel-allocator/allocation/strategy.go
```diff
@@ -49,6 +49,10 @@ var (
 		Name: "opentelemetry_allocator_targets_remaining",
 		Help: "Number of targets kept after filtering.",
 	})
+	TargetsUnassigned = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "opentelemetry_allocator_targets_unassigned",
+		Help: "Number of targets that could not be assigned due to missing node label.",
+	})
```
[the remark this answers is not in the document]
matej-g: It was intended for both.

I changed to gauge, I'd agree it is more useful. Seems like we use interchangeably counters and gauges for different target number metrics, might be good candidate for some harmonization.

cmd/otel-allocator/allocation/per_node_test.go
matej-g: Added tests as suggested.

jaronoff97:
@matej-g thank you! what was the reason for changing the metric type? that may break anyone who has existing dashboards for the TA, we should just rename it too if we do want it to be a gauge

jaronoff97 (commented):
Just one concern around the metric type change, otherwise this LGTM.

matej-g:
Hey @jaronoff97, this was based off of the suggestion in https://github.com/open-telemetry/opentelemetry-operator/pull/2430#discussion_r1457269919 - this is a new metric we're introducing with this change, so it should not be breaking

jaronoff97 (approved):
Ah that's right. Looks good to me! I'll wait on @swiatekm-sumo 's review to merge. Thank you so much!

swiatekm-sumo (approved):

jaronoff97:
@matej-g thanks for this awesome work!! I really appreciate it.

swiatekm-sumo:
:clap: indeed.

matej-g:
My pleasure, thanks you both for helping me move it over the finish line!

Issue #1825, opened 2023-06-14
LokiReceiver ports are not automatically exposed


fbuetler:
## Description

Analogously to `OTLPReceiverParser`, we need a `LokiReceiverParser` for the newly added `lokireceiver` in `OpenTelemetry Collector Contrib` [v0.76.3](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.76.3) to automatically add the ports to the Container, Service and Ingress.

The `lokireceiver` exposes a `http` and a `grpc` endpoint like `otlpreceiver`.

Current workaround is to add them to `Spec.Ports` of the `OpenTelemetryCollector` k8s custom resource as described in its [description](https://github.com/open-telemetry/opentelemetry-operator/blob/v0.79.0/config/crd/bases/opentelemetry.io_opentelemetrycollectors.yaml#L1852-L1857):
> Ports allows a set of ports to be exposed by the underlying
  v1.Service. By default, the operator will attempt to infer the required
  ports by parsing the .Spec.Config property but this property can
  be used to open additional ports that can't be inferred by the operator,
  like for custom receivers.

## Version

v0.79.0

## Looking over the edge

Alternatively, we could add, analogously to `GenericReceiver`, a receiver that can handle multiple protocols with multiple default ports. Overall, generalizing the `OTLPReceiverParser` as I think there might be multiple receivers exposing both a grpc and http endpoint. Example for this are `loki`, `jeager` and `skywalking`. For `jaeger` there is already `JaegerReceiverParser` (with some [additional protocols](https://github.com/open-telemetry/opentelemetry-operator/blob/v0.79.0/pkg/collector/parser/receiver_jaeger.go#L68-L89)) and for `skywalking` there is already a PR (https://github.com/open-telemetry/opentelemetry-operator/pull/1650) essentially duplicating the `OTLPReceiverParser`.

I volunteer to add `LokiReceiverParser` and to implement a "generic multiple protocols receiver", if the proposal from above is accepted.

jaronoff97:
@fbuetler would you be able to work on this?

yuriolisa:
@fbuetler, are you still willing to implement that?

PR #2437, opened 2023-12-13
add loki parser to parse servicePort from config

fbuetler:

Add `LokiReceiverParser` to parse `servicePort` from config.
By default, only the Otel ports are extracted, parsed, then configured into the collector container's list of service ports.
This feature allows extracting loki ports (grpc:9095/http:3100, see [default values](https://github.com/grafana/loki/blob/main/operator/internal/manifests/var.go#L20-L22)) into the container's port list.

Resolves https://github.com/open-telemetry/opentelemetry-operator/issues/1825


internal/manifests/collector/parser/receiver/receiver_loki.go
```diff
@@ -0,0 +1,131 @@
...
+			defaultPorts: []corev1.ServicePort{
+				{
+					Name:        naming.PortName(fmt.Sprintf("%s-http", o.name), defaultLokiHTTPPort),
+					Port:        defaultLokiHTTPPort,
+					TargetPort:  intstr.FromInt(int(defaultLokiHTTPPort)),
+					AppProtocol: &http,
+				},
+			},
+		},
+	} {
```
jaronoff97, line 90: the formatting here is a bit odd... why have another `{`?

jaronoff97 (approved):

internal/manifests/collector/parser/receiver/receiver_loki.go
jaronoff97: ah i see, yeah i usually have only seen this in the test blocks so my brain didn't parse that correctly. Thanks for explaining!

pavolloffay (approved):

PR #2438, opened 2023-12-13
[Chore] Fix targetallocator-features test case to work on OpenShift.

IshwarKanse:
The PR fixes the targetallocator-features test case to work on OpenShift by adding the required securityContext configurations. Note the namespace annotations openshift.io/sa.scc.uid-range=1000/1000 openshift.io/sa.scc.supplemental-groups=3000/1000 has no effect in Kubernetes


pavolloffay:
CI failed, I have restarted it

swiatekm-sumo (approved):

pavolloffay (approved):

PR #2444, opened 2023-12-15
Use typed config in v1alpha2

pavolloffay:

I have as well tried suggestions from https://github.com/kubernetes-sigs/controller-tools/issues/636#issuecomment-1173565102 

```go
runtime.RawExtension
```

```go
        // +kubebuilder:validation:Schemaless
	Spec interface{} `json:"spec"`
```

But nothing worked better than byte array

Resolves https://github.com/open-telemetry/opentelemetry-operator/issues/1707 


apis/v1alpha2/config.go
```diff
@@ -0,0 +1,69 @@
...
+			out[key] = append([]string(nil), val...)
+		default:
+			out[key] = val
+		}
+	}
+	return &out
+}
+
+// Config encapsulates collector config.
+type Config struct {
```
pavolloffay, line 43: The struct will have methods to simplify parsing - e.g. extract enabled component, extract pipeline name etc.

jaronoff97: i was wondering if we could match what the collector's config looks like right prior to the naming of various components. This would mean having explicit fields set for telemetry, pipelines, extensions, etc. ([see here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/config.go#L15-L24)).

Something like this:
```
type Config struct {
	Receivers []receiverConfig `mapstructure:"receivers"`
	Processors []processorConfig `mapstructure:"processors"`
	Exporters []exporterConfig `mapstructure:"exporters"`
	Connectors []connectorConfig `mapstructure:"connectors"`
	Telemetry telemetryConfig `mapstructure:"telemetry"`
	Extensions extensionsConfig `mapstructure:"extensions"`
	Pipelines pipelinesConfig `mapstructure:"pipelines"`
}
```

which would let us know the type of each component when doing logic like the port parsing

swiatekm-sumo: On one hand, the more structured we can make this config, the more confident we can be about rejecting invalid ones, which is a major win for users.

The potential negative is that we'll be heavily coupled to the collector at the schema level. A breaking change in the collector config schema would require us to do a corresponding breaking change in the CRD. Maybe that's ok though? Realistically, we'd need to do something like that regardless.

It would also prevent us from easily supporting alternative syntax which the collector may add without breaking changes, like https://github.com/open-telemetry/opentelemetry-collector/issues/9077.

pavolloffay: I like @jaronoff97's proposal. I was thinking about the same, but I wanted to do this gradually. I will play with it and add it to the current PR.

pavolloffay: > A breaking change in the collector config schema would require us to do a corresponding breaking change in the CRD

In the operator, we are trying to resolve breaking changes of the collector anyways. I believe that a strongly typed config will enable us to do.

jaronoff97: agreed with pavol. I think if / when the collector does those types of syntax changes we should have first-party support for the version that includes said changes :)

iblancasa (approved):

apis/v1alpha2/collector_webhook_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestEmptyObjectWarnings(t *testing.T) {
+	webhook := CollectorWebhook{}
```
jaronoff97, line 25: nit: could we do list tests like how we do in the other webhook tests? This will make it easier in the future to add more tests for these warnings.

apis/v1alpha2/config.go
```diff
@@ -0,0 +1,147 @@
...
+		if nulls := hasNullValue(c.Connectors.Object); len(nulls) > 0 {
+			nullKeys = append(nullKeys, addPrefix("connectors.", nulls)...)
+		}
+	}
+	// Make the return deterministic. The config uses maps therefore processing order is non-deterministic.
+	sort.Strings(nullKeys)
+	return nullKeys
+}
+
+func hasNullValue(cfg map[string]interface{}) []string {
```
jaronoff97, line 124: this is great – in the future in the defaulting webhook we could set the empty map using this as well :)

jaronoff97, line 124: nit: should we just have this work on the `AnyConfig` type?

apis/v1alpha2/collector_webhook_test.go
jaronoff97: this is very stylistic so don't feel obligated to do this.

jaronoff97 (commented):
a few small suggestions, the only one I think may be useful is just using the AnyConfig type

apis/v1alpha2/config.go
pavolloffay: here is a recursive call on line 131 if the value is map, so it would not work.

apis/v1alpha2/collector_webhook_test.go
```diff
@@ -0,0 +1,68 @@
...
+							Object: map[string]interface{}{
+								"foo": nil,
+							},
+						},
+					},
+				},
+			},
+
+			warnings: []string{
+				"Collector config spec.config has null objects: extensions.foo:, processors.batch:, processors.foo:. For compatibility tooling (kustomize and kubectl edit) it is recommended to use empty obejects e.g. batch: {}.",
```
jaronoff97, line 52: this is awesome :D

apis/v1alpha2/config.go
jaronoff97: Given that AnyConfig is just map[string]interface{}, wouldn't we just change 131 to work on `v.(AnyConfig)`

jaronoff97: regardless, i don't think this is blocking so it's fine as is ✅

jaronoff97 (approved):

justurbo:
When will OpenTelemetryCollector be available with opentelemetry.io/v1alpha~~1~~2?

Using Terraform `kubernetes_manifest`, operator [official setup](https://opentelemetry.io/docs/kubernetes/operator/automatic/), the kubernetes yaml file fails because `spec.config` is string.

PR #2450, opened 2023-12-18
Fix Ingress for HTTP OTLP receiver

### LINKED ISSUE #2449: Ingress created to expose the collector is not correct for the HTTP receiver
### Component(s)

operator

### What happened?

The Ingress object created to expose the Collector outside the cluster is not
created properly for the HTTP receiver.

Environment:
- Kind cluster: 1.24
- Ingress controller: [v1.0.1](https://raw.githubusercontent.com/kubernetes/ingress-nginx/controller-v1.0.1/deploy/static/provider/kind/deploy.yaml)

When I create this OpenTelemetry Collector instance:
```yaml
kind: OpenTelemetryCollector
metadata:
  name: simplest
spec:
  mode: "deployment"
  ingress:
    type: ingress
  config: |
    receivers:
      otlp:
        protocols:
          grpc:
          http:
    exporters:
      debug:

    service:
      pipelines:
        traces:
          receivers: [otlp]
          processors: []
          exporters: [debug]
```

This is the created Ingress instance:
```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  creationTimestamp: "2023-12-18T11:04:39Z"
  generation: 1
  labels:
    app.kubernetes.io/instance: demo.simplest
    app.kubernetes.io/name: simplest-ingress
  name: simplest-ingress
  namespace: demo
  ownerReferences:
  - apiVersion: opentelemetry.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: OpenTelemetryCollector
    name: simplest
    uid: 540cf5f8-29ec-4d8e-94b5-cf89749e8c9e
  resourceVersion: "1038"
  uid: 6d2f105f-8d56-4967-86db-e8c62f291cd3
spec:
  rules:
  - http:
      paths:
      - backend:
          service:
            name: simplest-collector
            port:
              name: otlp-grpc
        path: /otlp-grpc
        pathType: Prefix
      - backend:
          service:
            name: simplest-collector
            port:
              name: otlp-http
        path: /otlp-http
        pathType: Prefix
status:
  loadBalancer: {}
```

If I run `curl` to the exposed URL:
```sh
$ curl http://localhost/otlp-http/v1/traces
404 page not found
```
If I try to send traces from a Python application, I can see this error message:
```sh
Failed to export batch code: 404, reason: 404 page not found
```

If I do port forwarding to the 4318 port, and run `curl` for the same subpath:
```sh
$ curl http://localhost:4318/v1/traces
405 method not allowed, supported: [POST]
```

### Kubernetes Version

1.24

### Operator version

c8b2970

### Collector version

0.90.0

### Environment information

_No response_

### Log output

_No response_

### Additional context

_No response_

iblancasa:
Fixes #2449

### DIFF
--- a/.chloggen/2449-fix-ingress.yaml
+++ b/.chloggen/2449-fix-ingress.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix the path and annotations needed for the Ingress to expose the OTLP HTTP receiver.
+
+# One or more tracking issues related to the change
+issues: [2449]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/internal/manifests/collector/ingress.go
+++ b/internal/manifests/collector/ingress.go
@@ -53,11 +53,20 @@ func Ingress(params manifests.Params) (*networkingv1.Ingress, error) {
 		rules = createSubdomainIngressRules(params.OtelCol.Name, params.OtelCol.Spec.Ingress.Hostname, ports)
 	}
 
+	ingressAnnotations := params.OtelCol.Spec.Ingress.Annotations
+	if ingressAnnotations == nil {
+		ingressAnnotations = make(map[string]string)
+	}
+
+	if _, ok := ingressAnnotations["nginx.ingress.kubernetes.io/rewrite-target"]; !ok {
+		ingressAnnotations["nginx.ingress.kubernetes.io/rewrite-target"] = "/$2"
+	}
+
 	return &networkingv1.Ingress{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        naming.Ingress(params.OtelCol.Name),
 			Namespace:   params.OtelCol.Namespace,
-			Annotations: params.OtelCol.Spec.Ingress.Annotations,
+			Annotations: ingressAnnotations,
 			Labels: map[string]string{
 				"app.kubernetes.io/name":       naming.Ingress(params.OtelCol.Name),
 				"app.kubernetes.io/instance":   fmt.Sprintf("%s.%s", params.OtelCol.Namespace, params.OtelCol.Name),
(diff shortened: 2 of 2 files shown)


.chloggen/2449-fix-ingress.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
```
pavolloffay, line 2: This is a correct behaviour of the operator. 

Automatically putting ingress specific annotation should be a new feature

internal/manifests/collector/ingress.go
```diff
@@ -78,7 +87,7 @@ func createPathIngressRules(otelcol string, hostname string, ports []corev1.Serv
 	for i, port := range ports {
 		portName := naming.PortName(port.Name, port.Port)
 		paths[i] = networkingv1.HTTPIngressPath{
-			Path:     "/" + port.Name,
+			Path:     "/" + port.Name + "(/|$)(.*)",
```
pavolloffay, line 90: Can you explain why the path changes? 

This will also result in a breaking change.

.chloggen/2449-fix-ingress.yaml
iblancasa: I don't think so. Because the Ingress doesn't work otherwise.

internal/manifests/collector/ingress.go
iblancasa: The traces, for instance, are reported to `<<URL>>/v1/traces`. Without this, you can only access to `<<URL>>`. Everything you send to `<URL>/<subpath>` will be redirected to `<URL>`. Exposing the collector doesn't work right now.

.chloggen/2449-fix-ingress.yaml
iblancasa: Take a look to #2449

pavolloffay: The assumption was that users setup URL rewrite annotations

iblancasa: Where is that documented? Because I was not able to find anything related to that.

iblancasa: OK. I can revert the annotation, but we still need the other change.

internal/manifests/collector/ingress.go
jaronoff97: this fix will only work for nginx right? for anything using a default ingress provider this will have no effect? @pavolloffay this technically isn't a breaking change because we choose prefix currently, so anything after `/<port-name>` is already accepted and this regex doesn't change that because the current options are captured within the regex. ex:
```
/<port-name> is matched by /<port-name>(/|$)(.*)
/<port-name>/somethingelse is matched by /<port-name>(/|$)(.*)
```

.chloggen/2449-fix-ingress.yaml
Starefossen: The current implementation is broken as enabling ingress results in 404 for all paths.

Issue #2454, opened 2023-12-18
Add version label to target allocator resources


swiatekm-sumo:
### Component(s)

operator

### Is your feature request related to a problem? Please describe.

Collector resources get a `app.kubernetes.io/version` label, but target allocator resources don't. We should be consistent about this.

### Describe the solution you'd like

Add the missing label and ideally use `manifestutils.Labels` in the implementation.

swiatekm-sumo:
Resolved in #2455

PR #2455, opened 2023-12-18
Add version label to target allocator resources

swiatekm-sumo:
**Description:**
Add `app.kubernetes.io/version` label to all target allocator resources. Where the label was present, it now uses the target allocator version instead of the collector version. Use `manifestutils.Labels` as the implementation.

I've also used `manifestutils.SelectorLabels` where it makes sense.

**Link to tracking Issue:** #2454 

**Testing:**
Modified existing tests.


internal/manifests/targetallocator/labels.go
```diff
@@ -16,27 +16,15 @@ package targetallocator
...
-
-	base["app.kubernetes.io/managed-by"] = "opentelemetry-operator"
-	base["app.kubernetes.io/instance"] = naming.Truncate("%s.%s", 63, instance.Namespace, instance.Name)
-	base["app.kubernetes.io/part-of"] = "opentelemetry"
-	base["app.kubernetes.io/component"] = "opentelemetry-targetallocator"
-
-	if _, ok := base["app.kubernetes.io/name"]; !ok {
-		base["app.kubernetes.io/name"] = name
-	}
+	return manifestutils.Labels(instance.ObjectMeta, name, instance.Spec.TargetAllocator.Image, ComponentOpenTelemetryTargetAllocator, nil)
```
jaronoff97, line 24: this is awesome

internal/manifests/targetallocator/service.go
```diff
@@ -27,7 +27,7 @@ func Service(params manifests.Params) *corev1.Service {
 	name := naming.TAService(params.OtelCol.Name)
 	labels := Labels(params.OtelCol, name)
 
-	selector := Labels(params.OtelCol, name)
+	selector := SelectorLabels(params.OtelCol)
```
jaronoff97, line 30: to confirm, we should expect no changes for the selector labels right?

jaronoff97 (approved):

internal/manifests/targetallocator/service.go
swiatekm-sumo: There is one change - we no longer have `app.kubernetes.io/name` in there, same as for the collector. Think that's worth calling out?

jaronoff97: ah yeah its probably because we add it as a label on the deployment, just not for the selector (see [here](https://github.com/open-telemetry/opentelemetry-operator/blob/879c8469b09bc83b94c1ecc8b6bf724a47fecc3f/internal/manifests/manifestutils/labels.go#L68)) which seems fine to me. Let's call it out in the changelog notes to make it clear

internal/manifests/targetallocator/deployment.go
```diff
@@ -44,7 +44,7 @@ func Deployment(params manifests.Params) (*appsv1.Deployment, error) {
 		Spec: appsv1.DeploymentSpec{
 			Replicas: params.OtelCol.Spec.TargetAllocator.Replicas,
 			Selector: &metav1.LabelSelector{
-				MatchLabels: labels,
+				MatchLabels: SelectorLabels(params.OtelCol),
```
pavolloffay, line 47: Isn't this immutable field? https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates 

Can you confirm that this will not break during the upgrade?

swiatekm-sumo: I'll test it. If there's a problem with that, then it's probably for the best to leave the selector as is.

jaronoff97: @pavolloffay we do have logic in our reconciler to just delete and recreate when it sees this immutable change, so it should just work (with a small blip for the reconciler logic)

swiatekm-sumo: Deleting and recreating resources is a bit heavyhanded though, and I'm not sure if we want to do it just to have uniform selector labels between the collector and target allocator.

swiatekm-sumo: I brought this label back. I don't think it makes the code substantially messier - check it our for yourselves.

jaronoff97: fair enough! Maybe lets keep it as it is then, and leave a comment explaining the divergence between the two.

jaronoff97 (approved):
looks good to me, thanks for leaving the comment as well :)

pavolloffay (approved):

Issue #2422, opened 2023-12-06
Introduce TargetAllocator CRD


swiatekm:
### Component(s)

target allocator

### Describe the issue you're reporting

The Target Allocator subresource of the OpenTelemetryCollector CR has been growing in size over time, in particular due to various infrastructure-related attributes that users rightfully want to set for it. It’s time to split it into its own CRD, improving separation of concerns and reducing the size of the Collector CRD.

## Design

The TargetAllocator CRD should contain all the fields the Target Allocator subresource currently does, to begin with. We will later continue adding fields to it, while possibly deprecating some fields from the subresource. The subresource should always only contain a subset of the CRD fields. Internally, the operator will create a CRD based on the subresource, which is how we’ll maintain backwards compatibility.

```yaml
apiVersion: opentelemetry.io/v1alpha1
kind: TargetAllocator
metadata:
  name: example
spec:
  scrapeConfigs:
    - job_name: otel-collector
      scrape_interval: 10s
      static_configs:
      - targets:
        - 0.0.0.0:8888
        - 0.0.0.0:9999
  prometheusCR:
    enabled: true
  ..
```

### Relation to the Collector CR

A Collector can be configured to use a given target allocator by setting a label on the collector CR: 

```yaml
apiVersion: opentelemetry.io/v1beta1
kind: TargetAllocator
metadata:
  name: example
  labels:
    opentelemetry.io/target-allocator: example
spec:
  config:
    receivers:
      prometheus:
  ...
```
Collectors and Target Allocators are independent, neither owns the other, and each can in principle run on their own. By establishing a connection between them, certain aspects of their configuration will change under the hood.

## Rollout

We should be able to roll out most of the necessary changes without any user-facing change:

1. Add the type definitions for the TargetAllocator CRD
2. Generate the TargetAllocator CR from the subresource
3. Use the generated TargetAllocator CR to generate target allocator manifests
4. Add webhooks for TargetAllocator without enabling them
5. Enable the webhooks and reconciliation for the TargetAllocator CR

jaronoff97:
some thoughts:
* I think this should only be in v2 i.e. `apiVersion: opentelemetry.io/v1alpha2` This will let us take advantage of the work for #901 (embedding a common fields struct)
* Having this in v2 only would let us begin with some fields in the TA spec deprecated or removed
[...]

PR #2466, opened 2023-12-20
Use standard collector selectors in target allocator config

swiatekm-sumo:
**Description:**
Make the label selector for collectors in target allocator configuration use the K8s standard format. We can later explicitly expose this in the TargetAllocator CRD.

One annoying aspect of using the core struct is that it only has tags for json serialization, and the yaml marshaler just lowercases the name. So we end up with `matchlabels` rather than `matchLabels` or `match_labels`. I'm not sure the additional complexity of using a custom struct and then converting is worth it, given that target allocator is almost exclusively used and configured by the operator.

**Link to tracking Issue:** #2422

Resolves https://github.com/open-telemetry/opentelemetry-operator/issues/2422 

**Testing:**
Modified existing tests.

### DIFF
--- a/.chloggen/feat_targetallocator-selector.yaml
+++ b/.chloggen/feat_targetallocator-selector.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: target allocator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Use standard K8s label selectors for collectors in target allocator config
+
+# One or more tracking issues related to the change
+issues: [2422]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: This is a breaking change only for users of standalone target allocator. Operator users are unaffected.
--- a/.chloggen/feat_targetallocator-selector.yaml
+++ b/.chloggen/feat_targetallocator-selector.yaml
@@ -13,4 +13,6 @@ issues: [2422]
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
-subtext: This is a breaking change only for users of standalone target allocator. Operator users are unaffected.
+subtext: |
+  This is a breaking change only for users of standalone target allocator. Operator users are unaffected.
+  The operator is still compatible with previous target allocator versions, and will be for the next 3 releases.
--- a/cmd/otel-allocator/config/config.go
+++ b/cmd/otel-allocator/config/config.go
@@ -40,17 +40,17 @@ const DefaultConfigFilePath string = "/conf/targetallocator.yaml"
 const DefaultCRScrapeInterval model.Duration = model.Duration(time.Second * 30)
 
 type Config struct {
-	ListenAddr             string               `yaml:"listen_addr,omitempty"`
-	KubeConfigFilePath     string               `yaml:"kube_config_file_path,omitempty"`
-	ClusterConfig          *rest.Config         `yaml:"-"`
-	RootLogger             logr.Logger          `yaml:"-"`
-	CollectorSelector      metav1.LabelSelector `yaml:"collector_selector,omitempty"`
-	PromConfig             *promconfig.Config   `yaml:"config"`
-	AllocationStrategy     *string              `yaml:"allocation_strategy,omitempty"`
-	FilterStrategy         *string              `yaml:"filter_strategy,omitempty"`
-	PrometheusCR           PrometheusCRConfig   `yaml:"prometheus_cr,omitempty"`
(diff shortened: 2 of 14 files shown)


pavolloffay (approved):

.chloggen/feat_targetallocator-selector.yaml
```diff
@@ -0,0 +1,16 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Use standard K8s label selectors for collectors in target allocator config
+
+# One or more tracking issues related to the change
+issues: [2422]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: This is a breaking change only for users of standalone target allocator. Operator users are unaffected.
```
jaronoff97, line 16: this is acceptable IMO

cmd/otel-allocator/config/config.go
```diff
@@ -39,17 +40,17 @@ const DefaultConfigFilePath string = "/conf/targetallocator.yaml"
...
-	AllocationStrategy     *string            `yaml:"allocation_strategy,omitempty"`
-	FilterStrategy         *string            `yaml:"filter_strategy,omitempty"`
-	PrometheusCR           PrometheusCRConfig `yaml:"prometheus_cr,omitempty"`
-	PodMonitorSelector     map[string]string  `yaml:"pod_monitor_selector,omitempty"`
-	ServiceMonitorSelector map[string]string  `yaml:"service_monitor_selector,omitempty"`
+	ListenAddr             string               `yaml:"listen_addr,omitempty"`
+	KubeConfigFilePath     string               `yaml:"kube_config_file_path,omitempty"`
+	ClusterConfig          *rest.Config         `yaml:"-"`
+	RootLogger             logr.Logger          `yaml:"-"`
+	CollectorSelector      metav1.LabelSelector `yaml:"collector_selector,omitempty"`
```
jaronoff97, line 47: should this be reference? or should we remove `omitempty`? I'm not sure the TA can work without the collector selector.

internal/manifests/targetallocator/configmap.go
```diff
@@ -52,7 +52,9 @@ func ConfigMap(params manifests.Params) (*corev1.ConfigMap, error) {
 
 	taConfig := make(map[interface{}]interface{})
 	prometheusCRConfig := make(map[interface{}]interface{})
-	taConfig["label_selector"] = manifestutils.SelectorLabels(params.OtelCol.ObjectMeta, collector.ComponentOpenTelemetryCollector)
+	taConfig["collector_selector"] = map[string]any{
+		"matchlabels": manifestutils.SelectorLabels(params.OtelCol.ObjectMeta, collector.ComponentOpenTelemetryCollector),
+	}
```
jaronoff97, line 57: one issue that may arise with this change is that users cannot use the newer operator with an older TA version. This could be problematic in this scenario:
* a company runs a mirrored image of the TA due to a security requirement for the image in the collector CRD
* engineer bumps the operator version which no longer writes the former selector
* TA can no longer startup until they re-mirror and bump the TA image

I think for now it may make sense to just mark the label_selector as deprecated in the operator and will be removed entirely in v1alpha2. It is totally fine, however, to just remove it entirely from the TA.

jaronoff97 (commented):
two thoughts

cmd/otel-allocator/config/config.go
swiatekm-sumo: It should be, good catch. The semantics of `v1.LabelSelector`, as per the [documentation](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#LabelSelector), are: 

> A label selector is a label query over a set of resources. The result of matchLabels and matchExpressions are ANDed. An empty label selector matches all objects. A null label selector matches no objects.

So the target allocator should in fact work without a selector, it'll behave the same way as if there were no collectors to assign targets to.

internal/manifests/targetallocator/configmap.go
swiatekm-sumo: Do we support running different versions of the operator and targetallocator? I know we don't for the collector, even though things usually work anyway.

How would you mark the label_selector as deprecated? From the operator's perspective, it's just a key in an untyped map.

jaronoff97: I would just mark it in a comment with a TODO linking to an open issue, this could also be a featuregate that starts enabled by default and then we eventually just delete it?

I'm not sure we've discussed what the operators own compat matrix looks like, but the scenario I linked above could break a lot of users because you cannot usually update the TA image and the operator image in the same release. Doing either one without the other would result in a broken state.

swiatekm-sumo: A problem with this is that target allocator currently unmarshals its configuration in strict mode, so the old version can't deal with the new value either way. The way I see it, there's two possible approaches:

1. Do an explicit version check and apply the right value.
2. Merge a change that makes config unmarshalling non-strict, wait till that's released, and then merge this PR.

I think I like 2 better, WDYT?

jaronoff97 (approved):

internal/manifests/targetallocator/configmap.go
jaronoff97: argh 🤦 I think i like 2 more as well. Lets get that merged in before #2482

jaronoff97:
@swiatekm-sumo i think we're good to merge this now that #2488 has been merged right?

jaronoff97 (approved):

PR #2472, opened 2023-12-21
feat/add shareProcessName capability

bmiguel-teixeira:
**Description:**
Adding an additional field to the OpenTelemetryCollector CRD called `shareProcessNamespace` that essentially toggles the equivalent field in the PodSpec of the generated resources (DS, Deployment or STS).

No issue opened on my side.

The reason for this addition in to allow configuration reloads. The OtelCollectors allow config reload via SIGHUP, but there is currently no way to this via the `additionalContainers` field without this capability.

**Testing:**

Unit tests added.

**Documentation:**


swiatekm-sumo:
What do you need the SIGHUP reload for, if I may ask? I'm not opposed to adding this attribute, but reloading config via a signal seems like an antipattern in Kubernetes. We already restart collector Pods whenever configuration changes,

bmiguel-teixeira:
Hi @swiatekm-sumo 

While the bulk of the configuration is indeed static and managed by the controller, a small section of it, the filter section, 

```
  processors:
    filter/md: ${file:/var/conf/extraconf/dynamic.yaml}
´´´

is dynamic in our use case and usually varies from cluster to cluster.

P.S: pushed the missing manifests change.

Happy holidays.

**Comment by bmiguel-teixeira** [comment]
Hi @swiatekm-sumo

Pushed missing changelog file. Any comments/feedback on this?

Ty

File: apis/v1alpha1/opentelemetrycollector_types.go [hunk 1]
```diff
@@ -215,6 +215,9 @@ type OpenTelemetryCollectorSpec struct {
 	// HostNetwork indicates if the pod should run in the host networking namespace.
 	// +optional
 	HostNetwork bool `json:"hostNetwork,omitempty"`
+	// ShareNetworkProcess indicates if the pod's containers should share process namespace.
+	// +optional
+	ShareNetworkProcess bool `json:"shareNetworkProcess,omitempty"`
```
swiatekm-sumo, line 220: Why is this named `ShareNetworkProcess` and not just `ShareProcessNamespace`? We try to stick to Kubernetes core attribute names unless there's a good reason to deviate.

swiatekm-sumo (commented):
Could you also set this in one of the E2E tests?

apis/v1alpha1/opentelemetrycollector_types.go
bmiguel-teixeira: honestly, have no idea why. most likely bad copy paste... let me fix it.

bmiguel-teixeira: done. and added scenario e2e for this use case. cheers

swiatekm-sumo (commented):
@bmiguel-teixeira you've got some failures in the builder unit tests. Other than that, I think we're good to go.

bmiguel-teixeira:
hi @swiatekm-sumo 

Fixed unit tests. Unit and e2e all good locally. ty

swiatekm-sumo:
@bmiguel-teixeira you still have a minor lint issue.

bmiguel-teixeira:
@swiatekm-sumo done. 🤞

swiatekm-sumo (approved):
LGTM, thanks for your patience!

@jaronoff97 we probably want to add this attribute to common fields as well.

jaronoff97:
yep, i'll open a follow up issue for that, no need to further block this.

jaronoff97 (approved):

Issue #2477, opened 2023-12-22
Target Allocator is not capable of scaling out with least-weighted strategy


utr1903:
### Component(s)

target allocator

## Description

I'm not sure whether this is a bug or an intentional implemention, however the `least-weighted` strategy of the Target allocator is not capable of scaling out in terms of re-distributing the existing targets in case of new collector addition.

## Steps to Reproduce

1. Configure 5 scrape targets and deploy 2 collectors.
2. Let the collectors get their targets assigned.
  - `collector1` -> `target1`, `target3`, `target5`
  - `collector2` -> `target2`, `target4`
3. Deploy a 3rd collector.
4. Check target distribution.

## Expected Result

Expected outcome would be re-distribution of the targets across all 3 instances.

- `collector1` -> `target1`, `target4`
- `collector2` -> `target2`, `target5`
- `collector3` -> `target3`

## Actual Result

No re-distribution is triggered and the 3rd collector does not get assigned any target.

- `collector1` -> `target1`, `target3`, `target5`
- `collector2` -> `target2`, `target4`
- `collector3` -> ` `

### Kubernetes Version

1.27.1

### Operator version

0.90.0

### Collector version

0.91.0

### Implementation analysis

I've checked the code and if I'm not mistaken scaling out is not possible with the current implementation (please correct me if I'm wrong).

When newly added collector instances are found, it is detected [here](https://github.com/open-telemetry/opentelemetry-operator/blob/v0.90.0/cmd/otel-allocator/main.go#L175) and the [`SetCollectors()`](https://github.com/open-telemetry/opentelemetry-operator/blob/v0.90.0/cmd/otel-allocator/allocation/least_weighted.go#L272) function of the allocator is executed. It checks the differences of collectors between current and new states, finds a new one (`addition`) and therefore runs [`handleCollectors`](https://github.com/open-telemetry/opentelemetry-operator/blob/v0.90.0/cmd/otel-allocator/allocation/least_weighted.go#L179).

[...]

PR #2478, opened 2023-12-22
Add test for adding new collector after completed initial allocation

utr1903:
**Description:**
As mentioned in the linked issue, the `least-weighted` allocation strategy is at the moment not capable of scaling out. Shortly explained, after the targets are distributed across the initial set of collector instances, the newly added collector instance will not be assigned any targets.

This added test ensures that the current capability is covered. In case it will be decided in the future to change the implementation in order to achieve scaling capability, this test will fail and will thereby signal for a new test to be created.

**Link to tracking Issue:**
[2477](https://github.com/open-telemetry/opentelemetry-operator/issues/2477)

**Testing:**
The test `allocation/TestNoAssignmentToNewCollector` is added.

**Documentation:**
The test assures that the newly added collectors are not assigned any targets after a successful target distribution across the initially detected collectors.


jaronoff97 (approved):

utr1903:
I think this simple PR does not require any addition to change log. Could we add the `Skip Changelog` label if you agree as well?

cmd/otel-allocator/allocation/allocatortest.go
```diff
@@ -23,14 +23,27 @@ import (
...
+		label := model.LabelSet{
+			"i":     model.LabelValue(strconv.Itoa(i)),
+			"total": model.LabelValue(strconv.Itoa(n + startingIndex)),
+		}
+		newTarget := target.NewItem(fmt.Sprintf("test-job-%d", i), fmt.Sprintf("test-url-%d", i), label, "")
+		toReturn[newTarget.Hash()] = newTarget
+	}
+	return toReturn
+}
+
```
swiatekm-sumo, line 38: Is this necessary? The "collector" label isn't currently used in allocation tests, so you can simply delete it from `MakeNNewTargets` instead of creating a new function.

utr1903: @swiatekm-sumo Thanks for your review! I haven't checked all of the tests but I saw that `MakeNNewTargets` was being used also in `consistent-hashing`. So instead of breaking something that I haven't noticed, I created a new function.

Do you want me to completely override the old one with the new one? Or shall we just keep it as it is (assuming a test which considers the collector label will be implemented in the future anyway)?

swiatekm-sumo: I'd prefer is you modified the existing function. Since the label isn't used for anything, keeping it around is just technical debt that causes confusion for no benefit.

utr1903: Then let me revert all the changes regarding adding new function because what I was willing to have was to initialize targets without assigning the collectors (just like in the actual `least_weighted` implementation):

```
newTarget := target.NewItem(<jobName>, <targetUrl>, <label>, "")
```

Yet, now I see that it actually doesn't have an impact at all... I will only remove the collector label part :)

swiatekm-sumo (approved):

PR #2479, opened 2023-12-23
Use kustomize for e2e test setup

swiatekm-sumo:
**Description:**
Right now, we have a very ad-hoc and inconsistent way of doing setup for e2e tests:
* The operator image is set via kustomize
* target allocator and opamp bridge images are set by doing a text replace over the test manifests
* Feature gates for the operator are set by redeploying the operator in kuttl commands, in the test definition itself

This change introduces a single mechanism for e2e test setup. All the per-testsuite setup is done via kustomize, primarily through json patches on the operator args - both images and feature gates can be set this way.

As a result, we have:
* All test setup for our E2E tests is done via the `prepare-e2e` make target, or its flavors
* It is now straightforward to run different test suites with specific feature flags enabled or disabled
* Tests themselves do no setup, and external consumers like the Helm Chart can easily do their own
* We get rid of the hacky script for text-replacing images in manifests
* All the local changes are done to the `config/manager/kustomization.yaml` file, which makes them easy to control

**Testing:**
Went through a bunch of typical dev workflows manually.


config/manager/manager.yaml
```diff
@@ -28,7 +28,11 @@ spec:
     spec:
       containers:
       - args:
-        - --enable-leader-election
+        - "--metrics-addr=127.0.0.1:8080"
+        - "--enable-leader-election"
+        - "--zap-log-level=info"
+        - "--zap-time-encoding=rfc3339nano"
+        - "--feature-gates=+operator.autoinstrumentation.go,+operator.autoinstrumentation.nginx"
```
swiatekm-sumo, line 35: I moved this so I could patch these args in the manager kustomization.yaml, as opposed to the one in config/default. I don't see any reason for it not to be here.

.github/workflows/e2e.yaml
```diff
@@ -21,17 +21,22 @@ jobs:
...
          - e2e-instrumentation
          - e2e-upgrade
-         - e2e-prometheuscr
          - e2e-autoscale
-         - e2e-multi-instrumentation
          - e2e-pdb
          - e2e-opampbridge
+         - e2e-prometheuscr
+         - e2e-multi-instrumentation
+       include:
```
jaronoff97, line 35: This is very slick. I like it :)

jaronoff97 (approved):

jaronoff97:
This LGTM. The operator helm chart also runs these [E2E tests](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/.github/workflows/operator-test.yaml). Do we need to make any changes there?

swiatekm-sumo:
> This LGTM. The operator helm chart also runs these [E2E tests](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/.github/workflows/operator-test.yaml). Do we need to make any changes there?

Test run by the Helm Chart should work as they are. This will let us simplify their setup, which can happen in a later change.

yuriolisa (approved):

PR #2485, opened 2024-01-03
Changed default allocation & filtering strategy

### LINKED ISSUE #2477: Target Allocator is not capable of scaling out with least-weighted strategy

utr1903:
**Description:**
As discussed in the linked issue:
- default value for allocation strategy is changed from `least-weighted` to `consistent-hashing`
- default value for filtering strategy is changed from `none` to `relabel-config`

**Link to tracking Issue:**
closes #2477

**Testing:**
None required.

**Documentation:**
None required.


jaronoff97 (commented):
we'll also need to update the defaults [here](https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/manifests/targetallocator/configmap.go#L53-L61)

swiatekm-sumo:
> we'll also need to update the defaults [here](https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/manifests/targetallocator/configmap.go#L53-L61)

While we're at it, let's set them in the CRDs as well. Like so:
```go
	// AllocationStrategy determines which strategy the target allocator should use for allocation.
	// The current options are least-weighted and consistent-hashing. The default option is least-weighted
	// +optional
	// +kubebuilder:default:="least-weighted"
	AllocationStrategy OpenTelemetryTargetAllocatorAllocationStrategy `json:"allocationStrategy,omitempty"`
```

utr1903:
On it :)

utr1903:
Would these changes be OK regarding the allocation strategy?

If yes, shall I do the exact same structure for the filtering strategy? We can:
- convert it from to string to enum
- add it's own file as in `allocation_strategy.go` -> `// +kubebuilder:validation:Enum="";relabel-config`
- set default value to `relabel-config` -> `// +kubebuilder:default:=relabel-config`

jaronoff97 (approved):

jaronoff97:
@utr1903 Yes! that sounds great. I think i would maybe rename `allocation_strategy.go` to something like `allocator_constants.go` or something like that, otherwise sounds good :D

internal/manifests/targetallocator/configmap.go
```diff
@@ -53,7 +53,7 @@ func ConfigMap(params manifests.Params) (*corev1.ConfigMap, error) {
 	if len(params.OtelCol.Spec.TargetAllocator.AllocationStrategy) > 0 {
 		taConfig["allocation_strategy"] = params.OtelCol.Spec.TargetAllocator.AllocationStrategy
 	} else {
-		taConfig["allocation_strategy"] = v1alpha1.OpenTelemetryTargetAllocatorAllocationStrategyLeastWeighted
+		taConfig["allocation_strategy"] = v1alpha1.OpenTelemetryTargetAllocatorAllocationStrategyConsistentHashing
 	}
```
swiatekm-sumo, line 57: We've set the default on the CRD, so here we can unconditionally do:

```go
taConfig["allocation_strategy"] = params.OtelCol.Spec.TargetAllocator.AllocationStrategy
```

swiatekm-sumo (approved):
One minor comment, LGTM otherwise. :+1:

jaronoff97:
@utr1903 looks like the hashes in the config are wrong, otherwise i think this is good to merge.

Issue #2490, opened 2024-01-04
SecurityContext vs PodSecurityContext in the target allocator CR


jaronoff97:
### Component(s)

operator

## Description
Right now, the `securityContext` for the TA is using the `PodSecurityContext`. This naming creates a problem. The PodSecurityContext object has [these fields](https://github.com/kubernetes/api/blob/v0.29.0/core/v1/types.go#L4027). The SecurityContext object has [these fields.](https://github.com/kubernetes/api/blob/v0.29.0/core/v1/types.go#L7012) Unfortunately, there is a difference in these fields, here's the distinct set:

PODSECURITYCONTEXT
```
SupplementalGroups []int64 `json:"supplementalGroups,omitempty" protobuf:"varint,4,rep,name=supplementalGroups"`
FSGroup *int64 `json:"fsGroup,omitempty" protobuf:"varint,5,opt,name=fsGroup"`
Sysctls []Sysctl `json:"sysctls,omitempty" protobuf:"bytes,7,rep,name=sysctls"`
FSGroupChangePolicy *PodFSGroupChangePolicy `json:"fsGroupChangePolicy,omitempty" protobuf:"bytes,9,opt,name=fsGroupChangePolicy"`
```

SECURITYCONTEXT
```
Capabilities *Capabilities `json:"capabilities,omitempty" protobuf:"bytes,1,opt,name=capabilities"`
Privileged *bool `json:"privileged,omitempty" protobuf:"varint,2,opt,name=privileged"`
ReadOnlyRootFilesystem *bool `json:"readOnlyRootFilesystem,omitempty" protobuf:"varint,6,opt,name=readOnlyRootFilesystem"`
AllowPrivilegeEscalation *bool `json:"allowPrivilegeEscalation,omitempty" protobuf:"varint,7,opt,name=allowPrivilegeEscalation"`
ProcMount *ProcMountType `json:"procMount,omitempty" protobuf:"bytes,9,opt,name=procMount"`
```

Although there is a difference here, I would argue that the current bug is that our fields are misnamed and this breaking change is actually a bug fix. I believe we should adjust the TA to embed the fields the same way the [collector does](https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1alpha1/opentelemetrycollector_types.go#L152C38-L152C38) in the current version.

### Kubernetes Version

n/a

### Operator version

n/a

### Collector version

n/a

### Environment information

n/a

### Log output

```shell
n/a
```

[...]

PR #2492, opened 2024-01-08
Use container securityContext for target allocator

swiatekm-sumo:
**Description:**
The target allocator CR `securityContext` field is currently the Pod Security Context. This change makes it the container security context, as originally intended. 

**Link to tracking Issue:** #2490 

**Testing:**
Added a unit test and modified an E2E test to specifically check fields unique to the container security context.

### DIFF
--- a/.chloggen/fix-ta-security-context.yaml
+++ b/.chloggen/fix-ta-security-context.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: target allocator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Use container securityContext for target allocator
+
+# One or more tracking issues related to the change
+issues: [2490]
+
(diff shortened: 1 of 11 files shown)


.chloggen/fix-ta-security-context.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
```
jaronoff97, line 2: let's call this out as a breaking change and note in the subtext why this is also a bug fix.

internal/manifests/targetallocator/deployment.go
```diff
@@ -59,7 +59,6 @@ func Deployment(params manifests.Params) (*appsv1.Deployment, error) {
 					Tolerations:               params.OtelCol.Spec.TargetAllocator.Tolerations,
 					TopologySpreadConstraints: params.OtelCol.Spec.TargetAllocator.TopologySpreadConstraints,
 					Affinity:                  params.OtelCol.Spec.TargetAllocator.Affinity,
-					SecurityContext:           params.OtelCol.Spec.TargetAllocator.SecurityContext,
```
jaronoff97, line 62: note: we should bring back podsecuritycontext in a separate PR

jaronoff97 (approved):

PR #2493, opened 2024-01-08
Common Fields for V2 Types

### LINKED ISSUE #901: Embed a PodTemplate or Container[] list in the OpentelemetryCollector CRD
The `OpentelemetryCollector` CRD currently defines a few configuration options for env-var injection from `ConfigMap` or `Secret` resources and a couple of other options. But it does not expose most of the configuration of a `Pod`: You cannot control

* the `Pod`'s annotations (https://github.com/open-telemetry/opentelemetry-operator/issues/900)
* its resource limits and ranges (https://github.com/open-telemetry/opentelemetry-operator/issues/895)
* details of the storage configuration (https://github.com/open-telemetry/opentelemetry-operator/issues/897)
* `imagePullSecrets` (https://github.com/open-telemetry/opentelemetry-operator/issues/846)
* liveness probe settings (https://github.com/open-telemetry/opentelemetry-operator/issues/760)
* `readinessProbe`
* ...

These issues share a common underlying problem: The operator does not use a `PodTemplate` in the `OpentelemetryCollector` resource. It instead constructs the `Deployment` and embedded `PodTemplate` `spec.template` entirely using individual configuration settings.

That won't scale or be maintainable. What about security policies? What about ... anything you can name that can appear in a pod?

The future-resistant solution is to add an optional `PodTemplate` to the `OpentelemetryCollector` CRD, and deprecate the current keys `image`, `serviceAccount`, `env` etc.

jaronoff97:
**Description:**
Simplifies the shared fields between the CRDs.

**Link to tracking Issue:** resolves #901 

**Testing:** unit tests

**Documentation:** n/a

### DIFF
--- a/apis/v1alpha2/common.go
+++ b/apis/v1alpha2/common.go
@@ -1,0 +1,149 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1alpha2
+
+import (
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
+)
+
+type OpenTelemetryCommonFields struct {
+	// ManagementState defines if the CR should be managed by the operator or not.
+	// Default is managed.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default:=managed
+	ManagementState v1alpha1.ManagementStateType `json:"managementState,omitempty"`
+	// Resources to set on the OpenTelemetry Collector pods.
+	// +optional
+	Resources v1.ResourceRequirements `json:"resources,omitempty"`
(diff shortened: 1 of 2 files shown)


swiatekm-sumo (approved):

.chloggen/common-fields-v2.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Introduce common fields to the v2 types that can be reused for any CRDs.
```
pavolloffay, line 8: could we please change the v2 to v1alpha2?

apis/v1alpha2/common.go
```diff
@@ -0,0 +1,159 @@
...
+)
+
+type OpenTelemetryCommonFields struct {
+	// ManagementState defines if the CR should be managed by the operator or not.
+	// Default is managed.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default:=managed
+	ManagementState v1alpha1.ManagementStateType `json:"managementState,omitempty"`
```
pavolloffay, line 30: The type should be probably defined in this package rather than importing it from the previous version

.chloggen/common-fields-v2.yaml
jaronoff97: yep!

apis/v1alpha2/common.go
jaronoff97: sounds good, i'll duplicate these over.

pavolloffay (approved):

Issue #2489, opened 2024-01-04
Allow the bridge to report on individual collector pods


jaronoff97:
### Component(s)

opamp bridge

### Is your feature request related to a problem? Please describe.

Right now the bridge is only reporting health on the overall collector CRDs, but it would be great if the bridge could healthcheck against the running collectors when it sends the component health message. This would allow a consumer to know what pods are currently running for a collector pool.

### Describe the solution you'd like

Optionally enable the bridge to health check collector pods by using the collector selector in the CRD. Right now we can use

jaronoff97:
If we do this, we should also use the informer pattern rather than a ticker like we do today

PR #2496, opened 2024-01-08
Show pods in component health message

jaronoff97:
**Description:** Adds the ability to set pods in component health. This lets users know how many pods are running for each of their collector pools

**Link to tracking Issue:** Part 1 of #2489 

**Testing:** new unit tests

**Documentation:** n/a

### DIFF
--- a/cmd/operator-opamp-bridge/agent/agent.go
+++ b/cmd/operator-opamp-bridge/agent/agent.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -29,6 +30,7 @@ import (
 	"k8s.io/utils/clock"
 	"sigs.k8s.io/yaml"
 
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
 	"github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/config"
 	"github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/metrics"
 	"github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/operator"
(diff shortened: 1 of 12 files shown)


cmd/operator-opamp-bridge/config/config.go
```diff
@@ -269,7 +269,7 @@ func LoadFromFile(cfg *Config, configFile string) error {
 	if err != nil {
 		return err
 	}
-	if err = yaml.UnmarshalStrict(yamlFile, cfg); err != nil {
+	if err = yaml.Unmarshal(yamlFile, cfg); err != nil {
```
TylerHelmuth, line 272: What about this new feature required the function switch?

cmd/operator-opamp-bridge/agent/agent.go
```diff
@@ -102,20 +104,70 @@ func (agent *Agent) getHealth() *protobufs.ComponentHealth {
...
+			selMap[kv[0]] = kv[1]
+		}
+		return selMap
+	}
+	return map[string]string{
+		"app.kubernetes.io/managed-by": "opentelemetry-operator",
+		"app.kubernetes.io/instance":   fmt.Sprintf("%s.%s", col.GetNamespace(), col.GetName()),
+		"app.kubernetes.io/part-of":    "opentelemetry",
+		"app.kubernetes.io/component":  "opentelemetry-collector",
+	}
```
TylerHelmuth, line 151: Is this a safe default? Is it possible for user to tell the operator not to add these labels to the collector AND not provide any selectors of their own?

cmd/operator-opamp-bridge/config/config.go
jaronoff97: nothing, but this is something i noticed us doing that we probably shouldn't be. See [this PR for more](https://github.com/open-telemetry/opentelemetry-operator/pull/2466). TL;DR: doing this will allow us to be more flexible with the supplied configuration from the operator in the future.

cmd/operator-opamp-bridge/agent/agent.go
jaronoff97: this is entirely static from the operator and not user customizable. https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/manifests/manifestutils/labels.go#L76-L83

TylerHelmuth (approved):

PR #2514, opened 2024-01-13
Update the OpenTelemetry Java agent version to 2.0.0

opentelemetrybot:
Update the OpenTelemetry Java agent version to `2.0.0`.


TylerHelmuth (changes requested):
We'll want to think carefully about the implications of this bump to our users. Blocking until we can discuss. There are a lot of important, but breaking, changes.

TylerHelmuth:
Changelog for reference: https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v2.0.0

pavolloffay:
@TylerHelmuth I think only the following is relevant for the operator 

```
The default OTLP protocol has been changed from grpc to http/protobuf in order to align with the [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.28.0/specification/protocol/exporter.md#specify-protocol). You can switch to the grpc protocol using OTEL_EXPORTER_OTLP_PROTOCOL=grpc or -Dotel.exporter.otlp.protocol=grpc.
```

TylerHelmuth:
@pavolloffay I am more concerned about forcing the breaking changes on users.  If we update the default version to 2.0, when a user takes the latest operator version and eventually recreate their java deploys that opt into auto-instrumentation they will suddenly get _a lot_ of impactful breaking changes.  

Since Java has acknowledged how disruptive the 2.0 changes are by supporting 1.x versions for the next 6 months, I wonder if we should do the same.  I think we should default to the 1.x versions until only 2.x is supported.  If users want to use 2.0 auto-instrumenation in the meantime they are welcome to update the version themselves.

swiatekm-sumo:
@TylerHelmuth can you open an issue to house this discussion? Looks like we need to carefully consider what to do, and I'd rather do that in PR comments.

pavolloffay (approved):
Let's merge this PR and publish the image. This will enable users to use the 2.0.0. The operator will keep using the 1.x (the version used by the operator is hosted in versions.txt)

pavolloffay:
@swiatekm-sumo here is the ticket https://github.com/open-telemetry/opentelemetry-operator/issues/2529

TylerHelmuth:
@pavolloffay I see what you're saying.  I support publishing our java auto-instrumentation using 2.0, but keeping the default value to be 1.x. I think we'll need to rework some stuff so that we can publish both 2.x and 1.x versions for the next 6 months. Let me take a look at how we'd do that.

pavolloffay:
The image gets published when https://github.com/open-telemetry/opentelemetry-operator/blob/main/autoinstrumentation/java/version.txt#L1 changes. It should be fine to use it for both version streams.

TylerHelmuth (approved):

PR #2516, opened 2024-01-14
Add TargetAllocator CRD

swiatekm-sumo:
**Description:**

Add type definitions for the new TargetAllocator CRD. See the linked issue for the plan on how this will be rolled out.

**Link to tracking Issue:** #2422


jaronoff97 (commented):
I think this LGTM. It's a solid simple start that we can add to later. Are we going to add the collector selector in this PR or in a later one? Also, can we adjust the v1alpha2 collector's existing TA config to be a minimal version of this, or do we want to do that in a follow?

swiatekm-sumo:
> Are we going to add the collector selector in this PR or in a later one?

I'd like to add it alongside the conversion from the embedded TA to the v1alpha2 CRD.

> Also, can we adjust the v1alpha2 collector's existing TA config to be a minimal version of this, or do we want to do that in a follow?

I'm not completely convinced we want to do this at all tbh, it will make the conversion a lot more complicated.

jaronoff97 (approved):

apis/v1alpha2/targetallocator_types.go
```diff
@@ -0,0 +1,106 @@
...
+	// +optional
+	PrometheusCR TargetAllocatorPrometheusCR `json:"prometheusCR,omitempty"`
+}
+
+// TargetAllocatorPrometheusCR configures Prometheus CustomResource handling in the Target Allocator.
+type TargetAllocatorPrometheusCR struct {
+	// Enabled indicates whether to use a PrometheusOperator custom resources as targets or not.
+	// +optional
+	Enabled bool `json:"enabled,omitempty"`
+	// Interval between consecutive scrapes. Equivalent to the same setting on the Prometheus CRD.
```
pavolloffay, line 48: is there any precedence order for the interval defined here and on the service/pod monitors?

apis/v1alpha2/targetallocator_types.go
```diff
@@ -0,0 +1,106 @@
...
+	PodMonitorSelector *metav1.LabelSelector `json:"podMonitorSelector,omitempty"`
+	// ServiceMonitors to be selected for target discovery.
+	// This is a map of {key,value} pairs. Each {key,value} in the map is going to exactly match a label in a
+	// ServiceMonitor's meta labels. The requirements are ANDed.
+	// +optional
+	ServiceMonitorSelector *metav1.LabelSelector `json:"serviceMonitorSelector,omitempty"`
+}
+
+// TargetAllocatorStatus defines the observed state of Target Allocator.
+type TargetAllocatorStatus struct {
```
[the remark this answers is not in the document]
pavolloffay: +1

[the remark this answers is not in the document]
pavolloffay: It would be cool to book a ticket for it :)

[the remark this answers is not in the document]
swiatekm-sumo: I wonder how difficult it is to actually do this. We'd have to get the target allocator metrics, which while not exactly difficult, seems not very robust to me.

apis/v1alpha2/targetallocator_types.go
swiatekm-sumo: It's exactly the same as the respective setting on the Prometheus CR - a default that can be overridden in individual CRs.

pavolloffay: is it smth that should be documented here?

swiatekm-sumo: The comment already says it's equivalent to the Prometheus CR attribute, but I can also mention it's an overridable default.

pavolloffay (approved):

apis/v1alpha2/targetallocator_types.go
swiatekm-sumo: Done.

swiatekm-sumo:
I've also added a separate type for filter strategies and set the right default in kubebuilder tags.

pavolloffay (approved):

apis/v1alpha2/targetallocator_types.go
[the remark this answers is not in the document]
jaronoff97: certainly something to figure out in the future. https://github.com/open-telemetry/opentelemetry-operator/issues/2549

PR #2517, opened 2024-01-14
Create release only after artifacts are published

### LINKED ISSUE #2304: Releases should be created after all the artifacts are ready
### Component(s)

operator, target allocator, opamp bridge

We create and publish the release before our container images are published. Normally, this makes no difference, but it's bitten us a bit due to #2295, and would also cause problems if any of the builds failed.

Unfortunately, fixing this isn't that easy, as Github Actions only has very basic support for dependencies between workflows. In order for this to happen, we'd need to refactor our CI into reusable workflows, and then have all the building and publishing happen in a dedicated release workflow.

A simple fix in the meantime would be to create the release in draft mode and have the release manager manually publish it once everything's fine. WDYT @jaronoff97 @pavolloffay ?

swiatekm-sumo:
Resolves #2304. I've moved all the image publishing to a single workflow, and had the release workflow run on successful completion of that one.


pavolloffay (approved):

.github/workflows/publish-images.yaml
```diff
@@ -103,3 +103,143 @@ jobs:
           labels: ${{ steps.docker_meta.outputs.labels }}
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache
+  publish-targetalloactor:
```
jaronoff97, line 106:
```suggestion
  publish-targetallocator:
```

.github/workflows/release.yaml
```diff
@@ -1,11 +1,14 @@
 name: "Prepare the release"
 on:
-  push:
-    tags: [ 'v*' ]
+  workflow_run:
+    workflows: [Publish container images]
+    types:
```
jaronoff97, line 5: I think right now this would publish a release on every commit to main. The container publish runs on any tag and any commit to main. Is there a way we can only run this on a tag?

jaronoff97 (changes requested):

.github/workflows/release.yaml
swiatekm-sumo: Good catch. And unfortunately, it doesn't look like we can.

Looks like properly solving this involves creating a reusable workflow for publishing our images and creating separate workflows for publishing dev and release images, making use of that reusable workflow. Then we can trigger based only on release images. Or just have a single release workflow that does everything.

jaronoff97: i think consolidating to a single workflow makes sense to me. We would then just conditionally enable the actual publish part if its a tag.

swiatekm-sumo:
I'm going to rethink this and come up with a better approach.

PR #2532, opened 2024-01-18
internal/manifests/collector: switch internally to v1alphav2

frzifus:


iblancasa:
Could you add some description/link to the issue to explain what are you implementing here?

internal/manifests/collector/annotations.go
```diff
@@ -66,7 +68,11 @@ func PodAnnotations(instance v1alpha1.OpenTelemetryCollector) map[string]string
 	return podAnnotations
 }
 
-func getConfigMapSHA(config string) string {
-	h := sha256.Sum256([]byte(config))
+func getConfigMapSHA(config v1alpha2.Config) string {
```
pavolloffay, line 71: How does it impact the upgrade? It seems like the SHA is changing with the new version

pkg/sidecar/pod.go
```diff
@@ -39,7 +40,7 @@ func add(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelemetryCo
 		return pod, err
 	}
 
-	container := collector.Container(cfg, logger, otelcol, false)
+	container := collector.Container(cfg, logger, convert.V1Alpha1to2(otelcol), false)
```
pavolloffay, line 43: Shouldn't the signature of `add` change to the new type?

frzifus: I plan to change that in one of the follow up prs.
In this one I tried to limited the scope only to `internal/manifests/collector`.

internal/manifests/collector/annotations.go
frzifus: Thats actually a good point. I assume it does and the operator will update all pods? How does it work during an operator upgrade increasing the version from e.g. `0.88.0` to `0.89.0`? The config checksum would be the same, but the container images would be different?

internal/api/convert/v1alpha.go
```diff
@@ -0,0 +1,154 @@
...
+
+	collectorJson, err := yaml.YAMLToJSON([]byte(in.Spec.Config))
+	if err != nil {
+		return v1alpha2.OpenTelemetryCollector{}
+	}
+
+	cfg := &v1alpha2.Config{}
+	if err := json.Unmarshal(collectorJson, cfg); err != nil {
+		return v1alpha2.OpenTelemetryCollector{}
+	}
```
frzifus, line 42: I'm not quite sure whether it wouldn't make more sense to propagate the error. Wdyt?

internal/manifests/collector/container_test.go
```diff
@@ -108,12 +114,6 @@ service:
 		specPorts     []corev1.ServicePort
 		expectedPorts []corev1.ContainerPort
 	}{
-		{
-			description:   "bad spec config",
-			specConfig:    "🦄",
-			specPorts:     nil,
-			expectedPorts: []corev1.ContainerPort{},
```
frzifus, line 115: There is no more bad spec. :)

internal/manifests/collector/annotations.go
jaronoff97: yeah i think the SHA changing is fine, it would result in a new replica set iirc. Same as what would happen on a version upgrade. Acceptable IMO

internal/api/convert/v1alpha.go
jaronoff97: I think we should propagate it as it would imply an invalid config is being supplied.

internal/api/convert/v1alpha.go
```diff
@@ -0,0 +1,154 @@
...
+		ObjectMeta: copy.ObjectMeta,
+	}
+
+	collectorJson, err := yaml.YAMLToJSON([]byte(in.Spec.Config))
+	if err != nil {
+		return v1alpha2.OpenTelemetryCollector{}
+	}
+
+	cfg := &v1alpha2.Config{}
+	if err := json.Unmarshal(collectorJson, cfg); err != nil {
```
jaronoff97, line 40: We should be able to just yaml marshal here i believe. The bridge does [this right now](https://github.com/jdcrouse/opentelemetry-operator/blob/main/cmd/operator-opamp-bridge/operator/client.go#L124-L125)

internal/api/convert/v1alpha.go
```diff
@@ -0,0 +1,154 @@
...
+
+	if copy.Spec.Autoscaler != nil {
+		metrics := make([]v1alpha2.MetricSpec, len(copy.Spec.Autoscaler.Metrics))
+		for i, m := range copy.Spec.Autoscaler.Metrics {
+			metrics[i] = v1alpha2.MetricSpec{
+				Type: m.Type,
+				Pods: m.Pods,
+			}
+		}
+		out.Spec.OpenTelemetryCommonFields.Autoscaler = &v1alpha2.AutoscalerSpec{
```
jaronoff97, line 59: do we need to worry about the deprecated min and max replica fields?

internal/api/convert/v1alpha.go
```diff
@@ -0,0 +1,154 @@
...
+			SuccessThreshold:              copy.Spec.LivenessProbe.SuccessThreshold,
+			FailureThreshold:              copy.Spec.LivenessProbe.FailureThreshold,
+			TerminationGracePeriodSeconds: copy.Spec.LivenessProbe.TerminationGracePeriodSeconds,
+		}
+	}
+
+	out.Spec.Observability.Metrics.EnableMetrics = copy.Spec.Observability.Metrics.EnableMetrics
+
+	out.Spec.ConfigMaps = copy.Spec.ConfigMaps
+	out.Spec.DaemonSetUpdateStrategy = appsv1.DaemonSetUpdateStrategy{} // NOTE: N/A in v1alpha1
```
jaronoff97, line 149: This does exist in v1alpha1, it's just called [UpdateStrategy](https://github.com/jdcrouse/opentelemetry-operator/blob/main/apis/v1alpha1/opentelemetrycollector_types.go#L283)

internal/api/convert/v1alpha.go
```diff
@@ -0,0 +1,154 @@
...
+	out.Spec.DeploymentUpdateStrategy.Type = copy.Spec.DeploymentUpdateStrategy.Type
+	out.Spec.DeploymentUpdateStrategy.RollingUpdate = copy.Spec.DeploymentUpdateStrategy.RollingUpdate
```
jaronoff97, line 151: does it matter if any of these fields are nil and we're trying to update them? I would think we would panic.

internal/api/convert/v1alpha.go
```diff
@@ -0,0 +1,154 @@
...
+	"encoding/json"
+
+	appsv1 "k8s.io/api/apps/v1"
+	"sigs.k8s.io/yaml"
+
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha2"
+)
+
+func V1Alpha1to2(in v1alpha1.OpenTelemetryCollector) v1alpha2.OpenTelemetryCollector {
```
jaronoff97, line 27: Is there any prior work we can base this off of? I know prometheus has done a few version upgrades before.

internal/manifests/collector/container.go
```diff
@@ -36,14 +38,26 @@ import (
...
+
+	configYaml, err := yaml.JSONToYAML(configJson)
+	if err != nil {
+		logger.Error(err, "could not convert json to yaml")
+		return corev1.Container{}
+	}
+
 	// build container ports from service ports
-	ports, err := getConfigContainerPorts(logger, otelcol.Spec.Config)
+	ports, err := getConfigContainerPorts(logger, string(configYaml))
```
jaronoff97, line 60: should we just have this work off of the config type? Or is this something we want to do in a followup?

internal/manifests/collector/container.go
```diff
@@ -133,7 +147,7 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelem
 	}
 
 	var livenessProbe *corev1.Probe
-	if configFromString, err := adapters.ConfigFromString(otelcol.Spec.Config); err == nil {
+	if configFromString, err := adapters.ConfigFromString(string(configYaml)); err == nil {
```
jaronoff97, line 150: same here

internal/manifests/collector/daemonset.go
```diff
@@ -19,44 +19,46 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/open-telemetry/opentelemetry-operator/internal/api/convert"
 	"github.com/open-telemetry/opentelemetry-operator/internal/manifests"
 	"github.com/open-telemetry/opentelemetry-operator/internal/manifests/manifestutils"
 	"github.com/open-telemetry/opentelemetry-operator/internal/naming"
 )
 
 // DaemonSet builds the deployment for the given instance.
 func DaemonSet(params manifests.Params) *appsv1.DaemonSet {
```
jaronoff97, line 29: I wonder if we should just have the params here work off of the v1alpha2 type rather than converting it here.

internal/manifests/collector/horizontalpodautoscaler.go
```diff
@@ -93,19 +95,24 @@ func HorizontalPodAutoscaler(params manifests.Params) client.Object {
...
 				APIVersion: v1alpha1.GroupVersion.String(),
 				Kind:       "OpenTelemetryCollector",
-				Name:       naming.OpenTelemetryCollector(params.OtelCol.Name),
+				Name:       naming.OpenTelemetryCollector(otelCol.Name),
 			},
-			MinReplicas: params.OtelCol.Spec.Autoscaler.MinReplicas,
-			MaxReplicas: *params.OtelCol.Spec.Autoscaler.MaxReplicas,
-			Metrics:     metrics,
+			MinReplicas: otelCol.Spec.OpenTelemetryCommonFields.Autoscaler.MinReplicas,
+			MaxReplicas: func(max *int32) int32 {
```
jaronoff97, line 101: can we make this a separate function? Why did we not need to do this before?

internal/manifests/collector/rbac.go
```diff
@@ -50,7 +64,16 @@ func ClusterRole(params manifests.Params) *rbacv1.ClusterRole {
 }
 
 func ClusterRoleBinding(params manifests.Params) *rbacv1.ClusterRoleBinding {
-	configFromString, err := adapters.ConfigFromString(params.OtelCol.Spec.Config)
+	otelCol := convert.V1Alpha1to2(params.OtelCol)
```
jaronoff97, line 67: same as above... I do think this may be easier for us to just do the conversion when we construct manifest.Params

jaronoff97 (commented):
A few thoughts, but it does feel like we're on a good path here!

internal/api/convert/v1alpha.go
frzifus: That should actually be fine. Its safe to access e.g. `copy.Spec.DeploymentUpdateStrategy` since its a value. `.RollingUpdate` is a ptr. But if this one is `nil` it means we assign `nil` from the `copy` to `out`.

For all ptr fields, I added a `!= nil` check.

internal/manifests/collector/container.go
frzifus: Thats something I would like to do as a follow up. Originally I planned here also to submit a smaller PR. I would prefer to slowly work across the code base and do more intense changes in separate Prs.

internal/manifests/collector/daemonset.go
frzifus: I agree, but that change had other implications for example on `opampbridge` and `targetallocator`.

internal/manifests/collector/horizontalpodautoscaler.go
frzifus: The field `MaxReplicas` is in `v1alpha2` its now  a value. The field  `k8s.io/api/autoscaler/v2.HorizontalPodAutoscalerSpec.MaxReplicas` is a value too.

https://github.com/open-telemetry/opentelemetry-operator/blob/d6f3c31a072979783461dd9d3b23d3e12aeb1f26/apis/v1alpha1/opentelemetrycollector_types.go#L473-L479

https://github.com/open-telemetry/opentelemetry-operator/blob/d6f3c31a072979783461dd9d3b23d3e12aeb1f26/apis/v1alpha2/common.go#L47-L53

internal/manifests/collector/rbac.go
frzifus: https://github.com/open-telemetry/opentelemetry-operator/pull/2532#discussion_r1458052336

internal/manifests/collector/daemonset.go
frzifus: good question, I assume it tests and all the things need to be changed in at least the other sub packages.

internal/api/convert/v1alpha.go
frzifus: I did say no. Since you are still able to set min/max in the AutoscalerSpec.

internal/api/convert/v1alpha_test.go
```diff
@@ -0,0 +1,61 @@
...
+		},
+	}
+
+	cfgV2, err := V1Alpha1to2(cfgV1)
+	assert.Nil(t, err)
+	assert.NotNil(t, cfgV2)
+
+	jsonCfg, err := json.Marshal(&cfgV2.Spec.Config)
+	assert.Nil(t, err)
+	yamlCfg, err := yaml.JSONToYAML(jsonCfg)
```
jaronoff97, line 58: getting a test fail here

internal/api/convert/v1alpha_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/yaml.v3"
+
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
+)
+
+func Test_V1Alpha1to2_Config(t *testing.T) {
```
jaronoff97, line 27: can you add a case for when we fail to convert a v1alpha1 to a v1alpha2?

internal/manifests/collector/rbac.go
```diff
@@ -46,41 +58,49 @@ func ClusterRole(params manifests.Params) *rbacv1.ClusterRole {
...
 }
 
-func ClusterRoleBinding(params manifests.Params) *rbacv1.ClusterRoleBinding {
-	configFromString, err := adapters.ConfigFromString(params.OtelCol.Spec.Config)
+func ClusterRoleBinding(params manifests.Params) (*rbacv1.ClusterRoleBinding, error) {
+	otelCol, err := convert.V1Alpha1to2(params.OtelCol)
+	if err != nil {
+		return nil, err
+	}
+	confStr, err := yaml.Marshal(&otelCol.Spec.Config)
```
jaronoff97, line 69: thought: I wonder if we should just make this a method on the config object, given we have to do this in a few places

pkg/sidecar/pod.go
```diff
@@ -39,7 +40,11 @@ func add(cfg config.Config, logger logr.Logger, otelcol v1alpha1.OpenTelemetryCo
 		return pod, err
 	}
 
-	container := collector.Container(cfg, logger, otelcol, false)
+	otc, err := convert.V1Alpha1to2(otelcol)
+	if err != nil {
+		return corev1.Pod{}, err
+	}
+	container := collector.Container(cfg, logger, otc, false)
```
jaronoff97, line 47: I think the replace config needs to come after the conversion `otelColCfg, err := collector.ReplaceConfig(otelcol)`

internal/manifests/collector/rbac.go
frzifus: y, right now it would be useful. But as soon as `params` contain the `v1alpha2` it becomes irrelevant again.

frzifus:
Seems there are still some tests failing that compare strings.
e.g.:
```
- "collector.yaml": "receivers:\n  examplereceiver:\n    endpoint: \"0.0.0.0:12345\"\nexporters:\n  logging:\nservice:\n  pipelines:\n    metrics:\n      receivers: [examplereceiver]\n      exporters: [logging]\n",
+ "collector.yaml": "exporters:\n    logging: null\nreceivers:\n    prometheus:\n        config: {}\n        target_allocator:\n            collector_id: ${POD_NAME}\n            endpoint: http://test-targetallocator:80\n            interval: 30s\nservice:\n    pipelines:\n        metrics:\n            exporters:\n                - logging\n            receivers:\n                - prometheus\n",
```

internal/manifests/collector/daemonset_test.go
```diff
@@ -46,7 +46,7 @@ func TestDaemonSetNewDefault(t *testing.T) {
 	}
 
 	// test
-	d := DaemonSet(params)
+	d, _ := DaemonSet(params)
```
pavolloffay, line 49: testing the error should not hurt

internal/manifests/collector/daemonset_test.go
```diff
@@ -104,7 +104,7 @@ func TestDaemonsetHostNetwork(t *testing.T) {
 		Log: logger,
 	}
 	// test
-	d1 := DaemonSet(params1)
+	d1, _ := DaemonSet(params1)
```
pavolloffay, line 107: here as well and in the follow up lines

pavolloffay (commented):
LGTM I would add require statements for returned error that was added to many manifest files

internal/manifests/collector/rbac.go
jaronoff97: we'll still need to marshal the config regardless to populate the config map and also to generate a hash right?

frzifus:
Well, now the formatting in the e2e tests seems to be slightly different. :raccoon:

internal/manifests/collector/annotations.go
iblancasa: If this is a problem, it should be detected by the E2E tests

internal/manifests/collector/annotations.go
```diff
@@ -66,7 +68,11 @@ func PodAnnotations(instance v1alpha1.OpenTelemetryCollector) map[string]string
...
+	b, err := json.Marshal(&config)
+	if err != nil {
+		return "invalid"
```
iblancasa, line 74: How about returning an error?

internal/manifests/collector/config_replace.go
```diff
@@ -42,18 +42,22 @@ type Config struct {
...
-func ReplaceConfig(instance v1alpha1.OpenTelemetryCollector) (string, error) {
+func ReplaceConfig(instance v1alpha2.OpenTelemetryCollector) (string, error) {
+	cfgStr, err := instance.Spec.Config.Yaml()
+	if err != nil {
+		return "", err
+	}
 	// Check if TargetAllocator is enabled, if not, return the original config
 	if !instance.Spec.TargetAllocator.Enabled {
-		return instance.Spec.Config, nil
+		return string(cfgStr), nil
```
iblancasa, line 52: If the `Yaml` method returns a string, why do you need to convert it to `string`?

internal/manifests/collector/configmap.go
```diff
@@ -18,16 +18,21 @@ import (
...
+	"github.com/open-telemetry/opentelemetry-operator/internal/api/convert"
 	"github.com/open-telemetry/opentelemetry-operator/internal/manifests"
 	"github.com/open-telemetry/opentelemetry-operator/internal/manifests/manifestutils"
 	"github.com/open-telemetry/opentelemetry-operator/internal/naming"
 )
 
 func ConfigMap(params manifests.Params) (*corev1.ConfigMap, error) {
-	name := naming.ConfigMap(params.OtelCol.Name)
-	labels := manifestutils.Labels(params.OtelCol.ObjectMeta, name, params.OtelCol.Spec.Image, ComponentOpenTelemetryCollector, []string{})
+	otelCol, err := convert.V1Alpha1to2(params.OtelCol)
```
iblancasa, line 28: Instead of converting in each method, wouldn't it simpler to create the `manifest.Params` structure using the `v1alpha2` version of the collector?

internal/manifests/collector/container_test.go
iblancasa: How about this situation?
```golang
	configJson, err := json.Marshal(&otelcol.Spec.Config)
	if err != nil {
		logger.Error(err, "could not marshal config to json")
		return corev1.Container{}
	}
```

internal/manifests/collector/horizontalpodautoscaler_test.go
```diff
@@ -74,19 +80,31 @@ func TestHPA(t *testing.T) {
...
+				require.NoError(t, err)
 
 				hpa := raw.(*autoscalingv2.HorizontalPodAutoscaler)
 
 				// verify
 				assert.Equal(t, "my-instance-collector", hpa.Name)
 				assert.Equal(t, "my-instance-collector", hpa.Labels["app.kubernetes.io/name"])
-				assert.Equal(t, int32(3), *hpa.Spec.MinReplicas)
-				assert.Equal(t, int32(5), hpa.Spec.MaxReplicas)
+				// require.NotNil(t, hpa.Spec.MinReplicas)
```
iblancasa, line 105: Cam this be removed?

internal/manifests/collector/horizontalpodautoscaler.go
iblancasa: Agree about moving to a different function

internal/manifests/collector/rbac.go
frzifus: yes

internal/manifests/collector/configmap.go
frzifus: Sure, thats the final goal. But to do this know, we need to adapt ta, opamp and other packages too.

- https://github.com/open-telemetry/opentelemetry-operator/pull/2532#discussion_r1458043087
- https://github.com/open-telemetry/opentelemetry-operator/pull/2532#issue-2087265240

internal/manifests/collector/horizontalpodautoscaler_test.go
frzifus: Why should it be removed? The goal is to of this PR is to only prepare the switch from v1alpha1 to v1alpha2 in `internal/manifests/collector`. Since this is quite a big change, I want to avoid changing all kinds of logic, I do not have to touch.

internal/api/convert/v1alpha_test.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/yaml.v3"
+
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
+)
+
+func Test_V1Alpha1to2(t *testing.T) {
```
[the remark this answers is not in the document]
frzifus: I do not think so. When fields are missing or invalid, other unittests will detect it.

internal/api/convert/v1alpha.go
```diff
@@ -0,0 +1,150 @@
...
+	out.Spec.OpenTelemetryCommonFields.Lifecycle = copy.Spec.Lifecycle
+	out.Spec.OpenTelemetryCommonFields.TerminationGracePeriodSeconds = copy.Spec.TerminationGracePeriodSeconds
+	out.Spec.OpenTelemetryCommonFields.TopologySpreadConstraints = copy.Spec.TopologySpreadConstraints
+	out.Spec.OpenTelemetryCommonFields.HostNetwork = copy.Spec.HostNetwork
+	out.Spec.OpenTelemetryCommonFields.ShareProcessNamespace = copy.Spec.ShareProcessNamespace
+	out.Spec.OpenTelemetryCommonFields.PriorityClassName = copy.Spec.PriorityClassName
+	out.Spec.OpenTelemetryCommonFields.InitContainers = copy.Spec.InitContainers
+	out.Spec.OpenTelemetryCommonFields.AdditionalContainers = copy.Spec.AdditionalContainers
+
+	out.Spec.TargetAllocator.Replicas = copy.Spec.TargetAllocator.Replicas
```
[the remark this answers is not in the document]
frzifus: Whats the advantage?

internal/manifests/collector/container_test.go
frzifus: How could this become an issue in the future? Only in such a case, right? ([docs](https://pkg.go.dev/encoding/json))
- Channel, complex, and function values cannot be encoded in JSON. Attempting to encode such a value causes Marshal to return an UnsupportedTypeError.
- JSON cannot represent cyclic data structures and Marshal does not handle them. Passing cyclic structures to Marshal will result in an error.

internal/manifests/collector/horizontalpodautoscaler.go
iblancasa: It's just to make it more readable

internal/api/convert/v1alpha.go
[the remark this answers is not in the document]
iblancasa: Make it easier to read and follow

internal/manifests/collector/horizontalpodautoscaler_test.go
iblancasa: Because the line is commented. So, it seems it is not needed anymore.

internal/manifests/collector/horizontalpodautoscaler.go
frzifus: From my point of view this is quite a common thing.

internal/api/convert/v1alpha.go
frzifus: I would prefer to keep the change as minimal as possible. We can add/remove things afterwards.

internal/manifests/collector/annotations.go
```diff
@@ -40,14 +41,20 @@ func Annotations(instance v1alpha1.OpenTelemetryCollector) map[string]string {
...
-	annotations["opentelemetry-operator-config/sha256"] = getConfigMapSHA(instance.Spec.Config)
+	annotations["opentelemetry-operator-config/sha256"] = hash
 
-	return annotations
+	return annotations, nil
 }
 
 // PodAnnotations return the spec annotations for OpenTelemetryCollector pod.
-func PodAnnotations(instance v1alpha1.OpenTelemetryCollector) map[string]string {
+func PodAnnotations(instance v1alpha2.OpenTelemetryCollector) (map[string]string, error) {
```
jaronoff97, line 57: note: in the future, i think we should make the getConfigMapSHA a method on the collector itself that returns no error. Note worth doing in this PR, but doing so in the future would help us clean up a lot of the error checking this introduces.

internal/manifests/collector/deployment.go
```diff
@@ -19,54 +19,66 @@ import (
...
 	"github.com/open-telemetry/opentelemetry-operator/internal/manifests"
 	"github.com/open-telemetry/opentelemetry-operator/internal/manifests/manifestutils"
 	"github.com/open-telemetry/opentelemetry-operator/internal/naming"
 )
 
 // Deployment builds the deployment for the given instance.
-func Deployment(params manifests.Params) *appsv1.Deployment {
-	name := naming.Collector(params.OtelCol.Name)
-	labels := manifestutils.Labels(params.OtelCol.ObjectMeta, name, params.OtelCol.Spec.Image, ComponentOpenTelemetryCollector, params.Config.LabelsFilter())
+func Deployment(params manifests.Params) (*appsv1.Deployment, error) {
```
jaronoff97, line 29: I think after this change and #2516 go in, i will modify these signatures to not take in the whole `manifests.Params` object, but only the object they need (i.e. collector in this case)

internal/manifests/collector/poddisruptionbudget.go
```diff
@@ -18,37 +18,45 @@ import (
...
 
+	"github.com/open-telemetry/opentelemetry-operator/internal/api/convert"
 	"github.com/open-telemetry/opentelemetry-operator/internal/manifests"
 	"github.com/open-telemetry/opentelemetry-operator/internal/manifests/manifestutils"
 	"github.com/open-telemetry/opentelemetry-operator/internal/naming"
 )
 
-func PodDisruptionBudget(params manifests.Params) *policyV1.PodDisruptionBudget {
+func PodDisruptionBudget(params manifests.Params) (*policyV1.PodDisruptionBudget, error) {
+	otelCol, err := convert.V1Alpha1to2(params.OtelCol)
```
jaronoff97, line 28: can you open an issue for cleaning this up after this is merged?

tests/e2e/targetallocator-prometheuscr/00-assert.yaml
```diff
@@ -27,24 +27,23 @@ metadata:
...
+        prometheus:
+            endpoint: 0.0.0.0:9090
     receivers:
-      prometheus:
-        config: {}
-        target_allocator:
-          collector_id: ${POD_NAME}
-          endpoint: http://prometheus-cr-targetallocator:80
-          interval: 30s
+        prometheus:
```
jaronoff97, line 33: is there a reason we need to change the whitespacing for these now? I saw the indent call above in the serialization and wasn't sure why we need that.

jaronoff97 (approved):
a few thoughts/questions. But overall i think this LGTM. I do think we should merge this after #2567 so that we don't incur a major performance hit from all this marshalling business we're doing now.

internal/manifests/collector/annotations.go
frzifus: I made an issue for this:
- https://github.com/open-telemetry/opentelemetry-operator/issues/2591

internal/manifests/collector/deployment.go
frzifus: - https://github.com/open-telemetry/opentelemetry-operator/issues/2592

internal/manifests/collector/poddisruptionbudget.go
frzifus: - https://github.com/open-telemetry/opentelemetry-operator/issues/2592

internal/manifests/collector/rbac.go
```diff
@@ -18,22 +18,33 @@ import (
...
 
-func ClusterRole(params manifests.Params) *rbacv1.ClusterRole {
-	configFromString, err := adapters.ConfigFromString(params.OtelCol.Spec.Config)
+func ClusterRole(params manifests.Params) (*rbacv1.ClusterRole, error) {
+	otelCol, err := convert.V1Alpha1to2(params.OtelCol)
+	if err != nil {
+		return nil, err
+	}
+
+	confStr, err := otelCol.Spec.Config.Yaml()
```
pavolloffay, line 34: The typed config is serialized to yaml and then back unserialized to  `map[interface{}]interface{}`

The typed config could be easily converted to `map[interface{}]interface{}` which is accepted in most signatures

tests/e2e/targetallocator-prometheuscr/00-assert.yaml
frzifus: y, thats a bit weird. [This one](https://github.com/open-telemetry/opentelemetry-operator/pull/2532/files#diff-395c66bb78902810e7c0ed75332a8b8dd0a404cf09306d0b9aa0e633ebfc3969R94) works fine. But as soon as we use `yaml.Marshal` it falls back to ident = 4.

pavolloffay (commented):
Great progress 🎉 

Just one comment, I would prefer avoiding type config -> yaml -> type config translation.

tests/e2e/targetallocator-prometheuscr/00-assert.yaml
frzifus: The default `yaml.Marshal` creates a `newEncoder()` which is initliazed with ident = 4.

https://github.com/go-yaml/yaml/blob/f6f7691b1fdeb513f56608cd2c32c51f8194bf51/encode.go#L56-L67

internal/manifests/collector/rbac.go
pavolloffay: please book a follow up ticket for this

pavolloffay: https://github.com/open-telemetry/opentelemetry-operator/issues/2603

pavolloffay (approved):

PR #2537, opened 2024-01-18
[Chore] Fix targetallocator-features e2e case

IshwarKanse:

Small fix for the targetallocator-features test case to work on OpenShift. The test had started failing after changes in PR https://github.com/open-telemetry/opentelemetry-operator/pull/2492 

Tested on OpenShift 4.14, 4.13.

### DIFF
--- a/tests/e2e/targetallocator-features/00-assert.yaml
+++ b/tests/e2e/targetallocator-features/00-assert.yaml
@@ -59,7 +59,7 @@ spec:
       securityContext:
         runAsUser: 1000
         runAsGroup: 3000
-        fsGroup: 2000
+        fsGroup: 3000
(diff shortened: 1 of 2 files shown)


swiatekm-sumo:
The intent of #2492 was to set an attribute on the container security context that is not present in Pod security context, to ensure we're not mixing up these types. `privileged: false` is a default, so checking it in a test doesn't necessarily prove anything. Can you replace it with a different attribute, if it causes problems on OpenShift?

IshwarKanse:
@swiatekm-sumo Ok, I changed few attributes to check and differentiate between the security contexts.

swiatekm-sumo (approved):

Issue #2540, opened 2024-01-18
Add shareProcessName to common fields for v1alpha2


jaronoff97:
_Originally posted by @jaronoff97 in https://github.com/open-telemetry/opentelemetry-operator/issues/2472#issuecomment-1898866451_

PR #2541, opened 2024-01-18
add shareProcessNamespace capability to v1alpha2

frzifus:

Closes https://github.com/open-telemetry/opentelemetry-operator/issues/2540

### DIFF
--- a/apis/v1alpha2/common.go
+++ b/apis/v1alpha2/common.go
@@ -195,6 +195,9 @@ type OpenTelemetryCommonFields struct {
 	// HostNetwork indicates if the pod should run in the host networking namespace.
 	// +optional
 	HostNetwork bool `json:"hostNetwork,omitempty"`
+	// ShareProcessNamespace indicates if the pod's containers should share process namespace.
+	// +optional
+	ShareProcessNamespace bool `json:"shareProcessNamespace,omitempty"`
 	// If specified, indicates the pod's priority.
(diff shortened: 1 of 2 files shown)


jaronoff97 (commented):
need to run make generate

frzifus:
afaik we do not generate `v1alpha2`, right?

swiatekm-sumo:
> afaik we do not generate `v1alpha2`, right?

We don't build manifests, but we do generate the deepcopy code.

frzifus:
Is coping a primitive values not covered in line 600? Since `make generate` does not produce any output. While when I change the field to be a ptr, I can see that the generator adds a few lines to copy the value. Once I change it back and run `make generate` it gets removed. :racehorse: 
https://github.com/open-telemetry/opentelemetry-operator/blob/e2773befed10dc6388a887229e35c673b60c883b/apis/v1alpha2/zz_generated.deepcopy.go#L598-L600

swiatekm-sumo:
I was confused why the documentation change was in this PR, but it looks like it's because the original PRs didn't include it. Which is strange, because the CI should fail that. I'll look into it, but this PR seems fine even though the changes are confusing.

swiatekm-sumo (approved):

frzifus:
Oh, I thought the generated lines belong to `v1alpha2`. But It seems that not part of `api.md` :rabbit: :facepalm:

pavolloffay (approved):

Issue #2542, opened 2024-01-18
The stable HTTP Semantic Conventions impact our Auto-Instrumentation capabilities


TylerHelmuth:
### Component(s)

instrumentation

### Describe the issue you're reporting

[The HTTP semantic conventions are now stable](https://opentelemetry.io/blog/2023/http-conventions-declared-stable/) and they bring lots of impactful breaking changes. As OTel languages take the stable conventions by default our auto-instrumentation images will get updated and those breaking changes will make their way to our users.

Unlike users who manually add instrumentation libraries, the Operator currently adds a layer of obfuscation around what auto-instrumentation is actually installed.  In situations where the operator is installing an "agent", like for Java, .Net, or Go, there is an extra layer of obfuscation.

We need to be very careful with these updates so that we do not causes our users to upgrade unknowingly, causing their telemetry to change. Here are some of the solutions we discussed on the SIG meeting on 2023-01-18:

- Determine a migration plan, per language, for auto-instrumentation users.  Things we can do to help:
   - Documentation
   - Warnings in the webhook
   - Feature gates
   - Major version bumps of our images.
- Determine how long we will wait, per language, before setting the default language image to use instrumentation libraries with the breaking changes.
- Communicate concerns with the TC/GC via the Maintainers meeting

pavolloffay:
This is related to https://github.com/open-telemetry/opentelemetry-operator/issues/2529

swiatekm:
From what I've investigated, this change has already been implemented in some of the images we ship:

- [x] Python (in [v0.46b0](https://github.com/open-telemetry/opentelemetry-python/blob/main/CHANGELOG.md#version-1250046b0-2024-05-30))
- [x] Node (in [0.54.0](https://www.npmjs.com/package/@opentelemetry/instrumentation-http#semantic-conventions))
- [ ] .NET (in 1.4.0, but we ship 1.3.0 by default)
- [ ] Java (in 2.0.0, but we ship 1.x by default)
- [x] Go (in [v0.14.0](https://github.com/open-telemetry/opentelemetry-go-instrumentation/releases/tag/v0.14.0-alpha))

@TylerHelmuth does this look right to you? It seems like we need to come up with a way of informing our .NET and Java consumers about the breaking change effectively, and then we should be able to bump the defaults and close this issue.

swiatekm:
[...]

PR #2543, opened 2024-01-18
[chore] Update release process

TylerHelmuth:
Update release process to warn against bumping Java and .NET versions.

Related to https://github.com/open-telemetry/opentelemetry-operator/issues/2542

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -4,6 +4,7 @@ Steps to release a new version of the OpenTelemetry Operator:
 
 1. Change the `versions.txt`, so that it lists the target version of the OpenTelemetry Collector (operand), and the desired version for the target allocator and the operator. The `major.minor` should typically match, with the patch portion being possibly different.
 2. Change the `autoinstrumentation-*` versions in `versions.txt` as per the latest supported versions in `autoinstrumentation/`.
+   - :warning: DO NOT BUMP JAVA PAST `1.32.X` AND DO NOT BUMP .NET PAST `1.2.0`. :warning: Upgrades past these versions will introduce breaking HTTP semantic convention changes. 
 3. Run `make bundle DOCKER_USER=open-telemetry VERSION=0.38.0`, using the version that will be released.
 4. Change the compatibility matrix in the [readme](./README.md) file, using the OpenTelemetry Operator version to be released and the current latest Kubernetes version as the latest supported version. Remove the oldest entry.
 5. Add the changes to the changelog. Manually add versions of all operator components.


.github/workflows/auto-update-java-agent.yaml
```diff
@@ -51,12 +51,20 @@ jobs:
...
 
-      - name: Update version
+      - name: Update 1.x version
         env:
           VERSION: ${{ needs.check-versions.outputs.latest-version }}
+        if: ${{ startsWith(needs.check-versions.outputs.latest-version, "1" }}
         run: |
           echo $VERSION > autoinstrumentation/java/version.txt
 
+      - name: Update 2.x version
```
pavolloffay, line 61: What is the motivation for having a separate version for 2.x e.g.  `autoinstrumentation/java/2x-version.txt`? 

I find this unnecessary since autoinstrumentation/java/version.txt can be used for both versions and updated when a new image needs to be shipped.

RELEASE.md
```diff
@@ -4,6 +4,7 @@ Steps to release a new version of the OpenTelemetry Operator:
 
 1. Change the `versions.txt`, so that it lists the target version of the OpenTelemetry Collector (operand), and the desired version for the target allocator and the operator. The `major.minor` should typically match, with the patch portion being possibly different.
 2. Change the `autoinstrumentation-*` versions in `versions.txt` as per the latest supported versions in `autoinstrumentation/`.
+   - :warning: DO NOT BUMP JAVA PAST `1.32.X` AND DO NOT BUMP .NET PAST `1.2.0`. :warning: Upgrades past these versions will introduce breaking HTTP semantic convention changes.
```
pavolloffay, line 7: +1

Kielek: > @TylerHelmuth, http/network semantic convention was changed for ASP.NET, ASP.NET Core and HTTP Instrumentation. There is no support for old sem. conv.
> 
> Theoretically (I know that the real world is different), it is not a problem:
> 1. Previously http and network sem. conv. was not stable. There were no breaking changes after stabilization.
> 2. There were no stable instrumentation libraries in OTel for .NET. The version with new sem. conv. is first stable release.
> 3. All instrumentations are/were marked as not stable in automatic instrumentation. As long as they are not stable, we can release minor versions.
> 4. Operator is not stable (<1.0.0).
> 
> What is more, OTel Sem. conv. will accept a lot of breaking changes in the future for non-stable parts. If e.g. SQL/DB sem/ conv. will be marked as stable, do you recommend to release Java 3.0? If not, why? It is serious question, I do not have a good answer. Do you have a plan to block such upgrades here?
> 
> What's about dropping support for .NET6 and .NET7. Both of them will reach EOL this year. .NET team decided to support only currently supported versions (without rising major version). Do you have plan to also block the upgrade versions here?
> 
> IMO all these cases touches same issue problem and I do not have good answer. Still, theoretically we do not do anything against policies and we can make upgrades with clear statement in changelog.

_Originally posted by @Kielek in https://github.com/open-telemetry/opentelemetry-operator/issues/2538#issuecomment-1900352585_

.github/workflows/auto-update-java-agent.yaml
TylerHelmuth: Sorry my branch was messed up, I'll clean this up.

RELEASE.md
TylerHelmuth: See https://github.com/open-telemetry/opentelemetry-operator/pull/2538#issuecomment-1900747043

pavolloffay (approved):

jaronoff97 (approved):

PR #2421, opened 2023-12-06
Improve developer experience by minimizing local changes

pavolloffay:
**Description:** <Describe what has changed.>

Replaces #1794 

**Link to tracking Issue:** <Issue number if applicable>

**Testing:** <Describe what testing was performed and which tests were added.>

**Documentation:** <Describe the documentation added.>

PR #2544, opened 2024-01-18
Makefile: add api-docs to bundle rule 

frzifus:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -451,7 +451,7 @@ operator-sdk:
 
 # Generate bundle manifests and metadata, then validate generated files.
 .PHONY: bundle
-bundle: kustomize operator-sdk manifests set-image-controller
+bundle: kustomize operator-sdk manifests set-image-controller api-docs
 	$(OPERATOR_SDK) generate kustomize manifests -q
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -192,7 +192,7 @@ lint: golangci-lint
 
 # Generate code
 .PHONY: generate
-generate: controller-gen
+generate: api-docs controller-gen
```
pavolloffay, line 195: the `api-docs` traget was remvoed from generate to improve developer UX by minimizing local changes when developing the operator.  #2421

swiatekm-sumo: Was that actually necessary? `api-docs` is a noop if you haven't made any changes impacting the docs, so it should be fine to have it here.

pavolloffay: What does the generate update? Is it the deepcopy? 

I think the api-docs generate should be bundled with the target to generate CRD

swiatekm-sumo: `generate` only generates the deepcopy code, yes.

swiatekm-sumo: But I do agree that `api-docs` should be a part of `bundle` rather than `generate`.

swiatekm-sumo (approved):

PR #2553, opened 2024-01-19
apis/v1alpha2: add yaml marshal and unmarshal methods to config

### LINKED ISSUE #2551: make v1alpha2 config marshling work with gopkg.in/yaml
### Component(s)

operator

The marshaling and unmarshalling of `v1alpha2.Config` adds an extra object to the converted configuration when gopkg.in/yaml.v2 or gopkg.in/yaml.v3 is used. We should impl https://pkg.go.dev/gopkg.in/yaml.v3#Marshaler and https://pkg.go.dev/gopkg.in/yaml.v3#Unmarshaler.

frzifus:

Closes #2551


apis/v1alpha2/config_test.go
```diff
@@ -23,6 +23,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	go_yaml "gopkg.in/yaml.v2"
```
jaronoff97, line 26: can we use v3 here?

apis/v1alpha2/config.go
```diff
@@ -69,27 +69,45 @@ func (c *AnyConfig) MarshalJSON() ([]byte, error) {
 	return json.Marshal(c.Object)
 }
 
+// UnmarshalYAML implements an alternative parser this field.
```
jaronoff97, line 72: can you add a static type check for AnyConfig that it adheres to the YAML marshal/unmarshal interface?

frzifus: It is now based on go-yaml version 3. I have adapted the unmarshal signature accordingly and added a static type check.

apis/v1alpha2/config.go
```diff
@@ -69,27 +73,45 @@ func (c *AnyConfig) MarshalJSON() ([]byte, error) {
...
+	vals := map[string]interface{}{}
+	if err := value.Decode(&vals); err != nil {
+		return err
+	}
+	c.Object = vals
+	return nil
+}
+
+// MarshalYAML specifies how to convert this object into YAML.
+func (c *AnyConfig) MarshalYAML() (interface{}, error) {
```
pavolloffay, line 87: Does the PR fixe the issue without defining `UnmarshalYAML` and `MarshalYAML`

jaronoff97 (approved):

apis/v1alpha2/config.go
frzifus: What do you mean?

pavolloffay: Do we need to have defined these two methods? 

If we don't define them would the test added in this PR fail?

frzifus: Without the methods, the test would fail. Because the default unmarshal/marshal methods take the `Object` member into account.

apis/v1alpha2/config.go
```diff
@@ -128,7 +128,16 @@ func hasNullValue(cfg map[string]interface{}) []string {
...
 		}
 		if reflect.ValueOf(v).Kind() == reflect.Map {
-			nulls := hasNullValue(v.(map[string]interface{}))
+			val := make(map[string]interface{})
+			switch t := v.(type) {
+			case map[string]interface{}:
+				val = t
+			case map[interface{}]interface{}:
+				for kk, vv := range t {
+					val[fmt.Sprintf("%v", kk)] = vv
```
pavolloffay, line 137: This looks suspicious. 

When does this happen? It would be good to have a test for this.

frzifus: Oh well, seems that its also no longer required. I replaced it with a default section that returns an empty array. Instead of panic if its not a `map[string]interface{}`

pavolloffay:
The unit test failed

.chloggen/v1alpha2_config_yaml-go_marshling.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: add yaml marshal and unmarshal methods to v1alpha2.config
```
pavolloffay, line 8: This needs to be updated

.chloggen/v1alpha2_config_yaml-go_marshling.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
pavolloffay, line 2: this seems more like a bug_fix

swiatekm-sumo:
Do we even need a changelog entry here? I doubt users care what we're doing with our unpublished types.

pavolloffay:
yeah +1 on getting rid of it. It is mostly internal detail for now.

pavolloffay (approved):

pavolloffay:
unit tests  failed again

frzifus:
I will re-run it. Seems to be another flaky test. https://github.com/open-telemetry/opentelemetry-operator/issues/2556

PR #2563, opened 2024-01-24
Fix truncated descriptions in the CRDs

### LINKED ISSUE #2561: Documentation for `observability.metrics.enableMetrics` truncated
### Component(s)

operator, collector

### What happened?

I am currently trying to wrap my head around how to let the OpenTelemetry Collector expose a `ServiceMonitor`. While doing so, I checked the [API docs](https://github.com/open-telemetry/opentelemetry-operator/blob/0821383bfd392366c0e7065041e2c3af4ff57b33/docs/api.md#opentelemetrycollectorspecobservabilitymetrics) for the Collector CRD. In there, it contains the following description of the `enableMetrics` key:

> EnableMetrics specifies if ServiceMonitor or PodMonitor(for sidecar mode) should be created for the service managed by the OpenTelemetry Operator. The operator.observability.

This description is obviously truncated. The issue apparently was introduced in PR #1983. 

It would be great if the full docs entry could be added back as the instruction to enable the `operator.observability.prometheus` feature gate seems pretty vital to make use of this key.

Tangent: I'd really appreciate more thorough docs somewhere on how to set up the OTel Collector with its Prometheus exporter and a Service/PodMonitor!

### Kubernetes Version

1.27.7

### Operator version

v0.92.1

### Collector version

v0.92.0

## Environment
Microsoft AKS

### Log output

_No response_

### Additional context

_No response_

iblancasa:
Fixes #2561.


jaronoff97:
@iblancasa we had to do this because we were reaching a limit for the size of a crd's annotations field. I'm not sure there's a way around this... cc @pavolloffay

bundle/manifests/opentelemetry.io_opentelemetrycollectors.yaml
```diff
@@ -49,39 +49,52 @@ spec:
...
             type: string
           metadata:
             type: object
           spec:
             description: OpenTelemetryCollectorSpec defines the desired state of OpenTelemetryCollector.
             properties:
               additionalContainers:
-                description: AdditionalContainers allows injecting additional containers
-                  into the Collector's pod definition.
+                description: 'AdditionalContainers allows injecting additional containers
```
swiatekm-sumo, line 67: Looks like 500 still isn't enough for some of these descriptions. Maybe we should shorten them instead of increasing the limit, though?

iblancasa:
Ok, I see. Since the problem is with the generated docs... how about increasing the limit just for the docs? I can generate the CRDs without a limit in the description and generate the documentation from there. The CRDs shipped as part of the bundle will remain intact.

Also, @swiatekm-sumo suggestion sounds like something we should do too.

What do you think?

pavolloffay:
The max description length was decreased in https://github.com/open-telemetry/opentelemetry-operator/pull/1981. I would prefer a solution where we could decide which fields would be truncated.

iblancasa:
> The max description length was decreased in #1981. I would prefer a solution where we could decide which fields would be truncated.

I'm not sure that can be done except doing them shorter.

Makefile
```diff
@@ -479,8 +480,11 @@ bundle-push:
 api-docs: crdoc kustomize
 	@{ \
 	set -e ;\
+	TMP_MANIFEST_DIR=$$(mktemp -d) ; \
+	cp -r config/crd/* $$TMP_MANIFEST_DIR; \
+	$(MAKE) CRD_OPTIONS=$(CRD_OPTIONS),maxDescLen=800 MANIFEST_DIR=$$TMP_MANIFEST_DIR/bases manifests ;\
```
pavolloffay, line 485: Why here is `,maxDescLen=800` and the length is also defined in `CRD_OPTIONS ?= "crd:generateEmbeddedObjectMeta=true,maxDescLen=200"` ?

iblancasa: This will call the `manifests` target using `CRD_OPTIONS="crd:generateEmbeddedObjectMeta=true,maxDescLen=200",maxDescLen=800`. The second `maxDescLen` overrides the first one. By doing this, we can:
- Reuse the value of `CRD_OPTIONS`
- Use the new value for `maxDescLen` without polluting the bundle

pavolloffay: Do we actually need the max length for the api-docs?

swiatekm-sumo (approved):
@jaronoff97 @pavolloffay we need to merge this if we want to actually release v1beta1.

jaronoff97:
i'm good to merge this, @pavolloffay any objections?

pavolloffay:
@iblancasa could you please rebase this one?

pavolloffay (approved):

PR #2564, opened 2024-01-25
Use standard label selectors in target allocator config

swiatekm-sumo:
**Description:**
Use standard label selectors in target allocator config instead of plain maps. This is a step towards resolving #1907 in `v1alpha2` CRs, by using these selectors in the CRDs themselves as well.

I'd wanted to keep backwards compatibility with older target allocator versions to make upgrades easier, so I couldn't simply change the types of existing config fields. Instead, I decided to move all of these fields under the `prometheus_cr` section in the configuration, which imo is the more intuitive location for it anyway.

There's also a difference in semantics between our current `v1alpha1` CRD and standard behaviour for these selectors that I needed to handle. For selectors:

* a `nil` selector means "select nothing", which is also the default in prometheus-operator CRs
* an empty `{}` selector means "select everything"
* a selector with actual labels or expressions works as expected

However, for the label maps in our `v1alpha1` CRD, an empty or nil map means "select everything". As a result, I needed to make sure we put an empty selector in the configuration in that case. This is the reason they appear in the modified test code. In `v1alpha2`, we can change this behaviour to match prometheus-operator.

**Link to tracking Issue:** #1907

**Testing:**
Modified existing unit tests. Existing E2E tests cover the functionality.

### DIFF
--- a/internal/manifests/targetallocator/configmap.go
+++ b/internal/manifests/targetallocator/configmap.go
@@ -68,10 +68,20 @@ func ConfigMap(params manifests.Params) (*corev1.ConfigMap, error) {
 		prometheusCRConfig["scrape_interval"] = params.OtelCol.Spec.TargetAllocator.PrometheusCR.ScrapeInterval.Duration
 	}
 
+	prometheusCRConfig["service_monitor_selector"] = &metav1.LabelSelector{
+		MatchLabels: params.OtelCol.Spec.TargetAllocator.PrometheusCR.ServiceMonitorSelector,
+	}
+	// The below instruction is here for compatibility with the previous target allocator version
+	// TODO: Drop it after 3 more versions
 	if params.OtelCol.Spec.TargetAllocator.PrometheusCR.ServiceMonitorSelector != nil {
 		taConfig["service_monitor_selector"] = &params.OtelCol.Spec.TargetAllocator.PrometheusCR.ServiceMonitorSelector
 	}
 
+	prometheusCRConfig["pod_monitor_selector"] = &metav1.LabelSelector{
+		MatchLabels: params.OtelCol.Spec.TargetAllocator.PrometheusCR.PodMonitorSelector,
+	}
+	// The below instruction is here for compatibility with the previous target allocator version
+	// TODO: Drop it after 3 more versions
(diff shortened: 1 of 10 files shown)


internal/manifests/targetallocator/configmap.go
```diff
@@ -68,10 +68,20 @@ func ConfigMap(params manifests.Params) (*corev1.ConfigMap, error) {
 		prometheusCRConfig["scrape_interval"] = params.OtelCol.Spec.TargetAllocator.PrometheusCR.ScrapeInterval.Duration
 	}
 
+	prometheusCRConfig["service_monitor_selector"] = &metav1.LabelSelector{
```
jaronoff97, line 71: just to confirm my understanding here... by not having the nil check we follow what prometheus expects i.e. when service/pod monitor selector is nil nothing will be selected? whereas before we just didn't set it and assumed everything?

jaronoff97 (approved):

internal/manifests/targetallocator/configmap.go
swiatekm-sumo: > just to confirm my understanding here... by not having the nil check we follow what prometheus expects i.e. when service/pod monitor selector is nil nothing will be selected? whereas before we just didn't set it and assumed everything?

Prometheus' (and LabelSelector in general) behaviour is that `nil` == `nothing` and `empty` == `everything`. We on the other hand, took the map being empty or nil to mean `everything`, and there wasn't any way to set it to be `nothing`.

I've modified the configmap tests (and the builder and reconciliation tests) to show the empty label selectors. E2E tests check that this works as well - I know because I had a bug here earlier, and they caught it.

swiatekm-sumo: In a sense, I'm moving the logic from target allocator here. If you look at the target allocator changes, it used to have this for the Prometheus Spec:

```go
ServiceMonitorSelector: &metav1.LabelSelector{
	MatchLabels: cfg.ServiceMonitorSelector,
},
```

Now we have this here, accomplishing the same thing.

jaronoff97: gotcha, thanks for explaining.

PR #2565, opened 2024-01-25
Generate TargetAllocator CR from Collector CR

swiatekm-sumo:
**Description:**
Generate a `v1alpha2.TargetAllocator` based on `v1alpha1.Collector`. The change doesn't have any effect, it simply adds the function and modifies the TargetAllocator CRD.

One noteworthy change that I'd like some feedback on is the addition of scrape configs to the TargetAllocator CRD. This is something we've missed, and which is necessary for the CR to have feature parity with the subresource. Unfortunately this makes things a bit messy, as it's necessary to parse the Collector configuration to get these values, but I don't think there's any alternative.

Currently missing from this PR:
- [x] more unit tests
- [x] rebase on #2532
- [x] rebase on #2564 

**Link to tracking Issue:** #2516 

**Testing:**
Aside from unit tests, I also have another change which converts all TA manifest generation to use the TargetAllocator CR, and all E2E tests pass there.


apis/v1alpha2/targetallocator_types.go
```diff
@@ -35,8 +39,12 @@ type TargetAllocatorSpec struct {
 	// +optional
 	// +kubebuilder:default:=relabel-config
 	FilterStrategy TargetAllocatorFilterStrategy `json:"filterStrategy,omitempty"`
-	// ServiceAccount indicates the name of an existing service account to use with this instance. When set,
-	// the operator will not automatically create a ServiceAccount for the TargetAllocator.
+	// ScrapeConfigs define static Prometheus scrape configurations for the target allocator.
+	// To use dynamic configurations from ServiceMonitors and PodMonitors, see the PrometheusCR section.
+	// +optional
+	// +listType=atomic
+	ScrapeConfigs []ScrapeConfig `json:"scrapeConfigs,omitempty"`
```
jaronoff97, line 46: this makes sense to add, but I also wonder if having the collector selector above should hint the operator to pull the configmap for that collector? Is that too difficult / cumbersome / confusing?

apis/v1alpha2/targetallocator_types.go
```diff
@@ -82,6 +90,46 @@ type TargetAllocatorStatus struct {
 	Messages []string `json:"messages,omitempty"`
 }
 
+// ScrapeConfig is a Prometheus scrape config definition.
+type ScrapeConfig map[interface{}]interface{}
```
jaronoff97, line 94: should we use the upstream prometheus type?

jaronoff97 (commented):
a few thoughts

apis/v1alpha2/targetallocator_types.go
swiatekm-sumo: That should also work, the way I imagined it is that the operator would set this field based on the annotation. I'm not actually sure what the idiomatic solution to this kind of problem is, but I'm defaulting to being more explicit.

apis/v1alpha2/targetallocator_types.go
swiatekm-sumo: I thought about it, but that type is actually quite complicated, and we specifically decided not to parse the prometheus receiver config into upstream Prometheus structs in the past. We'd be forced to do that again here.

pavolloffay: Can it be changed to `map[string]interface{}` which is generally easier to work with

apis/v1alpha2/targetallocator_types.go
pavolloffay: Could we at least put a link to p8s docs where people can find how to define scrapeconfigs?

apis/v1alpha2/targetallocator_types.go
swiatekm-sumo: I think so. I originally had it as that, but ran into some problems with kubebuilder generation which were ultimately unrelated. Let me try again.

jaronoff97 (approved):
this LGTM. @pavolloffay thoughts?

swiatekm-sumo:
I'm wondering if it wouldn't be better to first merge #2621, then add a v1alpha2 version of the embedded TA spec, take care of conversion, and then generating the CR will be much simpler. WDYT?

jaronoff97:
Agreed, i think this becomes simpler with #2621

swiatekm-sumo:
I'd also like to merge #2623 first, then the conversion in this PR becomes trivial.

swiatekm-sumo:
This is now ready for review. I've rebased my manifest generation changes on it, and everything works there as well.

jaronoff97 (approved):

swiatekm-sumo:
I did some reconciliation tests and ended up switching to `AnyConfig` for the ScrapeConfigs field. I couldn't make it any nicer, so in the end there wasn't any reason to have two implementations of the same thing.

jaronoff97 (approved):

pavolloffay (approved):

PR #2575, opened 2024-01-29
Fixed Otel Controller - Auxiliares resources deletion

### LINKED ISSUE #2568: HPA is only added with autoscaler setting, the reverse is ignored
### Component(s)

_No response_

## Description
Adding the autoscaler to a collector deploys a HPA crd, but it does not remove the CRD if the autoscaler section is removed.
operator logs show this instead.
`hpa field is unset in Spec, skipping autoscaler creation`

## Steps to Reproduce
Deploy a Collector with the autoscaler in the spec.
Validate HPA is deployed for collector.
Remove autoscaler from spec.
HPA CRD remains.

## Expected Result
HPA CRD for collector to get deleted.

## Actual Result
HPA CRD remains.

### Kubernetes Version

v1.25.15-gke.1115000

### Operator version

v0.92.1

### Collector version

v0.92.0

## Environment
OS: (e.g., "Ubuntu 20.04")

### Log output

```shell
hpa field is unset in Spec, skipping autoscaler creation
```

### Additional context

_No response_

### LINKED ISSUE #2587: PVC does not increase after initially set
### Component(s)

_No response_

## Description

After deploying a collector with a pvc of 5Gi, Increasing it later doesn't not increase the PVC.

## Steps to Reproduce
Deploy Collector and set the volumeClaimTemplates storage to `5Gi`.
Update the CRD and increase to `50Gi`
Check PVCs to validate new size.

## Expected Result

PVCs increase to `50Gi`

## Actual Result
PVSs remain the same, `5Gi`.

### Kubernetes Version

v1.27.7-gke.1121000

### Operator version

v0.92.1

### Collector version

v0.92.0

## Environment
GKE

### Log output

_No response_

### Additional context

I have to manually delete the PVC before they get created with the new size.

### LINKED ISSUE #2651: Disabling ingress does not remove the created ingress objects
### Component(s)

collector

## Description

If you disable a previously enabled ingress for a Collector the ingress resource will not be deleted from the cluster.

## Steps to Reproduce

1. Create a collector with an ingress
2. Remove the ingress and apply the collector to the cluster

## Expected Result

The previously created ingress resource should be removed.

## Actual Result

The previously created ingress is not removed.

### Kubernetes Version

v1.28.3-gke.1286000

### Operator version

0.93.0

### Collector version

n/a

## Environment
Cloud: GKE

### Log output

_No response_

### Additional context

_No response_

yuriolisa:

Fixed HPA deletion
Resolves #2568 
Resolves #2651 
Resolves #2587 

Added new e2e test to check this step.
- Created a new bundle due to the new RBAC for `PersistentVolumes` and `PersistentVolumesClaims`.
-  Fixed the labels used by Ingress resources and their unit tests.
- Created new function `findOtelOwnedObjects` which will find all the owned objects and serve as an auxiliary resource to the  `reconcileDesiredObjects` function to keep or delete the expected objects.
- Changed the parameters of the function `HorizontalPodAutoscaler` from `client.Object`  to `*autoscalingv2.HorizontalPodAutoscaler` to keep the same pattern we have been using for another resources.
- Added the networkv1 package to the scheme. (`networkingv1.AddToScheme(scheme)`). 
- Update Chainsaw from 0.1.4 to 0.1.7 in order to get the `patch` operation enabled.
- Created new e2e tests for `Ingress` and `HorizontalPodAutoscalers` to test the deletion process.
- Added the testdata.ServiceMonitorCRD, testdata.PodMonitorCRD to the  `controllers/suite_test.go`


swiatekm-sumo (approved):

internal/manifests/collector/horizontalpodautoscaler.go
```diff
@@ -18,19 +18,18 @@ import (
...
-	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
 	"github.com/open-telemetry/opentelemetry-operator/internal/manifests"
 	"github.com/open-telemetry/opentelemetry-operator/internal/manifests/manifestutils"
 	"github.com/open-telemetry/opentelemetry-operator/internal/naming"
 )
 
-func HorizontalPodAutoscaler(params manifests.Params) client.Object {
+func HorizontalPodAutoscaler(params manifests.Params) *autoscalingv2.HorizontalPodAutoscaler {
```
jaronoff97, line 28: why was this preventing the deletion from happening?

frzifus: @yuriolisa did you try to run the test without the interface removal?

yuriolisa: Yes, I did it. Tested with the latest version and applied and deleted the e2e-autoscale tests.

controllers/common.go
```diff
@@ -76,9 +87,71 @@ func BuildOpAMPBridge(params manifests.Params) ([]client.Object, error) {
...
+	otelCol, err := convert.V1Alpha1to2(params.OtelCol)
+	if err != nil {
+		return nil, err
+	}
+
+	listOps := &client.ListOptions{
+		Namespace:     otelCol.Namespace,
+		LabelSelector: labels.SelectorFromSet(manifestutils.Labels(otelCol.ObjectMeta, naming.Collector(otelCol.Name), otelCol.Spec.Image, collector.ComponentOpenTelemetryCollector, params.Config.LabelsFilter())),
+	}
+	hpaList := &autoscalingv2.HorizontalPodAutoscalerList{}
```
jaronoff97, line 101: do we also need to do PDBs?

tests/e2e-autoscale/autoscale/04-error.yaml
```diff
@@ -0,0 +1,28 @@
+---
+apiVersion: autoscaling/v2
+kind: HorizontalPodAutoscaler
```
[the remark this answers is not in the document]
yuriolisa: RIght!

controllers/opentelemetrycollector_controller.go
```diff
@@ -139,6 +148,7 @@ func (r *OpenTelemetryCollectorReconciler) SetupWithManager(mgr ctrl.Manager) er
 		Owns(&appsv1.Deployment{}).
 		Owns(&appsv1.DaemonSet{}).
 		Owns(&appsv1.StatefulSet{}).
+		Owns(&networkingv1.Ingress{}).
```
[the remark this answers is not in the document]
yuriolisa: So am I.

controllers/common.go
yuriolisa: Sure, good catch.

controllers/common.go
```diff
@@ -76,9 +87,78 @@ func BuildOpAMPBridge(params manifests.Params) ([]client.Object, error) {
...
+		routesList := &routev1.RouteList{}
+		err := r.List(ctx, routesList, listOps)
+		if err != nil {
+			return nil, fmt.Errorf("Error listing Routes: %w", err)
+		}
+		for i := range routesList.Items {
+			ownedObjects[routesList.Items[i].GetUID()] = &routesList.Items[i]
+		}
+	}
+	if params.OtelCol.Spec.PodDisruptionBudget != nil {
```
jaronoff97, line 144: i don't think we want the nil check here, as we actually care about the case where the PDB is nil but it previously wasn't

jaronoff97 (commented):
one comment. Otherwise, were you able to confirm locally that this does as we intend?

controllers/common.go
pavolloffay: +1

Then shouldn't be removed as well the check for p8s objects (` params.OtelCol.Spec.Observability.Metrics.EnableMetrics && featuregate.PrometheusOperatorIsAvailable.IsEnabled()`)?

controllers/common.go
```diff
@@ -76,9 +87,78 @@ func BuildOpAMPBridge(params manifests.Params) ([]client.Object, error) {
...
+	}
+
+	return ownedObjects, nil
+}
+
 // reconcileDesiredObjects runs the reconcile process using the mutateFn over the given list of objects.
-func reconcileDesiredObjects(ctx context.Context, kubeClient client.Client, logger logr.Logger, owner metav1.Object, scheme *runtime.Scheme, desiredObjects ...client.Object) error {
+func reconcileDesiredObjects(ctx context.Context, kubeClient client.Client, logger logr.Logger, owner metav1.Object, scheme *runtime.Scheme, desiredObjects []client.Object, ownedObjects map[types.UID]client.Object) error {
 	var errs []error
+	pruneObjects := ownedObjects
```
pavolloffay, line 161: Why there is this assignment? I didn't find `owhnedObjects` being used in the function. 

Why  `ownedObjects map[types.UID]client.Object` is a map? I didn't find the map was actually useful

yuriolisa: I did this assignment to have an auxiliary map to control the flow for which objects should be deleted. However, you're right, let me change it. 
Regarding the map, I'm using it to store and compare the objects by UID and proceed with the reconciliation. Do you have any suggestions?

tests/e2e-autoscale/autoscale/04-delete.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: kuttl.dev/v1beta1
```
jaronoff97, line 1: we're on chainsaw now :)

yuriolisa: Thanks for the heads-up. I've just submitted the e2e using Chainsaw instead.

controllers/suite_test.go
```diff
@@ -120,6 +121,11 @@ func TestMain(m *testing.M) {
 		os.Exit(1)
 	}
 
+	if err = autoscalingv2.AddToScheme(testScheme); err != nil {
```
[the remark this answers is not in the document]
yuriolisa: Removed it.

controllers/common.go
```diff
@@ -76,8 +87,75 @@ func BuildOpAMPBridge(params manifests.Params) ([]client.Object, error) {
 	return resources, nil
 }
 
+func (r *OpenTelemetryCollectorReconciler) findOtelOwnedObjects(ctx context.Context, params manifests.Params) (map[types.UID]client.Object, error) {
```
jaronoff97, line 90: nit: this method should be moved to the collector controller rather than be in common here IMO

controllers/common.go
```diff
@@ -76,8 +87,75 @@ func BuildOpAMPBridge(params manifests.Params) ([]client.Object, error) {
...
+	hpaList := &autoscalingv2.HorizontalPodAutoscalerList{}
+	err := r.List(ctx, hpaList, listOps)
+	if err != nil {
+		return nil, fmt.Errorf("error listing HorizontalPodAutoscalers: %w", err)
+	}
+	for i := range hpaList.Items {
+		ownedObjects[hpaList.Items[i].GetUID()] = &hpaList.Items[i]
+	}
+
+	if params.OtelCol.Spec.Observability.Metrics.EnableMetrics && featuregate.PrometheusOperatorIsAvailable.IsEnabled() {
```
jaronoff97, line 106: in this case we should just check for the featuregate, not the enableMetrics check. If someone were to enable and then disable we want to clean up the metrics.

jaronoff97 (commented):
just two small things, otherwise this LGTM!! Thank you so much Yuri.

controllers/common.go
yuriolisa: @jaronoff97, I've changed what you suggested and added the references for the PersistentVolumes and their claims, and e2e tests for Volumes.

pavolloffay:
The CI is still failing :/

pavolloffay (approved):

Makefile
```diff
@@ -198,7 +198,7 @@ generate: controller-gen
 # end-to-tests
 .PHONY: e2e
 e2e: chainsaw
-	$(CHAINSAW) test --test-dir ./tests/e2e
+	$(CHAINSAW) test --test-dir ./tests/e2e/ingress
```
swiatekm-sumo, line 201: I think this might've been committed by mistake.

yuriolisa: Right, but already removed.

jaronoff97 (approved):

jaronoff97:
@yuriolisa thank you so much for your heroics here. I really appreciate the work and all the fixes that come from this.

PR #2578, opened 2024-01-29
Opamp Bridge can read env variables on config generation

### LINKED ISSUE #2577: Environment variable interpolation in bridge config
### Component(s)

opamp bridge

I want to populate headers in the opamp bridge from environment variables. The opamp bridge should be able to substitute environment variables during config generation

The [LoadFromFile](https://github.com/open-telemetry/opentelemetry-operator/blob/ee8db9d4641ab0d0b001748786c2cdf875e227d6/cmd/operator-opamp-bridge/config/config.go#L267) function should substitute templated environment variable names with their assigned value using the go `os` package

_No response_

### Additional context

_No response_

jdcrouse:
The opamp bridge now substitutes env variables in its yaml config with their respective value from the local environment upon startup

**Link to tracking Issue:** closes #2577

**Testing:** I added a unit test which includes env variables as args and expects that they are parsed into headers according to the updated `agentwithheaders.yaml`

**Documentation:** n/a

### DIFF
--- a/cmd/operator-opamp-bridge/config/config.go
+++ b/cmd/operator-opamp-bridge/config/config.go
@@ -18,7 +18,6 @@ import (
 	"crypto/rand"
 	"errors"
 	"fmt"
-	"github.com/oklog/ulid"
 	"io/fs"
 	"net/url"
 	"os"
@@ -26,6 +25,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"github.com/oklog/ulid/v2"
 	opampclient "github.com/open-telemetry/opamp-go/client"
 	"github.com/open-telemetry/opamp-go/protobufs"
 	"github.com/spf13/pflag"
--- a/cmd/operator-opamp-bridge/config/config.go
+++ b/cmd/operator-opamp-bridge/config/config.go
@@ -269,11 +269,11 @@ func LoadFromFile(cfg *Config, configFile string) error {
 	if err != nil {
 		return err
 	}
-	envExpandedYaml := os.ExpandEnv(string(yamlFile))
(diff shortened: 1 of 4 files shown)


cmd/operator-opamp-bridge/config/config.go
```diff
@@ -18,14 +18,14 @@ import (
 	"crypto/rand"
 	"errors"
 	"fmt"
+	"github.com/oklog/ulid"
```
jaronoff97: This import is in the wrong place

jaronoff97: I think we also probs want the v2

cmd/operator-opamp-bridge/config/config.go
```diff
@@ -269,7 +269,11 @@ func LoadFromFile(cfg *Config, configFile string) error {
 	if err != nil {
 		return err
 	}
-	if err = yaml.Unmarshal(yamlFile, cfg); err != nil {
+	envExpandedYaml := []byte(os.ExpandEnv(string(yamlFile)))
+	if err = yaml.UnmarshalStrict(envExpandedYaml, cfg); err != nil {
+		return fmt.Errorf("error unmarshaling YAML: %w", err)
+	}
```
jaronoff97, line 275: No need for this extra call

cmd/operator-opamp-bridge/config/config.go
jdcrouse: yeah this was some goland side effect i didnt intend to change, i reverted that bit

jaronoff97 (approved):

jaronoff97 (approved):

PR #2579, opened 2024-01-29
Document empty or nil prometheus service/pod monitors

pavolloffay:

### DIFF
--- a/apis/v1alpha1/opentelemetrycollector_types.go
+++ b/apis/v1alpha1/opentelemetrycollector_types.go
@@ -380,11 +380,13 @@ type OpenTelemetryTargetAllocatorPrometheusCR struct {
 	// PodMonitors to be selected for target discovery.
 	// This is a map of {key,value} pairs. Each {key,value} in the map is going to exactly match a label in a
 	// PodMonitor's meta labels. The requirements are ANDed.
+	// Empty or nil map matches all pod monitors.
 	// +optional
 	PodMonitorSelector map[string]string `json:"podMonitorSelector,omitempty"`
 	// ServiceMonitors to be selected for target discovery.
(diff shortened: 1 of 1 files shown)


apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -380,11 +380,13 @@ type OpenTelemetryTargetAllocatorPrometheusCR struct {
 	// PodMonitors to be selected for target discovery.
 	// This is a map of {key,value} pairs. Each {key,value} in the map is going to exactly match a label in a
 	// PodMonitor's meta labels. The requirements are ANDed.
+	// Empty or nil map matches all pod monitors.
```
jaronoff97, line 383: an explicit nil means select nothing iirc.

jaronoff97 (commented):
i think we need a make generate here for the docs/types

apis/v1alpha1/opentelemetrycollector_types.go
swiatekm-sumo: It doesn't, there's no way to set nothing with these label maps. I think you're mixing this up with proper LabelSelectors in `v1alpha2`.

pavolloffay:
> i think we need a make generate here for the docs/types

`make api-docs bundle  generate ` does not make any change

pavolloffay:
And all checks are passing

apis/v1alpha1/opentelemetrycollector_types.go
pavolloffay: @swiatekm-sumo so is it correct? Can it be merged or do I need to make any change?

jaronoff97: 🤦 i am mixing those up.

jaronoff97 (approved):

PR #2602, opened 2024-02-03
Add Dependabot config for NodeJS autoinstrumentation

akselleirv:
**Description:** 

Added Dependabot for bumping NodeJS updates for the autoinstrumentation with grouping on bump type.

**Testing:** 

Tested the config on my fork. Please see PRs there for how it look like:

- https://github.com/akselleirv/opentelemetry-operator/pull/2
- https://github.com/akselleirv/opentelemetry-operator/pull/5

### DIFF
--- a/.chloggen/cleanup-roles.yaml
+++ b/.chloggen/cleanup-roles.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: collector
+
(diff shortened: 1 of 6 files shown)


TylerHelmuth:
This is really interesting.  We've typically only done a release when the nodejs API/SDK has a bump, but I guess there is nothing stopping us from bumping when only the instrumentation has a release.

pavolloffay:
Can any version be independently bumped https://github.com/open-telemetry/opentelemetry-operator/blob/main/autoinstrumentation/nodejs/package.json? 


We might also need to change how we are versioning and when we are releasing the node auto-instrumentation https://github.com/open-telemetry/opentelemetry-operator/blob/main/.github/workflows/publish-autoinstrumentation-nodejs.yaml#L29. Right now it is released only if `sdk-node` is released.

Issue #2592, opened 2024-02-01
v1alpha2: Minimise signatures of object generation functions


frzifus:
### Component(s)

operator

### Describe the issue you're reporting

Minimise signatures of object generation functions and clean up `convert.V1Alpha1to2(params.OtelCol)` calls.

Comments: 
- https://github.com/open-telemetry/opentelemetry-operator/pull/2532#discussion_r1474773719
- https://github.com/open-telemetry/opentelemetry-operator/pull/2532#discussion_r1474776362

PR #2621, opened 2024-02-12
Clean up signatures

jaronoff97:
**Description:** 
This PR cleans up all the places that need to convert the v1alpha1 type to the v1alpha2 by minimizing the calls to the conversion function. Other than the type changes all over, the main thing that changed is the tests which previously built a whole params object, which really only needed to build a collector object.

**Link to tracking Issue:** closes https://github.com/open-telemetry/opentelemetry-operator/issues/2592

**Testing:** unit testing

**Documentation:** N/A

### DIFF
--- a/controllers/reconcile_test.go
+++ b/controllers/reconcile_test.go
@@ -69,24 +69,15 @@ var (
 
 type check[T any] func(t *testing.T, params T)
 
-func newParamsAssertNoErr(t *testing.T, taContainerImage string, file string) v1alpha1.OpenTelemetryCollector {
-	p, err := newParams(taContainerImage, file)
-	assert.NoError(t, err)
-	if len(taContainerImage) == 0 {
-		p.Spec.TargetAllocator.Enabled = false
-	}
-	return p
-}
-
 func TestOpenTelemetryCollectorReconciler_Reconcile(t *testing.T) {
-	addedMetadataDeployment := paramsWithMode(v1alpha1.ModeDeployment)
+	addedMetadataDeployment := testCollectorWithMode(v1alpha1.ModeDeployment)
 	addedMetadataDeployment.Labels = map[string]string{
 		labelName: labelVal,
 	}
 	addedMetadataDeployment.Annotations = map[string]string{
 		annotationName: annotationVal,
 	}
-	deploymentExtraPorts := paramsWithModeAndReplicas(v1alpha1.ModeDeployment, 3)
+	deploymentExtraPorts := testCollectorWithModeAndReplicas(v1alpha1.ModeDeployment, 3)
 	deploymentExtraPorts.Spec.Ports = append(deploymentExtraPorts.Spec.Ports, extraPorts)
 	deploymentExtraPorts.Spec.DeploymentUpdateStrategy = appsv1.DeploymentStrategy{
 		RollingUpdate: &appsv1.RollingUpdateDeployment{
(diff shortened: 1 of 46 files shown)


controllers/reconcile_test.go
```diff
@@ -67,28 +67,28 @@ var (
 	}
 )
 
-type check func(t *testing.T, params manifests.Params)
+type check[T any] func(t *testing.T, params T)
 
-func newParamsAssertNoErr(t *testing.T, taContainerImage string, file string) manifests.Params {
+func newParamsAssertNoErr(t *testing.T, taContainerImage string, file string) v1alpha1.OpenTelemetryCollector {
```
swiatekm-sumo, line 72: We should rename this function if we're changing the return type.

controllers/suite_test.go
```diff
@@ -215,51 +215,44 @@ func TestMain(m *testing.M) {
 	os.Exit(code)
 }
 
-func paramsWithMode(mode v1alpha1.Mode) manifests.Params {
+func paramsWithMode(mode v1alpha1.Mode) v1alpha1.OpenTelemetryCollector {
```
swiatekm-sumo, line 218: We should rename this function if we're changing the return type.

controllers/suite_test.go
```diff
@@ -215,51 +215,44 @@ func TestMain(m *testing.M) {
...
 	replicas := int32(2)
 	return paramsWithModeAndReplicas(mode, replicas)
 }
 
-func paramsWithModeAndReplicas(mode v1alpha1.Mode, replicas int32) manifests.Params {
+func paramsWithModeAndReplicas(mode v1alpha1.Mode, replicas int32) v1alpha1.OpenTelemetryCollector {
```
swiatekm-sumo, line 223: We should rename this function if we're changing the return type.

controllers/suite_test.go
```diff
@@ -215,51 +215,44 @@ func TestMain(m *testing.M) {
...
+			Mode:     mode,
 		},
-		Scheme:   testScheme,
-		Log:      logger,
-		Recorder: record.NewFakeRecorder(10),
 	}
 }
 
-func newParams(taContainerImage string, file string) (manifests.Params, error) {
+func newParams(taContainerImage string, file string) (v1alpha1.OpenTelemetryCollector, error) {
```
swiatekm-sumo, line 255: We should rename this function if we're changing the return type.

controllers/suite_test.go
```diff
@@ -270,95 +263,76 @@ func newParams(taContainerImage string, file string) (manifests.Params, error) {
...
+			Replicas: &replicas,
+			Config:   string(configYAML),
 		},
-		Scheme: testScheme,
-		Log:    logger,
 	}, nil
 }
 
-func paramsWithHPA(minReps, maxReps int32) manifests.Params {
+func paramsWithHPA(minReps, maxReps int32) v1alpha1.OpenTelemetryCollector {
```
swiatekm-sumo, line 298: We should rename this function if we're changing the return type.

controllers/suite_test.go
```diff
@@ -270,95 +263,76 @@ func newParams(taContainerImage string, file string) (manifests.Params, error) {
...
 			},
 		},
-		Scheme:   testScheme,
-		Log:      logger,
-		Recorder: record.NewFakeRecorder(10),
 	}
 }
 
-func paramsWithPolicy(minAvailable, maxUnavailable int32) manifests.Params {
+func paramsWithPolicy(minAvailable, maxUnavailable int32) v1alpha1.OpenTelemetryCollector {
```
swiatekm-sumo, line 335: We should rename this function if we're changing the return type.

swiatekm-sumo (approved):
LGTM, some functions should be renamed but I'm fine approving this as-is.

controllers/suite_test.go
jaronoff97: agreed, will make the change :)

PR #2623, opened 2024-02-13
Add v1alpha2 embedded Target Allocator struct

swiatekm-sumo:
Add and use the embedded Target Allocator struct for v1alpha2. The changes mostly involve label selectors for Prometheus CR.

Worth noting that this involves a change in semantics for said selectors, as an empty selector now selects nothing, which is in line with the standard, but different from v1alpha1.


apis/v1alpha2/allocation_strategy.go
```diff
@@ -31,5 +31,5 @@ const (
 	TargetAllocatorAllocationStrategyConsistentHashing TargetAllocatorAllocationStrategy = "consistent-hashing"
 
 	// TargetAllocatorFilterStrategyRelabelConfig targets will be consistently drops targets based on the relabel_config.
-	TargetAllocatorFilterStrategyRelabelConfig TargetAllocatorFilterStrategy = "consistent-hashing"
+	TargetAllocatorFilterStrategyRelabelConfig TargetAllocatorFilterStrategy = "relabel-config"
```
swiatekm-sumo, line 34: This is a small bug fix, the value was simply wrong.

apis/v1alpha2/opentelemetrycollector_types.go
```diff
@@ -165,6 +166,84 @@ type OpenTelemetryCollectorSpec struct {
 	DeploymentUpdateStrategy appsv1.DeploymentStrategy `json:"deploymentUpdateStrategy,omitempty"`
 }
 
+// TargetAllocatorEmbedded defines the configuration for the Prometheus target allocator, embedded in the
+// OpenTelemetryCollector spec.
+type TargetAllocatorEmbedded struct {
```
jaronoff97, line 171: should we have this use the common fields?

jaronoff97 (commented):
one question, otherwise looking good!

apis/v1alpha2/opentelemetrycollector_types.go
swiatekm-sumo: I'd rather do that in the TargetAllocator CR proper, and slowly deprecate most of the common fields in this embedded struct. For the migration, I'd like to keep them the same for simplicity.

jaronoff97 (approved):
The semantics change is fine to me IMO, when we actually start using this in #2565 we should include this in a changelog maybe.

PR #2625, opened 2024-02-14
Ensure unique names for ClusterRoles in e2e tests

swiatekm-sumo:
ClusterRoles with the same name in different tests can cause hard to debug test failures due to race conditions. With this fix, all ClusterRoles have unique names.


pavolloffay (approved):
Could we add cleanup of those roles as well?

swiatekm-sumo:
> Could we add cleanup of those roles as well?

I *think* kuttl cleans them up. It has a `skipDelete` flag which is disabled by default.

PR #2607, opened 2024-02-08
chore: migrate from kuttl to chainsaw

eddycharly:
This PR migrates e2e tests from kuttl to chainsaw.

Chainsaw was created because kuttl stopped to be actively maintained (it's actually a single maintainer now).

We are also solving most of the kuttl limitations that exist today and wrote a command to automatically migrate from kuttl to chainsaw. It supercharges e2e tests with better logs, config maps assertions, [assertions trees](https://kyverno.io/blog/2023/12/13/kyverno-chainsaw-exploring-the-power-of-assertion-trees/) and many more things 💪

This is a relatively young project but it has already been adopted by a couple of e2e tests heavy projects (kyverno, keptn, fairwinds rbac-manager, grafana-operator, redis-operator and others).

To give more context, you can see similar PRs at:
- https://github.com/keptn/lifecycle-toolkit/pull/3000
- https://github.com/operator-framework/operator-sdk/pull/6676
- https://github.com/grafana/tempo-operator/pull/786 in this PR the tempo-operator maintainers asked some good questions and I explained the current status, that I suggest taking a look at.

Sadly, [kuttl](https://github.com/kudobuilder/kuttl) haven't done a release since jan 2023.

Here you can find a good video showcasing chainsaw: https://www.youtube.com/watch?v=Ejof-wtAdQM&ab_channel=CamilaMacedo

To make things more concrete, there are a couple of issues in this repo, a few files are not valid and kuttl doesn't complain because unmarshalling is not strict so the non-valid fields are simply ignored. This is a problem because the behaviour of the tests is not the expected one, for example:
- https://github.com/open-telemetry/opentelemetry-operator/blob/c534eafdf30e7bd8c484ea70ffc7d5a6a9294e41/tests/e2e-opampbridge/opampbridge/00-assert.yaml#L67-L70
- https://github.com/open-telemetry/opentelemetry-operator/blob/94c8420ea6403ab359af9eb5c20ddef8d243a9d2/tests/e2e-prometheuscr/create-sm-prometheus-exporters/07-delete.yaml#L6-L8

If you need more infos :point_down:
- GitHub: https://github.com/kyverno/chainsaw
- Docs: https://kyverno.github.io/chainsaw
- Slack: https://kubernetes.slack.com/archives/C067LUFL43U

PR #2630, opened 2024-02-15
chore: migrate from kuttl to chainsaw

eddycharly:
This PR migrates e2e tests from kuttl to chainsaw.

Chainsaw was created because kuttl stopped to be actively maintained (it's actually a single maintainer now).

We are also solving most of the kuttl limitations that exist today and wrote a command to automatically migrate from kuttl to chainsaw. It supercharges e2e tests with better logs, config maps assertions, [assertions trees](https://kyverno.io/blog/2023/12/13/kyverno-chainsaw-exploring-the-power-of-assertion-trees/) and many more things 💪

This is a relatively young project but it has already been adopted by a couple of e2e tests heavy projects (kyverno, keptn, fairwinds rbac-manager, grafana-operator, redis-operator and others).

To give more context, you can see similar PRs at:
- https://github.com/keptn/lifecycle-toolkit/pull/3000
- https://github.com/operator-framework/operator-sdk/pull/6676
- https://github.com/grafana/tempo-operator/pull/786 in this PR the tempo-operator maintainers asked some good questions and I explained the current status, that I suggest taking a look at.

Sadly, [kuttl](https://github.com/kudobuilder/kuttl) haven't done a release since jan 2023.

Here you can find a good video showcasing chainsaw: https://www.youtube.com/watch?v=Ejof-wtAdQM&ab_channel=CamilaMacedo

To make things more concrete, there are a couple of issues in this repo, a few files are not valid and kuttl doesn't complain because unmarshalling is not strict so the non-valid fields are simply ignored. This is a problem because the behaviour of the tests is not the expected one, for example:
- https://github.com/open-telemetry/opentelemetry-operator/blob/c534eafdf30e7bd8c484ea70ffc7d5a6a9294e41/tests/e2e-opampbridge/opampbridge/00-assert.yaml#L67-L70
- https://github.com/open-telemetry/opentelemetry-operator/blob/94c8420ea6403ab359af9eb5c20ddef8d243a9d2/tests/e2e-prometheuscr/create-sm-prometheus-exporters/07-delete.yaml#L6-L8

If you need more infos :point_down:
- GitHub: https://github.com/kyverno/chainsaw
- Docs: https://kyverno.github.io/chainsaw
- Slack: https://kubernetes.slack.com/archives/C067LUFL43U


eddycharly:
Replaces https://github.com/open-telemetry/opentelemetry-operator/pull/2607 🙈

eddycharly:
For reference, we should really avoid this kind of thing https://github.com/open-telemetry/opentelemetry-operator/blob/b32c855e4af40eb6f7652dc9326fc16845c35689/tests/e2e/prometheus-config-validation/00-promreceiver-allocatorconfig.yaml#L16-L19

Chainsaw doesn't know anything about resources created with `kubectl` and therefore won't clean them up.

pavolloffay (approved):

swiatekm-sumo:
> For reference, we should really avoid this kind of thing
> 
> https://github.com/open-telemetry/opentelemetry-operator/blob/b32c855e4af40eb6f7652dc9326fc16845c35689/tests/e2e/prometheus-config-validation/00-promreceiver-allocatorconfig.yaml#L16-L19
> 
> Chainsaw doesn't know anything about resources created with `kubectl` and therefore won't clean them up.

We can use templating for it, right? That can happen as a follow-up PR though.

CONTRIBUTING.md
```diff
@@ -118,11 +118,11 @@ KUBEBUILDER_ASSETS=$(./bin/setup-envtest use -p path 1.23) go test ./pkg...
 
 ### End to end tests
 
-To run the end-to-end tests, you'll need [`kind`](https://kind.sigs.k8s.io) and [`kuttl`](https://kuttl.dev). Refer to their documentation for installation instructions.
+To run the end-to-end tests, you'll need [`kind`](https://kind.sigs.k8s.io) and [`chainsaw`](https://kyverno.github.io/chainsaw). Refer to their documentation for installation instructions.
```
swiatekm-sumo, line 121: I think we can just delete this line and adjust the next one. All of these tools are now installed automatically by the make targets.

Makefile
```diff
@@ -403,6 +404,18 @@ CRDOC = $(shell pwd)/bin/crdoc
 crdoc: ## Download crdoc locally if necessary.
 	$(call go-get-tool,$(CRDOC), fybrik.io/crdoc,v0.5.2)
 
+.PHONY: chainsaw
+chainsaw: ## Find or download chainsaw
+ifeq (, $(shell which chainsaw))
+	@{ \
+	set -e ;\
+	go install github.com/kyverno/chainsaw@v0.1.4 ;\
+	}
```
swiatekm-sumo, line 413: We have a small utility to make this nicer.
```suggestion
$(call go-get-tool,$(CHAINSAW), github.com/kyverno/chainsaw,v0.1.4)
```

Can we also move the version to its own variable, like the other tools?

eddycharly:
@swiatekm-sumo we can use templating indeed, will do in a follow-up

Makefile
```diff
@@ -403,6 +404,18 @@ CRDOC = $(shell pwd)/bin/crdoc
...
+CHAINSAW ?= $(GOBIN)/chainsaw
```
swiatekm-sumo, line 414: If we use the method suggested above, then this can simply be `$(LOCALBIN)/chainsaw`, and this whole conditional may be simplified, as we already detect whether chainsaw is installed globally before this point.

tests/e2e-instrumentation/instrumentation-apache-httpd/chainsaw-test.yaml
```diff
@@ -0,0 +1,36 @@
...
+        - openshift.io/sa.scc.uid-range=1000/1000
+        - --overwrite
+        entrypoint: kubectl
+    - command:
+        args:
+        - annotate
+        - namespace
+        - ${NAMESPACE}
+        - openshift.io/sa.scc.supplemental-groups=3000/1000
+        - --overwrite
```
swiatekm-sumo, line 25: We should be able to use a namespace template to apply these annotations, right?

swiatekm-sumo (commented):
LGTM overall, left some minor comments and questions about the Makefile and potential improvements to the tests using chainsaw's features.

tests/e2e-instrumentation/instrumentation-apache-httpd/chainsaw-test.yaml
eddycharly: Correct

Makefile
eddycharly: You are probably right, i didn't spend much time on this part. Will look at it in more depth after lunch (or in a follow up if it's ok).

Makefile
eddycharly: 👍

Makefile
swiatekm-sumo: Follow up is fine as well, e2e tests are modified often and it's not very practical to keep this PR blocked for minor issues.

eddycharly:
Ok, i'll address those comments in follow ups once this one merges :)

CONTRIBUTING.md
eddycharly: Follow up PR if it's ok.

Makefile
eddycharly: Follow up PR if it's ok.

jaronoff97:
Incredible work here @eddycharly !!! Thank you very much for your contributions. I'm looking forward to those follows to clean this up further.

Issue #2628, opened 2024-02-15
Operator Panic and CrashLoop for invalid prometheus exporter endpoint Collector


Starefossen:
## Description

Operator exists with the following panic and enters CrashLoopBackoff after applying an `OpenTelemetryCollector` (see bellow) and stays crashing until the `OpenTelemetryCollector` is deleted from the cluster. It is not possible to edit the `OpenTelemetryCollector` due to the failing webhook.

## Steps to Reproduce

Apply the following OpenTelemetryCollector resource:

```yaml
apiVersion: opentelemetry.io/v1alpha1
kind: OpenTelemetryCollector
metadata:
  name: opentelemetry-collector-management-internet
  namespace: my-system
spec:
  config: |
    receivers:
      otlp:
        protocols:
          http:
            endpoint: "http://localhost:4318/"
    processors:
      batch: {}
      memory_limiter:
        check_interval: 5s
        limit_mib: 4000
        spike_limit_mib: 500
      attributes:
        actions:
          - key: source
            value: internet
            action: insert
    exporters:
      prometheus:
        endpoint: prometheus
      otlp:
        endpoint: http://tempo:4317
        tls:
          insecure: true
      loki:
        endpoint: http://loki:3100/loki/api/v1/push
        tls:
          insecure: true
    service:
      pipelines:
        metrics:
          receivers: [ otlp ]
          processors: [ batch, memory_limiter ]
          exporters: [ prometheus ]
        traces:
          receivers: [ otlp ]
          processors: [ batch, memory_limiter ]
          exporters: [ otlp ]
        logs:
          receivers: [ otlp ]
          processors: [ batch, memory_limiter ]
          exporters: [ loki ]
  deploymentUpdateStrategy: {}
  ingress:
    route: {}
  managementState: managed
  mode: deployment
  observability:
    metrics: {}
  podDisruptionBudget:
    maxUnavailable: 1
  podSecurityContext:
    fsGroup: 65532
    runAsGroup: 65532
    runAsNonRoot: true
    runAsUser: 65532
    seccompProfile:
      type: RuntimeDefault
  ports:
  - name: otlphttp
    port: 4318
    protocol: TCP
    targetPort: 4318
  replicas: 1
  resources: {}
  securityContext:
    allowPrivilegeEscalation: false
    capabilities:
      drop:
      - ALL
    readOnlyRootFilesystem: true
    runAsNonRoot: true
    runAsUser: 65532
    seccompProfile:
[...]

PR #2653, opened 2024-02-21
Fix/config exporter endpoint panic

dexter0195:
**Description:** 
This PR fixes a CrashLoopBackoff caused by a panic occurring when, in the `spec.config`, the endpoint for the prometheus exporter is not valid. 

I integrated a simple check in `internal/manifests/collector/parser/exporter/exporter_prometheus.go` on the function `func (o *PrometheusExporterParser) Ports() ([]corev1.ServicePort, error)` to check if `singlePortFromConfigEndpoint` is returning a nil pointer.

**Link to tracking Issue(s):** https://github.com/open-telemetry/opentelemetry-operator/issues/2628

- Resolves: 2628

**Testing:** 
- adds unit tests on the `exporter` package for Prometheus.


internal/manifests/collector/parser/exporter/exporter_test.go
```diff
@@ -0,0 +1,80 @@
...
+			},
+			want: []v1.ServicePort{},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.testName, func(t *testing.T) {
+			if got, _ := tt.parser.Ports(); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("Ports(%v, = %v, want %v", tt.parser, got, tt.want)
+			}
```
swiatekm-sumo, line 77: We're using testify, so this is easier to accomplish via:
```suggestion
                        assert.Equal(t, tt.want, tt.parser.Ports())
```

swiatekm-sumo (approved):
LGTM, minor quibble about test assertions aside. Thank you for your contribution!

jaronoff97 (approved):

Issue #2608, opened 2024-02-08
`observability.metrics.DisablePrometheusAnnotations` does not work


verejoel:
## Description
Setting the new `DisablePrometheusAnnotations` config option to `true` in the latest version of the operator does not work.

## Steps to Reproduce
Operator version: 0.93.0
CRDs have been updated on K8s for this version.

Add the following stanza to an existing `OpenTelemetryCollector` and observe that the deployment is not rolled out, nor are annotations removed.
```yaml
spec:
  observability:
    metrics:
      DisablePrometheusAnnotations: true

```

I also tried redeploying the operator, to no avail.

## Expected Result
The prometheus scrape annotations should not be present on the deployment if this option is enabled.

## Actual Result
Prometheus scrape annotations are present, operator does not perform reconciliation

## Supplemental
Can we also fix the naming of this resource in the OpenAPI spec to not start with a capital letter?

### Kubernetes Version

1.28.3

### Operator version

1.93.0

### Collector version

Custom build

verejoel:
I tested this also with the `--feature-gates=operator.observability.prometheus` flag enabled with the same result

jaronoff97:
Are there any error logs from the operator?

pavolloffay:
The implementation of this feature is very simple https://github.com/open-telemetry/opentelemetry-operator/blob/769c9843728b56af6d2042784716c0a4ebd3c5f2/internal/manifests/collector/annotations.go#L31 

Could you pleae paste collector deployment with `DisablePrometheusAnnotations: true`

rivToadd:
I just ran a validation 
```
apiVersion: opentelemetry.io/v1alpha1
kind: OpenTelemetryCollector
metadata:
  name: my-collector
spec:
  mode: statefulset
  replicas: 1
  observability:
    metrics:
      enableMetrics: true
      # DisablePrometheusAnnotations: true
  config: |
    receivers:
      jaeger:
        protocols:
          grpc:
    processors:

    exporters:
      debug:

    service:
      pipelines:
        traces:
          receivers: [jaeger]
          processors: []
          exporters: [debug]
        ```
        give it a go with this spec. I was successful with it

jaronoff97:
an aside, I do think we should rename `DisablePrometheusAnnotations` to be lowercased to match all of the other fields we have.

NickAnge:
Hey everyone. 👋  We are facing the issue. 
[...]

PR #2664, opened 2024-02-26
bugfix: fixing disableprometheusannotation casing

rivToadd:

change disableprometheusannotation to lower camel case format

https://github.com/open-telemetry/opentelemetry-operator/issues/2608

- Resolves: #2608

### DIFF
--- a/internal/manifests/collector/annotations.go
+++ b/internal/manifests/collector/annotations.go
@@ -27,8 +27,25 @@ func Annotations(instance v1beta1.OpenTelemetryCollector) (map[string]string, er
 	// new map every time, so that we don't touch the instance's annotations
 	annotations := map[string]string{}
 
+	var enableAnnotations bool
+	// enableAnnotations is true only if both disable variables are false.
+	// DisableAutomaticPrometheusAnnotations takes precedence over DisablePrometheusAnnotations.
+	if instance.Spec.Observability.Metrics.DisableAutomaticPrometheusAnnotations {
+		enableAnnotations = !instance.Spec.Observability.Metrics.DisableAutomaticPrometheusAnnotations
+	} else if instance.Spec.Observability.Metrics.DisablePrometheusAnnotations {
+		enableAnnotations = !instance.Spec.Observability.Metrics.DisablePrometheusAnnotations
+	} else {
+		enableAnnotations = true
+	}
+	if enableAnnotations {
+		// Set default Prometheus annotations
+		annotations["prometheus.io/scrape"] = "true"
+		annotations["prometheus.io/port"] = "8888"
+		annotations["prometheus.io/path"] = "/metrics"
+	}
+
 	// Enable Prometheus annotations by default if DisablePrometheusAnnotations|DisableAutomaticPrometheusAnnotations is nil or true
-	if !instance.Spec.Observability.Metrics.DisablePrometheusAnnotations || !instance.Spec.Observability.Metrics.DisableAutomaticPrometheusAnnotations {
+	if !instance.Spec.Observability.Metrics.DisableAutomaticPrometheusAnnotations {
 		// Set default Prometheus annotations
 		annotations["prometheus.io/scrape"] = "true"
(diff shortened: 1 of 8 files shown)


Starefossen:
This should be noted as a breaking change (if merged).

yuriolisa:
@Toaddyan, please replace this change from v1alpha1 to v1beta1 and use the [conversion](https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/api/convert/v1alpha.go).

pavolloffay:
+1 we cannot merge this in v1alpha1 as it is a breaking change

jaronoff97:
@pavolloffay IMO this is an acceptable breaking change given the functionality as written will never work as intended. I would prefer we keep this as is to allow v1alpha1 users to use this feature as expected.

Starefossen:
I am not weight for or against the change but it should be properly marked. Maybe there should be a versioning policy established for the operator to explain what level of stability users can expect granted the api clearly stays alpha – but so does much within the cloud native ecosystem so at some point one becomes desensitized to the meaning of the word.

jaronoff97:
From the SIG meeting:
* on the topic of this PR
  * what will happen when after making this CR change?
  * if the upper case doesn't work, then this is a bug fix. If not this is a breaking change
  * the base assumption is that we should _never_ making a breaking change in the CRD
  * could we make an upgrade routine to convert from older version to newer version
  * **maybe define a second one with the lowercase...** (this is what we think should be done)
    * mark the old field deprecated
    * remove all references
    * add a warning for the old field
* For CRDs
  * a breaking change cannot be made without changing the version
  * stability around operator features
    * featureflags exist, but they're not exhaustive and they exist to change _behavior_ not really to opt-in/opt-out
    * we could document it, but it could rot pretty easily...
    * something like kubebuilder struct tags maybe?
  * we don't have stability guarantees currently, we will take this topic async.

internal/manifests/collector/annotations.go
```diff
@@ -27,8 +27,8 @@ func Annotations(instance v1beta1.OpenTelemetryCollector) (map[string]string, er
 	// new map every time, so that we don't touch the instance's annotations
 	annotations := map[string]string{}
 
-	// Enable Prometheus annotations by default if DisablePrometheusAnnotations is nil or true
-	if !instance.Spec.Observability.Metrics.DisablePrometheusAnnotations {
+	// Enable Prometheus annotations by default if DisablePrometheusAnnotations|DisableAutomaticPrometheusAnnotations is nil or true
+	if !instance.Spec.Observability.Metrics.DisablePrometheusAnnotations || !instance.Spec.Observability.Metrics.DisableAutomaticPrometheusAnnotations {
```
jaronoff97, line 31: IMO we should only use the new line and just remove references to this other than the warning. Otherwise if the first annotation is unset this would always be true, nullifying the new flag

internal/manifests/collector/annotations.go
```diff
@@ -27,8 +27,25 @@ func Annotations(instance v1beta1.OpenTelemetryCollector) (map[string]string, er
 	// new map every time, so that we don't touch the instance's annotations
 	annotations := map[string]string{}
 
-	// Enable Prometheus annotations by default if DisablePrometheusAnnotations is nil or true
-	if !instance.Spec.Observability.Metrics.DisablePrometheusAnnotations {
+	var enableAnnotations bool
+	// enableAnnotations is true only if both disable variables are false.
+	// DisableAutomaticPrometheusAnnotations takes precedence over DisablePrometheusAnnotations.
+	if instance.Spec.Observability.Metrics.DisableAutomaticPrometheusAnnotations {
```
jaronoff97, line 33: you'll need to add this to the v1beta1 type too

rivToadd: looks fixed in beta version

jaronoff97:
per my discussion with @rivToadd, this isn't needed anymore as the fix is in v1beta1 now.

PR #2688, opened 2024-02-29
ci: create release based on changes to versions.txt

swiatekm-sumo:
Automatically create a draft release if the latest tagged version and the version from versions.txt don't match. This changes the release workflow dependencies a bit.

Right now we have:
* Create release job is triggered on tag.
* When it succeeds, the operatorhub publishing job runs in the same workflow.
* All the other jobs are independently triggered on tag.

After this change:
* Create release is triggered on changes to `version.txt` in `main`.
* If needed, it creates a draft release and tag.
* Other jobs are triggered by the tag.
* Once the release manager publishes the release, this triggers the operatorhub publishing workflow.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,21 +1,45 @@
-name: "Prepare the release"
+name: "Create the release"
 on:
   push:
-    tags: [ 'v*' ]
+    branches:
+      - 'release/**'
+      - 'main'
+    paths:
+      - 'versions.txt'
 
 jobs:
+  get-versions:
+    runs-on: ubuntu-22.04
+    outputs:
+      current_version: ${{ steps.get-versions.outputs.current_version }}
+      desired_version: ${{ steps.get-versions.outputs.desired_version }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yaml
```diff
@@ -1,21 +1,45 @@
-name: "Prepare the release"
+name: "Create the release"
 on:
   push:
-    tags: [ 'v*' ]
+    branches:
+      - 'release/**'
```
jaronoff97, line 5: does this mean we should start using release branches?

jaronoff97 (approved):
one small question, otherwise this looks great to me!

.github/workflows/release.yaml
swiatekm-sumo: No, it means we can if we need to. Without this entry we wouldn't have the option.

jaronoff97: thanks for the clarification, ill wait on pavol's thoughts before merging.

pavolloffay (approved):

frzifus (approved):
Thats great stuff :)

jaronoff97:
merging!

Issue #2627, opened 2024-02-14
Propagated annotations causes issues for Deployment, StatefulSet and DaemonSet


ryanohnemus:
### Component(s)

operator

## Description

I am running into a repeat of #605, but for annotations instead. Annotation changes on the OpenTelemetryCollector Custom Resource cause the deployment (etc) to be updated with this annotation (which is changing the template) and will cause opentelemetry resources to restart.

Can the work that was completed for adding labelsFilter also be completed for annotations via an annotationsFilter so that these do not propagate to the Deployment (etc) resources that are being created?

This is especially troublesome for k8s deployments that use a form of configuration sync that applies an annotation to all resources to specify the latest version they were synced from, such as google's anthos config-sync which applies a [configmanagement.gke.io/token](https://cloud.google.com/anthos-config-management/docs/reference/labels-and-annotations) annotation each time config (related or unrelated to the OpenTelemetryCollector CR) is updated.

## Steps to Reproduce
Add any annotation to an OpenTelemetryCollector CR that use `mode: deployment` and your deployments will restart.

## Expected Result
A way to filter out which annotations are propagated to the managed resource similar to labelsFilter.

## Actual Result
No filtering available, so deployments are restarted when annotations are updated.

### Kubernetes Version

1.28.1

### Operator version

0.92.0

### Collector version

0.92.0

## Environment
OS: (e.g., "Ubuntu 20.04")
Compiler(if manually compiled): (e.g., "go 14.2")

PR #2692, opened 2024-02-29
Created annotations filter

yuriolisa:

Created ability to filter out Annotations 

- Resolves: #2627 

Unit tests have been modified accordingly. 

### DIFF
--- a/internal/config/options.go
+++ b/internal/config/options.go
@@ -50,6 +50,7 @@ type options struct {
 	operatorOpAMPBridgeImage            string
 	openshiftRoutesAvailability         openshift.RoutesAvailability
 	labelsFilter                        []string
+	annotationsFilter                   []string
 }
 
 func WithAutoDetect(a autodetect.AutoDetect) Option {
@@ -161,11 +162,11 @@ func WithOpenShiftRoutesAvailability(os openshift.RoutesAvailability) Option {
 	}
 }
 
-func WithLabelFilters(labelFilters []string) Option {
+func WithSetFilters(setFilters []string, setType string) Option {
 	return func(o *options) {
 
 		filters := []string{}
-		for _, pattern := range labelFilters {
+		for _, pattern := range setFilters {
 			var result strings.Builder
 
 			for i, literal := range strings.Split(pattern, "*") {
@@ -174,14 +175,18 @@ func WithLabelFilters(labelFilters []string) Option {
 				if i > 0 {
 					result.WriteString(".*")
 				}
-
 				// Quote any regular expression meta characters in the
 				// literal text.
 				result.WriteString(regexp.QuoteMeta(literal))
 			}
 			filters = append(filters, result.String())
 		}
+		switch filterType := setType; filterType {
+		case "labels":
+			o.labelsFilter = filters
+		case "annotations":
+			o.annotationsFilter = filters
+		}
 
-		o.labelsFilter = filters
 	}
 }
--- a/internal/manifests/manifestutils/annotations.go
+++ b/internal/manifests/manifestutils/annotations.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package collector
+package manifestutils
 
 import (
 	"crypto/sha256"
@@ -23,7 +23,7 @@ import (
 )
 
 // Annotations return the annotations for OpenTelemetryCollector pod.
-func Annotations(instance v1beta1.OpenTelemetryCollector) (map[string]string, error) {
+func Annotations(instance v1beta1.OpenTelemetryCollector, filterAnnotations []string) (map[string]string, error) {
 	// new map every time, so that we don't touch the instance's annotations
 	annotations := map[string]string{}
 
@@ -36,9 +36,11 @@ func Annotations(instance v1beta1.OpenTelemetryCollector) (map[string]string, er
 	}
 
 	// allow override of prometheus annotations
-	if nil != instance.Annotations {
-		for k, v := range instance.Annotations {
-			annotations[k] = v
+	if nil != instance.ObjectMeta.Annotations {
+		for k, v := range instance.ObjectMeta.Annotations {
+			if !isFilteredSet(k, filterAnnotations) {
+				annotations[k] = v
+			}
 		}
 	}
 
@@ -54,16 +56,18 @@ func Annotations(instance v1beta1.OpenTelemetryCollector) (map[string]string, er
 }
 
 // PodAnnotations return the spec annotations for OpenTelemetryCollector pod.
-func PodAnnotations(instance v1beta1.OpenTelemetryCollector) (map[string]string, error) {
+func PodAnnotations(instance v1beta1.OpenTelemetryCollector, filterAnnotations []string) (map[string]string, error) {
 	// new map every time, so that we don't touch the instance's annotations
 	podAnnotations := map[string]string{}
-
-	// allow override of pod annotations
-	for k, v := range instance.Spec.PodAnnotations {
-		podAnnotations[k] = v
+	if nil != instance.Spec.PodAnnotations {
+		for k, v := range instance.Spec.PodAnnotations {
+			if !isFilteredSet(k, filterAnnotations) {
+				podAnnotations[k] = v
+			}
+		}
 	}
 
-	annotations, err := Annotations(instance)
+	annotations, err := Annotations(instance, filterAnnotations)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/manifests/manifestutils/annotations_test.go
+++ b/internal/manifests/manifestutils/annotations_test.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package collector
+package manifestutils
 
 import (
 	"testing"
@@ -45,9 +45,9 @@ func TestDefaultAnnotations(t *testing.T) {
 	}
 
 	// test
-	annotations, err := Annotations(otelcol)
+	annotations, err := Annotations(otelcol, []string{})
 	require.NoError(t, err)
-	podAnnotations, err := PodAnnotations(otelcol)
+	podAnnotations, err := PodAnnotations(otelcol, []string{})
 	require.NoError(t, err)
 
 	//verify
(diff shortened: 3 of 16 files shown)


internal/config/options.go
```diff
@@ -161,11 +162,11 @@ func WithOpenShiftRoutesAvailability(os openshift.RoutesAvailability) Option {
 	}
 }
 
-func WithLabelFilters(labelFilters []string) Option {
+func WithSetFilters(setFilters []string, setType string) Option {
```
jaronoff97, line 165: I think rather than having a single method for both label and annotation, I would prefer a structure that looks like this:
```
WithLabelFilters(filters []string) Option
WithAnnotationFilters(filters []string) Option
```

both of these methods would call some shared function that does the logic for constructing the regex and filter list. That way we don't need to a potentially non-exhaustive switch and a caller is explicit about which config filter is being set.

internal/manifests/manifestutils/annotations.go
jaronoff97, line 1: very welcome package move here :)

jaronoff97: does the TA have this problem too?

internal/manifests/manifestutils/annotations_test.go
```diff
@@ -159,3 +159,26 @@ func TestAnnotationsPropagateDown(t *testing.T) {
 	assert.Equal(t, "mycomponent", podAnnotations["myapp"])
 	assert.Equal(t, "pod_annotation_value", podAnnotations["pod_annotation"])
 }
+
+func TestAnnotationsFilter(t *testing.T) {
+	otelcol := v1beta1.OpenTelemetryCollector{
+		ObjectMeta: metav1.ObjectMeta{
+			Annotations: map[string]string{"test.bar.io": "foo",
```
jaronoff97, line 166:
```suggestion
			Annotations: map[string]string{
				"test.bar.io": "foo",
```

internal/manifests/manifestutils/annotations_test.go
```diff
@@ -159,3 +159,26 @@ func TestAnnotationsPropagateDown(t *testing.T) {
...
+				"test.io/path": "/test",
+			},
+		},
+		Spec: v1beta1.OpenTelemetryCollectorSpec{
+			Mode: "deployment",
+		},
+	}
+
+	// This requires the filter to be in regex match form and not the other simpler wildcard one.
+	annotations, err := Annotations(otelcol, []string{".*.bar.io"})
```
jaronoff97, line 177:
```suggestion
	annotations, err := Annotations(otelcol, []string{".*\.bar\.io"})
```
small regex nit

jaronoff97 (commented):
a few small things, also can you augment or create an e2e test for this? do we have one for the label filter that we could tack this on to?

internal/manifests/manifestutils/annotations.go
yuriolisa: Actually, yes, because the TA doesn't have labels and annotations filters at all.

jaronoff97: lets book a ticket for that as a follow up

internal/manifests/opampbridge/annotations.go
```diff
@@ -0,0 +1,61 @@
...
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
+	"github.com/open-telemetry/opentelemetry-operator/internal/manifests/manifestutils"
+)
+
+const configMapHashAnnotationKey = "opentelemetry-opampbridge-config/hash"
+
+// Annotations returns the annotations for the OPAmpBridge Pod.
+func Annotations(instance v1alpha1.OpAMPBridge, configMap *v1.ConfigMap, filterAnnotations []string) map[string]string {
```
jaronoff97, line 30: i wonder if we should also move annotations.go and labels.go to a shared thing in manifestutils?

yuriolisa:
> a few small things, also can you augment or create an e2e test for this? do we have one for the label filter that we could tack this on to?

Added some e2e for Labels and Annotations filters. 

Do you think we should change the parameter from labels to labels-filter? I know it's a breaking change, however, this name makes more sense for that purpose.

internal/manifests/opampbridge/annotations.go
jaronoff97: yeah that makes sense, want to open an issue about that?

jaronoff97:
@yuriolisa i think if we make that change (renaming labels) we should do it in a follow up

jaronoff97:
also, looks like the e2e tests are failing on make:
`make: unrecognized option '--annotations=*filter.out'`

yuriolisa:
> @yuriolisa i think if we make that change (renaming labels) we should do it in a follow-up

So, if we intend to do a follow-up, I decided to set the annotations parameters as annotations-filter.

yuriolisa:
> also, looks like the e2e tests are failing on make: `make: unrecognized option '--annotations=*filter.out'`

Thanks for catching-up on that, it should be labels instead, once we don't have any e2e labels filter yet.

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,2 @@
 resources:
-- manager.yaml
+- manager.yaml
```
jaronoff97, line 2: small nit
```suggestion
- manager.yaml

```

jaronoff97 (approved):
one small nit, otherwise this LGTM. given the size of this, I'm going to wait to merge on another approver to give it a ✅

swiatekm-sumo (approved):
All of this looks good to me, although I think we ended up putting way too many changes into this PR. For example, the config hash annotation for opamp bridge definitely should've been a separate PR. I'm fine with merging as-is, though I wouldn't mind splitting it either.

The main thing that's actually unclear to me, is what the format of the `--annotations-filter` value is supposed to be. It looks like a glob, but then we parse it into a regex in a weird way. I'd like us to document it better, even if it's just a description of the flag.

yuriolisa:
> All of this looks good to me, although I think we ended up putting way too many changes into this PR. For example, the config hash annotation for opamp bridge definitely should've been a separate PR. I'm fine with merging as-is, though I wouldn't mind splitting it either.
> 
> The main thing that's actually unclear to me, is what the format of the `--annotations-filter` value is supposed to be. It looks like a glob, but then we parse it into a regex in a weird way. I'd like us to document it better, even if it's just a description of the flag.

@swiatekm-sumo, thank you for pointing out these topics. I will be aware next time to split into smaller parts. I've just sent an improvement to the parameter. Let me know wdyt.

main.go
```diff
@@ -145,7 +145,7 @@ func main() {
...
 	stringFlagOrEnv(&autoInstrumentationNginx, "auto-instrumentation-nginx-image", "RELATED_IMAGE_AUTO_INSTRUMENTATION_NGINX", fmt.Sprintf("ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-apache-httpd:%s", v.AutoInstrumentationNginx), "The default OpenTelemetry Nginx instrumentation image. This image is used when no image is specified in the CustomResource.")
 	pflag.StringArrayVar(&labelsFilter, "label", []string{}, "Labels to filter away from propagating onto deploys")
-	pflag.StringArrayVar(&annotationsFilter, "annotations-filter", []string{}, "Annotations to filter away from propagating onto deploys")
+	pflag.StringArrayVar(&annotationsFilter, "annotations-filter", []string{}, "Annotations to filter away from propagating onto deploys. It should be a string array that contains any match of the regular expression pattern. Example: --annotations-filter=*filter.out will filter out annotations that looks like: annotation.filter.out: true")
```
swiatekm-sumo, line 148: I think mentioning regexes here just makes things more confusing. The pattern here is not a regex, it's a literal string with `*` wildcards. 

```suggestion
	pflag.StringArrayVar(&annotationsFilter, "annotations-filter", []string{}, "Annotations to filter away from propagating onto deploys. It should be a string array containing patterns, which are literal strings optionally containing a * wildcard character. Example: --annotations-filter=*filter.out will filter out annotations that looks like: annotation.filter.out: true")
```

swiatekm-sumo (approved):

ryanohnemus:
@yuriolisa Thank you for adding this!!

yuriolisa:
> @yuriolisa Thank you for adding this!!

No problem, help to contribute; I hope it's gonna helpful for your use case.

ryanohnemus:
I just tested this out using the latest v0.96.0 image and the annotations aren't being filtered out as I'd expect. I'm not sure what i'm missing.

I setup a kind cluster via: 
```
kind create cluster --config kind-1.29.yaml
kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.14.4/cert-manager.yaml
```

then I downloaded the latest install here: 

```curl -Lo otel-op.yaml https://github.com/open-telemetry/opentelemetry-operator/releases/latest/download/opentelemetry-operator.yaml```

I then edited that to add the --annotations-filter arg below for the operator's Deployment: 
```
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/name: opentelemetry-operator
    control-plane: controller-manager
  name: opentelemetry-operator-controller-manager
  namespace: opentelemetry-operator-system
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: opentelemetry-operator
      control-plane: controller-manager
  template:
    metadata:
      labels:
        app.kubernetes.io/name: opentelemetry-operator
        control-plane: controller-manager
    spec:
      containers:
      - args:
        - --metrics-addr=127.0.0.1:8080
        - --enable-leader-election
        - --zap-log-level=info
        - --feature-gates=+operator.autoinstrumentation.go,+operator.autoinstrumentation.nginx
        - --annotations-filter="config.*.gke.io.*"
        image: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator:0.96.0
```

Finally I added a an OpenTelemetryCollector resource that has an annotation I was expecting filtered out: 
```
kind: OpenTelemetryCollector
metadata:
  name: simplest
  annotations:
    configmanagement.gke.io/token: "asdfasdf"
spec:
  config: |
    receivers:
      otlp:
        protocols:
          grpc:
          http:
    processors:
      memory_limiter:
        check_interval: 1s
        limit_percentage: 75
        spike_limit_percentage: 15
      batch:
        send_batch_size: 10000
        timeout: 10s

    exporters:
      debug:

    service:
      pipelines:
        traces:
          receivers: [otlp]
          processors: []
          exporters: [debug]
```

The `configmanagement.gke.io/token: asdfasdf` is still added to the template for the deployment of the collector and is still added to the pod.

Am I using the `--annotations-filter` option incorrectly?

Issue #2734, opened 2024-03-07
Clarify which RBAC permissions are needed for different Target Allocator features


swiatekm-sumo:
### Component(s)

target allocator

### Describe the issue you're reporting

The [target allocator README](https://github.com/open-telemetry/opentelemetry-operator/tree/main/cmd/otel-allocator#rbac) only provides RBAC permissions required to use Prometheus CRs with the target allocator. 

The allocator needs a much smaller set of permissions for static targets. It does need the specified Pod/Endpoint/... permissions if the user wants to use Kubernetes Service Discovery in their scrape config, but this isn't very clear in the documentation.

We should be clear about what is required for each of these use cases.

pavolloffay:
Related to https://github.com/open-telemetry/opentelemetry-operator/issues/424

PR #2739, opened 2024-03-08
Clarify RBAC in TargetAllocator readme. Issue ref: #2734

avillela:
**Description:** 

Updated TargetAllocator README to clarify RBAC.

- Resolves: #2734 

**Testing:**

n/a - no Operator code was modified

**Documentation:**

Updated TargetAllocator README


cmd/otel-allocator/README.md
```diff
@@ -52,7 +54,9 @@ flowchart RL
...
 
-The easiest way to do this is by going to the [Prometheus Operator’s Releases page](https://github.com/prometheus-operator/prometheus-operator/releases), grabbing a copy of the latest `bundle.yaml` file (for example, [this one](https://github.com/prometheus-operator/prometheus-operator/releases/download/v0.66.0/bundle.yaml)), and stripping out all of the YAML except the ServiceMonitor and PodMonitor YAML definitions.
+The easiest way to do this is to grab a copy of the individual [`PodMonitor`](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/charts/crds/crds/crd-podmonitors.yaml) YAML and [`ServiceMonitor`](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/charts/crds/crds/crd-servicemonitors.yaml) YAML custom resource definitions (CRDs) from the [Kube Prometheus Operator’s Helm chart](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack/charts).
```
nicolastakashi, line 57: nit: I think we could link to the prometheus operator repo itself, Since it's the source of truth to the other repos.

- https://github.com/prometheus-operator/prometheus-operator/blob/main/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml

- https://github.com/prometheus-operator/prometheus-operator/blob/main/example/prometheus-operator-crd/monitoring.coreos.com_podmonitors.yaml

avillela: @nicolastakashi according to the [README of the Prometheus Operator helm chart](https://github.com/prometheus-operator/prometheus-operator/tree/main/helm):

> The helm charts originally developed as part of this repository have been moved to [prometheus-community/kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack).
>
> The multiple charts built in this repository have been combined into a single chart that installs prometheus operator, prometheus, alertmanager, grafana as well as the multitude of exporters necessary to monitor a cluster.

Since that's the case, I would suggest keeping the URLs as-is

matej-g (approved):
Nice work, thanks for the enhancement 🙇

cmd/otel-allocator/README.md
```diff
@@ -118,14 +122,32 @@ OpenTelemetry Collector Operator-specific config.
...
+      enabled: true
+```
+
+> 🚨 **Note**: The Collector part of this same CR *also* has a serviceAccount key which only affects the collector and *not*
+the TargetAllocator.
+
+If you omit the `ServiceAccount` name, the TargetAllocator creates a `ServiceAccount` for you. The `ServiceAccount`’s default name is a concatenation of the Collector name and the `-collector` suffix. By default, this `ServiceAccount` has no defined policy, so you’ll need to create your own `ClusterRole` and `ClusterRoleBinding` for it, as per below.
+
+The role below will provide the minimum access required for the Target Allocator to query all the targets it needs based on any Prometheus configurations:
+
```
swiatekm-sumo, line 150: I think you have the roles mixed up here. The role below is specifically for Prometheus CRs.

.chloggen/avillela-2734.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
```
swiatekm-sumo, line 1: We don't normally add changelog entries for documentation updates. Do you think it's worthwhile to have this one?

pavolloffay: +

pavolloffay (approved):
LGTM,

I would recommend removing the changelog entry before merging

.chloggen/avillela-2734.yaml
avillela: The CI build process erred out when I left it out 🤷‍♀️

cmd/otel-allocator/README.md
avillela: Ooops...copy pasta error. Fixed now!

pavolloffay:
> @pavolloffay The CI process erred out when I left it out initially, which is why I put it in. Is there a way to bypass that?

It's expected, maintainers are adding skip changelog label for PRs that don't need a changelog

avillela:
> > @pavolloffay The CI process erred out when I left it out initially, which is why I put it in. Is there a way to bypass that?
> 
> It's expected, maintainers are adding skip changelog label for PRs that don't need a changelog

Ah! TIL 😄

.chloggen/avillela-2734.yaml
avillela: Removed the changelog.

cmd/otel-allocator/README.md
```diff
@@ -118,34 +122,32 @@ OpenTelemetry Collector Operator-specific config.
...
+    prometheusCR:
+      enabled: true
 ```
-In addition, the TargetAllocator needs the same permissions as a Prometheus instance would to find the matching targets
-from the CR instances.
+
+> 🚨 **Note**: The Collector part of this same CR *also* has a serviceAccount key which only affects the collector and *not*
+the TargetAllocator.
+
+If you omit the `ServiceAccount` name, the TargetAllocator creates a `ServiceAccount` for you. The `ServiceAccount`’s default name is a concatenation of the Collector name and the `-collector` suffix. By default, this `ServiceAccount` has no defined policy, so you’ll need to create your own `ClusterRole` and `ClusterRoleBinding` for it, as per below.
```
jaronoff97, line 147:
```suggestion
If you omit the `ServiceAccount` name, the TargetAllocator creates a `ServiceAccount` for you. The `ServiceAccount`’s default name is a concatenation of the Collector name and the `-targetallocator` suffix. By default, this `ServiceAccount` has no defined policy, so you’ll need to create your own `ClusterRole` and `ClusterRoleBinding` for it, as per below.
```

cmd/otel-allocator/README.md
```diff
@@ -118,34 +122,32 @@ OpenTelemetry Collector Operator-specific config.
...
+The TargetAllocator service is named based on the `OpenTelemetryCollector` CR name. For example, if your Collector CR name is `my-collector`, then the TargetAllocator `service` and `deployment` will each be named `my-collector-targetallocator`, and the `pod` will be named `my-collector-targetallocator-<pod_id>`. `collector_id` should be unique per
 collector instance, such as the pod name. The `POD_NAME` environment variable is convenient since this is supplied
 to collector instance pods by default.
 
 
 ### RBAC
-The ServiceAccount that the TargetAllocator runs as, has to have access to the CRs and the namespaces to watch for the pod and service monitors. A role like this will provide that
-access.
+
+Before the TargetAllocator can start scraping, you need to set up a Kubernetes RBAC (role-based access controls). This means that you need to have a `ServiceAccount` and corresponding cluster roles so that the TargetAllocator has access to all of the necessary resources to pull metrics from.
```
jaronoff97, line 132:
```suggestion
Before the TargetAllocator can start scraping, you need to set up Kubernetes RBAC (role-based access controls) resources. This means that you need to have a `ServiceAccount` and corresponding cluster roles so that the TargetAllocator has access to all of the necessary resources to pull metrics from.
```

jaronoff97 (commented):
two small changes, otherwise this looks great!

cmd/otel-allocator/README.md
avillela: @jaronoff97 I just re-ran my demo, and I noticed that it created a SA with the `-collector` suffix. Is that something different? Per the screen shot below, the `opentelemetry-targetallocator-sa` was created by me, and `otelcol-collector` was automagically created. I don't see any SA with the `-targetallocator` suffix. Just wondering if I'm missing something?

Screenshot 2024-03-11 at 15 33 02

PR #2747, opened 2024-03-11
[chore] require Go 1.22

swiatekm-sumo:
#2741 requires it.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/open-telemetry/opentelemetry-operator
 
-go 1.21
+go 1.22
 
 retract v1.51.0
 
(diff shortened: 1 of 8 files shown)


jaronoff97 (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/open-telemetry/opentelemetry-operator
 
-go 1.21
+go 1.22
```
frzifus, line 3: Do  we want to define a guideline like on the collector when to bump this version?
See: https://github.com/open-telemetry/opentelemetry-collector#compatibility

frzifus (commented):
I am asking to define it, since bumping it "randomly" could cause issues when users productive their own opentelemetry-operator distro.

cc @pavolloffay

Issue #1758, opened 2023-05-19
Support opentelemetry Javaagent extensions


dreaminghk:
Hi everyone! We are looking into implement an custom propagator to propagate x-request-id for ISTIO as an opentelemetry extension.

As instructed in [this page](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/examples/extension#build-and-add-extensions), we can implement that, built it as a jar and build a custom java instrumentation image that include that jar.

But the OpentelemetryOperator has to instruct the client to use it by adding -Dotel.javaagent.extensions

```
java -javaagent:path/to/opentelemetry-javaagent.jar \
     -Dotel.javaagent.extensions=build/libs/opentelemetry-java-instrumentation-extension-demo-1.0-all.jar
     -jar myapp.jar
```

I have tried something like the following in the **Instrumentation** Object:
```
  java:
    image: ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-java:latest
    env:
      - name: JAVA_TOOL_OPTIONS
        value: "$(JAVA_TOOL_OPTIONS) -Dotel.javaagent.extensions=build/libs/opentelemetry-java-instrumentation-extension-demo-1.0-all.jar"
```
However, only env names starting with OTEL_ and SPLUNK_ are supported. 

Does anyone has any idea on how to instruct the application to use a opentelemetry extension? Thank you!

jaronoff97:
If you have a custom image with your code already, all you should need to do is change the java image in the instrumentation object rather than change the environment variables. ([docs](https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1alpha1/instrumentation_types.go#L113-L116))

example:

```
  java:
    image: <your-custom-image>:<tag>
```

dreaminghk:
Thanks @jaronoff97 for your reply! In that case, looks like we will need to build the propagator into the OpenTelemetry Java Agent to produce a single jar file.

pavolloffay:
Let's reopen this, as the java extension questions come up often. We should consider supporting it and document it. 

We can still require users to provide a custom image with java agent but additionally copy the extension as well and allow configuring it.

crossoverJie:
```dockerfile
FROM busybox


ADD open-telemetry/opentelemetry-javaagent.jar /javaagent.jar

# Copy extensions to specify a path.
[...]

PR #2761, opened 2024-03-14
Instrumentations support Java extensions

crossoverJie:

```dockerfile
FROM busybox
ADD ex-1.0.0.jar /extensions/ex-1.0.0.jar
RUN chmod -R go+r /extensions/ex-1.0.0.jar
```

```yaml
kind: Instrumentation
metadata:
  name: my-instrumentation
spec:
  java:
    image: foo/a
    extensions:
      image: foo/b
      dir: /extensions
```
Operator will copy jars from the `${java.extensions}` directory to `/otel-auto-instrumentation/extensions`.

JVM parameters will also be added to the `JAVA_TOOL_OPTIONS` environment variable: `-Dotel.javaagent.extensions=/otel-auto-instrumentation-java/extensions.`

- Resolves: #1758

**Testing:** https://github.com/crossoverJie/opentelemetry-operator/blob/6727627c897d89ecdc93718f4a8417a0d69a1c48/pkg/instrumentation/javaagent_test.go#L89

**Documentation:** Later.


apis/v1alpha1/instrumentation_types.go
```diff
@@ -133,6 +133,18 @@ type Java struct {
 	// Resources describes the compute resource requirements.
 	// +optional
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// Extensions defines java specific extensions.
+	// +optional
+	Extensions *Extensions `json:"extensions,omitempty"`
```
pavolloffay, line 139: I would prefer if this would be a list.

apis/v1alpha1/instrumentation_types.go
```diff
@@ -133,6 +133,18 @@ type Java struct {
...
+}
+
+type Extensions struct {
+	// Image is a container image with extensions auto-instrumentation JAR.
+	Image string `json:"image"`
```
pavolloffay, line 144: Does the extension needs to be rebuild for every java agent version?

pavolloffay:
Please add a changelog for this

apis/v1alpha1/instrumentation_types.go
crossoverJie: I don't think it's necessary; the user just needs to package based on the javaagent version he's using.

crossoverJie: For some open source extensions, maybe it should provide a compatibility list for javaagent.

apis/v1alpha1/instrumentation_types.go
crossoverJie: You are right; this makes it easier to integrate some third-party extensions.

crossoverJie:
> Please add a changelog for this

Done, PTAL

pkg/instrumentation/javaagent.go
```diff
@@ -45,6 +46,11 @@ func injectJavaagent(javaSpec v1alpha1.Java, pod corev1.Pod, index int) (corev1.
 		}
 	}
 
+	javaJVMArgument := javaAgent
+	if len(javaSpec.Extensions) > 0 {
+		javaJVMArgument = javaAgent + fmt.Sprintf(" -Dotel.javaagent.extensions=%s/extensions", javaInstrMountPath)
```
pavolloffay, line 51: https://opentelemetry.io/docs/languages/java/automatic/configuration/#extensions

If all extensions are copied to the same directory it can cause overrides.

pavolloffay (commented):
LGTM, we just need to figure out how to avoid overriding of the extensions

pkg/instrumentation/javaagent.go
crossoverJie: Sorry, I'm not sure what the confusion is here.

I think the user should make sure that the jars in the `/extensions` directory are unique, even if he doesn't use the operator.

pavolloffay:
@crossoverJie the basic checks failed. You probably need to regenerate manifests

.chloggen/support-extensions.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: auto-instrumentation
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Instrumentations support Java extensions
```
pavolloffay, line 8: nit: Support Java auto-instrumentation extensions.

apis/v1alpha1/instrumentation_types.go
```diff
@@ -133,6 +133,18 @@ type Java struct {
 	// Resources describes the compute resource requirements.
 	// +optional
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// Extensions defines java specific extensions.
```
pavolloffay, line 137: @crossoverJie all extensions are copied to a single directory, therefore we should document it because one extension can override another one.

pavolloffay (commented):
LGTM just some documentation nits

.chloggen/support-extensions.yaml
crossoverJie: Done.

apis/v1alpha1/instrumentation_types.go
crossoverJie: I have added related documents.

pavolloffay (approved):

swiatekm-sumo (approved):

PR #2774, opened 2024-03-21
Clean up local workflow

swiatekm-sumo:
Clean up and reorganize make targets used most frequently during local development. The intent is to make the new contributor experience better and to avoid unnecessary dependencies between the targets.

The actual changes are as follows:

* `make test` no longer runs anything other than tests. Previously it also ran `fmt`, `vet` and `generate`.
* `make manager` no longer does anything other than building the manager binary. Previously it depended on some other unnecessary targets.
* `make precommit` is a new target that includes all the local checks, excluding e2e tests.
* Updated CONTRIBUTING.md to reflect the changes and removed some outdated info.


CONTRIBUTING.md
```diff
@@ -36,13 +51,6 @@ The following `make` target is run on CI to verify the project structure:
 make ensure-generate-is-noop
 ```
 
-### Pre-requisites
-* Install [Go](https://golang.org/doc/install).
-* Install [Kustomize](https://kubectl.docs.kubernetes.io/installation/kustomize/).
```
jaronoff97, line 41: is kustomize no longer needed?

CONTRIBUTING.md
```diff
@@ -8,6 +8,21 @@ We gratefully welcome improvements to documentation as well as to code.
 
 ## Getting Started
 
+### Pre-requisites
+
+* [Go](https://golang.org/doc/install).
+* Docker version 23.0.0 or greater.
```
jaronoff97, line 14: i also would recommend using kind for local development, i think that's a requirement in there?

jaronoff97: though i guess this is mentioned below...

CONTRIBUTING.md
```diff
@@ -8,6 +8,21 @@ We gratefully welcome improvements to documentation as well as to code.
...
+* `make lint` to run linters
+* `make fmt` to format Go code
+* `make vet` to run `go vet`
+* `make generate` to generate code and manifests based on Go struct definitions for CRDs.
+
+`make precommit` includes all of the above.
+
 ### Workflow
 
 It is recommended to follow the ["GitHub Workflow"](https://guides.github.com/introduction/flow/). When using [GitHub's CLI](https://github.com/cli/cli), here's how it typically looks like:
```
jaronoff97, line 28: should we include running precommit as part of the workflow?

CONTRIBUTING.md
swiatekm-sumo: It is, but we install it locally as needed, so the user doesn't need to do anything.

CONTRIBUTING.md
swiatekm-sumo: That's not strictly speaking a requirement. If you run `prepare-e2e` for example, kind will be installed automatically, so you only need docker. I think I'd prefer to add a separate section on how to run various commands in your own cluster, as it's more involved.

CONTRIBUTING.md
swiatekm-sumo: Done.

jaronoff97 (approved):

frzifus (approved):

Makefile
```diff
@@ -94,26 +94,24 @@ ensure-generate-is-noop: set-image-controller generate bundle
...
-test: generate fmt vet ensure-generate-is-noop envtest
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(KUBE_VERSION) -p path)" go test ${GOTEST_OPTS} ./...
+# No lint here, as CI runs it separately
+.PHONY: ci
+ci: generate fmt vet test ensure-generate-is-noop
 
 # Build manager binary
 .PHONY: manager
-manager: generate fmt vet
+manager:
```
pavolloffay, line 105: `generate` generates deepcopy methods for the `api` package wich seems fundamental for  manager binary

swiatekm-sumo: I don't think there's any reason to run it every time we build the manager, though. We don't do this for other binaries.

pavolloffay: The majority of contributors don't know what the `generate` command does. There is not even documentation that for the change on the objects in the  `api` package the command should be run.

swiatekm-sumo: This is actually documented here: https://github.com/open-telemetry/opentelemetry-operator/blob/main/CONTRIBUTING.md#make-changes-to-the-project-manifests.

But in general, I think this is ok. If the contributor should've run `generate`, but didn't, the CI will catch that, as will running `make precommit`.

What I'm trying to achieve with this PR is to not surprise users with changes to files in version control. Running tests or building the operator binary should not do this, in my view, even if it may be convenient at times.

pavolloffay: The makefile was scaffolded by the operator-sdk e.g. (https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v4/memcached-operator/Makefile#L145)


I am not sure how this improves dev experience if people will run manger with outdated generate methods.

pavolloffay: I am fine splitting manifest generation from the compilation. But I would expect that compilation produces up-to-date binaries.

swiatekm-sumo: I can see that argument. In that case, should I also add `generate` to the opamp bridge make target?

swiatekm-sumo: For the record, I definitely disagree with running `fmt` as part of this target, so I guess me and operator-sdk are enemies now. 🗡️

pavolloffay: I think we should be consistent and if opapm requires the generate command it should be added as well.

swiatekm-sumo: Added `generate` as a dependency of both `manager` and `opamp-bridge`.

Issue #2775, opened 2024-03-21
Allow for custom job labels in the observability spec


jaronoff97:
### Component(s)

collector, target allocator, opamp

### Is your feature request related to a problem? Please describe.

It would be great if we could specify the job label on the service monitor so that the emitted job name is customizable. Right now it ends in -collector-monitoring which isn't ideal for most backends IMO.

### Describe the solution you'd like

Add a field to the observability spec that allows specifying the jobLabel on the service per the [SM spec](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#servicemonitorspec)

swiatekm-sumo:
Are we sure we want to open the gates for more customizations of our generated CRs? In principle, we could make all the ServiceMonitor and PodMonitor attributes configurable via the MetricsConfigSpec, but in my view the user should simply create their own CRs at some point.

I'm also a bit concerned about coupling the MetricsConfigSpec to Prometheus too much. Right now it creates a Prometheus CR, but in the future we probably want to default to an otel-native way of exposing collector telemetry via the Otel Go SDK: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/observability.md#how-we-expose-telemetry.

jaronoff97:
that's a very good point... I think ill close this for now. I do think it would be great if we had the ability to set collector-self telemetry options via the metricsconfigspec, but also maybe even via a CR that can be set as a global option for all collector pools... I'll think on this.

PR #2776, opened 2024-03-21
Add joblabel customization

jaronoff97:
**Description:** Makes the job label customizable

- Resolves: #2775 

**Testing:** Unit and e2e

**Documentation:** api docs


swiatekm-sumo (commented):
The change itself looks good to me, but I'm not sure we should be making it. See my comment under the issue.

PR #2750, opened 2024-03-11
Golang instrumentation featuregates into cli

yuriolisa:
**Description:** <Describe what has changed.>

Changed featuregates.autoinstrumentation.go to cli. 
**Link to tracking Issue(s):** <Issue number if applicable>

- Resolves: #2675 

**Testing:** <Describe what testing was performed and which tests were added.>

**Documentation:** <Describe the documentation added.>

PR #2777, opened 2024-03-21
Change nginx instrumentation into command line

dexter0195:
**Description:**
Change nginx instrumentation into command line

https://github.com/open-telemetry/opentelemetry-operator/issues/2676

- Resolves: #2676

**Testing:**

**Documentation:**


dexter0195:
Hey 😃 
Can you help me understand why the e2e are all failing at the step prepare-e2e please 🙏 ?

TylerHelmuth:
The instrumentation was off by default, so you'll need to update some tests to re-enable it.  You can look at https://github.com/open-telemetry/opentelemetry-operator/pull/2750 as an example.

main.go
```diff
@@ -140,6 +141,7 @@ func main() {
 	pflag.BoolVar(&enableApacheHttpdInstrumentation, constants.FlagApacheHttpd, true, "Controls whether the operator supports Apache HTTPD auto-instrumentation")
 	pflag.BoolVar(&enableDotNetInstrumentation, constants.FlagDotNet, true, "Controls whether the operator supports dotnet auto-instrumentation")
 	pflag.BoolVar(&enablePythonInstrumentation, constants.FlagPython, true, "Controls whether the operator supports python auto-instrumentation")
+	pflag.BoolVar(&enableNginxInstrumentation, constants.FlagNginx, true, "Controls whether the operator supports nginx auto-instrumentation")
```
TylerHelmuth, line 144: This should be disabled by default.

dexter0195:
> The instrumentation was off by default, so you'll need to update some tests to re-enable it. You can look at #2750 as an example.

Hey @TylerHelmuth thank you for the advice, I had a look at the pull request you said and performed the required changes.

However the tests were still failing so I found that we were activating the feature gate in config/manager/manager.yaml , which after its removal was no longer valid. So I removed it and no the e2e tests are passing ✅ 

Please tell me if the solution I found was the right one or if there is something else I should do 🙏 
Thank you

swiatekm-sumo:
It looks like we have both the Go and the Nginx autoinstrumentations enabled in our release manifests, I believe unintentionally? If I recall correctly, the intent was to just enable them in E2E tests, right @TylerHelmuth ? In that case, we should probably just accept this state of affairs and set the flags themselves to be enabled by default. WDYT @jaronoff97 @pavolloffay ?

To be clear, I think this PR can be merged as-is, it's not changing the current behaviour.

TylerHelmuth:
> If I recall correctly, the intent was to just enable them in E2E tests

Correct.

swiatekm-sumo (approved):

TylerHelmuth (approved):

dexter0195:
Hey 😃 
thank you for the review and your comments ! Do you think we can merge this ?

.chloggen/autoinstrumentation-nginx-cli-flag.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: change nginx instrumentation feature gate into command line flag --enable-nginx-instrumentation
```
pavolloffay, line 8: Please mention here what is the feature gate flag that is being changed.

dexter0195: Done ✅

jaronoff97 (approved):

PR #2778, opened 2024-03-22
Instrumentations support select

### LINKED ISSUE #2744: Instrumentation support selector
### Component(s)

auto-instrumentation

`Instrumentation` applies to the specified Pod, similar to this:
https://skywalking.apache.org/docs/skywalking-swck/latest/java-agent-injector/#1-label-selector-and-container-matcher

```yaml
kind: Instrumentation
metadata:
  name: my-instrumentation
spec:
  selector:
    app: appname
```

_No response_

### Additional context

_No response_

crossoverJie:

```yaml
kind: Instrumentation
metadata:
  name: my-instrumentation
spec:
  selector:
    app: appname
  propagators:
    - tracecontext
  java:
    image: image
```

Support select.

**Link to tracking Issue(s):** #2744

- Resolves: #2744


apis/v1alpha1/instrumentation_types.go
```diff
@@ -22,6 +22,10 @@ import (
 
 // InstrumentationSpec defines the desired state of OpenTelemetry SDK and instrumentation.
 type InstrumentationSpec struct {
+	// Selector is the selector label of injected Object
+	// +optional
+	Selector map[string]string `json:"selector,omitempty"`
```
swiatekm-sumo, line 27: Can we use a https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#LabelSelector for this instead? This is the effective K8s standard for selecting resources based on labels.

swiatekm (commented):
I'd also like to see a E2E test for this behaviour.

crossoverJie:
> I'd also like to see a E2E test for this behaviour.

I'll add it later.

apis/v1alpha1/instrumentation_types.go
crossoverJie: Done, PTAL.

tests/e2e-instrumentation/instrumentation-select/01-install-app-select.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: apps/v1
```
swiatekm-sumo, line 1: Can you also add a second Pod here without the label, and assert that it doesn't get instrumentation injected? You can use chainsaw to simply check that the container count is 1 without checking the full manifest.

pkg/instrumentation/podmutator.go
```diff
@@ -392,18 +394,37 @@ func (pm *instPodMutator) getInstrumentationInstance(ctx context.Context, ns cor
...
+		if isMatch {
+			availableInstrument = append(availableInstrument, ins)
+		}
+	}
+
+	switch s := len(availableInstrument); {
 	case s == 0:
 		return nil, errNoInstancesAvailable
-	case s > 1:
-		return nil, errMultipleInstancesPossible
```
swiatekm-sumo, line 405: This is a significant change in behavior. Until now, if we had multiple instrumentations applicable to the same Pod, we'd just error. After this change, we'd be applying one of them without clear criteria for the choice. I can see how this makes sense with the rest of your PR, as making sure your selector are mutually exclusive sounds annoying. 

For this PR, I'd like to keep this the way it was, and we can discuss changing it in a follow-up.

tests/e2e-instrumentation/instrumentation-select/01-install-app-select.yaml
crossoverJie: Done.

pkg/instrumentation/podmutator.go
crossoverJie: Yes, this is a breaking change.

I'll create another thread to discuss this when this PR is complete.

swiatekm (approved):
LGTM, minus some unnecessary changes from E2E tests.

.chloggen/instrumentation-support-select.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: auto-instrumentation
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Instrumentations support select
```
swiatekm-sumo, line 8: It'd also be nice to make this more descriptive. How about:

```suggestion
note: Support label selectors in the Instrumentation CR
```

bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -488,6 +488,8 @@ spec:
                 - --zap-log-level=info
                 - --zap-time-encoding=rfc3339nano
                 - --feature-gates=+operator.autoinstrumentation.go,+operator.autoinstrumentation.nginx
+                - --target-allocator-image=ghcr.io/open-telemetry/opentelemetry-operator/target-allocator:v0.96.0-20-g4f69d5b
+                - --operator-opamp-bridge-image=ghcr.io/open-telemetry/opentelemetry-operator/operator-opamp-bridge:v0.96.0-20-g4f69d5b
```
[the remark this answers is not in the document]
crossoverJie: Already deleted.

iblancasa (approved):

swiatekm (commented):
I noticed that we're not following the standard semantics of v1.LabelSelector:

```
A label selector is a label query over a set of resources. The result of matchLabels and matchExpressions are ANDed. An empty label selector matches all objects. A null label selector matches no objects.
```

Whereas the way we have this implemented, `null` means "everything" and there's no way to select nothing. Could we switch to the standard semantics? I don't think there's a way to get kubebuilder to set this default for us, so we'd need to do it in the defaulting webhook, but we'd get to skip the conditional instructions in the Pod mutator. What do you think?

crossoverJie:
> Could we switch to the standard semantics?

Good suggestion. I agree with you.



> ```
> An empty label selector matches all objects. A null label selector matches no objects.
> ```

If we want to implement this semantics, the current test case will need to add empty selectors to pass the e2e test.

image

The implement code:
image

This will change the default behavior; is this necessary?

crossoverJie:
Are there any new updates now?

jaronoff97:
@swiatekm-sumo i think it may be okay for the instrumentations to select none by default, given that's technically the current behavior right?

swiatekm-sumo:
> @swiatekm-sumo i think it may be okay for the instrumentations to select none by default, given that's technically the current behavior right?

I don't think that's the case? Looking at the code being modified in this PR, it just picks the instrumentation defined in the namespace if the annotation isn't specific. Setting the default to `nil` would stop this from working. 

Incidentally, we should document that this selector only applies to this situation. As is, if the selector doesn't fit the Pod, and the Pod asks for a specific Instrumentation via the annotation, it will still get it.

crossoverJie:
> So maybe it's better to keep your current logic and make this change when we promote Instrumentation to beta?

I prefer this approach; maybe we can finish the current issue first.

apis/v1alpha1/instrumentation_types.go
```diff
@@ -22,6 +22,10 @@ import (
 
 // InstrumentationSpec defines the desired state of OpenTelemetry SDK and instrumentation.
 type InstrumentationSpec struct {
+	// Selector is the selector label of injected Object
+	// +optional
```
swiatekm-sumo, line 26:
```suggestion
	// Selector is the label selector for affected Pods.
	// Unlike standard label selectors, `nil` means `everything`, and this is also the default.
	// This may change in a future CRD version.
	// +optional
```

swiatekm (approved):
I'm ok keeping this as is, but we should document how we diverge from standard LabelSelector behavior.

pkg/instrumentation/podmutator.go
```diff
@@ -20,8 +20,11 @@ import (
 	"fmt"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/labels"
```
pavolloffay, line 23: should be merged with the block below

apis/v1alpha1/instrumentation_types.go
```diff
@@ -22,6 +22,12 @@ import (
 
 // InstrumentationSpec defines the desired state of OpenTelemetry SDK and instrumentation.
 type InstrumentationSpec struct {
+	// Selector is the label selector for affected Pods.
+	// Unlike standard label selectors, `nil` means `everything`, and this is also the default.
+	// This may change in a future CRD version.
+	// +optional
+	Selector *metav1.LabelSelector `json:"selector,omitempty"`
```
pavolloffay, line 29: Are all pods from all namespaces matched? 

What is the precedence order when annotation `instrumentation.opentelemetry.io/` is also present on the pod/namespace?

swiatekm-sumo: I believe this feature only comes into play when the annotation is not set and we try to pick an Instrumentation from the namespace the Pod is in. The annotation has precedence. But we should document this behaviour somewhere.

pkg/instrumentation/podmutator.go
swiatekm-sumo: I wonder why golangci-lint doesn't complain about this. We do have the `goimports` linter enabled.

crossoverJie: 😂 I have used `goimports` and `lint` and have not found this issue, and now I have manually adjusted it.

apis/v1alpha1/instrumentation_types.go
crossoverJie: > I believe this feature only comes into play when the annotation is not set and we try to pick an Instrumentation from the namespace the Pod is in

https://github.com/open-telemetry/opentelemetry-operator/blob/c65629b3b9fcb722e770eb58b73ea4f5cc8bc63f/pkg/instrumentation/podmutator.go#L368-L377

Did I understand it wrong? I think the premise for this feature to take effect is that the `instrumentation.opentelemetry.io/inject-xx` is equal to true.

swiatekm-sumo: What I meant to say was that this mechanism only applies if the annotation doesn't specify which instrumentation should be used. Apologies for the confusion.

crossoverJie: Thanks; I understand that the documentation has been supplemented.

crossoverJie:
The E2e test failed, but I can pass it locally, any suggestions?

swiatekm-sumo:
> @swiatekm-sumo The E2E test failed, but I can pass it locally, any suggestions?

Can you try recreating the test cluster locally to see if tests still pass? It's possible you're using an older docker image in tests.

crossoverJie:
> Can you try recreating the test cluster locally to see if tests still pass? It's possible you're using an older docker image in tests.

Thanks; I have changed the e2e test.

swiatekm-sumo:
@crossoverJie I think you're running into trouble merging changes in generated manifests. I suggest running `make precommit` and including the output in the merge commit.

crossoverJie:
> @crossoverJie I think you're running into trouble merging changes in generated manifests. I suggest running `make precommit` and including the output in the merge commit.

Thanks, done.

swiatekm (approved):

swiatekm-sumo:
@crossoverJie can you rebase on main? That should get rid of the EasyCLA complaint.

crossoverJie:
> @crossoverJie can you rebase on main? That should get rid of the EasyCLA complaint.

Thanks for your reminder, done.

swiatekm-sumo:
I don't get why EasyCLA complains about my email address. Feel free to just drop the `Co-authored-by: Mikołaj Świątek <mail+sumo@mikolajswiatek.com>`  messages from the commits it complains about. I don't want this to block this PR.

tests/e2e-instrumentation/instrumentation-select/01-assert-select-multi.yaml
```diff
@@ -0,0 +1,77 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  annotations:
+    instrumentation.opentelemetry.io/inject-java: "true"
```
pavolloffay, line 5: I don't understand the use-case to have the `Selector` in the CR while the inject annotation is still needed. 

What is the use-case this feature solves? Why do we need to maintain two approaches to solve the same use-case?

pavolloffay (changes requested):

tests/e2e-instrumentation/instrumentation-select/01-assert-select-multi.yaml
crossoverJie: This test case is to test that when there are multiple `Instrumentation` (they all use `Selector`), the `Selector` feature can still work as expected.

jackshirazi:
Nice PR. Is this blocked on some particular change needed?

jaronoff97:
@jackshirazi we're discussing the design of this in the linked issue #2744

PR #2787, opened 2024-03-25
Check for the permissions instead of using a CLI flag

### LINKED ISSUE #2588: Replace the create-rbac-permissions flag by just checking if the SA has permissions
### Component(s)

_No response_

After fixing #2525, the `ClusterRole` and `ClusterRoleBinding` are not part of the operator manifest anymore.

During a SIG call, we agreed to document that those permissions need to be added to the operator service account if you want the operator to, automatically, generate the RBAC resources for those processors that need it. Also, instead of having the `create-rbac-permissions`  flag, we can check if the SA has permissions to handle those resources and reconcile them.

_No response_

### Additional context

_No response_

### LINKED ISSUE #2833: kubernetes detector doesn't exist for resourcedetection
### Component(s)

collector

### What happened?

While creating some E2E tests for https://github.com/open-telemetry/opentelemetry-operator/pull/2787, I found one of the detectors `resourcedetection` for `resourcedetection` is `kubernetes`. It doesn't exist and it is `k8snode`.

### Kubernetes Version

.

### Operator version

.

### Collector version

.

## Environment
OS: (e.g., "Ubuntu 20.04")

### Log output

_No response_

### Additional context

_No response_

### LINKED ISSUE #2862: Collectors with same name create ClusterRoleBinding with same name
### Component(s)

_No response_

## Description
When two OTEL Collectors are created with the same name and the operator can create `ClusterRoleBindings`, the first one will be overridden by the second one. The permissions will mismatch and the namespace.

## Steps to Reproduce
Create two Collectors with the same name in different namespaces with the ability to create RBAC resources. The second one will override the first one.

### Kubernetes Version

.

### Operator version

.

### Collector version

.

## Environment
OS: (e.g., "Ubuntu 20.04")

### Log output

_No response_

### Additional context

_No response_

iblancasa:
Closes #2588
Closes #2833 
Closes #2862


internal/autodetect/operator.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	namespace, err := GetOperatorNamespace()
+	if err != nil {
+		return "", err
+	}
+
+	deploymentName := "opentelemetry-operator"
```
swiatekm-sumo, line 51: There's no guarantee that this is the operator's Deployment name. The user can set this to whatever they want. I think we need a more robust detection mechanism here.

main.go
```diff
@@ -135,7 +136,7 @@ func main() {
 	pflag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
-	pflag.BoolVar(&createRBACPermissions, "create-rbac-permissions", false, "Automatically create RBAC permissions needed by the processors")
+	pflag.BoolVar(&createRBACPermissions, "create-rbac-permissions", false, "Automatically create RBAC permissions needed by the processors (deprecated)")
```
swiatekm-sumo, line 139: I thought we wanted to keep this flag, just set the default to true eventually, and only create the RBAC resources if we have permission *and* the flag is set. This way, users who don't want the operator to do this can easily disable it.

internal/autodetect/operator.go
iblancasa: Any idea about how I can implement this?

main.go
iblancasa: My understanding from the SIG call (as I reflected in #2588) was that we wanted to just check if the SA has the correct permissions. If users want to disable this feature, they will just remove the permissions from the SA (there is no other reason to keep those permissions otherwise, right?)

swiatekm-sumo: I don't actually recall that discussion :sweat_smile: , but that sounds fair enough as long as we output a log about what the operator is doing. We should also make this clear in the documentation, so users don't need to hunt around the code to understand the operator's behaviour.

internal/autodetect/operator.go
swiatekm-sumo: I think the most idiomatic way is to use the downward API to inject this into the operator deployment as an environment variable:

```yaml
      env:
        - name: SERVICE_ACCOUNT_NAME
          valueFrom:
            fieldRef:
              fieldPath: spec.serviceAccountName
```

If we want to do this without this kind of helper, then I think you need to read something from the environment and then look for the right Pod based on that using the client. Not sure what the way with the least assumptions is.

apis/v1alpha1/collector_webhook.go
```diff
@@ -378,7 +376,7 @@ func (c CollectorWebhook) validateTargetAllocatorConfig(ctx context.Context, r *
 		if subjectAccessReviews, err := c.reviewer.CheckPolicyRules(ctx, r.GetNamespace(), r.Spec.TargetAllocator.ServiceAccount, targetAllocatorCRPolicyRules...); err != nil {
 			return nil, fmt.Errorf("unable to check rbac rules %w", err)
 		} else if allowed, deniedReviews := rbac.AllSubjectAccessReviewsAllowed(subjectAccessReviews); !allowed {
-			return warningsGroupedByResource(deniedReviews), nil
+			return rbac.WarningsGroupedByResource(deniedReviews), nil
```
pavolloffay, line 379: I might be missing here something so take this with a grain of salt.

I thought that an admission warning should be printed if a component is enabled that requires custom role and the operator cannot create it.

iblancasa: The idea was to check if the service account of the operator has the needed permissions to create RBAC resources when the application starts. If it has those permissions, we will try to create them when needed (every time we reconcile an OTEL Collector, for instance). Else, we will skip.

So, if you change the permissions to the service account to enable/disable this feature, you will need to restart the OTEL Operator deployment.

internal/autodetect/operator.go
```diff
@@ -0,0 +1,37 @@
...
+		return "", err
+	}
+	return string(nsBytes), nil
+}
+
+func GetOperatorServiceAccount() (string, error) {
+	saEnvVar := "SERVICE_ACCOUNT_NAME"
+	sa := os.Getenv(saEnvVar)
+	if sa == "" {
+		return sa, fmt.Errorf("%s env variable not found", saEnvVar)
```
swiatekm-sumo, line 34: So if this doesn't exist, we just log the error and continue as if we didn't have the permissions? That sounds reasonable to me.

swiatekm-sumo (commented):
LGTM, but I'd like to see a e2e or envtest test for this.

internal/autodetect/operator.go
iblancasa: Yes

iblancasa:
> LGTM, but I'd like to see a e2e or envtest test for this.

Sure. I'm implementing them.

internal/autodetect/operator.go
```diff
@@ -0,0 +1,37 @@
...
+func GetOperatorNamespace() (string, error) {
+	nsBytes, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
+	if err != nil {
+		return "", err
+	}
+	return string(nsBytes), nil
+}
+
+func GetOperatorServiceAccount() (string, error) {
+	saEnvVar := "SERVICE_ACCOUNT_NAME"
```
jaronoff97, line 31: can you make this a const in the file here?

main.go
```diff
@@ -410,3 +427,32 @@ func tlsConfigSetting(cfg *tls.Config, tlsOpt tlsConfig) {
 	}
 	cfg.CipherSuites = cipherSuiteIDs
 }
+
+func checkRbacPermissions(reviewer *rbac.Reviewer, ctx context.Context) (admission.Warnings, error) {
```
jaronoff97, line 431: Instead of keeping this method here, could we move it to the autodetect package and follow the pattern we have for environment detection? That would allow us to keep with our current style for these checks:
```
if cfg.RBACPermissions() == permissions.Allowed {
	setupLog.Info("Has necessary RBAC permissions.")
	...
} else {
	setupLog.Info("Does not have necessary RBAC permissions, skipping.")
}
```

internal/rbac/format.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	v1 "k8s.io/api/authorization/v1"
+)
+
+// WarningsGroupedByResource is a helper to take the missing permissions and format them as warnings.
+func WarningsGroupedByResource(reviews []*v1.SubjectAccessReview) []string {
```
jaronoff97, line 25: can we add a unit test for this?

jaronoff97 (changes requested):

internal/autodetect/main.go
```diff
@@ -29,14 +34,17 @@ var _ AutoDetect = (*autoDetect)(nil)
...
 type autoDetect struct {
-	dcl discovery.DiscoveryInterface
+	ctx      context.Context
+	dcl      discovery.DiscoveryInterface
+	reviewer *rbac.Reviewer
 }
 
 // New creates a new auto-detection worker, using the given client when talking to the current cluster.
-func New(restConfig *rest.Config) (AutoDetect, error) {
+func New(restConfig *rest.Config, ctx context.Context, reviewer *rbac.Reviewer) (AutoDetect, error) {
```
jaronoff97, line 47: suggestion: it's never good to keep context in a struct, we should instead pass context when its needed

internal/autodetect/main.go
```diff
@@ -83,3 +93,15 @@ func (a *autoDetect) OpenShiftRoutesAvailability() (openshift.RoutesAvailability
 
 	return openshift.RoutesNotAvailable, nil
 }
+
+func (a *autoDetect) RBACPermissions() (autoRbac.Availability, error) {
+	w, err := autoRbac.CheckRbacPermissions(a.reviewer, a.ctx)
```
jaronoff97, line 98: this should take in a context

internal/autodetect/main_test.go
```diff
@@ -104,7 +105,7 @@ func TestDetectPlatformBasedOnAvailableAPIGroupsPrometheus(t *testing.T) {
 		}))
 		defer server.Close()
 
-		autoDetect, err := autodetect.New(&rest.Config{Host: server.URL})
+		autoDetect, err := autodetect.New(&rest.Config{Host: server.URL}, nil, nil)
```
jaronoff97, line 108: can you add a test for rbac here too?

main.go
```diff
@@ -284,17 +237,78 @@ func main() {
...
+		setupLog.Info("The operator has not permissions to create RBAC resources, skipping.")
+	} else {
+		createRBACPermissions = true
+	}
+
+	cfg := config.New(
+		config.WithLogger(ctrl.Log.WithName("config")),
+		config.WithVersion(v),
+		config.WithCollectorImage(collectorImage),
+		config.WithCreateRBACPermissions(createRBACPermissions),
```
jaronoff97, line 270: we don't need this option anymore and should instead use the autodetect's RBACPermissions() method

main.go
```diff
@@ -284,17 +237,78 @@ func main() {
...
 	if err != nil {
-		setupLog.Error(clientErr, "failed to create kubernetes clientset")
+		createRBACPermissions = false
+		setupLog.Info("error checking if the operator has permissions to create RBAC resources", "reason", err)
+	} else if rbacAvailable == autoRbac.NotAvailable {
+		createRBACPermissions = false
+		setupLog.Info("The operator has not permissions to create RBAC resources, skipping.")
+	} else {
+		createRBACPermissions = true
+	}
```
jaronoff97, line 264: same here, we shouldn't need this anymore

Makefile
```diff
@@ -207,6 +207,19 @@ generate: controller-gen
...
+e2e-automatic-rbac: chainsaw
+	# Add extra needed permissions
+	kubectl apply -f hack/rbac-operator-permissions.yaml
+	kubectl rollout restart deployment opentelemetry-operator-controller-manager -n opentelemetry-operator-system
+	kubectl rollout status deployment opentelemetry-operator-controller-manager -n opentelemetry-operator-system
+	go run hack/check-operator-ready.go
+	$(CHAINSAW) test --test-dir ./tests/e2e-automatic-rbac
+	# Cleanup
+	kubectl delete -f hack/rbac-operator-permissions.yaml
+	kubectl rollout restart deployment opentelemetry-operator-controller-manager -n opentelemetry-operator-system
```
swiatekm-sumo, line 221: I think I'd prefer if this was done the same way feature gates are enabled in e2e tests - by adding a patch to the kustomization. I think you can just add in the permissions you need that way, without needing to redeclare the whole ClusterRole, which is brittle.

As a general rule, when we do custom setup in e2e tests, we should do it the same way as we'd imagine users applying the same customization in their clusters.

swiatekm-sumo (approved):
Nitpick about the test setup, LGTM otherwise.

Makefile
iblancasa: Done

hack/rbac-operator-permissions.yaml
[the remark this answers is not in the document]
iblancasa: Actually, this is not needed anymore because we apply it with `kustomize`

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
+  newTag: 0.97.1-7-gcafbc81a
```
swiatekm-sumo, line 8: `make reset` to get rid of this.

pavolloffay:
The e2e tests failed

.chloggen/replace-create-rbac-permissions-by-checking-the-sa-permissions.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Automatically enable RBAC creation if operator SA can create clusterroles and bindings. We should also add link to the manifest that defines the RBAC. --create-rbac-permissions flag is noop and deprecated now.
```
pavolloffay, line 8: `We should also add link to the manifest that defines the RBAC` should not be there. I was asking to add a link to the RBAC users can create for the operator SA

iblancasa: Sorry.I don't know how I commited this.

iblancasa: The problem with the link is that we will duplicate the information because we are applying it to `kustomize` too. And we need to apply the patches as I did in this PR because otherwise the PolicyRules list will be overridden.

swiatekm-sumo (approved):

config/manager/kustomization.yaml
jaronoff97: @iblancasa bump on this

config/rbac/extra-permissions-operator/namespaces.yaml
```diff
@@ -0,0 +1,15 @@
+- op: add
+  path: /rules/-
```
jaronoff97, line 2: where does this come from?

internal/autodetect/main.go
```diff
@@ -83,3 +91,15 @@ func (a *autoDetect) OpenShiftRoutesAvailability() (openshift.RoutesAvailability
 
 	return openshift.RoutesNotAvailable, nil
 }
+
+func (a *autoDetect) RBACPermissions(ctx context.Context) (autoRbac.Availability, error) {
+	w, err := autoRbac.CheckRbacPermissions(ctx, a.reviewer)
```
jaronoff97, line 96: nit: can we make the naming convention consistent? I.e. RBAC vs Rbac (I think the all caps probs makes more sense)

tests/e2e-automatic-rbac/processor-k8sattributes/02-assert.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: simplest-chainsaw-k8sattributes-cluster-role
```
jaronoff97, line 4: can we also confirm:
* the binding
* the service account name

jaronoff97 (commented):
a few more minor things! :bow:

config/rbac/extra-permissions-operator/namespaces.yaml
iblancasa: Do you mean "why do the operator needs permission for namespaces"? If yes, because the `k8sattributes` needs it. And a SA cannot provide permissions over a resource it has not permissions.

jaronoff97: i mean was this manually generated?

jaronoff97: Im just not familiar with how this will be used

iblancasa: > i mean was this manually generated?

Yes.

I was asked to apply the permissions for the operator in a similar way that users will apply them. https://github.com/open-telemetry/opentelemetry-operator/pull/2787/files/a12aade2316541c7f0be674704c53980378b2321#r1560704098

config/manager/kustomization.yaml
jaronoff97: @iblancasa this is still here, we need to remove it prior to merge

config/rbac/extra-permissions-operator/nodes.yaml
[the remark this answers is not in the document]
swiatekm-sumo: I think these can go into the `/tests/e2e-automatic-rbac` directory. My request to use patches instead of overriding the whole default ClusterRole aimed to keep e2e test setup consistent. It's good if our e2e tests use the kustomize manifests roughly the way we expect users to, but these files don't literally need to be user-facing in any way.

Does that make sense?

jaronoff97:
Just a few more Qs, once those are answered I think we're ✅

pavolloffay:
The PR needs to be rebased

tests/e2e-automatic-rbac/processor-resourcedetection/chainsaw-test.yaml
```diff
@@ -0,0 +1,24 @@
...
+    - apply:
+        file: 01-install.yaml
+    - assert:
+        file: 01-assert.yaml
+  - name: k8snode-detector
+    try:
+    - apply:
+        file: 02-install.yaml
+    - assert:
+        file: 02-assert.yaml
```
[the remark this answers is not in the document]
iblancasa: I have been thinking about that since I created the test. Is it ok to use the `contrib` image for that? I can do it in a different PR yes.

config/rbac/extra-permissions-operator/nodes.yaml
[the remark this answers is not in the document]
iblancasa: I put the files under `config/rbac` because this error is thrown by `kustomize`:
```sh
Error: accumulating resources: accumulation err='accumulating resources from '../rbac': '/home/iblancasa/projects/opentelemetry-operator/config/rbac' must resolve to a file': recursed accumulation of path '/home/iblancasa/projects/opentelemetry-operator/config/rbac': trouble configuring builtin PatchTransformer with config: `
path: ../../tests/e2e-automatic-rbac/extra-permissions-operator/namespaces.yaml
target:
  kind: ClusterRole
  name: manager-role
`: security; file '/home/iblancasa/projects/opentelemetry-operator/tests/e2e-automatic-rbac/extra-permissions-operator/namespaces.yaml' is not in or below '/home/iblancasa/projects/opentelemetry-operator/config/rbac'
error: no objects passed to apply
```

tests/e2e-automatic-rbac/processor-resourcedetection/chainsaw-test.yaml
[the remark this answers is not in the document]
jaronoff97: does the k8s distro have all the things we need?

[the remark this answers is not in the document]
iblancasa: Oh, actually it should. Didn't remember you added/are adding that. Thanks!

iblancasa:
Are any other changes required?

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
```
pavolloffay, line 3: this needs to be removed

.chloggen/2833-fix-detector-resourcedetectionprocessor.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: collector
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Use the k8snode detector instead of kubernetes for the automatic RBAC creation for the resourcedetector"
```
pavolloffay, line 8: Why the kubernetes does not work? 

is this a breaking change?

pavolloffay (approved):

.chloggen/2833-fix-detector-resourcedetectionprocessor.yaml
pavolloffay: whas `kubernetes` deprecated or removed from the processor? It would be good to note that

iblancasa: No. It was a mistake made by the operator. `kubernetes` has never exist` It was `k8snode`.

swiatekm-sumo (approved):
Minor nitpick about the e2e tests, but it shouldn't block merging this PR. Thank you for being so patient with this review process!

jaronoff97 (approved):

pavolloffay (approved):

Issue #2793, opened 2024-03-27
TargetAllocator returned scrape config crashes otel MetricsReceiver when relabel or metric relabel action has action:keepequal or dropequal 


rashmichandrashekar:
### Component(s)

target allocator

## Description
TargetAllocator's responce crashes metricsreceiver when relabel or metric relabel action has action:keepequal or dropequal.
When the metricsreceiver calls the TA service to receive config and unmarshal the response, it fails because of the unmarshal implementation. https://github.com/prometheus/prometheus/blob/25a8d576717f4a69290d6f6755b4a90cfaab08ff/model/relabel/relabel.go#L144C3-L144C46
This is related to this issue - https://github.com/prometheus/prometheus/issues/12534

## Steps to Reproduce
Use configmap or pod or service monitor CRD that use action keepequal or dropequal.

## Expected Result
The action should not result in otel metrics receiver crashing

## Actual Result
Prometheus Metrics Receiver crashes after receiving config from TA with the following error - 
"level":"error","ts":1707850745.82048,"caller":"prometheusreceiver/metrics_receiver.go:167","msg":"Failed to retrieve job list","kind":"receiver","name":"prometheus","data_typ
e":"metrics","error":"keepequal action requires only 'source_labels' and `target_label`, and no other fields"

### Kubernetes Version

1.26.12

### Operator version

0.96.0

### Collector version

0.96.0

swiatekm-sumo:
I'd love to solve this in a robust way, where we can be confident we're generating valid config, not just for these specific relabel actions. I think we might be able to imitate what prometheus-operator does, maybe Prometheus itself also has a config serialization mechanism defined that we could reuse.

PR #2794, opened 2024-03-27
Fix for unmarshalling error for keepequal/dropequal

rashmichandrashekar:
**Description:** 
 Fixing a bug - Fix for targetAllocator returned scrape config crashing otel MetricsReceiver when relabel or metric relabel action has action:keepequal or dropequal

[2793
](https://github.com/open-telemetry/opentelemetry-operator/issues/2793)
- Resolves: #2793

Tested that it doesn't crash metricsreceiver


cmd/otel-allocator/server/server.go
```diff
@@ -120,8 +120,44 @@ func (s *Server) UpdateScrapeConfigResponse(configs map[string]*promconfig.Scrap
 	if err != nil {
 		return err
 	}
+
+	var jobToScrapeConfig map[string]interface{}
+	err = json.Unmarshal(jsonConfig, &jobToScrapeConfig)
+	if jobToScrapeConfig != nil {
```
swiatekm-sumo, line 126: Instead of doing this scary looking surgery, we should probably do something similar to what prometheus-operator does, and manually build the scrapeconfig out of appropriate yaml structs: https://github.com/prometheus-operator/prometheus-operator/blob/4e3b2bcea44cbfdbd5c4975d777240285721dc6b/pkg/prometheus/promcfg.go#L1645/

rashmichandrashekar: Thanks @swiatekm-sumo. But that would mean that whenever there is an update to the scrape config, we would need to  manually update this as well which seems tedious. Also, I dont see any libraries I can reuse. Until it is fixed in prometheus, could this be considered as a fix since the marshal implementation simplifies the conversion and we can rely on the interface implementation to reliably convert to yaml? Once that is done, we can just remove this.

rashmichandrashekar: @swiatekm-sumo - Following up. Please let me know your thoughts. Thanks!

swiatekm-sumo: I understand that your approach fixes this specific problem, but is still leaves us uncertain if we don't have other similar problems that we have yet to encounter. I'm not against merging this specific fix as-is, but I'd like to look for a more systemic solution, where we can be assured our serialized scrape configs are correct. @jaronoff97 @pavolloffay wdyt?

rashmichandrashekar: Thanks @swiatekm-sumo. As a part of this fix, I also did test out all the relabel actions available with prometheus configuration. These 2 were the only actions that cause problems and since this is unique to a bug, should not likely reoccur. If there will be other issues related to marshalling, it would also affect all prometheus users dependent on it and the ideal fix should be from the prometheus implementation of marshalling. We should also be able to reliably take dependency on the native prometheus marshalling/unmarshalling implementation without duplication of effort. Please let me know what you think. Thanks again!

rashmichandrashekar: Sure @swiatekm-sumo. I have added tests checking that unmarshaling doesn't result in errors for any prometheus action supported. Please take a look. Thank you!

swiatekm-sumo (commented):
I think I'm ok with this fix, although I'd like to open an issue to find a better solution as well. But I'd like @pavolloffay and @jaronoff97 to give their opinions as well.

cmd/otel-allocator/server/server_test.go
```diff
@@ -612,6 +613,50 @@ func TestServer_Readiness(t *testing.T) {
 	}
 }
 
+func TestServer_ScrapeConfigRespose(t *testing.T) {
+	tests := []struct {
```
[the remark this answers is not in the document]
rashmichandrashekar: Thanks @swiatekm-sumo. Added a few more cases. Please let me know if this looks good.

cmd/otel-allocator/server/server.go
```diff
@@ -120,8 +121,45 @@ func (s *Server) UpdateScrapeConfigResponse(configs map[string]*promconfig.Scrap
...
+		if scrapeConfig["metric_relabel_configs"] != nil {
+			metricRelabelConfigs := scrapeConfig["metric_relabel_configs"].([]interface{})
+			for _, metricRelabelConfig := range metricRelabelConfigs {
+				metricRelabelConfig := metricRelabelConfig.(map[string]interface{})
+				// Dropping regex key from the map since unmarshalling this on the client(metrics_receiver.go) results in error
+				// because of the bug here - https://github.com/prometheus/prometheus/issues/12534
+				if metricRelabelConfig["action"] == "keepequal" || metricRelabelConfig["action"] == "dropequal" {
+					delete(metricRelabelConfig, "regex")
+				}
+			}
```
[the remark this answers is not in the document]
rashmichandrashekar: Thanks @swiatekm-sumo. Moved this to a helper function.

swiatekm-sumo (commented):
Please fix the linter error, otherwise this looks good to me.

swiatekm-sumo (approved):
Thanks for fixing this, and being patient with my feedback. As I said earlier, I consider this a temporary fix until we can figure out a more robust way of dealing with the problem. However, fixing the bug is more important than other considerations here in my view.

jaronoff97:
Could you add a comment to the function signature explaining why we need to do this, linking this issue https://github.com/prometheus/prometheus/issues/12534

rashmichandrashekar:
> Thanks for fixing this, and being patient with my feedback. As I said earlier, I consider this a temporary fix until we can figure out a more robust way of dealing with the problem. However, fixing the bug is more important than other considerations here in my view.

Thanks @swiatekm-sumo. Your feedback makes sense to me.

rashmichandrashekar:
> Could you add a comment to the function signature explaining why we need to do this, linking this issue [prometheus/prometheus#12534](https://github.com/prometheus/prometheus/issues/12534)

Thanks @jaronoff97. Added the comment. Please take a look.

cmd/otel-allocator/server/server.go
```diff
@@ -107,7 +107,12 @@ func (s *Server) Shutdown(ctx context.Context) error {
 	return s.server.Shutdown(ctx)
 }
 
-func RemoveRelabelAction(jsonConfig []byte) ([]byte, error) {
+// This is needed specifically for keepequal/dropequal actions because even though the user doesn't specify the
```
jaronoff97, line 110:
```suggestion
// RemoveRegexFromRelabelAction is needed specifically for keepequal/dropequal actions because even though the user doesn't specify the
```

rashmichandrashekar:
@jaronoff97 - I have committed your suggestion. Please merge if all looks good, Thanks again!

jaronoff97:
@rashmichandrashekar before I merge, did you confirm this worked by running the TA locally and the scrape_configs response worked with the collector?

rashmichandrashekar:
> @rashmichandrashekar before I merge, did you confirm this worked by running the TA locally and the scrape_configs response worked with the collector? If not, it may be helpful to add an e2e here...

@jaronoff97 - Yes, I did test out that this works by running TA and collector locally :)

jaronoff97 (approved):

Issue #2796, opened 2024-03-28
Remove Target Allocator Rewrite Feature Gate in v0.99.0


swiatekm-sumo:
### Component(s)

collector, target allocator

### Describe the issue you're reporting

The feature gate has been in beta for a while. We should promote it to Stable, and remove it in a subsequent version.

jaronoff97:
closed by #2797

swiatekm-sumo:
I'd like to use this issue to track removing the flag in 0.99.0.

jaronoff97:
@swiatekm-sumo sounds good, would you mind rewriting the title to match that?

PR #2797, opened 2024-03-28
Promote operator.collector.rewritetargetallocator feature flag to stable

swiatekm-sumo:
**Description:**
Promote the flag to stable, with the intent of removing it before the next release.

**Link to tracking Issue(s):** https://github.com/open-telemetry/opentelemetry-operator/issues/2796

**Testing:**
Removed tests for the case where the flag is disabled.

### DIFF
--- a/pkg/featuregate/featuregate.go
+++ b/pkg/featuregate/featuregate.go
@@ -53,9 +53,10 @@ var (
 	// automatically be rewritten when the target allocator is enabled.
 	EnableTargetAllocatorRewrite = featuregate.GlobalRegistry().MustRegister(
 		"operator.collector.rewritetargetallocator",
-		featuregate.StageBeta,
+		featuregate.StageStable,
 		featuregate.WithRegisterDescription("controls whether the operator should configure the collector's targetAllocator configuration"),
 		featuregate.WithRegisterFromVersion("v0.76.1"),
+		featuregate.WithRegisterToVersion("v0.98.0"),
 	)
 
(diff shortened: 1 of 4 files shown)


pkg/featuregate/featuregate.go
```diff
@@ -53,9 +53,10 @@ var (
 	// automatically be rewritten when the target allocator is enabled.
 	EnableTargetAllocatorRewrite = featuregate.GlobalRegistry().MustRegister(
 		"operator.collector.rewritetargetallocator",
-		featuregate.StageBeta,
+		featuregate.StageStable,
 		featuregate.WithRegisterDescription("controls whether the operator should configure the collector's targetAllocator configuration"),
 		featuregate.WithRegisterFromVersion("v0.76.1"),
+		featuregate.WithRegisterToVersion("v0.97.0"),
```
jaronoff97, line 59: I think we usually do 2 versions from the current one

jaronoff97 (approved):

cmergenthaler:
@swiatekm-sumo Would you mind explaining why this PR is a breaking change for the user? Does it already influence people using the feature-gate or will it only when removing the feature gate?

swiatekm-sumo:
> @swiatekm-sumo Would you mind explaining why this PR is a breaking change for the user? Does it already influence people using the feature-gate or will it only when removing the feature gate in an upcoming release?

Before this change, the feature flag was in Beta, and it was possible to disable it. After the change to Stable, this is no longer possible, and attempting to do so will cause an error. So this is a breaking change for users who had this flag explicitly disabled.

Issue #2376, opened 2023-11-20
Use k8s 1.28 native sidecar for OTEL collector sidecar mode


pavolloffay:
### Component(s)

collector

### Is your feature request related to a problem? Please describe.

The k8s 1.28 introduced "native sidecar" mode (init-container with `restartPolicy: Always`) https://kubernetes.io/blog/2023/08/25/native-sidecar-containers/. 

The approach makes sure the sidecar starts/finishes before the application. 

### Describe the solution you'd like

Use the native sidecar for collector sidecar on k8s 1.28 and above.

frzifus:
I would like to give that a try.

angelokurtis:
Hey @frzifus, are you currently working on this task? If not, I'd be interested in contributing to it.

im-aIex:
Any update on getting this support added? Or any known workarounds to get the new `restartPolicy: Always` included in the meantime?

jasonious:
We are running into this while using gcsfuse annotation which creates a native sidecar, and then the pods can't be terminated. Any estimated timelines for when/if the solution will be merged in? It appears to not have any movement on it for a few months.

frzifus:
Hey, I will try to get this ready on the weekend. :)

PR #2801, opened 2024-03-29
add featuregate for k8s 1.28 native sidecar container

frzifus:

- Resolves: #2376 

### DIFF
--- a/.chloggen/native_sidecar.yaml
+++ b/.chloggen/native_sidecar.yaml
@@ -2,7 +2,7 @@
 change_type: enhancement
 
 # The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
-component: pkg/sidecar
+component: collector
 
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
 note: Add native sidecar injection behind a feature gate which is disabled by default.
@@ -13,4 +13,9 @@ issues: [2376]
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
-subtext:
+subtext: |
+  Native sidecars are supported since Kubernetes version `1.28` and are availabe by default since `1.29`.
+  To use native sidecars on Kubernetes v1.28 make sure the "SidecarContainers" feature gate on kubernetes is enabled.
+  If native sidecars are available, the operator can be advised to use them by adding adding
+  the `--feature-gates=operator.sidecarcontainers.native` to the Operator args.
+  In the future this may will become availabe as deployment mode on the Collector CR. See [#3356](https://github.com/open-telemetry/opentelemetry-operator/issues/3356)
--- a/.chloggen/native_sidecar.yaml
+++ b/.chloggen/native_sidecar.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: pkg/sidecar
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add native sidecar injection behind a feature gate which is disabled by default.
+
+# One or more tracking issues related to the change
+issues: [2376]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -46,6 +46,9 @@ jobs:
            setup: "add-operator-arg OPERATOR_ARG='--feature-gates=operator.targetallocator.mtls' add-certmanager-permissions prepare-e2e"
          - group: e2e-automatic-rbac
            setup: "add-rbac-permissions-to-operator prepare-e2e"
+         - group: e2e-native-sidecar
+           setup: "add-operator-arg OPERATOR_ARG='--feature-gates=operator.sidecarcontainers.native' prepare-e2e"
+           kube-version: "1.31"
     steps:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v4
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -48,7 +48,7 @@ jobs:
            setup: "add-rbac-permissions-to-operator prepare-e2e"
          - group: e2e-native-sidecar
            setup: "add-operator-arg OPERATOR_ARG='--feature-gates=operator.sidecarcontainers.native' prepare-e2e"
-           kube-version: "1.31"
+           kube-version: "1.29"
     steps:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v4
--- a/pkg/featuregate/featuregate.go
+++ b/pkg/featuregate/featuregate.go
@@ -35,7 +35,7 @@ var (
 	EnableNativeSidecarContainers = featuregate.GlobalRegistry().MustRegister(
 		"operator.sidecarcontainers.native",
 		featuregate.StageAlpha,
-		featuregate.WithRegisterDescription("controls whether the operator supports sidecar containers as init containers"),
+		featuregate.WithRegisterDescription("controls whether the operator supports sidecar containers as init containers. Should only be enabled on k8s v1.29+"),
 	)
 	// PrometheusOperatorIsAvailable is the feature gate that enables features associated to the Prometheus Operator.
 	PrometheusOperatorIsAvailable = featuregate.GlobalRegistry().MustRegister(
--- a/pkg/featuregate/featuregate.go
+++ b/pkg/featuregate/featuregate.go
@@ -25,6 +25,18 @@ const (
 )
 
 var (
+	// EnableNativeSidecarContainers is the feature gate that controls whether a
+	// sidecar should be injected as a native sidecar or the classic way.
+	// Native sidecar containers have been available since kubernetes v1.28 in
+	// alpha and v1.29 in beta.
+	// It needs to be enabled with +featureGate=SidecarContainers.
+	// See:
+	// https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-gates-for-alpha-or-beta-features
+	EnableNativeSidecarContainers = featuregate.GlobalRegistry().MustRegister(
+		"operator.sidecarcontainers.native",
+		featuregate.StageAlpha,
+		featuregate.WithRegisterDescription("controls whether the operator supports sidecar containers as init containers"),
+	)
 	// PrometheusOperatorIsAvailable is the feature gate that enables features associated to the Prometheus Operator.
 	PrometheusOperatorIsAvailable = featuregate.GlobalRegistry().MustRegister(
 		"operator.observability.prometheus",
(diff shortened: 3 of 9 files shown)


pkg/sidecar/pod.go
```diff
@@ -47,7 +48,15 @@ func add(cfg config.Config, logger logr.Logger, otelcol v1beta1.OpenTelemetryCol
 		container.Env = append(container.Env, attributes...)
 	}
 	pod.Spec.InitContainers = append(pod.Spec.InitContainers, otelcol.Spec.InitContainers...)
-	pod.Spec.Containers = append(pod.Spec.Containers, container)
+
+	if featuregate.EnableNativeSidecarContainers.IsEnabled() {
+		policy := corev1.ContainerRestartPolicyAlways
+		container.RestartPolicy = &policy
+		// TODO(frzifus): Add StartupProbe
```
frzifus, line 55: Should we pre-define a startup probe that checks the `${service.telemetry.metrics.address}/metrics` endpoint, expose it in the CRD or do something else?

wdyt @open-telemetry/operator-approvers ?

swiatekm-sumo: Is there a reason to treat it any differently than we do readiness probes and liveness probes? We can even default to the readiness probe if not set.

frzifus: Nope, but reusing the readiness probe sounds good to me :+1:

jaronoff97: I don't think there's a good reason... is there much of a difference between these things? Maybe in v1beta1 we should just use the upstream definition?

jaronoff97 (commented):
I think this makes sense... is there a way to write an e2e for this?

swiatekm-sumo:
> I think this makes sense... is there a way to write an e2e for this?

Sure, you'd need to add a new test suite and run it with the gate enabled. Then checking if the sidecar is an initContainer should be straightforward.

pkg/sidecar/pod.go
frzifus: Using the upstream definition would extend ours by the [ProbeHandler](https://github.com/kubernetes/api/blob/5147c1a32f6a0b9b155bb84e59f933e0ff8a3792/core/v1/types.go#L2836).

frzifus:
@swiatekm whats the best way to configure a k8s featuregate for a specific e2e test suite?

Do we need to create a new e.g. `kind-1.30-native-sidecar.yaml` and register it here?
https://github.com/open-telemetry/opentelemetry-operator/blob/a515234b34557a325f0f944ac97a7f22a9b8f9fb/.github/workflows/e2e.yaml#L22-L24

Do we need then to create a new step like `e2e-test-sidecar` here too?
https://github.com/open-telemetry/opentelemetry-operator/blob/a515234b34557a325f0f944ac97a7f22a9b8f9fb/.github/workflows/e2e.yaml#L14

swiatekm:
> @swiatekm whats the best way to configure a k8s featuregate for a specific e2e test suite?
> 
> Do we need to create a new e.g. `kind-1.30-native-sidecar.yaml` and register it here?
> 
> https://github.com/open-telemetry/opentelemetry-operator/blob/a515234b34557a325f0f944ac97a7f22a9b8f9fb/.github/workflows/e2e.yaml#L22-L24
> 
> Do we need then to create a new step like `e2e-test-sidecar` here too?
> 
> https://github.com/open-telemetry/opentelemetry-operator/blob/a515234b34557a325f0f944ac97a7f22a9b8f9fb/.github/workflows/e2e.yaml#L14

You should do the same thing we do for enabling other options in e2e tests. See [here](https://github.com/open-telemetry/opentelemetry-operator/blob/154719ba1669c4ff6d50c6d9368c83cd18400ca4/.github/workflows/e2e.yaml#L37).

You do need to add a new entry to the e2e test matrix, because you need a separate operator deployment with the flag enabled. Then you add a make target that adds the command-line option to the operator Deployment (same as we do for other operator flags), and use that make target for setup in the e2e tests.

pkg/featuregate/featuregate.go
```diff
@@ -25,6 +25,19 @@ const (
...
 
 var (
+	// EnableNativeSidecarContainers is the feature gate that controls whether a
+	// sidecar should be injected as a native sidecar or the classic way.
+	// Native sidecar containers have been available since kubernetes v1.28 in
+	// alpha and v1.29 in beta.
+	// It needs to be enabled with +featureGate=SidecarContainers.
+	// See:
+	// https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-gates-for-alpha-or-beta-features
+	EnableNativeSidecarContainers = featuregate.GlobalRegistry().MustRegister(
```
pavolloffay, line 35: I wonder if we could on k8s 1.28+ always use the native sidecar pattern and keep using the existing approach on older k8s versions.

This way no feature flag is needed and we would always use a recommended approach.

frzifus: We can do that, but we would expect different results in our e2e tests. Which makes it hart to execute the same tests on both platform versions.

swiatekm: I think I'm ok with that as well. I can't think of any issues it might cause.

frzifus: ok, then I will enable it for the 1.28+ e2e tests. Would be nice it it makes it into [v0.111.0](https://github.com/open-telemetry/opentelemetry-operator/issues/3323).

frzifus:
Seems we dont need to enable it in Kind. 
https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/

> Provided that your cluster has the SidecarContainers [feature gate](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) enabled (the feature is active by default since Kubernetes v1.29), you can specify a restartPolicy for containers listed in a Pod's initContainers field.

pkg/sidecar/pod.go
```diff
@@ -71,6 +82,16 @@ func remove(pod corev1.Pod) corev1.Pod {
...
+
+	// NOTE: we also remove init containers (native sidecars) since k8s 1.28.
+	// This should have no side effects.
+	var initContainers []corev1.Container
+	for _, initContainer := range pod.Spec.InitContainers {
+		if initContainer.Name != naming.Container() {
+			initContainers = append(initContainers, initContainer)
+		}
+	}
+	pod.Spec.InitContainers = initContainers
```
swiatekm, line 94: How about using https://pkg.go.dev/slices#DeleteFunc? Also, shouldn't this only happen if the feature flag is set?

pkg/sidecar/pod.go
```diff
@@ -81,5 +102,12 @@ func existsIn(pod corev1.Pod) bool {
 			return true
 		}
 	}
+	// NOTE: we also check init containers (native sidecars) since k8s 1.28.
+	// This should have no side effects.
+	for _, container := range pod.Spec.InitContainers {
+		if container.Name == naming.Container() {
+			return true
+		}
+	}
```
swiatekm, line 111: How about https://pkg.go.dev/slices#ContainsFunc? Both here and above. Also, shouldn't this only happen if the feature flag is set?

pkg/sidecar/pod.go
```diff
@@ -64,20 +76,28 @@ func remove(pod corev1.Pod) corev1.Pod {
...
 	}
-	pod.Spec.Containers = containers
 	return pod
 }
 
 // existsIn checks whether a sidecar container exists in the given pod.
 func existsIn(pod corev1.Pod) bool {
-	for _, container := range pod.Spec.Containers {
-		if container.Name == naming.Container() {
+	fn := func(c corev1.Container) bool { return c.Name == naming.Container() }
```
jaronoff97, line 92: this is a nit, but given we use this a few times here, it may be easier to just make it a file scoped var.

pkg/sidecar/pod_test.go
```diff
@@ -19,17 +19,107 @@ import (
...
 
 	"github.com/open-telemetry/opentelemetry-operator/apis/v1beta1"
 	"github.com/open-telemetry/opentelemetry-operator/internal/config"
 	"github.com/open-telemetry/opentelemetry-operator/internal/naming"
+	"github.com/open-telemetry/opentelemetry-operator/pkg/featuregate"
 )
 
 var logger = logf.Log.WithName("unit-tests")
 
+func sidecarFeatureGate(t *testing.T) {
```
jaronoff97, line 35:
```suggestion
func enableSidecarFeatureGate(t *testing.T) {
```

tests/e2e-native-sidecar/00-assert.yaml
```diff
@@ -0,0 +1,21 @@
...
+kind: Pod
+metadata:
+  annotations:
+    sidecar.opentelemetry.io/inject: "true"
+  name: myapp
+spec:
+  containers:
+  - name: myapp
+  initContainers:
+  - name: otc-container
```
jaronoff97, line 12: should we verify that this has the correct restart policy?

tests/e2e-native-sidecar/00-install.yaml
```diff
@@ -0,0 +1,41 @@
...
+apiVersion: v1
+kind: Pod
+metadata:
+  name: myapp
+  annotations:
+    sidecar.opentelemetry.io/inject: "true"
+spec:
+  containers:
+  - name: myapp
+    image: jaegertracing/vertx-create-span:operator-e2e-tests
```
jaronoff97, line 38: should we use one of our own images here?

swiatekm (approved):

.chloggen/native_sidecar.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: pkg/sidecar
```
pavolloffay, line 5: collector

.chloggen/native_sidecar.yaml
```diff
@@ -0,0 +1,16 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add native sidecar injection behind a feature gate which is disabled by default.
+
+# One or more tracking issues related to the change
+issues: [2376]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
pavolloffay, line 16: Could you please document here the feature gate as a flag e.g. `--`. Could you please document which k8s version support by default native sidecar?

frzifus:
Updated the changelog @pavolloffay

.github/workflows/e2e.yaml
```diff
@@ -46,6 +46,9 @@ jobs:
            setup: "add-operator-arg OPERATOR_ARG='--feature-gates=operator.targetallocator.mtls' add-certmanager-permissions prepare-e2e"
          - group: e2e-automatic-rbac
            setup: "add-rbac-permissions-to-operator prepare-e2e"
+         - group: e2e-native-sidecar
+           setup: "add-operator-arg OPERATOR_ARG='--feature-gates=operator.sidecarcontainers.native' prepare-e2e"
+           kube-version: "1.31"
```
pavolloffay, line 51: could we test this in 1.29?

pkg/featuregate/featuregate.go
```diff
@@ -25,6 +25,18 @@ const (
...
+		"operator.sidecarcontainers.native",
+		featuregate.StageAlpha,
+		featuregate.WithRegisterDescription("controls whether the operator supports sidecar containers as init containers"),
```
pavolloffay, line 38: could you please mention here it should be used on k8s 1.29?

pavolloffay (commented):
LGTM,

just some docs nits

frzifus:
Changed it to k8s 1.29 and we now mention in the fg desc that it should only be enabled on k8s v1.29+.

pavolloffay (approved):

Issue #2802, opened 2024-03-31
Node autoinstrumentation image includes dev dependencies


Zirak:
### Component(s)

auto-instrumentation

### Describe the issue you're reporting

When building the node autoinstrumentation image, its `devDepencencies` are included:

```
# 60d37e18ce is HEAD at time of writing
$ docker run -it --rm --name otel-autoinstro-node -w /autoinstrumentation autoinstrumentation-nodejs:60d37e18ce /bin/sh

/autoinstrumentation # du -ahd2 | sort -h
[ ... ]
796.0K  ./node_modules/copyfiles
808.0K  ./node_modules/@hapi
840.0K  ./node_modules/resolve
1.1M    ./node_modules/path-scurry
1.9M    ./node_modules/rimraf
2.9M    ./node_modules/protobufjs
3.6M    ./node_modules/@types
3.7M    ./node_modules/@grpc
31.3M   ./node_modules/typescript
112.5M  ./node_modules/@opentelemetry
169.3M  .
169.3M  ./node_modules
```

Here's the result of building the image without typescript, typescript + rimraf, and finally all dev dependencies:

```
$ docker image ls --format 'table {{.Repository}} {{.Tag}} {{.Size}}'
autoinstrumentation-nodejs 60d37e18ce 272MB
autoinstrumentation-nodejs no-ts 208MB
autoinstrumentation-nodejs no-ts-rmrf 201MB
autoinstrumentation-nodejs no-dev 200MB
```

I'll open up a PR to remove these dependencies as part of the build. It seems like every PR must have a tracking issue alongside it, hence this issue; apologies if I misunderstood the process.

PR #2803, opened 2024-03-31
Remove nodejs autoinstrumentation dev dependencies from image

Zirak:
**Description:**

The autoinstrumentation has several build-time dependencies which aren't necessary for the final image, like typescript. Comparison of builds without typescript, typescript + rimraf, and finally all dev dependencies:

```
$ docker image ls --format 'table {{.Repository}} {{.Tag}} {{.Size}}'
autoinstrumentation-nodejs 60d37e18ce 272MB
autoinstrumentation-nodejs no-ts 208MB
autoinstrumentation-nodejs no-ts-rmrf 201MB
autoinstrumentation-nodejs no-dev 200MB
```

We extend the existing `postcompile` script. After building into `build/workspace`, we `prune` to remove extraneous packages. Doing that requires the `package.json` in the build directory.

- Resolves: #2802 

**Testing:** Locally instrumenting a nodejs application.

### DIFF
--- a/.chloggen/js-build-remove-dev.yaml
+++ b/.chloggen/js-build-remove-dev.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: auto-instrumentation
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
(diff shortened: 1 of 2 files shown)


swiatekm-sumo (approved):
LGTM, but someone with a better understanding of the node instrumentation should double-check before we merge.

I do wonder if we won't break anyone with this change. Technically, nobody should be using these dependencies, but you never know.

Zirak:
Thanks! 

> Technically, nobody should be using these dependencies, but you never know.

Interesting! From my understanding the only possible user for these dependencies is the nodejs agent itself. Since the modules end up in `/autoinstrumentation/node_modules`, an end user shouldn't really be exposed to them.

pavolloffay (approved):

Issue #2810, opened 2024-04-04
Otel operator crashes if podmonitor and servicemonitor are not installed


jaronoff97:
### Component(s)

collector

## Description
When you install the operator on a fresh cluster it will crash if you do not have the podmonitor and servicemonitors installed

## Steps to Reproduce
* create a kind cluster
* install cert manager
* install operator

## Expected Result
:) happy operator

## Actual Result
:( sad operator:
```
{"level":"error","ts":"2024-04-04T18:40:39Z","logger":"controller-runtime.source.EventHandler","msg":"if kind is a CRD, it should be installed before calling Start","kind":"ServiceMonitor.monitoring.coreos.com","error":"no matches for kind \"ServiceMonitor\" in version \"monitoring.coreos.com/v1\"","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/source.(*Kind).Start.func1.1\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.2/pkg/internal/source/kind.go:63\nk8s.io/apimachinery/pkg/util/wait.loopConditionUntilContext.func2\n\t/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.29.3/pkg/util/wait/loop.go:87\nk8s.io/apimachinery/pkg/util/wait.loopConditionUntilContext\n\t/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.29.3/pkg/util/wait/loop.go:88\nk8s.io/apimachinery/pkg/util/wait.PollUntilContextCancel\n\t/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.29.3/pkg/util/wait/poll.go:33\nsigs.k8s.io/controller-runtime/pkg/internal/source.(*Kind).Start.func1\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.2/pkg/internal/source/kind.go:56"}
[...]

PR #2811, opened 2024-04-04
Only add to scheme when prometheus is available

jaronoff97:
**Description:** Changes the prometheus operator integration to fail by logging (rather than panicking) and as a result only allowing the creation of prometheus resources when it is safe to do so.

- Resolves: #2810 

**Testing:** unit tests, e2e testing, local confirmation

**Documentation:** n/a

### DIFF
--- a/.chloggen/disable-looking-for-podmonitor-servicemonitor-when-not-needed.yaml
+++ b/.chloggen/disable-looking-for-podmonitor-servicemonitor-when-not-needed.yaml
@@ -5,7 +5,7 @@ change_type: bug_fix
 component: collector
 
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: resolves a problem for trying to use a scheme that isn't present.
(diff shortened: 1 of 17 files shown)


.chloggen/disable-looking-for-podmonitor-servicemonitor-when-not-needed.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: collector
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: resolves a problem for trying to use a scheme that isn't present.
```
pavolloffay, line 8: This could be more descriptive and mention p8s CRDs

TylerHelmuth (approved):

Issue #2770, opened 2024-03-20
Annotations Filter Added in v0.96.0 does not filter annotations


ryanohnemus:
### Component(s)

collector

### What happened?

I am hoping this is possibly just user error, but I attempted to use the annotations-filter added in the v0.96.0 release. 
I tested this out using the latest v0.96.0 image and the annotations aren't being filtered out as I'd expect.

I setup a kind cluster using the kind-1.29.yaml included in the repo:

```
# setup kind
kind create cluster --config kind-1.29.yaml

# add required cert-manager install
kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.14.4/cert-manager.yaml

# Download the latest operator install
curl -Lo otel-op.yaml https://github.com/open-telemetry/opentelemetry-operator/releases/latest/download/opentelemetry-operator.yaml
```

I then edited the `otel-op.yaml` to add the `--annotations-filter="config.*.gke.io.*"` arg below for the operator's Deployment:

```
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/name: opentelemetry-operator
    control-plane: controller-manager
  name: opentelemetry-operator-controller-manager
  namespace: opentelemetry-operator-system
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: opentelemetry-operator
      control-plane: controller-manager
  template:
    metadata:
      labels:
        app.kubernetes.io/name: opentelemetry-operator
        control-plane: controller-manager
    spec:
      containers:
      - args:
        - --metrics-addr=127.0.0.1:8080
        - --enable-leader-election
        - --zap-log-level=info
        - --zap-time-encoding=rfc3339nano
        - --feature-gates=+operator.autoinstrumentation.go,+operator.autoinstrumentation.nginx
        - --annotations-filter="config.*.gke.io.*"
        image: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator:0.96.0
```

```
 # Applied operator to the cluster and confirmed the operator pod args contained the annotations-filter above: 
 kubectl apply -f otel-op.yaml
 ```

Finally I added an OpenTelemetryCollector resource that has an annotation `configmanagement.gke.io/token` I was expecting filtered out:
```
apiVersion: opentelemetry.io/v1alpha1
kind: OpenTelemetryCollector
metadata:
  name: simplest
  annotations:
    configmanagement.gke.io/token: "asdfasdf"
spec:
  config: |
[...]

PR #2828, opened 2024-04-08
Fixed labels and annotations filter

yuriolisa:

Fixing the Labels and Annotations filter.

- Resolves: #2770 


internal/manifests/manifestutils/labels.go
```diff
@@ -26,8 +25,9 @@ import (
 
 func IsFilteredSet(sourceSet string, filterSet []string) bool {
 	for _, pattern := range filterSet {
-		match, _ := regexp.MatchString(pattern, sourceSet)
-		return match
+		if strings.Contains(sourceSet, pattern) {
+			return true
+		}
```
swiatekm-sumo, line 30: Rather than fixing it, it looks more like you're changing the behaviour to not support wildcards at all? That's more a breaking change than a fix.

yuriolisa: I know it's a break change, but if we think about efficiency and simplicity to end-user, I would say it's better. WDYT?

jaronoff97: is there a reason this functionality doesn't work as intended today? The user seems to be reporting an issue with just a single filter matching, not multiple so I'm not sure this is going to fix it.

yuriolisa: I could keep it, but it will keep confusing the users to set the proper regex model to match labels or annotations. Regarding the multiple filters, it was reported [here](https://github.com/open-telemetry/opentelemetry-operator/issues/2770#issuecomment-2031921173).

swiatekm-sumo: The problem is that the syntax is poorly defined right now. I think the intent was for it to be glob-like, so can we use https://github.com/bmatcuk/doublestar and be done with it? I believe this would keep current semantics and fix the bug.

jaronoff97: yes, i see the multi-filter issue, but the user is also reporting that single filter isn't working either

yuriolisa: Right, @jaronoff97, it's because it is confusing how to enable it properly.

yuriolisa: > https://github.com/bmatcuk/doublestar

I will have a look if we can use it to keep the current semantics.

yuriolisa: Despite the doublestar being functional, it’s not meeting our requirements for the filters. They check like /*/, but some annotations don’t have any slashes, which the package doesn’t filter properly.
So, I decided to leave the same way, using regexp but using a pre-compilation of the pattern and fixing the multiple annotations.

internal/manifests/manifestutils/labels.go
```diff
@@ -18,16 +18,26 @@ import (
...
+
 func IsFilteredSet(sourceSet string, filterSet []string) bool {
-	for _, pattern := range filterSet {
-		match, _ := regexp.MatchString(pattern, sourceSet)
-		return match
+	for _, basePattern := range filterSet {
+		pattern, compileErr := regexp.Compile(basePattern)
+		match := pattern.MatchString(sourceSet)
+		if compileErr != nil {
+			logger.Error(compileErr, "could not compile the regexp pattern")
```
jaronoff97, line 36: is there a logger could pass in as a parameter instead here?

jaronoff97 (commented):
one question, otherwise this is looking pretty good. Do we need to update any e2e tests?

internal/manifests/manifestutils/labels.go
yuriolisa: I don't know if it's a good idea, once we call this function for each annotation/label we put in our resources.

yuriolisa:
> one question, otherwise this is looking pretty good. Do we need to update any e2e tests?

No, I did change the github action passing the proper parameter when we setup the operator's test.

internal/manifests/manifestutils/labels.go
yuriolisa: Sounds good. I've changed the function signature to support the logger we propagate.

jaronoff97 (approved):

internal/manifests/manifestutils/labels.go
jaronoff97: thank you for doing all that parameter drilling :)

.chloggen/fix-filter-labels-annotations.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'bug_fix'
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: operator
```
pavolloffay, line 5: there is no component operator.

internal/manifests/manifestutils/labels.go
```diff
@@ -18,28 +18,36 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/go-logr/logr"
```
pavolloffay, line 21: the API imports in this file are not ordered. the `metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"` should be here

internal/status/collector/collector.go
```diff
@@ -15,22 +15,24 @@
...
+	"context"
+	"strconv"
 
 	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
 	"github.com/open-telemetry/opentelemetry-operator/internal/manifests/collector"
 	"github.com/open-telemetry/opentelemetry-operator/internal/manifests/manifestutils"
 	"github.com/open-telemetry/opentelemetry-operator/internal/naming"
 	"github.com/open-telemetry/opentelemetry-operator/internal/version"
+
+	"github.com/go-logr/logr"
```
pavolloffay, line 29: these should be above the internal imports

pavolloffay:
The PR needs to be rebased as well

.chloggen/fix-filter-labels-annotations.yaml
pavolloffay: This change is probably for the collector

PR #2831, opened 2024-04-09
feat: change java feature gate to cli flag

dexter0195:
**Description:** 
Change java instrumentation into command line

https://github.com/open-telemetry/opentelemetry-operator/issues/2673

- Resolves: #2673

**Testing:**

**Documentation:**


jaronoff97:
@dexter0195 looks like some unit test failures for the podmutator_test.go

swiatekm-sumo (commented):
Looks like you have some unit tests failing because the new option is not enabled by default in the config constructor. Simplest way to fix these failures would be to enable it by default in `config.New`.

dexter0195:
Hey @swiatekm-sumo thank you for the tip ! 😃 
I set the java intrumentation flag to default true in the config as you said and it worked ! I also needed to add a call to the config contructor in the java tests, see commit https://github.com/open-telemetry/opentelemetry-operator/pull/2831/commits/776ffb674bb8d9c724d1a61cc1a07c259066add8 

Please let me know what you think 🙏

swiatekm-sumo (approved):
LGTM, thanks for doing this!

internal/config/main.go
```diff
@@ -74,6 +75,7 @@ func New(opts ...Option) Config {
 		operatorOpAMPBridgeConfigMapEntry: defaultOperatorOpAMPBridgeConfigMapEntry,
 		logger:                            logf.Log.WithName("config"),
 		version:                           version.Get(),
+		enableJavaInstrumentation:         true,
```
jaronoff97, line 78: is there a reason we need to set this here if our default for the option is true?

iblancasa (approved):

Issue #2835, opened 2024-04-10
Use the new k8s distro


jaronoff97:
### Component(s)

collector

### Is your feature request related to a problem? Please describe.

We now have a kubernetes distro! 🎉 let's use that

### Describe the solution you'd like

change all our core uses to the k8s distro

pavolloffay:
At the moment the operator uses the core as a default distro https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/distributions/otelcol/manifest.yaml 

The proposal is to use https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/distributions/otelcol-k8s/manifest.yaml 

This will be a breaking change for users. The k8s distribution does not have
receivers: 
* kafka
* noop

exporters:
* noop
* logging (deprecated)
* opencensus
* zipkin
*  kafka
* prometheus
* prometheusremotewrite

extensions:
* ballast

processors:
* span

swiatekm-sumo:
So we do need to migrate after all. Thanks for being vigilant about this. In that case, do we go back to the original plan we hatched at KubeCon? That is, set the image to core while converting from v1alpha1 to v1beta1?

swiatekm-sumo:
Looking at that list, we actually use the `prometheus` exporter in some tests, which is a bit annoying. We'd need to keep using either core or contrib there.

jaronoff97:
I do think it may be worth updating the k8s distro to include the prom-remote-write exporter and prom exporter as both of those are things we know operator installations use

pavolloffay:
I would agree about the p8s exporter.

swiatekm-sumo:
The [k8s distribution rules](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-k8s#opentelemetry-collector-kubernetes-distro) state: `Only exporters that use OTLP are allowed. `. That rule can probably be relaxed, but judging from the process of establishing the k8s distribution, I wouldn't count on that happening in less than 3-4 weeks. So if we want to make this change alongside the v1beta1 transition, then we either delay it yet again, or we go with the k8s distribution as-is and pursue adding the additional exporters independently.

Could we also do a cleverer conversion method, where we check which components the Collector is using, and keep core as the default image only if those components aren't present in the k8s distribution?

@TylerHelmuth what do you think? It sounded like you gave some thought to doing this transition in the Helm Chart.

swiatekm-sumo:
[...]

PR #2836, opened 2024-04-10
Change to use kube distro

jaronoff97:
**Description:** Uses the new kubernetes distro

- Resolves: #2835 

**Testing:** local

**Documentation:** maybe we should add a note...


TylerHelmuth:
We'll need the image released before this PR will work

tests/e2e/smoke-pod-labels/00-install.yaml
```diff
@@ -3,7 +3,7 @@ kind: OpenTelemetryCollector
 metadata:
   name: testlabel
 spec:
-  image: ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-contrib:0.94.0
```
swiatekm-sumo, line 6: Do we want to change this? I thought the point was to use a non-default here? It also won't work because the first k8s distro release will be 0.98.0.

swiatekm-sumo (commented):
This change is independent of v1beta1 CRDs, but we want to roll it out simultaneously, right?

.chloggen/kube-distro.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
pavolloffay, line 2: This should be a breaking change. We discussed that the k8s distro should be enabled for the first time with v1beta1 release

jaronoff97: not a breaking change :) we're going from core to k8s distro

pavolloffay (approved):

TylerHelmuth (approved):

jaronoff97:
@pavolloffay you'll need to merge this into your release PR, all the tests will probably fail because the image only exists for 0.97.0. Actually this will be an interesting exercise for what actually needs a running collector (@swiatekm-sumo)

PR #2849, opened 2024-04-12
Upgrade golangci-lint to 1.57.2 and enable gci linter

swiatekm-sumo:
`gci` can actually enforce import grouping and order. I've upgraded golangci-lint, enabled gci, and fixed all the issues it flagged. Let me know if we're ok with this ordering. I also cleaned up the golangci-lint config file of deprecated and ineffective settings.


jaronoff97 (approved):
sooths my brain, thank you

PR #2735, opened 2024-03-07
Remove unnecessary config marshalling wherever possible

jaronoff97:
**Description:** <Describe what has changed.>

**Link to tracking Issue(s):** <Issue number if applicable>

- Resolves: #issue-number

**Testing:** <Describe what testing was performed and which tests were added.>

**Documentation:** <Describe the documentation added.>

PR #2851, opened 2024-04-13
Remove `operator.collector.rewritetargetallocator` feature flag

swiatekm-sumo:
**Description:**
This feature gate is slated for removal in 0.99.0. 

After merging this, we should rethink how the prometheus config validation should work. It's probably still too strict in some cases the way it is now.

**Link to tracking Issue(s):** #2796 

- Resolves: #2796

**Testing:**
Removed some tests for the non-enabled case.

### DIFF
--- a/internal/manifests/targetallocator/adapters/config_to_prom_config.go
+++ b/internal/manifests/targetallocator/adapters/config_to_prom_config.go
@@ -15,7 +15,6 @@
 package adapters
 
 import (
-	"errors"
 	"fmt"
 	"net/url"
 	"regexp"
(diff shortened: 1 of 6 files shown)


internal/manifests/targetallocator/adapters/config_to_prom_config.go
```diff
@@ -297,20 +296,12 @@ func AddTAConfigToPromConfig(prometheus map[interface{}]interface{}, taServiceNa
 }
 
 // ValidatePromConfig checks if the prometheus receiver config is valid given other collector-level settings.
-func ValidatePromConfig(config map[interface{}]interface{}, targetAllocatorEnabled bool, targetAllocatorRewriteEnabled bool) error {
+func ValidatePromConfig(config map[interface{}]interface{}, targetAllocatorEnabled bool) error {
+	// TODO: Rethink this validation, now that target allocator rewrite is enabled permanently.
```
jaronoff97, line 300: we can probably build off of my PR [here](https://github.com/open-telemetry/opentelemetry-operator/pull/2735) and create a known type for this and just insert it via a struct or something

jaronoff97 (approved):

Issue #2603, opened 2024-02-03
Remove unnecessary config marshalling in v1alpha2


pavolloffay:
### Component(s)

collector

### Describe the issue you're reporting

Created from https://github.com/open-telemetry/opentelemetry-operator/pull/2532

crossoverJie:
@pavolloffay 
I'd like to fix this issue.

We can create a hardcoded method: 

```go
func (c *Config) ToMap() map[string]interface{} {
	result := make(map[string]interface{})

	result["receivers"] = c.Receivers.Object
	result["exporters"] = c.Exporters.Object
	if c.Processors != nil {
		result["processors"] = c.Processors.Object
	}

	if c.Connectors != nil {
		result["connectors"] = c.Connectors.Object
	}
	if c.Extensions != nil {
		result["extensions"] = c.Extensions.Object

	}
	serviceMap := make(map[string]interface{})
	if c.Service.Extensions != nil {
		serviceMap["extensions"] = c.Service.Extensions
	}
	if c.Service.Telemetry != nil {
		serviceMap["telemetry"] = c.Service.Telemetry
	}
	serviceMap["pipelines"] = c.Service.Pipelines.Object
	result["service"] = serviceMap

	return result
}
```

Or use serialization method:

```go
func (c *Config) ToMap() (map[string]interface{}, error) {
	var result map[string]interface{}
	v, err := json.Marshal(c)
	if err != nil {
		return nil, err
	}
	err = json.Unmarshal(v, &result)
	if err != nil {
		return nil, err
	}
	return result, err
}
```

But this will still cause multiple calls to the serialization and deserialization method.

BWT, maybe we can replace `map[interface{}]interface{}` with `map[string]interface{}` so we can avoid redundant type assertions.

https://github.com/open-telemetry/opentelemetry-operator/blob/e58df33539f7b0af378b1998e6e4e107d86f00d3/internal/manifests/collector/adapters/config_validate.go#L68

jaronoff97:
@crossoverJie yep, already working on this :)

pavolloffay:
@jaronoff97 is this done?

jaronoff97:
Almost! We still have [this code block](https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/manifests/collector/config_replace.go#L45-L88) that we should probably try and fix

Issue #2399, opened 2023-11-28
Block bad resources from reconciliation


jaronoff97:
### Component(s)

collector

### Is your feature request related to a problem? Please describe.

Following up from #1972 , we should check for manifest errors in the webhook(s) and prevent objects from being created when we know they will not work.

### Describe the solution you'd like

Modify the existing validating webhooks to run the manifest generators checking for any errors that may exist.

jaronoff97:
@taniyourstruly is going to take this one on after her great work on #1972

PR #2856, opened 2024-04-15
Replace pulling metrics port with using direct config

jaronoff97:
**Description:** 
This PR makes it possible to pull the set self-monitoring port from the config directly as opposed to working off interfaces and strings.

NOTE: I'm trying to minimize changes, because the next one is going to be... large... [example](https://github.com/open-telemetry/opentelemetry-operator/compare/main...jaronoff97:opentelemetry-operator:not-doing-scrap?expand=1)

This is the first part of #2603 

**Testing:**
Unit tests

**Documentation:**
n/a


iblancasa (approved):

apis/v1beta1/config.go
```diff
@@ -135,6 +127,57 @@ func (c Config) nullObjects() []string {
...
+	// Resource specifies user-defined attributes to include with all emitted telemetry.
+	// Note that some attributes are added automatically (e.g. service.version) even
+	// if they are not specified here. In order to suppress such attributes the
+	// attribute must be specified in this map with null YAML value (nil string pointer).
+	Resource map[string]*string `json:"resource,omitempty" yaml:"resource,omitempty"`
+}
+
+// GetTelemetry serves as a helper function to access the fields we care about in the underlying telemetry struct.
+// This exists to avoid needing to worry extra fields in the telemetry struct.
+func (s *Service) GetTelemetry() *Telemetry {
```
swiatekm-sumo, line 164: I wonder if we should have a method on Config to convert it to a more constrained and strictly typed struct containing only information we make use of in the operator. Not a request to do it in this PR, just a thought.

On a separate note, should we cache this to avoid the marshalling and unmarshalling whenever it's accessed?

apis/v1beta1/config.go
```diff
@@ -135,6 +127,57 @@ func (c Config) nullObjects() []string {
...
+	// Convert map to JSON bytes
+	jsonData, err := json.Marshal(s.Telemetry)
+	if err != nil {
+		return nil
+	}
+	t := &Telemetry{}
+	// Unmarshal JSON into the provided struct
+	if err := json.Unmarshal(jsonData, t); err != nil {
+		return nil
+	}
```
swiatekm-sumo, line 177: Do you think we should check this in the admission webhook, so we can be reasonably sure it works? Or is it better to be more permissive and then simply ignore functionality that depends on this value?

apis/v1beta1/config.go
jaronoff97: Yeah that feels like a good follow up to this – both of those are things we should figure out _after_ we simplify all of the methods that use only maps and strings. Caching would get simpler if we only pass around that constrained struct (because we would probably only create it once)

apis/v1beta1/config.go
jaronoff97: I think its better to be permissive IMO, it's much closer to what we do today (i.e. it's fine if someone doesn't set this)... I think we could maybe introduce a "strict" operator mode as part of [this issue's](https://github.com/open-telemetry/opentelemetry-operator/issues/2399) resolution

swiatekm-sumo (approved):

apis/v1beta1/config.go
swiatekm-sumo: I was thinking about the situation where the parsing fails, that is, the format is inconsistent with what we expect. I *think*, other than a breaking change in the respective collector core package, this would be safe to do.

With that said, breaking changes in collector core packages can happen, and it'd be unfortunate for us to need to always do a breaking change in the CRD as a result.

jaronoff97: Yeah I think once everything in the collector is marked stable we could do this... they're working on it!

PR #2859, opened 2024-04-16
Add missing per-node strategy to TA CRD

swiatekm-sumo:
Partially addresses #2860.

### DIFF
--- a/apis/v1beta1/allocation_strategy.go
+++ b/apis/v1beta1/allocation_strategy.go
@@ -16,7 +16,7 @@ package v1beta1
 
 type (
 	// TargetAllocatorAllocationStrategy represent a strategy Target Allocator uses to distribute targets to each collector
-	// +kubebuilder:validation:Enum=least-weighted;consistent-hashing
+	// +kubebuilder:validation:Enum=least-weighted;consistent-hashing;per-node
 	TargetAllocatorAllocationStrategy string
 	// TargetAllocatorFilterStrategy represent a filtering strategy for targets before they are assigned to collectors
 	// +kubebuilder:validation:Enum="";relabel-config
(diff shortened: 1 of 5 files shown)


apis/v1beta1/allocation_strategy.go
```diff
@@ -30,6 +30,9 @@ const (
 	// TargetAllocatorAllocationStrategyConsistentHashing targets will be consistently added to collectors, which allows a high-availability setup.
 	TargetAllocatorAllocationStrategyConsistentHashing TargetAllocatorAllocationStrategy = "consistent-hashing"
 
+	// TargetAllocatorAllocationStrategyConsistentHashingPerNode targets will be assigned to the collector on the node they reside on (use only with daemon set).
+	OpenTelemetryTargetAllocatorAllocationStrategyPerNode TargetAllocatorAllocationStrategy = "per-node"
```
pavolloffay, line 34: It would be good to align the name with the other ones e.g. `TargetAllocatorAllocationStrategyPerNode`

pavolloffay:
I would consider changing the conversion from code in case a new strategy is added to the v1beta1 in future to avoid failures

pavolloffay:
it can be done in a separate PR

apis/v1beta1/allocation_strategy.go
```diff
@@ -30,6 +30,9 @@ const (
 	// TargetAllocatorAllocationStrategyConsistentHashing targets will be consistently added to collectors, which allows a high-availability setup.
 	TargetAllocatorAllocationStrategyConsistentHashing TargetAllocatorAllocationStrategy = "consistent-hashing"
 
+	// TargetAllocatorAllocationStrategyConsistentHashingPerNode targets will be assigned to the collector on the node they reside on (use only with daemon set).
```
pavolloffay, line 33: The comment should be fixed

swiatekm-sumo:
> I would consider changing the conversion from code in case a new strategy is added to the v1beta1 in future to avoid failures

You mean like, writing an explicit conversion functions between the v1beta1 and v1alpha1 enums? Sure, I can do that.

Issue #2867, opened 2024-04-17
Target allocator example from readme fails


pavolloffay:
### Component(s)

target allocator

### What happened?

https://github.com/open-telemetry/opentelemetry-operator?tab=readme-ov-file#target-allocator

```
kubectl apply -f - <<EOF
apiVersion: opentelemetry.io/v1alpha1
kind: OpenTelemetryCollector
metadata:
  name: collector-with-ta
spec:
  mode: statefulset
  targetAllocator:
    enabled: true
  config: |
    receivers:
      prometheus:
        config:
          scrape_configs:
          - job_name: 'otel-collector'
            scrape_interval: 10s
            static_configs:
            - targets: [ '0.0.0.0:8888' ]
            metric_relabel_configs:
            - action: labeldrop
              regex: (id|name)
            - action: labelmap
              regex: label_(.+)
              replacement: $$1

    exporters:
      debug: {}

    service:
      pipelines:
        metrics:
          receivers: [prometheus]
          processors: []
          exporters: [debug]
EOF
```


### Kubernetes Version

minikube version: v1.30.1

### Operator version

main

### Collector version

0.98.0

### Environment information

## Environment
OS: (e.g., "Ubuntu 20.04")
Compiler(if manually compiled): (e.g., "go 14.2")


### Log output

```shell
k get pods                                                                                                                                                                                                                                                                                                                                                1 ↵ ploffay@fedora
NAME                                                 READY   STATUS             RESTARTS      AGE
collector-with-ta-collector-0                        0/1     CrashLoopBackOff   4 (8s ago)    104s
collector-with-ta-targetallocator-5bcd4d4fb5-cwq2t   0/1     CrashLoopBackOff   4 (17s ago)   104s

```
k logs collector-with-ta-targetallocator-5bcd4d4fb5-cwq2t                                                                                                                                                                                                                                                                                                     ploffay@fedora
Failed to load config: error unmarshaling YAML: "69141" is invalid 'replacement' for labelmap action%

```

```
```


### Additional context

_No response_

**Comment by pavolloffay** [comment]
```
k get cm collector-with-ta-targetallocator -o yaml                                                                                                                                                                                                                                                                                                            ploffay@fedora
apiVersion: v1
data:
  targetallocator.yaml: |
    allocation_strategy: consistent-hashing
    collector_selector:
      matchlabels:
        app.kubernetes.io/component: opentelemetry-collector
[...]

PR #2872, opened 2024-04-17
Fix a metric relabel config unescaping bug

swiatekm-sumo:
**Description:**
Fixed a bug where unescaping `$$` values in prometheus receiver configuration wouldn't apply to metric relabel configs if target relabel configs weren't also present.

**Link to tracking Issue(s):** #2867 

- Resolves: #2867 

**Testing:**
Added more unit tests to the unescaping function. Also added a test to TargetAllocator CR generation, and modified a E2E test to specifically check this configuration

### DIFF
--- a/internal/manifests/targetallocator/adapters/config_to_prom_config.go
+++ b/internal/manifests/targetallocator/adapters/config_to_prom_config.go
@@ -134,62 +134,58 @@ func UnescapeDollarSignsInPromConfig(cfg string) (map[interface{}]interface{}, e
 
 	for i, scrapeConfig := range scrapeConfigs {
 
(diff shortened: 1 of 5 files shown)


internal/manifests/targetallocator/adapters/config_to_prom_config.go
```diff
@@ -134,62 +134,58 @@ func UnescapeDollarSignsInPromConfig(cfg string) (map[interface{}]interface{}, e
 
 	for i, scrapeConfig := range scrapeConfigs {
 
-		relabelConfigsProperty, ok := scrapeConfig["relabel_configs"]
```
jaronoff97, line 137: this logic is so :( i want to clean it up in a future PR if we can

jaronoff97 (approved):

PR #2573, opened 2024-01-29
Add missing label servicemonitor

yuriolisa:
**Description:** <Describe what has changed.>

Add missing label for Service/Pod Monitors
**Link to tracking Issue:** <Issue number if applicable>
Resolves #2251 
**Testing:** <Describe what testing was performed and which tests were added.>
Added unit and e2e tests. 
**Documentation:** <Describe the documentation added.>

PR #2878, opened 2024-04-18
Fix labels for Service Monitors

### LINKED ISSUE #2877: Missing notation for Services broke integration with ServiceMonitor
### Component(s)

collector

## Description
When we create a collector instance with `prometheus` exporter, the created `ServiceMonitor` requires the following label selectors:
```yaml
  selector:
    matchLabels:
      app.kubernetes.io/instance: chainsaw-otlp-metrics.cluster-collector
      app.kubernetes.io/part-of: opentelemetry
      operator.opentelemetry.io/collector-monitoring-service: Exists
```

The label ` operator.opentelemetry.io/collector-monitoring-service: Exists` is available in the `Monitoring` service but not in the `collector` nor `collector-headless` services. Due to that, the `ServiceMonitor` cannot match the `Service` for the OTEL Collector and scrape the metrics from the Prometheus exporter.

### Kubernetes Version

.

### Operator version

.

### Collector version

.

## Environment
OS: (e.g., "Ubuntu 20.04")

### Log output

_No response_

### Additional context

_No response_

### LINKED ISSUE #2965: Collector service monitor does not scape prometheus exporter
### Component(s)

collector

### What happened?

## Description

```yaml
  observability:
    metrics:
      enableMetrics: true
```

Creates service monitor, however only pipeline collector metrics are scraped (port 8888).



## Steps to Reproduce
```
kubectl apply -f - <<EOF
kind: OpenTelemetryCollector
metadata:
  name: simplest
spec:
  observability:
    metrics:
      enableMetrics: true
  config:
    receivers:
      otlp:
        protocols:
          grpc: {}
          http: {}
    processors:
      memory_limiter:
        check_interval: 1s
        limit_percentage: 75
        spike_limit_percentage: 15
      batch:
        send_batch_size: 10000
        timeout: 10s

    exporters:
      debug: {}
      prometheus:
        endpoint: 0.0.0.0:8889
        resource_to_telemetry_conversion:
          enabled: true # by default resource attributes are dropped

    service:
      pipelines:
        metrics:
          receivers: [otlp]
          processors: []
          exporters: [debug, prometheus]
EOF
```

```yaml
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  creationTimestamp: "2024-05-17T07:17:43Z"
  generation: 1
  labels:
    app.kubernetes.io/instance: default.simplest
    app.kubernetes.io/name: simplest-collector
    app.kubernetes.io/part-of: opentelemetry
    app.kubernetes.io/version: latest
  name: simplest-collector
  namespace: default
  ownerReferences:
  - apiVersion: opentelemetry.io/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: OpenTelemetryCollector
    name: simplest
    uid: 4f7a1341-4c03-45e4-b6f0-6f0036eba791
  resourceVersion: "52828"
  uid: 7e9c6a00-7c95-4b67-8894-7f9a98f661b0
spec:
  endpoints:
  - port: monitoring
  - port: prometheus
  namespaceSelector:
    matchNames:
    - default
  selector:
    matchLabels:
      app.kubernetes.io/instance: default.simplest
      app.kubernetes.io/part-of: opentelemetry
      operator.opentelemetry.io/collector-monitoring-service: Exists  # <---- this matches only the monitoring service

```


```yaml
k get service simplest-collector-monitoring -o yaml                                                                                                                                                                                                                                                                                                           ploffay@fedora
apiVersion: v1
kind: Service
metadata:
  annotations:
  creationTimestamp: "2024-05-17T07:17:43Z"
  labels:
    app.kubernetes.io/instance: default.simplest
    app.kubernetes.io/name: simplest-collector-monitoring
    app.kubernetes.io/part-of: opentelemetry
    app.kubernetes.io/version: latest
    operator.opentelemetry.io/collector-monitoring-service: Exists
  name: simplest-collector-monitoring
  namespace: default
  ownerReferences:
  - apiVersion: opentelemetry.io/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: OpenTelemetryCollector
    name: simplest
    uid: 4f7a1341-4c03-45e4-b6f0-6f0036eba791
  resourceVersion: "52825"
  uid: f9805b36-d45e-4014-9408-a22f2d634962
spec:
  clusterIP: 10.96.227.230
  clusterIPs:
  - 10.96.227.230
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: monitoring
    port: 8888
    protocol: TCP
    targetPort: 8888
  selector:
    app.kubernetes.io/instance: default.simplest
    app.kubernetes.io/part-of: opentelemetry
  sessionAffinity: None
  type: ClusterIP
status:
  loadBalancer: {}

```


```yaml
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: "2024-05-17T07:17:43Z"
  labels:
    app.kubernetes.io/instance: default.simplest
    app.kubernetes.io/name: simplest-collector
    app.kubernetes.io/part-of: opentelemetry
    app.kubernetes.io/version: latest
  name: simplest-collector
  namespace: default
  ownerReferences:
  - apiVersion: opentelemetry.io/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: OpenTelemetryCollector
    name: simplest
    uid: 4f7a1341-4c03-45e4-b6f0-6f0036eba791
  resourceVersion: "52817"
  uid: dcfa3688-b807-4a9a-963c-ae5ef646871f
spec:
  clusterIP: 10.108.22.138
  clusterIPs:
  - 10.108.22.138
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - appProtocol: grpc
    name: otlp-grpc
    port: 4317
    protocol: TCP
    targetPort: 4317
  - appProtocol: http
    name: otlp-http
    port: 4318
    protocol: TCP
    targetPort: 4318
  - name: prometheus
    port: 8889
    protocol: TCP
    targetPort: 8889
  selector:
    app.kubernetes.io/instance: default.simplest
    app.kubernetes.io/part-of: opentelemetry
  sessionAffinity: None
  type: ClusterIP
status:
  loadBalancer: {}

```


## Expected Result

## Actual Result


### Kubernetes Version

latest

### Operator version

0.99.0

### Collector version

0.99.0

### Environment information

## Environment
OS: (e.g., "Ubuntu 20.04")


### Log output

_No response_

### Additional context

_No response_

iblancasa:
**Description:**: fix the labels for the Service Monitor integration. Exclude the headless Service from the integration with SM.

Resolves #2965 
Resolves #2877


iblancasa:
@IshwarKanse can you check if the behaviour is correct now? The test passed in my cluster but I need another couple of eyes.

internal/manifests/collector/service.go
```diff
@@ -31,9 +31,7 @@ import (
 
 // headless and monitoring labels are to differentiate the headless/monitoring services from the clusterIP service.
 const (
-	headlessLabel   = "operator.opentelemetry.io/collector-headless-service"
```
jaronoff97, line 34: IMO we should keep these labels to not make a breaking change for people who rely on these

internal/manifests/collector/servicemonitor.go
```diff
@@ -52,7 +52,8 @@ func ServiceMonitor(params manifests.Params) (*monitoringv1.ServiceMonitor, erro
 	name := naming.ServiceMonitor(params.OtelCol.Name)
 	labels := manifestutils.Labels(params.OtelCol.ObjectMeta, name, params.OtelCol.Spec.Image, ComponentOpenTelemetryCollector, []string{})
 	selectorLabels := manifestutils.SelectorLabels(params.OtelCol.ObjectMeta, ComponentOpenTelemetryCollector)
-	selectorLabels[monitoringLabel] = valueExists
+	// This label is to match only some services and exclude others like headless
+	selectorLabels[matchSMLabel] = "true"
```
jaronoff97, line 56: this is how we potentially create double targets which we should avoid.

jaronoff97 (changes requested):
IMO rather than creating one service monitor for the self scrape and the prom exporter (which I think are on different ports) we should just make two service monitors. I worry that this is going to break users because it will cause dual scraping

internal/manifests/collector/service.go
iblancasa: Ok, I'll change it.

internal/manifests/collector/servicemonitor.go
iblancasa: Why? I can change it but I want to understand.

iblancasa:
> > @IshwarKanse can you check if the behaviour is correct now? The test passed in my cluster but I need another couple of eyes.
> 
> AFAIK, we fixed that [here](https://github.com/open-telemetry/opentelemetry-operator/pull/2573).

The SM has the label for the monitor Service but not for the collector. So, it doesn't match the collector and the Prometheus exporter is never scrapped.

internal/manifests/collector/servicemonitor.go
jaronoff97: if service A and service B both point to the same ports on the same collector a servicemonitor that matches both services will generate two targets for the same collector pod

jaronoff97: yep, exactly right:

```mermaid
stateDiagram-v2
    classDef page fill:green
    classDef svc fill:blue

    ServiceMonitor --> collector_svc_monitoring
    ServiceMonitor  --> collector_svc
    collector_svc_monitoring --> collector_crd
    collector_svc --> collector_crd
    collector_crd --> collector_pod_1
    collector_crd --> collector_pod_2
    collector_crd --> collector_pod_3

    class collector_svc, collector_svc_monitoring svc
    class collector_pod_1, collector_pod_2, collector_pod_3 page
```

pavolloffay:
it seems like e2e tests are failing

IshwarKanse:
The PR needs to be updated.

IshwarKanse:
@iblancasa We need to update this test as well [e2e-openshift/otlp-metrics-traces ](https://github.com/open-telemetry/opentelemetry-operator/tree/main/tests/e2e-openshift/otlp-metrics-traces)

internal/manifests/collector/servicemonitor.go
```diff
@@ -29,30 +30,38 @@ import (
...
+	endpoints := []monitoringv1.Endpoint{
+		{
+			Port: "monitoring",
+		},
+	}
+	return createServiceMonitor(name, params, monitoringLabel, endpoints)
+}
 
-	if params.OtelCol.Spec.Mode == v1beta1.ModeSidecar {
+func createServiceMonitor(name string, params manifests.Params, label string, endpoints []monitoringv1.Endpoint) (*monitoringv1.ServiceMonitor, error) {
```
jaronoff97, line 54: can you add a doc string for explaining what the values should be here?

internal/manifests/collector/service.go
```diff
@@ -32,6 +32,7 @@ import (
 // headless and monitoring labels are to differentiate the headless/monitoring services from the clusterIP service.
 const (
 	headlessLabel   = "operator.opentelemetry.io/collector-headless-service"
+	headfulLabel    = "operator.opentelemetry.io/collector-service"
 	monitoringLabel = "operator.opentelemetry.io/collector-monitoring-service"
```
jaronoff97, line 36: This _may_ be a breaking change, but i'm not 100% sure... I think it may be better for us to do something like:
```
operator.opentelemetry.io/collector-service-type: "headless|monitoring|base"
```

or something like that. I think changing the labels on a service are fine as long as you don't change the selector right?

jaronoff97 (commented):
approach broadly makes sense, a few thoughts/questions.

internal/manifests/collector/service.go
iblancasa: It should be OK if users are just relying on the Service Monitors but not sure if they are doing something else.
How about keeping the `operator.opentelemetry.io/collector-monitoring-service` and `operator.opentelemetry.io/collector-headless-service` labels, mark them as deprecated at least in the documentation and use the `operator.opentelemetry.io/collector-service-type` for the `ServiceMonitors`?

jaronoff97: i would be okay with that... @swiatekm-sumo do you have any opinions?

internal/manifests/collector/podmonitor.go
```diff
@@ -107,3 +93,22 @@ func metricsEndpointsFromConfig(logger logr.Logger, otelcol v1beta1.OpenTelemetr
...
 	return metricsEndpoints
 }
+
+func shouldCreatePodMonitor(params manifests.Params) bool {
+	l := params.Log.WithValues(
+		"params.OtelCol.name", params.OtelCol.Name,
+		"params.OtelCol.namespace", params.OtelCol.Namespace,
+	)
+
+	if !params.OtelCol.Spec.Observability.Metrics.EnableMetrics {
```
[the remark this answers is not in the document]
iblancasa: Why a double condition + no log message is easier to read than the current approach?

The current is easy to follow and understand why the method returns `false`.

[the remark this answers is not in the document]
pavolloffay: A single log message with all parameters makes it clear to understand the overall state. 

The function has 2 states (`true` and `false`), aligning the condition to it makes it easier to understand

internal/manifests/collector/service.go
swiatekm-sumo: Correct me if I'm wrong, but the problem we're trying to solve here is that we don't want the monitoring ServiceMonitor to select the headless Service, right? In that case, wouldn't it be simplest to add an explicit exclusion to the label selector?

iblancasa: This PR tries to solve two different issues:
1. Exclude the headless service from the ServiceMonitor
2. The current `ServiceMonitor` is not finding the `collector` service (so, the Prometheus exporter is not working with that automatically created `ServiceMonitor`). It worked before.

swiatekm-sumo: Ok, I'm a bit lost with this PR, but this is how I think it should work:

    We have separate ServiceMonitors for Collector metrics and metrics from Prometheus exporter.
    They don't share any metrics, although they can share the same label selector.
    In the above case, the ports should be different.
    Both of them should exclude the headless Service based on its label.

Is there a reason this can't work? @jaronoff97 let me know what you think as well.

jaronoff97: yes, that's what I think the solution should be too – we would probably have each of the servicemonitor's ports use the proper names too

iblancasa: Ok, since we agree, I'll implement the approach suggested by @jaronoff97

swiatekm-sumo (approved):
LGTM, minor nitpick about the changelog.

pavolloffay (approved):

pavolloffay:
@jaronoff97 could you please re-review?

internal/manifests/collector/service.go
```diff
@@ -29,21 +29,31 @@ import (
 	"github.com/open-telemetry/opentelemetry-operator/internal/naming"
 )
 
-// headless and monitoring labels are to differentiate the headless/monitoring services from the clusterIP service.
+// headless and monitoring labels are to differentiate the base/headless/monitoring services from the clusterIP service.
 const (
-	headlessLabel   = "operator.opentelemetry.io/collector-headless-service"
-	monitoringLabel = "operator.opentelemetry.io/collector-monitoring-service"
```
jaronoff97, line 35: @iblancasa i think we should keep these around for now, mark them as deprecated and remove them in two+ versions rather than deleting them outright which would constitute a breaking change.

jaronoff97 (commented):
Just one thought to prevent this from breaking users

internal/manifests/collector/service.go
iblancasa: Done

tests/e2e-prometheuscr/create-sm-prometheus-exporters/05-assert.yaml
```diff
@@ -9,19 +32,18 @@ metadata:
...
     matchNames:
     - create-sm-prometheus
   selector:
     matchLabels:
+      app.kubernetes.io/component: opentelemetry-collector
       app.kubernetes.io/instance: create-sm-prometheus.simplest
       app.kubernetes.io/managed-by: opentelemetry-operator
       app.kubernetes.io/part-of: opentelemetry
-      app.kubernetes.io/component: opentelemetry-collector
-      operator.opentelemetry.io/collector-monitoring-service: "Exists"
```
jaronoff97, line 24: @iblancasa i think these tests need to be updated too

jaronoff97 (approved):

iblancasa:
I think it should be fine now. I'm having some problems running this locally due to some network restrictions...

jaronoff97:
@iblancasa looks like you missed one test https://github.com/open-telemetry/opentelemetry-operator/actions/runs/9131574679/job/25111089581?pr=2878

iblancasa:
@jaronoff97 thanks! I fixed the issue now. I added the label in an extra service.

pavolloffay (approved):

PR #2896, opened 2024-04-24
Introduce typed pipeline

jaronoff97:
**Description:** Introduces a typed pipeline for the collector. This allows us to really easily construct a list of enabled components. 

NOTE! One decision we need to make:

By using a struct like this, we have to figure out how to handle an empty list of processors. I chose to always include it on marshal i.e. if someone has a pipeline configuration like this:
```

service:
  extensions:
    - oauth2client
  pipelines:
    traces:
      receivers:
        - otlp
      exporters:
        - otlp/auth
```

we will then unmarshal and remarshal it to be:

```

service:
  extensions:
    - oauth2client
  pipelines:
    traces:
      receivers:
        - otlp
      processors: []
      exporters:
        - otlp/auth

```

IMO this is fine and better than the reverse where we set it explicitly to an empty list and then don't remarshal it with that empty list. If someone has a way to make this work both ways, let me know :) 

## DECISION
From the SIG we decided this is fine, it's only visible to a user in their configmap, but no functionality difference.

part of #2603 

**Testing:** unit testing

**Documentation:** n/a


apis/v1beta1/config.go
```diff
@@ -75,6 +87,37 @@ func (c *AnyConfig) MarshalJSON() ([]byte, error) {
...
 
+// Pipeline is a struct of component type to a list of component IDs.
+type Pipeline struct {
+	Receivers  []string `json:"receivers" yaml:"receivers"`
+	Processors []string `json:"processors" yaml:"processors"`
+	Exporters  []string `json:"exporters" yaml:"exporters"`
+}
+
+// GetEnabledComponents constructs a list of enabled components by component type.
+func (c *Config) GetEnabledComponents() map[ComponentType]map[string]interface{} {
```
jaronoff97, line 98: eventually this will replace the way we do this today in config_validate.go but we can't do that for now...

pavolloffay: How are the connectors handled?

jaronoff97: connectors don't have their own definition in a pipeline, they are defined in the config and [used as exporters and receivers](https://github.com/open-telemetry/opentelemetry-collector/tree/main/connector#usage), this will work as expected for them as a result :)

apis/v1beta1/config.go
```diff
@@ -75,6 +87,37 @@ func (c *AnyConfig) MarshalJSON() ([]byte, error) {
 	return json.Marshal(c.Object)
 }
 
+// Pipeline is a struct of component type to a list of component IDs.
+type Pipeline struct {
+	Receivers  []string `json:"receivers" yaml:"receivers"`
+	Processors []string `json:"processors" yaml:"processors"`
+	Exporters  []string `json:"exporters" yaml:"exporters"`
+}
```
swiatekm-sumo, line 95: I think the fields are marshalled in the order they're declared in the struct. So if you change the order to alphabetical, we can avoid a bunch of test changes due to ordering.

jaronoff97: I had put them in this order because of how people conceptually think of a pipeline, but i guess this isn't really meant to be user read, so it doesn't really matter. Will change it back!

apis/v1beta1/config_test.go
```diff
@@ -145,11 +145,11 @@ func TestConfigYaml(t *testing.T) {
...
-			Pipelines: AnyConfig{
-				Object: map[string]interface{}{
-					"receivers":  []string{"otlp"},
-					"processors": []string{"modify_2000"},
-					"exporters":  []string{"otlp/exporter", "con"},
+			Pipelines: map[string]*Pipeline{
+				"traces": {
+					Receivers:  []string{"otlp"},
+					Processors: []string{"modify_2000"},
+					Exporters:  []string{"otlp/exporter", "con"},
```
swiatekm-sumo, line 152: This was incorrect, right?

apis/v1beta1/config.go
```diff
@@ -75,6 +87,37 @@ func (c *AnyConfig) MarshalJSON() ([]byte, error) {
...
+	Receivers  []string `json:"receivers" yaml:"receivers"`
+}
+
+// GetEnabledComponents constructs a list of enabled components by component type.
+func (c *Config) GetEnabledComponents() map[ComponentType]map[string]interface{} {
+	toReturn := map[ComponentType]map[string]interface{}{
+		ComponentTypeReceiver:  {},
+		ComponentTypeProcessor: {},
+		ComponentTypeExporter:  {},
+	}
```
swiatekm-sumo, line 103: There's a slight ambiguity here. We return connectors as both exporters and receivers, which is technically correct as far as their role in the pipeline goes, but I wonder if it won't cause confusion in the future. If I were a caller of this function, I'd expect to only get actual receivers in the Receiver map.

swiatekm-sumo (commented):
LGTM overall, with one question I'd like to answer before we merge.

apis/v1beta1/config_test.go
jaronoff97: yes, fully incorrect before :)

apis/v1beta1/config.go
jaronoff97: iirc this is technically the same as we would do today. After we do #2915 we could probably separate these out more effectively. Right now this is only used for parsing logic (generating RBAC and Ports) but obv in the future this should probably contain enabled extensions and connectors.

swiatekm-sumo (approved):

pavolloffay (approved):

PR #2898, opened 2024-04-25
Change default otel image to the K8s distro

swiatekm:
**Description:**
This is meant as an example of how the default image conversion between v1alpha1 and v1beta1 could work. It explicitly lists all the components and checks configuration for invalid ones for the target distro. This way, if the configs are compatible, we can freely convert back and forth without changing anything. If the configs are not compatible, we explicitly set the `Image` at the source CR's default.

Note that some of our e2e tests use the prometheus exporter, which isn't available in the K8s distro, so them passing shows the mechanism actually working.

**Link to tracking Issue(s):** #2835


swiatekm-sumo:
During the SIG meeting on 25.04.2024, we've decided to approach this migration differently.

PR #2899, opened 2024-04-25
Fix hashing bug

jaronoff97:
**Description:** 
Fixes a bug where we were propagating the last applied configuration down the stack which was causing unnecessary reconciliation. Also fixes a bug where we accidentally always short circuited.

n/a

**Testing:** 
n/a

**Documentation:** 
n/a


internal/config/options.go
```diff
@@ -219,8 +219,7 @@ func WithLabelFilters(labelFilters []string) Option {
 
 func WithAnnotationFilters(annotationFilters []string) Option {
 	return func(o *options) {
-
-		filters := []string{}
+		filters := o.annotationsFilter
```
swiatekm-sumo, line 222: Why change this? Do we not want `WithAnnotationFilters` to replace existing filters?

swiatekm-sumo (commented):
Is there a reason you don't want a changelog entry for this change? Seems like a normal bugfix to me.

internal/config/options.go
jaronoff97: this is so that I can use the default filter above, otherwise it [will always be overridden](https://github.com/open-telemetry/opentelemetry-operator/blob/main/main.go#L229)

jaronoff97: but also, I kind of expect these to be additive

pavolloffay (approved):

frzifus (approved):

internal/config/options.go
swiatekm-sumo: Ok, fair enough. Can you document that this function is additive, then?

swiatekm-sumo (approved):

PR #2902, opened 2024-04-26
Add document explaining v1beta1 changes

swiatekm-sumo:
I ended up adding a more general "CRD changelog" document, with the first entry explaining the v1alpha1 -> v1beta1 transition. We can put other, similar sections in here in the future, for example introducing new CRDs.

### DIFF
--- a/docs/crd-changelog.md
+++ b/docs/crd-changelog.md
@@ -3,11 +3,33 @@
 This document explains major changes made in new CRD versions. It is intended to help users migrate and take
 advantage of the new features.
 
-## opentelemetry.io/v1beta1 OpenTelemetryCollector
+## OpenTelemetryCollector.opentelemetry.io/v1beta1 
 
 ### Migration
 
-There is no need for any user action. The operator will automatically convert existing `v1alpha1` resources to `v1beta1`.
+There is no need for any immediate user action. The operator will continue to support existing `v1alpha1` resources.
+
+In addition, any newly applied `v1alpha1` resource will be converted to `v1beta1` and stored in the new API version.
+
+The plan is to remove support for `v1alpha1` in a future operator version, so users should migrate promptly. In order to migrate fully to `v1beta1`:
+
+1. Update any manifests you have stored outside the cluster, for example in your infrastructure git repository.
+2. Apply them, so they're all stored as `v1beta1`.
+3. Update the OpenTelemetryCollector CRD to only store `v1beta1`
+   ```bash
+   kubectl patch customresourcedefinitions opentelemetrycollectors.opentelemetry.io  \
+     --subresource='status' \
+     --type='merge' \
+     -p '{"status":{"storedVersions":["v1beta1"]}}'
+   ```
+For a more thorough explanation of how and why this migration works, see the relevant [Kubernetes documentation][crd_migration_guide].
+
+#### Operator Lifecycle Manager
+
+If you're installing the opentelemetry-operator in OpenShift using OLM, be advised that
+**only `AllNamespaces` install mode is now supported**, due to the conversion webhook from `v1beta1` to `v1alpha1`.
+See [OLM docs](https://olm.operatorframework.io/docs/tasks/install-operator-with-olm/) and
+[OLM operator groups docs](https://olm.operatorframework.io/docs/advanced-tasks/operator-scoping-with-operatorgroups/).
 
 ### Structured Configuration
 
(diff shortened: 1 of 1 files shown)


docs/crd-changelog.md
```diff
@@ -0,0 +1,138 @@
+# CRD Changelog
+
+This document explains major changes made in new CRD versions. It is intended to help users migrate and take
+advantage of the new features.
+
+## opentelemetry.io/v1beta1 OpenTelemetryCollector
```
pavolloffay, line 6: This is commonly written as OpenTelemetryCollector.opentelemetry.io/v1beta1

docs/crd-changelog.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Migration
+
+There is no need for any user action. The operator will automatically convert existing `v1alpha1` resources to `v1beta1`.
+
+### Structured Configuration
+
+The `Config` field containing the Collector configuration is a string in `v1alpha1`. This has some downsides:
```
pavolloffay, line 14: I would call out the limitation with kustomize and other tools

docs/crd-changelog.md
```diff
@@ -0,0 +1,138 @@
...
+    service:
+      pipelines:
+        traces:
+          receivers: [otlp]
+          processors: [memory_limiter, batch]
+          exporters: [debug]
+```
+
+> [!NOTE]  
+> Empty maps, like `debug:` in the above configuration, should have an explicit value of `{}`.
```
pavolloffay, line 88: !! :)

docs/crd-changelog.md
```diff
@@ -0,0 +1,160 @@
...
+1. Update any manifests you have stored outside the cluster, for example in your infrastructure git repository.
+2. Apply them, so they're all stored as `v1beta1`.
+3. Update the OpenTelemetryCollector CRD to only store `v1beta1`
+   ```bash
+   kubectl patch customresourcedefinitions opentelemetrycollectors.opentelemetry.io  \
+     --subresource='status' \
+     --type='merge' \
+     -p '{"status":{"storedVersions":["v1beta1"]}}'
+   ```
+For a more thorough explanation of how and why this migration works, see the relevant [Kubernetes documentation][crd_migration guide].
```
jaronoff97, line 24: shouldn't there be a link for the kube documentation there?

jaronoff97: actually this may just be the github preview renderer not working

jaronoff97: no, nvm i think this should be []() syntax right?

docs/crd-changelog.md
swiatekm-sumo: I called it out below, would you prefer it closer to the start of the section?

docs/crd-changelog.md
swiatekm-sumo: Yeah, it was missing an underscore.

docs/crd-changelog.md
```diff
@@ -0,0 +1,160 @@
...
+This document explains major changes made in new CRD versions. It is intended to help users migrate and take
+advantage of the new features.
+
+## OpenTelemetryCollector.opentelemetry.io/v1beta1 
+
+### Migration
+
+There is no need for any immediate user action. The operator will automatically convert existing `v1alpha1` resources to `v1beta1`.
+
+However, the conversion procedure will leave your resources stored as both versions. Support for `v1alpha1` will be removed in the future.
```
pavolloffay, line 12: This is not accurate, but rather misleading. 

>  the conversion procedure will leave your resources stored as both versions.

The custom resources stored as v1alpha1 are not automatically updated to v1beta1. The updated or newly written cutom resources will be stored as v1beta1 because it is the new storage version. 

I would also link https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version

docs/crd-changelog.md
```diff
@@ -0,0 +1,160 @@
+# CRD Changelog
+
+This document explains major changes made in new CRD versions. It is intended to help users migrate and take
+advantage of the new features.
+
+## OpenTelemetryCollector.opentelemetry.io/v1beta1 
+
+### Migration
+
+There is no need for any immediate user action. The operator will automatically convert existing `v1alpha1` resources to `v1beta1`.
```
pavolloffay, line 10: >  The operator will automatically convert existing `v1alpha1` resources to `v1beta1`.

I would rephrase this. Yes there is conversion webhook but for migration from v1alpha1 to v1beta1 an update on the CR must happen

docs/crd-changelog.md
swiatekm-sumo: Ok, how about:

```
There is no need for any immediate user action. The operator will continue to support existing `v1alpha1` resources.

In addition, any newly applied `v1alpha1` resource will be converted to `v1beta1` and stored as both versions. The plan is to remove support for `v1alpha1` in a future version, so users should migrate promptly.

(detailed migration guide here)
```

jaronoff97 (approved):

pavolloffay (approved):

PR #2904, opened 2024-04-27
[chore] Move TargetAllocator CRD to v1alpha1

swiatekm-sumo:
Move the TargetAllocator CRD to v1alpha1. Since the CRD has a bunch of definitions in common with v1beta1.OpenTelemetryCollector, I've moved those to a `common` package. In the process, I've also gotten rid off some unnecessary conversions between v1beta1 and v1alpha1 enums.

Right now, this also includes changes from #2883. That should be merged first, then I'll rebase.

Supersedes #2894.


.chloggen/fix_load-initial-servicemonitors.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: target allocator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix target allocator readiness check
```
pavolloffay, line 8: Does not not seem to be accurate

apis/common/common.go
```diff
@@ -12,7 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package v1beta1
+// +kubebuilder:object:generate=true
+// +groupName=opentelemetry.io
+package common
```
pavolloffay, line 17: I am concerned about using a common package for multiple versions. I haven't seen this being the case in other projects or k8s. All symbols are usually clearly versioned.

swiatekm-sumo: Would it help to alias them? I don't really want to copy all the shared structs and then define pointless conversions between them.

.chloggen/fix_load-initial-servicemonitors.yaml
swiatekm-sumo: This is from #2883, it'll disappear once that's merged. This PR is a bit messy in general right now, I published it to see if there's any fundamental objections to my approach here.

swiatekm-sumo:
@pavolloffay this is the closest I can get to having mostly versioned symbols. The common structs are part of an internal package. I've aliased the enums, but for more complex types this causes more problems than it solves. The alternative would be to copy the shared structs to `v1alpha1`, which I don't really like the idea of.

apis/v1beta1/targetallocator_types.go
```diff
@@ -15,125 +15,28 @@
...
-	TargetAllocatorAllocationStrategy string
-	// TargetAllocatorFilterStrategy represent a filtering strategy for targets before they are assigned to collectors
-	// +kubebuilder:validation:Enum="";relabel-config
-	TargetAllocatorFilterStrategy string
+	// TargetAllocatorPrometheusCR configures Prometheus CustomResource handling in the Target Allocator.
+	TargetAllocatorPrometheusCR common.TargetAllocatorPrometheusCR
+	// TargetAllocatorAllocationStrategy represent a strategy Target Allocator uses to distribute targets to each collector.
+	TargetAllocatorAllocationStrategy common.TargetAllocatorAllocationStrategy
+	// TargetAllocatorFilterStrategy represent a filtering strategy for targets before they are assigned to collectors.
+	TargetAllocatorFilterStrategy common.TargetAllocatorFilterStrategy
```
jaronoff97, line 27: should this remain here? Or is this because we need to maintain a v1alpha1 and v1beta1 for this now?

jaronoff97:
I had one question about the TA_types.go file. My only gripe with this is referencing the internal package from files that we do publicly expose. Is it not possible to just reference the v1beta1 type from the v1alpha1? It looks like that's what [prometheus operator](https://github.com/prometheus-operator/prometheus-operator/blob/5a8ab71c21639f471595088c12f58fc32160c21c/pkg/apis/monitoring/v1alpha1/prometheusagent_types.go#L29-L31) may do?

apis/common/common.go
pavolloffay: internal could work, but I thought that the symbols will be exposed in the CRD

swiatekm-sumo:
Closing in favor of #2918

PR #2915, opened 2024-04-30
Define components more clearly in v1beta1

jaronoff97:
**Description:** 
Defines components as a map from component name to an AnyConfig

Part of #2603 

**Testing:** unit tests

**Documentation:** n/a


apis/v1beta1/config.go
```diff
@@ -121,16 +136,16 @@ func (c *Config) GetEnabledComponents() map[ComponentType]map[string]interface{}
 // Config encapsulates collector config.
 type Config struct {
 	// +kubebuilder:pruning:PreserveUnknownFields
-	Receivers AnyConfig `json:"receivers" yaml:"receivers"`
+	Receivers map[string]*AnyConfig `json:"receivers" yaml:"receivers"`
```
pavolloffay, line 139: Could you explain why this change is needed? `AnyConfig` is already a map 

```go
type AnyConfig struct {
	Object map[string]interface{} `json:"-" yaml:",inline"`
}
```

jaronoff97:
I cannot make this work for the life of me. Closing and will try a different approach :sob:

PR #2918, opened 2024-05-01
[chore] Move TargetAllocator CRD to v1alpha1

swiatekm-sumo:
Alternative approach to #2904. Here we keep the shared stuff in `v1beta1`, similar to how prometheus-operator does it.


pavolloffay:
> Here we keep the shared stuff in v1beta1, similar to how prometheus-operator does it.

Could you please share a link to the prom operator?

swiatekm-sumo:
> > Here we keep the shared stuff in v1beta1, similar to how prometheus-operator does it.
> 
> Could you please share a link to the prom operator?

https://github.com/prometheus-operator/prometheus-operator/blob/5a8ab71c21639f471595088c12f58fc32160c21c/pkg/apis/monitoring/v1alpha1/prometheusagent_types.go#L29-L31

pavolloffay (approved):

jaronoff97 (approved):

Issue #1669, opened 2023-04-20
Target Allocator - ServiceMonitor scheme


rhysxevans:
Hi

I have the TA up and with the `PrometheusCR` enabled, this is to look at options of migrating to using the OTEL collector for more purposes.

At present I am getting data as expected except for service monitors (to be fair in my test lab I only have a couple of these setup) that contain a "non default" scheme and have authentication data in them. eg

```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  creationTimestamp: "2023-04-06T18:07:21Z"
  generation: 1
  labels:
    app: first-cluster
  name: first-cluster
  namespace: opensearch-first-cluster
  resourceVersion: "991661"
  uid: cfad90db-7361-4b05-96de-b89b36537dbc
spec:
  endpoints:
  - basicAuth:
      password:
        key: password
        name: first-cluster-opensearch-monitoring
      username:
        key: username
        name: first-cluster-opensearch-monitoring
    interval: 30s
    path: /_prometheus/metrics
    port: http
    scheme: https
    tlsConfig:
      insecureSkipVerify: true
  jobLabel: opensearch_first-cluster
  namespaceSelector:
    matchNames:
    - opensearch-first-cluster
  selector:
    matchLabels:
      opster.io/opensearch-cluster: first-cluster
```

```
apiVersion: v1
kind: Service
metadata:
  annotations:
[...]

PR #2921, opened 2024-05-01
Ta https server

ItielOlenick:
Added an additional HTTPS server with mTLS to serve scrape_config with secret values. This resolves one part of the issue. An additional PR will be opened for the generation and mounting of certs.

- 1st PR towards a solution for #1669

Tested in cluster with self-issued certificates. Successfully retrieved the redacted scrape_config using the existing HTTP server and obtained the scrape_config with actual secret values from the new HTTPS server.

Not yet added. Will be added once the entire feature is available.


cmd/otel-allocator/server/server.go
```diff
@@ -155,32 +216,48 @@ func RemoveRegexFromRelabelAction(jsonConfig []byte) ([]byte, error) {
...
 }
 
 // ScrapeConfigsHandler returns the available scrape configuration discovered by the target allocator.
-func (s *Server) ScrapeConfigsHandler(c *gin.Context) {
+func (s *Server) ScrapeConfigsHandler(c *gin.Context, httpsRouter bool) {
 	s.mtx.RLock()
 	result := s.scrapeConfigResponse
+	if httpsRouter {
+		result = s.ScrapeConfigMarshalledSecretResponse
+	}
```
swiatekm-sumo, line 260: I think that instead of adding this parameter, we can just check if the request is encrypted, and respond based on that.

cmd/otel-allocator/server/server.go
```diff
@@ -155,32 +216,48 @@ func RemoveRegexFromRelabelAction(jsonConfig []byte) ([]byte, error) {
...
+			return err
+		}
+
+		s.mtx.Lock()
+		if marshalSecretValue {
+			s.ScrapeConfigMarshalledSecretResponse = jsonConfigNew
+		} else {
+			s.scrapeConfigResponse = jsonConfigNew
+		}
+		s.mtx.Unlock()
```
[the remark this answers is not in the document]
ItielOlenick: We can marshal both by setting `marshalSecretValues := []bool{false, true} ` and removing
```
if s.httpsServer != nil {
    marshalSecretValues = append(marshalSecretValues, true)
}
```
The loop is there to avoid duplication.

cmd/otel-allocator/main.go
```diff
@@ -189,6 +193,18 @@ func main() {
 				setupLog.Error(shutdownErr, "Error on server shutdown")
 			}
 		})
+	runGroup.Add(
+		func() error {
+			err := srv.StartHTTPS()
+			setupLog.Info("HTTPS Server failed to start")
+			return err
```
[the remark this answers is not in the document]
ItielOlenick: I missed adding the conditional
`if cfg.HTTPS.Enabled`. Will add it.
I believe the HTTPS server wont start without the configuration.

cmd/otel-allocator/server/server.go
ItielOlenick: Good point. That we be cleaner.

cmd/otel-allocator/server/server.go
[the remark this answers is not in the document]
swiatekm-sumo: If we want to avoid duplication, maybe it's better to factor it out into a function? I don't find the loop approach very readable, personally.

cmd/otel-allocator/config/config.go
```diff
@@ -164,3 +203,26 @@ func ValidateConfig(config *Config) error {
 	}
 	return nil
 }
+
+func TLSConfig(CAFile, CertFile, KeyFile string) (*tls.Config, error) {
```
swiatekm-sumo, line 207: nit: can we turn this into a method on the `HTTPSServerConfig` struct?

cmd/otel-allocator/config/config.go
```diff
@@ -36,11 +38,14 @@ import (
...
-	DefaultAllocationStrategy                = "consistent-hashing"
-	DefaultFilterStrategy                    = "relabel-config"
+	DefaultResyncTime                          = 5 * time.Minute
+	DefaultConfigFilePath       string         = "/conf/targetallocator.yaml"
+	DefaultCRScrapeInterval     model.Duration = model.Duration(time.Second * 30)
+	DefaultAllocationStrategy                  = "consistent-hashing"
+	DefaultFilterStrategy                      = "relabel-config"
+	DefaultHttpsCAFilePath                     = "/conf/tls/ca.crt"
+	DefaultHttpsTLSCertFilePath                = "/conf/tls/tls.crt"
+	DefaultHttpsTLSKeyFilePath                 = "/conf/tls/tls.key"
```
swiatekm-sumo, line 48: I wonder if these should have default values at all. What is the benefit of having these?

cmd/otel-allocator/main.go
```diff
@@ -86,8 +86,18 @@ func main() {
...
 	srv := server.NewServer(log, allocator, cfg.ListenAddr)
 
+	if cfg.HTTPS.Enabled {
+		tlsConfig, confErr := config.TLSConfig(cfg.HTTPS.TLSCertFilePath, cfg.HTTPS.TLSKeyFilePath, cfg.HTTPS.CAFilePath)
+		if confErr != nil {
+			setupLog.Error(confErr, "Unable to initialize TLS configuration")
+			os.Exit(1)
+		}
+		srv = server.NewServer(log, allocator, cfg.ListenAddr, server.WithTLSConfig(tlsConfig, cfg.HTTPS.ListenAddr))
+	}
```
swiatekm-sumo, line 99: How about:

```go
httpOptions := []server.Option{}
if cfg.HTTPS.Enabled {
    tlsConfig, confErr := config.TLSConfig(cfg.HTTPS.TLSCertFilePath, cfg.HTTPS.TLSKeyFilePath, cfg.HTTPS.CAFilePath)
    if confErr != nil {
      setupLog.Error(confErr, "Unable to initialize TLS configuration")
      os.Exit(1)
    }
    httpOptions = append(httpOptions, server.WithTLSConfig(tlsConfig, cfg.HTTPS.ListenAddr))          
}
srv := server.NewServer(log, allocator, cfg.ListenAddr, ...httpOptions)
```

swiatekm-sumo (commented):
Left some comments, but overall this looks good to me. @jaronoff97 can you have a look as well?

cmd/otel-allocator/main.go
jaronoff97: +1

cmd/otel-allocator/config/config.go
ItielOlenick: Following the style of having a default value for `DefaultConfigFilePath`.
Should these be different?

swiatekm-sumo: Honestly, that one probably shouldn't have a default either. There are users of standalone target allocator out there, and I try to avoid coupling this configuration too hard to what the operator wants. For these values in particular, since HTTPS is disabled by default, I think the user should be forced to supply them.

cmd/otel-allocator/main.go
```diff
@@ -86,8 +86,20 @@ func main() {
...
 
+	httpOptions := []server.Option{}
+	if cfg.HTTPS.Enabled {
+		tlsConfig, confErr := cfg.HTTPS.NewTLSConfig()
+		if confErr != nil {
+			setupLog.Error(confErr, "Unable to initialize TLS configuration")
+			os.Exit(1)
+		}
+		httpOptions = append(httpOptions, server.WithTLSConfig(tlsConfig, cfg.HTTPS.ListenAddr))
+		srv = server.NewServer(log, allocator, cfg.ListenAddr, httpOptions...)
```
swiatekm-sumo, line 100: You can move this line out of the conditional block, and delete the one in line 90.

ItielOlenick: 100%

swiatekm-sumo (approved):
LGTM! Thank you for your contribution and patience with the review process.

ItielOlenick:
> LGTM! Thank you for your contribution and patience with the review process.

Sure thing. 
Perhaps for the rest of the changes we can have a short discussion about the implementation? Specifically cret management.

swiatekm-sumo:
> > LGTM! Thank you for your contribution and patience with the review process.
> 
> Sure thing. Perhaps for the rest of the changes we can have a short discussion about the implementation? Specifically cret management.

Sure, let's do that in the issue.

cmd/otel-allocator/main.go
```diff
@@ -189,6 +199,20 @@ func main() {
 				setupLog.Error(shutdownErr, "Error on server shutdown")
 			}
 		})
+	if cfg.HTTPS.Enabled {
```
jaronoff97, line 202: do we need to put the above srv.Start() in an else block or are we okay to potentially run both a secure and insecure simultaneously? I guess doing it this way would make for a better migration from http to https.

cmd/otel-allocator/server/server.go
```diff
@@ -172,15 +208,38 @@ func (s *Server) UpdateScrapeConfigResponse(configs map[string]*promconfig.Scrap
...
+// in to a JSON format for consumers to use.
+func (s *Server) UpdateScrapeConfigResponse(configs map[string]*promconfig.ScrapeConfig) error {
+	err := s.MarshalScrapeConfig(configs, false)
+	if err != nil {
+		return err
+	}
+	err = s.MarshalScrapeConfig(configs, true)
+	if err != nil {
+		return err
+	}
```
jaronoff97, line 232: should we switch on if we're on a secure server here instead of trying one and then the other?

ItielOlenick: As we are having both secure and insecure servers running simultaneously, it made sense to keep (and update) both versions

cmd/otel-allocator/main.go
ItielOlenick: Based on previous conversation, it was suggested to have both HTTP and HTTPS running at the same time.

cmd/otel-allocator/server/server.go
jaronoff97: wouldn't this result in us dual writing the scrape config though? Is there no way for us to know if it's the https or http server responding?

ItielOlenick: It will. We are writing both but serving the relevant version based on if the connection is secure or not - https://github.com/open-telemetry/opentelemetry-operator/pull/2921/files/70929155ab0a15976a3180f6524e5c5d5db70128#diff-d586dd0d67f5b7b16ee4093cad5611ced5099b2e6c31e2a47f07248c6048ddaaR240

jaronoff97: ohhhh i see now, thanks for explaining :)

jaronoff97 (approved):

PR #2923, opened 2024-05-03
Check labels annotations filter

yuriolisa:

Fixing the Labels and Annotations filter.

- Resolves: #2770 


internal/manifests/manifestutils/labels.go
```diff
@@ -25,8 +25,9 @@ import (
 )
 
 func IsFilteredSet(sourceSet string, filterSet []string) bool {
```
jaronoff97, line 27: do we need to compile the pattern here? Is it not enough to just match string on it?

yuriolisa: Theoretically, we shouldn't, but without doing that, the regexp match doesn't work as expected.

jaronoff97 (approved):

pavolloffay:
@yuriolisa the PR needs to be rebased

pavolloffay (approved):

swiatekm-sumo (approved):

internal/manifests/manifestutils/labels.go
swiatekm-sumo: What is the difference between them?

Intuinewin:
Hi, the `webhook-port` flag was removed in this PR !!!

Issue #268, opened 2021-05-10
Configure zap EncoderConfig for operator


kaskavalci:
Operator uses default configuration for zap encoder configuration. `levelKey` in this config defaults to `level`. This creates an issue in GCP Logging as it expects `severity` for log level. Below is an example of a log level interpreted as error. 

I've noticed operator uses `sigs.k8s.io/controller-runtime/pkg/log/zap` package to bind directly from environment variables. As far as I can see, that function does not support full configuration of `zapcore.EncoderConfig`. But, it seems it is possible to configure zap completely by using the zap package from controller-runtime. 

If such a feature is reasonable, I'm happy to pick it up to contribute.

```json
{
  "insertId": "z96t3fg2d841ax",
  "jsonPayload": {
    "opentelemetry-collector": "0.24.0",
    "go-os": "linux",
    "opentelemetry-operator": "0.24.0",
    "build-date": "2021-04-22T20:20:45Z",
    "go-version": "go1.15.11",
    "ts": 1620630610.4453816,
    "msg": "Starting the OpenTelemetry Operator",
    "go-arch": "amd64",
    "level": "info"
  },
  "timestamp": "2021-05-10T07:10:10.445834703Z",
  "severity": "ERROR",
  "receiveTimestamp": "2021-05-10T07:10:16.030283595Z"
}
```

jpkrohling:
Are you still interested in this one, @kaskavalci?

kaskavalci:
I do but it's been difficult to integrate this work with summer plans. I put a self imposed deadline but that didn't work either :) Is the issue prioritized? I'm happy to leave it to others or pick it up ASAP I find some time on my hands.

jpkrohling:
It's not prioritized, I would love to have you do it whenever you have time for it.

kaskavalci:
@jpkrohling I'm sorry to keep this holding on me. It seems I cannot allocate any time to ship this feature. So, it's open for grabs!

yuriolisa:
@jpkrohling, @kaskavalci. Could you please clarify what the expected results in this issue are?

yuriolisa:
@jpkrohling.

crossoverJie:
Is this the expected outcome of the issue?

Current:
```json
{"level":"info","ts":"2024-03-12T14:39:02+08:00","msg":"test"}
```

Expect:
```json
{"severity":"info","ts":"2024-03-12T14:41:24+08:00","msg":"test"}
```

[...]

Issue #873, opened 2022-05-16
Document how to enable debug logging in collector


ringerc:
The collector docs don't appear to mention how to control the collector's own log verbosity.

It uses Zap logging, and the k8s collector framework, so the admin may pass `--zap-log-level=debug` to set fine grained logging. This is documented for the operator SDK at https://sdk.operatorframework.io/docs/building-operators/golang/references/logging/

Mentioning the `--zap-log-level` param and linking to the SDK docs would help.

But it'd be nice to tell users how to actually set it, since Kustomize is a typical deployment method, but doing so through Kustomize is non-obvious. See https://github.com/operator-framework/operator-sdk/pull/5766/files for my patch against the operator SDK to add an explanation to its docs. A snippet like

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- https://url-of-opentelemetry-operator/
patches:
  - target:
      kind: Deployment
      name: opentelemetry-operator-controller-manager
      namespace: opentelemetry-opertor-system
    patch: |-
      - op: add
        path: "/spec/template/spec/containers/1/args/-"
        value: "--zap-log-level=debug"
```

or a `kubectl` hot-patch like

```
kubectl patch -n opentelemetry-operator-system $(\
  kubectl get -n opentelemetry-operator-system \
    -l app.kubernetes.io/name=opentelemetry-operator deployment -o name \
  ) \
  --type=json \
  -p '[{"op": "add", "path": "/spec/template/spec/containers/1/args/-", "value": "--zap-log-level=debug"}]' \
  --dry-run=client 
```

(To actually apply, remove `--dry-run=client` from the above)

TBH, it'd be a *lot* nicer to do this with an env-var binding, not a command-line argument. But documenting a simple way to adjust logging would help.

I don't see any increased logging with the above, so I'm not certain it's actually correct.

pavolloffay:
@ringerc could you please comment?

ringerc:
@pavolloffay  I'm suggesting that the docs for the operator cover how to raise the debug level of the operator. Managing the log level of the deployed collector is simple and already well documented.

It should IMO at least mention the `--zap-log-level=debug` option in the README, and link to the collector SDK docs for details.

But it'd be really nice to provide a quick explanation of how to enable it for the most common deployment methods of the operator, like I gave for Kustomize above.

[...]

PR #2927, opened 2024-05-06
Added new Log Enconder Config

yuriolisa:

Added 5 new parameters to enable a customized JSON log Encoder: 
`--enable-customized-log-encoder` = It will enable the feature to set customized `keys` and `formats` to the logs, otherwise, the operator will use the zap-devel config. 
`--log-message-key` = It will replace the Key of the Message field (e.g. "msg") 
`--log-level-key` = It will replace the Key of the Level field (e.g. "severity")
`--log-time-key`= It will replace the Key of the Time field (e.g. "timestamp")
`--log-time-format` = It will replace the Format of the Time field (e.g. "epoch", "epochmillis", "epochnano", "iso8601", "rfc3339" or "rfc3339nano")
`--log-level-format` = It will replace the Format of the Level field (e.g. "uppercase", "lowercase")

- Resolves: #268 

### DIFF
--- a/DEBUG.md
+++ b/DEBUG.md
@@ -1,0 +1,34 @@
+# Debug tips to the OpenTelemetry Operator
+
+A tip during a troubleshooting process is always welcome. Therefore, we prepared this documentation to help you identify possible issues and improve the application's reliability.
+
+## Customizing Logs Output
+By the default the Operator's log format is console like you can see below:
+```bash
+2024-05-06T11:55:11+02:00	INFO	setup	Prometheus CRDs are installed, adding to scheme.
+2024-05-06T11:55:11+02:00	INFO	setup	Openshift CRDs are not installed, skipping adding to scheme.
+2024-05-06T11:55:11+02:00	INFO	setup	the env var WATCH_NAMESPACE isn't set, watching all namespaces
+2024-05-06T11:55:11+02:00	INFO	Webhooks are disabled, operator is running an unsupported mode	{"ENABLE_WEBHOOKS": "false"}
+2024-05-06T11:55:11+02:00	INFO	setup	starting manager
+```
+
+If it is necessary to customize the log format, so you can use one of the following parameters:
+- `--zap-devel`:                                        Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn). Production Mode defaults(encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error) (default false)
+- `--zap-encoder`:                               Zap log encoding (one of 'json' or 'console')
+- `--zap-log-level`                              Zap Level to configure the verbosity of logging. Can be one of 'debug', 'info', 'error', or any integer value > 0 which corresponds to custom debug levels of increasing verbosity
+- `--zap-stacktrace-level`                        Zap Level at and above which stacktraces are captured (one of 'info', 'error', 'panic').
+- `--zap-time-encoding`                   Zap time encoding (one of 'epoch', 'millis', 'nano', 'iso8601', 'rfc3339' or 'rfc3339nano'). Defaults to 'epoch'.
+- The following parameters are effective only if the `--zap-encoder=json`:
+  - `zap-message-key`: The message key to be used in the customized Log Encoder
+  - `zap-level-key`: The level key to be used in the customized Log Encoder
+  - `zap-time-key`: The time key to be used in the customized Log Encoder
+  - `zap-level-format`: The level format to be used in the customized Log Encoder
+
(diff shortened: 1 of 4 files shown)


swiatekm-sumo (commented):
Some of these can be set via existing flags: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/log/zap#Options.BindFlags. If we need to add more, can we use the existing naming convention?

yuriolisa:
> Some of these can be set via existing flags: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/log/zap#Options.BindFlags. If we need to add more, can we use the existing naming convention?

I'm aware of these `BindFlags` options, but when you set a new Enconderconfig you have to populate these values as well, due to that I've put these parameters. 
The [name convention](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.31.0/specification/logs/data-model.md#field-severitynumber:~:text=Log%20and%20Event%20Record%20Definition) is a good option, however, the use case for that issue is to deliver an option to the users to customize their operator logs' fields.

swiatekm-sumo:
> > Some of these can be set via existing flags: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/log/zap#Options.BindFlags. If we need to add more, can we use the existing naming convention?
> 
> I'm aware of these `BindFlags` options, but when you set a new Enconderconfig you have to populate these values as well, due to that I've put these parameters. The [name convention](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.31.0/specification/logs/data-model.md#field-severitynumber:~:text=Log%20and%20Event%20Record%20Definition) is a good option, however, the use case for that issue is to deliver an option to the users to customize their operator logs' fields.

I think you can use `Options.EncoderConfigOptions` to modify the encoder config without overwriting it? Right now, we have a `--zap-time-encoding` which appears to do the same thing as your `--log-time-format`. We should either use the existing flag or remove it and use the new one.

main.go
```diff
@@ -163,8 +169,22 @@ func main() {
 	pflag.IntVar(&webhookPort, "webhook-port", 9443, "The port the webhook endpoint binds to.")
 	pflag.StringVar(&tlsOpt.minVersion, "tls-min-version", "VersionTLS12", "Minimum TLS version supported. Value must match version names from https://golang.org/pkg/crypto/tls/#pkg-constants.")
 	pflag.StringSliceVar(&tlsOpt.cipherSuites, "tls-cipher-suites", nil, "Comma-separated list of cipher suites for the server. Values are from tls package constants (https://golang.org/pkg/crypto/tls/#pkg-constants). If omitted, the default Go cipher suites will be used")
+	pflag.BoolVar(&custLogEncoder, "enable-customized-log-encoder", false, "Enable the use of a customized Json Log Encoder")
```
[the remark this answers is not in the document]
yuriolisa: It makes sense. I will remove it.

main.go
```diff
@@ -163,8 +168,20 @@ func main() {
 	pflag.IntVar(&webhookPort, "webhook-port", 9443, "The port the webhook endpoint binds to.")
 	pflag.StringVar(&tlsOpt.minVersion, "tls-min-version", "VersionTLS12", "Minimum TLS version supported. Value must match version names from https://golang.org/pkg/crypto/tls/#pkg-constants.")
 	pflag.StringSliceVar(&tlsOpt.cipherSuites, "tls-cipher-suites", nil, "Comma-separated list of cipher suites for the server. Values are from tls package constants (https://golang.org/pkg/crypto/tls/#pkg-constants). If omitted, the default Go cipher suites will be used")
+	pflag.StringVar(&encodeMessageKey, "zap-message-key", "message", "The message key to be used in the customized Log Encoder")
+	pflag.StringVar(&encodeLevelKey, "zap-level-key", "level", "The level key to be used in the customized Log Encoder")
+	pflag.StringVar(&encodeTimeKey, "zap-time-key", "timestamp", "The time key to be used in the customized Log Encoder")
+	pflag.StringVar(&encodeLevelFormat, "zap-level-format", "uppercase", "The level format to be used in the customized Log Encoder")
```
swiatekm-sumo, line 174: Are the defaults here the same as we currently have? That is, if we don't set any of these, there's no change in behaviour?

yuriolisa: Yes, but we can't compare it with the current log because we use the `console` encoder by default, while these changes will be effective when we set the `zap-encoder=json`.

swiatekm-sumo (approved):
LGTM, modulo the lint failure.

jaronoff97 (commented):
@yuriolisa can you also please update some docs too for this change? This would assist with the feedback we got in https://github.com/open-telemetry/opentelemetry-operator/issues/873

DEBUG.md
jaronoff97, line 1: 🚀 🎉

jaronoff97 (approved):

PR #2932, opened 2024-05-07
Allow setting of PDB for per-node

### LINKED ISSUE #2900: Target Allocator multiple Replicas with per-node allocationStrategy
### Component(s)

_No response_

### **Component(s)**
target allocator

### **Description**
The documentation says:

> Replicas is the number of pod instances for the underlying TargetAllocator. This should only be set to a value other than 1 if a strategy that allows for high availability is chosen. Currently, the only allocation strategy that can be run in a high availability mode is consistent-hashing.

Is there a reason the per-node allocation does not support multiple replicas currently?
In contrast to least-weighted, there should be no problems with contradicting allocations.

jaronoff97:
**Description:** 
Makes it possible to set PDBs for per-node strategy

resolves #2900 

**Testing:** unit tests added

**Documentation:** n/a


swiatekm-sumo (approved):

diranged:
Thanks @jaronoff97 ... I appreciate the change. I still don't actually think there should be controls on whether or not I am allowed to define a PDB, but at least this will help us solve the issue we were running into! :)

jaronoff97:
@diranged you're always able to define a PDB outside what the operator provides :) this is for the user convenience for something within the bounds of what we know will work. I'd be happy to review a follow up / discuss further during our next sig call!

iblancasa (approved):

matej-g (approved):
Anything still needed here? Looks good to me 👍

swiatekm-sumo (approved):

PR #2938, opened 2024-05-08
Cleanup cluster roles and bindings 

pavolloffay:

- Resolves: #issue-number

### DIFF
--- a/controllers/opentelemetrycollector_controller.go
+++ b/controllers/opentelemetrycollector_controller.go
@@ -35,6 +35,7 @@ import (
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"github.com/open-telemetry/opentelemetry-operator/apis/v1beta1"
 	"github.com/open-telemetry/opentelemetry-operator/internal/autodetect/openshift"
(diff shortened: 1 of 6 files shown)


controllers/reconcile_test.go
```diff
@@ -681,7 +681,7 @@ func TestOpAMPBridgeReconciler_Reconcile(t *testing.T) {
 							exists, err = populateObjectIfExists(t, &v1.Service{}, namespacedObjectName(naming.OpAMPBridgeService(params.Name), params.Namespace))
 							assert.NoError(t, err)
 							assert.True(t, exists)
-							exists, err = populateObjectIfExists(t, &v1.ServiceAccount{}, namespacedObjectName(naming.ServiceAccount(params.Name), params.Namespace))
+							exists, err = populateObjectIfExists(t, &v1.ServiceAccount{}, namespacedObjectName(naming.OpAMPBridgeServiceAccount(params.Name), params.Namespace))
```
pavolloffay, line 684: This was accidentally passing bc the SA was there from the previous test.

controllers/opentelemetrycollector_controller.go
```diff
@@ -127,7 +128,42 @@ func (r *OpenTelemetryCollectorReconciler) findOtelOwnedObjects(ctx context.Cont
...
+		}
+		for k, v := range clusterObjects {
+			ownedObjects[k] = v
+		}
+	}
+	return ownedObjects, nil
+}
 
+// The cluster scope objects do not have owner reference.
+func (r *OpenTelemetryCollectorReconciler) findClusterRoleObjects(ctx context.Context, params manifests.Params) (map[types.UID]client.Object, error) {
```
jaronoff97, line 144: Given we have to do this for a fair amount of resources now, I think it may be worth it to make this a generic method that works on [T client.Object]

controllers/reconcile_test.go
```diff
@@ -73,14 +73,14 @@ var (
 type check[T any] func(t *testing.T, params T)
 
 func TestOpenTelemetryCollectorReconciler_Reconcile(t *testing.T) {
```
jaronoff97, line 75: can you add a test in here for the rbac specific finalizer logic?

pavolloffay: Added a test

controllers/opentelemetrycollector_controller.go
pavolloffay: This needs to be a list e.g. `[]metav1.List`` but then I am not able to get the UID for the map

jaronoff97: I think you may be able to `[]v1.Object` (from the [metav1 package](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Object))

pavolloffay: Then we cannot use the `client.List` and rely on labels like other methods in this file. Instead we would need to change to `client.Get` and pass as well the `[] types.NamespacedName`.

pavolloffay: I cannot make the method more generic and look simple

jaronoff97 (approved):

Issue #2940, opened 2024-05-09
When using OpenTelemetryCollector `nodeSelector` and later removing it, collector pod doesn't clean up the nodeSelector


janario:
### Component(s)

collector

## Description

When adding OpenTelemetryCollector.spec.nodeSelector and later removing it, it doesn't get applied to the collector Pods.

## Steps to Reproduce

Scenario,
1. create an OpenTelemetryCollector with spec.nodeSelector
 a. it will create pods(from DaemonSet/Deployment/Statefulset) with the nodeSelector ✅ 
2. change OpenTelemetryCollector to remove nodeSelector
 a. it is not removed from pods(from DaemonSet/Deployment/Statefulset) 
3. change OpenTelemetryCollector to add another nodeSelector (different key/value from step 1)
 a. it is adding another nodeSelector to the pod and keeping both (step 1 and step 3)

## Expected Result

Collector Pods should reflect exactly the OpenTelemetryCollector.spec.nodeSelector

## Actual Result

Collector Pods get nodeSelectors but never removed to reflect the OpenTelemetryCollector.spec.nodeSelector.

### Kubernetes Version

1.26

### Operator version

0.98.0

### Collector version

0.98.0

## Environment
OS: (e.g., "Ubuntu 20.04")
Compiler(if manually compiled): (e.g., "go 14.2")

PR #2941, opened 2024-05-09
Fix Pod NodeSelector changes from OpenTelemetryCollector

janario:

- Resolves: https://github.com/open-telemetry/opentelemetry-operator/issues/2940

With e2e, create and update collectors (Deployment/Daemonset/Statefulset) and add/remove nodeSelector.

### DIFF
--- a/internal/manifests/mutate.go
+++ b/internal/manifests/mutate.go
@@ -178,6 +178,10 @@ func mergeWithOverride(dst, src interface{}) error {
 	return mergo.Merge(dst, src, mergo.WithOverride)
 }
 
+func mergeWithOverwriteWithEmptyValue(dst, src interface{}) error {
+	return mergo.Merge(dst, src, mergo.WithOverwriteWithEmptyValue)
+}
+
 func mutateSecret(existing, desired *corev1.Secret) {
 	existing.Labels = desired.Labels
 	existing.Annotations = desired.Annotations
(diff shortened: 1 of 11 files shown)


swiatekm-sumo:
A reconciliation test for OpampBridge is failing, I'm guessing because it expects an annotation to exist that your change removes. @jaronoff97 would know better.

janario:
> A reconciliation test for OpampBridge is failing, I'm guessing because it expects an annotation to exist that your change removes. @jaronoff97 might be able to narrow it down more.

yes, previously I was trying to merge-override-empty with the entire PodTemplate.
But a few things like labels and annotations are expected to stay even when not managed by the operator.

In the last commit, I changed to merge-override-empty only to the nodeSelector and we add more things as we find later.

internal/manifests/mutate.go
```diff
@@ -270,7 +274,9 @@ func mutateDaemonset(existing, desired *appsv1.DaemonSet) error {
 	if existing.CreationTimestamp.IsZero() {
 		existing.Spec.Selector = desired.Spec.Selector
 	}
-
+	if err := mergeWithOverwriteWithEmptyValue(&existing.Spec.Template.Spec.NodeSelector, desired.Spec.Template.Spec.NodeSelector); err != nil {
```
jaronoff97, line 277: shouldn't this step come after the following mergeWithOverride? Otherwise we would override this again in the next step?

jaronoff97 (approved):

swiatekm-sumo (approved):
This looks good to me, but we really should figure out a more holistic way of dealing with this problem.

jaronoff97:
@swiatekm-sumo agreed, though I think that's something we can consider in a follow up.

Issue #2943, opened 2024-05-09
Support for OpenTelemetryCollector `readinessProbe`


janario:
### Component(s)

collector

### Is your feature request related to a problem? Please describe.

We were testing high-availability of the collector (managed by the operator) while we roll out changes and we noticed that the Pods created by the operator don't have readinessProbe configured.

### Describe the solution you'd like

Support readinessProbe for the collector pods using the default as provided by the chart https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.yaml#L397

janario:
We should have a PR for this one by tomorrow :-)

PR #2944, opened 2024-05-10
Support for collector readinessProbe

janario:

This PR introduces a possibility to customize readinessProbe for OpenTelemetryCollector as well as adds the default just like it is done at liveness and collector chart https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.yaml#L397

**Link to tracking Issue(s):** https://github.com/open-telemetry/opentelemetry-operator/issues/2943

- Resolves: https://github.com/open-telemetry/opentelemetry-operator/issues/2943

**Testing:**

### DIFF
--- a/apis/v1beta1/collector_webhook.go
+++ b/apis/v1beta1/collector_webhook.go
@@ -289,45 +289,14 @@ func (c CollectorWebhook) validate(ctx context.Context, r *OpenTelemetryCollecto
 		return warnings, fmt.Errorf("a valid Ingress hostname has to be defined for subdomain ruleType")
 	}
 
-	if r.Spec.LivenessProbe != nil {
-		if r.Spec.LivenessProbe.InitialDelaySeconds != nil && *r.Spec.LivenessProbe.InitialDelaySeconds < 0 {
-			return warnings, fmt.Errorf("the OpenTelemetry Spec LivenessProbe InitialDelaySeconds configuration is incorrect. InitialDelaySeconds should be greater than or equal to 0")
-		}
-		if r.Spec.LivenessProbe.PeriodSeconds != nil && *r.Spec.LivenessProbe.PeriodSeconds < 1 {
-			return warnings, fmt.Errorf("the OpenTelemetry Spec LivenessProbe PeriodSeconds configuration is incorrect. PeriodSeconds should be greater than or equal to 1")
-		}
-		if r.Spec.LivenessProbe.TimeoutSeconds != nil && *r.Spec.LivenessProbe.TimeoutSeconds < 1 {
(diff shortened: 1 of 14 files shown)


swiatekm-sumo:
I wouldn't add this to `v1alpha`. That version is now deprecated, and we shouldn't add any more features to it. WDYT @pavolloffay @jaronoff97

jaronoff97:
+1 we should only add this to v1beta1.

janario:
Got it! 
Removed from alpha version.

Just one detail to see if I got it right.

Deploying alpha won't have a way to configure anything about readiness but it will enable if healthcheck extension is available.

In case the user needs to customize it, they will required to update to beta.

Right?

apis/v1beta1/collector_webhook.go
```diff
@@ -309,6 +309,26 @@ func (c CollectorWebhook) validate(ctx context.Context, r *OpenTelemetryCollecto
...
+		if r.Spec.ReadinessProbe.SuccessThreshold != nil && *r.Spec.ReadinessProbe.SuccessThreshold < 1 {
+			return warnings, fmt.Errorf("the OpenTelemetry Spec ReadinessProbe SuccessThreshold configuration is incorrect. SuccessThreshold should be greater than or equal to 1")
+		}
+		if r.Spec.ReadinessProbe.FailureThreshold != nil && *r.Spec.ReadinessProbe.FailureThreshold < 1 {
+			return warnings, fmt.Errorf("the OpenTelemetry Spec ReadinessProbe FailureThreshold configuration is incorrect. FailureThreshold should be greater than or equal to 1")
+		}
+		if r.Spec.ReadinessProbe.TerminationGracePeriodSeconds != nil && *r.Spec.ReadinessProbe.TerminationGracePeriodSeconds < 1 {
+			return warnings, fmt.Errorf("the OpenTelemetry Spec ReadinessProbe TerminationGracePeriodSeconds configuration is incorrect. TerminationGracePeriodSeconds should be greater than or equal to 1")
+		}
+	}
```
swiatekm-sumo, line 331: This check is basically identical to the one above, right? Could we factor it out to a separate function, `validateProbe` maybe?

swiatekm-sumo (approved):
LGTM, with one nitpick.

swiatekm-sumo:
> Deploying alpha won't have a way to configure anything about readiness but it will enable if healthcheck extension is available.
> 
> In case the user needs to customize it, they will required to update to beta.
> 
> Right?

Yes, that's correct. I think it's fine, especially with a change as safe as this.

apis/v1beta1/collector_webhook.go
janario: Good point.

adjusted it ✅

jaronoff97 (approved):

pavolloffay (approved):

janario:
Thank you all for the review ;-) 

Looking forward to get this and a few other improvements in the next releases 💪 ;-)

Issue #2871, opened 2024-04-17
Specify the collector configuration in an external ConfigMap


matthagenbuch:
### Component(s)

collector

### Is your feature request related to a problem? Please describe.

When changing the configuration of a collector, I have no option for a controlled rollout / rollback process in the case of an error. Once the new configuration is applied to my OpenTelemetryCollector resource, the old configuration is overwritten and all running collector pods now see the new configuration. 

This makes it difficult to control the rollout process, as any pod that restarts for any reason will load with the new configuration. It also makes it impossible to roll back to the previous configuration without again updating the OpenTelemetryCollector resource.

### Describe the solution you'd like

For other services, we "version" the configuration in ConfigMap resources. Each time a configuration is updated, a new ConfigMap with a unique name is created and the deployment/statefulset is updated to reference it. This ensures that running pods see no change in configuration, and rolling back can be done by an external tool such as Argo Rollouts since the old configuration is still on the cluster.

Ideally, I'd like to be able to specify the OpenTelemetryCollector configuration by passing in a ConfigMap, which I manage myself. This would allow me to use the same versioning / rollback system as for other services I own.

### Describe alternatives you've considered

The OpenTelemetry Operator could automatically create and manage versioned ConfigMaps for the collector configuration. This might be a useful OOTB feature for some users, but in my case I would prefer to manage ConfigMaps external to the operator in order to guarantee consistency with my other services.

### Additional context

There is a similar [open issue](https://github.com/open-telemetry/opentelemetry-operator/issues/1089) for specifying multiple configurations to be merged. The solutions proposed there which involve specifying an external ConfigMap would also resolve this feature request.

Currently, the operator does have a `configmaps` field which could be used to mount the collector configuration. However, the operator [explicitly doesn't support](https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/manifests/collector/container.go#L82) overriding the `--config` arg, so we can't use these ConfigMaps for collector configuration.

swiatekm-sumo:
[...]

PR #2946, opened 2024-05-10
Keep multiple versions of Collector Config

matthagenbuch:

From the issue:

> When changing the configuration of a collector, I have no option for a controlled rollout / rollback process in the case of an error. Once the new configuration is applied to my OpenTelemetryCollector resource, the old configuration is overwritten and all running collector pods now see the new configuration.
> 
> This makes it difficult to control the rollout process, as any pod that restarts for any reason will load with the new configuration. It also makes it impossible to roll back to the previous configuration without again updating the OpenTelemetryCollector resource.

This PR resolves the issue by keeping 1 or more previous versions of the Collector ConfigMap. The ConfigMap name is suffixed with an 8-character hash of the Config object. The full hash is truncated to 8 characters to leave room in the 63-character name limit for the collector name.

The number of config versions to keep can be specified using the new ConfigVersions field on the CRD. The operator will keep the current version + `ConfigVersions` previous versions. Cleanup of old versions is handled using the reconciliation code path.

This solution was preferred to support for externally managed ConfigMaps based on conversations at the SIG meeting. External ConfigMaps would break some functionality of the operator which requires the collector config to be known by the operator.

- Resolves: #2871

 local manual testing, unit tests, e2e tests

New configuration value on OpenTelemetryCollector CRD for setting the number of versions to keep.


apis/v1beta1/opentelemetrycollector_types.go
```diff
@@ -94,6 +94,11 @@ type OpenTelemetryCollectorSpec struct {
 	// +required
 	// +kubebuilder:pruning:PreserveUnknownFields
 	Config Config `json:"config"`
+	// ConfigVersions defines the number versions to keep for the collector config. Each config version is stored in a separate ConfigMap.
+	// Defaults to 3. The minimum value is 1.
+	// +optional
+	// +kubebuilder:default:=3
```
swiatekm-sumo, line 100: We can validate the minimum using a validation marker:
```suggestion
	// +kubebuilder:default:=3
	// +kubebuilder:validation:Minimum:=1
```

apis/v1beta1/opentelemetrycollector_types.go
```diff
@@ -94,6 +94,11 @@ type OpenTelemetryCollectorSpec struct {
...
+	ConfigVersions *int32 `json:"configVersions,omitempty"`
```
swiatekm-sumo, line 101: Does this need to be a pointer? It looks like it should always be set.

controllers/builder_test.go
```diff
@@ -86,6 +87,10 @@ service:
 	goodConfig := v1beta1.Config{}
 	err := go_yaml.Unmarshal([]byte(goodConfigYaml), &goodConfig)
 	require.NoError(t, err)
+
+	goodConfigHash, _ := manifestutils.GetConfigMapSHA(goodConfig)
+	goodConfigHash = goodConfigHash[:8]
```
swiatekm-sumo, line 92: Nice. These tests were always a pain to update because of this.

controllers/opentelemetrycollector_controller.go
```diff
@@ -128,6 +129,30 @@ func (r *OpenTelemetryCollectorReconciler) findOtelOwnedObjects(ctx context.Cont
...
+	if params.OtelCol.Spec.ConfigVersions != nil {
+		configVersionsToKeep = max(1, int(*params.OtelCol.Spec.ConfigVersions))
+	}
+
+	for i := range configMapList.Items {
+		if i > configVersionsToKeep {
+			ownedObjects[configMapList.Items[i].GetUID()] = &configMapList.Items[i]
+		}
+	}
+
```
swiatekm-sumo, line 155: Can we move this to a separate function? It's relatively self-contained and `findOtelOwnedObjects` is already a bit long.

swiatekm-sumo (commented):
Overall this looks good to me, and is pleasantly much simpler than I expected. Left some comments mostly around the CRD field.

apis/v1beta1/opentelemetrycollector_types.go
matthagenbuch: I changed it to `int`, good point

controllers/opentelemetrycollector_controller.go
matthagenbuch: done!

controllers/builder_test.go
```diff
@@ -223,13 +228,13 @@ service:
...
 					ObjectMeta: metav1.ObjectMeta{
-						Name:      "test-collector",
+						Name:      "test-collector-" + goodConfigHash,
 						Namespace: "test",
 						Labels: map[string]string{
 							"app.kubernetes.io/component":  "opentelemetry-collector",
 							"app.kubernetes.io/instance":   "test.test",
 							"app.kubernetes.io/managed-by": "opentelemetry-operator",
-							"app.kubernetes.io/name":       "test-collector",
+							"app.kubernetes.io/name":       "test-collector-" + goodConfigHash,
```
jaronoff97, line 237: The name of the object changing is fine, but I think we should keep this app name consistent with the collector intended to mount it. I.e. the guarantee that `kubectl get all -l app.kubernetes.io/name=<col-name>` is one we should aim to keep

controllers/opentelemetrycollector_controller.go
```diff
@@ -128,9 +129,39 @@ func (r *OpenTelemetryCollectorReconciler) findOtelOwnedObjects(ctx context.Cont
...
+	for i := range ownedConfigMaps {
+		ownedObjects[ownedConfigMaps[i].GetUID()] = &ownedConfigMaps[i]
+	}
+
 	return ownedObjects, nil
 }
 
+// getConfigMapsToReconcile returns a list of ConfigMaps to reconcile based on the number of ConfigMaps to keep.
+// It keeps the newest ConfigMap, the `configVersionsToKeep` next newest ConfigMaps, and returns the remainder.
+func (r *OpenTelemetryCollectorReconciler) getConfigMapsToReconcile(configVersionsToKeep int, configMapList *corev1.ConfigMapList) []corev1.ConfigMap {
```
jaronoff97, line 147:
```suggestion
func (r *OpenTelemetryCollectorReconciler) getConfigMapsToRemove(configVersionsToKeep int, configMapList *corev1.ConfigMapList) []corev1.ConfigMap {
```

I think this function name better describes what this is doing – i.e. returning any configmaps that need to be cleaned up.

controllers/reconcile_test.go
```diff
@@ -495,7 +499,9 @@ func TestOpenTelemetryCollectorReconciler_Reconcile(t *testing.T) {
 					result: controllerruntime.Result{},
 					checks: []check[v1alpha1.OpenTelemetryCollector]{
 						func(t *testing.T, params v1alpha1.OpenTelemetryCollector) {
-							exists, err := populateObjectIfExists(t, &v1.ConfigMap{}, namespacedObjectName(naming.Collector(params.Name), params.Namespace))
+							configHash, _ := getConfigMapSHAFromString(params.Spec.Config)
```
jaronoff97, line 502: can you add a test into this reconcile file to ensure we clean up the old configmaps effectively? If you set the configmap versions to keep as one, then you can check that updating ensure that guarantee.

apis/v1beta1/opentelemetrycollector_types.go
```diff
@@ -94,6 +94,12 @@ type OpenTelemetryCollectorSpec struct {
 	// +required
 	// +kubebuilder:pruning:PreserveUnknownFields
 	Config Config `json:"config"`
+	// ConfigVersions defines the number versions to keep for the collector config. Each config version is stored in a separate ConfigMap.
+	// Defaults to 3. The minimum value is 1.
+	// +optional
+	// +kubebuilder:default:=3
+	// +kubebuilder:validation:Minimum:=1
+	ConfigVersions int `json:"configVersions,omitempty"`
```
jaronoff97, line 102: @swiatekm-sumo do you think we should also add a validation for configVersions in the webhook to ensure that it's always >1 or do we trust kubebuilder enough here?

controllers/reconcile_test.go
jaronoff97: if it's too much work let's just book a follow-up item given you have the e2e test. We're trying to rely more on these tests where we can as e2e testing can be a bit cumbersome.

jaronoff97 (commented):
a few minor thoughts, otherwise this looks awesome!

apis/v1beta1/opentelemetrycollector_types.go
swiatekm-sumo: I trust OpenAPI validation, which is what it comes down to. In general, I'd like to rely more on these validations and less on the webhook, given that the latter can be disabled.

controllers/reconcile_test.go
matthagenbuch: I started working on this but it seems like I either need to create a new test for v1beta1 objects (which is easy enough but duplicates a lot of boilerplate), or convert the existing reconcile test to v1beta1 which would be too much work. WDYT?

jaronoff97: ah let's just book a followup for this then, I will probs need to do that conversion :)

matthagenbuch: https://github.com/open-telemetry/opentelemetry-operator/issues/2948

controllers/builder_test.go
matthagenbuch: fair point, i changed this

.chloggen/matth.versioned_config.yaml
```diff
@@ -0,0 +1,16 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Keep multiple previous versions of the Collector ConfigMap, configurable via the ConfigVersions field.
+
+# One or more tracking issues related to the change
+issues: [2871]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
pavolloffay, line 16: It would be great to describe why this feature is useful and how it can be used

internal/manifests/collector/configmap_test.go
```diff
@@ -17,6 +17,8 @@ package collector
 import (
 	"testing"
 
+	"github.com/open-telemetry/opentelemetry-operator/internal/manifests/manifestutils"
+	"github.com/open-telemetry/opentelemetry-operator/internal/naming"
```
pavolloffay, line 21: these imports should be moved to a separate group

swiatekm-sumo: We should update the gci config to enforce this.

.chloggen/matth.versioned_config.yaml
matthagenbuch: good idea, I added a more detailed description here

jaronoff97 (approved):

matthagenbuch:
@swiatekm-sumo @pavolloffay any other thoughts on this?

swiatekm-sumo (approved):
:rocket:

pavolloffay (approved):

PR #2964, opened 2024-05-16
Verify ServiceMonitor and PodMonitor are installed in prom cr availability check

akselleirv:
**Description:**
Currently the Prometheus CRs availability checks today that the API group `monitoring.coreos.com` is installed, and then adds the ServiceMonitor and PodMonitor to the scheme. However, I only have the ServiceMonitor CRD installed which results in the operator is unable to start. 

This change adds verification that both the expected resources are installed and not only that the API group is installed.

### DIFF
--- a/.chloggen/verify-prom-crd-resources.yaml
+++ b/.chloggen/verify-prom-crd-resources.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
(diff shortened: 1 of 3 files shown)


jaronoff97 (commented):
Great approach here! Can you add a changelog entry in too please? This can be done with `make chlog-new`. It will ask you to link an issue, but you can just link this PR # if you'd like

swiatekm-sumo (approved):
LGTM, thanks for your contribution!

pavolloffay (approved):

PR #2971, opened 2024-05-20
Make deletion logic generic

jaronoff97:
**Description:** Simplifies deletion logic by making it generic to the type

**Link to tracking Issue(s):** n/a

**Testing:** Unit testing

**Documentation:** n/a


swiatekm-sumo (approved):
Looks good to me.

Something I realized though is that I don't really understand why we need additional deletion logic for these specific resources. Is it because their existence is contingent on CR settings? I wonder if there isn't a controller-runtime facility for this.

pavolloffay:
e.g. the cluster scope resources do not use owner references

PR #2972, opened 2024-05-20
Introduce simplified parsers

jaronoff97:
**Description:** 
This is the first part of a multi-PR effort to simplify how parsing configuration blocks works. This is accomplished by massively reducing the toil of creating new parsers and limiting the state objects need to hold on to. Right now, none of these new components are being used, as that will be isolated for a follow up.

**Link to tracking Issue(s):** #2603 

**Testing:** many unit tests

**Documentation:** n/a


jaronoff97:
Screenshot 2024-05-20 at 3 30 59 PM

iblancasa (commented):
Thanks for this work. Ping me if you need help for the rest of components

jaronoff97:
@iblancasa can do :D I want to land this, then do exporters, then do processors and then migrate over to use this. You can see an example of the usage [here](https://github.com/jaronoff97/opentelemetry-operator/compare/another-try-sob...jaronoff97:opentelemetry-operator:migrate-to-use?expand=1)

internal/components/single_endpoint.go
```diff
@@ -0,0 +1,95 @@
...
+// configuration. It doesn't self-register and should be created/used directly.
+type SingleEndpointParser struct {
+	name string
+
+	svcPort *corev1.ServicePort
+}
+
+func (s *SingleEndpointParser) Ports(logger logr.Logger, config interface{}) ([]corev1.ServicePort, error) {
+	singleEndpointConfig := &SingleEndpointConfig{}
+	if err := LoadMap[*SingleEndpointConfig](config, singleEndpointConfig); err != nil {
```
swiatekm-sumo, line 65: Is it necessary to explicitly specialize this function call? I'd expect the compiler to be able to figure it out.

internal/components/multi_endpoint.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	return ports, nil
+}
+
+func (m *MultiPortReceiver) ParserType() string {
+	return ComponentType(m.name)
+}
+
+func (m *MultiPortReceiver) ParserName() string {
+	return fmt.Sprintf("__%s", m.name)
```
swiatekm-sumo, line 70: Why the underscores here?

swiatekm-sumo (commented):
Looks good to me overall, had some questions, and some comments about style.

internal/components/multi_endpoint.go
iblancasa: All the parsers had `__` as a prefix: https://github.com/open-telemetry/opentelemetry-operator/blob/59a7c258dcf3517870a72762e31e8c118b036f59/internal/manifests/collector/parser/receiver/receiver_aws-xray.go#L23

internal/components/component.go
```diff
@@ -0,0 +1,124 @@
...
+var (
+	PortNotFoundErr = errors.New("port should not be empty")
+)
+
+type PortRetriever interface {
+	GetPortNum() (int32, error)
+	GetPortNumOrDefault(logr.Logger, int32) int32
+}
+
+type PortBuilderOption func(portBuilder *corev1.ServicePort)
```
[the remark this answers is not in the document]
jaronoff97: yeah i copied this from a past implementation where the portBuilder was a struct i had defined but then i realized there was no reason for the struct.

internal/components/component.go
```diff
@@ -0,0 +1,127 @@
...
+	Ports(logger logr.Logger, config interface{}) ([]corev1.ServicePort, error)
+
+	// ParserType returns the name of this parser
+	ParserType() string
+
+	// ParserName is an internal name for the parser
+	ParserName() string
+}
+
+func LoadMap[T any](m interface{}, in T) error {
```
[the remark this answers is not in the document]
jaronoff97: what would the direct conversion look like? do you mean doing it when we parse the config?

internal/components/multi_endpoint.go
jaronoff97: yeah this is for compatibility, all of that is internal so it doesn't __really__ matter but I thought i would keep it consistent to not break other tests in the future.

internal/components/component.go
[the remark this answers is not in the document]
swiatekm-sumo: What this function is doing, in essence, is turning a map into a given hierarchical struct. It would be cleaner to either use similar generic conversion logic (which is a bunch of reflection internally) without the serialization step. [Mapstructure](https://github.com/mitchellh/mapstructure) instead of json, in essence.

internal/components/single_endpoint.go
jaronoff97: im gonna get rid of this entirely in favor of mapstructure :) way simpler.

internal/components/component.go
```diff
@@ -0,0 +1,115 @@
...
+func PortFromEndpoint(endpoint string) (int32, error) {
+	var err error
+	var port int64
+
+	r := regexp.MustCompile(":[0-9]+")
+
+	if r.MatchString(endpoint) {
+		portStr := r.FindString(endpoint)
+		cleanedPortStr := strings.Replace(portStr, ":", "", -1)
+		port, err = strconv.ParseInt(cleanedPortStr, 10, 32)
```
swiatekm-sumo, line 81: Port numbers are 16-bit integers, so this should really be `16` instead of `32`. We should also check if the value isn't negative.

In general, is there a reason not to use https://pkg.go.dev/net#SplitHostPort for parsing here, instead of the regex?

swiatekm-sumo (approved):
LGTM, minor comment on the port parsing notwithstanding.

internal/components/component.go
jaronoff97: this was copied over from the existing code, but i can swap it :)

jaronoff97: actually the problem with net.SplitHostPort is that it doesn't handle the case where it contains a scheme at the beginning (obv not a major issue) but that probably is why it wasn't used initially.

pavolloffay (approved):

PR #2982, opened 2024-05-22
Bump operator-sdk to 1.30

vasireddy99:
**Description:** PR to bump operator-sdk to 1.30


swiatekm-sumo:
Is this *actually* supposed to be synced with the K8s version @jaronoff97 @pavolloffay ? The latest operator-sdk release is [v1.34.2](https://github.com/operator-framework/operator-sdk/releases/tag/v1.34.2) so it looks like these versions were identical by accident.

pavolloffay:
The operator-sdk version is not aligned with k8s

swiatekm-sumo:
@vasireddy99 I don't think you need to do this as part of K8s 1.30 support, so it's fine to close this PR.

Issue #3007, opened 2024-06-03
Annotations Filter does not Filter Annotations


ryanohnemus:
### Component(s)

collector

## Description
When using an annotations filter with the opentelemetry-operator the annotations are not actually being filtered out in the Deployment that is created. 

## Steps to Reproduce
I setup a kind cluster using the kind-1.29.yaml included in the repo:

```
# setup kind
kind create cluster --config kind-1.29.yaml

# add required cert-manager install
kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.14.4/cert-manager.yaml

# Download the latest operator install
curl -Lo otel-op.yaml https://github.com/open-telemetry/opentelemetry-operator/releases/latest/download/opentelemetry-operator.yaml
```

I then edited the otel-op.yaml to add the `--annotations-filter="config.*.gke.io.*"` arg below for the operator's Deployment:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/name: opentelemetry-operator
    control-plane: controller-manager
  name: opentelemetry-operator-controller-manager
  namespace: opentelemetry-operator-system
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: opentelemetry-operator
      control-plane: controller-manager
  template:
    metadata:
      labels:
        app.kubernetes.io/name: opentelemetry-operator
        control-plane: controller-manager
    spec:
      containers:
      - args:
        - --metrics-addr=127.0.0.1:8080
        - --enable-leader-election
        - --zap-log-level=info
        - --zap-time-encoding=rfc3339nano
        - --enable-nginx-instrumentation=true
        - --annotations-filter="config.*.gke.io.*"
        env:
        - name: SERVICE_ACCOUNT_NAME
          valueFrom:
            fieldRef:
              fieldPath: spec.serviceAccountName
        image: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator:0.101.0
 ```
 
 I then applied operator to the cluster and confirmed the operator pod args contained the annotations-filter above: 
 `kubectl apply -f otel-op.yaml`
Finally I added an OpenTelemetryCollector resource that has an annotation configmanagement.gke.io/token I was expecting filtered out:
```
apiVersion: opentelemetry.io/v1alpha1
kind: OpenTelemetryCollector
metadata:
  name: simplest
  annotations:
[...]

PR #3008, opened 2024-06-03
Fix annotation/label filter setting

jaronoff97:
**Description:** 
Fixes a bug where annotations/label filter regexes were not being applied correctly because they were being escaped. This result in something `hello.*world` to become `hello\..*world` in a way that was opaque to the user.

- Resolves: #3007 

**Testing:** Unit testing, manual testing

**Documentation:** n/a


swiatekm-sumo (approved):

pavolloffay (approved):

yuriolisa:
Could you please set a wider test [here](https://github.com/open-telemetry/opentelemetry-operator/blob/dc38bf1ee15e0be1b2442a7fcba7716a872e6a7f/.github/workflows/e2e.yaml#L43) with a similar regex mentioned in the issue? 
```bash
config.*.gke.io.*
```

yuriolisa (approved):

ChristianCiach:
*Technically* this PR broke the annotation/label filters.

The CLI help text of the annotations-filter option reads like this:

> Annotations to filter away from propagating onto deploys. It should be a string array containing patterns, which are literal strings optionally containing a * wildcard character.

This PR changed this option from accepting glob-style wildcards to accepting literal rexexp only. In other words, this PR broke every user that tried to do something like this:

```
--annotations-filter=something*
```

These would be needed to rewritten like this to restore the documented behaviour:

```
--annotations-filter=something.*
```

Also, where this was previously a literal match:

```
--annotations-filter=app.kubernetes.io/instance
```

This would now also match annotations like `appxkubernetesyio/instance`. 

I think it's great to be able to use regex filters now, but this is still a breaking change (which is probably fine, since the operator is still v0), but the CLI help text is now wrong. You cannot use `*` as a wildcard anymore. You would now have to use `.*`. Most people may not have noticed this change, because their `*` probably happened to have a `.` in front (which was previously a literal dot, but now means "any single character").

PR #3023, opened 2024-06-10
Factor out some fields from OpenTelemetryCommonFields

swiatekm-sumo:
Remove Autoscaler, the TargetAllocator CRD doesn't need it. Move VolumeClaimTemplates to its own StatefulSet-specific struct.

This only affects the unpublished TargetAllocator CRD in practice.

The alternative this change would be to reject these fields at admission. This isn't ideal because admission webhooks can be disabled, and with the OpenTelemetryCommonFields embedding approach, there's no way to indicate which fields are supported in the autogenerated documentation.


pavolloffay (approved):

apis/v1beta1/common.go
```diff
@@ -115,10 +115,6 @@ type OpenTelemetryCommonFields struct {
 	// Replicas is the number of pod instances for the underlying replicaset. Set this if you are not using autoscaling.
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
-	// Autoscaler specifies the pod autoscaling configuration to use
-	// for the workload.
-	// +optional
-	Autoscaler *AutoscalerSpec `json:"autoscaler,omitempty"`
```
jaronoff97, line 121: Theoretically someone may want an auto scaler for the TA, no?

swiatekm-sumo: Theoretically, yes, but I don't really want to support it just based on that possibility. Right now scaling the TA only really scales the HTTP server, and that's really not a very performance-sensitive component. So I'd like to avoid it until someone actually asks.

swiatekm-sumo: Theoretically, yes, but I don't really want to support it just based on that possibility. Right now scaling the TA only really scales the HTTP server, and that's really not a very performance-sensitive component. So I'd like to avoid it until someone actually asks.

jaronoff97 (approved):

PR #3024, opened 2024-06-10
Enable Prometheus CR via the TA config file

swiatekm-sumo:
Prometheus CR was arbitrarily enabled by a command-line flag instead of the configuration file setting. Use the config file instead for consistency. This also lets us avoid adding the PrometheusCR section altogether if it's not enabled.

I've also added proper support for `Args` in the TargetAllocator CRD. This is currently a no-op, as this field cannot be set via the embedded TA struct.

### DIFF
--- a/cmd/otel-allocator/config/config.go
+++ b/cmd/otel-allocator/config/config.go
@@ -109,9 +109,10 @@ func LoadFromCLI(target *Config, flagSet *pflag.FlagSet) error {
 		return err
 	}
 
-	target.PrometheusCR.Enabled, err = getPrometheusCREnabled(flagSet)
-	if err != nil {
-		return err
+	if prometheusCREnabled, changed, flagErr := getPrometheusCREnabled(flagSet); flagErr != nil {
+		return flagErr
+	} else if changed {
+		target.PrometheusCR.Enabled = prometheusCREnabled
 	}
 
 	target.HTTPS.Enabled, err = getHttpsEnabled(flagSet)
--- a/cmd/otel-allocator/config/flags.go
+++ b/cmd/otel-allocator/config/flags.go
@@ -69,8 +69,13 @@ func getListenAddr(flagSet *pflag.FlagSet) (string, error) {
 	return flagSet.GetString(listenAddrFlagName)
 }
 
-func getPrometheusCREnabled(flagSet *pflag.FlagSet) (bool, error) {
-	return flagSet.GetBool(prometheusCREnabledFlagName)
+func getPrometheusCREnabled(flagSet *pflag.FlagSet) (value bool, changed bool, err error) {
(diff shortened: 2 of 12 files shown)


cmd/otel-allocator/config/config.go
```diff
@@ -109,9 +109,10 @@ func LoadFromCLI(target *Config, flagSet *pflag.FlagSet) error {
 		return err
 	}
 
-	target.PrometheusCR.Enabled, err = getPrometheusCREnabled(flagSet)
-	if err != nil {
-		return err
+	if prometheusCREnabled, changed, flagErr := getPrometheusCREnabled(flagSet); flagErr != nil {
```
swiatekm-sumo, line 112: Note: This needs to be fixed for other flags as well, here I limited the fix to the relevant one.

cmd/otel-allocator/config/flags.go
```diff
@@ -69,8 +69,13 @@ func getListenAddr(flagSet *pflag.FlagSet) (string, error) {
 	return flagSet.GetString(listenAddrFlagName)
 }
 
-func getPrometheusCREnabled(flagSet *pflag.FlagSet) (bool, error) {
-	return flagSet.GetBool(prometheusCREnabledFlagName)
+func getPrometheusCREnabled(flagSet *pflag.FlagSet) (value bool, changed bool, err error) {
+	if changed = flagSet.Changed(prometheusCREnabledFlagName); !changed {
+		value, err = false, nil
```
jaronoff97, line 74: shouldn't the middle value just be false here?

jaronoff97:
just two thoughts, otherwise this looks good!

cmd/otel-allocator/config/flags.go
```diff
@@ -69,8 +69,13 @@ func getListenAddr(flagSet *pflag.FlagSet) (string, error) {
 	return flagSet.GetString(listenAddrFlagName)
 }
 
-func getPrometheusCREnabled(flagSet *pflag.FlagSet) (bool, error) {
-	return flagSet.GetBool(prometheusCREnabledFlagName)
+func getPrometheusCREnabled(flagSet *pflag.FlagSet) (value bool, changed bool, err error) {
+	if changed = flagSet.Changed(prometheusCREnabledFlagName); !changed {
```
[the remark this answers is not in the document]
swiatekm-sumo: Because if it's not, we'll unconditionally replace the value from the config file with the default value from the flag, even if the flag is unset. We only want to give flags priority if they're actually set.

cmd/otel-allocator/config/flags.go
swiatekm-sumo: Yes, I could just do a plain `return` here, but imo this is more explicit. If you disagree, I can remove it, I don't feel too strongly about it.

jaronoff97 (approved):

Issue #3036, opened 2024-06-12
appArmorProfile is removed from SecurityContext


jbmolle:
## Description
When creating a Pod or a Deploy with appArmorProfile into securityContext field, the appArmorProfile is removed by the mutating webhook.
Check https://github.com/kubernetes/kubernetes/issues/125069

## Steps to Reproduce
1) Create a pod with a Yaml file containing 
```
securityContext:
  appArmorProfile:
    type: Unconfined
```
The security context is there with kubectl apply -f xxx.yaml -oyaml --dry-run=server
2)Install opentelemetry-operator
3) Run again kubectl apply -f xxx.yaml -oyaml --dry-run=server, the securityContext appArmorProfile is gone

## Expected Result
appArmorProfile should be present in securityContext

## Actual Result
appArmorProfile is not there

### Kubernetes Version

1.30.2

### Operator version

0.102.0

### Collector version

0.102.0

## Environment
OS: Ubuntu 22.04 / Ubuntu 24.04

swiatekm-sumo:
As I mentioned in https://github.com/open-telemetry/opentelemetry-operator/pull/3037#issuecomment-2163479802, this is blocked on #3041, which is blocked by bumping our required Go version to 1.22 (#2757).

Issue #2757, opened 2024-03-12
Define a policy how the go version gets updated


frzifus:
### Is your feature request related to a problem? Please describe.

Do we want to define a guideline like on the collector when to bump this version? Context => https://github.com/open-telemetry/opentelemetry-operator/pull/2747#pullrequestreview-1927974514

### Describe the solution you'd like

Maybe similar to the opentelemetry-collector? 
See: https://github.com/open-telemetry/opentelemetry-collector#compatibility

swiatekm-sumo:
The main barrier towards supporting each Go version until EOL are dependencies. In particular, as seen in #2741, prometheus-operator packages can be quite aggressive when it comes to bumping their Go version requirements. As these dependencies account for our support for prometheus Custom Resources, not tracking upstream closely can get us into situations where we don't support some new attribute, but this lack of support isn't very discoverable for users.

In spite of what I claimed during our recent SIG meeting, the degradation is graceful, and within a given version, CRs are both forwards and backwards compatible with a given operator version. If the operator doesn't know about a particular attribute, it's simply ignored during deserialization. But we (obviously) still don't support it, and don't give any indication of this to users.

@jaronoff97 @pavolloffay how about we put the prometheus-operator version in the changelog alongside the instrumentation versions? Then we'd behave roughly the same way as prometheus-operator itself.

jaronoff97:
@swiatekm-sumo i would think we would put the compat matrix next to our matrix for kubernetes support, but yes I think that's a good idea. We'll also want to add some instructions for updating this in the release. I think I'm okay being a bit behind prometheus-operator given how eager they are, but I don't want us stuck many versions behind for more than a couple of months. I don't have strong feelings here though.

pavolloffay:
I like the alignment with the collector requirements for the go version. We could give it a try and see if it would impact our use of the promethues operator dependency. 

We should also put prometheus operator into our compatibility matrix.

PR #3037, opened 2024-06-12
Add appArmorProfile to securityContext

jbmolle:
**Description:** Add appArmorProfile in securityContext definitions so it is not removed by openteletry-operator

**Link to tracking Issue(s):**  #3036

- Resolves:  #3036

**Testing:** None

**Documentation:** None

### DIFF
--- a/bundle/manifests/opentelemetry.io_opampbridges.yaml
+++ b/bundle/manifests/opentelemetry.io_opampbridges.yaml
@@ -597,6 +597,15 @@ spec:
                     required:
                     - type
                     type: object
+                  appArmorProfile:
+                    properties:
+                      localhostProfile:
+                        type: string
+                      type:
+                        type: string
+                    required:
+                    - type
+                    type: object
                   supplementalGroups:
                     items:
                       format: int64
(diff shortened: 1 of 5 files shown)


swiatekm-sumo:
We use the SecurityContext definition from the Kubernetes API. From what I can see, this was added in Kubernetes 1.30, and we haven't yet upgraded to its API libraries. The reason is that they require Go 1.22, and we still support 1.21 as per #2757.

jbmolle:
@swiatekm-sumo Thanks for the feedback, that's very helpful!
I've actually found the problem when I tried to update Cilium to version 1.15.5 which uses appArmorProfile instead of annotations.
The init containers from Cilium can't start because opentelemetry-operator modifies the pod yaml sent to kube-apiserver.
Is there a way to exclude the pod YAML from being mutated by the webhook? (With an annotation maybe?)

swiatekm-sumo:
I don't think there's a way to do that, unfortunately. It's not so much that the operator modifies the Pod, but rather deserializes the Pod data into a struct that doesn't have the appArmor field, so the field is effectively ignored and the operator isn't even aware it exists. Best you can do if you want to solve this in the short term, would be to fork the operator, update the API libraries, and build with Go 1.22.

swiatekm-sumo:
@jbmolle can we close this PR? #3036 was fixed by #3061.

jbmolle:
Yes sure

PR #3039, opened 2024-06-14
Remove collectorSelector field from TA CRD

swiatekm-sumo:
Doing this as per the decision made in https://github.com/open-telemetry/opentelemetry-operator/issues/2422#issuecomment-2154655906. We're going to use a label to manage this relationship instead.

On a separate note, I'd like to give the TargetAllocator builder a different Params struct, so we can more clearly indicate this relationship. That's a more involved change, though, and I'll make it separately.


swiatekm-sumo:
> I agree about the params change. I was thinking we would even make the CR being reconciled a generic type for the params struct i.e. `type Params[T client.Object] struct` or something like that... client.Object probably isn't right though

I don't think that's enough in this case. What I actually want is to have Params contain a TargetAllocator struct and a slice of Collectors connected to it.

Actually making the builder code more generic isn't that difficult - I'll submit the change I have very soon.

PR #3040, opened 2024-06-15
Support more common attributes in TargetAllocator CRD

swiatekm-sumo:
Add support for all attributes in `OpenTelemetryCommonFields` to TargetAllocator CRD (#2516). Note that this is a no-op for the embedded TargetAllocator, as it doesn't let the user set any of these fields. The only change there is that we now explicitly set `DNSPolicy` and `ShareProcessNamespace`.

For containers, we now support:
* `imagePullPolicy`
* `ports`
* `envFrom`
* `volumeMounts`
* `lifecycle`

For Pods, we now support:
* `initContainers`
* `additionalContainers`
* `hostNetwork`
* `shareProcessNamespace`
* `priorityClassName`
* `terminationGracePeriodSeconds`

@jaronoff97 I think with this, it makes sense to factor this attribute generation out of the collector/targetallocator specific packages, into common.


jaronoff97 (approved):
this looks great, moving more stuff to be shared is a great idea wherever we can

PR #3043, opened 2024-06-17
Add a separate Params struct for Target Allocator

swiatekm-sumo:
Make the builder definitions more generic to allow different Params structures for different builders. Use this to define a separate Params structure for the TargetAllocator.

### DIFF
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -51,8 +51,37 @@ func BuildCollector(params manifests.Params) ([]client.Object, error) {
 	builders := []manifests.Builder[manifests.Params]{
 		collector.Build,
 	}
+	var resources []client.Object
+	for _, builder := range builders {
+		objs, err := builder(params)
+		if err != nil {
+			return nil, err
+		}
+		resources = append(resources, objs...)
+	}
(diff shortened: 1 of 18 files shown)


iblancasa (approved):

controllers/common.go
```diff
@@ -60,12 +59,28 @@ func BuildCollector(params manifests.Params) ([]client.Object, error) {
 		}
 		resources = append(resources, objs...)
 	}
+	if params.OtelCol.Spec.TargetAllocator.Enabled {
+		taParams := targetallocator.Params{
+			Client:          params.Client,
+			Scheme:          params.Scheme,
+			Recorder:        params.Recorder,
+			Log:             params.Log,
+			Config:          params.Config,
```
jaronoff97, line 68: Given that these things are generic to any params, i wonder if it would be better to define a common type that always has these and the generic parameter is that type such that a TA params would just be 
```
type Params {
     manifestutils.Params
     Collector ...
     TargetAllocator ....
}
```

jaronoff97 (approved):

PR #3069, opened 2024-06-24
Migrate to use new parsers

jaronoff97:
**Description:** Enables the operator to use the new parsers

- Part of: #2603 

**Testing:** Unit tests


apis/v1beta1/config.go
```diff
@@ -133,6 +139,53 @@ type Config struct {
 	Service    Service    `json:"service" yaml:"service"`
 }
 
+// getPortsForComponentTypes gets the ports for the given ComponentType(s).
+// NOTE FOR REVIEWERS: we could also do this by introducing something in components that is aware of which retriever
```
jaronoff97, line 143: Seeking opinions here

apis/v1beta1/config.go
```diff
@@ -133,6 +139,53 @@ type Config struct {
...
+			retriever = receivers.ReceiverFor
+			cfg = c.Receivers
+		case ComponentTypeExporter:
+			retriever = exporters.ParserFor
+			cfg = c.Exporters
+		case ComponentTypeProcessor:
+			break
+		}
+		for componentName := range enabledComponents[cType] {
+			// NOTE FOR REVIEWERS:
```
jaronoff97, line 162: Seeking opinions here

internal/manifests/collector/adapters/config_to_ports.go
```diff
@@ -37,105 +25,3 @@ const (
 func (c ComponentType) String() string {
```
jaronoff97, line 25: this will be deleted in a followup!

apis/v1beta1/config.go
swiatekm-sumo: Passing in the config is fine imo, but I'd make the parsers themselves aware of their own component type in a follow up. They kind of know anyway right now, just with the double underscore prefix.

apis/v1beta1/config.go
```diff
@@ -133,6 +139,58 @@ type Config struct {
 	Service    Service    `json:"service" yaml:"service"`
 }
 
+// getPortsForComponentTypes gets the ports for the given ComponentType(s).
+// NOTE FOR REVIEWERS: we could also do this by introducing something in components that is aware of which retriever
+// is being used and pass the right config in where necessary...
+func (c *Config) getPortsForComponentTypes(logger logr.Logger, componentType ...ComponentType) ([]corev1.ServicePort, error) {
```
swiatekm-sumo, line 145: We should try to use the same terminology as the collector here. `receiver` is the component [`kind`](https://github.com/open-telemetry/opentelemetry-collector/blob/227fb823dbf33e6a62b842aef5a70e10ed1db84f/component/component.go#L88), whereas `prometheus` is the component [`type`.](https://github.com/open-telemetry/opentelemetry-collector/blob/227fb823dbf33e6a62b842aef5a70e10ed1db84f/component/config.go#L100)

apis/v1beta1/config.go
swiatekm-sumo: I guess this depends on where we want to handle finding the right component config. At this level, we *know* it exists, as we're iterating over the map. The retriever would have to check and potentially return an error. Is there any other place we need this logic in? If it's just here, I'd keep it as is.

internal/components/component.go
```diff
@@ -113,3 +116,16 @@ func ConstructServicePort(current *corev1.ServicePort, port int32) corev1.Servic
 		Protocol:    current.Protocol,
 	}
 }
+
+func GetPortsForConfig(logger logr.Logger, config map[string]interface{}, retriever ParserRetriever) ([]corev1.ServicePort, error) {
+	var ports []corev1.ServicePort
+	for componentName, componentDef := range config {
```
swiatekm-sumo, line 122: This is really the type of this component, not the name.

apis/v1beta1/config.go
jaronoff97: I think i will clear up the type/name stuff in a followup because it's a bit bigger of a change.

internal/components/component.go
jaronoff97: This is the name because it's what we're getting from the config which could be the form `name=type/nameoverride`

apis/v1beta1/config.go
```diff
@@ -24,18 +24,24 @@ import (
...
 	"gopkg.in/yaml.v3"
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/open-telemetry/opentelemetry-operator/internal/components"
+	"github.com/open-telemetry/opentelemetry-operator/internal/components/exporters"
+	"github.com/open-telemetry/opentelemetry-operator/internal/components/receivers"
 )
 
-type ComponentType int
+type ComponentKind int
```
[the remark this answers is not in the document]
jaronoff97: this was per @swiatekm-sumo comment here https://github.com/open-telemetry/opentelemetry-operator/pull/3069#discussion_r1656980967

internal/components/component.go
swiatekm-sumo: In that case, the collector term for this is the `component ID`, and we should use it as well. Doesn't necessarily have to be this PR.

swiatekm-sumo (approved):

pavolloffay (approved):

pavolloffay:
@jaronoff97 it is approved, can we merge it?

Issue #3014, opened 2024-06-05
Issue with Deploying Otel Operator and Instrumentation CR instance via Helm


ZiedChekir:
### Component(s)

Instrumentation CR
## Description

When deploying the otel operator and Instrumentation CR instance via a custom Helm chart, there is a labeling conflict. The CR is labeled as `app.kubernetes.io/managed-by=Helm`, while according to the OpenTelemetry documentation, it needs to be managed by the OpenTelemetry operator with the label `app.kubernetes.io/managed-by=opentelemetry-operator`. This conflict causes the auto-instrumentation process to fail as the operator cannot add the necessary annotations to the Instrumentation CR, which are necessary for injecting the correct image into the auto-instrumentation initContainer.

## Steps to Reproduce

Deploy the Otel operator and Instrumentation CR using a custom Helm chart.

## Expected Result

The Otel operator should manage the Instrumentation CR, adding the necessary annotations, and successfully inject the appropriate image into the auto-instrumentation initContainer.

## Actual Result
I get an error in the deployment resource saying `spec.initcontainers[0].image is a required value` which means the application looked at the instrumenation CR instance but did not find the autoinstrumenation images in the annotations since it is managed by Helm

### Kubernetes Version

1.29

### Operator version

0.98.0

### Collector version

0.98.0

jaronoff97:
[Discussions here](https://cloud-native.slack.com/archives/C03HVLM8LAH/p1717619485184129?thread_ts=1717611417.583469&cid=C03HVLM8LAH)

jaronoff97:
Thanks for the added context!! i think this is all because of how we upgrade things because [this logic](https://github.com/open-telemetry/opentelemetry-operator/blob/4cd4454d7260b7804cc0877d66ba36608889d3d4/pkg/instrumentation/upgrade/upgrade.go#L111-L159) only runs when that is set. We filter based on this label for upgrades for [the collector](https://github.com/open-telemetry/opentelemetry-operator/blame/main/pkg/collector/upgrade/upgrade.go#L45-L50) and [instrumentations](https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/instrumentation/upgrade/upgrade.go#L83-L87). This means that despite setting the right images on the operator, we never set those correctly on the instrumentation.

jaronoff97:
this will have to be breaking change, but I think that's acceptable given the current state isn't great. cc @open-telemetry/operator-approvers thoughts?

PR #3074, opened 2024-06-25
Remove unnecessary instances of app.kubernetes.io/managed-by

jaronoff97:
**Description:** 
This PR resolves a bug where we applying the `app.kubernetes.io/managed-by` to resources unnecessarily. We also relied on this label to be present which caused a nasty bug with instrumentation. Because we expect this to be on all CRDs and we use that for querying for upgrades, when we go to query for instrumentations, any that were applied by helm (which sets this label itself) would be omitted, causing them to fail to be upgraded. This change is technically breaking, but is also a bug fix because we are no longer failing the user expectation for auto instrumentation fixes.

- Resolves: #3014

**Testing:** Unit, manual


kristinapathak (approved):

apis/v1beta1/collector_webhook_test.go
```diff
@@ -111,9 +111,7 @@ func TestCollectorDefaultingWebhook(t *testing.T) {
 			otelcol: OpenTelemetryCollector{},
 			expected: OpenTelemetryCollector{
 				ObjectMeta: metav1.ObjectMeta{
-					Labels: map[string]string{
-						"app.kubernetes.io/managed-by": "opentelemetry-operator",
-					},
+					Labels: map[string]string{},
```
fyuan1316, line 114: Hi @jaronoff97 ,
Just curious, after removing the filtering logic. For resources with label "app.kubernetes.io/managed-by": "Helm", could it be possible that both the helm chart and the OTel operator are managed at the same time resulting in a common modification of a resource (e.g., operator modifies the resource first, helm upgrade modifies it again)

swiatekm-sumo: Different applications making changes to the same resource is a normal occurence in K8s, that by itself is not an issue. What we're fixing here is that the value of this label is a lie currently. The operator does not manage its own CRs in general, it only manages the ones it creates.

swiatekm-sumo (approved):
I think this is ok, but I'm not 100% confident there's no edge case when upgrading to the operator version with this change.

apis/v1beta1/collector_webhook_test.go
jaronoff97: yeah the issue today is that a collector or instrumentation created by helm will not be upgraded correctly by the operator because the label we filter for doesn't exist

jaronoff97:
@swiatekm-sumo i think if we were going the other way (beginning to filter) that would be problematic, but this should encapsulate _all_ collector CRs now, not just the ones created without helm.

jpkrohling:
Hey folks, are there plans to merge this? Someone asked me about this one today, they would benefit from this fix.

jaronoff97:
@jpkrohling i still need to do some testing on this. I want to be sure we're not going to break existing installations more than we need to. I'll see if I can prioritize that for this week, but I'm juggling a few responsibilities currently.

kennytrytek-wf:
> I want to be sure we're not going to break existing installations

@jpkrohling Release https://github.com/open-telemetry/opentelemetry-operator/releases/tag/v0.107.0 has already broken existing installations, since upgrade 105 is not being applied due to this bug. If the feature flag that was removed in the collector is present in the custom resource, then the collector pod will begin crashing due to invalid configuration. A workaround is to edit the CR to remove the offending flag, though this is tedious if you manage many collectors.

jaronoff97:
@kennytrytek-wf thanks for the report. I will make this my priority and this will block our next release. Sincere apologies for the tediousness there.

kennytrytek-wf:
@jaronoff97 Thank you! We're currently updating all our CRs. :)

pkg/collector/upgrade/upgrade.go
```diff
@@ -73,23 +68,36 @@ func (u VersionUpgrade) ManagedInstances(ctx context.Context) error {
...
+				var upgradedErr error
+				upgraded, upgradedErr = u.ManagedInstance(ctx, col)
+				if upgradedErr != nil {
+					return upgradedErr
+				}
+				if err := u.Client.Update(ctx, &upgraded); err != nil {
+					return err
+				}
+				return nil
+			})
```
jaronoff97, line 89: i found a bug in the upgrade that was preventing the 105 from being applied. The patch never applied the removal of the unify env feature gate which meant that the 107 release was causing anything that had that gate to fail rollout. When i moved to update, i started getting the good ol' object has changed bug, so i wrapped this in the retry on conflict method from [this PR](https://github.com/open-telemetry/opentelemetry-operator/pull/2383)

```
{"level":"ERROR","timestamp":"2024-08-23T20:36:50Z","logger":"collector-upgrade","message":"failed to apply changes to instance","name":"opentelemetry-kube-stack-cluster-stats","namespace":"default","error":"Operation cannot be fulfilled on opentelemetrycollectors.opentelemetry.io \"opentelemetry-kube-stack-cluster-stats\": the object has been modified; please apply your changes to the latest version and try again","stacktrace":"github.com/open-telemetry/opentelemetry-operator/pkg/collector/upgrade.VersionUpgrade.ManagedInstances\n\t/Users/jacob.aronoff/workspace/playground/opentelemetry-operator/pkg/collector/upgrade/upgrade.go:88\nmain.addDependencies.func1\n\t/Users/jacob.aronoff/workspace/playground/opentelemetry-operator/main.go:458\nsigs.k8s.io/controller-runtime/pkg/manager.RunnableFunc.Start\n\t/Users/jacob.aronoff/workspace/playground/opentelemetry-operator/vendor/sigs.k8s.io/controller-runtime/pkg/manager/manager.go:307\nsigs.k8s.io/controller-runtime/pkg/manager.(*runnableGroup).reconcile.func1\n\t/Users/jacob.aronoff/workspace/playground/opentelemetry-operator/vendor/sigs.k8s.io/controller-runtime/pkg/manager/runnable_group.go:226"}
```

If anyone can explain why patch was working to _add_ the new args but fail to _remove_ the unwanted arg I would love to know.

swiatekm (approved):

swiatekm:
At this point I think we should either split the upgrade fix into its own PR or add another changelog entry. What we're doing here is way beyond just removing the labels.

apis/v1alpha1/instrumentation_webhook.go
```diff
@@ -91,10 +91,6 @@ func (w InstrumentationWebhook) defaulter(r *Instrumentation) error {
 	if r.Labels == nil {
 		r.Labels = map[string]string{}
 	}
-	if r.Labels["app.kubernetes.io/managed-by"] == "" {
-		r.Labels["app.kubernetes.io/managed-by"] = "opentelemetry-operator"
```
pavolloffay, line 95: We should document this, perhaps in a different changelog entry?

frzifus (approved):

iblancasa (approved):

kennytrytek-wf:
Thank you for the quick turnaround on getting this merged and released!

jaronoff97:
no problem sorry it took so long. Testing this was really difficult 😭

Issue #2658, opened 2024-02-23
Provide Access to Pod DNSConfig


forestsword:
### Component(s)

collector

### Is your feature request related to a problem? Please describe.

I need to be able to add the DNSConfig to the collector pods by way of the Operator. Today this is not exposed.

### Describe the solution you'd like

That [`DNSConfig`](https://pkg.go.dev/k8s.io/api@v0.29.2/core/v1#PodDNSConfig) is exposed as a part of the `OpenTelemetryCollector` API.

### Describe alternatives you've considered

I don't know of another way aside from a custom mutating webhook.

forestsword:
As a late follow up, it would be also be helpful to have [`DNSPolicy`](https://pkg.go.dev/k8s.io/api@v0.29.2/core/v1#DNSPolicy) configurable.

pavolloffay:
@forestsword could you please describe the use-case you are trying to solve?

forestsword:
Sure, I'm following some internal best practices for our Kubernetes clusters where we use dnsmasq to cache and prevent overloading coredns.

The policy needs to be set to `None` and a custom dns configuration provided to use a local dnsmasq instead.

jaronoff97:
@forestsword will [this resolve](https://github.com/open-telemetry/opentelemetry-operator/pull/3077/files#diff-34db83fc5c513943d88ada8c831ee83c992c8250ff9e791d57ac985f73623de4) your issue?

PR #3077, opened 2024-06-26
Adding PodDNSConfig

yuriolisa:

Enabling PodDnsConfig for OpenTelemetry Collector, TargetAllocator and OpAMPBridge.

- Resolves: #2658 


swiatekm-sumo (approved):
LGTM

jaronoff97 (approved):

apis/v1alpha1/opampbridge_types.go
```diff
@@ -105,6 +105,8 @@ type OpAMPBridgeSpec struct {
 	// https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/
 	// +optional
 	TopologySpreadConstraints []v1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+	// PodDNSConfig defines the DNS parameters of a pod in addition to those generated from DNSPolicy.
```
jaronoff97, line 108: @yuriolisa should we also make it possible to configure DNSPolicy?

yuriolisa: @jaronoff97, actually not, because all the three possible values are contemplated [here](https://github.com/open-telemetry/opentelemetry-operator/pull/3077/files#diff-34db83fc5c513943d88ada8c831ee83c992c8250ff9e791d57ac985f73623de4) following the precedence.

yuriolisa: Just adding that I prefer this approach because it helps the end user avoid configuration mistakes.

jaronoff97: is there any other case where the user would want to set the DNSPolicy separately?

yuriolisa: IMO, no. It's just to offer a safe place to deliver the combination of HostNetwork, DNSPolicy, and DNSConfig.

internal/manifests/manifestutils/dns.go
```diff
@@ -15,14 +15,22 @@
...
 // Get the Pod DNS Policy depending on whether we're using a host network.
-func GetDNSPolicy(hostNetwork bool) corev1.DNSPolicy {
+func GetDNSPolicy(hostNetwork bool, dnsConfig corev1.PodDNSConfig) corev1.DNSPolicy {
 	dnsPolicy := corev1.DNSClusterFirst
 	if hostNetwork {
 		dnsPolicy = corev1.DNSClusterFirstWithHostNet
 	}
+	// If local DNS configuration is set, takes precedence of hostNetwork.
+	if dnsConfig.Nameservers != nil {
+		dnsPolicy = corev1.DNSNone
```
jaronoff97, line 33: it's also possible that someone would want to use nameservers but not have the DNSPolicy set to be None https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-dns-config though yeah this probably doesn't make a ton of sense... let's do this for now and if someone really wants the customization we can cross that bridge when we get to it.

jaronoff97 (approved):

Issue #2985, opened 2024-05-23
Bridge should pull from v1beta1 too


jaronoff97:
### Component(s)

opamp

### Is your feature request related to a problem? Please describe.

The OpAMP Bridge currently only queries for v1alpha1 collectors, we should also check for v1beta1 now too!

### Describe the solution you'd like

Modify the code to query for v1beta1 and v1alpha1

yuriolisa:
@jaronoff97, I can get this.

yuriolisa:
> @yuriolisa my colleague has been working on this and should have a PR soon for it actually :)

Great!! Thanks for the heads-up!

jaronoff97:
@gdfast is working on it :)

PR #3080, opened 2024-06-27
OpAMP Bridge: Applier support for OTel CRD v1beta1 API Version

gdfast:
**Description:** Updates the opamp bridge's config applier interface and client to support opentelemetry.io/v1beta1, explicitly removing support for applying OpenTelemetryCollector configurations of the opentelemetry.io/v1alph1 version

**Link to tracking Issue(s):** [2985](https://github.com/open-telemetry/opentelemetry-operator/issues/2985)

- Resolves: https://github.com/open-telemetry/opentelemetry-operator/issues/2985

**Testing:** Updated existing client and agent unit tests; functionally tested the bridge by running it and making sure it didn't hit any errors at steady state. Did the same after launching collectors with v1beta1 and v1alpha1 specs.

**Documentation:** n/a


cmd/operator-opamp-bridge/operator/client.go
```diff
@@ -76,82 +76,134 @@ func NewClient(name string, log logr.Logger, c client.Client, componentsAllowed
...
+	c.log.Info("Received new config", "name", name, "namespace", namespace)
+
+	if len(configmap.Body) == 0 {
+		return errors.NewBadRequest("invalid config to apply: config is empty")
+	}
+
+	var collector v1beta1.OpenTelemetryCollector
+	err := yaml.Unmarshal(configmap.Body, &collector)
+	if err != nil {
+		return errors.NewBadRequest(fmt.Sprintf("failed to unmarshal config into v1beta1 API Version: %v", err))
```
jaronoff97, line 89: nit: we should just use fmt.Errorf here instead

cmd/operator-opamp-bridge/operator/client.go
```diff
@@ -165,36 +217,51 @@ func (c Client) Delete(name string, namespace string) error {
...
 
-func (c Client) GetInstance(name string, namespace string) (*v1alpha1.OpenTelemetryCollector, error) {
+func (c Client) listCollectors(ctx context.Context, opts ...client.ListOption) (v1beta1.OpenTelemetryCollectorList, error) {
+	collectorsList := v1beta1.OpenTelemetryCollectorList{}
+	err := c.k8sClient.List(ctx, &collectorsList, opts...)
+	if err != nil {
+		return v1beta1.OpenTelemetryCollectorList{}, err
+	}
+	return collectorsList, nil
+}
```
jaronoff97, line 259: thought: i'm not sure we're getting much value from this as a helper and I think we should just stick with using the List directly given it is p much the same LOC.

jaronoff97 (commented):
two minor thoughts, certainly nothing blocking. Overall this looks great though! Pretty clean implementation. Were you able to confirm the behavior of the bridge when running in a cluster that has both v1alpha1 and v1beta1 collectors? We don't have any e2e tests for this today, but it would be great to confirm this manually.

cmd/operator-opamp-bridge/operator/client.go
gdfast: I liked what this did for the readability of the main `ListInstances` code to have separate, one-line assignments for the lists of managed collectors and reporting collectors. You're right though that it's a small change that's not saving many LOC, I'll re-work it.

cmd/operator-opamp-bridge/operator/client.go
gdfast: Can do. Can you tell me why though? I thought we wanted to return a `errors.NewBadRequest` if the user's input (in this case the config) isn't valid based on our (updated) rules

jaronoff97: it's usually when i see something that's just making an error that's a Sprintf it's a golangci lint rule to just use fmt.Errorf. You are right though that NewBadRequest is closer to what Kubernetes would actually return here, so let's just stick with it! :)

jaronoff97 (approved):

swiatekm (approved):

gdfast:
> two minor thoughts, certainly nothing blocking. Overall this looks great though! Pretty clean implementation. Were you able to confirm the behavior of the bridge when running in a cluster that has both v1alpha1 and v1beta1 collectors? We don't have any e2e tests for this today, but it would be great to confirm this manually.

Just wanted to note that I talked live with Jacob and he guided me through some manual, functional testing. Jacob is also adding e2e tests that will increase coverage.

jaronoff97 (approved):

Issue #3094, opened 2024-07-02
OpAMP Bridge e2e testing


jaronoff97:
### Component(s)

opamp

### Is your feature request related to a problem? Please describe.

It would be great if we could run e2e tests for the OpAMP bridge to verify that it can run effectively

### Describe the solution you'd like

introduce a new image for a test opamp server that can be queried against to verify that the opamp connection was successful.

PR #3095, opened 2024-07-02
Add new image for bridge server e2e tests

jaronoff97:
**Description:** 
Creates and builds an e2e test image for an OpAMP server with endpoints available for e2e testing

- Resolves: #3094

**Testing:** This is testing

**Documentation:** n/a


.github/workflows/publish-autoinstrumentation-e2e-images.yaml
jaronoff97, line 1: this was renamed!

tests/test-e2e-apps/bridge-server/data/agent.go
jaronoff97, line 1: much of this code was copied from [here](https://github.com/open-telemetry/opamp-go/blob/ed38d5f4bf930b57e04581919fbfb676aaa5a5af/internal/examples/server/)

jaronoff97:
a follow up will use the published test image!

pavolloffay (approved):

swiatekm (approved):
This is a fair amount of code, so I think we should set up some basic tooling for it. We should run golangci-lint on it, for one. I'm ok with that happening in a follow-up though.

Issue #3042, opened 2024-06-17
Is it possible to deploy opentelemetry-operator without clusterrole, using role instead


chunglun:
### Describe the issue you're reporting

Is it possible to deploy opentelemetry-operator without clusterrole, using role instead?
My company doesn't want to grant me cluster admin privilege and ask me to use role and rolebing only to deploy this operator; however, first error happened 

go@v0.29.3/tools/cache/reflector.go:229: failed to list *v1.PersistentVolume: persistentvolumes is forbidden: User "system:serviceaccount:..." cannot list resource "persistentvolumes" in API group "" at the cluster scope

How could I solve it.....

swiatekm-sumo:
You mean for the operator itself? You should be able to, but as a Role only covers RBAC in a specific namespace, the operator deployed this way will only be able to affect resource in this namespace. You'll want to set the WATCH_NAMESPACE env variable to that namespace as well.

pavolloffay:
Let us know if that worked. I have seen this request before and it would be great to document it.

chunglun:
In fact, I moved most of the setting in ClusterRole to Role but keep the get/list/watch persistentVolume in ClusterRole.

Operator was successfully deployed and Center Collector was successfully deployed too.

Currently I’m solving the problem of auto instrumentation due to Instrument deployed but no pods generates. Not sure what makes auto instrumentation not work, still work on it.

iblancasa:
> Currently I’m solving the problem of auto instrumentation due to Instrument deployed but no pods generated. Not sure what makes auto instrumentation not work, still work on it.

The `Instrumentation` CR doesn't deploy any pod. It creates an `InitContainer` in those pods where you annotate to inject the autoinstrumentation.

chunglun:
> > Currently I’m solving the problem of auto instrumentation due to Instrument deployed but no pods generated. Not sure what makes auto instrumentation not work, still work on it.
> 
> The `Instrumentation` CR doesn't deploy any pod. It creates an `InitContainer` in those pods where you annotate to inject the autoinstrumentation.

!!! I will try again tomorrow to see if there is an InitContainer. Thanks so much for this helpful information.

alita1991:
Hi, I have the same scenario to cover and encountered the same issues when I tried to bootstrap some collectors via the operator. Probably I need to create a new role with enough access or I miss something else that I'm not aware of. Right now, the operator is configured to watch only a specific namespace, where the operator is deployed.

[...]

PR #3097, opened 2024-07-03
Don't unnecessarily take ownership of PersistentVolumes and PersistentVolumeClaims

swiatekm:
**Description:**
Fixed a bug where we'd unnecessarily watch PVs and PVCs without ever creating any. For mysterious reasons, this caused issues when running the operator scoped to a single namespace, using a Role for RBAC. See #3042.

**Link to tracking Issue(s):** #3042

### DIFF
--- a/.chloggen/fix_dont-own-volumes.yaml
+++ b/.chloggen/fix_dont-own-volumes.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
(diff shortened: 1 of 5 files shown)


jaronoff97 (approved):

iblancasa (approved):

iblancasa (changes requested):
This was added here https://github.com/open-telemetry/opentelemetry-operator/pull/2575

Because of this: https://github.com/open-telemetry/opentelemetry-operator/issues/2587

Are we sure we can remove this without any implications?

swiatekm:
@yuriolisa can you have a look as the author of #2575? It doesn't look like we're actually deleting PVCs left after deleting a StatefulSet, nor do I think we should be.

yuriolisa (approved):
LGTM

PR #3101, opened 2024-07-03
Add TargetAllocator CR webhook

swiatekm:
**Description:**
Implement the TargetAllocator webhook. I've moved some functions to seperate modules to be able to reuse them between the added webhook and the existing collector webhook. Maybe we can have a common validator for OpenTelemetryCommonFields in general?

The webhook is not yet enabled in the manifests, so this change has no user-facing effect.

**Link to tracking Issue(s):** #2516


apis/v1beta1/targetallocator_rbac.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/open-telemetry/opentelemetry-operator/internal/rbac"
+)
+
+var (
+
+	// targetAllocatorCRPolicyRules are the policy rules required for the CR functionality.
+	targetAllocatorCRPolicyRules = []*rbacv1.PolicyRule{
+		{
+			APIGroups: []string{"monitoring.coreos.com"},
+			Resources: []string{"servicemonitors", "podmonitors"},
```
yuriolisa, line 32: I'm adding [here](https://github.com/open-telemetry/opentelemetry-operator/pull/3103/files)  the Probe/ScrapeConfig support. I believe you can add both resources on CRPolicyRules.

pavolloffay (approved):

PR #3105, opened 2024-07-05
fix javaagent version

tormodatt:
**Description:** I noticed that the javaagent was running an old version.

- Resolves: #issue-number

### DIFF
--- a/versions.txt
+++ b/versions.txt
@@ -15,7 +15,7 @@ operator-opamp-bridge=0.103.0
 
 # Represents the current release of Java instrumentation.
 # Should match autoinstrumentation/java/version.txt
-autoinstrumentation-java=1.32.1
+autoinstrumentation-java=2.5.0
 
 # Represents the current release of NodeJS instrumentation.
 # Should match value in autoinstrumentation/nodejs/package.json


versions.txt
```diff
@@ -15,7 +15,7 @@ operator-opamp-bridge=0.103.0
 
 # Represents the current release of Java instrumentation.
 # Should match autoinstrumentation/java/version.txt
```
tormodatt, line 17: it is already 2.5.0 here: https://github.com/open-telemetry/opentelemetry-operator/blob/main/autoinstrumentation/java/version.txt

pavolloffay: We are intentionally pinning 1.x due to breaking change on the semantic conventions

pavolloffay (changes requested):

versions.txt
tormodatt: right. is there any info about this and the plan for upgrade somewhere?

the comment should probably be updated in any case?

pavolloffay: It should be documented in release.yaml. 

Users can override the image in the instrumentation CR

tormodatt: aha, yes I can see that there is a comment about it in RELEASE.md. might help others if it were a bit easier to find though.

should I just close this PR then?

TylerHelmuth: There is also this pinned issue: https://github.com/open-telemetry/opentelemetry-operator/issues/2542

TylerHelmuth: @tormodatt closing is appropriate as we are not ready to accept making 2.x the default image.

PR #3106, opened 2024-07-05
Allow custom annotations on service accounts

pavolloffay:

Allow custom annotations on service accounts. OpenShift 4.16 adds an annotation (`openshift.io/internal-registry-pull-secret-ref: simplest-collector-dockercfg-jwq66`) to SA which then configures pull secret. 

See https://docs.openshift.com/container-platform/4.16/release_notes/ocp-4-16-release-notes.html   (`Legacy service account API token secrets are no longer generated for each service account`)

- Resolves: #issue-number

### DIFF
--- a/controllers/reconcile_test.go
+++ b/controllers/reconcile_test.go
@@ -96,6 +96,9 @@ func TestOpenTelemetryCollectorReconciler_Reconcile(t *testing.T) {
 			},
 		},
 	}
+	deploymentExtraPorts.Annotations = map[string]string{
+		"new-annotation": "new-value",
+	}
 	ingressParams := testCollectorAssertNoErr(t, "test-ingress", "", testFileIngress)
 	ingressParams.Spec.Ingress.Type = "ingress"
 	updatedIngressParams := testCollectorAssertNoErr(t, "test-ingress", "", testFileIngress)
@@ -209,7 +212,7 @@ func TestOpenTelemetryCollectorReconciler_Reconcile(t *testing.T) {
 							exists, err = populateObjectIfExists(t, sa, namespacedObjectName(naming.ServiceAccount(params.Name), params.Namespace))
 							assert.NoError(t, err)
 							assert.True(t, exists)
-							assert.Equal(t, map[string]string{annotationName: "true", "user-defined-annotation": "value"}, sa.Annotations)
+							assert.Equal(t, map[string]string{annotationName: "true", "user-defined-annotation": "value", "new-annotation": "new-value"}, sa.Annotations)
 						},
 					},
 					wantErr:     assert.NoError,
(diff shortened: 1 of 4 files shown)


rubenvp8510 (approved):

internal/manifests/mutate.go
```diff
@@ -194,7 +194,6 @@ func mutateConfigMap(existing, desired *corev1.ConfigMap) {
 }
 
 func mutateServiceAccount(existing, desired *corev1.ServiceAccount) {
-	existing.Annotations = desired.Annotations
```
jaronoff97, line 197: doesn't this mean that if a user were to add an annotation that is expected to propagate to their serviceaccount, we would no longer do this? Also, if we remove an annotation it will no longer be removed? should we do some type of merge here?

pavolloffay: > doesn't this mean that if a user were to add an annotation that is expected to propagate to their serviceaccount, we would no longer do this? 

This PR actually makes sure that the user-added annotations on SA are not reverted by the operator. The mutate.go by default keeps user-defined annotations but for some kinds those annotations are removed (e.g. SA) and this PR relaxes that for SA.

jaronoff97: yes, but what about annotations added to the desired SA? i.e. the current logic allows us to update the existing SA with the desired new annotations

pavolloffay: The logic in this PR makes sure that the desired annotations (set by the operator) are enforced and the user-defined annotations are not removed.

jaronoff97: can you add a test for the serviceaccount annotation updating to verify that in [here](https://github.com/open-telemetry/opentelemetry-operator/blob/main/controllers/reconcile_test.go#L77)?

pavolloffay: I have added a test

controllers/reconcile_test.go
```diff
@@ -198,6 +204,12 @@ func TestOpenTelemetryCollectorReconciler_Reconcile(t *testing.T) {
 								"app.kubernetes.io/managed-by": "opentelemetry-operator",
 								"app.kubernetes.io/part-of":    "opentelemetry",
 							})
+
+							sa := &v1.ServiceAccount{}
```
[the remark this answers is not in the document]
pavolloffay: I think the test already covers what was changed in the PR.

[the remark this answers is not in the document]
jaronoff97: not entirely – my worry is that on updates we will no longer be able to add _new_ annotations to service accounts because we no longer take the desired SA annotations and put them on the existing SA. this is good for external systems that are adding annotations as they will not be overridden, but would result in an odd user experience for ourselves where users can not add new annotations to their existing SA via the otelcol CR.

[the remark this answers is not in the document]
pavolloffay: https://github.com/open-telemetry/opentelemetry-operator/blob/6c923e39f36ae03cc8e10ce44a6d26c8eb32f530/internal/manifests/mutate.go#L66 enforces that desired annotations take precedence. 

The single line this PR removes ensures that user-defined annotations are added as well.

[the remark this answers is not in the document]
pavolloffay: I have added an extra annotation to `deploymentExtraPorts` and it is propagated to the SA after the update

Issue #3126, opened 2024-07-14
Use 0.0.0.0 as otlp receiver default address


frzifus:
### Component(s)

collector

### Is your feature request related to a problem? Please describe.

The 0.104.0 collector release enables `component.UseLocalHostAsDefaultHost` by default. Since this would be a breaking change for the users, the operator in 0.104.0 reverts this change by updating fields like `:4317` or empty fields to `0.0.0.0:4317`.

### Describe the solution you'd like

Maybe it does make sense to use the collector default `localhost` in sidecar scenarios and the `$POD_IP` for all the other modes. 

---

### Update
Add upgrade routine and default webhook entry for the following components(source https://github.com/open-telemetry/opentelemetry-collector/issues/8510#issuecomment-1735292387):
- [x] [otlp receiver]()
 - [x]    [healtcheck extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/39d1e5398ff9f20681adb1c46f667b8f61c74143/extension/healthcheckextension/factory.go#L19)
---
- [ ] [awsproxy extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/39d1e5398ff9f20681adb1c46f667b8f61c74143/extension/awsproxy/factory.go#L19)
- [ ]     [jmx receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/39d1e5398ff9f20681adb1c46f667b8f61c74143/receiver/jmxreceiver/factory.go#L19)
 - [ ]    [loki receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/39d1e5398ff9f20681adb1c46f667b8f61c74143/receiver/lokireceiver/factory.go#L23)
 - [ ]    [jaeger receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/39d1e5398ff9f20681adb1c46f667b8f61c74143/receiver/jaegerreceiver/factory.go#L33)
 - [ ]    [zipkin receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/39d1e5398ff9f20681adb1c46f667b8f61c74143/receiver/zipkinreceiver/factory.go#L20)
 - [ ]    [awsxray receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/39d1e5398ff9f20681adb1c46f667b8f61c74143/receiver/awsxrayreceiver/factory.go#L35)
[...]

PR #3147, opened 2024-07-18
Use 0.0.0.0 as otlp receiver default address

frzifus:
* Adds a feature gate `operator.collector.UseLocalHostAsDefaultHost` that, when enabled, binds unspecified receiver listener addresses to all interfaces
* Contains impl. for OTLP Receiver

- Belongs to: https://github.com/open-telemetry/opentelemetry-operator/issues/3126

### DIFF
--- a/apis/v1beta1/collector_webhook.go
+++ b/apis/v1beta1/collector_webhook.go
@@ -30,6 +30,7 @@ import (
 	"github.com/open-telemetry/opentelemetry-operator/internal/config"
 	ta "github.com/open-telemetry/opentelemetry-operator/internal/manifests/targetallocator/adapters"
 	"github.com/open-telemetry/opentelemetry-operator/internal/rbac"
+	"github.com/open-telemetry/opentelemetry-operator/pkg/featuregate"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


frzifus:
Something like this @jaronoff97.

apis/v1beta1/collector_webhook.go
```diff
@@ -503,3 +510,59 @@ func ComponentUseLocalHostAsDefaultHost(otelcol *OpenTelemetryCollector) {
...
+// Unset endpoint entries are extended by the listener address 0.0.0.0.
+// Affected components are:
+// * OTLP Receiver.
+func DoNotUseLocalHostAsDefaultHost(otelcol *OpenTelemetryCollector) error {
+	return defaultOTLPAddress(otelcol)
+}
+
+// defaultOTLPAddress binds configured otlp receivers to 0.0.0.0 if nothing else
+// is explicitly defined.
+func defaultOTLPAddress(otelcol *OpenTelemetryCollector) error {
```
jaronoff97, line 524: this is kinda similar to the logic i have for the parsers, im wondering if we could reuse that in the defaulting webhook? like we could have the parsers return the default config and set that in the default webhook?

frzifus: I think that makes sense. In case we agree on the general idea, I will spend some time to build it into that. :smiley:

jaronoff97: yeah i think the general idea makes sense to me, i think doing it in the parsers with the introduction of a new method should be pretty clean

frzifus: @jaronoff97 you may want to have a look on the part changed in [internal/components](https://github.com/open-telemetry/opentelemetry-operator/pull/3147/files#diff-67ec9fd2d57bc0210df6e8f71d0970da4c107134a88f4ae26338d025a2d4265b).

Its just a draft. What I like is that the component parser becomes a bit more flexible. What I do not like is that `components.WithPortMapping` has no effect on the default. From my point of view that could lead to confusion.

On the other hand I do not want to glue some magic defaulting into the `ComponentPortParser` which only has the intention to parse ports.

Any recommendations?

Issue #3149, opened 2024-07-18
Create a must-gather for the Operator


iblancasa:
### Is your feature request related to a problem? Please describe.

Some operators have this `must-gather` tool to collect information about the components deployed by one operator.

For instance: https://github.com/sustainable-computing-io/kepler-operator/tree/v1alpha1/must-gather

The idea behind this issue is to implement one for the OpenTelemetry operator.

### Describe the solution you'd like

.

jaronoff97:
@iblancasa can you link some more documentation about this?

iblancasa:
@jaronoff97 I added some extra info to the first message.

PR #3150, opened 2024-07-18
Create a must-gather for the operator

iblancasa:
**Description:**  adds `must-gather` tool for troubleshooting.

- Resolves: #3149

I created it using Golang so we can add it directly to the operator container image.


.chloggen/3149-add-must-gather.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Add a must gather application to help troubleshoot"
```
pavolloffay, line 8: s/application/utility

.chloggen/3149-add-must-gather.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+# One or more tracking issues related to the change
+issues: [3149]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: "The new application was added to the operator container so it can be run without adding any new container. It can be run from the /must-gather path"
```
pavolloffay, line 16: The must gather utility was added to the operator container image.

Could we explain here how the utility should be run? E.g. can user run it from the host and use the default kube config?

cmd/gather/main.go
pavolloffay, line 1: Could we add a readme to this directory with some links to must gather docs and how to use the utility?

Dockerfile
```diff
@@ -15,6 +15,7 @@ COPY --from=certificates /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-ce
 
 # Copy binary built on the host
 COPY bin/manager_${TARGETARCH} manager
+COPY bin/must-gather_${TARGETARCH} must-gather
```
swiatekm, line 18: Does this need to be a part of the operator image? I'd prefer not to make this image larger for all users.

cmd/gather/README.md
```diff
@@ -0,0 +1,29 @@
+# OpenTelemetry Operator Must-Gather
```
swiatekm, line 1: In general, is this utility intended to only be used on OpenShift? The issue makes it sound like it's more generic, but this README specifically restricts it to OpenShift.

iblancasa: You can use it without OpenShift:
> Note that you can use this utility too to gather information about the objects deployed by the OpenTelemetry Operator if you don't use OpenShift.

Dockerfile
iblancasa: It is a good practice and can help with debuggability. But I'm ok with removing it or adding something to include the binary in a conditional way.

swiatekm: How big is it?

cmd/gather/README.md
swiatekm: That should be closer to the top of the document in my opinion. If you only read the opening paragraph, you walk away with the impression that it's only for OpenShift, and won't read to the end to learn otherwise.

Dockerfile
iblancasa:
```
.rwxr-xr-x@ 39M iblancasa 27 Aug 15:54 bin/must-gather_amd64
```

iblancasa: I can add a new container or, maybe, add something to add the binary only if you want. Like a `makefile` variable or something like that.

jaronoff97: yeah i think if we can indeed build this as a separate image that would be better. Especially given not all users will want/need this functionality. We also can modify the openshift bundle to automatically use this if we'd like too.

cmd/gather/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## What is a Must-Gather?
+
+The `must-gather` tool is a utility that collects logs, cluster information, and resource configurations related to a specific operator or application in an OpenShift cluster. It helps cluster administrators and developers diagnose issues by providing a snapshot of the cluster's state related to the targeted component. More information [in the official documentation](https://docs.openshift.com/container-platform/4.16/support/gathering-cluster-data.html).
+
+## Usage
+
+First, you will need to build and push the image:
+```sh
+make container-must-gather container-must-gather-push
```
swiatekm, line 15: For the record, I do think we should publish this image (and probably version it the same way we do the operator?) and not force users to build it themselves.

iblancasa: Ok, I can do. Just wanted to avoid maintaining another image.

swiatekm: Well, how much maintenence the image actually needs depends on how elaborate we make the release process. I think it's simplest to use the same versioning as we do for the operator, but I can also see building it on each commit if the files have changed, and just tagging it with the commit hash. Realistically, users will always want the latest one after all.

iblancasa: We can do it per release. I just applied the changes. Thanks for the feedback!

.chloggen/3149-add-must-gather.yaml
```diff
@@ -0,0 +1,16 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Add a must gather utility to help troubleshoot"
+
+# One or more tracking issues related to the change
+issues: [3149]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: "The new utility was added to the operator container so it can be run without adding any new container. It can be run from the /must-gather path"
```
pavolloffay, line 16: I am missing something. I don't see that the must gather is part of the operator docker image.

.github/workflows/publish-must-gather.yaml
```diff
@@ -0,0 +1,92 @@
...
+            echo "Building must-gather for $arch"
+            make must-gather ARCH=$arch
+          done
+
+      - name: Docker meta
+        id: docker_meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            otel/opentelemetry-operator-must-gather
```
pavolloffay, line 45: Why are we publishing the image if the binary should be part of the operator container (written in release notes)?

.chloggen/3149-add-must-gather.yaml
jaronoff97: we moved it out of the operator image, i think we need to update this too.

.github/workflows/publish-must-gather.yaml
swiatekm: Release notes are wrong, I asked for it to be removed from the manager container. If you disagree we can reopen that discussion - my argument was that it made the image 40% larger, and I don't think that's justifiable for a debugging tool.

iblancasa: Release notes fixed

swiatekm (commented):
I'm fine with this PR as is, and this kind of tool doesn't really need extensive testing. But I would like to have some assurance that it actually works on a basic level. Is there some kind of E2E test we could do to verify that?

iblancasa:
> I'm fine with this PR as is, and this kind of tool doesn't really need extensive testing. But I would like to have some assurance that it actually works on a basic level. Is there some kind of E2E test we could do to verify that?

I added some unit tests and I think that should be enough.

swiatekm (approved):
:+1:

.github/workflows/publish-must-gather.yaml
```diff
@@ -0,0 +1,92 @@
+name: "Publish Python Auto-Instrumentation"
```
pavolloffay, line 1: this should be changed

.chloggen/3149-add-must-gather.yaml
```diff
@@ -0,0 +1,16 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Add a must gather utility to help troubleshoot"
+
+# One or more tracking issues related to the change
+issues: [3149]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: "The new utility is available as part of a new container image"
```
pavolloffay, line 16: I would like us to provide more info  in the release notes. Can you provide here the docker image name as well? And perhaps how to use it?

.chloggen/3149-add-must-gather.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: operator
```
pavolloffay, line 5: there is no operator component. Which CRs can be debugged with the must gather?

.github/workflows/publish-must-gather.yaml
```diff
@@ -0,0 +1,92 @@
...
+            arch=${platform#*/}
+            echo "Building must-gather for $arch"
+            make must-gather ARCH=$arch
+          done
+
+      - name: Docker meta
+        id: docker_meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
```
pavolloffay, line 44: I think we should remove `otel/opentelemetry-operator-must-gather` and publish only to ghcr like we do for other images

.github/workflows/publish-must-gather.yaml
```diff
@@ -0,0 +1,92 @@
...
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version: '~1.22.4'
+
+      - name: Unshallow
+        run: git fetch --prune --unshallow
+
+      - name: Set env vars for the job
```
pavolloffay, line 27: It does not seem that these env vars are used anywhere

cmd/gather/README.md
```diff
@@ -0,0 +1,36 @@
...
+```sh
+oc adm must-gather --image=<must-gather-image> -- /usr/bin/must-gather --operator-namespace opentelemetry-operator-system
+```
+
+### Using it as a CLI
+
+You only need to build and run:
+```sh
+make must-gather
+./bin/must-gather --help
```
pavolloffay, line 33: the make command attaches the arch suffix

cmd/gather/README.md
```diff
@@ -0,0 +1,36 @@
...
+```
+
+This is the recommended way to do it if you are not using OpenShift.
```
pavolloffay, line 36: I would move this sentence to the top of the section

cmd/gather/main.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	err = cluster.GetOperatorDeploymentInfo()
+	if err != nil {
+		log.Fatalln(err)
+	}
+	err = cluster.GetOLMInfo()
+	if err != nil {
+		log.Fatalln(err)
+	}
+	err = cluster.GetOpenTelemetryCollectors()
```
pavolloffay, line 78: Does it add collector logs?

pavolloffay (commented):
a couple of nits and a question

.github/workflows/publish-must-gather.yaml
iblancasa: Oops.

.chloggen/3149-add-must-gather.yaml
iblancasa: `OpenTelemetryCollector` and `Instrumentation`. I'll add them

cmd/gather/main.go
iblancasa: Just the collector. If you want, I can also add the collector logs but I would prefer to add it as part of a new PR. Otherwise we will never merge this one.

iblancasa:
@pavolloffay I addressed your comments. Is this ok to merge?

.chloggen/3149-add-must-gather.yaml
```diff
@@ -0,0 +1,25 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: instrumentation, collector
```
pavolloffay, line 5: auto-instrumentation

cmd/gather/main.go
pavolloffay: Please do in a follow up PR

pavolloffay (commented):
just a nit for changelog

pavolloffay (approved):

Issue #3157, opened 2024-07-23
When pod.name is not empty, service.instance.id incorrectly 


dominicqi:
https://github.com/open-telemetry/opentelemetry-operator/blob/09d25e17590bbe388f5a62620fe6ef73fae0d371/pkg/instrumentation/sdk.go#L299

Eg:  
When the workload type is StatefulSet, the OTEL_RESOURCE_ATTRIBUTES_POD_NAME env will not exist.
service.instance.id=namespace.$(OTEL_RESOURCE_ATTRIBUTES_POD_NAME).container

PR #3158, opened 2024-07-23
Fix service instance id if Pod Name not empty

dominicqi:
**Description:** 
Fix service.instance.id correctly if the pod.name already exists.

**Link to tracking Issue(s):**  #3157 

- Resolves: #3157 

**Testing:** Change some tests that the pod name is not empty, OTEL_RESOURCE_ATTRIBUTES_POD_NAME does not exist, service.instance.id  should be the actual pod name


pkg/instrumentation/sdk.go
```diff
@@ -472,7 +472,11 @@ func (i *sdkInjector) createResourceMap(ctx context.Context, otelinst v1alpha1.I
 	k8sResources[semconv.K8SPodNameKey] = pod.Name
 	k8sResources[semconv.K8SPodUIDKey] = string(pod.UID)
 	k8sResources[semconv.K8SNodeNameKey] = pod.Spec.NodeName
-	k8sResources[semconv.ServiceInstanceIDKey] = createServiceInstanceId(ns.Name, fmt.Sprintf("$(%s)", constants.EnvPodName), pod.Spec.Containers[index].Name)
+	podName := pod.Name
+	if podName == "" {
+		podName = fmt.Sprintf("$(%s)", constants.EnvPodName)
+	}
+	k8sResources[semconv.ServiceInstanceIDKey] = createServiceInstanceId(ns.Name, podName, pod.Spec.Containers[index].Name)
```
swiatekm, line 479: Instead of adding a special case here to account for a different special case later added in `injectCommonSDKConfig`, could we not always set `OTEL_RESOURCE_ATTRIBUTES_POD_NAME` from the downward API, and then always use that value in the resource map? That should be simpler and more consistent.

dominicqi: Yes, it is better.

dominicqi: Fixed @swiatekm

pkg/instrumentation/sdk.go
```diff
@@ -295,18 +295,18 @@ func (i *sdkInjector) injectCommonSDKConfig(ctx context.Context, otelinst v1alph
...
+		ValueFrom: &corev1.EnvVarSource{
+			FieldRef: &corev1.ObjectFieldSelector{
+				FieldPath: "metadata.name",
 			},
-		})
-		resourceMap[string(semconv.K8SPodNameKey)] = fmt.Sprintf("$(%s)", constants.EnvPodName)
-	}
+		},
+	})
+	resourceMap[string(semconv.K8SPodNameKey)] = fmt.Sprintf("$(%s)", constants.EnvPodName)
```
swiatekm, line 307: We should also stop setting this in `createResourceMap`.

swiatekm (commented):
Still needs a bit of cleanup, otherwise LGTM.

pkg/instrumentation/sdk.go
dominicqi: https://github.com/open-telemetry/opentelemetry-operator/blob/934c1d8ef90dbf0d570c52e115148c09c63d3959/pkg/instrumentation/sdk.go#L416-L420

We can retain this, because other places are using it.

dominicqi: also `nginx` and `httpd`

swiatekm (approved):

jaronoff97 (approved):

jaronoff97:
@dominicqi thank you! :bow:

PR #3159, opened 2024-07-23
bump opamp go to v0.15.0

jaronoff97:
**Description:** 
Bumps to latest opamp go

- Resolves #3019 

**Testing:** unit tests updated

### DIFF
--- a/tests/e2e-opampbridge/opampbridge/00-install.yaml
+++ b/tests/e2e-opampbridge/opampbridge/00-install.yaml
@@ -15,7 +15,6 @@ spec:
       containers:
         - name: e2e-test-app-bridge-server
           image: ghcr.io/open-telemetry/opentelemetry-operator/e2e-test-app-bridge-server:ve2e
-          imagePullPolicy: Always
           ports:
             - containerPort: 4320
             - containerPort: 4321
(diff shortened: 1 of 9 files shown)


yuriolisa (approved):

tests/e2e-opampbridge/opampbridge/00-install.yaml
```diff
@@ -14,7 +14,7 @@ spec:
     spec:
       containers:
         - name: e2e-test-app-bridge-server
-          image: ghcr.io/open-telemetry/opentelemetry-operator/e2e-test-app-bridge-server:main
+          image: ghcr.io/open-telemetry/opentelemetry-operator/e2e-test-app-bridge-server:ve2e
```
swiatekm, line 17: Where in this PR are we actually building this version? Shouldn't this be treated the same way other versions are - that is, injected via kustomize?

jaronoff97: we build this via the updated make command for prepare-e2e which is always run with `make prepare-e2e VERSION=e2e` resulting in this being built and loaded into kind. I initially didn't do this via Kustomize because I wasn't 100% sure how to do that (right now a lot of that is done to override the operator deployment) but let me see if i can do it for this one too.

swiatekm: Maybe the CI runs that command this way, but there's no such requirement for local tests. Maybe we can just hardcode this version? Not like this image is used outside of e2e tests.

swiatekm (approved):

Issue #3151, opened 2024-07-22
The annotations-filter startup parameter is not used for configmap, rbac, service, serviceaccount resources.


chinaran:
### Component(s)

collector

## Description

annotations-filter has a default configuration that filters kubectl.kubernetes.io/last-applied-configuration.

## Steps to Reproduce

deploy a otel-collector instance.

## Expected Result

None of the resources installed by opentelemetry-operator have kubectl.kubernetes.io/last-applied-configuration annotations.

## Actual Result

The configmap, rbac, service, serviceaccount resources are annotated kubectl.kubernetes.io/last-applied-configuration.

### Kubernetes Version

v1.29.2

### Operator version

v0.104.0

### Collector version

v0.104.0

## Environment
OS:  MacOS Kind

### Additional context

I found that the hpa resource filters the kubectl.kubernetes.io/last-applied-configuration annotation.
Reference: https://github.com/open-telemetry/opentelemetry-operator/blob/v0.104.0/internal/manifests/collector/horizontalpodautoscaler.go# L31.

```go
annotations, err := manifestutils.Annotations(params.OtelCol, params.Config.AnnotationsFilter())
if err != nil {
	return nil, err
}
```

When I try to fix this issue using similar code, I have the following two queries:

1. whether Prometheus annotations should be added only in PodAnnotations.

2. Should the opentelemetry-operator-config/sha256 annotation be added to all resources.

```go
// Annotations return the annotations for OpenTelemetryCollector pod.
func Annotations(instance v1beta1.OpenTelemetryCollector, filterAnnotations []string) (map[string]string, error) {
	// new map every time, so that we don't touch the instance's annotations
	annotations := map[string]string{}

	// Enable Prometheus annotations by default if DisablePrometheusAnnotations is nil or true
	if !instance.Spec.Observability.Metrics.DisablePrometheusAnnotations {
		// Set default Prometheus annotations
		annotations["prometheus.io/scrape"] = "true"
		annotations["prometheus.io/port"] = "8888"
		annotations["prometheus.io/path"] = "/metrics"
	}

	// allow override of prometheus annotations
	if nil != instance.ObjectMeta.Annotations {
		for k, v := range instance.ObjectMeta.Annotations {
			if !IsFilteredSet(k, filterAnnotations) {
				annotations[k] = v
			}
		}
	}

	hash, err := GetConfigMapSHA(instance.Spec.Config)
	if err != nil {
		return nil, err
	}

[...]

PR #3163, opened 2024-07-25
feat: configmap, rbac, service, serviceaccount apply annotations-filter

chinaran:

- Resolves: #3151

### DIFF
--- a/internal/manifests/collector/daemonset_test.go
+++ b/internal/manifests/collector/daemonset_test.go
@@ -266,7 +266,7 @@ func TestDaemonsetFilterAnnotations(t *testing.T) {
 	d, err := DaemonSet(params)
 	require.NoError(t, err)
 
-	assert.Len(t, d.ObjectMeta.Annotations, 1)
+	assert.Len(t, d.ObjectMeta.Annotations, 0)
 	for k := range excludedAnnotations {
 		assert.NotContains(t, d.ObjectMeta.Annotations, k)
 	}
--- a/internal/manifests/collector/daemonset_test.go
+++ b/internal/manifests/collector/daemonset_test.go
@@ -55,9 +55,6 @@ func TestDaemonSetNewDefault(t *testing.T) {
 	// verify
 	assert.Equal(t, "my-instance-collector", d.Name)
 	assert.Equal(t, "my-instance-collector", d.Labels["app.kubernetes.io/name"])
-	assert.Equal(t, "true", d.Annotations["prometheus.io/scrape"])
-	assert.Equal(t, "8888", d.Annotations["prometheus.io/port"])
-	assert.Equal(t, "/metrics", d.Annotations["prometheus.io/path"])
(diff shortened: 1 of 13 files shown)


yuriolisa (approved):

jaronoff97:
@chinaran can you add a change log entry when you get a chance please? Thank you for your contribution!

chinaran:
@jaronoff97 Thanks for the reminder, I've added the changelog.

internal/manifests/collector/daemonset_test.go
```diff
@@ -55,9 +55,6 @@ func TestDaemonSetNewDefault(t *testing.T) {
 	// verify
 	assert.Equal(t, "my-instance-collector", d.Name)
 	assert.Equal(t, "my-instance-collector", d.Labels["app.kubernetes.io/name"])
-	assert.Equal(t, "true", d.Annotations["prometheus.io/scrape"])
-	assert.Equal(t, "8888", d.Annotations["prometheus.io/port"])
-	assert.Equal(t, "/metrics", d.Annotations["prometheus.io/path"])
```
jaronoff97, line 60: can you add an assertion that this is still on the pod spec?

jaronoff97: same for statefulset and deployment

jaronoff97 (approved):

internal/manifests/manifestutils/annotations.go
```diff
@@ -27,15 +27,6 @@ func Annotations(instance v1beta1.OpenTelemetryCollector, filterAnnotations []st
...
 
-	// Enable Prometheus annotations by default if DisablePrometheusAnnotations is nil or true
-	if !instance.Spec.Observability.Metrics.DisablePrometheusAnnotations {
-		// Set default Prometheus annotations
-		annotations["prometheus.io/scrape"] = "true"
-		annotations["prometheus.io/port"] = "8888"
-		annotations["prometheus.io/path"] = "/metrics"
-	}
-
-	// allow override of prometheus annotations
```
swiatekm, line 38: Adding the ConfigMap Hash annotation should also be moved to `PodAnnotations`. We don't need that annotation anywhere else, and this should also reduce the amount of test changes.

chinaran: I somewhat agree with you about adding opentelemetry-operator-config/sha256 annotation only to PodAnnotations.
@jaronoff97 @swiatekm 
 can we agree on that?

iblancasa (approved):

internal/manifests/manifestutils/annotations.go
jaronoff97: yes, i think that makes sense.

chinaran: @swiatekm I've changed it, please check again.

swiatekm (approved):

jaronoff97:
@chinaran thank you for your contribution!

chinaran:
> @chinaran thank you for your contribution!

Glad to be part of it!

PR #3165, opened 2024-07-25
instrumentation/python: set OTEL_EXPORTER_OTLP_PROTOCOL to http/protobuf

xrmx:

Instead of setting the environment variable for each specific signal use the common one that it is also the one that is set by the distro.

Added test case for the new environment variable set

None


xrmx:
I'm not able to sort out what's wrong with the e2e-tests failure but looks unrelated.

jaronoff97:
@xrmx fixed that, e2e can be flaky at times.

jaronoff97 (approved):

pkg/instrumentation/python.go
```diff
@@ -71,6 +72,15 @@ func injectPythonSDK(pythonSpec v1alpha1.Python, pod corev1.Pod, index int) (cor
 		})
 	}
 
+	// Set OTEL_EXPORTER_OTLP_PROTOCOL to http/protobuf if not set by user because it is what our autoinstrumentation supports.
+	idx = getIndexOfEnv(container.Env, envOtelExporterOTLPProtocol)
```
TylerHelmuth, line 76: If we set the top-level env var we no longer need to set the signal specific-env vars.

xrmx: Yep, will drop signal specific env vars

TylerHelmuth (approved):

pkg/instrumentation/python.go
xrmx: Updated PR

xrmx:
I don't understand what this means:

```
pkg/instrumentation/python.go:26: File is not `gci`-ed with --skip-generated -s standard -s default -s prefix(github.com/open-telemetry/opentelemetry-operator) (gci)
```

TylerHelmuth:
@xrmx it means the import statements at the top of the file are not arranged correctly into 3 sections.

xrmx:
> @xrmx it means the import statements at the top of the file are not arranged correctly into 3 sections.

Was not that, the issue was that there were too many spaces in the const definitions before the `"` :sweat_smile:

iblancasa (approved):

swiatekm (approved):

.chloggen/3165-python-otlp-protocol-env-var.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: collector
```
pavolloffay, line 5: this should be set to auto-instrumentation

xrmx, line 5:
```suggestion
component: auto-instrumentation
```

xrmx: Updated, thanks

xrmx:
Hello, can I do something to move this forward? BTW merged main and e2e tests failure looks unrelated. Thanks!

pavolloffay (approved):

Issue #3171, opened 2024-07-26
OpAMP Bridge Server incompletely loaded out configs


3-commerce:
### Component(s)

opamp

### What happened?

When OpAMP Server displays the current configuration of Otel Collector config. Some parts of configuration `spec.config` incompletely loaded out.

## Description

OpAMP Server load out configs part incompletely

## Steps to Reproduce
1. Install open-telemetry/opentelemetry-kube-stack (Otel Operator, Otel Col, Instrumentation, OpAMPBridge)
2. Setup a VM to host the [OpAMP Server](https://github.com/open-telemetry/opamp-go/tree/main/internal/examples/server) and disable the cert tls before run it.
3. Configure the otel col and point the endpoint on opamp bridge to VM host `ws://IP:4320/v1/opamp`
4. Check the current agent connected in opamp server vm

## Expected Result
<img width="278" alt="Screenshot 2024-07-27 at 00 30 40" src="https://github.com/user-attachments/assets/fcfa608b-3bad-4841-8cbb-2badb7c0e1fe">

## Actual Result
<img width="269" alt="Screenshot 2024-07-27 at 00 31 08" src="https://github.com/user-attachments/assets/917e47ae-066e-4e3a-9d60-999fa187261b">

### Kubernetes Version

1.29.6

### Operator version

0.103.0

### Collector version

0.105.0

## Environment
Op AMP Server
OS: Ubuntu 24.04 LTS
Compiler(if manually compiled): go1.22.5 linux/amd64

### Log output

```shell
## Otel Operator

{"level":"INFO","timestamp":"2024-07-26T15:59:08Z","message":"Starting workers","controller":"opentelemetrycollector","controllerGroup":"opentelemetry.io","controllerKind":"OpenTelemetryCollector","worker count":1}
{"level":"INFO","timestamp":"2024-07-26T15:59:08Z","message":"Starting workers","controller":"opampbridge","controllerGroup":"opentelemetry.io","controllerKind":"OpAMPBridge","worker count":1}
```

## Otel Collector
```
2024-07-26T15:56:29.312Z	info	otlpreceiver@v0.105.0/otlp.go:152	Starting HTTP server	{"kind": "receiver", "name": "otlp", "data_type": "traces", "endpoint": "0.0.0.0:4318"}
2024-07-26T15:56:29.312Z	info	service@v0.105.0/service.go:224	Everything is ready. Begin running and processing data.
[...]

PR #3178, opened 2024-07-29
change to pointer type for anyconfig to allow for effective marshalling

jaronoff97:
**Description:** This PR changes the type for Receivers and Exporters to be a pointer to AnyConfig, this was preventing the bridge from effectively marshalling the configuration to send. (see issue for details). This is technically a breaking change for people that use the v1beta1 type directly, however, it's a no-op for users of the collector. 

- Resolves: #3171

**Testing:** Unit test and e2e test updated

**Documentation:** n/a


swiatekm:
Why exactly is the current state an issue for the opamp bridge? I recall us changing this recently, and I don't want to do so again without a good reason. At the very least we should leave a comment explaining the decision.

jaronoff97:
@swiatekm the problem today is that because receivers and exporters is not a pointer, when we go to marshal the configuration, the custom Marshal methods (which have pointer receivers) are never called. This results in the outgoing config as `{}` rather than the actual configuration block.

jaronoff97:
The thing we needed to fix recently here was how processors are defaulted iirc https://github.com/open-telemetry/opentelemetry-operator/commit/107d2c31a61f1cea3a1d6b21241c5fee7ff79f41

cmd/operator-opamp-bridge/agent/agent.go
```diff
@@ -291,7 +291,7 @@ func (agent *Agent) getEffectiveConfig(ctx context.Context) (*protobufs.Effectiv
 	}
 	instanceMap := map[string]*protobufs.AgentConfigFile{}
 	for _, instance := range instances {
-		marshaled, err := yaml.Marshal(instance)
+		marshaled, err := yaml.Marshal(&instance)
```
jaronoff97, line 294: From @jmacd 
> when you put a value into an interface, it becomes something you can't address, but if you put a pointer in, you can still address it. The `allowAddr` variable changes in the expression above. you can't mutate values in interfaces either, because they are not addressable

jmacd (approved):

.chloggen/fix-opamp-bridge-missing-config.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: opamp
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fixes the type for AnyConfig in the collector to allow for proper marshalling
```
swiatekm, line 8: This description is now out of date I think, we're now just fixing a marshalling bug in the bridge.

swiatekm (approved):

PR #3180, opened 2024-07-29
Enhanced Webhook Reconciliation Errors

### LINKED ISSUE #2399: Block bad resources from reconciliation

taniyourstruly:
Closes #2399  

Added reconciliation errors for webhook events in the operator


apis/v1beta1/collector_webhook.go
```diff
@@ -454,13 +468,32 @@ func checkAutoscalerSpec(autoscaler *AutoscalerSpec) error {
 	return nil
 }
 
-func SetupCollectorWebhook(mgr ctrl.Manager, cfg config.Config, reviewer *rbac.Reviewer, metrics *Metrics) error {
+type BuildValidator func(c OpenTelemetryCollector) (admission.Warnings, error)
+
+func NewCollectorWebhook(
```
jaronoff97, line 473: may be worth mentioning that this is mostly used for testing purposes.

main.go
```diff
@@ -19,6 +19,7 @@ import (
 	"crypto/tls"
 	"flag"
 	"fmt"
+	"github.com/open-telemetry/opentelemetry-operator/internal/autodetect"
```
jaronoff97, line 22: import order is wrong here

main.go
```diff
@@ -391,7 +394,19 @@ func main() {
 
 		}
 
-		if err = otelv1beta1.SetupCollectorWebhook(mgr, cfg, reviewer, crdMetrics); err != nil {
+		bv := func(collector otelv1beta1.OpenTelemetryCollector) (admission.Warnings, error) {
```
jaronoff97, line 397: It looks like we're not using the admission.Warnings at all here, should we move the errors to the warnings, or just get rid of the warnings and only do errors?

apis/v1beta1/collector_webhook.go
```diff
@@ -438,13 +446,34 @@ func checkAutoscalerSpec(autoscaler *AutoscalerSpec) error {
 	return nil
 }
 
-func SetupCollectorWebhook(mgr ctrl.Manager, cfg config.Config, reviewer *rbac.Reviewer, metrics *Metrics) error {
+// BuildValidator is mostly used for testing purposes.
```
jaronoff97, line 449: can you add a real description here for the purpose of this type?

apis/v1beta1/collector_webhook.go
```diff
@@ -438,13 +446,34 @@ func checkAutoscalerSpec(autoscaler *AutoscalerSpec) error {
...
+// +kubebuilder:object:generate=false
+type BuildValidator func(c OpenTelemetryCollector) admission.Warnings
+
+func NewCollectorWebhook(
```
jaronoff97, line 453: Can you note that this is mostly for testing

apis/v1beta1/collector_webhook.go
```diff
@@ -438,13 +446,34 @@ func checkAutoscalerSpec(autoscaler *AutoscalerSpec) error {
...
+		logger:   logger,
+		scheme:   scheme,
+		cfg:      cfg,
+		reviewer: reviewer,
+		bv:       bv,
+	}
+}
+
+func SetupCollectorWebhook(mgr ctrl.Manager, cfg config.Config, reviewer *rbac.Reviewer, metrics *Metrics, bv BuildValidator) error {
 	cvw := &CollectorWebhook{
```
jaronoff97, line 470: Can we replace instance of direct struct creation with the new constructor?

main.go
```diff
@@ -404,7 +407,22 @@ func main() {
 
 		}
 
-		if err = otelv1beta1.SetupCollectorWebhook(mgr, cfg, reviewer, crdMetrics); err != nil {
+		bv := func(collector otelv1beta1.OpenTelemetryCollector) admission.Warnings {
```
jaronoff97, line 410: Maybe we should put this method on the collector controller such that we can just pass a reference to its function rather than inlining this closure.

apis/v1beta1/collector_webhook_test.go
```diff
@@ -1244,20 +1304,58 @@ func TestOTELColValidatingWebhook(t *testing.T) {
 			},
 			expectedErr: "the OpenTelemetry Collector mode is set to statefulset, which does not support the attribute 'deploymentUpdateStrategy'",
 		},
+		{
+			name: "missing port for ingress type",
```
taniyourstruly, line 1308: The changes in this file are mostly package fixes. This is an added test to check that webhook reconciliation errors are working properly.

.chloggen/enhanced-webhook.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: added reconciliation errors for webhook events
```
iblancasa, line 8: Please, can you elaborate a bit more?

main.go
iblancasa: I agree

main.go
```diff
@@ -366,13 +366,15 @@ func main() {
...
 		Client:   mgr.GetClient(),
 		Log:      ctrl.Log.WithName("controllers").WithName("OpenTelemetryCollector"),
 		Scheme:   mgr.GetScheme(),
 		Config:   cfg,
 		Recorder: mgr.GetEventRecorderFor("opentelemetry-operator"),
-	}).SetupWithManager(mgr); err != nil {
+	})
+
+	if err = collectorReconciler.SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "OpenTelemetryCollector")
```
iblancasa, line 378: I think this change can be reverted, right?

main.go
```diff
@@ -404,7 +406,10 @@ func main() {
 
 		}
 
-		if err = otelv1beta1.SetupCollectorWebhook(mgr, cfg, reviewer, crdMetrics); err != nil {
+		reconciler := controllers.NewReconciler(controllers.Params{})
+		bv := otelv1beta1.SetupCollectorWebhook(mgr, cfg, reviewer, crdMetrics, reconciler.Validate)
+
+		if err = bv; err != nil {
 			setupLog.Error(err, "unable to create webhook", "webhook", "OpenTelemetryCollector")
```
iblancasa, line 413: Can't we create the reconciler inside the `SetupCollectorWebhook` method? Because I think we aren ot using anything from this scope to create the `reconciler` object, right?

controllers/opentelemetrycollector_controller.go
```diff
@@ -168,7 +188,7 @@ func (r *OpenTelemetryCollectorReconciler) getConfigMapsToRemove(configVersionsT
 	return ownedConfigMaps
 }
 
-func (r *OpenTelemetryCollectorReconciler) getParams(instance v1beta1.OpenTelemetryCollector) (manifests.Params, error) {
+func (r *OpenTelemetryCollectorReconciler) GetParams(instance v1beta1.OpenTelemetryCollector) (manifests.Params, error) {
```
iblancasa, line 191: Do we still need to change the visibility of this method?

apis/v1beta1/collector_webhook_test.go
```diff
@@ -35,34 +41,48 @@ import (
...
-
-	"github.com/open-telemetry/opentelemetry-operator/internal/config"
-	"github.com/open-telemetry/opentelemetry-operator/internal/rbac"
 )
 
 var (
 	testScheme = scheme.Scheme
 )
 
+var cfgYaml = `receivers:
```
iblancasa, line 50: This is used just in the `TestOTELColValidatingWebhook` method, right? Can we move it to there?

apis/v1beta1/collector_webhook_test.go
```diff
@@ -548,29 +546,34 @@ func TestOTELColValidatingWebhook(t *testing.T) {
 	three := int32(3)
 	five := int32(5)
 
-	cfg := Config{}
+	if err := v1beta1.AddToScheme(testScheme); err != nil {
+		fmt.Printf("failed to register scheme: %v", err)
+		os.Exit(1)
+	}
```
iblancasa, line 552:
```suggestion
	v1beta1.AddToScheme(testScheme)
    require.NoError(e, err)
```

iblancasa (approved):

controllers/opentelemetrycollector_controller.go
```diff
@@ -75,6 +76,25 @@ type Params struct {
 	Config   config.Config
 }
 
+func (r *OpenTelemetryCollectorReconciler) Validate(otelcol v1beta1.OpenTelemetryCollector) admission.Warnings {
+	var warnings admission.Warnings
+	cfg := config.New(
+		config.WithCollectorImage("default-collector"),
+		config.WithTargetAllocatorImage("default-ta-allocator"),
+	)
+	params := manifests.Params{
```
jaronoff97, line 85: you should just use this method https://github.com/open-telemetry/opentelemetry-operator/blob/65477b55214a53495343e1d2a7e39c85b3c33290/controllers/opentelemetrycollector_controller.go#L191-L208

apis/v1beta1/collector_webhook.go
```diff
@@ -407,14 +415,32 @@ func checkAutoscalerSpec(autoscaler *AutoscalerSpec) error {
...
-		reviewer: reviewer,
-		logger:   mgr.GetLogger().WithValues("handler", "CollectorWebhook", "version", "v1beta1"),
-		scheme:   mgr.GetScheme(),
+// BuildValidator enables running the manifest generators for the collector reconciler
+// Kubebuilder is used for testing
+// +kubebuilder:object:generate=false
+
+type BuildValidator func(c OpenTelemetryCollector) admission.Warnings
+
+func NewCollectorWebhook(
```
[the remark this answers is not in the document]
jaronoff97: i think we don't need the testing comment anymore.

jaronoff97 (approved):
one minor thing, otherwise this looks great!

swiatekm (commented):
Thanks for working on this change, it should greatly improve the feedback we give users via the validation webhook.

Something I'm concerned about in this PR is that we pass a lot of objects to the builder via the `Params` struct. I *think* these aren't really used and the function is effectively pure or close to it, but for this use case we really don't want any side effects. As an innocent example, if any of the builders emits a log line, that log line will appear in the operator's output (because it's using the main logger), which we don't want.

I think the short-term solution is to avoid creating the reconciler, run `collector.Build` directly in the validator, and just provide fake objects to `Params` other than the otelcol instance and config.

The long-term solution would involve restricting what params the manifest generation process actually needs and only passing those - actually making it pure. WDYT @jaronoff97 ?

jaronoff97:
@swiatekm i think that's reasonable... that's what tani had before and then i advised to re-use but I think it would be fine to just fake it for now. @taniyourstruly i would go for what is recommended there and then open an issue for the long term solution mentioned.

controllers/opentelemetrycollector_controller.go
```diff
@@ -79,13 +79,16 @@ type Params struct {
...
-	if err != nil {
-		warnings = append(warnings, err.Error())
-		return warnings
+	cfg := config.New(
+		config.WithCollectorImage("default-collector"),
+		config.WithTargetAllocatorImage("default-ta-allocator"),
+	)
+	params := manifests.Params{
+		Log:     logr.Discard(),
+		Config:  cfg,
```
jaronoff97, line 88: i think for the config we can propagate that https://github.com/open-telemetry/opentelemetry-operator/blob/b6f0911cda24680d2f1ae521d469b01b4f4e13e6/controllers/opentelemetrycollector_controller.go#L194

controllers/opentelemetrycollector_controller.go
```diff
@@ -75,6 +76,22 @@ type Params struct {
 	Config   config.Config
 }
 
+func (r *OpenTelemetryCollectorReconciler) Validate(otelcol v1beta1.OpenTelemetryCollector) admission.Warnings {
```
swiatekm, line 79: Is there any reason for this to be a method on the reconciler? Looks like the only thing we're using from it is the config, and the webhook has access to that anyway. I'd just make it a function in `collector_webhook.go`. That would also let us avoid the setup changes.

jaronoff97: we can't do it in the webhook because then we would have an import cycle (through the collector.Build) unfortunately, previously this was an anonymous function in the main.go file, but it felt a bit odd to put it there IMO.

swiatekm: Ah, makes sense. Honestly, I think I'd like it better as an anonymous function in main, or in some other package. I don't like adding it to the reconciler, which is completely unrelated functionally, and I don't really like creating a runtime dependency between it and the webhook.

That said, I do want to merge this PR, so if you and @taniyourstruly feel strongly about keeping it the way it is, I can accept it.

jaronoff97: i think we can make it an anon func or just a named func in main.

swiatekm (approved):
I'm ok with merging this, but we should open an issue to limit the amount of parameters we pass to the builders.

jaronoff97 (approved):

PR #3185, opened 2024-07-31
[Chore] Fix monitoring test case.

IshwarKanse:
**Description:**
*Fix the monitoring test case.

### DIFF
--- a/tests/e2e-openshift/monitoring/01-assert.yaml
+++ b/tests/e2e-openshift/monitoring/01-assert.yaml
@@ -1,10 +1,6 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  annotations:
-    prometheus.io/path: /metrics
-    prometheus.io/port: "8888"
-    prometheus.io/scrape: "true"
   labels:
     app.kubernetes.io/component: opentelemetry-collector
     app.kubernetes.io/managed-by: opentelemetry-operator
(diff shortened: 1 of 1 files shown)


pavolloffay:
The CI seems to be stuck

tests/e2e-opampbridge/opampbridge/00-install.yaml
```diff
@@ -14,7 +14,7 @@ spec:
     spec:
       containers:
         - name: e2e-test-app-bridge-server
-          image: ghcr.io/open-telemetry/opentelemetry-operator/e2e-test-app-bridge-server:ve2e
+          image: ghcr.io/open-telemetry/opentelemetry-operator/e2e-test-app-bridge-server:main
           ports:
```
jaronoff97, line 18: @IshwarKanse i need to do this so that we can use the bridge-server image built during CI rather than the one off of main. Is there a way to load the image into openshift?

IshwarKanse: Ok, np. I'll update the OpenShift CI to build the image as well and use it in the job run. I'll update this PR removing this change.

IshwarKanse: Done, removed the change.

jaronoff97 (approved):

tests/e2e-opampbridge/opampbridge/00-install.yaml
jaronoff97: thank you :bow:

iblancasa (approved):

Issue #3198, opened 2024-08-06
Otel operator creating invalid PDB


taer:
### Component(s)

collector

### What happened?

I configured my otel collector with a pdb, shown below in the config setting section. In the YAML for the `OpenTelemetryCollector`, I asked for a PDB with min-available of 1

k8s is mad about this, w/ the error showing up below in the logs section. The PDB that it is creating seems to have a default PDB w/ a `maxUnavailable: 1`  It then merges my minAvailable and makes an invalid PDB

If the user specifies a pdb in the OpenTelmetry yaml config, then it should be respected instead of merged.
From this [link](https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/api.md#opentelemetrycollectorspecpoddisruptionbudget)   I cannot find a way to disable the max option. But that max option shouldn't be present if I specified a min

### Kubernetes Version

1.30

### Operator version

0.64.1

### Collector version

.103.1

### Environment information

OpenTelemetry CRD has this
```
apiVersion: opentelemetry.io/v1beta1
kind: OpenTelemetryCollector
metadata:
  name: opentelemetry
spec:
  podDisruptionBudget:
    minAvailable: 1
```

### Log output

```shell
k8s events has this

observability                  9m31s (x9 over 20m)     Warning   Error                          OpenTelemetryCollector/opentelemetry                                  (combined from similar events): failed to create objects for opentelemetry: PodDisruptionBudget.policy "opentelemetry-collector" is invalid: spec: Invalid value: policy.PodDisruptionBudgetSpec{MinAvailable:(*intstr.IntOrString)(0xc03d6aad60), Selector:(*v1.LabelSelector)(0xc03d6aad80), MaxUnavailable:(*intstr.IntOrString)(0xc03d6aada0), UnhealthyPodEvictionPolicy:(*policy.UnhealthyPodEvictionPolicyType)(nil)}: minAvailable and maxUnavailable cannot be both set
```
```


### Additional context

_No response_

jaronoff97:
hm our [default is to set maxUnavailable to 1](https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1beta1/collector_webhook.go#L95-L106), but that should be overridden by your config. We also just replace the spec [here](https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/manifests/mutate.go#L233) so it's odd that this is the case. I'll see if I can reproduce this.

jaronoff97:
[...]

PR #3200, opened 2024-08-06
Fix pdb defaulting logic

jaronoff97:
**Description:** Fixes a bug where we setting the field on the CRD itself which was causing it to be added permanently to the CR in the defaulting webhook. This meant that when a user went to apply a change and didn't want that default anymore we would crash.

- Resolves:  #3198 

**Testing:** Unit tests, e2e tests, manual


internal/manifests/collector/poddisruptionbudget_test.go
```diff
@@ -29,38 +29,84 @@ import (
 )
 
 func TestPDB(t *testing.T) {
-	type test struct {
-		name           string
-		MinAvailable   *intstr.IntOrString
+	type expected struct {
```
jaronoff97, line 32: i swapped around where the defaulting is happening so now we should test that here.

iblancasa (approved):

tests/e2e-pdb/pdb/00-assert.yaml
```diff
@@ -6,4 +6,4 @@ spec:
   selector:
     matchLabels:
       app.kubernetes.io/component: opentelemetry-collector
-  minAvailable: 1
+  maxUnavailable: 1
```
pavolloffay, line 9: Shall we document that by default this is set?

pavolloffay (approved):

Issue #3201, opened 2024-08-07
The labels-filter startup parameter, flag is --label, which is inconsistent with the example and variable name.


chinaran:
### Component(s)

collector

## Description

The labels-filter startup parameter, flag is --label, which is inconsistent with the example and variable name.

## Expected Result

--labels-filter

## Actual Result

--label

### Kubernetes Version

v1.29.2

### Operator version

v0.105.0

### Collector version

v0.104.0

### Additional context

Reference Code: https://github.com/open-telemetry/opentelemetry-operator/blob/v0.105.0/main.go#L173

```go
pflag.StringArrayVar(&labelsFilter,
 "label",
 []string{},
 "Labels to filter away from propagating onto deploys.
  It should be a string array containing patterns,
 which are literal strings optionally containing a * wildcard character.
  Example: --labels-filter=.*filter.out will filter out labels that looks like: label.filter.out: true")
```

pavolloffay:
@chinaran would you like to fix it?

PR #3203, opened 2024-08-07
fix: example for labels-filter startup parameter --label

chinaran:
**Description:** The labels-filter startup parameter, flag is --label, which is inconsistent with the example and variable name.

- Resolves: #3201


iblancasa (changes requested):
This can be a breaking change for those using the current value. We should change the example, not the flag

pavolloffay:
+1 on aligning the example with the flag

chinaran:
@iblancasa @pavolloffay 
Modifying flag directly does cause breaking changes, so I modified the example now.
There are just some inconsistencies between flag and expressed functions, especially when compared with the `--annotations-filter` flag.

chinaran:
@iblancasa @pavolloffay
Additionally, there is currently a unit testing issue and I've also submitted a PR: https://github.com/open-telemetry/opentelemetry-operator/pull/3207

iblancasa (approved):

yuriolisa (approved):

jaronoff97 (approved):
This is fine for now, though I feel like we should align the naming in the future.

Issue #3112, opened 2024-07-08
Add support for k8s labels such as app.kubernetes.io/name


zeitlinger:
### Component(s)

auto-instrumentation

### Is your feature request related to a problem? Please describe.

k8s labels such as app.kubernetes.io/name are recommended to set the app name and similar

See https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

this should be supported by the otel operator

### Describe the solution you'd like

This mapping should work for the following labels:

- `app.kubernetes.io/name` - `service.name`
- `app.kubernetes.io/instance` - `service.instance.id`
- `app.kubernetes.io/version` - `service.version`
- `app.kubernetes.io/part-of` - `service.namespace`

And for the following annotations:

- `opentelemetry.io/resource.foo` - `foo` (for any key) - note that the [k8sattributes processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/README.md) can also process those annotations, see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/34040

jaronoff97:
@zeitlinger is this different from [this issue](https://github.com/open-telemetry/opentelemetry-operator/issues/2181)? Wondering if i can merge these together.

zeitlinger:
> @zeitlinger is this different from [this issue](https://github.com/open-telemetry/opentelemetry-operator/issues/2181)? Wondering if i can merge these together.

didn't know about it 
looks very similar

- the prefix in my pr is `opentelemetry.io/resource.` instead of `resource.opentelemetry.io/` - I don't have a strong preference
- I also have support for common k8s variables (see description)
- I think I've handled more edge cases in my pr

jaronoff97:
I think i prefer the `resource.opentelemetry.io/`, and agreed, I think your PR handles more. I'm going to close that issue and PR in favor of this one.

jaronoff97:
@zeitlinger i added the discuss-at-sig label as it seems from Tyler's comment on the PR that there's some bigger picture stuff we need to figure out in order to not break users. We have a SIG meeting tomorrow (8/29) at noon EST where we will discuss this.

zeitlinger:
> @zeitlinger i added the discuss-at-sig label as it seems from Tyler's comment on the PR that there's some bigger picture stuff we need to figure out in order to not break users. We have a SIG meeting tomorrow (8/29) at noon EST where we will discuss this.

great - I'll join

PR #2330, opened 2023-11-09
using opentelemetry.io* namespace for extracing resource and attribute labels from k8s Api

Horiodino:
**Description:** 
i have added recognizeUserDefinedValues()  that will recognize the pods with pentelemetry.io/ prefix
#2181

**Testing:** i haven't tested it yet , im  not sure is it right so please tell any changes to made

PR #3204, opened 2024-08-07
read resource attributes from annotations

zeitlinger:
Fixes https://github.com/open-telemetry/opentelemetry-operator/issues/3112

**Testing:**  Added integration test

**Documentation:** will be added to https://opentelemetry.io/docs/kubernetes/operator/automatic/ once released


jaronoff97:
per our discussion, @zeitlinger will rebase and expand on the capabilities introduced by #2330 :)

pkg/instrumentation/sdk_test.go
```diff
@@ -152,6 +152,16 @@ func TestSDKInjection(t *testing.T) {
...
 						},
 					},
+					Labels: map[string]string{
+						"app.kubernetes.io/name":     "app-name",
+						"app.kubernetes.io/instance": "app-id",
+						"app.kubernetes.io/version":  "v1",
+						"app.kubernetes.io/part-of":  "shop",
+					},
+					Annotations: map[string]string{
+
```
jaronoff97, line 162: nit: can we remove the space here

jaronoff97 (commented):
Can you also update the README.md, explaining the changes and also explaining the precedence of deciding these variables?

pkg/instrumentation/sdk_test.go
zeitlinger: done

jaronoff97 (approved):
@zeitlinger last thing – can you update the README.md with some more information about the new behavior? Maybe also include it in the subtext of the chlog entry

pkg/instrumentation/sdk.go
```diff
@@ -419,8 +422,22 @@ func chooseServiceName(pod corev1.Pod, resources map[string]string, index int) s
 	return pod.Spec.Containers[index].Name
 }
 
+func chooseLabelOrAnnotation(pod corev1.Pod, resource attribute.Key, labelKey string) string {
```
TylerHelmuth, line 425: The parameter name is `labelKey`, but when invoked the `constants.AnnotationAppName` value is used.  Then when indexing the annotations the resource is used, not the `labelKey`.  Is this wired up correctly?

pkg/instrumentation/sdk.go
```diff
@@ -395,6 +395,9 @@ func (i *sdkInjector) injectCommonSDKConfig(ctx context.Context, otelinst v1alph
 }
 
 func chooseServiceName(pod corev1.Pod, resources map[string]string, index int) string {
+	if name := chooseLabelOrAnnotation(pod, semconv.ServiceNameKey, constants.AnnotationAppName); name != "" {
+		return name
+	}
```
TylerHelmuth, line 400: I am concerned that this would be a breaking change.

pkg/instrumentation/sdk.go
```diff
@@ -432,8 +449,12 @@ func chooseServiceVersion(pod corev1.Pod, index int) string {
 
 // creates the service.instance.id following the semantic defined by
 // https://github.com/open-telemetry/semantic-conventions/pull/312.
-func createServiceInstanceId(namespaceName, podName, containerName string) string {
-	var serviceInstanceId string
+func createServiceInstanceId(pod corev1.Pod, namespaceName, podName, containerName string) string {
+	serviceInstanceId := chooseLabelOrAnnotation(pod, semconv.ServiceInstanceIDKey, constants.AnnotationAppInstance)
+	if serviceInstanceId != "" {
+		return serviceInstanceId
+	}
```
TylerHelmuth, line 456: I'm concerned that this would be a breaking change.

pkg/instrumentation/sdk.go
zeitlinger: yes, I'll include that in the changelog

pkg/instrumentation/sdk.go
zeitlinger: labelkey is looked up in labels:

https://github.com/open-telemetry/opentelemetry-operator/blob/bef08935239ce6c7859fa9c0d7ee59f074c8c306/pkg/instrumentation/sdk.go#L429

zeitlinger:
> @zeitlinger last thing – can you update the README.md with some more information about the new behavior? Maybe also include it in the subtext of the chlog entry

done

pavolloffay (approved):

pkg/instrumentation/sdk.go
TylerHelmuth: @jaronoff97 @pavolloffay I am pretty worried about this breaking change.  Changes to `service.name` will change emitted telemetry, which could affect end-users alerts/SLOs/dashboards/etc. This change will (likely) not  be detected with unit or e2e tests before a user updates their operator. Additionally, the operator's auto-instrumentation injection is already a huge obfuscation on top of already obfuscated auto-instrumentation, which means that our end-users are even less likely to understand the impact of this change.

I would much rather prefer this work be implemented in such a way that the annotation is used as the service name if no other service name is set OR if some other explicit configuration field is set that enables this behavior.

TylerHelmuth: The reason I call this out as a breaking change in the current implementation is that it is now defaulting to the annotation value if it exists.  Since it is a common annotation, it will probably be on most end-user's pods, resulting in the potential breaking change.

I would much rather this be the last option in this list to avoid the breaking change.

TylerHelmuth: In addition, there is no stable semantic convention that ranks which value should be used as the `service.name`. I'd prefer avoiding breaking changes on this topic until there is a stable semantic convention, so that we can minimize the number of breaking changes.

zeitlinger: we could solve this problem if we let the user define the mapping they prefer - so it would be an opt-in feature

jaronoff97: from SIG:
I think we should move forward with this by putting it within a new "defaults" section in the instrumentation CRD as so:
```
kind: Instrumentation
metadata:
  name: my-instrumentation
spec:
  defaults:
    useLabelsForResource: true
```

We also must then document in the readme the precedence for the automatic variables we set for users and how this boolean changes that.

pkg/instrumentation/sdk.go
zeitlinger: changed constants to have `label` prefix

pkg/instrumentation/sdk.go
zeitlinger: fixed

pavolloffay:
CI failed

README.md
```diff
@@ -733,7 +733,56 @@ spec:
   containers:
   - name: main-container
     image: your-image:tag
- ```
+```
+
+You can also use common labels to set resource attributes.
```
jaronoff97, line 739: let's have this under a new section

README.md
```diff
@@ -733,7 +733,56 @@ spec:
...
+
+#### Priority for setting resource attributes
+
+The priority for setting resource attributes is as follows (first found wins):
+
+- Resource attributes set via `OTEL_RESOURCE_ATTRIBUTES` and `OTEL_SERVICE_NAME` environment variables
+- Resource attributes calculated from the pod's metadata (e.g. `k8s.pod.name`)
+- Resource attributes set via the `Instrumentation` CR (in the `spec.resource.resourceAttributes` section)
+- Resource attributes set via annotations (with the `resource.opentelemetry.io/` prefix)
+- Resource attributes set via labels (e.g. `app.kubernetes.io/name`)
```
jaronoff97, line 783: let's change this to be a numbered list

jaronoff97: where does the new `useLabelsForResourceAttributes` fall into this? it is the last rule? if so, we should specify that you need to set defaults.useLabelsForResourceAttributes=true

apis/v1alpha1/instrumentation_types.go
```diff
@@ -114,6 +117,12 @@ type Sampler struct {
 	Argument string `json:"argument,omitempty"`
 }
 
+// Defaults defines default values for the instrumentation.
+type Defaults struct {
+	// UseLabelsForResourceAttributes defines whether to use common labels for resource attributes.
```
jaronoff97, line 122: let's include the mapping rules in this doc string too

pkg/instrumentation/apachehttpd.go
```diff
@@ -95,7 +95,7 @@ func injectApacheHttpdagent(_ logr.Logger, apacheSpec v1alpha1.ApacheHttpd, pod
 			MountPath: apacheAgentConfDirFull,
 		})
 		// remove resource requirements since those are then reserved for the lifetime of a pod
-		// and we definitely do not need them for the init container for cp command
+		// ]and we definitely do not need them for the init container for cp command
```
jaronoff97, line 98:
```suggestion
		// and we definitely do not need them for the init container for cp command
```

pkg/instrumentation/sdk.go
```diff
@@ -275,14 +278,14 @@ func (i *sdkInjector) injectCommonEnvVar(otelinst v1alpha1.Instrumentation, pod
 // and appIndex should be the same value.  This is true for dotnet, java, nodejs, and python instrumentations.
 // Go requires the agent to be a different container in the pod, so the agentIndex should represent this new sidecar
 // and appIndex should represent the application being instrumented.
-func (i *sdkInjector) injectCommonSDKConfig(ctx context.Context, otelinst v1alpha1.Instrumentation, ns corev1.Namespace, pod corev1.Pod, agentIndex int, appIndex int) corev1.Pod {
+func (i *sdkInjector) injectCommonSDKConfig(ctx context.Context, otelinst v1alpha1.Instrumentation, ns corev1.Namespace, pod corev1.Pod, useLabelsForResourceAttributes bool, agentIndex int, appIndex int) corev1.Pod {
```
jaronoff97, line 297: we shouldn't need the bool flag on this method if it's already in the otelinst object

pkg/instrumentation/sdk.go
```diff
@@ -410,7 +413,10 @@ func (i *sdkInjector) injectCommonSDKConfig(ctx context.Context, otelinst v1alph
 	return pod
 }
 
-func chooseServiceName(pod corev1.Pod, resources map[string]string, index int) string {
+func chooseServiceName(pod corev1.Pod, useLabelsForResourceAttributes bool, resources map[string]string, index int) string {
```
jaronoff97, line 416: can we add a docstring for this method that includes the precedence?

zeitlinger: also added a docstring to similar methods

zeitlinger:
@jaronoff97 please check again

README.md
```diff
@@ -717,7 +717,9 @@ spec:
 EOF
 ```
 
-### Setting instrumentation resource attributes via namespace annotations
+### Configure resource attributes
```
jaronoff97, line 720:
```suggestion
## Configure resource attributes
```

pkg/instrumentation/sdk.go
```diff
@@ -488,22 +547,14 @@ func (i *sdkInjector) createResourceMap(ctx context.Context, otelinst v1alpha1.I
...
+	k8sResources[semconv.ServiceInstanceIDKey] = createServiceInstanceId(pod, useLabelsForResourceAttributes, ns.Name, fmt.Sprintf("$(%s)", constants.EnvPodName), pod.Spec.Containers[index].Name)
 	i.addParentResourceLabels(ctx, otelinst.Spec.Resource.AddK8sUIDAttributes, ns, pod.ObjectMeta, k8sResources)
+
 	for k, v := range k8sResources {
 		if !existingRes[string(k)] && v != "" {
 			res[string(k)] = v
 		}
 	}
-
-	for k, v := range pod.GetAnnotations() {
```
jaronoff97, line 499: as far as precedence goes, I do think we want this to remain last right? i.e. if someone is explicitly setting `"resource.opentelemetry.io/service.name": "my-service"` I think that should have final say. Otherwise, this would be a breaking change

jaronoff97 (commented):
another thought, also if we could add something in to an existing e2e test that would be great :) thank you! FWIW, i do think I will be taking on a refactor to make this logic a bit more straightforward in the future.

pkg/instrumentation/sdk.go
zeitlinger: right - also for the labels

zeitlinger: done

zeitlinger:
> another thought, also if we could add something in to an existing e2e test that would be great :) 

done @jaronoff97

pkg/instrumentation/sdk.go
```diff
@@ -488,22 +535,29 @@ func (i *sdkInjector) createResourceMap(ctx context.Context, otelinst v1alpha1.I
...
 		}
 	}
 
-	for k, v := range pod.GetAnnotations() {
-		if strings.HasPrefix(k, constants.OtelAnnotationNamespace) {
-			key := strings.TrimSpace(strings.TrimPrefix(k, constants.OtelAnnotationNamespace))
-			if _, ok := res[key]; !ok {
-				res[key] = v
+	// attributes and labels from the pod have the highest precedence (except for values set in environment variables)
+	for k, v := range pod.Annotations {
```
jaronoff97, line 548: why not use the method call here?

pkg/instrumentation/sdk.go
```diff
@@ -488,22 +535,29 @@ func (i *sdkInjector) createResourceMap(ctx context.Context, otelinst v1alpha1.I
...
+		if strings.HasPrefix(k, constants.ResourceAttributeAnnotationPrefix) {
+			k = strings.TrimPrefix(k, constants.ResourceAttributeAnnotationPrefix)
```
jaronoff97, line 550: we shouldn't modify the iteration variable here

pkg/instrumentation/sdk.go
```diff
@@ -488,22 +535,29 @@ func (i *sdkInjector) createResourceMap(ctx context.Context, otelinst v1alpha1.I
...
+			if !existingRes[k] && k != string(semconv.ServiceNameKey) {
```
jaronoff97, line 551: don't we want to let the override happen here?

pkg/instrumentation/sdk.go
zeitlinger: which method call?

pkg/instrumentation/sdk.go
zeitlinger: we treat the user defined enviroment variables as the highest priority - otherwise it would be a breaking change.
Not sure if it would also be surprising.

https://github.com/zeitlinger/opentelemetry-operator/blob/c1966659f203516c87991715c95f1dc00a372aa1/README.md?plain=1#L783

zeitlinger:
@jaronoff97 please check again

pkg/instrumentation/sdk.go
jaronoff97: `pod.GetAnnotations()`

pkg/instrumentation/sdk_test.go
```diff
@@ -259,7 +529,7 @@ func TestSDKInjection(t *testing.T) {
 							Env: []corev1.EnvVar{
 								{
 									Name:  "OTEL_SERVICE_NAME",
-									Value: "explicitly_set",
+									Value: "explicit-name",
```
jaronoff97, line 532: OOC – why change this for the test?

jaronoff97 (approved):
two minor questions, neither blocking. overall i think this looks good. Going to wait on @TylerHelmuth's review before merging.

pkg/instrumentation/sdk_test.go
zeitlinger: to make sure it's coming from the right place

tests/e2e/env-vars/00-install-instrumentation.yaml
```diff
@@ -4,4 +4,7 @@ metadata:
   name: sdk-only
 spec:
   exporter:
-    endpoint: http://localhost:4317
\ No newline at end of file
+    endpoint: http://localhost:4317
+  defaults:
+    useLabelsForResourceAttributes: true
```
TylerHelmuth, line 9: Please add a new test with this flag set, I'd like to preserve the test that uses the existing behavior

zeitlinger: done @TylerHelmuth

TylerHelmuth (approved):

swiatekm (approved):

jaronoff97:
@zeitlinger thank you for your contribution! 🙇

PR #3206, opened 2024-08-07
Uses the new parsing structure for RBAC parsing

jaronoff97:
**Description:** Uses the new parsing structure for getting parsing rules

- Resolves: n/a

**Testing:** Many unit tests

**Documentation:** n/a


iblancasa:
Is there anything I can do to help you with this?

internal/components/component.go
```diff
@@ -34,37 +35,44 @@ var (
...
+type Option[T any] struct {
 	protocol    corev1.Protocol
 	appProtocol *string
 	targetPort  intstr.IntOrString
 	nodePort    int32
 	name        string
 	port        int32
+	portParser  PortParser[T]
+	rbacGen     RBACRuleGenerator[T]
 }
```
jaronoff97, line 60: Note: I need the generics for the parsing stuff, but it really crowds the type signatures. (see the receivers/helpers.go file) I could remove the functions from the options struct, and put it in the constructor, but that feels counter intuitive to the options struct.

internal/components/component.go
```diff
@@ -75,22 +83,32 @@ func (o *Option) GetServicePort() *corev1.ServicePort {
 	}
 }
 
-type PortBuilderOption func(*Option)
+func WithRBACRuleGenerator[T any](r RBACRuleGenerator[T]) PortBuilderOption[T] {
```
iblancasa, line 86: TBH I don't feel confortable with everything returing "PortBuilder<Something>". For some components we don't have to build anything port related (some processors) but we need to get RBAC. I was not sure about this in the other PR but I though you were thinking about something different and didn't raise my concerns.

jaronoff97: I think i should just rename this to not be "PortBuilderOption" but rather "ParserOption" which is more accurate.

internal/components/component.go
```diff
@@ -34,66 +33,17 @@ var (
...
 
-// PortParser is a function that returns a list of servicePorts given a config of type T.
-type PortParser[T any] func(logger logr.Logger, name string, o *Option, config T) ([]corev1.ServicePort, error)
-
 type PortRetriever interface {
 	GetPortNum() (int32, error)
 	GetPortNumOrDefault(logr.Logger, int32) int32
 }
 
-type Option struct {
```
jaronoff97, line 45: moved this to builder.go

internal/components/generic_parser.go
```diff
@@ -30,8 +31,20 @@ var (
...
+}
+
+func (g *GenericParser[T]) GetRBACRules(logger logr.Logger, config interface{}) ([]rbacv1.PolicyRule, error) {
+	if g.rbacGen == nil {
+		return nil, nil
+	}
+	var parsed T
+	if err := mapstructure.Decode(config, &parsed); err != nil {
+		return nil, err
+	}
```
jaronoff97, line 46: By using generics here, we can let mapstructure decode take care of all the marshalling for us into the exact right type – no more map assertions!

iblancasa (approved):
Great work! Thanks for taking care of this. When do we plan to start using the new infra and remove the old one?

jaronoff97:
this changes to use it for RBAC parsing, hoping to delete the old stuff in a different PR.

internal/components/multi_endpoint.go
```diff
@@ -71,21 +72,46 @@ func (m *MultiPortReceiver) ParserName() string {
 	return fmt.Sprintf("__%s", m.name)
 }
 
-func NewMultiPortReceiver(name string, opts ...MultiPortOption) *MultiPortReceiver {
+func (m *MultiPortReceiver) GetRBACRules(logr.Logger, interface{}) ([]rbacv1.PolicyRule, error) {
+	return nil, nil
+}
+
+type MultiPortBuilder[T any] []Builder[T]
```
swiatekm, line 79: See my previous comments about type parameter naming.

swiatekm (commented):
I'm not that familiar with the component parsing code, so I mostly limited my review to naming and whether the abstractions make sense to me in a vacuum. The test code looks good to be at a glance, but I'd feel more assured if someone with more expertise reviewed it before merging.

internal/components/generic_parser.go
```diff
@@ -31,7 +31,7 @@ var (
 // functionality to idempotent functions.
 type GenericParser[T any] struct {
 	name       string
-	option     *Option[T]
+	option     *Settings[T]
```
swiatekm, line 34: This should be named `settings` now, I think.

swiatekm (approved):

Issue #3090, opened 2024-07-01
Enable "enable-multi-instrumentation" flag by default


iblancasa:
### Component(s)

collector, auto-instrumentation

### Is your feature request related to a problem? Please describe.

The features that are enabled by `enable-multi-instrumentation` have been there for a while and it might be a good time to enable them by default.

The only risk I can see is those user workloads using the  `instrumentation.opentelemetry.io/container-names` would not be instrumented when they are restarted or recreated. We can implement a warning as part of the validation webhook + something as part of the upgrade routines in order to replace this annotation with the language-specific ones.

### Describe the solution you'd like

.

pavolloffay:
@swiatekm @jaronoff97 any objections?

swiatekm:
As far as I can see, this would be a breaking change, as the flag actually changes the behaviour, and different annotations must be used for multi-container instrumentation for the same language. Can we not avoid this breaking change, or at least allow the old annotations to work while the multi-instrumentation flag is enabled?

pavolloffay:
@iblancasa can we write an upgrade procedure to migrate already injected agents?

iblancasa:
> . Can we not avoid this breaking change, or at least allow the old annotations to work while the multi-instrumentation flag is enabled?
@swiatekm With a migration procedure should be enough. But I agree that it would be better if we could make the `instrumentation.opentelemetry.io/container-names` annotation work when the flag is enabled. I'll try to send a PR to make it work when the annotation is there (I'm not sure if that would be considered a breaking change too).

> @iblancasa can we write an upgrade procedure to migrate already injected agents?
@pavolloffay It is what I proposed in my first comment:
```
We can implement a warning as part of the validation webhook + something as part of the upgrade routines in order to replace this annotation with the language-specific ones.
```
I'll create a PR to address this.

artem-nefedov:
It seems that recent changes broke something, check https://github.com/open-telemetry/opentelemetry-operator/pull/3213#issuecomment-2383700953

iblancasa:
https://github.com/open-telemetry/opentelemetry-operator/pull/3213#issuecomment-2383700953

@artem-nefedov can you provide more information like:
1. Is it multiinstrumentation enabled in your environment?
2. Any log messages from the operator?
3. Is the `initContainer` created?

iblancasa:
[...]

PR #3213, opened 2024-08-12
Not ignore the container-names annotation when the multi-instrumentation is enabled

iblancasa:
**Description:** previously, the `container-names` annotation was ignored when multi-instrumentation was enabled in the operator (even if for that Instrumentation object a single instrumentation was used). Now, it is not ignored.

If that annotation is used simultaneously with the ones used for multi-instrumentation, the `areContainerNamesConfiguredForMultipleInstrumentations` will raise an error.

- Relates: #3090

This is the first step to enable the multi-instrumentation by default.


pkg/instrumentation/podmutator.go
```diff
@@ -329,36 +298,27 @@ func (pm *instPodMutator) Mutate(ctx context.Context, ns corev1.Namespace, pod c
...
+		insts.Java.Containers = appendContainers(generalContainerNames, ns.ObjectMeta, pod.ObjectMeta, annotationInjectJavaContainersName)
+		insts.NodeJS.Containers = appendContainers(generalContainerNames, ns.ObjectMeta, pod.ObjectMeta, annotationInjectNodeJSContainersName)
+		insts.Python.Containers = appendContainers(generalContainerNames, ns.ObjectMeta, pod.ObjectMeta, annotationInjectPythonContainersName)
+		insts.DotNet.Containers = appendContainers(generalContainerNames, ns.ObjectMeta, pod.ObjectMeta, annotationInjectDotnetContainersName)
+		insts.Go.Containers = appendContainers(generalContainerNames, ns.ObjectMeta, pod.ObjectMeta, annotationInjectGoContainersName)
+		insts.ApacheHttpd.Containers = appendContainers(generalContainerNames, ns.ObjectMeta, pod.ObjectMeta, annotationInjectApacheHttpdContainersName)
+		insts.Nginx.Containers = appendContainers(generalContainerNames, ns.ObjectMeta, pod.ObjectMeta, annotationInjectNginxContainersName)
+		insts.Sdk.Containers = appendContainers(generalContainerNames, ns.ObjectMeta, pod.ObjectMeta, annotationInjectSdkContainersName)
```
swiatekm, line 314: I would feel a bit better if we operated on slices of strings here instead of a single string containing a comma separated list of values. Do you think we could do that without making the code more complicated?

swiatekm (commented):
LGTM, with one nitpick.

pkg/instrumentation/podmutator.go
iblancasa: Yes. I don't like the current "string list" too much. Let me check if I can do something to improve it...

.chloggen/feature_3090.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: auto-instrumentation
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Not ignore the container-names annotation when the multi-instrumentation is enabled"
```
pavolloffay, line 8: Please put here the entire annotation or in the subtext. It will be clearer for end users.

pavolloffay:
@iblancasa could we deprecate/remove `--enable-multi-instrumentation` flag so that it is enabled by default and users can combine annotations?

pavolloffay (commented):
Perhaps some documentation needs to be updated?

iblancasa:
> @iblancasa could we deprecate/remove `--enable-multi-instrumentation` flag so that it is enabled by default and users can combine annotations?

I'll do it as part of other PR because I don't want this one to be really big. Just to make it easy to review.

iblancasa:
> Perhaps some documentation needs to be updated?

I'm not sure. I read the `README.md` file and I think it should be ok but I'm open to make changes on that. Please let me know what do you think it should be included.

pkg/instrumentation/podmutator.go
iblancasa: @swiatekm I added some changes to improve how this is managed (and avoid some `splits`, `joins` in multiple parts of the code). I think it looks better now.

iblancasa:
Checking the errors in the E2E tests...

pkg/instrumentation/helper.go
```diff
@@ -133,3 +129,32 @@ func volumeSize(quantity *resource.Quantity) *resource.Quantity {
...
+	if err != nil {
+		return fmt.Errorf("error while checking for instrumentation container annotations %w", err)
+	}
+	if !matched {
+		return fmt.Errorf("not valid characters included in the instrumentation container annotation %s", containersAnnotation)
+	}
+	return nil
+}
+
+func configureLanguageContainers(inst *instrumentationWithContainers, annotation string, ns metav1.ObjectMeta, pod metav1.ObjectMeta) error {
```
swiatekm, line 148: I think a better name for this would be `setContainersFromAnnotation`. We should also add a docstring here imo.

pkg/instrumentation/podmutator.go
```diff
@@ -190,6 +202,35 @@ func (langInsts *languageInstrumentations) setInstrumentationLanguageContainers(
...
+	}
+	if err := configureLanguageContainers(&langInsts.ApacheHttpd, annotationInjectApacheHttpd, ns, pod); err != nil {
+		return err
+	}
+	if err := configureLanguageContainers(&langInsts.Nginx, annotationInjectNginx, ns, pod); err != nil {
+		return err
+	}
+	if err := configureLanguageContainers(&langInsts.Sdk, annotationInjectSdk, ns, pod); err != nil {
+		return err
+	}
```
swiatekm, line 232: This looks like it could become a loop instead.

swiatekm (approved):
LGTM, some minor nitpicks.

pavolloffay (approved):

artem-nefedov:
This PR may have broken handling of language-specific annotations.

I have deployment with the following pod annotations:

```yaml
      instrumentation.opentelemetry.io/inject-sdk: otel-system/otel-instrumentation
      instrumentation.opentelemetry.io/sdk-container-names: test-app
```

This works fine with versions up to v0.107.0, but after upgrade to v0.108.0 it no longer injects variables.

PR #3222, opened 2024-08-16
[all] add support for Windows nodes

### LINKED ISSUE #642: Windows nodes support
Is there any plan to add Kubernetes Windows nodes support?
We have an application running on Windows nodes and would like to export traces. 
Currently traces from the application are pushed to a Linux node using address hardcode in configuration. 
It would be great to have a DeamonSet with corresponding tolerations to be able to send traces to the same node.

atoulme:
**Description:** 
Add support to build Windows images of the Docker images used by the operator and add them as an additional manifest platform on pushed images.

Fixes #642

**Testing:**
Not much. I however have a k8s environment with Windows nodes to try things in.

### DIFF
--- a/.chloggen/support_windows.yaml
+++ b/.chloggen/support_windows.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: all
+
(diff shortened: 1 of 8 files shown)


swiatekm (commented):
Just building the images doesn't resolve #642. The main blocker for that issue is testing, as we're not comfortable releasing or supporting untested images. As per Jacob's comment [here](https://github.com/open-telemetry/opentelemetry-operator/issues/642#issuecomment-2025671628), for us to accept this you either need to find a way to run tests in a cluster with Windows nodes in our CI, or commit to maintaining such tests externally, similar to what Red Hat does for OpenShift.

We're open to implementing this, but it's unfortunately not that simple.

PR #3234, opened 2024-08-21
Bump k8s dependencies to 1.31.0

swiatekm:
Also bump controller-runtime to 0.19.0. This also required upgrading controller-tools to 0.16.1.

Supersedes #3233.


swiatekm:
I don't really understand what's going on with the `managementState` field here. The field *is* marked as required, so the change to the CRDs is correct, but I'm not sure why it wasn't the case in the first place. The E2E test failure is a bit confusing. `kubectl replace` doesn't apply defaults, so it makes sense that it would fail validation, but a plain `apply` also doesn't work, for reasons which are unclear to me.

config/rbac/role.yaml
```diff
@@ -152,31 +153,7 @@ rules:
...
-  - opentelemetry.io
-  resources:
   - opentelemetrycollectors/finalizers
-  verbs:
-  - get
-  - patch
-  - update
-- apiGroups:
-  - opentelemetry.io
-  resources:
```
iblancasa, line 179: Why are those not needed anymore?

config/crd/bases/opentelemetry.io_targetallocators.yaml
```diff
@@ -3284,6 +3313,8 @@ spec:
                   type: object
                 type: array
                 x-kubernetes-list-type: atomic
+            required:
+            - managementState
```
iblancasa, line 3317: Does this mean that users will always need to provide the `.spec.managementState` field when creating an `OpenTelemetryCollector`?

config/rbac/role.yaml
swiatekm: This section was merged with the one for Instrumentations. I guess the new version of controller-tools is better at arranging RBAC rules.

config/crd/bases/opentelemetry.io_targetallocators.yaml
swiatekm: No, `managementState` has a default, and that applies if it's not set by the user. Otherwise, all the E2E tests would fail.

Worth noting that I had to change one test which did a `kubectl replace`. That command does not apply defaults, and in that case you do need to set `managementState` manually. I just changed it into an `apply` instead.

iblancasa (approved):

jaronoff97 (approved):

PR #3240, opened 2024-08-27
[chore] Add runbooks file

bogdan-at-adobe:
Added a runbooks.md file primarily to have an endpoint to point to from [here](https://github.com/bogdan-at-adobe/opentelemetry-helm-charts/blob/7f64f3fad27888839cdf0059b597b5d5f688e318/charts/opentelemetry-operator/templates/prometheusrule.yaml#L49) since it expects a url.

### DIFF
--- a/docs/runbooks.md
+++ b/docs/runbooks.md
@@ -1,0 +1,23 @@
+# OpenTelemetry Operator Runbooks
+
+See [`README.md`](../README.md) for more details about the OpenTelemetry Operator.
+
+## Manager Rules
+
+### [ReconcileErrors](#reconcileerrors)
+
+|||
+|-:|-|
+| Meaning | The OpenTelemetry Operator is unable to succeed in the reconciliaton step, probably because of a missconifgured OpenTelemetryCollector. |
+| Impact | No impact on already running deployments or new correct ones. |
+| Diagnosis | Check manager logs for reasons why this might happen. <br> <br> Example: `failed to create objects for <podName>: Deployment.apps "<deploymentName>" is invalid: spec.template.spec.containers[0].env[1].valueFrom: Invalid value: "": may not be specified when "value" is not empty`|
+| Mitigation | Find out which OpenTelemetryCollector is causing the errors and fix the config. |
+
+### [WorkqueueDepth](#workqueuedepth)
+
+|||
+|-:|-|
+| Meaning | The working queue for the operator is larger than 0. |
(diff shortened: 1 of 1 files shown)


friedrichg (approved):

docs/runbooks.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Manager Rules
+
+### [ReconcileErrors](#reconcileerrors)
+
+|||
+|-:|-|
+| Meaning | The OpenTelemetry Operator is unable to succeed in the reconciliaton step, probably because of a missconifgured OpenTelemetryCollector. |
+| Impact | No impact on already running deployments or new correct ones. |
+| Diagnosis | Check manager logs for reasons why this might happen. <br> <br> Example: `failed to create objects for <podName>: Deployment.apps "<deploymentName>" is invalid: spec.template.spec.containers[0].env[1].valueFrom: Invalid value: "": may not be specified when "value" is not empty`|
```
jaronoff97, line 13: should we include a real example of this? I believe the operator should be preventing most of these already, so having a clear example and how its solved may be useful. (we're also working on preventing these from happening with #2193)

bogdan-at-adobe: That is a real text from an alert we had. Do you mean i should add the OtelCol config that caused the alert?

jaronoff97: oh I hadn't seen this from the operator before so I assumed it was contrived, if it was legit then we can keep it here.

bogdan-at-adobe: I could remove it. Now that I think more of it i feel like it only adds confusion. It would only help someone with that exact problem if i added some more resolution steps for that particular error. The operator logs are descriptive enough to find the problem.

jaronoff97:
@bogdan-at-adobe thanks for the PR here! I _think_ it may make more sense to keep this in the [operator docs on the [opentelemetry.io](http://opentelemetry.io/) site](https://opentelemetry.io/docs/kubernetes/operator/troubleshooting/). That feels like a place that will also be more discoverable for people not just coming from a prometheus rule. If in the future we needed to move the place for the runbooks, the alerts would then be pointing to the wrong location and we have no way to do rerouting from github unfortunately. You can open the PR against [this repo]([github.com/open-telemetry/opentelemetry.io/](http://github.com/open-telemetry/opentelemetry.io/)) Let me know if that makes sense.

bogdan-at-adobe:
@jaronoff97 I agree. I will close this. I have opened a PR in the correct spot. Thank you for the suggestion!

PR #3243, opened 2024-08-29
Move TA config parsing to configmap manifest builder

swiatekm:
**Description:**
This is in preparation for the Target Allocator CRD. For that CRD, the Collector instance may be specified via a label or ownership relation. Therefore, we need to pull the necessary data from the Collector instance during TA config generation. It's more convenient to do this in all cases, which is what this change does.

No current behaviour should change as a result.

**Link to tracking Issue(s):** #2516 

**Testing:**

Added tests for some new helper functions I introduced.

### DIFF
--- a/internal/manifests/targetallocator/configmap.go
+++ b/internal/manifests/targetallocator/configmap.go
@@ -15,6 +15,7 @@
 package targetallocator
 
 import (
+	"github.com/mitchellh/mapstructure"
 	"gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


internal/manifests/targetallocator/configmap.go
```diff
@@ -93,3 +104,78 @@ func ConfigMap(params Params) (*corev1.ConfigMap, error) {
...
+	return collectorGlobalConfig.Object, nil
+}
+
+func getScrapeConfigs(taScrapeConfigs []v1beta1.AnyConfig, collectorConfig v1beta1.Config) ([]v1beta1.AnyConfig, error) {
+	scrapeConfigs := []v1beta1.AnyConfig{}
+
+	// we take scrape configs from both the target allocator spec and the collector config
+	if len(taScrapeConfigs) > 0 {
+		scrapeConfigs = append(scrapeConfigs, taScrapeConfigs...)
+	}
```
jaronoff97, line 127: of note – if someone just creates a targetallocator crd they no longer need to escape scrape configs... this may make it easier for us to use upstream libraries for parsing prometheus in the future.

internal/manifests/targetallocator/configmap.go
```diff
@@ -93,3 +104,78 @@ func ConfigMap(params Params) (*corev1.ConfigMap, error) {
...
+		return nil, err
+	}
+
+	collectorScrapeConfigs, err := getScrapeConfigsFromOtelConfig(configStr)
+	if err != nil {
+		return nil, err
+	}
+	scrapeConfigs = append(scrapeConfigs, collectorScrapeConfigs...)
+
+	return scrapeConfigs, nil
```
jaronoff97, line 140:
```suggestion
	return append(scrapeConfigs, collectorScrapeConfigs...), nil
```

Issue #3261, opened 2024-09-04
awsxray-receiver defaults to TCP protocol instead of UDP


jholm117:
### Component(s)

collector

### Describe the issue you're reporting

The awsxray-receiver is defaulting to TCP instead of UDP.

I will have a PR up for this shortly.

PR #1433, opened 2023-02-01
Default to UDP protocol for statsd receiver

PeterSzegedi:
The statsd receiver's service and deployment should default to UDP 8125 instead of the TCP 8125. 
Currently the collector pod/container is unreachable, when someone tries to spin up a statsd receiver because the pod is listening on the UDP port, but the service and the deployment is exposing TCP.
 
The TCP protocol has not been implemented for the statsd receiver, so having a default to TCP is not useful [link](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/statsdreceiver/receiver.go#L78)

I have built the controller image on my local minikube and it seems to fix the issue.

Let me know how, if any changes are needed to get this merged. 

Thanks,
Peter

PR #3263, opened 2024-09-04
Default to UDP protocol for xray receiver

jholm117:

The awsxray receiver's service and deployment should default to UDP 2000 instead of TCP 2000.
Currently the collector pod/container is unreachable, when someone tries to spin up an xray receiver because the pod is listening on the UDP port, but the service and the deployment is exposing TCP.

From the awsxray-receiver [README](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/awsxrayreceiver#transport-optional):

> This should always be "udp" as X-Ray SDKs only send segments using UDP.

I looked to #1433 for implementation guidance.

Let me know how, if any changes are needed to get this merged.

- Resolves: #3261 


internal/manifests/collector/parser/receiver/receiver_aws-xray.go
```diff
@@ -25,11 +26,12 @@ const parserNameAWSXRAY = "__awsxray"
 // NewAWSXrayReceiverParser builds a new parser for AWS xray receivers, from the contrib repository.
 func NewAWSXrayReceiverParser(logger logr.Logger, name string, config map[interface{}]interface{}) parser.ComponentPortParser {
```
jaronoff97, line 27: this is unused now and will be deleted soon, you can safely remove this

jholm117: Ahh, that explains a lot. I've removed the change, thanks for pointing it out!

jaronoff97 (approved):

PR #3270, opened 2024-09-06
Add arm64 artefacts to dotnet autoinstrumentation container.

Mpdreamz:
This will allow the operator to auto instrument dotnet applications on `arm64` containers. (Including k8s on OSX M1/M2/M3 etc)

Validated manually as part of Elastic's distribution of OpenTelemetry .NET

https://github.com/elastic/elastic-otel-dotnet/pull/155

CC @Kielek & @RassK who did the bulk of the work actually creating the artefacts 🙏

### DIFF
--- a/.chloggen/dotnet-instrumenation-arm64.yml
+++ b/.chloggen/dotnet-instrumenation-arm64.yml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add arm64 artefacts to dotnet autoinstrumentation container. 
+
+# One or more tracking issues related to the change
+issues: [5868]
+
+# (Optional) One or more lines of additional information to render under the primary note.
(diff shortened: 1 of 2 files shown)


frzifus (approved):

Kielek (commented):
1. As I know, there is a possibility to have separate images for architectures (arm64/x64). It could reduce image size to download. I am not telling that it is mandatory for this PR.
2. It will not work without additional changes on the operator side. See code around https://github.com/open-telemetry/opentelemetry-operator/blob/5497a49a541ac49bc992403be14e06469cfb9d3a/pkg/instrumentation/dotnet.go#L48-L51

swiatekm (commented):
It doesn't look like this fully addresses the issue. The operator will still set the `CORECLR_PROFILER_PATH` to the x64 path.

I think a clean way to implement this would be:

1. Make the image multi-platform, changing the path to just `linux`. Add a `linux-x64` symbolic link for backwards compatibility.
2. Use the new path when injecting instrumentation.

For 1, you just need to add the built-in `ARG TARGETARCH` and use this variable in your urls.

Mpdreamz:
Just a note that I will get back to this PR either this or next week.

pl4nty:
I have this working with v0.113 by setting `CORECLR_PROFILER_PATH` manually, and bumping the version to 1.9.0 https://github.com/open-telemetry/opentelemetry-operator/blob/99b6c6fc4c472fd6b54080f9e2737fdf38573ae9/versions.txt#L30

Issue #3267, opened 2024-09-06
Allow references to other volumes (eg. PVCs) for instrumention auto-inject


jnarezo:
### Component(s)

auto-instrumentation

### Is your feature request related to a problem? Please describe.

In an environment where you **can't provision an emptyDir above ~150Mi** (the size of the nodejs instrumentation), it's impossible to use auto-instrumentation, since the emptyDir would be too small to transfer the instrumentation from the injected init container to the main container.

### Describe the solution you'd like

Changing the Instrumentation resource spec to replace `Instrumentation.spec.<lang>.volumeSizeLimit` with something more generic like `Instrumentation.spec.<lang>.volume` that references a [Volume](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#volume-v1-core).

This solution should allow a reference to a pre-existing or ephemeral volume help alleviate this issue.

### Describe alternatives you've considered

I've also considered:
- Allow some switch between emptyDir and a reference to a PVC
  - Not as flexible as just allowing the user to specify a k8s **core** spec
- Forcing user to change their environment
  - Operator is responsible for creating the volume to transfer instrumentation, it makes more sense for it to be customizable/solved at this level

jnarezo:
I don't mind tackling this myself + opening a PR, but would like to hear your opinions/suggestions for the overall feature. 🙂

iblancasa:
> Changing the Instrumentation resource spec to replace Instrumentation.spec.<lang>.volumeSizeLimit with something more generic like Instrumentation.spec.<lang>.volume that references a [Volume](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#volume-v1-core).

I think those should be two independent configurations. Not a replacement.

>Forcing user to change their environment
Operator is responsible for creating the volume to transfer instrumentation, it makes more sense for it to be customizable/solved at this level

It is more or less what the operator is currently doing, right? It creates the Volumen to inject the instrumentation.

I don't get the `Forcing user to change their environment`, can you give context?

jnarezo:
> I don't get the Forcing user to change their environment, can you give context?

Sorry, I meant to say that I considered these alternative **workarounds**, however I don't think they're good solutions.

With my second bullet point, a workaround could be "forcing the user to change their environment," which means forcing the user to work with their cluster admins to change policies and increase the emptyDir size limit policy.

However, that's not really a solution so that's why I opened this issue.

---

[...]

PR #3285, opened 2024-09-12
Add user-specified instrumentation volume

jnarezo:
**Description:**

Adds the ability for user to specify a custom volume for instrumentation / auto-injection to `Instrumentation.spec.<lang>.volumeSizeLimit`.

Adds validation to prevent `Volume` and `VolumeSizeLimit` being defined simultaneously to prevent overlapping behavior (ex. Do we use `Volume.Size` or `VolumeSizeLimit`?)

- Resolves: #3267 

**Testing:**

Unit tests for validation with different combinations of Volume/VolumeSizeLimit being defined and not defined

**Documentation:**

Updated API docs with new spec.


pkg/instrumentation/helper.go
```diff
@@ -123,6 +124,22 @@ func isInstrWithoutContainers(inst instrumentationWithContainers) int {
 	return 0
 }
 
+func instrVolume(volume corev1.Volume, name string, volumeSizeLimit *resource.Quantity) (corev1.Volume, error) {
```
swiatekm, line 127: Could you add a docstring for this function?

pkg/instrumentation/helper.go
```diff
@@ -123,6 +124,22 @@ func isInstrWithoutContainers(inst instrumentationWithContainers) int {
...
+	if !reflect.ValueOf(volume).IsZero() && volumeSizeLimit != nil {
+		return volume, fmt.Errorf("both Volume and VolumeSizeLimit cannot be defined simultaneously")
+	} else if !reflect.ValueOf(volume).IsZero() {
+		return volume, nil
+	}
```
swiatekm, line 132: This check should happen in the instrumentation admission webhook. Here, during Pod injection, we should use the highest priority setting instead of returning an error, in my opinion. In this case, use the volume if it's set, otherwise use the emptydir.

swiatekm (commented):
These changes look good to me overall. Some things missing before we can merge this PR:

* You need to add a changelog entry. Run `make chlog-new` and fill in the template.
* Run `make precommit` and fix any issues it brings up. Right now, the CI is failing because the bundle needs to be updated after changes to CRDs.
* We should check this volume mechanism in E2E tests. It's fine to modify an existing test.

pkg/instrumentation/helper.go
jnarezo: Added docstring

pkg/instrumentation/helper.go
jnarezo: Right, I definitely agree. Moved validation logic to webhook + slimmed down helper logic.

jnarezo:
> These changes look good to me overall. Some things missing before we can merge this PR:
> 
> * You need to add a changelog entry. Run `make chlog-new` and fill in the template.
> * Run `make precommit` and fix any issues it brings up. Right now, the CI is failing because the bundle needs to be updated after changes to CRDs.
> * We should check this volume mechanism in E2E tests. It's fine to modify an existing test.

Thanks, got it working locally! Was using golang v1.23.x and pipeline was broken on me, so I downgraded.

swiatekm:
The PR looks great, thanks for the contribution!

There's one big problem with it, and to be clear, this is really Kubernetes' fault. As it turns out, the OpenAPI spec for `corev1.Volume` is actually quite large, and we're including a copy of it for each language instrumentation. As a result, we're making the Instrumentation CRD *much* larger. There are consequences to this - for example, if the CRD is too big to fit in a K8s annotation (the limit is ~256kb), then it can't be applied using `kubectl apply` anymore. Your change pushes the Instrumentation CRD above that limit.

I'm not sure what we can do to alleviate this. We *could* make the `Volume` attribute common to all instrumentations. It would make it less flexible, but maybe that's ok? @jaronoff97 @pavolloffay WDYT?

jnarezo:
Hey sorry I dropped this for a bit, but I saw that there was no response on your question.

I redesigned the feature to accept a `volumeClaimTemplate` which is restricted to a **generic ephemeral volume** to match the solution proposed in the issue.

Although more restrictive that originally planned, I think this still provides about the functionality as an emptyDir and will still allow users with emptyDir restrictions to use a different type of volume, supported by storage drivers.

I think this should work for now. An alternative could be using an existing PVC, but there may be issues with files already existing from previous autoinstrumentation so I did not pursue it further.

What do you think?

jnarezo:
Also, restricting the volume allowed me to get the manifest down to ~80KB.

swiatekm:
@jnarezo this looks like a reasonable solution to me. Pending any more fundamental changes to the CRD, where we perhaps move all of these common values to the top level, this looks like a good compromise. WDYT @jaronoff97 @pavolloffay ?

Looks like you have some conflicts and failing e2e tests, could you have a look at that?

swiatekm (approved):

jnarezo:
> @jnarezo this looks like a reasonable solution to me. Pending any more fundamental changes to the CRD, where we perhaps move all of these common values to the top level, this looks like a good compromise. WDYT @jaronoff97 @pavolloffay ?
> 
> Looks like you have some conflicts and failing e2e tests, could you have a look at that?

Yeah! Sorry, looks like my e2e differed a bit locally. Thanks for working with me on this.

jaronoff97 (approved):
just one thought, otherwise this looks good. thank you :bow:

apis/v1alpha1/instrumentation_types.go
```diff
@@ -278,6 +298,10 @@ type ApacheHttpd struct {
 	// +optional
 	Image string `json:"image,omitempty"`
 
+	// VolumeClaimTemplate defines a ephemeral volume used for auto-instrumentation.
+	// If omitted, an emptyDir is used with size limit VolumeSizeLimit
+	VolumeClaimTemplate corev1.PersistentVolumeClaimTemplate `json:"volumeClaimTemplate,omitempty"`
+
 	// VolumeSizeLimit defines size limit for volume used for auto-instrumentation.
 	// The default size is 200Mi.
 	VolumeSizeLimit *resource.Quantity `json:"volumeLimitSize,omitempty"`
```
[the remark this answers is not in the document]
swiatekm: Let's open a separate issue for that, I don't think it should be within the scope of this PR.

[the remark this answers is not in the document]
swiatekm: https://github.com/open-telemetry/opentelemetry-operator/issues/3382

PR #3286, opened 2024-09-12
Improve OpenTelemetry Collector JSON Marshalling

### LINKED ISSUE #3281: Wrong marshal for OpenTelemetryCollector
## Description
I'm working on something using the OpenTelemetry Collector API provided by this repository and I found some configuration fields were not marshaled properly.

So, for an OpenTelemetry Collector like this:
```yaml
kind: OpenTelemetryCollector
metadata:
  name: instance
  namespace: open-cluster-management-observability
spec:
  config:
    exporters:
      debug:
    processors:
    receivers:
      jaeger:
        protocols:
            grpc:
      otlp:
        protocols:
          grpc:
          http:
    service:
      pipelines:
        traces:
          exporters:
            - debug
          processors: []
          receivers:
            - otlp
            - jaeger
  mode: deployment
```

I can see the `protocols` field from the `otlp` receiver is dropped.

If we try directly in a script, it doesn't reproduce. But it can be reproduced by creating the resource first in Kubernetes, retrieving it, and doing the marshaling.

This seems to be expected from the changes introduced in https://github.com/open-telemetry/opentelemetry-operator/pull/2444

iblancasa:
Fixes #3281


.chloggen/3281-marshal-config.yaml
```diff
@@ -0,0 +1,16 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix Configuration marshalling for v1beta1 OpenTelemetryCollector instances.
+
+# One or more tracking issues related to the change
+issues: [3281]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
pavolloffay, line 16: Could you please explain what has changed in details here? 

Eg. fragment of a CR that was submitted to cluster and what will be stored in etcd and retrieved again by kubectl.

apis/v1beta1/config.go
```diff
@@ -88,11 +88,24 @@ func (c *AnyConfig) UnmarshalJSON(b []byte) error {
...
+
+	for k, v := range c.Object {
+		if v == nil {
+			nonNilMap[k] = nil
+		} else if emptyMap, ok := v.(map[string]interface{}); ok && len(emptyMap) == 0 {
+			nonNilMap[k] = emptyMap
+		} else {
+			nonNilMap[k] = v
+		}
+	}
```
swiatekm, line 106: It seems to me that this just makes a shallow copy of `c.Object`. Does that actually change anything?

apis/v1beta1/config.go
```diff
@@ -238,6 +251,21 @@ func (c *Config) Yaml() (string, error) {
...
+		return nil, err
+	}
+	return json.Marshal(&struct {
+		*Alias
+		Receivers json.RawMessage `json:"receivers,omitempty"`
+	}{
+		Alias:     (*Alias)(&c),
+		Receivers: receiversJSON,
+	})
+}
```
swiatekm, line 267: Why are receivers singled out here? Shouldn't this issue potentially apply to all component kinds?

swiatekm (commented):
I don't understand what is actually going on in this PR. From the tests, it seems like everything works correctly, but I don't know why. I've left some comments asking questions, but in general I think you should exhaustively explain the weird marshaling functions in docstrings.

iblancasa:
> I don't understand what is actually going on in this PR. From the tests, it seems like everything works correctly, but I don't know why.

I tried to explain it in the changelog:
>   Previously, the v1beta1 OpenTelemetryCollector CR's config field was incorrectly
  marshalled, potentially causing issues when storing in etcd or retrieving via kubectl.
  This fix ensures the config is properly marshaled, maintaining the correct
  structure and data types. For example, a config with nested maps and arrays will now
  be correctly preserved when stored and retrieved, instead of being flattened or
  losing structure.

The problem happens when you want to create one OpenTelemetry Collector instance like this one:
```yaml
kind: OpenTelemetryCollector
metadata:
  name: otel
spec:
    config:
        receivers:
            otlp:
                http:
                grpc:
```
Then, you create it in the cluster with, for instance, `kubectl`.
```sh
kubectl apply -f otelcol.yaml
```
When you get the OpenTelemetry Collector instance, you will get this:
```sh
$ kubectl get otelcol -o yaml otel
kind: OpenTelemetryCollector
metadata:
  name: otel
spec:
    config:
        receivers:
            otlp:
```

We have a warning mentioning this can happen, but it can be difficult for users to see.

I'll add more info to the changelog.

> I've left some comments asking questions, but in general I think you should exhaustively explain the weird marshaling functions in docstrings.

I just added some comments to the source code.

apis/v1beta1/config.go
```diff
@@ -280,6 +286,27 @@ func (c *Config) Yaml() (string, error) {
 	return buf.String(), nil
 }
 
+// MarshalJSON marshalls the Config field.
+func (c Config) MarshalJSON() ([]byte, error) {
+	// When you marshal a struct that embeds itself or a type that directly or indirectly refers back to itself,
+	// Go's JSON marshaling can lead to infinite recursion. To avoid this, we create an alias of the struct
+	// (Config), so Go no longer considers it the same type.
+	// This allows you to embed the struct safely within itself without triggering that infinite loop.
```
swiatekm, line 294: So the problem is that `Receivers` can contain `Config`?

apis/v1beta1/config.go
```diff
@@ -90,11 +90,17 @@ func (c *AnyConfig) UnmarshalJSON(b []byte) error {
...
 // MarshalJSON specifies how to convert this object into JSON.
-func (c *AnyConfig) MarshalJSON() ([]byte, error) {
-	if c == nil {
+func (c AnyConfig) MarshalJSON() ([]byte, error) {
+	if c.Object == nil {
 		return []byte("{}"), nil
 	}
-	return json.Marshal(c.Object)
+
+	nonNilMap := make(map[string]interface{}, len(c.Object))
```
jaronoff97, line 98: at this point don't we know that c.Object is nonNil?

iblancasa:
> I'm still not entirely sure how or why this works... maybe we can talk about it in slack?

Just messaged you.

iblancasa:
The E2E tests are failing because there was a problem while downloading a dependency or something.

iblancasa:
@jaronoff97 @swiatekm do you have any other questions or concerns regarding this PR?

.chloggen/3281-marshal-config.yaml
```diff
@@ -0,0 +1,38 @@
...
+      otlp:
+        protocols:
+          grpc:
+          http:
+
+  Would be retrieved from the Kubernetes cluster as:
+
+  receivers:
+      otlp:
+        protocols:
```
jaronoff97, line 36: i actually just ran into this and was very confused 😃

apis/v1beta1/config.go
```diff
@@ -280,6 +286,27 @@ func (c *Config) Yaml() (string, error) {
...
 
+// MarshalJSON marshalls the Config field.
+func (c Config) MarshalJSON() ([]byte, error) {
+	// When you marshal a struct that embeds itself or a type that directly or indirectly refers back to itself,
+	// Go's JSON marshaling can lead to infinite recursion. To avoid this, we create an alias of the struct
+	// (Config), so Go no longer considers it the same type.
+	// This allows you to embed the struct safely within itself without triggering that infinite loop.
+	type Alias Config
+	// Ensure we call the custom marshaller for AnyConfig for the Receivers
+	receiversJSON, err := json.Marshal(c.Receivers)
```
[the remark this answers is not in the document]
jaronoff97: @iblancasa bump on this

apis/v1beta1/config_test.go
```diff
@@ -26,11 +30,227 @@ import (
...
+	require.NoError(t, err)
+	defer func() {
+		errStop := testEnv.Stop()
+		require.NoError(t, errStop)
+	}()
+
+	k8sClient, err := client.New(cfg, client.Options{Scheme: testScheme})
+	if err != nil {
+		fmt.Printf("failed to setup a Kubernetes client: %v", err)
+		os.Exit(1)
```
jaronoff97, line 62: ive found [this fake client package](https://github.com/open-telemetry/opentelemetry-operator/blob/7ee69c59d6727bddf8b1af71da544eac7df08a9b/cmd/operator-opamp-bridge/operator/client_test.go#L46-L59) a lot more usable

tests/e2e-targetallocator/targetallocator-features/00-assert.yaml
```diff
@@ -20,7 +20,7 @@ spec:
           items:
           - key: collector.yaml
             path: collector.yaml
-          name: stateful-collector-85dbe673
+          name: stateful-collector-57be076a
```
jaronoff97, line 23: OOC why are the hashes changing? id be interested to understand how the full content is being changed.

jaronoff97:
@swiatekm reference https://medium.com/picus-security-engineering/custom-json-marshaller-in-go-and-common-pitfalls-c43fa774db05

iblancasa:
Closing this PR because the changes introduced in #3333 workaround the problem.

Since it seems it was a problem for so many people to follow what's going on this PR, I'll add an explanation here just as a reference.

## The issue

After creating an `OpenTelemetryCollector` instance from a YAML file like this one:
```yaml
kind: OpenTelemetryCollector
metadata:
  name: otel
spec:
    config:
        receivers:
            otlp:
                protocols:
                    http:
                    grpc:
```

If you get the `OpenTelemetryCollector` instance from Kubernetes:
```sh
kubectl get otelcol -o yaml otel
```
this is what you get:
```yaml
kind: OpenTelemetryCollector
metadata:
  name: otel
spec:
    config:
        receivers:
            otlp:
                protocols:
```
The values for `protocols` are missing. This does not just apply to `otlp` receiver but to all those configurations with a key but not a value -like what happens for `grpc` or `http` in the previous example.

This is not supported by other tools like `kustomize` but, if you are an OpenTelemetry Collector user, you would expect to be able to use the same kind of config that you would use with the regular Go binary.

The operator shows a warning in this situation.

## The fix
The fix is based on [this blog article]( https://medium.com/picus-security-engineering/custom-json-marshaller-in-go-and-common-pitfalls-c43fa774db05).

I was implementing the `MarshalJSON` for the `OpenTelemetry`, `OpenTelemetrySpec` and `Config` structs to ensure the custom `MarshalJSON` method from `AnyConfig` would be called -because it seems in some scenarios the custom `MarshalJSON` implementations are not called when one struct is part of another, I don't get why-. I was forcing the call to the custom `MarshalJSON` only for receivers because are the only kind of components where this situation can be reproduced. Why? Because receiver components are the only ones that contain configurations where keys without values are valid.

I added a unit test to ensure this situation, which seems to be reproducible only when the object is created + retrieved in Kubernetes, didn't reproduce again.

## The unexpected workaround
With  #3333, all those keys that would not have any value, have now a value by default. So, the issue is there but users will not suffer it anymore.

swiatekm:
@iblancasa I think I finally understand, thanks for being so patient with us! Could we actually keep the tests you've written, and skip them for the time being? We've been talking about alternative solutions to this issue, and the tests would be very helpful as benchmarks. I'd also hate for your hard work to go to waste this way.

Issue #1934, opened 2023-07-18
PrometheusSpec configuration for Target Allocator


swiatekm:
Target Allocator can generate Prometheus configurations based on prometheus-operator Custom Resources, and it does so using code imported from prometheus-operator. In order to generate this configuration, the monitoring CRs like ServiceMonitor aren't sufficient, as some of the configuration lives in the Prometheus CR. The default scrape interval is a good example, see #1925.

We should decide which fields from the Prometheus CR we want to support in the Target Allocator spec. There's quite a few fields in https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#commonprometheusfields, and most of them either don't apply to Target Allocator, or do apply, but are already set in the Collector spec.

The purpose of this issue is to track which fields we should add, and their implementation status.

- [x] `serviceMonitorSelector`
- [x] `serviceMonitorNamespaceSelector`
- [x] `podMonitorSelector`
- [x] `podMonitorNamespaceSelector`
- [ ] `scrapeTimeout`
- [ ] `overrideHonorLabels`
- [ ] `overrideHonorTimestamps`
- [ ] `ignoreNamespaceSelectors`
- [ ] `enforcedNamespaceLabel`
- [ ] `enforcedSampleLimit`
- [ ] `enforcedTargetLimit`
- [ ] `enforcedLabelLimit`
- [ ] `enforcedLabelNameLengthLimit`
- [ ] `enforcedLabelValueLengthLimit`
- [ ] `enforcedBodySizeLimit`
- [ ] `minReadySeconds`
- [ ] `excludedFromEnforcement`
- [ ] `podTargetLabels`

swiatekm-sumo:
An initial list of what seems applicable:

- `serviceMonitorSelector`
- `serviceMonitorNamespaceSelector`
- `podMonitorSelector`
- `podMonitorNamespaceSelector`
- `scrapeTimeout`
- `overrideHonorLabels`
- `overrideHonorTimestamps`
- `ignoreNamespaceSelectors`
- `enforcedNamespaceLabel`
- `enforcedSampleLimit`
- `enforcedTargetLimit`
- `enforcedLabelLimit`
- `enforcedLabelNameLengthLimit`
- `enforcedLabelValueLengthLimit`
- `enforcedBodySizeLimit`
- `minReadySeconds`
- `excludedFromEnforcement`
- `podTargetLabels`

jaronoff97:
[...]

PR #3292, opened 2024-09-16
Assert protocol is propagated

jaronoff97:
**Description:** 
This is a new check to assert that we are indeed setting the protocol globally in the generated scrape config file

**Link to tracking Issue(s):** n/a

### DIFF
--- a/tests/e2e-targetallocator/targetallocator-prometheuscr/00-install.yaml
+++ b/tests/e2e-targetallocator/targetallocator-prometheuscr/00-install.yaml
@@ -136,6 +136,9 @@ spec:
     receivers:
       prometheus:
         config:
+          global:
+            scrape_interval: 30s
+            scrape_protocols: ['PrometheusProto','OpenMetricsText1.0.0','OpenMetricsText0.0.1','PrometheusText0.0.4']
           scrape_configs: []
 
     processors:
--- a/tests/e2e-targetallocator/targetallocator-prometheuscr/01-install.yaml
+++ b/tests/e2e-targetallocator/targetallocator-prometheuscr/01-install.yaml
@@ -45,7 +45,9 @@ spec:
           args:
             - /bin/sh
             - -c
-            - curl -s http://prometheus-cr-targetallocator/scrape_configs | grep "prometheus-cr"
+            - >
+              curl -s http://prometheus-cr-targetallocator/scrape_configs |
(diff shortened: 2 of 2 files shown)


cmd/otel-allocator/config/config.go
```diff
@@ -58,6 +59,79 @@ type Config struct {
...
+		return prom
+	}
+
+	// the prometheus-operator provides no automatic conversion for global fields
+	if len(cfg.PromConfig.GlobalConfig.ScrapeProtocols) > 0 {
+		var scrapeProtocols []monitoringv1.ScrapeProtocol
+		for _, protocol := range cfg.PromConfig.GlobalConfig.ScrapeProtocols {
+			scrapeProtocols = append(scrapeProtocols, monitoringv1.ScrapeProtocol(protocol))
+		}
+		prom.Spec.CommonPrometheusFields.ScrapeProtocols = scrapeProtocols
```
swiatekm, line 89: I don't think this is the correct thing to do. In my view, `GlobalConfig` should only affect the raw scrape configs, while the respective `Prometheus` fields (which only affect prometheus-operator CRs) should be separately configured.

The ambiguity of how `GlobalConfig` should affect the prometheus-operator world is why I was reluctant to include it in the first place.

jaronoff97: I can see that, however, i think the confusion / importance of doing it this way is that we don't have the luxury of having it automatically propagated to the prometheus instance. i.e. when a prometheus instance sets the global config and is being configured via the prometheus operator, any scrape configs generated by the prometheus CRDs to be added to the prometheus instance will be using the global config defined by said prometheus instance. 

As I was writing this, i was wondering if this is necessary given it's the collector that's doing the scraping. So if a user sets the global config on the prometheus receiver in the collector, shouldn't that be used when scraping a target overriding the scrape_configs received by the TA?

jaronoff97: I need to look into this, because if that's the case this would be unnecessary right? Otherwise, I _do_ think we need this.

jaronoff97: yeah that makes sense to me, and would probably better in the long run anyway. I had to look through some more code to make sense of this, but I agree that this is probably the way to go.

jaronoff97:
Closing this in favor of the resolution of #1934

Issue #3279, opened 2024-09-11
Tracking issues apis/v1alpha1 Sampling CR


frzifus:
### Describe the issue you're reporting

https://docs.google.com/document/d/1neyY2bsDmu6gsnrkwnNfnLXvU1UfsJ_am20mNWSGKQY/edit

PR #3297, opened 2024-09-20
rfc: Opinionated OpenTelemetry Operator Sampling CR

frzifus:

- Belongs to: #3279

https://github.com/frzifus/opentelemetry-operator/blob/rfc/sampling/docs/rfcs/opinionated_sampling_cr.md


docs/rfcs/opinionated_sampling_cr.md
```diff
@@ -0,0 +1,140 @@
...
+    # By default, the size is 0 and the cache is inactive.
+    # If using, configure this as much higher than num_traces so decisions for
+    # trace IDs are kept longer than the span data for the trace.
+    sampled_cache_size: 10000
+  # Components defines the template of all requirements to configure scheduling
+  # of all components to be deployed.
+  components:
+    loadbalancer:
+      # Defines if the component is managed by the operator
+      managementState: managed
```
iblancasa, line 97: Why is it needed to set this field at the component level instead of just doing it at the full object level?

frzifus: Actually Ive no strong opinion on that.

docs/rfcs/opinionated_sampling_cr.md
```diff
@@ -0,0 +1,140 @@
...
+The introduction of this CRD is intended to significantly simplify trace sampling in a kubernetes enviroment.
+
+## Summary
+
+Provide a sampler CR in v1alpha1 that can be used to simplify the sampling of traces in kubernetes. 
+
+## Goals and non-goals
+
+**Goals**
+- Provide an opinionated CR to simply sampling configuration in distributed environments
```
jaronoff97, line 22:
```suggestion
- Provide an opinionated CR to simplify sampling configuration in distributed environments
```

docs/rfcs/opinionated_sampling_cr.md
```diff
@@ -0,0 +1,140 @@
...
+
+As a cluster administrator I want to reduce the amount of traffic caused by generating telemetry data.
+
+### CASE 2
+
+As a cluster administrator I want to be in control of the collector resources while allowing a user to change sampling policies.
+
+### CASE 3
+
+As a user I want to be able to filter relevant data without much specific open telemetry knowledge.
```
jaronoff97, line 44:
```suggestion
As a user I want to be able to filter relevant data without much specific OpenTelemetry knowledge. 
```

docs/rfcs/opinionated_sampling_cr.md
```diff
@@ -0,0 +1,140 @@
...
+
+As a user I want to be able to filter relevant data without much specific open telemetry knowledge. 
+
+## Struct Design
+
+This proposal introduces a new CR which serve as an abstraction setting up an opentelemetry collector based sampling layer. Similar to the following figure.
+
+![sampling arch](./images/arch_sampling_crd.png)
+
+This custom resource creates an environment that allows us to apply e.g. tailbased sampling in a distributed environment. The operator takes care of creating an optional otel LB service and sampling instances similar to the figure shown above.
```
jaronoff97, line 52: Is there a reason we prefer tail sampling in the collector versus a global sampling policy that we apply to instrumentations? I guess not everything will be otel instrumented...

docs/rfcs/opinionated_sampling_cr.md
```diff
@@ -0,0 +1,140 @@
...
+
+LB instances will be pre-configured to distribute traces based on a given routing key like the traceID to the sampler instances.
+
+A policy is used to define which telemetry data should be sampled. Available policies can be found in the [tailbased sampling description](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/tailsamplingprocessor).
```
jaronoff97, line 56: i also want us to be able to support the probabilistic sampler which requires _no_ sticky load balancing configuration https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/probabilisticsamplerprocessor. How would this design work with that?

docs/rfcs/opinionated_sampling_cr.md
```diff
@@ -0,0 +1,140 @@
...
+    # OTLP exporter endpoint
+    endpoint: "jaeger:4317"
+```
+
+## Rollout Plan
+
+1. Introduction of the CRD in v1alpha1.
+2. First controller implementation.
+3. Implementation of e2e tests.
+4. CRD becomes part of the operator bundle.
```
jaronoff97, line 133: Let's adjust this to include some more iteration and design changes prior to being in the main bundle.

docs/rfcs/opinionated_sampling_cr.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Summary
+
+Provide a sampler CR in v1alpha1 that can be used to simplify the sampling of traces in kubernetes. 
+
+## Goals and non-goals
+
+**Goals**
+- Provide an opinionated CR to simply sampling configuration in distributed environments
+- Allow managing access using RBAC to different parts of the collector configuration
```
jaronoff97, line 23: It's not clear to me how the design solves this goal.

jaronoff97 (commented):
A few thoughts. Overall, I think i'm unclear how the CR solves the goals / use cases. It feels like this is very cluster admin focused but doesn't give users many options for customization. What if we used some annotations on the pods to determine something about the sampling configuration? Also, how would a user opt into using the created sampler architecture?

I think I would prefer a design that's a bit more composable for cluster admins / automatic for users. I.e.

Cluster admins:
* determine which services should be instrumented via a label selector (maybe this is in its own CR)
* define a `SamplerPolicy` CR
  * I like the components section, but I want this CR to be more flexible in terms of architecture based on sampler type
    * it would also be nice of users could bring their own collectors in the mix.
  *  I also want support for sampling at the SDK level in here, as IMO that's the biggest gains in terms of efficiency

Users:
* Optionally add an otel-instrumentation specific label to get instrumentation if the cluster admin requires on
* Adds annotations to override the instrumentation policy within their pod template. The cluster admin should be able to define the limits for these (something like a valid range for sample % or max cardinality, etc.)

LMK if you want to talk through this more :) I like where this is going though.

PR #3302, opened 2024-09-23
Enable multi instrumentation by default

### LINKED ISSUE #3090: Enable "enable-multi-instrumentation" flag by default

iblancasa:
Closes #3090


IshwarKanse (approved):

.chloggen/3090-enable-multiinstrumentation-by-default.yaml
```diff
@@ -0,0 +1,16 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Enable multi instrumentation by default.
+
+# One or more tracking issues related to the change
+issues: [3090]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
swiatekm, line 16: We should elaborate how this change can break things for users here.

swiatekm (approved):

.chloggen/3090-enable-multiinstrumentation-by-default.yaml
jaronoff97: +1

jaronoff97:
should we add some warnings as we discussed [here](https://github.com/open-telemetry/opentelemetry-operator/issues/3090#issuecomment-2206875946)?

iblancasa:
> should we add some warnings as we discussed [here](https://github.com/open-telemetry/opentelemetry-operator/issues/3090#issuecomment-2206875946)?

I don't think so since #3213 was implemented. With this, the warnings are not needed because users will not need to do anything.

.chloggen/3090-enable-multiinstrumentation-by-default.yaml
iblancasa: I think now is much better.

iblancasa:
Can we merge this?

swiatekm:
@jaronoff97 @pavolloffay can you have a look as well? I don't want to merge a change like this without at least another maintainer approval.

jaronoff97 (approved):
I think this LGTM. i don't believe there's anything here where an existing user would be functionally broken though right? Just more possible error states where users try to do both methods?

jaronoff97:
also, looks like tests are failing

iblancasa:
> I think this LGTM. i don't believe there's anything here where an existing user would be functionally broken though right?

I don't think so. 

> Just more possible error states where users try to do both methods?

Yes. And that should not be happening right now so..

> also, looks like tests are failing

Yes. It is related to #3307.

PR #3307, opened 2024-09-27
fix: fix container names annotations for sdk, apache, nginx

Joibel:
Found through code inspection, the ApacheHttpd, Nginx and SDK injectors do not honour their container-names annotations.

I have changed the annotation used for the container list to match the expected annotation constant which already existed in each case but was unused in the code base.

This is a breaking change if anyone is using the enablement flag with container names for these 3 injectors, which I believe would work as truthy would treat a list of container names as true.

### DIFF
--- a/.chloggen/container-names.yaml
+++ b/.chloggen/container-names.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: auto-instrumentation
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix ApacheHttpd, Nginx and SDK injectors to honour their container-names annotations.
+
+# One or more tracking issues related to the change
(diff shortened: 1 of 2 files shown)


iblancasa (approved):
Good catch!

Please, could you add a changelog?

IshwarKanse:
@Joibel Thanks for the fix, I'll update the tests after this fix.

swiatekm:
What I find worrisome is that changing this doesn't fail any tests. Could you add a unit test? Ideally, we'd also have an e2e test exercise this function.

IshwarKanse:
@swiatekm The [test](https://github.com/open-telemetry/opentelemetry-operator/blob/main/tests/e2e-multi-instrumentation/instrumentation-multi-multicontainer/01-install-app.yaml) should have checked for the ApacheHTTP, Nginx and SDK language specific instrumentation annotation. It currently only tests Java, NodeJs and Python that is the reason no tests are failing. I'll update it to check for the missing ones.

Joibel:
> Good catch!
> 
> Please, could you add a changelog?

Yes, I will do.

Joibel:
I've added a changelog. I'm not that familiar with the tests - I could possibly write one but it looks like @IshwarKanse has offered so I'll leave it with them.

swiatekm (approved):

PR #3315, opened 2024-09-30
Add FIPS flag

pavolloffay:

- Resolves: #issue-number

Fips enablement is install time configuration on the cluster

* https://learn.microsoft.com/en-us/azure/aks/enable-fips-nodes 
* https://docs.openshift.com/container-platform/4.16/installing/overview/installing-fips.html
* https://ubuntu.com/security/certifications/docs/fips-enablement
* https://discuss.kubernetes.io/t/howto-enable-fips-mode-operation/25067

### DIFF
--- a/apis/v1beta1/collector_webhook.go
+++ b/apis/v1beta1/collector_webhook.go
@@ -27,6 +27,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	"github.com/open-telemetry/opentelemetry-operator/internal/config"
+	"github.com/open-telemetry/opentelemetry-operator/internal/fips"
 	ta "github.com/open-telemetry/opentelemetry-operator/internal/manifests/targetallocator/adapters"
 	"github.com/open-telemetry/opentelemetry-operator/internal/rbac"
 )
@@ -48,6 +49,7 @@ type CollectorWebhook struct {
 	reviewer *rbac.Reviewer
 	metrics  *Metrics
 	bv       BuildValidator
+	fips     fips.FipsCheck
 }
 
 func (c CollectorWebhook) Default(_ context.Context, obj runtime.Object) error {
@@ -290,6 +292,11 @@ func (c CollectorWebhook) Validate(ctx context.Context, r *OpenTelemetryCollecto
 		return warnings, fmt.Errorf("the OpenTelemetry Collector mode is set to %s, which does not support the attribute 'deploymentUpdateStrategy'", r.Spec.Mode)
 	}
 
+	components := r.Spec.Config.GetEnabledComponents()
+	if notAllowedComponents := c.fips.Check(components[KindReceiver], components[KindExporter], components[KindProcessor], components[KindExtension]); notAllowedComponents != nil {
+		return nil, fmt.Errorf("the collector configuration contains not FIPS compliant components: %s. Please remove it from the config", notAllowedComponents)
+	}
+
 	return warnings, nil
 }
 
@@ -423,6 +430,7 @@ func NewCollectorWebhook(
 	reviewer *rbac.Reviewer,
 	metrics *Metrics,
 	bv BuildValidator,
+	fips fips.FipsCheck,
 ) *CollectorWebhook {
 	return &CollectorWebhook{
 		logger:   logger,
@@ -431,11 +439,12 @@ func NewCollectorWebhook(
 		reviewer: reviewer,
 		metrics:  metrics,
 		bv:       bv,
+		fips:     fips,
 	}
 }
 
-func SetupCollectorWebhook(mgr ctrl.Manager, cfg config.Config, reviewer *rbac.Reviewer, metrics *Metrics, bv BuildValidator) error {
-	cvw := NewCollectorWebhook(mgr.GetLogger().WithValues("handler", "CollectorWebhook", "version", "v1beta1"), mgr.GetScheme(), cfg, reviewer, metrics, bv)
+func SetupCollectorWebhook(mgr ctrl.Manager, cfg config.Config, reviewer *rbac.Reviewer, metrics *Metrics, bv BuildValidator, fipsCheck fips.FipsCheck) error {
+	cvw := NewCollectorWebhook(mgr.GetLogger().WithValues("handler", "CollectorWebhook", "version", "v1beta1"), mgr.GetScheme(), cfg, reviewer, metrics, bv, fipsCheck)
 	return ctrl.NewWebhookManagedBy(mgr).
 		For(&OpenTelemetryCollector{}).
 		WithValidator(cvw).
--- a/apis/v1beta1/collector_webhook.go
+++ b/apis/v1beta1/collector_webhook.go
@@ -49,7 +49,7 @@ type CollectorWebhook struct {
 	reviewer *rbac.Reviewer
 	metrics  *Metrics
 	bv       BuildValidator
-	fips     fips.FipsCheck
+	fips     fips.FIPSCheck
 }
 
 func (c CollectorWebhook) Default(_ context.Context, obj runtime.Object) error {
@@ -293,7 +293,7 @@ func (c CollectorWebhook) Validate(ctx context.Context, r *OpenTelemetryCollecto
 	}
 
 	components := r.Spec.Config.GetEnabledComponents()
-	if notAllowedComponents := c.fips.Check(components[KindReceiver], components[KindExporter], components[KindProcessor], components[KindExtension]); notAllowedComponents != nil {
+	if notAllowedComponents := c.fips.DisabledComponents(components[KindReceiver], components[KindExporter], components[KindProcessor], components[KindExtension]); notAllowedComponents != nil {
 		return nil, fmt.Errorf("the collector configuration contains not FIPS compliant components: %s. Please remove it from the config", notAllowedComponents)
 	}
 
@@ -430,7 +430,7 @@ func NewCollectorWebhook(
 	reviewer *rbac.Reviewer,
 	metrics *Metrics,
 	bv BuildValidator,
-	fips fips.FipsCheck,
+	fips fips.FIPSCheck,
 ) *CollectorWebhook {
 	return &CollectorWebhook{
 		logger:   logger,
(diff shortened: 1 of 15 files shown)


internal/fips/check.go
```diff
@@ -0,0 +1,102 @@
...
+	if comp := isBlackListed(fips.processors, processors); comp != "" {
+		disabled = append(disabled, comp)
+	}
+	if comp := isBlackListed(fips.extensions, extensions); comp != "" {
+		disabled = append(disabled, comp)
+	}
+	return disabled
+}
+
+func isBlackListed(blackListed map[string]bool, cfg map[string]interface{}) string {
```
jaronoff97, line 77:
```suggestion
// isDenied returns a list of denied components
func (fips FipsCheck) isDenied(cfg map[string]interface{}) []string {
```

internal/fips/check.go
```diff
@@ -0,0 +1,102 @@
...
+	for id := range cfg {
+		component := strings.Split(id, "/")[0]
+		if blackListed[component] {
+			return component
+		}
+	}
+	return ""
+}
+
+func isFipsEnabled() bool {
```
jaronoff97, line 87: Im wondering if this should follow our usual pattern for conditional functionality that's autodetected ([see here](https://github.com/open-telemetry/opentelemetry-operator/tree/main/internal/autodetect)) we could then remove the enabled check from here and just rely on the main.go operator to determine if we should run this.

internal/webhook/podmutation/webhookhandler_suite_test.go
```diff
@@ -18,6 +18,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"github.com/open-telemetry/opentelemetry-operator/internal/fips"
```
jaronoff97, line 21: minor: import order is wrong here

main.go
```diff
@@ -180,6 +182,7 @@ func main() {
 	pflag.StringVar(&encodeLevelKey, "zap-level-key", "level", "The level key to be used in the customized Log Encoder")
 	pflag.StringVar(&encodeTimeKey, "zap-time-key", "timestamp", "The time key to be used in the customized Log Encoder")
 	pflag.StringVar(&encodeLevelFormat, "zap-level-format", "uppercase", "The level format to be used in the customized Log Encoder")
+	pflag.StringVar(&fipsDisabledComponents, "fips-disabled-components", "uppercase", "Disabled collector components when operator runs on FIPS enabled platform. Example flag value =receiver.foo,receiver.bar,exporter.baz")
```
jaronoff97, line 185: should we name this `disabled-components`? I thought we had discussed wanting this for non-fips users?

internal/fips/check.go
```diff
@@ -0,0 +1,102 @@
...
+func listToMap(list []string) map[string]bool {
+	m := map[string]bool{}
+	for _, v := range list {
+		m[v] = true
+	}
+	return m
+}
+
+// Check checks if a submitted components are back lister or not.
+func (fips FipsCheck) Check(receivers map[string]interface{}, exporters map[string]interface{}, processors map[string]interface{}, extensions map[string]interface{}) []string {
```
jaronoff97, line 57: should we use the v1beta1 types here rather than maps?

main.go
pavolloffay: > should we name this disabled-components? I thought we had discussed wanting this for non-fips users?

I don't want to make this feature generic for the follow up reasons:

If we wanted to make this generic we would have to introduce another flag that would conditionally enable the disabled components flag. Such flag could check env var, however users would need to rebuild the operator image and introduce init script that would set the env var based on the content of `/proc/sys/crypto/fips_enabled`

The FIPS feature on k8s is a well-defined concept. See the links in the PR descriptions. Making its own configuration option on the operator makes sense to me, rather than allowing the use case with generic option that require no trivial configuration.

internal/fips/check.go
pavolloffay: that would introduce a circular import. 

Important is that the use of this API in the webhook does not require any type conversion.

internal/fips/check.go
pavolloffay: +

pavolloffay:
@jaronoff97 I have updated the PR with the requested changes

internal/fips/fipscheck.go
```diff
@@ -0,0 +1,88 @@
...
+
+type noopFIPSCheck struct{}
+
+func (noopFIPSCheck) DisabledComponents(receivers map[string]interface{}, exporters map[string]interface{}, processors map[string]interface{}, extensions map[string]interface{}) []string {
+	return nil
+}
+
+// NewFipsCheck creates new FipsCheck.
+// It checks if FIPS is enabled on the platform in /proc/sys/crypto/fips_enabled.
+func NewFipsCheck(FIPSEnabled bool, receivers, exporters, processors, extensions []string) FIPSCheck {
```
jaronoff97, line 41: at the call site, we should just create the FipsCheck if its enabled, otherwise we set the interface to be nil i.e.

```
var fipsCheck fips.FIPSCheck
if ad.FIPSEnabled(ctx) {
    fips.NewFipsCheck(receivers, exporters, processors, extensions)
}
...
```

apis/v1beta1/collector_webhook_test.go
```diff
@@ -113,6 +114,7 @@ func TestValidate(t *testing.T) {
 			getReviewer(test.shouldFailSar),
 			nil,
 			bv,
+			fips.NewFipsCheck(false, nil, nil, nil, nil),
```
jaronoff97, line 117: we should be able to just set this equal to nil now

apis/v1beta1/collector_webhook.go
```diff
@@ -290,6 +292,11 @@ func (c CollectorWebhook) Validate(ctx context.Context, r *OpenTelemetryCollecto
 		return warnings, fmt.Errorf("the OpenTelemetry Collector mode is set to %s, which does not support the attribute 'deploymentUpdateStrategy'", r.Spec.Mode)
 	}
 
+	components := r.Spec.Config.GetEnabledComponents()
+	if notAllowedComponents := c.fips.DisabledComponents(components[KindReceiver], components[KindExporter], components[KindProcessor], components[KindExtension]); notAllowedComponents != nil {
```
jaronoff97, line 296: given fips is an interface we should check if its nil prior to calling this. Doing it this way will also mean we can remove the noop fips check

apis/v1beta1/collector_webhook_test.go
jaronoff97: (same with the below)

internal/fips/fipscheck.go
```diff
@@ -0,0 +1,88 @@
...
+	if comp := isDisabled(fips.processors, processors); comp != "" {
+		disabled = append(disabled, comp)
+	}
+	if comp := isDisabled(fips.extensions, extensions); comp != "" {
+		disabled = append(disabled, comp)
+	}
+	return disabled
+}
+
+func isDisabled(blackListed map[string]bool, cfg map[string]interface{}) string {
```
jaronoff97, line 80:
```suggestion
func isDisabled(denyList map[string]bool, cfg map[string]interface{}) string {
```

internal/fips/fipscheck.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"strings"
+)
+
+type FIPSCheck interface {
+	DisabledComponents(receivers map[string]interface{}, exporters map[string]interface{}, processors map[string]interface{}, extensions map[string]interface{}) []string
+}
+
+// FipsCheck holds configuration for FIPS black list.
```
jaronoff97, line 25:
```suggestion
// FipsCheck holds configuration for FIPS deny lists.
```

main.go
pavolloffay: no, that would not work

internal/fips/fipscheck.go
pavolloffay: I prefer to avoid dealing with `nil` checks in the webhook or wherever the check is used.

jaronoff97: that's how we handle other conditional checks that the webhook is supposed to be running today ([here](https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1beta1/collector_webhook.go#L48-L50)) so it feels odd to special case this. I suppose if we want to avoid that we could just create the NoopCheck here instead. I don't love the bool in the signature, it feels like a leak in terms of responsibility for the package.

main.go
pavolloffay: The operator needs to be shipped with preset `fips-disabled-components` and the components should be disabled only when running on FIPS platform. Customers don't have option to modify this flag when the operator is installed via OLM.

jaronoff97: is the worry that if it's always settable that a user could override the disabled components? I'm trying to understand why it should only be disabled on FIPS Platform vs a user trying to disable components on any platform. I don't quite understand the distinction/requirement. And in this case if it's going to be shipped with preset components should this be a build argument?

pavolloffay: > And in this case if it's going to be shipped with preset components should this be a build argument?

It is going to be shipped with preset flag in the bundle/CSV.

jaronoff97: ah i see. okay. I'm still not sure _why_ this means we can't just have it be `disabled-components` and have it be required if FIPS is true and optional if its set for the rest of the bundle though. I don't think it's blocking though.

pavolloffay:
@jaronoff97 PR updated. I have removed the `noopFips` and used `nil`

jaronoff97 (commented):
one comment, otherwise looks good!

internal/fips/fipscheck.go
```diff
@@ -0,0 +1,81 @@
...
+// FipsCheck holds configuration for FIPS deny list.
+type fipsCheck struct {
+	receivers  map[string]bool
+	exporters  map[string]bool
+	processors map[string]bool
+	extensions map[string]bool
+}
+
+// NewFipsCheck creates new FipsCheck.
+func NewFipsCheck(FIPSEnabled bool, receivers, exporters, processors, extensions []string) FIPSCheck {
```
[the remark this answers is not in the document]
pavolloffay: oh I forgot this, removed

Issue #3319, opened 2024-10-02
Move validation to be part of the CRD


jaronoff97:
### Component(s)

collector, auto-instrumentation, target allocator, opamp

### Is your feature request related to a problem? Please describe.

CRDs can have validation rules now, it would be great to move our existing webhooks (for whatever we can) into [these](https://kubernetes.io/blog/2022/09/23/crd-validation-rules-beta/).

### Describe the solution you'd like

Migrate to the kubebuilder's support for CRD validation rules. This would probably require us to bump our min supported version ...

mercybassey:
Hi @jaronoff97 I am an Outreachy applicant, and I would like to work on this issue. I'll review the project and keep you updated on my progress. In the meantime, I would appreciate any insights or guidance you could provide to help me address this issue.

jaronoff97:
@mercybassey sounds good! one issue with this is that feature went beta in Kubernetes 1.25 and stable in kubernetes 1.29 ([link](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules)) and right now our [minimum supported version is 1.23](https://github.com/open-telemetry/opentelemetry-operator?tab=readme-ov-file#opentelemetry-operator-vs-kubernetes-vs-cert-manager-vs-prometheus-operator). We will probably need to bump that minimum version as part of this. We have an open issue here for that #2798

mercybassey:
@jaronoff97 Thank you for the update! I’d like to clarify if I can go ahead with migrating the validation rules to the CRDs while keeping an eye on the progress of issue #2798 regarding the version bump.

mercybassey:
> Assuming the `x-kubernetes-validations` field is ignored on K8s <1.25 (it should be, but we need to verify), then I think we can add these validations, without removing existing ones from the webhook. There isn't really any downside to having both for users. It does represent additional maintenance burden for us, but I wouldn't expect it to be very noticable.
> 
[...]

PR #3321, opened 2024-10-03
Move validation to be part of the CRD

mercybassey:

- Resolves: #3319 

### DIFF
--- a/apis/v1beta1/opentelemetrycollector_types.go
+++ b/apis/v1beta1/opentelemetrycollector_types.go
@@ -113,7 +113,7 @@ type OpenTelemetryCollectorSpec struct {
 	// functionality is only available if one of the valid modes is set.
 	// Valid modes are: deployment, daemonset and statefulset.
 	// +optional
-	// +kubebuilder:validation:Enum=deployment;daemonSet;statefulSet;
+	// +kubebuilder:validation:Enum=deployment;daemonSet;statefulSet
 	Ingress Ingress `json:"ingress,omitempty"`
 	// Liveness config for the OpenTelemetry Collector except the probe handler which is auto generated from the health extension of the collector.
 	// It is only effective when healthcheckextension is configured in the OpenTelemetry Collector pipeline.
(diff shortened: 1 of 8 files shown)


mercybassey:
@swiatekm Based on my changes, I just wanted to know if am on the right track before I proceed for all CRDs. Thanks.

apis/v1beta1/opentelemetrycollector_types.go
```diff
@@ -90,10 +90,10 @@ type OpenTelemetryCollectorSpec struct {
 	// +optional
 	TargetAllocator TargetAllocatorEmbedded `json:"targetAllocator,omitempty"`
 	// Mode represents how the collector should be deployed (deployment, daemonset, statefulset or sidecar)
-	// +optional
+	// +kubebuilder:validation:Enum=ModeDeployment;ModeDaemonSet;ModeStatefulSet;ModeSidecar
```
swiatekm, line 93: This specific change is superfluous. If you look at the [generated CRD](https://github.com/open-telemetry/opentelemetry-operator/blob/91759b0138c6b58329b3420ef159688ff0070e4c/config/crd/bases/opentelemetry.io_opentelemetrycollectors.yaml#L2328), this validation is already there. If we look at what's missing compared to the [webhook](https://github.com/open-telemetry/opentelemetry-operator/blob/91759b0138c6b58329b3420ef159688ff0070e4c/apis/v1beta1/collector_webhook.go#L60), we should add a marker here to set the default to `deployment`.

Keep in mind that the markers deal with raw values of the field as they appear in the resource in K8s, not the aliases we have for them in Go code. So you should use `deployment` instead of `ModeDeployment`.

swiatekm (commented):
You're making changes in the right place, but I'm not sure they're the right changes for the effect you want.

Try running `make bundle` and examining the result of your changes to the [generated CRDS](https://github.com/open-telemetry/opentelemetry-operator/tree/main/config/crd/bases). This should give you a better sense of whether you're on the right path.

mercybassey:
@swiatekm I addressed your comment and added more validation rules. I'll appreciate your review. Thank you.

apis/v1beta1/opentelemetrycollector_types.go
```diff
@@ -110,6 +113,7 @@ type OpenTelemetryCollectorSpec struct {
 	// functionality is only available if one of the valid modes is set.
 	// Valid modes are: deployment, daemonset and statefulset.
 	// +optional
+	// +kubebuilder:validation:Enum=deployment;daemonSet;statefulSet;
```
pavolloffay, line 116: this is not correct

mercybassey: This is a mistake. I'll remove the extra  `;`.

mercybassey:
@pavolloffay I have fixed it.

Issue #3306, opened 2024-09-26
Remove ComponentUseLocalHostAsDefaultHost()


diranged:
### Component(s)

collector

## Description

The [`ComponentUseLocalHostAsDefaultHost()`](https://github.com/open-telemetry/opentelemetry-operator/blob/v0.109.0/apis/v1beta1/collector_webhook.go#L446-L464) function should be removed at this point... the operator forces the `component.UseLocalHostAsDefaultHost` flag to be added. The problem is that in the OTEL Collector [0.110.0](https://github.com/open-telemetry/opentelemetry-collector-releases/releases/tag/v0.110.0),  that flag is no longer settable:

```
Error: invalid argument "+processor.resourcedetection.hostCPUSteppingAsString,+exporter.prometheusremotewritexporter.RetryOn429,-component.UseLocalHostAsDefaultHost" for "--feature-gates" flag: feature gate "component.UseLocalHostAsDefaultHost" is stable, can not be disabled
```

### Kubernetes Version

1.30.0

### Operator version

0.109.0

### Collector version

1.111.0

## Environment
OS: (e.g., "Ubuntu 20.04")
Compiler(if manually compiled): (e.g., "go 14.2")

jaronoff97:
this should be part of our next release, cc @frzifus would you be able to do this to close https://github.com/open-telemetry/opentelemetry-operator/issues/3126 as well?

frzifus:
yes!

ryanmt:
Is there any way to prevent this feature flag from being added to the produced resources w/o a new release? I'm trying to overcome an issue similar to this one for the helm chart from the install manifests. https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1339

jaronoff97:
@ryanmt we will need to do a new release as that's how the images get published, though i guess you could use the image from main. We haven't released 0.110.0 yet, so i would recommend not misaligning those versions.

henrikrexed:
Hi,
i'm having the same issue , i forced to use a collector v 0.110 to get the latest functions of the transform processor. 
When would the feature flalg UseLocalHostAsdeFualthost would be removed and when is the next release of the operator planned?

jaronoff97:
@henrikrexed @frzifus is on it right now and will be done hopefully by EOW this week.

odubajDT:
[...]

PR #3325, opened 2024-10-04
Use 0.0.0.0 as otlp receiver default address

frzifus:

This change sets the otlp receiver endpoint 0.0.0.0 as default, unless otherwise configured by the user.

This PR does not change the default endpoint for other receivers. 

The following steps:
- Add default addresses for other receivers
- Remove ComponentUseLocalHostAsDefaultHost feature gate flag

- OTLP part of https://github.com/open-telemetry/opentelemetry-operator/issues/3126
- Unblocks the work on https://github.com/open-telemetry/opentelemetry-operator/issues/3306


internal/components/multi_endpoint.go
```diff
@@ -72,6 +72,23 @@ func (m *MultiPortReceiver) ParserName() string {
 	return fmt.Sprintf("__%s", m.name)
 }
 
+func (m *MultiPortReceiver) GetDefaultConfig(logger logr.Logger, config interface{}) (interface{}, error) {
+	multiProtoEndpointCfg := &MultiProtocolEndpointConfig{}
+	if err := mapstructure.Decode(config, multiProtoEndpointCfg); err != nil {
+		return nil, err
+	}
+	for protocol, ec := range multiProtoEndpointCfg.Protocols {
+		// TODO: Should we take default receiver address from settngs?
```
jaronoff97, line 81: yes you can just take it from there

internal/components/generic_parser.go
```diff
@@ -27,15 +28,48 @@ var (
...
-	rbacGen      RBACRuleGenerator[T]
-	livenessGen  ProbeGenerator[T]
-	readinessGen ProbeGenerator[T]
+	name            string
+	settings        *Settings[T]
+	portParser      PortParser[T]
+	rbacGen         RBACRuleGenerator[T]
+	livenessGen     ProbeGenerator[T]
+	readinessGen    ProbeGenerator[T]
+	defaultsApplier Defaulter[T]
```
jaronoff97, line 56: youll need to set a functionthat matches the signature here for the single endpoint config.

frzifus: That should be here:
```golang
func NewSinglePortParserBuilder(name string, port int32) Builder[*SingleEndpointConfig] {
	return NewBuilder[*SingleEndpointConfig]().WithPort(port).WithName(name).WithPortParser(ParseSingleEndpoint).WithDefaultsApplier(AddressDefaulter)
}

func NewSilentSinglePortParserBuilder(name string, port int32) Builder[*SingleEndpointConfig] {
	return NewBuilder[*SingleEndpointConfig]().WithPort(port).WithName(name).WithPortParser(ParseSingleEndpointSilent).WithDefaultsApplier(AddressDefaulter)
}
```

jaronoff97: ah 👍

internal/components/multi_endpoint.go
```diff
@@ -72,6 +73,37 @@ func (m *MultiPortReceiver) ParserName() string {
...
+		if err != nil {
+			return nil, err
+		}
+		tmp[protocol] = res
+	}
+
+	for protocol, ec := range tmp {
+		multiProtoEndpointCfg.Protocols[protocol] = ec
+	}
+	return config, mapstructure.Decode(multiProtoEndpointCfg, &config)
```
frzifus, line 104: In case we need to marshal `listenaddress`, I think we should remove the entry here.

apis/v1beta1/config.go
```diff
@@ -225,6 +226,72 @@ func (c *Config) getPortsForComponentKinds(logger logr.Logger, componentKinds ..
...
+					return err
+				}
+				out := make(map[string]interface{}, 0)
+				if err := yaml.Unmarshal(got, out); err != nil {
+					return err
+				}
+				// NOTE: The underlying struct compoents.SingleEndpointConfig adds this listenaddress
+				// field. It is marshaled due to internal use. To avoid adding invalid fields to the
+				// collector config, this temporary workaround removes this field.
+				// TODO: Try to get rid of it or move it into the parser.GetDefaultConfig method.
```
jaronoff97, line 264: i think if we added yaml tags for omitempty this would just work.

apis/v1beta1/config.go
```diff
@@ -225,6 +226,72 @@ func (c *Config) getPortsForComponentKinds(logger logr.Logger, componentKinds ..
...
+				removeKeysRecursively(out, "listenaddress")
```
jaronoff97, line 265: i think if we added yaml tags for omitempty this would just work.

apis/v1beta1/config.go
```diff
@@ -225,6 +226,72 @@ func (c *Config) getPortsForComponentKinds(logger logr.Logger, componentKinds ..
...
+			continue
+		}
+		for componentName := range enabledComponents[componentKind] {
+			parser := retriever(componentName)
+			if newCfg, err := parser.GetDefaultConfig(logger, cfg.Object[componentName]); err != nil {
+				return err
+			} else {
+				// NOTE: The internally used parser returns components.SingleEndpointConfig
+				// as a map value. The following lines normalize this value..
+				got, err := yaml.Marshal(newCfg)
```
jaronoff97, line 253: are we able to use mapstructure here instead?

apis/v1beta1/config.go
jaronoff97: or see above.

internal/components/generic_parser.go
```diff
@@ -30,12 +30,29 @@ var (
...
+
+	if g.settings.defaultRecAddr == "" {
+		return config, nil
+	}
+
+	var parsed T
+	if err := mapstructure.Decode(config, &parsed); err != nil {
+		return nil, err
+	}
+	return g.defaultsApplier(logger, g.settings.defaultRecAddr, g.settings.port, parsed)
```
jaronoff97, line 55: lets use g.settings.GetServicePort() here instead per 
https://github.com/open-telemetry/opentelemetry-operator/pull/3325/files#diff-05ed6f1e9e944cb19f7bb12a2ed9ce66c1e75ecabb6807cf3e03e0dd1afdf360R99

internal/components/multi_endpoint.go
```diff
@@ -72,6 +73,37 @@ func (m *MultiPortReceiver) ParserName() string {
 	return fmt.Sprintf("__%s", m.name)
 }
 
+func (m *MultiPortReceiver) GetDefaultConfig(logger logr.Logger, config interface{}) (interface{}, error) {
```
jaronoff97, line 76: dont we want to just delegate to `MultiAddressDefaulter` here?

internal/components/component.go
```diff
@@ -49,6 +49,10 @@ type RBACRuleGenerator[ComponentConfigType any] func(logger logr.Logger, config
 // It's expected that type Config is the configuration used by a parser.
 type ProbeGenerator[ComponentConfigType any] func(logger logr.Logger, config ComponentConfigType) (*corev1.Probe, error)
 
+// Defaulter is a function that applies given defaults to the passed Config.
+// It's expected that type Config is the configuration used by a parser.
+type Defaulter[ComponentConfigType any] func(logger logr.Logger, defaultRecAddr string, port int32, config ComponentConfigType) (ComponentConfigType, error)
```
jaronoff97, line 54: i think it _may_ be simpler to just return a map[string]interface here

apis/v1beta1/config.go
frzifus: I had that before but did end up with:
```diff
Diff:
--- Expected
+++ Actual
@@ -115,9 +115,11 @@
      (string) (len=4) "otlp": (map[string]interface {}) (len=1) {
-      (string) (len=9) "protocols": (map[string]interface {}) (len=2) {
-       (string) (len=4) "grpc": (map[string]interface {}) (len=1) {
-        (string) (len=8) "endpoint": (string) (len=12) "0.0.0.0:4317"
-       },
-       (string) (len=4) "http": (map[string]interface {}) (len=1) {
-        (string) (len=8) "endpoint": (string) (len=12) "0.0.0.0:4318"
-       }
+      (string) (len=9) "protocols": (map[string]*components.SingleEndpointConfig) (len=2) {
+       (string) (len=4) "grpc": (*components.SingleEndpointConfig)({
+        Endpoint: (string) (len=12) "0.0.0.0:4317",
+        ListenAddress: (string) ""
+       }),
+       (string) (len=4) "http": (*components.SingleEndpointConfig)({
+        Endpoint: (string) (len=12) "0.0.0.0:4318",
+        ListenAddress: (string) ""
+       })
```

frzifus: Well. Thats now obsolete.

internal/components/generic_parser.go
```diff
@@ -30,12 +30,29 @@ var (
...
+
+	if g.settings.defaultRecAddr == "" {
+		return config, nil
+	}
+
+	var parsed T
+	if err := mapstructure.Decode(config, &parsed); err != nil {
+		return nil, err
+	}
+	return g.defaultsApplier(logger, func(string) (string, int32) { return g.settings.defaultRecAddr, g.settings.GetServicePort().Port }, parsed)
```
jaronoff97, line 55: this signature feels a bit odd. The multi endpoint struct doesn't use the generic parser at all, so i think we could get away with the signature just taking in a defaultRecAddr and a port int. The multi endpoint would just use the protocol it's been given..\

apis/v1beta1/config.go
```diff
@@ -225,6 +225,36 @@ func (c *Config) getPortsForComponentKinds(logger logr.Logger, componentKinds ..
 	return ports, nil
 }
 
+// getPortsForComponentKinds gets the ports for the given ComponentKind(s).
```
pavolloffay, line 228: the comment should be updated

internal/components/generic_parser.go
frzifus: I see the point. Earlier it was that way, but I tried to reuse the same `Defaulter` on the `GenericParser`  and `MultiPortReceiver`. While we would change the signature to this:
```diff
- type Defaulter[ComponentConfigType any] func(logger logr.Logger, addrProv AddressProvider, config ComponentConfigType) (map[string]interface{}, error)
+ type Defaulter[ComponentConfigType any] func(logger logr.Logger, addrProv AddressProvider, config *SingleEndpointConfig) (map[string]interface{}, error)
 ```
Its sufficent for the `AddressDefaulter` but not for the `MultiAddressDefaulter`. Since there we may have different  addresses set and/or different default ports.

The only thing that comes to my would be moving some of the logic into the `MultiPortReceiver) GetDefaultConfig` and use the  same AddressDefaulter.

jaronoff97 (approved):

pavolloffay (approved):

jvoravong (approved):

Issue #2264, opened 2023-10-24
Python autoinstrumentation for musl libc based application containers


ilyamochalov:
### Component(s)

instrumentation

### Is your feature request related to a problem? Please describe.

Python autoinstrumentation for musl libc based application containers fails with the following error:
```sh
#16 2.190 ImportError: Error relocating /autoinstrumentation/psutil/_psutil_linux.abi3.so: __sched_cpufree: symbol not found
#16 2.191 Failed to auto initialize opentelemetry
#16 2.191 Traceback (most recent call last):
#16 2.191   File "/autoinstrumentation/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py", line 39, in initialize
#16 2.191     _load_instrumentors(distro)
#16 2.191   File "/autoinstrumentation/opentelemetry/instrumentation/auto_instrumentation/_load.py", line 91, in _load_instrumentors
#16 2.191     raise exc
#16 2.191   File "/autoinstrumentation/opentelemetry/instrumentation/auto_instrumentation/_load.py", line 87, in _load_instrumentors
#16 2.191     distro.load_instrumentor(entry_point, skip_dep_check=True)
#16 2.191   File "/autoinstrumentation/opentelemetry/instrumentation/distro.py", line 62, in load_instrumentor
#16 2.191     instrumentor: BaseInstrumentor = entry_point.load()
#16 2.191   File "/autoinstrumentation/pkg_resources/__init__.py", line 2518, in load
#16 2.191     return self.resolve()
#16 2.191   File "/autoinstrumentation/pkg_resources/__init__.py", line 2524, in resolve
#16 2.191     module = __import__(self.module_name, fromlist=['__name__'], level=0)
#16 2.191   File "/autoinstrumentation/opentelemetry/instrumentation/system_metrics/__init__.py", line 79, in <module>
#16 2.191     import psutil
#16 2.191   File "/autoinstrumentation/psutil/__init__.py", line 102, in <module>
#16 2.191     from . import _pslinux as _psplatform
#16 2.191   File "/autoinstrumentation/psutil/_pslinux.py", line 25, in <module>
#16 2.191     from . import _psutil_linux as cext
#16 2.191 ImportError: Error relocating /autoinstrumentation/psutil/_psutil_linux.abi3.so: __sched_cpufree: symbol not found
```

Root cause: current autoinstrumentation build packaged for BSD libc. 

### Describe the solution you'd like

[...]

PR #2266, opened 2023-10-24
Python autoinstrumentation for musl libc based application containers

ilyamochalov:
**Description:**  adding musl build to python autoinstrumentation

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-operator/issues/2264 

**Testing:** updated e2e test and also added unit tests cases

**Documentation:** updated the README file

PR #3332, opened 2024-10-07
Python auto-instrumentation: handle musl based containers

xrmx:

Build and and inject musl based python auto-instrumentation if proper annotation is configured:

instrumentation.opentelemetry.io/otel-python-platform: "musl"

This takes a different approach that the stale PR at #2266:
- does not change directory where python sdk is installed in the docker container for default distribution (glibc)

- Resolves: #2264

**Testing:**  unit tests and e2e are green. Tested locally on minikube that by deploying an operator from this branch, a custom image for the auto-instrumentation was able to get metrics (that depends on psutil that uses a binary extension) out of a python alpine container with the `instrumentation.opentelemetry.io/otel-python-platform: "musl"` annotation and a stacktrace without it.
Also tested that the glibc based is copied to the container if the instrumentation image does not have the musl one.

**Documentation:** <Describe the documentation added.> Will add docs to opentelemetry.io

### DIFF
--- a/pkg/instrumentation/annotation.go
+++ b/pkg/instrumentation/annotation.go
@@ -30,6 +30,7 @@ const (
 	annotationInjectNodeJSContainersName      = "instrumentation.opentelemetry.io/nodejs-container-names"
 	annotationInjectPython                    = "instrumentation.opentelemetry.io/inject-python"
 	annotationInjectPythonContainersName      = "instrumentation.opentelemetry.io/python-container-names"
+	annotationPythonPlatform                  = "instrumentation.opentelemetry.io/otel-python-platform"
 	annotationInjectDotNet                    = "instrumentation.opentelemetry.io/inject-dotnet"
 	annotationDotNetRuntime                   = "instrumentation.opentelemetry.io/otel-dotnet-auto-runtime"
 	annotationInjectDotnetContainersName      = "instrumentation.opentelemetry.io/dotnet-container-names"
--- a/pkg/instrumentation/python.go
+++ b/pkg/instrumentation/python.go
@@ -23,19 +23,23 @@ import (
 )
 
 const (
-	envPythonPath               = "PYTHONPATH"
-	envOtelTracesExporter       = "OTEL_TRACES_EXPORTER"
-	envOtelMetricsExporter      = "OTEL_METRICS_EXPORTER"
-	envOtelLogsExporter         = "OTEL_LOGS_EXPORTER"
-	envOtelExporterOTLPProtocol = "OTEL_EXPORTER_OTLP_PROTOCOL"
-	pythonPathPrefix            = "/otel-auto-instrumentation-python/opentelemetry/instrumentation/auto_instrumentation"
-	pythonPathSuffix            = "/otel-auto-instrumentation-python"
-	pythonInstrMountPath        = "/otel-auto-instrumentation-python"
-	pythonVolumeName            = volumeName + "-python"
-	pythonInitContainerName     = initContainerName + "-python"
+	envPythonPath                    = "PYTHONPATH"
+	envOtelTracesExporter            = "OTEL_TRACES_EXPORTER"
+	envOtelMetricsExporter           = "OTEL_METRICS_EXPORTER"
+	envOtelLogsExporter              = "OTEL_LOGS_EXPORTER"
+	envOtelExporterOTLPProtocol      = "OTEL_EXPORTER_OTLP_PROTOCOL"
+	glibcLinuxAutoInstrumentationSrc = "/autoinstrumentation/."
+	muslLinuxAutoInstrumentationSrc  = "/autoinstrumentation-musl/."
+	pythonPathPrefix                 = "/otel-auto-instrumentation-python/opentelemetry/instrumentation/auto_instrumentation"
+	pythonPathSuffix                 = "/otel-auto-instrumentation-python"
+	pythonInstrMountPath             = "/otel-auto-instrumentation-python"
+	pythonVolumeName                 = volumeName + "-python"
+	pythonInitContainerName          = initContainerName + "-python"
+	glibcLinux                       = "glibc"
+	muslLinux                        = "musl"
 )
 
-func injectPythonSDK(pythonSpec v1alpha1.Python, pod corev1.Pod, index int) (corev1.Pod, error) {
+func injectPythonSDK(pythonSpec v1alpha1.Python, pod corev1.Pod, index int, platform string) (corev1.Pod, error) {
 	volume := instrVolume(pythonSpec.VolumeClaimTemplate, pythonVolumeName, pythonSpec.VolumeSizeLimit)
 
 	// caller checks if there is at least one container.
@@ -44,6 +48,16 @@ func injectPythonSDK(pythonSpec v1alpha1.Python, pod corev1.Pod, index int) (cor
 	err := validateContainerEnv(container.Env, envPythonPath)
 	if err != nil {
 		return pod, err
+	}
+
+	autoInstrumentationSrc := ""
+	switch platform {
+	case "", glibcLinux:
+		autoInstrumentationSrc = glibcLinuxAutoInstrumentationSrc
+	case muslLinux:
+		autoInstrumentationSrc = muslLinuxAutoInstrumentationSrc
+	default:
+		return pod, fmt.Errorf("provided instrumentation.opentelemetry.io/otel-python-platform annotation value '%s' is not supported", platform)
 	}
 
 	// inject Python instrumentation spec env vars.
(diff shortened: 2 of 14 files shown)


tests/e2e-instrumentation/instrumentation-python-musl/00-install-collector.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: opentelemetry.io/v1alpha1
```
xrmx, line 1: Adding specific e2e tests is not needed because the `e2e-test-app-python` docker image is already based on alpine right? looks like tests are failing atm

iblancasa: >  looks like tests are failing on main

Can you elaborate this a bit more?


> Adding specific e2e tests is not needed because the e2e-test-app-python docker image is already based on alpine right? 

Maybe this was something where I failed. Our idea is, at some point, add verifications to know if the libraries were injected properly and verify they are emitting data.

iblancasa (commented):
Missing changelog.

tests/e2e-instrumentation/instrumentation-python-musl/00-install-collector.yaml
xrmx: BTW the other thing that should be kept in sync is the Python version of the two instrumentation because the ABI changes between python versions.

iblancasa: > At the moment the `e2e-test-app-python` is based on alpine but the python instrumentation image is glibc based. This is a problem because binary extensions are not portable between different C libraries (among other incompatibilities). So this PR builds them and copies one for musl or glibc depending on the configuration.

Didn't notice this when I added the images. As mentioned in the previous comment, the idea is to add real E2E checking if the instrumentation is generating real data. Since we are not checking this, issues like the one you saw are happening. This is something we need to fix with that image.

You can reuse that image (since it is `musl` based) for your E2E test. We need to add a new one for `glibc`.

xrmx: > You can reuse that image (since it is `musl` based) for your E2E test. We need to add a new one for `glibc`.

I'm already using that image in the musl e2e :+1:

pkg/instrumentation/annotation.go
```diff
@@ -30,6 +30,7 @@ const (
 	annotationInjectNodeJSContainersName      = "instrumentation.opentelemetry.io/nodejs-container-names"
 	annotationInjectPython                    = "instrumentation.opentelemetry.io/inject-python"
 	annotationInjectPythonContainersName      = "instrumentation.opentelemetry.io/python-container-names"
+	annotationPythonWheelKind                 = "instrumentation.opentelemetry.io/python-wheel-kind"
```
swiatekm, line 33: I don't like the name of this annotation. For dotnet, it's called `otel-dotnet-auto-runtime`, which is also not great, but we should aim at some consistency here. That aside, `python-wheel-kind` doesn't say anything useful to the user, in my opinion.

swiatekm (commented):
Could you have a look at the e2e test failures?

pkg/instrumentation/annotation.go
xrmx: Do you mean to use something like `otel-python-distribution` that takes (musl / glibc)? Or want to use `otel-python-auto-runtume`? runtime may make more sense in .net world than in python though

swiatekm: I think the name should indicate that what we're really looking for here is the type of libc the application was built against. @pavolloffay @jaronoff97 @iblancasa any thoughts on this?

iblancasa: I agree

xrmx: Pushed out a new version using `otel-python-platform` as config name with `glibc` and `musl` as accepted values

xrmx:
AFAICS e2e tests are failing because the python autoinstrumentation docker image does not have the musl based installation, is it the case because the image used in tests is not built from git?

swiatekm:
> AFAICS e2e tests are failing because the python autoinstrumentation docker image does not have the musl based installation, is it the case because the image used in tests is not built from git?

Autoinstrumentation tests use the default image, we don't build all the autoinstrumentation images from source for E2E tests, though we probably should.

What I would suggest here:
1. Verify locally that your current tests pass with the new image
2. Modify tests in this PR to only check if the container manifest is correct (and not that it's running)
3. Merge this PR
4. Operator release happens
5. Open a new PR, changing the tests back to their current state

I know it's a hassle, but it's simpler than all the alternatives.

xrmx:
> > AFAICS e2e tests are failing because the python autoinstrumentation docker image does not have the musl based installation, is it the case because the image used in tests is not built from git?
> 
> Autoinstrumentation tests use the default image, we don't build all the autoinstrumentation images from source for E2E tests, though we probably should.
> 
> What I would suggest here:
> 
>     1. Verify locally that your current tests pass with the new image
> 
>     2. Modify tests in this PR to only check if the container manifest is correct (and not that it's running)
> 
>     3. Merge this PR
> 
>     4. Operator release happens
> 
>     5. Open a new PR, changing the tests back to their current state
> 
> 
> I know it's a hassle, but it's simpler than all the alternatives.

Other than fixing the tests though this is breaking backward compatibility with older images and maybe we can avoid that?

xrmx:
Ok finally tested this manually by deploying an operator from this branch, a custom image for the auto-instrumentation and was able to get metrics (that depends on psutil that uses a binary extension) out of a python alpine container with the `        instrumentation.opentelemetry.io/otel-python-platform: "musl"` annotation and a stackstrace without it.

autoinstrumentation/python/Dockerfile
```diff
@@ -1,12 +1,12 @@
 # To build one auto-instrumentation image for Python, please:
-# - Ensure the packages are installed in the `/autoinstrumentation` directory. This is required as when instrumenting the pod,
-#   one init container will be created to copy all the content in `/autoinstrumentation` directory to your app's container. Then
+# - Ensure the packages are installed in the `/autoinstrumentation,{musl}` directory. This is required as when instrumenting the pod,
```
xrmx, line 2:
```suggestion
# - Ensure the packages are installed in the `/autoinstrumentation,{-musl}` directory. This is required as when instrumenting the pod,
```

pmcollins (approved):
I'm not an approver here but this looks great @xrmx. Do we have a place where we can document any of this for users of this functionality?

xrmx:
> I'm not an approver here but this looks great @xrmx. Do we have a place where we can document any of this for users of this functionality?

Thanks for reviewing, will update https://opentelemetry.io/docs/kubernetes/operator/automatic/ and https://opentelemetry.io/docs/zero-code/python/operator/

pkg/instrumentation/python.go
```diff
@@ -113,10 +127,13 @@ func injectPythonSDK(pythonSpec v1alpha1.Python, pod corev1.Pod, index int) (cor
 				},
 			}})
 
+		// copy the default glibc based dir if the autoInstrumentationSrc does not exist to not break compat with old images
+		shellCmd := fmt.Sprintf("test -d %[1]s && cp -r %[1]s %[2]s || cp -r %[3]s %[2]s", autoInstrumentationSrc, pythonInstrMountPath, glibcLinuxAutoInstrumentationSrc)
```
swiatekm, line 131: Honestly, while this makes tests pass and makes life easier for us, I think it's bad for users. If a user sets the annotation to `musl` and provides an image without the artifacts, we *should* fail and let them know.

xrmx: It's fine for me. If we agree on the general approach would it be fine to get a release out with the updated image installing the musl directory and then open a PR with the code making use of it?

swiatekm: Yes, that sounds good to me.

xrmx: PR updating the python image here https://github.com/open-telemetry/opentelemetry-operator/pull/3384

iblancasa (approved):
Missing changelog.

xrmx:
> Missing changelog.

It's the first hunk in the diff

tests/e2e-instrumentation/instrumentation-python-musl/01-assert.yaml
```diff
@@ -0,0 +1,80 @@
...
+    volumeMounts:
+    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
+      readOnly: true
+    - mountPath: /otel-auto-instrumentation-python
+      name: opentelemetry-auto-instrumentation-python
+  - args:
+    - --config=env:OTEL_CONFIG
+    name: otc-container
+  initContainers:
+  - name: opentelemetry-auto-instrumentation-python
```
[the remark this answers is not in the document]
xrmx: @swiatekm I have no clue on how to do that, any hint?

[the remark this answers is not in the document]
swiatekm: You can add the container command here, it contains the source directory.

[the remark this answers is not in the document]
xrmx: Added, thanks!

xrmx:
> The changes look good to me, thanks for being patient with our feedback on this PR! One small thing I think is still missing is documenting this new annotation in the README here: https://github.com/open-telemetry/opentelemetry-operator?tab=readme-ov-file#opentelemetry-auto-instrumentation-injection.

Updated README and rebased, thanks!

I think you have been more patient with me than the other way around :sweat_smile:

swiatekm (approved):

PR #3333, opened 2024-10-07
set default address for all parsed receivers

frzifus:
**Description:** <Describe what has changed.>

**Link to tracking Issue(s):** <Issue number if applicable>

- Resolves: https://github.com/open-telemetry/opentelemetry-operator/issues/3126

**Testing:** <Describe what testing was performed and which tests were added.>

**Documentation:** <Describe the documentation added.>

PR #3337, opened 2024-10-08
Fixes bug where we tried to default config for the wrong components

jaronoff97:
**Description:** 
Patch for the recent fix in #3333

- Resolves: n/a

**Testing:** 
Unit tests added, manually tested locally.

### DIFF
--- a/apis/v1beta1/collector_webhook.go
+++ b/apis/v1beta1/collector_webhook.go
@@ -43,13 +43,14 @@ var (
 // +kubebuilder:object:generate=false
 
 type CollectorWebhook struct {
-	logger   logr.Logger
-	cfg      config.Config
-	scheme   *runtime.Scheme
-	reviewer *rbac.Reviewer
-	metrics  *Metrics
-	bv       BuildValidator
-	fips     fips.FIPSCheck
+	logger            logr.Logger
+	cfg               config.Config
+	scheme            *runtime.Scheme
+	reviewer          *rbac.Reviewer
+	metrics           *Metrics
+	bv                BuildValidator
+	fips              fips.FIPSCheck
+	disableDefaulting bool
 }
 
 func (c CollectorWebhook) Default(_ context.Context, obj runtime.Object) error {
@@ -99,6 +100,9 @@ func (c CollectorWebhook) Default(_ context.Context, obj runtime.Object) error {
 	// This results in a default state of unmanaged preventing reconciliation from continuing.
 	if len(otelcol.Spec.ManagementState) == 0 {
 		otelcol.Spec.ManagementState = ManagementStateManaged
+	}
+	if c.disableDefaulting {
+		return nil
 	}
 	return otelcol.Spec.Config.ApplyDefaults(c.logger)
 }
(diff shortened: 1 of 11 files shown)


apis/v1beta1/collector_webhook.go
```diff
@@ -100,6 +101,9 @@ func (c CollectorWebhook) Default(_ context.Context, obj runtime.Object) error {
 	if len(otelcol.Spec.ManagementState) == 0 {
 		otelcol.Spec.ManagementState = ManagementStateManaged
 	}
+	if c.disableDefaulting {
+		return nil
+	}
```
frzifus, line 106: Do we actually need this?
From my point of view we should limit the defaulting only to components we actually configured a parser for.
In that case, we use `0.0.0.0` to default if at least a parser+port is configured.

jaronoff97: i think so, i believe there will be users that plainly don't want us touching their config at all. I can imagine a scenario where a user has a custom receiver which keeps getting an endpoint to set to 0.0.0.0 which doesn't allow for setting to that host or doesn't have an endpoint... though maybe to @swiatekm's point we should just make this a featuregate and eventually enable it for everyone. I do think that makes sense too... I'm going to do that instead.

frzifus: When we limit it to known parsers, it should not be an issue if a collector contains custom components.
Regarding the general modification I would not worry to much. We also do other defaulting and so on.

In case you want to run the collector only on localhost (default behaviour),  you still can configure the endpoint explicit.

If we still want to add an option, I would prefer something to overwrite the default value. E.g.  replace `0.0.0.0` with `$POD_IP` or `localhost`.

internal/components/generic_parser.go
```diff
@@ -44,15 +44,15 @@ func (g *GenericParser[T]) GetDefaultConfig(logger logr.Logger, config interface
 		return config, nil
 	}
 
-	if g.settings.defaultRecAddr == "" {
+	if g.settings.defaultRecAddr == "" || g.settings.port == 0 {
```
frzifus, line 47: :+1:

pkg/featuregate/featuregate.go
```diff
@@ -40,6 +40,13 @@ var (
 		featuregate.WithRegisterDescription("enables feature to set GOMEMLIMIT and GOMAXPROCS automatically"),
 		featuregate.WithRegisterFromVersion("v0.100.0"),
 	)
+	// EnableConfigDefaulting is the feature gate that enables the operator to default the endpoint for known components.
+	EnableConfigDefaulting = featuregate.GlobalRegistry().MustRegister(
+		"operator.collector.default.config",
+		featuregate.StageBeta,
+		featuregate.WithRegisterDescription("enables the operator to default the endpoint for known components"),
+		featuregate.WithRegisterFromVersion("v0.110.0"),
+	)
```
frzifus, line 49: Since no e2e breaks, I assume its enabled by default? :smile:

frzifus (approved):

pkg/featuregate/featuregate.go
jaronoff97: yes :)

swiatekm (approved):

Issue #3345, opened 2024-10-11
Incorrect validation and error message of `stabilizationWindowSeconds` in autoscaler behaviour


onematchfox:
## Description

Received the following error message when trying to configure `scaleUp` policy for an `OpenTelemetryCollector` with 
`stabilizationWindowSeconds: 0`.

```
admission webhook "vopentelemetrycollectorcreateupdatebeta.kb.io" denied the request: the OpenTelemetry Spec autoscale configuration is incorrect, scaleUp should be one or more
```

There are 2 issues here. 

1) The error message is misleading in that it is not the `scaleUp` object that should be one or more, but rather the value of `stabilizationWindowSeconds`.
2) As per [k8s.io/api/autoscaling/v2/types.go](https://github.com/kubernetes/kubernetes/blob/948afe5ca072329a73c8e79ed5938717a5cb3d21/staging/src/k8s.io/api/autoscaling/v2/types.go#L186)
  > StabilizationWindowSeconds must be greater than or equal to zero and less than or equal to 3600 (one hour).

## Steps to Reproduce

Attempt to apply following manifest:

```
apiVersion: opentelemetry.io/v1beta1
kind: OpenTelemetryCollector
metadata:
  name: foo
spec:
  mode: deployment
  autoscaler:
    minReplicas: 1
    maxReplicas: 10
    targetCPUUtilization: 70
    targetMemoryUtilization: 60
    behavior:
      scaleUp:
        policies:
          - periodSeconds: 15
            type: Pods
            value: 4
          - periodSeconds: 15
            type: Percent
            value: 100
        selectPolicy: Max
        stabilizationWindowSeconds: 0
  config:
    receivers:
      otlp:
        protocols:
          grpc:
            endpoint: 0.0.0.0:4317
          http:
            endpoint: 0.0.0.0:4318
    processors:
      memory_limiter:
        check_interval: 1s
        limit_percentage: 75
        spike_limit_percentage: 15
      batch:
        send_batch_size: 10000
        timeout: 10s
    exporters:
      debug: {}
    service:
      pipelines:
        traces:
          receivers: [otlp]
          processors: [memory_limiter, batch]
          exporters: [debug]
```

## Expected Result

One should be apply to set `stabilizationWindowSeconds` to `0` (which is the default for `scaleUp` behaviour).

## Actual Result

See error message in description.

### Kubernetes Version

1.30.4

### Operator version

0.109.0

### Collector version

0.109.0

PR #3346, opened 2024-10-11
fix(collector-webhook): ensure `stabilizationWindowSeconds` validation matches `k8s.io/api/autoscaling/v2` requirements

onematchfox:
**Description:** 

This PR:
- ensures that the validation of `autoscaler.behaviour.scale[Down|Up].stabilizationWindowSeconds` matches the requirements of the upstream module [k8s.io/api/autoscaling/v2/types.go](https://github.com/kubernetes/kubernetes/blob/948afe5ca072329a73c8e79ed5938717a5cb3d21/staging/src/k8s.io/api/autoscaling/v2/types.go#L186)
- adjusts the error message to report that `scale[Down|Up].stabilizationWindowSeconds` in specific is invalid rather than referring to the entire object.

- Resolves: #3345

**Testing:** 

Existing tests have been updated to match the new requirements.


swiatekm:
Thank you for the contribution! Could you add a changelog entry for this fix? `make chlog-new` will generate a template you can fill out.

The change itself looks good to me. One thing I don't really understand is why the struct we're importing from [k8s.io/api/autoscaling](https://github.com/kubernetes/kubernetes/blob/948afe5ca072329a73c8e79ed5938717a5cb3d21/staging/src/k8s.io/api/autoscaling/v2/types.go#L186) doesn't come with the right validations in the first place. If we declared it on our own, we could easily validate it via OpenAPI in the CRD, not sure why Kubernetes itself can't do that.

onematchfox:
> Could you add a changelog entry for this fix? `make chlog-new` will generate a template you can fill out.

Done. Sorry I missed that.

> One thing I don't really understand is why the struct we're importing from [k8s.io/api/autoscaling](https://github.com/kubernetes/kubernetes/blob/948afe5ca072329a73c8e79ed5938717a5cb3d21/staging/src/k8s.io/api/autoscaling/v2/types.go#L186) doesn't come with the right validations in the first place. If we declared it on our own, we could easily validate it via OpenAPI in the CRD, not sure why Kubernetes itself can't do that.

Yeah, I'm not sure about that either. All I can add is that in trying to understand why this code was there I tracked it back to #1077.

swiatekm (approved):

PR #3364, opened 2024-10-17
Fix upgrade testing

jaronoff97:
**Description:** Fixes a bug in the upgrade testing that ran all the upgrades despite a version being set. This will now let us run each upgrades in isolation (future upgrades will not effect the current one.) i.e. a test for 0.9.0 would have to take 0.15.0 into account, this is no longer the case.


frzifus (approved):
Love it, thank you!!

PR #3368, opened 2024-10-18
Add support for merging GO* variables when existing ones are present in the manifest. 

HackToHell:
The default GOMEMLIMIT is set to 100pc which might not work with custom extensions/contrib collector which reccomends a 80pc. This allows users to set a custom value in the manifest and still not have duplicates.

### DIFF
--- a/internal/manifests/collector/container.go
+++ b/internal/manifests/collector/container.go
@@ -156,16 +156,28 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1beta1.OpenTeleme
 	}
 
 	if featuregate.SetGolangFlags.IsEnabled() {
-		envVars = append(envVars, corev1.EnvVar{
-			Name: "GOMEMLIMIT",
-			ValueFrom: &corev1.EnvVarSource{
-				ResourceFieldRef: &corev1.ResourceFieldSelector{
-					Resource:      "limits.memory",
-					ContainerName: naming.Container(),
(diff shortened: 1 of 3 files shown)


internal/manifests/collector/container.go
```diff
@@ -156,16 +156,28 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1beta1.OpenTeleme
...
 	if featuregate.SetGolangFlags.IsEnabled() {
-		envVars = append(envVars, corev1.EnvVar{
-			Name: "GOMEMLIMIT",
-			ValueFrom: &corev1.EnvVarSource{
-				ResourceFieldRef: &corev1.ResourceFieldSelector{
-					Resource:      "limits.memory",
-					ContainerName: naming.Container(),
+		skipGoMemLimit, skipGoMaxprocs := false, false
+		for _, v := range envVars {
+			if v.Name == "GOMEMLIMIT" {
```
frzifus, line 161: What does happen when the golangFlags fg is enabled and the env variable is set in the CR?

swiatekm:
I think I'd rather solve this in a more generic way. That is, ensure that we don't set any env variables the user has set. You can run into the same issue as this PR tries to address if you try to set `POD_NAME` on your own. We should probably just deduplicate the env variables at the end, giving priority to the user's. WDYT @jaronoff97 ?

Issue #3370, opened 2024-10-18
Create a ServiceMonitor for the operator metrics endpoint


iblancasa:
### Is your feature request related to a problem? Please describe.

Currently, we are creating the `ServiceMonitor` for the OpenTelemetry Operator's metrics endpoint statically from a manifest file. However, this approach has a significant limitation: users may install the operator in a namespace other than the default namespace provided in the manifest. When users specify tlsConfig, this leads to a failure in scraping metrics, as the serverName field in the ServiceMonitor depends on the namespace where the operator was deployed.

### Describe the solution you'd like

Enhance the operator to create the ServiceMonitor programmatically, ensuring it is configured automatically based on the namespace where the operator is installed. This will dynamically adjust the `serverName` and ensure metrics scraping works regardless of the installation namespace.

pavolloffay:
Cross-linking this with operator-sdk issue to handle SM by OLM https://github.com/operator-framework/operator-sdk/issues/6846

PR #2996, opened 2024-05-29
Create OpenShift dashboard to visualize Collector metrics

iblancasa:
**Description:** this PR adds the creation of OpenShift dashboards to monitor OpenTelemetry Collector instances.

- Resolves: #2995

PR #3371, opened 2024-10-18
Create ServiceMonitor for operator metrics programmatically

iblancasa:
- Resolves: #3370


internal/operator-metrics/metrics.go
```diff
@@ -0,0 +1,104 @@
...
+			Endpoints: []monitoringv1.Endpoint{
+				{
+					BearerTokenFile: "/var/run/secrets/kubernetes.io/serviceaccount/token",
+					Interval:        "30s",
+					Path:            "/metrics",
+					Scheme:          "https",
+					ScrapeTimeout:   "10s",
+					TargetPort:      &intstr.IntOrString{IntVal: 8443},
+					TLSConfig: &monitoringv1.TLSConfig{
+						CAFile: "/etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt",
```
pavolloffay, line 82: is this the configuration of OCP in-cluster monitoring stack? 

If so shall this feature be enabled only on OCP?

internal/operator-metrics/metrics.go
```diff
@@ -0,0 +1,104 @@
...
+			},
+		},
+	}
+
+	err = d.kubeClient.Create(ctx, &sm)
+	if err != nil {
+		return fmt.Errorf("error creating service monitor: %w", err)
+	}
+
+	<-ctx.Done()
```
pavolloffay, line 97: why do we have this here?

config/overlays/openshift/kustomization.yaml
```diff
@@ -1,10 +1,21 @@
 resources:
 - ../../default
 
+labels:
```
pavolloffay, line 4: why are these labels added?

iblancasa: A lot of resources were missing them.

internal/operator-metrics/metrics.go
iblancasa: A similar approach to how the OCP dashboard is created: https://github.com/open-telemetry/opentelemetry-operator/pull/2996

This is run as part of a different goroutine. When the context is done (the operator is stopped or something) this function continues, executing the deleting mechanism.

bundle/openshift/manifests/opentelemetry-operator-prometheus-rules_monitoring.coreos.com_v1_prometheusrule.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    app.kubernetes.io/managed-by: operator-lifecycle-manager
```
pavolloffay, line 5: Why is this label being added? Is it required for some use-case, does it fixes any bug? It's not being used in our bundle https://github.com/open-telemetry/opentelemetry-operator/tree/main/bundle/openshift/manifests

iblancasa: I just removed them. I thought we wanted to have them since some objects have  them but others not

pavolloffay: >  since some objects have them

What objects have them? I didn't find it in https://github.com/open-telemetry/opentelemetry-operator/tree/main/bundle/openshift/manifests

iblancasa: I just recheck the thing. It seems they are added later or I don´t know exactly where I saw them. But I removed that part of the PR anyway.

IshwarKanse:
@iblancasa I tested the PR with operator deployed in non-standard namespaces and enabling the in-cluster monitoring by adding label openshift.io/cluster-monitoring="true". The metrics are scraped by the in-cluster monitoring as well as the targets are shown up for the servicemonitors that are created. 

We already have added [test case](https://github.com/open-telemetry/opentelemetry-operator/tree/main/tests/e2e-openshift/monitoring) to check the metrics exported by prometheus exporter and using the user workload monitoring stack. I'll add the operator monitoring as well to that test case and verify the metrics using the in-cluster monitoring stack.

iblancasa:
Thanks @IshwarKanse !

bundle/openshift/manifests/opentelemetry-operator-prometheus_rbac.authorization.k8s.io_v1_role.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
```
pavolloffay, line 5: are these two annotations needed?

internal/operator-metrics/metrics.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return OperatorMetrics{}, err
+	}
+
+	return OperatorMetrics{
+		kubeClient: kubeClient,
+	}, nil
+}
+
+func (d OperatorMetrics) Start(ctx context.Context) error {
```
pavolloffay, line 64: nit: maybe change `d` to something more aligned with the `OperatorMetrics` name?

internal/operator-metrics/metrics.go
```diff
@@ -0,0 +1,142 @@
...
+			SafeTLSConfig: monitoringv1.SafeTLSConfig{
+				ServerName: &serviceName,
+			},
+		}
+	} else {
+		t := true
+		tlsConfig = &monitoringv1.TLSConfig{
+			SafeTLSConfig: monitoringv1.SafeTLSConfig{
+				// kube-rbac-proxy uses a self-signed cert by default
+				InsecureSkipVerify: &t,
```
pavolloffay, line 87: I think we talked about this internally as a bug that we should not be using the `InsecureSkipVerify` it is also not present here https://github.com/os-observability/konflux-opentelemetry/blob/main/bundle-patch/manifests/opentelemetry-operator-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml#L22

main.go
```diff
@@ -435,6 +436,16 @@ func main() {
 				setupLog.Error(err, "Error init CRD metrics")
 			}
 
+			if cfg.PrometheusCRAvailability() == prometheus.Available {
```
pavolloffay, line 439: Why is this in the block for ` os.Getenv("ENABLE_WEBHOOKS") != "false" ` The operator metrics should be enabled regardless of the webhooks.

pavolloffay: There is `addDependencie` where we call ` mgr.Add(`

internal/operator-metrics/metrics.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// nolint #nosec
+	// bearerTokenFile declares the path for bearer token file for service monitors.
+	bearerTokenFile string = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+
+	// openshiftInClusterMonitoringNamespace declares the namespace for the OpenShift in-cluster monitoring.
+	openshiftInClusterMonitoringNamespace string = "openshift-monitoring"
+)
+
+type OperatorMetrics struct {
```
pavolloffay, line 49: could you please do type assertion that it implements the Runnable interface? to make it clear what contract it fulfills

internal/operator-metrics/metrics.go
```diff
@@ -0,0 +1,142 @@
...
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	// namespaceFile is the path to the namespace file for the service account.
+	namespaceFile = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+
+	// caBundleConfigMap declares the name of the config map for the CA bundle.
+	caBundleConfigMap string = "serving-certs-ca-bundle"
```
pavolloffay, line 36: the types can be removed

internal/operator-metrics/metrics.go
iblancasa: If you review the code, you will see this will be only executed if the ConfigMap containing the CA doesn´t exist. This cover non-OpenShift scenarios.

internal/operator-metrics/metrics.go
pavolloffay: the comment is resolved with 👍🏼 , but the type assertion seems to be mising.

internal/operator-metrics/metrics_test.go
```diff
@@ -0,0 +1,103 @@
...
+	metrics := OperatorMetrics{kubeClient: client}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	errChan := make(chan error)
+	go func() {
+		errChan <- metrics.Start(ctx)
+	}()
+
+	// Wait a bit to allow the Start method to run
+	time.Sleep(100 * time.Millisecond)
```
pavolloffay, line 69: this is in general a bad practice to use static sleep statements (e.g. might become flaky and it usually unnecessary  adds to test time.

iblancasa: Any suggestion about how to improve this?

main.go
iblancasa: Oh I see now. Thanks!

internal/operator-metrics/metrics_test.go
iblancasa: I just saw a way.

internal/operator-metrics/metrics.go
iblancasa: Now it should be there.

iblancasa:
@pavolloffay anything else?

pavolloffay (approved):

Issue #1814, opened 2023-06-07
Setting JAVA_TOOL_OPTIONS via configmap is not possible (replaced,ignored)


5V715:
when using the operator in combination with ENVVAR from ConfigMap the original value of JAVA_TOOL_OPTIONS is ignored and therefore lost. see the append yaml and comments.

``` yaml
Containers:
  some-app:
    Container ID: xxx
    Image: xxx
    Image ID: xxx
    Port: 8080/TCP
    Host Port: 0/TCP
    ...
    Environment Variables from:
      some-app-config  ConfigMap  Optional: false # (<--) this also contains a value for JAVA_TOOL_OPTIONS
    Environment:
      AWS_METADATA_SERVICE_TIMEOUT: 5
      AWS_METADATA_SERVICE_NUM_ATTEMPTS: 20
      JAVA_TOOL_OPTIONS: -javaagent:/otel-auto-instrumentation/javaagent.jar # <-- replaces the old value from config map
```

pavolloffay:
Correct, the implementation at the moment does not check env vars defined in the config map. 

No objections on supporting it.

pavolloffay:
Actually in this case I am not sure what would be the behaviour. I would say the operator should make changes to the config map - the problem would be to deal with reverting those changes.

Maybe it could copy the env var value from the config map if the modification is needed?

KitSutliff:
Here’s a proposed implementation and a quick question.  With a little more info I would be happy to knock this out.

Proposed implementation:
1. When the operator receives a pod we will check for config maps that reference the given pod
2. If a config map is present operator will look for JAVA_TOOL_OPTS env var
3. I there is a JAVA_TOOL_OPTS env var in the ConfigMap, that will be used to populate the JAVA_TOOL_OPTS env var in the container

Should the value form the ConfigMap overwrite the value form the container or should it be appended in some way?

KitSutliff:
Awesome, thank you for the clarification!  I'll start knocking this out.

pavolloffay:
Perfect, can we make it generic enough and reuse for other/all variables? 


We should as well document the behavior and that these env vars cannot be hardcoded in the image https://github.com/open-telemetry/opentelemetry-operator/issues/1884

jaronoff97:
[...]

Issue #1393, opened 2023-01-24
NodeJS Auto instrumentation overrides NODE_OPTIONS when defined from ConfigMap


edenkoveshi:
## Description
I have a node container for which I use the env var `NODE_OPTIONS`
However, it is defined via a ConfigMap under `containers[0].envFrom` and not `containers[0].env`
The NodeJS auto instrumentation modifies the pod to and sets `NODE_OPTIONS` value under `containers[0].env` to the default value (`--require /otel-auto-instrumentation/autoinstrumentation.js`) instead of concatenating it like it should.
I have found [https://github.com/open-telemetry/opentelemetry-operator/issues/1094](this issue). To my understanding, the operator checks only the `containers.env` property and not `envFrom`, which results in this conjestion.
Though I know I can manually set the `NODE_OPTIONS` env under `containers.env` and it would probably work, I need my env variables to be managed with `envFrom`, and I don't think overriding these vars is the right behaviour for the operator.

## Steps to Reproduce
Create a NodeJS pod and a ConfigMap containing `NODE_OPTIONS` key and some value in it.
Instrument the pod using the operator auto instrumentation.

## Expected Result
The instrumentation `NODE_OPTIONS` should be concatenated with mine.

## Actual Result
The instrumentation `NODE_OPTIONS` overrides mine.

shovelend:
We need to enable the `--openssl-legacy-provider` option for some of our services since when auto-instrumentation is enabled, legacy encryption/decryption algorithms don't work. 
However, the `NODE_OPTIONS` environment variable is overridden by the injector, so we have to pass the option down to the `node` command manually inside the pod. Would be a lot better to have the `NODE_OPTIONS` env vars concatenated instead.

oldium:
I prepared a fix for this in #3373

PR #3373, opened 2024-10-19
Add handling of environment variables from ConfigMaps and Secrets

oldium:

Added detection and reading of environment variables defined in PODs resources like ConfigMaps and Secrets. Supports both `envFrom` and `valueFrom` forms.

- Resolves:
  - #1393
  - #1814

**Testing:**

Added `kubebuilder` tests for new code, tested by deploying the generated image to OpenShift cluster.

This is an enhancement, which can be read also as a bug fix (missing feature), so I have not added any documentation.

**Implementation notes/questions:**

1. The `ConfigMap` support is in first commit, the `Secret` support is in the second commit. They work the same, they both support “if exists” checks and are able to extend the existing value (append) into a newly created (or updated) environment variable.
2. I was not sure about the `Container` class naming (or have some common methods prefix?), so please suggest how the main class/file should be named - I can change it easily. The class handles everything regarding environment variables on the container currently.
3. I was not sure about the usage of namespace name. The `sdkInjector.inject` method gets one in arguments, but Apache and Nginx auto-instrumentations are ignoring it and taking namespace name from POD definition or from PODs resource map, which looks wrong (at minimum this is suspicious). I took the namespace name from the POD and if not found, I fall back to the one supplied as a parameter to `inject` method.
4. I was not sure about the reading of Secrets (second commit) - whether it is secure to handle them fully. The implementation is as full as for ConfigMaps, so the values can be detected (“if exists” checks) and extended (“append when exists”) into a new or updated environment variable (so Secret is converted to env var!). If necessary, I can change the implementation to just recognise them (record the existence) but fail when the value should be extended.

### DIFF
--- a/.chloggen/enhancement-envs-from-configmap.yaml
+++ b/.chloggen/enhancement-envs-from-configmap.yaml
@@ -1,0 +1,19 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: auto-instrumentation
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add support for detecting and reading environment variables from POD's ConfigMap resources.
+
+# One or more tracking issues related to the change
+issues: [1393, 1814]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Now the auto-instrumentation will see the environment variables defined via POD's `env.valueFrom.configMapKeyRef` and
+  `envFrom.configMapRef` fields, so the auto-instrumentation will be able to prevent overriding them, or it will be able
+  to extend the existing definition.
--- a/.chloggen/enhancement-envs-from-secret.yaml
+++ b/.chloggen/enhancement-envs-from-secret.yaml
@@ -1,0 +1,19 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: auto-instrumentation
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add support for detecting and reading environment variables from POD's Secret resources.
+
+# One or more tracking issues related to the change
+issues: [1393, 1814]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Now the auto-instrumentation will see the environment variables defined via POD's `env.valueFrom.secretKeyRef` and
+  `envFrom.secretRef` fields, so the auto-instrumentation will be able to prevent overriding them, or it will be able to
+  extend the existing definition.
--- a/pkg/instrumentation/apachehttpd.go
+++ b/pkg/instrumentation/apachehttpd.go
@@ -59,19 +59,13 @@ const (
 	6) Inject mounting of volumes / files into appropriate directories in application container
 */
 
-func injectApacheHttpdagent(_ logr.Logger, apacheSpec v1alpha1.ApacheHttpd, pod corev1.Pod, useLabelsForResourceAttributes bool, index int, otlpEndpoint string, resourceMap map[string]string) corev1.Pod {
+func injectApacheHttpdagent(_ logr.Logger, apacheSpec v1alpha1.ApacheHttpd, pod corev1.Pod, useLabelsForResourceAttributes bool, container Container, otlpEndpoint string, resourceMap map[string]string) corev1.Pod {
 
 	volume := instrVolume(apacheSpec.VolumeClaimTemplate, apacheAgentVolume, apacheSpec.VolumeSizeLimit)
 
-	// caller checks if there is at least one container
-	container := &pod.Spec.Containers[index]
-
 	// inject env vars
 	for _, env := range apacheSpec.Env {
-		idx := getIndexOfEnv(container.Env, env.Name)
-		if idx == -1 {
-			container.Env = append(container.Env, env)
-		}
+		container.appendEnvVarIfNotExists(&pod, env)
 	}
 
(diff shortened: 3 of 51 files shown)


oldium:
Tested with custom built image `oldium/opentelemetry-operator:configmaps-2024-10-20-1` on our test cluster.

oldium:
Handling Secrets requires updated permissions, otherwise it ends up with:

```
failed to list *v1.Secret: secrets is forbidden: User "system:serviceaccount:opentelemetry-operator-system:opentelemetry-operator-controller-manager" cannot list resource "secrets" in API group "" at the cluster scope
```

I am really in doubt on how to handle Secrets, whether to (1) read them (needs permission update) and put into env vars when appended, or (2) to read them (needs permission update), but only record existence and fail when appending, or (3) ignore Secrets completely as it is now.

Please suggest what to do. Thanks.

oldium:
So the Secrets code works fine with the following kustomize patch to the Release asset:

```
patches:
- patch: |-
    - op: add
      path: /rules/2/resources/1
      value: secrets
    - op: add
      path: /rules/2/verbs/2
      value: get
  target:
    kind: ClusterRole
    name: opentelemetry-operator-manager-role
```

jaronoff97 (commented):
can you adjust an E2E test to ensure this logic works as expected?

oldium:
> can you adjust an E2E test to ensure this logic works as expected?

You mean to add e2e tests for ConfigMaps and Secrets to ensure it works, right?

pavolloffay:
@oldium please take a look at #3379 , it's an alternative solution that is much simpler.

oldium:
@pavolloffay If I read it correctly, this is a solution for appending to an existing value. I am solving wider issue - my code is able to actually read ConfigMaps and Secrets, so when `OTEL_SERVICE_NAME` is defined in `ConfigMap`, it is not overwritten by OTEL Operator.

oldium:
So, to summarise. Your PR solves issue for the case when the environemntal variable is defined via `valueFrom`. My PR solves issue for the case when the environmental variable is defined either via `envFrom` or `valueFrom`.

oldium:
We use ConfigMaps in base directory and customize them with `kubectl kustomize` in overlay directories for all environments and we do it for all our services. We set all environmental properties via ConfigMaps, they are easily overridable like this. Currently we are forced to set OTEL-specific environmental variables in Deployments and Argo-Rollouts via patches, because the value in ConfigMap is ignored. We only would like to use ConfigMaps for environmental variables, because this is feature of Kubernetes after all.

oldium:
> As #3379 shows, we don't actually need to read external ConfigMaps and Secrets to achieve the latter.

You cannot put the `OTEL_SERVICE_NAME` and other environmental values in single (overridable) `ConfigMap` without directly referencing all the variables separately for #3379 to work.

Kubernetes can load the whole ConfigMap with `envFrom` field, this is much easier to handle. We have over 100 services running in 50 environments (test, staging, production), so actually 100 base directories and 5000 overrides, so having ConfigMaps (which is simple key-value text file) referenced by `envFrom` in the service's Deployment/Rollout in base directory is simply much easier to handle than taking care of multiple `valueFrom` for each environmental value loaded from ConfigMap.

This is a long-standing maintaintenance pain for us - this Kubernetes feature works otherwise fine, but OTEL Operator misses it. Please consider adding this feature.

swiatekm:
Allright, that makes sense. I'd like to support what you're asking for, but I'd really like to avoid all the machinery in this PR. We'll discuss it during our next SIG meeting and hopefully find some alternative. For the Java and Node option variables specifically, I think #3379 can solve it as-is:

1. You set `JAVA_TOOL_OPTIONS` in your ConfigMap, which you attach via `envFrom`.
2. In the container `env`, you add `JAVA_TOOL_OPTIONS: $(JAVA_TOOL_OPTIONS)`
3. Operator sees the variable defined and merges it with its own options.

A slightly annoying workaround, but it should be easy to add to your kustomize base.

oldium:
> Allright, that makes sense. I'd like to support what you're asking for, but I'd really like to avoid all the machinery in this PR. We'll discuss it during our next SIG meeting and hopefully find some alternative. For the Java and Node option variables specifically, I think #3379 can solve it as-is:
> 
> 1. You set `JAVA_TOOL_OPTIONS` in your ConfigMap, which you attach via `envFrom`.
> 2. In the container `env`, you add `JAVA_TOOL_OPTIONS: $(JAVA_TOOL_OPTIONS)`
> 3. Operator sees the variable defined and merges it with its own options.
> 
> A slightly annoying workaround, but it should be easy to add to your kustomize base.

For like 100 services with 50 overlays each, it is better to have the values present just once. It is also easier to grep for the value afterwards when necessary. So if I have to choose between two JAVA_TOOL_OPTIONS (one placeholder value in the deployment, one actual value in the ConfigMap), I will choose the direct value in the deployment and skip the ConfigMap altogether. That is because not all services have JAVA_TOOL_OPTIONS, not all services have OTEL_SERVICE_NAME and so on - some have the value just in test overlays, some in base, some override it just for test environments and the like. It is really per-service and per-environment decision.

Before OpenTelemetry we used just ConfigMaps and were “happy” (almost, you know).

oldium:
I have rebased the code and added e2e tests.

oldium:
> > I am solving wider issue - my code is able to actually read ConfigMaps and Secrets, so when for example OTEL_SERVICE_NAME is defined in ConfigMap (via envFrom), it is not overwritten by OTEL Operator.
> 
> This is the current behavior of the operator. It checks all env vars and if any `OTEL_` is specified it will honor it. The operator does not need to look into the config map for this.

This is not how it works. If you define `OTEL_SERVICE_NAME` in ConfigMap, which is referenced by `envFrom` from the Deployment, the OTEL Operator creates a new `OTEL_SERVICE_NAME` env var in a POD, because it does not see anything what is defined inside the ConfigMap. OTEL Operator only sees env vars defined in the Deployment, i.e. the POD directly.

My colleague just faced the issue (again, sigh!) and had to move the `OTEL_SERVICE_NAME` environment variable from ConfigMap into Deployment.

pavolloffay:
I see what you mean https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables 

```
apiVersion: v1
kind: Pod
metadata:
  name: dapi-test-pod
spec:
  containers:
    - name: test-container
      image: registry.k8s.io/busybox
      command: [ "/bin/sh", "-c", "env" ]
      envFrom:
      - configMapRef:
          name: special-config
  restartPolicy: Never
```

This will define all configmap keys as env vars.

pavolloffay:
@oldium we are concerned with this PR and we would like to avoid reading the configmaps in the operator. 

There is issue for discussing alternative approaches https://github.com/open-telemetry/opentelemetry-operator/issues/3392

Issue #552, opened 2021-11-17
Test reporting of data in e2e Instrumentation tests


pavolloffay:
The e2e instrumentation tests verify the structure of the workload deployment (if all env vars are set etc.). However the test should execute an endpoint on the test app that will trigger span generation and then the test should verify if the span arrived to the collector or some mock storage.

@anuraaga the integration tests in the java auto-instrumentation are dosing something similar, is there something we could reuse from that project - e.g. the mock storage?

anuraaga:
I think this might be a good segway into filing an OTEP or something for a project-wide black-box instrumentation testing suite. I can't think of anything we would implement in an e2e test that wouldn't apply to every language at the same time - sending a request, verifying sent OTLP is a very common concern.

jpkrohling:
> then the test should verify if the span arrived to the collector or some mock storage

At the collector would be easy, as we have metrics already with the number of spans seen by each component, IIRC.

pavolloffay:
I would prefer to get the reported data. This is what the java auto-instrumentation tests do. The SDK reports to a collector and the collector reports to a service with in-memory storage that can be queried.

jaronoff97:
@pavolloffay what's the load to make this work? We now do something similar with the target allocator e2e tests

PR #3381, opened 2024-10-22
Actually e2e test java

jaronoff97:
**Description:** 
Uses chainsaw to seed and then check for data in the e2e test for the java instrumentation only right now

**Link to tracking Issue(s):** #552

- Resolves: #issue-number

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -473,7 +473,7 @@ KUSTOMIZE_VERSION ?= v5.0.3
 CONTROLLER_TOOLS_VERSION ?= v0.16.1
 GOLANGCI_LINT_VERSION ?= v1.57.2
 KIND_VERSION ?= v0.20.0
-CHAINSAW_VERSION ?= v0.2.8
+CHAINSAW_VERSION ?= v0.2.11
 
 .PHONY: install-tools
 install-tools: kustomize golangci-lint kind controller-gen envtest crdoc kind operator-sdk chainsaw
--- a/tests/e2e-instrumentation/instrumentation-java-multicontainer/00-install-collector.yaml
+++ b/tests/e2e-instrumentation/instrumentation-java-multicontainer/00-install-collector.yaml
@@ -20,3 +20,6 @@ spec:
         traces:
           receivers: [otlp]
           exporters: [debug]
+        metrics:
+          receivers: [otlp]
+          exporters: [debug]
--- a/tests/e2e-instrumentation/instrumentation-java-multicontainer/00-install-instrumentation.yaml
+++ b/tests/e2e-instrumentation/instrumentation-java-multicontainer/00-install-instrumentation.yaml
@@ -9,11 +9,9 @@ spec:
     - name: OTEL_EXPORTER_OTLP_ENDPOINT
       value: http://localhost:4317
     - name: OTEL_EXPORTER_OTLP_TIMEOUT
-      value: "20"
+      value: "20000"
     - name: OTEL_TRACES_SAMPLER
-      value: parentbased_traceidratio
-    - name: OTEL_TRACES_SAMPLER_ARG
-      value: "0.85"
+      value: always_on
     - name: SPLUNK_TRACE_RESPONSE_HEADER_ENABLED
       value: "true"
   exporter:
--- a/tests/e2e-instrumentation/instrumentation-java-multicontainer/chainsaw-test.yaml
+++ b/tests/e2e-instrumentation/instrumentation-java-multicontainer/chainsaw-test.yaml
@@ -49,3 +49,37 @@ spec:
     catch:
       - podLogs:
           selector: app=my-java-multi
+  - name: "wait for app to be ready"
+    try:
+      - sleep:
+          duration: 5s
+  - name: Make a request to the app
+    try:
+      - script:
+          content: |
+            #!/bin/bash
+            # set -ex
+            pod=$(kubectl get pods -n $NAMESPACE | awk '{print $1}' | tail -n 1)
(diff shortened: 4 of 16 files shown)


tests/e2e-instrumentation/instrumentation-java-multicontainer/00-install-instrumentation.yaml
```diff
@@ -9,11 +9,9 @@ spec:
     - name: OTEL_EXPORTER_OTLP_ENDPOINT
       value: http://localhost:4317
     - name: OTEL_EXPORTER_OTLP_TIMEOUT
-      value: "20"
+      value: "20000"
```
jaronoff97, line 12: this is to account for the collector not being ready fast enough

tests/e2e-instrumentation/instrumentation-java-multicontainer/00-install-instrumentation.yaml
```diff
@@ -9,11 +9,9 @@ spec:
...
     - name: OTEL_TRACES_SAMPLER
-      value: parentbased_traceidratio
-    - name: OTEL_TRACES_SAMPLER_ARG
-      value: "0.85"
+      value: always_on
```
jaronoff97, line 14: don't want our test data to be sampled

tests/e2e-instrumentation/instrumentation-java-tls/00-install-collector.yaml
```diff
@@ -41,3 +41,6 @@ spec:
         traces:
           receivers: [otlp]
           exporters: [debug]
+        metrics:
```
jaronoff97, line 44: i was going to attempt the TLS test, but it seemed way too hard to do on a first-pass

tests/e2e-instrumentation/instrumentation-java-multicontainer/00-install-collector.yaml
```diff
@@ -20,3 +20,6 @@ spec:
         traces:
           receivers: [otlp]
           exporters: [debug]
+        metrics:
```
jaronoff97, line 23: java auto inst can send metrics too, why not accept them :)

Makefile
```diff
@@ -473,7 +473,7 @@ KUSTOMIZE_VERSION ?= v5.0.3
 CONTROLLER_TOOLS_VERSION ?= v0.16.1
 GOLANGCI_LINT_VERSION ?= v1.57.2
 KIND_VERSION ?= v0.20.0
-CHAINSAW_VERSION ?= v0.2.8
+CHAINSAW_VERSION ?= v0.2.11
```
pavolloffay, line 476: this version does not work on go 1.22. I would wait with updating before we upgrade the go version

tests/e2e-instrumentation/instrumentation-java-multicontainer/00-install-collector.yaml
pavolloffay: +1

tests/e2e-instrumentation/instrumentation-java-multicontainer/chainsaw-test.yaml
```diff
@@ -49,3 +49,37 @@ spec:
...
+      - sleep:
+          duration: 5s
+  - name: Make a request to the app
+    try:
+      - script:
+          content: |
+            #!/bin/bash
+            # set -ex
+            pod=$(kubectl get pods -n $NAMESPACE | awk '{print $1}' | tail -n 1)
+            kubectl get --raw /api/v1/namespaces/$NAMESPACE/pods/${pod}:8080/proxy/
```
pavolloffay, line 63: this is new to me :) 

@IshwarKanse does it work on OCP?

Makefile
swiatekm: If that's a concern for us, we should install it by downloading the release binary instead. Our supported Go version shouldn't get in the way of tooling upgrades.

tests/e2e-instrumentation/instrumentation-java-multicontainer/00-install-instrumentation.yaml
swiatekm: What happens if it's not ready? Shouldn't this work correctly either way?

tests/e2e-instrumentation/instrumentation-java-multicontainer/chainsaw-test.yaml
swiatekm: Looking at this, I really wonder if we shouldn't use e2e_framework and/or terratest for these tests. The main problem is really getting data out of the cluster, as we don't have a remote we can easily query. I can think of a few solutions to do this in kind, but they won't work on OpenShift. It'd be best if these could simply be Go tests we could run on a host against any cluster.

IshwarKanse: This can be implemented natively using Chainsaw 
and avoid using the script. https://github.com/kyverno/chainsaw/blob/main/testdata/e2e/examples/proxy/chainsaw-test.yaml 

We were also discussing on implementing some kind of custom store in Chainsaw which gathers all the signals and would be useful for asserting. https://docs.google.com/document/d/1G4uDMrC3okF2q2qyNYTg2qj7yxNBjSTGgn8pBEXEeiw/edit#heading=h.8mm7ay5rgjs6 

@eddycharly Do we have any decision on what custom store we can use for testing observability data. ?

jaronoff97:
superseded by Ishwar's recent PRs.

Issue #3155, opened 2024-07-22
Missing RBAC permissions in receivers and extensions


grandwizard28:
### Component(s)

collector

### Describe the issue you're reporting

There seems to be missing rbac permissions for the collector when trying to use receivers which query the K8s API.
There have been 2 instances which I have noticed:

1. The `k8sobserver` extension does not get the required rbac permissions. If I add the k8sattributes processor, then the permissions seem to be added.
2. The `kubeletstats` receiver is throwing a 403 forbidden when trying to query node metrics:
```
kubelet request GET https://xxxxxxx:10250/stats/summary failed - \"403 Forbidden\", response: \"Forbidden (user=system:serviceaccount:observability:<name-of-collector>, verb=get, resource=nodes, subresource=stats)\"", "scraper": "kubeletstats"
```

#### Expectation
The operator should automatically add permissions for the extensions/receivers to work. If not, what is the recommended method to add additional permissions to the collector?

#### Setup
Running a bunch of collectors as dameonset managed by the operator

jaronoff97:
We don't yet support automatic RBAC for the kubelet stats receiver so for now you'll need to follow the steps in the [kubelet stats documentation](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/kubeletstatsreceiver/README.md#role-based-access-control). That being said, this is definitely something we could/should do, i'll add it to our backlog :)

grandwizard28:
I mentioned thek8sobserver extension also in this ticket.

jaronoff97:
Maybe just change this issue's title to include it, thank you.

iblancasa:
I'll work on this as soon as #3206 is merged :)

PR #3388, opened 2024-10-24
Add automatic RBAC creation for kubeletstats receiver

iblancasa:
- Resolves: #3155

**Testing:** added unit and E2E tests.


internal/components/receivers/scraper_test.go
```diff
@@ -29,7 +29,6 @@ func TestScraperParsers(t *testing.T) {
 		defaultPort  int
 	}{
 		{"prometheus", "__prometheus", 0},
-		{"kubeletstats", "__kubeletstats", 0},
```
jaronoff97, line 32: this should still be a scraper right?

tests/e2e-automatic-rbac/receiver-kubeletstats/01-assert.yaml
jaronoff97, line 1: should we just put the tests from this folder into the extra-permissions-operator test?

internal/components/receivers/kubeletstats.go
```diff
@@ -0,0 +1,85 @@
...
+		config.Metrics.K8sPodCPURequestUtilization.Enabled,
+		config.Metrics.K8sPodMemoryLimitUtilization.Enabled,
+		config.Metrics.K8sPodMemoryRequestUtilization.Enabled,
+	}
+	for _, metric := range metrics {
+		if metric {
+			prs = append(prs, nodesProxyPr)
+			return prs, nil
+		}
+	}
```
swiatekm, line 83: This is a bit less efficient, but imo more clear. WDYT?

```suggestion
        resourceMetricEnabled := false
	metricsEnabled := []bool{
		config.Metrics.K8sContainerCPULimitUtilization.Enabled,
		config.Metrics.K8sContainerCPURequestUtilization.Enabled,
		config.Metrics.K8sContainerMemoryLimitUtilization.Enabled,
		config.Metrics.K8sContainerMemoryRequestUtilization.Enabled,
		config.Metrics.K8sPodCPULimitUtilization.Enabled,
		config.Metrics.K8sPodCPURequestUtilization.Enabled,
		config.Metrics.K8sPodMemoryLimitUtilization.Enabled,
		config.Metrics.K8sPodMemoryRequestUtilization.Enabled,
	}
	for _, metricEnabled := range metricsEnabled {
		resourceMetricEnabled = resourceMetricEnabled || metricEnabled
	}
	
	if len(config.ExtraMetadataLabels) > 0 || resourceMetricEnabled {
	         prs = append(prs, nodesProxyPr)
	}
	
	return prs, nil
```

tests/e2e-automatic-rbac/receiver-kubeletstats/01-assert.yaml
iblancasa: No. That folder contains some patchs for `kustomize` for the E2E tests.

internal/components/receivers/scraper_test.go
iblancasa: I don't think so. It was removed from being a scraper in `internal/components/receivers/helpers.go` and now is treated differently.

internal/components/receivers/kubeletstats.go
iblancasa: Actually, it looks more confusing to me 😅. The current approach is like "if this thing is defined, just return". And now you need to read the full thing... but if you have a strong opinion about this, I can change it.

swiatekm (approved):

internal/components/receivers/scraper_test.go
iblancasa: Reverted

jaronoff97 (approved):

Issue #1842, opened 2023-06-15
[target allocator] Support the `Probe` and `ScrapeConfig` Prometheus Operator CRDs


matej-g:
Off the back of the suggestion in https://github.com/open-telemetry/opentelemetry-operator/pull/1710#discussion_r1226753826

In the Prometheus Operator CRDs, there are couple more CRDs that it would be nice to support:
- [`Probe`](https://prometheus-operator.dev/docs/operator/design/#probe)
- [`ScrapeConfig`](https://github.com/prometheus-operator/prometheus-operator/pull/5279)

Actions to take:
- Evaluate if we want to support both
- Implement

jaronoff97:
I'm definitely in support of this, it also should be much easier to do now that we use the upstream prom-operator methods. @yuriolisa was interested in testing this out.

PR #3394, opened 2024-10-25
Scrape config and probe support in target allocator

jaronoff97:
**Description:** Enables support for pulling scrape configs and probes in the target allocator

- Resolves: #1842 

**Testing:** Unit


jaronoff97:
@swiatekm how do we want to handle preventing this being a breaking change https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1beta1/targetallocator_rbac.go#L31-L33, i.e. the TA starts without the rbac it needs and crashes?

swiatekm:
> @swiatekm how do we want to handle preventing this being a breaking change https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1beta1/targetallocator_rbac.go#L31-L33, i.e. the TA starts without the rbac it needs and crashes?

We should treat the fact that it crashes as a bug and ensure it keeps running.  I thought we had an issue for this somewhere.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/open-telemetry/opentelemetry-operator
 
-go 1.22.0
+go 1.23
+
+toolchain go1.23.2
```
swiatekm, line 5: FYI, as per the Go support policy, we can't set 1.23 as the minimum until 1.24 is released.

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/open-telemetry/opentelemetry-operator
 
 go 1.22.0
 
+toolchain go1.22.5
```
swiatekm, line 5: If we're going to suggest a toolchain, how about 1.22.8, which is the latest in the 1.22 line?

cmd/otel-allocator/watcher/promOperator_test.go
```diff
@@ -946,7 +1080,7 @@ func TestNamespaceLabelUpdate(t *testing.T) {
 
 	select {
 	case <-events:
-	case <-time.After(time.Second):
+	case <-time.After(3 * time.Second):
```
swiatekm, line 1083: Can we standardize on 5 seconds for these kinds of tests? They usually complete much faster, and having to wait 5 seconds for a fail is a fair price to pay for lack of random flaky failures on GHA runners.

swiatekm (approved):
LGTM, minor nitpicks.

frzifus (approved):

PR #3395, opened 2024-10-26
HTTPD instrumentation: Safe modification of httpd.conf

CostisC:
**Description:** 
The `Include /usr/local/apache2/conf/opentemetry_agent.conf` may be appended on the same line with the last httpd.conf statement, if there is not a blank last line in httpd.conf. This results in a corrupted httpd configuration, and the service will not start.   
 
It is safer that an empty line comes before the addition of `Include ...opentemetry_agent.conf`, as a protection measure against such cases.  

**Link to tracking Issue(s):** N/A

**Testing:** Change httpd.conf to not have an blank last line, and then restart the instrumented httpd pod. It fails to start up.  

**Documentation:** N/A


swiatekm:
Thank you for the contribution! Could you add a changelog entry to this PR? Running `make chlog-new` in the operator root directory will create a template file in `.chloggen` that you can fill in.

swiatekm (approved):

PR #3402, opened 2024-10-28
Generate only TargetAllocator CR from Collector CR

swiatekm:
**Description:**

Currently when the target allocator is enabled on the collector CR, the collector reconciler generates the target allocator manifests directly. This PR introduces a feature flag that causes it to generate a Target Allocator CR instead, and lets the Target Allocator reconciler take care of the rest.

All the changes are gated behind the flag, including actually enabling the TA controller and webhook. Users who don't set the flag are completely unaffected.

I'm not adding a changelog entry for this, because it's only really meant for testing. I'll add the entry alongside the documentation when we actually enable the controller and webhook unconditionally.

- Related to: #2422

**Testing:**
Added: 
* new unit tests
* new builder tests for cases we're already checking
* controller tests using envtest
* run e2e tests with the flag enabled

### DIFF
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -35,6 +35,7 @@ import (
 	"github.com/open-telemetry/opentelemetry-operator/internal/manifests/collector"
 	"github.com/open-telemetry/opentelemetry-operator/internal/manifests/opampbridge"
 	"github.com/open-telemetry/opentelemetry-operator/internal/manifests/targetallocator"
+	"github.com/open-telemetry/opentelemetry-operator/pkg/featuregate"
 )
 
 func isNamespaceScoped(obj client.Object) bool {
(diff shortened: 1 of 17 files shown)


controllers/targetallocator_controller.go
```diff
@@ -55,17 +63,42 @@ type TargetAllocatorReconcilerParams struct {
...
+	return p, nil
+}
+
+func (r *TargetAllocatorReconciler) getCollector(ctx context.Context, instance v1alpha1.TargetAllocator) (*v1beta1.OpenTelemetryCollector, error) {
+	var collector v1beta1.OpenTelemetryCollector
+	ownerReferences := instance.GetOwnerReferences()
+	collectorIndex := slices.IndexFunc(ownerReferences, func(reference metav1.OwnerReference) bool {
+		return reference.Kind == "OpenTelemetryCollector"
+	})
+	if collectorIndex != -1 {
```
jaronoff97, line 90: in the future this is where we'll do the label selector get i assume?

controllers/targetallocator_controller.go
```diff
@@ -121,32 +153,58 @@ func (r *TargetAllocatorReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
-		builder.Owns(&monitoringv1.PodMonitor{})
+		ctrlBuilder.Owns(&monitoringv1.ServiceMonitor{})
+		ctrlBuilder.Owns(&monitoringv1.PodMonitor{})
 	}
 
-	return builder.Complete(r)
+	// watch collectors which have embedded Target Allocator enabled
+	// we need to do this separately from collector reconciliation, as changes to Config will not lead to changes
+	// in the TargetAllocator CR
+	ctrlBuilder.Watches(
```
jaronoff97, line 187: didn't realize the controller has this capability, very cool :)

internal/manifests/mutate.go
```diff
@@ -56,6 +58,7 @@ var (
 // - HorizontalPodAutoscaler
 // - Route
 // - Secret
+// - TargetAllocator
```
jaronoff97, line 61: to confirm my understanding here... we need this because the collector CR will directly create the TA CR and that will be part of the mutate loop now?

controllers/common.go
```diff
@@ -59,22 +60,23 @@ func BuildCollector(params manifests.Params) ([]client.Object, error) {
...
-			Recorder:        params.Recorder,
-			Log:             params.Log,
-			Config:          params.Config,
-			Collector:       &params.OtelCol,
-			TargetAllocator: *params.TargetAllocator,
-		}
-		taResources, err := BuildTargetAllocator(taParams)
-		if err != nil {
-			return nil, err
+	if !featuregate.CollectorUsesTargetAllocatorCR.IsEnabled() {
```
jaronoff97, line 63: can you add a comment here specifying why this is negated?

jaronoff97 (approved):
nothing blocking, just a few minor questions/suggestions

controllers/targetallocator_controller.go
swiatekm: Yep, that's coming in a follow-up.

internal/manifests/mutate.go
swiatekm: Yes, exactly.

controllers/common.go
swiatekm: Sure. It's because here we have the collector builder also build the TA manifests. It should only do that if the flag is disabled; otherwise it should just build the TargetAllocator CR, which happens in the normal path.

jaronoff97: yep, understood. this is more for future committers incase they don't see that.

swiatekm: I expect to get rid of this gate pretty quickly. It's really only here to make it easier for review and merge these changes without enabling anything by default for users.

PR #3403, opened 2024-10-28
Create service for extensions

Ankit152:

The Otel Operator doesn't creates a service for extensions. This will help the operator to create Service for extensions which will be consumed by an Ingress so that users can interact with it directly. This is related to deployment of Jaeger V2 in k8s.

- https://github.com/jaegertracing/jaeger-operator/issues/2717
- https://github.com/open-telemetry/opentelemetry-operator/issues/3460

### DIFF
--- a/.chloggen/service-extension.yaml
+++ b/.chloggen/service-extension.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: collector
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: support for creating a service for extensions when ports are specified.
+
+# One or more tracking issues related to the change
+issues: [3460]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/apis/v1beta1/config.go
+++ b/apis/v1beta1/config.go
@@ -207,7 +207,11 @@ func (c *Config) getPortsForComponentKinds(logger logr.Logger, componentKinds ..
 			continue
 		case KindExtension:
 			retriever = extensions.ParserFor
-			cfg = *c.Extensions
+			if c.Extensions == nil {
+				cfg = AnyConfig{}
+			} else {
+				cfg = *c.Extensions
+			}
 		}
 		for componentName := range enabledComponents[componentKind] {
 			// TODO: Clean up the naming here and make it simpler to use a retriever.
--- a/apis/v1beta1/config.go
+++ b/apis/v1beta1/config.go
@@ -206,7 +206,8 @@ func (c *Config) getPortsForComponentKinds(logger logr.Logger, componentKinds ..
 		case KindProcessor:
 			continue
 		case KindExtension:
-			continue
+			retriever = extensions.ParserFor
+			cfg = *c.Extensions
 		}
 		for componentName := range enabledComponents[componentKind] {
 			// TODO: Clean up the naming here and make it simpler to use a retriever.
@@ -316,6 +317,10 @@ func (c *Config) GetReceiverPorts(logger logr.Logger) ([]corev1.ServicePort, err
 
 func (c *Config) GetExporterPorts(logger logr.Logger) ([]corev1.ServicePort, error) {
 	return c.getPortsForComponentKinds(logger, KindExporter)
+}
+
+func (c *Config) GetExtensionPorts(logger logr.Logger) ([]corev1.ServicePort, error) {
+	return c.getPortsForComponentKinds(logger, KindExtension)
 }
 
 func (c *Config) GetAllPorts(logger logr.Logger) ([]corev1.ServicePort, error) {
--- a/apis/v1beta1/config.go
+++ b/apis/v1beta1/config.go
@@ -327,8 +327,12 @@ func (c *Config) GetExtensionPorts(logger logr.Logger) ([]corev1.ServicePort, er
 	return c.getPortsForComponentKinds(logger, KindExtension)
 }
 
-func (c *Config) GetAllPorts(logger logr.Logger) ([]corev1.ServicePort, error) {
+func (c *Config) GetReceiverAndExporterPorts(logger logr.Logger) ([]corev1.ServicePort, error) {
 	return c.getPortsForComponentKinds(logger, KindReceiver, KindExporter)
+}
+
+func (c *Config) GetAllPorts(logger logr.Logger) ([]corev1.ServicePort, error) {
+	return c.getPortsForComponentKinds(logger, KindReceiver, KindExporter, KindExtension)
 }
 
 func (c *Config) GetEnvironmentVariables(logger logr.Logger) ([]corev1.EnvVar, error) {
--- a/internal/components/extensions/helpers.go
+++ b/internal/components/extensions/helpers.go
@@ -55,6 +55,9 @@ var (
 				return components.ParseSingleEndpointSilent(logger, name, defaultPort, &config.SingleEndpointConfig)
 			}).
 			MustBuild(),
+		components.NewSinglePortParserBuilder("jaeger_query", 16686).
+			WithTargetPort(16686).
+			MustBuild(),
 	}
 )
 
--- a/internal/manifests/collector/service.go
+++ b/internal/manifests/collector/service.go
@@ -108,6 +108,34 @@ func MonitoringService(params manifests.Params) (*corev1.Service, error) {
 	}, nil
 }
 
+func ExtensionService(params manifests.Params) (*corev1.Service, error) {
+	name := naming.Service(params.OtelCol.Name)
+	labels := manifestutils.Labels(params.OtelCol.ObjectMeta, name, params.OtelCol.Spec.Image, ComponentOpenTelemetryCollector, []string{})
+
+	annotations, err := manifestutils.Annotations(params.OtelCol, params.Config.AnnotationsFilter())
+	if err != nil {
+		return nil, err
+	}
+
+	ports, err := params.OtelCol.Spec.Config.GetExtensionPorts(params.Log)
+	if err != nil {
+		return nil, err
+	}
+
+	return &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        name,
+			Namespace:   params.OtelCol.Namespace,
+			Labels:      labels,
+			Annotations: annotations,
+		},
+		Spec: corev1.ServiceSpec{
+			Ports:    ports,
+			Selector: manifestutils.SelectorLabels(params.OtelCol.ObjectMeta, ComponentOpenTelemetryCollector),
+		},
+	}, nil
+}
+
 func Service(params manifests.Params) (*corev1.Service, error) {
 	name := naming.Service(params.OtelCol.Name)
 	labels := manifestutils.Labels(params.OtelCol.ObjectMeta, name, params.OtelCol.Spec.Image, ComponentOpenTelemetryCollector, []string{})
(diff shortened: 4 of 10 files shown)


apis/v1beta1/config.go
```diff
@@ -282,6 +283,10 @@ func (c *Config) GetExporterPorts(logger logr.Logger) ([]corev1.ServicePort, err
 	return c.getPortsForComponentKinds(logger, KindExporter)
 }
 
+func (c *Config) GetExtensionPorts(logger logr.Logger) ([]corev1.ServicePort, error) {
+	return c.getPortsForComponentKinds(logger, KindExtension)
+}
```
yurishkuro, line 288: note: not all extensions create extra servers, an extensions could be just a library for use elsewhere, e.g. encoding extensions, or even jaeger storage.

iblancasa: It's fine. There is a mechanism to register those components that provide a port or something special like RBAC.

internal/manifests/collector/service.go
```diff
@@ -108,6 +108,34 @@ func MonitoringService(params manifests.Params) (*corev1.Service, error) {
...
+func ExtensionService(params manifests.Params) (*corev1.Service, error) {
+	name := naming.Service(params.OtelCol.Name)
+	labels := manifestutils.Labels(params.OtelCol.ObjectMeta, name, params.OtelCol.Spec.Image, ComponentOpenTelemetryCollector, []string{})
+
+	annotations, err := manifestutils.Annotations(params.OtelCol, params.Config.AnnotationsFilter())
+	if err != nil {
+		return nil, err
+	}
+
+	ports, err := params.OtelCol.Spec.Config.GetExtensionPorts(params.Log)
```
iblancasa, line 120: If there are no ports... maybe we can just not create the service

internal/components/extensions/helpers.go
```diff
@@ -55,6 +55,9 @@ var (
 				return components.ParseSingleEndpointSilent(logger, name, defaultPort, &config.SingleEndpointConfig)
 			}).
 			MustBuild(),
+		components.NewSinglePortParserBuilder("jaeger_query", 16686).
```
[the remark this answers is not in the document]
Ankit152: Do you want me to add a file say `jaegerquery.go` under `internal/components/extensions`?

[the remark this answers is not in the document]
Ankit152: The source code for the extension is present [here](https://github.com/jaegertracing/jaeger/blob/main/cmd/jaeger/internal/extension/jaegerquery/config.go).

internal/manifests/collector/service.go
pavolloffay: +1

pavolloffay (commented):
great progress, 

Please add unit tests and e2e test

internal/manifests/collector/service.go
Ankit152: Do you mean something similar to this?
```
if len(ports) == 0 {
    return nil, nil
}
```

iblancasa: Yes. That's what I meant.

I'm not sure about the HA. Do we need GRPC for this?

Ankit152: > Yes. That's what I meant.

I have already implemented that and written proper unit tests for the same. 
Although, I have a suggestion/doubt. Won't it be a good idea to add labels in the extension service similar to [these](https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/manifests/collector/service.go#L32-L36)?

Ankit152: > Do we need GRPC for this?

@yurishkuro can you answer this?

internal/manifests/collector/service_test.go
```diff
@@ -321,6 +321,161 @@ func TestMonitoringService(t *testing.T) {
...
+				Spec: v1beta1.OpenTelemetryCollectorSpec{
+					Config: v1beta1.Config{
+						Service: v1beta1.Service{
+							Extensions: []string{"jaeger_query"},
+						},
+						Extensions: &v1beta1.AnyConfig{
+							Object: map[string]interface{}{
+								"jaeger_query": map[string]interface{}{
+									"grpc": map[string]interface{}{
+										"endpoint": "0.0.0.0:16686",
```
yurishkuro, line 374: why this port? The default ports are
```
Query server started	{"kind": "extension", "name": "jaeger_query", "http_addr": "[::]:16686", "grpc_addr": "[::]:16685"}
```

Ankit152: It's just used to test whether the function is working correctly or not.

apis/v1beta1/config.go
```diff
@@ -318,6 +323,10 @@ func (c *Config) GetExporterPorts(logger logr.Logger) ([]corev1.ServicePort, err
...
+
 func (c *Config) GetAllPorts(logger logr.Logger) ([]corev1.ServicePort, error) {
 	return c.getPortsForComponentKinds(logger, KindReceiver, KindExporter)
```
swiatekm, line 331: You should add extensions here as well. Then you'll automatically get the ports added to the container spec.

tests/e2e/extension/00-install.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: opentelemetry.io/v1beta1
+kind: OpenTelemetryCollector
+metadata:
+  name: jaeger-inmemory
+spec:
+  image: jaegertracing/jaeger-snapshot:latest
+  ports:
+  - name: jaeger
+    port: 16686
```
swiatekm, line 9: Ideally, this should happen automatically.

tests/e2e/extension/00-assert.yaml
```diff
@@ -0,0 +1,272 @@
+apiVersion: v1
```
swiatekm, line 1: This assert is too verbose. Feel free to delete any attributes you aren't actively checking in this test.

swiatekm (commented):
This looks good to me overall, left a few comments and questions.

tests/e2e/extension/00-install.yaml
Ankit152: Removed it.

tests/e2e/extension/00-assert.yaml
Ankit152: Alright!

apis/v1beta1/config.go
Ankit152: The idea is good but I am already exposing `extension` ports in the new service. Just out of curiosity, will this have any advantage? 🤔

internal/manifests/collector/service.go
```diff
@@ -108,6 +110,40 @@ func MonitoringService(params manifests.Params) (*corev1.Service, error) {
 	}, nil
 }
 
+func ExtensionService(params manifests.Params) (*corev1.Service, error) {
+	name := naming.Service(params.OtelCol.Name)
+	labels := manifestutils.Labels(params.OtelCol.ObjectMeta, name, params.OtelCol.Spec.Image, ComponentOpenTelemetryCollector, []string{})
+	labels[extensionService] = valueExists
```
iblancasa, line 116: I don't think you actually need this. We are using this label to create `ServiceMonitors`.

.chloggen/service-extension.yaml
```diff
@@ -0,0 +1,16 @@
...
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: collector
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: support for creating a service for extensions when ports are specified.
+
+# One or more tracking issues related to the change
+issues: []
```
iblancasa, line 11: Missing issue id

Ankit152: Created an issue and attached it in the chloggen.

apis/v1beta1/config.go
Ankit152: Added.

tests/e2e/extension/00-assert.yaml
```diff
@@ -0,0 +1,264 @@
...
+            name: otc-internal
+        dnsConfig: {}
+        dnsPolicy: ClusterFirst
+        restartPolicy: Always
+        schedulerName: default-scheduler
+        securityContext: {}
+        serviceAccount: jaeger-inmemory-collector
+        serviceAccountName: jaeger-inmemory-collector
+        shareProcessNamespace: false
+        terminationGracePeriodSeconds: 30
```
swiatekm, line 87:
```suggestion
```

tests/e2e/extension/00-assert.yaml
swiatekm: I think I wasn't clear enough about what I wanted, apologies for that. I've included some suggested changes in my new review to get my point across better.

Ankit152: No worries. I have added all the suggested changes. However, for services I have intentionally kept them verbose to test their correctness. 🙂 

Please review the PR. 🙇🏼

iblancasa (approved):

swiatekm (approved):

internal/manifests/collector/service_test.go
```diff
@@ -321,6 +321,161 @@ func TestMonitoringService(t *testing.T) {
 	})
 }
 
+func TestExtensionService(t *testing.T) {
```
jaronoff97, line 324: nit... can you change this to be a list test instead here and also assert on the service created to be sure it has the right ports set?

Ankit152: Okay, I understood about asserting the ports. But can you elaborate more on `list test instead`? Do you mean something like this:

```
func TestExtensionService(t *testing.T) {
  tests := []struct{
      name.            string
      params          manifests.params
      ports              []corev1.ServicePort
      expectedErr  error
}{
      {
          ....
          ....
      },
      {
          ....
          ....
      },
}

for _,tc := range tests {
    t.Run(tc.name, func (t *testing.T){
        // assertion
    })
}
```

jaronoff97: yep! you can see an example [here](https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/manifests/collector/service_test.go#L56-L62) for these.

Ankit152: I have added the changes. Kindly review them!

Ankit152: @jaronoff97 @iblancasa @swiatekm if the changes looks good, then can we merge this one please?

swiatekm (approved):

internal/manifests/collector/service_test.go
```diff
@@ -321,6 +322,206 @@ func TestMonitoringService(t *testing.T) {
...
+								Extensions: []string{"jaeger_query"},
+							},
+							Extensions: &v1beta1.AnyConfig{
+								Object: map[string]interface{}{
+									"jaeger_query": map[string]interface{}{
+										"http": map[string]interface{}{
+											"endpoint": "0.0.0.0:16686",
+										},
+										"grpc": map[string]interface{}{
+											"endpoint": "0.0.0.0:16686",
```
jaronoff97, line 426: wouldnt we expect grpc and http to not share an endpoint?

jaronoff97 (approved):
one minor comment, otherwise this looks good!

internal/manifests/collector/service_test.go
Ankit152: Actually, `jaeger-query` won't be using GRPC endpoint. I have intentionally kept that to see if it is throwing an error or not.

jaronoff97: got it, @yurishkuro would you mind reviewing this and then i can merge it in?

yurishkuro: @jaronoff97 you should not block on my review - I know nothing about how operators are implemented internally, and this change is about that, not about any Jaeger knowledge, so my review would be completely superficial.

jaronoff97: i mostly wanted the jaeger knowledge and the superficial ✔️ just to close out the comments you left from before. but no problem, thanks for letting me know :)

Ankit152:
Thank you very much everyone for the kind feedback and reviews! 🙇🏼

PR #3331, opened 2024-10-07
Added support for PHP auto-instrumentation

SergeyKleyman:
**Description:** Adding support for auto-instrumentation of PHP application in the same way as it's already implemented for other runtimes such as Java, .NET, etc.

- Resolves: https://github.com/open-telemetry/opentelemetry-php/issues/932

**Testing:** <Describe what testing was performed and which tests were added.>

**Documentation:**
- Updated ./README.md

PR #3409, opened 2024-10-30
[chore] Build docker image for PHP auto-intrumentation

SergeyKleyman:
**Description:** Part 1 of #3331 that is adding support for auto-instrumentation of PHP application in the same way as it's already implemented for other runtimes such as Java, .NET, etc.
[As requested](https://github.com/open-telemetry/opentelemetry-operator/pull/3331#issuecomment-2396969274) this PR handles the part which is publishing the image.

- Work towards resolving: https://github.com/open-telemetry/opentelemetry-php/issues/932


.github/workflows/publish-autoinstrumentation-php.yaml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Login to GitHub Package Registry
+        uses: docker/login-action@v3
+        if: ${{ github.event_name == 'push' }}
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Prepare files for docker image
+        run: ./autoinstrumentation/php/prepare_files_for_docker_image.sh --ext-ver ${{ env.VERSION }} --dest-dir ${PWD}/autoinstrumentation/php/files_for_docker_image
```
iblancasa, line 67: How about running this inside the Dockerfile? As part of a multi-stage build

SergeyKleyman: Thank you - good idea. I'll try - I wonder if it will work considering that the script runs other docker containers...

frzifus: if it runs other containers, I think it doesnt make much sense to mount the docker sock. It simply makes it harder to execute it on machines with only e.g. podman available.

SergeyKleyman:
I marked autoinstrumentation/php/prepare_files_for_docker_image.sh as executable and added "[chore]" to the PR title to skip changelog check. Please take a look.

.github/workflows/publish-autoinstrumentation-php.yaml
swiatekm: I'd like to second the request to run all of this as a multi-stage docker build instead. That will make it much easier to maintain. I see that the script requires running docker images - in my view, you should instead rework it so the Dockerfile itself accepts the PHP version and libc flavor as arguments. If you can't fit everything in a single Dockerfile, multiple different ones are also fine.

SergeyKleyman: The problem with hardcoding all the possible combinations of parameters such PHP-version x C-lib-variant is that it will grow exponentially if/when we would want to add additional dimension such as for example CPU architecture (for example [Java also supports `linux/arm64,linux/s390x,linux/ppc64le`](https://github.com/open-telemetry/opentelemetry-operator/blob/main/.github/workflows/publish-autoinstrumentation-java.yaml#L74) some/all of those might be relevant to PHP. It will be very hard to maintain all these combinations.

SergeyKleyman: A possible alternative is to generate Dockerfile with all the combinations but I am not sure if this complication is worth it.

swiatekm: CPU architecture is covered by docker itself - as you can see for our Java image, we only have one Dockerfile, and build a multi-platform image for all those architectures transparently.

In all honesty, if this is a concern then I'm worried about the user experience for these images. It's already unfortunate that for some languages, the user needs to set an annotation if their application is based on musl, but at least those instrumentation artifacts still come from the same docker image. Is it not possible for us to do that here? Build a single image containing all the versions, and then pick between them at runtime?

SergeyKleyman: > Is it not possible for us to do that here? Build a single image containing all the versions, and then pick between them at runtime?

Maybe I misunderstood what you mean but the result of this PR will be one docker image that will contain all the required files for all the combinations of PHP-version x C-lib-variant. The relevant combination will be selected by in the separate upcoming PR in [the way similar to how it's currently implemented for .NET](https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/instrumentation/dotnet.go#L79).

Regarding CPU architecture I am still not sure how multi-platform image will work in this context. I used .NET as the closest example to PHP (.NET also has two parts: native and pure .NET) and I see that [.NET has x64 hardcoded in Dockerfle](https://github.com/open-telemetry/opentelemetry-operator/blob/main/autoinstrumentation/dotnet/Dockerfile#L23) even though [ARM64 in mentioned in the workflow](https://github.com/open-telemetry/opentelemetry-operator/blob/main/.github/workflows/publish-autoinstrumentation-dotnet.yaml#L73). Future `pkg/instrumentation/php.go` will need to copy the corresponding PHP-version x C-lib-variant x CPU-architecture combination and I am not sure from where CPU-architecture parameter will be coming. Could you please clarify that?

I can change this PR to generate multi-stage `Dockerfile` with stages for all the combinations - will that approach be preferable? If so should `Dockerfile` be committed or generated during the workflow?

swiatekm: Docker natively supports multi-architecture OCI manifests, and with buildx it can cross-build images for different architectures using QEMU. The .NET image you looked at is only available for x64 - it's our CI build process for it that's wrong. But if you look at the [Python Dockerfile](https://github.com/open-telemetry/opentelemetry-operator/blob/main/autoinstrumentation/python/Dockerfile), you won't see any reference to the CPU architecture. However, an [actual build](https://github.com/open-telemetry/opentelemetry-operator/actions/runs/11556099952/job/32163006131) shows images for all the listed architectures being produced. The reason this works is that all the base images used in the Python Dockerfile have versions for each architecture, and when we tell Docker to build for arm64, it will fetch the arm64 versions of each of those images and run the build in a ARM64 QEMU VM. Does that make sense? This is one of the reason we're so insistent on doing everything in a Dockerfile, it makes supporting multiple architectures much easier.

That said, looking at your build script more, I can see why it wouldn't be that easy to switch to that method. I'll experiment with this process a bit and see if I can make it more streamlined.

swiatekm: And on a separate note, could you cleanly rebase your changes on `main`? Right now it looks like you have a very messy merge in there.

SergeyKleyman: Thank you for clarification. I understand that multiple images built by the workflow using QEMU each image for the corresponding CPU architecture, the part about which I am not clear is how the corresponding image is selected at the runtime? Namely how [`pkg/instrumentation/python.go`](https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/instrumentation/python.go) knows which image to copy files from? Will the correct image with auto-instrumentation files be selected based on CPU architecture used by docker image with the instrumented application? Do I understand correctly that that determination will occur after `pkg/instrumentation/python.go` execution?

swiatekm: > Thank you for clarification. I understand that multiple images built by the workflow using QEMU each image for the corresponding CPU architecture, the part about which I am not clear is how the corresponding image is selected at the runtime? Namely how [pkg/instrumentation/python.go](https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/instrumentation/python.go) knows which image to copy files from? Will the correct image with auto-instrumentation files be selected based on CPU architecture used by docker image with the instrumented application? Do I understand correctly that that determination will occur after pkg/instrumentation/python.go execution?

The operator doesn't know the CPU architecture of the image. It could find out, but it doesn't care what it is. The container runtime on the K8s Node (containerd in most cases) will simply download the right image for the Node's CPU architecture. 

> Won't switching to multi-stage image approach (either by generating Dockerfile or writing it manually) handle CPU architecture automatically?

It will, but a Dockerfile with 6+ build stages, one for each combination of libc+php, is going to be messy and repetitive. I'm wondering if there's an elegant way of handing this.

SergeyKleyman: I agree that neither of the discussed proposals is perfect but this docker image is just a bag of files and it's not used directly by the end user.
Maybe for now we can implement one of the proposals that is good enough, we will get the feature out, receive feedback and then if/when necessary we can improve upon it?

swiatekm:
@SergeyKleyman can you clean this branch up so Github doesn't show all the changes from main?

SergeyKleyman:
> @SergeyKleyman can you clean this branch up so Github doesn't show all the changes from main?

Sure thing - I am not an Git expert though so maybe you can help me out with exact git commands to achieve that? I think I caused the current situation because I rebased my branch on main and then I did pull-then-push while maybe the right way would have been to force-push without pull ? I just thought that force-pushing to published branch is kind of "bad" thing. So maybe you can help me with the following two questions:

1. Do I understand correctly that the right way to rebase on the current main would have been to rebase local clone and then force-push?
2. How can I fix the current state so that this PR shows only my changes on top of main?

Thank you in advance.

SergeyKleyman:
> 1. Do I understand correctly that the right way to rebase on the current main would have been to rebase local clone and then force-push?
> 2. How can I fix the current state so that this PR shows only my changes on top of main?

It seems merging main to my branch fixed the issue so I assume I should have just merged main on the first attempt instead of rebasing on main.

SergeyKleyman:
@swiatekm Could you please take a look if there is anything else still needs to be done?

jaronoff97 (approved):
I think this LGTM. the bash code was actually incredibly readable so good on you :D 

One minor thing, let's include php in the [versions.txt](https://github.com/open-telemetry/opentelemetry-operator/blob/main/versions.txt#L39-L41) file which rn just won't be read by anything.

SergeyKleyman:
Could you please let me know what is the next step to get this PR merged?

swiatekm:
@pavolloffay could you also have a look?

@SergeyKleyman no need to do anything else on your end. Supporting autoinstrumentation in a new language is simply a big decision, and we want to have wide consensus before merging it.

autoinstrumentation/php/prepare_files_for_docker_image.sh
```diff
@@ -0,0 +1,187 @@
...
+}
+
+select_composer_json_for_PHP_version() {
+    local PHP_version="${1:?}"
+    case "${PHP_version}" in
+        8.0|8.1|8.2)
+            echo "composer_PHP_${PHP_version}.json"
+            ;;
+        8.3)
+            echo "composer_PHP_8.2.json"
```
frzifus, line 129: Just to clarify, its correct we want the `8.2` filein case of `8.3`?

frzifus (approved):
LGTM

autoinstrumentation/php/prepare_files_for_docker_image.sh
SergeyKleyman: Yes, that is on purpose. Should I add a comment to that code line clarifying that?

The reason for having separate `composer_PHP_8.1.json` and `composer_PHP_8.2.json` is because [`open-telemetry/opentelemetry-auto-pdo` auto-instrumentation is supported only for PHP 8.2 and later](https://packagist.org/packages/open-telemetry/opentelemetry-auto-pdo).

brettmc: This relates to which PHP versions can/cannot observe internal functions, which arrived in 8.2, so it's really `<8.2` and `>=8.2`
I think that needs a comment or improved file names.

swiatekm: Please add the comment, and then we should be good to merge.

brettmc (approved):

autoinstrumentation/php/prepare_files_for_docker_image.sh
```diff
@@ -0,0 +1,217 @@
...
+    current_user_id="$(id -u)"
+    local current_user_group_id
+    current_user_group_id="$(id -g)"
+    docker run --rm \
+        -v "${dest_dir_for_current_args}:/dest_dir" \
+        "${PHP_docker_image}" sh -c "\
+        mkdir -p /app && cd /app \
+        ${install_compiler_command} \
+        && pecl install opentelemetry-${opentelemetry_extension_version} \
+        && cp /usr/local/lib/php/extensions/no-debug-non-zts-*/opentelemetry.so /dest_dir/ \
```
BenElferink, line 104: Better to extract the dir dynamically:
```
cp "$(php-config --extension-dir)/opentelemetry.so" /dest_dir/opentelemetry.so
```

PR #3411, opened 2024-10-30
Allow setting target allocator via label

swiatekm:
**Description:**
Allow connecting the OpenTelemetryCollector and TargetAllocator CRs by setting a label on the Collector CR.

- Related to: #2422 

**Testing:**
Added some basic unit tests and an e2e test showing how label changes affect the Collector manifests.

### DIFF
--- a/controllers/opentelemetrycollector_controller.go
+++ b/controllers/opentelemetrycollector_controller.go
@@ -38,6 +38,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
 	"github.com/open-telemetry/opentelemetry-operator/apis/v1beta1"
 	"github.com/open-telemetry/opentelemetry-operator/internal/autodetect/openshift"
 	"github.com/open-telemetry/opentelemetry-operator/internal/autodetect/prometheus"
@@ -49,6 +50,8 @@ import (
 	collectorStatus "github.com/open-telemetry/opentelemetry-operator/internal/status/collector"
 	"github.com/open-telemetry/opentelemetry-operator/pkg/featuregate"
 )
+
+const collectorTargetAllocatorLabelName = "opentelemetry.io/target-allocator"
 
 var (
 	ownedClusterObjectTypes = []client.Object{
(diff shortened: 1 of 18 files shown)


controllers/opentelemetrycollector_controller.go
```diff
@@ -50,6 +51,8 @@ import (
 	"github.com/open-telemetry/opentelemetry-operator/pkg/featuregate"
 )
 
+const collectorTargetAllocatorLabelName = "opentelemetry.io/target-allocator"
```
jaronoff97, line 54: should we put this [here](https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/constants/env.go)?

controllers/opentelemetrycollector_controller.go
```diff
@@ -179,14 +182,28 @@ func (r *OpenTelemetryCollectorReconciler) GetParams(instance v1beta1.OpenTeleme
...
+	targetAllocator, err := r.getTargetAllocator(ctx, p)
 	if err != nil {
 		return p, err
 	}
 	p.TargetAllocator = targetAllocator
 	return p, nil
 }
 
+func (r *OpenTelemetryCollectorReconciler) getTargetAllocator(ctx context.Context, params manifests.Params) (*v1alpha1.TargetAllocator, error) {
+	otelcol := params.OtelCol
```
jaronoff97, line 194: why do we need the shallow copy here?

swiatekm: This was just for readability I think, but it doesn't look necessary anymore. It's a pointer anyway, so no copy.

main.go
```diff
@@ -440,7 +440,7 @@ func main() {
 
 		bv := func(collector otelv1beta1.OpenTelemetryCollector) admission.Warnings {
 			var warnings admission.Warnings
-			params, newErr := collectorReconciler.GetParams(collector)
+			params, newErr := collectorReconciler.GetParams(context.Background(), collector)
```
frzifus, line 443: Does the underlying implementation handle timeouts and all that stuff?

frzifus (approved):
:partying_face: :partying_face:

main.go
swiatekm: It does, but now that you mention it, this should really be the context in which the caller runs. I'll modify the build validation function itself to take a context argument instead.

jaronoff97 (approved):

main.go
swiatekm: Fixed.

PR #3416, opened 2024-11-01
Add Node.js support for `--import` flag

### LINKED ISSUE #3414: Support `--import` flag for Node.js instrumentation
### Component(s)

auto-instrumentation

Currently it's only possible to instrument a Node.js application that uses ESM, or to create a custom auto-instrumentation image with asynchronous initialisation logic, because the `--require` flag injected in the `NODE_OPTIONS` environment variable doesn't support these use cases.

Adding a new `useImport` or similar flag to the Node.js auto-instrumentation that injects `--import /otel-auto-instrumentation-nodejs/autoinstrumentation.mjs` instead of `--require /otel-auto-instrumentation-nodejs/autoinstrumentation.js`, and adding the new file to the provided image that delegates to `autoinstrumentation.js` but also registers hooks for ESM instrumentation.

_No response_

### Additional context

I'd be happy to work on this.

raphael-theriault-swi:
**Description:**

This adds a new `useImport` option for Node.js autoinstrumentation that injects and `--import` flag to `NODE_OPTIONS` instead of the default `--require` flag. The behaviour with the provided image is the same as what `--require` does with the added feature of also registering hooks to instrument ESM code.

This also opens the door for custom images to run asynchronous code in their instrumentation code, which isn't possible with `--require`.

- Resolves: #3414

**Testing:**

Added both unit tests and a new e2e test that exercise the feature.

**Documentation:**

The new option is documented and was added to the generated docs.


.chloggen/node-import.yaml
```diff
@@ -0,0 +1,16 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add Node.js support for `--import` flag
+
+# One or more tracking issues related to the change
+issues: [3414]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
iblancasa, line 16: Can you add here what that flag does?

raphael-theriault-swi: Done !

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,14 @@
...
+images:
+- name: controller
+  newName: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
+patches:
+- patch: '[{"op":"add","path":"/spec/template/spec/containers/0/args/-","value":"--target-allocator-image=ghcr.io/open-telemetry/opentelemetry-operator/target-allocator:"}]'
+  target:
+    kind: Deployment
+- patch: '[{"op":"add","path":"/spec/template/spec/containers/0/args/-","value":"--operator-opamp-bridge-image=ghcr.io/open-telemetry/opentelemetry-operator/operator-opamp-bridge:"}]'
+  target:
+    kind: Deployment
```
swiatekm, line 14: Can you run `make reset` to get rid of this?

raphael-theriault-swi: Doesn't seem to clear this

swiatekm: Looks like you added these values without an image version somehow. Could you try resetting this file to `main`, staging the change, and then running `make generate manifests bundle api-docs reset` ?

swiatekm (commented):
This looks good to me overall, but I'd feel more confident if one of @open-telemetry/javascript-maintainers could have a look at the JS specific parts.

bundle/community/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -478,6 +478,8 @@ spec:
                 - --zap-log-level=info
                 - --zap-time-encoding=rfc3339nano
                 - --enable-nginx-instrumentation=true
+                - '--target-allocator-image=ghcr.io/open-telemetry/opentelemetry-operator/target-allocator:'
+                - '--operator-opamp-bridge-image=ghcr.io/open-telemetry/opentelemetry-operator/operator-opamp-bridge:'
```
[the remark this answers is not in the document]
raphael-theriault-swi: Managed to get it to output what I assume is correct

apis/v1alpha1/instrumentation_types.go
```diff
@@ -217,6 +217,11 @@ type NodeJS struct {
 	// Resources describes the compute resource requirements.
 	// +optional
 	Resources corev1.ResourceRequirements `json:"resourceRequirements,omitempty"`
+
+	// UseImport overrides the default injected --require flag with an --import flag that supports ESM.
+	// Requires Node.js 18 or later.
+	// +optional
+	UseImport bool `json:"useImport,omitempty"`
```
trentm, line 224: One subtlety that might impact the choice of the config var name.
The feature for the user is whether a module import hook is added (via `module.register()`) to **support instrumentating ESM code**. 

That is actually independent of whether `--import something.mjs` or `--require something.js` is used to run the OTel setup when Node.js is starting up. It is definitely possible to call `module.register("@opentelemetry/instrumentation/hooks.mjs");` from the CommonJS `autoinstrumentation.js` (loaded via `--require`).

So I wonder if a config name something like `instrumentEsm` or something like that would be clearer.

---

There are a few subtleties discussed at https://github.com/open-telemetry/opentelemetry-js/issues/4933, but I don't think you need to read and grok all that.  **Eventually** we'd like it so that [there is just one the one obvious way](https://github.com/open-telemetry/opentelemetry-js/issues/4933#issuecomment-2299946388) to setup OTel -- and that would be via `--import some-esm-file-that-calls-module.register.mjs`. However, for now we need to make the ESM support opt-in because:

- it requires newer versions of node
- it is possible the user explicitly does *not* want ESM instrumentation enabled because it is still new and can have overhead and bugs

.chloggen/node-import.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+# One or more tracking issues related to the change
+issues: [3414]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The new UseImport configuration overrides the default injected --require flag with an --import flag that supports ESM.
+  Requires Node.js 18 or later.
```
trentm, line 18: The important Node.js feature here is whether `module.register()` is supported and that version range is `^18.19.0 || ^20.6.0 || >=22`. It would be good to be specific here, so some user of an earlier minor version of v18 or v20 doesn't expect this to work.

autoinstrumentation/nodejs/src/autoinstrumentation.mts
```diff
@@ -0,0 +1,4 @@
+import "./autoinstrumentation.js";
+
+import { register } from "node:module";
+register("@opentelemetry/instrumentation/hooks.mjs");
```
trentm, line 4: The current code will crash if used on a verison of node that doesn't yet have `module.register`:

```
% cat foo.mjs
import { register } from "node:module";
console.log('reg', register);

% /Users/trentm/.nvm/versions/node/v18.0.0/bin/node foo.mjs
file:///Users/trentm/.../foo.mjs:1
import { register } from "node:module";
         ^^^^^^^^
SyntaxError: The requested module 'node:module' does not provide an export named 'register'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:128:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:194:5)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:409:24)
    at async loadESM (node:internal/process/esm_loader:85:5)
    at async handleMainPromise (node:internal/modules/run_main:61:12)

Node.js v18.0.0
```

An option might be to gracefully warn (untested code):

```suggestion
import { diag } from '@opentelemetry/api';
import * as module from 'node:module';
if (typeof module.register === 'function') {
    module.register('@opentelemetry/instrumentation/hooks.mjs');
} else {
    diag.warn(`OpenTelemetry Operator auto-instrumentation could not instrument ESM code: Node.js ${process.version} does not support 'module.register()'`);
}
```

raphael-theriault-swi: Does the SDK setup a logger for the diag API by default ? If not I'm thinking this should probably be a `console.warn`

trentm: The NodeSDK sets a DiagConsoleLogger if the `OTEL_LOG_LEVEL` env var is set. (Per https://github.com/open-telemetry/opentelemetry-js/pull/3693 it does not set a diag logger at all if the env var isn't set at all. So by *default* logging is effectively off.) `node -r @opentelemetry/auto-instrumentations-node/register ...` *does* turn it on by default (https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/metapackages/auto-instrumentations-node/src/register.ts#L23-L26).

Using the diag logger is the intended mechanism for logging by the OTel JS packages, but that doesn't necessarily have to mean that the OTel Operator uses it. I don't have a strong opinion whether the diag logger or `console.*` is used here.

apis/v1alpha1/instrumentation_types.go
trentm: > a custom image could also require the use of an import flag for top level await support (which is actually our usecase for this addition)

Ah, I misunderstood the use case then. Using `import` in the var name makes sense then. The docs for this feature should mention the two things then: (1) `--import ...` is used to load "autoinstrumentation.mjs" (so custom versions can use ESM and top-level await), (2) the default "autoinstrumentation.mjs" differs from "autoinstrumentation.js" in that it registers the hook for ESM instrumentation.

raphael-theriault-swi:
@trentm Thoughts on the updated wording ?

trentm (approved):
I  think the JS parts of this look good now. (I haven't reviewed the test code at all.)

swiatekm:
@raphael-theriault-swi could you regenerate the manifests? You have some conflicts there.

swiatekm:
@raphael-theriault-swi looks like you have a slightly malformed changelog entry. Can you have a look? It's complaining about tabs vs spaces.

raphael-theriault-swi:
This should do it

swiatekm (approved):

swiatekm (commented):
@raphael-theriault-swi it looks like the new e2e test you added is consistently failing. Can you have a look?

My suspicion is that it's because we don't actually build the autoinstrumentation images for e2e tests, and your new test depends on the image having the mjs file.

autoinstrumentation/nodejs/src/autoinstrumentation.mts
```diff
@@ -0,0 +1,10 @@
+import "./autoinstrumentation.js";
```
pavolloffay, line 1: @raphael-theriault-swi how does the `--import` flag with with the upstream `opentelemetry/auto-instrumentations-node` ? We are trying to minimize the nodejs code and rely only on that package.


Related ticket: https://github.com/open-telemetry/opentelemetry-operator/issues/3465

pavolloffay (changes requested):

autoinstrumentation/nodejs/src/autoinstrumentation.mts
pavolloffay: @trentm could you please take a look here? cc) @JamieDanielson

trentm: @pavolloffay Are you asking if `node --import @opentelemetry/auto-instrumentations-node/register ...` works? If so, I think yes it should work fine because `node --import something ...` works with `something` being either ESM or CommonJS code. I haven't tried it, though.

However, whether to move the operator to using `@opentelemetry/auto-instrumentations-node` instead of the bootstrap code in "autoinstrumentation.js" doesn't need to be part of *this*  PR, right?

pavolloffay: > However, whether to move the operator to using @opentelemetry/auto-instrumentations-node instead of the bootstrap code in "autoinstrumentation.js" doesn't need to be part of this PR, right?

The issue is that, if we keep adding new features here it will be harder to migrate them to `auto-instrumentations-node` (or other node packages)

raphael-theriault-swi: Like I mentioned above switching to `auto-instrumentations-node` would be not require any change to the code added by this PR now or in the future.

raphael-theriault-swi:
@swiatekm @pavolloffay Would this be more likely to get merged if I also transition the rest of the image to using `auto-instrumentations-node` ?

pavolloffay (approved):

swiatekm:
> Hello, interested in seeing this merged! Is any help required to get it over the line?

I think the change has been approved conceptually, but in the meantime the nodejs instrumentation package has moved to using `auto-instrumentations-node`. So @raphael-theriault-swi would need to rebase this PR on main.

On a more positive note, we now run E2E tests on instrumentation images built from PRs, so if this breaks anything obviously, we'll know.

swiatekm (approved):

swiatekm:
@pavolloffay can you have a final look before we merge?

raphael-theriault-swi:
quick bump @pavolloffay

pavolloffay (approved):

Issue #3420, opened 2024-11-05
Automatic RBAC support for k8sevents receiver


iblancasa:
### Component(s)

collector

### Describe the issue you're reporting

Create automatically the RBAC rules when k8seventsreceiver is present in the configuration.

PR #3421, opened 2024-11-05
Add automatic RBAC creation for k8sevents receiver

iblancasa:
- Resolves: #3420

### DIFF
--- a/internal/components/receivers/k8sevents.go
+++ b/internal/components/receivers/k8sevents.go
@@ -1,0 +1,79 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package receivers
+
+import (
+	"github.com/go-logr/logr"
+	rbacv1 "k8s.io/api/rbac/v1"
+)
+
+type k8seventsConfig struct{}
+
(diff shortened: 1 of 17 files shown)


internal/components/receivers/k8sevents.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"github.com/go-logr/logr"
+	rbacv1 "k8s.io/api/rbac/v1"
+)
+
+type k8seventsConfig struct{}
+
+func generatek8seventsRbacRules(_ logr.Logger, _ k8seventsConfig) ([]rbacv1.PolicyRule, error) {
+	// The k8s Events Receiver needs get permissions on the following resources always.
+	return []rbacv1.PolicyRule{
```
swiatekm, line 26: Do we know why this component requires all of these permissions? It looks like it only sets a watch on Events themselves.

iblancasa: I just followed the component documentation: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/k8seventsreceiver/README.md#rbac

I agree they are a lot. Maybe I can work on the future in the contrib repo to see if we really need that, fix the documentation and fix the permissions where. WDYT?

swiatekm (approved):

iblancasa:
Can we merge this?

tests/e2e-automatic-rbac/receiver-k8sevents/01-install.yaml
```diff
@@ -0,0 +1,18 @@
...
+spec:
+  config: |
+    receivers:
+      k8s_events:
+    processors:
+    exporters:
+      debug:
+    service:
+      pipelines:
+        traces:
```
zarbis, line 15: This shouldn't work and should fail with `Error: failed to build pipelines: failed to create "k8s_events" receiver for data type "traces": telemetry type is not supported`.

I wonder if this is the reason why no CR/CRB is created if I deploy exactly this manifest, but with `logs` instead of `traces`?

zarbis: Nevermind, my operator SA was missing permissions... but this typo is legit.

swiatekm: Worth fixing, but for the purpose of this test it doesn't matter. We don't even care if the collector starts, just that the generated ClusterRole is correct.

Issue #3429, opened 2024-11-06
Add automatic RBAC creation for k8sobjects receiver


iblancasa:
### Component(s)

collector

### Describe the issue you're reporting

Create automatically the RBAC rules when k8sobjectsreceiver is present in the configuration.

PR #3430, opened 2024-11-06
Add automatic RBAC creation for k8sobjects receiver

iblancasa:
**Description:** 
- Resolves: #3429

### DIFF
--- a/internal/components/receivers/k8sobjects.go
+++ b/internal/components/receivers/k8sobjects.go
@@ -30,7 +30,7 @@ type k8sObject struct {
 }
 
 func generatek8sobjectsRbacRules(_ logr.Logger, config k8sobjectsConfig) ([]rbacv1.PolicyRule, error) {
-	// The k8s Objects Receiver needs get permissions on the following resources always.
+	// https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver#rbac
 	prs := []rbacv1.PolicyRule{}
 	for _, obj := range config.Objects {
 		permissions := []string{"list"}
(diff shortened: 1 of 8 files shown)


internal/components/receivers/k8sobjects.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+type k8sObject struct {
+	Name  string `yaml:"name"`
+	Mode  string `yaml:"mode"`
+	Group string `yaml:"group,omitempty"`
+}
+
+func generatek8sobjectsRbacRules(_ logr.Logger, config k8sobjectsConfig) ([]rbacv1.PolicyRule, error) {
+	// The k8s Objects Receiver needs get permissions on the following resources always.
```
swiatekm, line 33: Comment left over from k8sevents? We could replace it with a link to the documentation specifying why we require these specific RBAC rules.

swiatekm (approved):
LGTM, minor nitpick about a comment.

pavolloffay (approved):

tests/e2e-automatic-rbac/receiver-k8sobjects/01-assert.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
pavolloffay, line 2: This is fine now, but it would be better to do a real e2e tests and asserting the k8sobjects actually can collect data

tests/e2e-automatic-rbac/receiver-k8sobjects/01-install.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: opentelemetry.io/v1alpha1
```
pavolloffay, line 1: can we start using v1beta1 in tests?

PR #3441, opened 2024-11-10
Create Ingress for extensions

### LINKED ISSUE #3438: [Collector] Consume extension service with Ingress to give users a way to interact with them
### Component(s)

collector

OpenTelemetry operator accepts custom [configs](https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1beta1/opentelemetrycollector_types.go#L102) in the spec to create all the resources on its behalf. Jaeger V2 follows the setting up of parameters using config files rather than passing them through the CLI flags in the command. The OTEL operator created resources like pods, deployments, service accounts, and services when custom configs were passed to create an OTEL instance for Jaeger V2. However, to interact with the UI, a forwarding port was required which is not something we should follow in a production environment. 

To cater this, a new [method](https://github.com/open-telemetry/opentelemetry-operator/pull/3403) is being implemented that will create a new service explicitly for extensions. The newly created service will be consumed by the ingress to give users a way to directly interact with the ui (in Jaeger's case it's `jaeger_query`). 

Since, the Operator has a parameter [ingress](https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1beta1/opentelemetrycollector_types.go#L113) which creates an Ingress according the specification set by the users themselves. Is it a good a idea to use the existing functionality?

Other way of creating ingress could be similar to the way [extension service](https://github.com/open-telemetry/opentelemetry-operator/pull/3403) is being implemented. In this, we can create a method e.g., `ExtensionIngress` which will only be responsible for creating ingress for extensions. This will also leave less room for errors as it will automate the entire flow and users won't have to worry about them at all. 

### Additional context

Jaeger V2: https://github.com/jaegertracing/jaeger-operator/issues/2717
Extension Service: https://github.com/open-telemetry/opentelemetry-operator/pull/3403

Ankit152:

- Resolves: #3438 
- https://github.com/jaegertracing/jaeger-operator/issues/2717

### DIFF
--- a/internal/manifests/collector/ingress.go
+++ b/internal/manifests/collector/ingress.go
@@ -50,9 +50,9 @@ func Ingress(params manifests.Params) (*networkingv1.Ingress, error) {
 	var rules []networkingv1.IngressRule
 	switch params.OtelCol.Spec.Ingress.RuleType {
 	case v1beta1.IngressRuleTypePath, "":
-		rules = []networkingv1.IngressRule{createPathIngressRules(params.OtelCol.Name, params.OtelCol.Spec.Ingress.Hostname, ports)}
+		rules = []networkingv1.IngressRule{createPathIngressRules(params.OtelCol.Name, params.OtelCol.Spec.Ingress.Hostname, ports, "service")}
 	case v1beta1.IngressRuleTypeSubdomain:
-		rules = createSubdomainIngressRules(params.OtelCol.Name, params.OtelCol.Spec.Ingress.Hostname, ports)
+		rules = createSubdomainIngressRules(params.OtelCol.Name, params.OtelCol.Spec.Ingress.Hostname, ports, "service")
 	}
 
 	return &networkingv1.Ingress{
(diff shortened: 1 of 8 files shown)


Ankit152:
It's a POC and work is still in progress.
The discussion around the implementation is still on-going.

internal/manifests/collector/ingress.go
```diff
@@ -70,6 +70,38 @@ func Ingress(params manifests.Params) (*networkingv1.Ingress, error) {
...
+		return nil, nil
+	}
+
+	rules := createSubdomainIngressRules(name, "", ports)
+
+	return &networkingv1.Ingress{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        name,
+			Namespace:   params.OtelCol.Namespace,
+			Annotations: params.OtelCol.Spec.Ingress.Annotations, // can the spec annotations be used?
```
iblancasa, line 94: What do you mean here?

Ankit152: I was just wondering if we can use the `OtelCol.ObjectMeta.Annotations` instead of `Ingress.Annotations` as this function will be responsible for creating ingress for extension services.

Ankit152:
Folks since the service for extensions is merged, I would like to discuss about the implementation of Ingress. 

cc @pavolloffay @iblancasa @swiatekm @jaronoff97 @yurishkuro

Ankit152:
Sure @swiatekm 
It will be very great if we discuss the implementation on the issue itself.

frzifus (commented):
> Folks since the service for extensions is merged, I would like to discuss about the implementation of Ingress.

We can put it on the agenda for the SIG call if you like.

frzifus:
> The change looks fine to me. What I'm not convinced of is whether we need a whole separate configuration vs a boolean switch to include extension ports in the existing Ingress. I'd like to discuss that in the issue.

Can a boolean work when the extension may changes?

swiatekm:
> > The change looks fine to me. What I'm not convinced of is whether we need a whole separate configuration vs a boolean switch to include extension ports in the existing Ingress. I'd like to discuss that in the issue.
> 
> Can a boolean work when the extension may changes?

Not sure what you mean? A boolean switch would mean we'd just add the extension paths to the existing receiver Ingress.

frzifus:
hey @Ankit152, the contribution was discussed in the SIG call today.

There three things came up:
- Can this lead to unwanted extensions being exposed by mistake? (e.g. pprof and other things? Even if there is no port-parser implementation for something specific, it may gets added in the future and would be exposed automatically.)
- Do we need a 2nd ingress entry? Can we maybe reuse the exsting one with a different sub domain? (e.g. `customXYZ.extension.*`?)
- Would it be sufficent to just provide a `extensionWhitelist` that will be exposed? (OT: might be also a good idea to add a receiver whitelist then. But thats another topic to discuss)

Ankit152:
Hey @frzifus 
Sorry for delayed response as I was out for year end recharge. I don't mind adding any modifications to the existing PR as this was just raised to discuss the implementation. 

@pavolloffay can you please help addressing the questions?

chinaran:
any update?

Issue #3446, opened 2024-11-11
Operator fails when the pod is restarted and the Service Monitor for operator metrics was already created


iblancasa:
### Component(s)

collector

## Description
Reported by @IshwarKanse (great job!)

## Steps to Reproduce
- Deploy the operator
- Change something from the deployment. For instance, add an arg
- Wait until the new pod fails

## Actual Result
```sh
{"level":"INFO","timestamp":"2024-11-11T10:18:28.446310667Z","message":"All workers finished","controller":"opampbridge","controllerGroup":"opentelemetry.io","controllerKind":"OpAMPBridge"}
{"level":"INFO","timestamp":"2024-11-11T10:18:28.446315312Z","message":"Stopping and waiting for caches"}
W1111 10:18:28.446400       1 reflector.go:484] pkg/mod/k8s.io/client-go@v0.31.2/tools/cache/reflector.go:243: watch of *v1.ConfigMap ended with: an error on the server ("unable to decode an event from the watch stream: context canceled") has prevented the request from succeeding
{"level":"INFO","timestamp":"2024-11-11T10:18:28.446627238Z","message":"Stopping and waiting for webhooks"}
{"level":"INFO","timestamp":"2024-11-11T10:18:28.446652862Z","logger":"controller-runtime.webhook","message":"Shutting down webhook server with timeout of 1 minute"}
{"level":"INFO","timestamp":"2024-11-11T10:18:28.446740103Z","message":"Stopping and waiting for HTTP servers"}
{"level":"INFO","timestamp":"2024-11-11T10:18:28.446760527Z","logger":"controller-runtime.metrics","message":"Shutting down metrics server with timeout of 1 minute"}
{"level":"INFO","timestamp":"2024-11-11T10:18:28.44679548Z","message":"shutting down server","name":"health probe","addr":"[::]:8081"}
{"level":"INFO","timestamp":"2024-11-11T10:18:28.446854274Z","message":"Wait completed, proceeding to shutdown the manager"}
{"level":"ERROR","timestamp":"2024-11-11T10:18:28.454439947Z","message":"error received after stop sequence was engaged","error":"leader election lost","stacktrace":"sigs.k8s.io/controller-runtime/pkg/manager.(*controllerManager).engageStopProcedure.func1\n\t/Users/ikanse/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.1/pkg/manager/internal.go:512"}
[...]

PR #3447, opened 2024-11-11
Fix error when the operator metrics ServiceMonitor already exists

iblancasa:
- Resolves: #3446

**Testing:** @IshwarKanse is working on a test to ensure this is detected in early stages.

### DIFF
--- a/internal/operator-metrics/metrics.go
+++ b/internal/operator-metrics/metrics.go
@@ -19,8 +19,11 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/go-logr/logr"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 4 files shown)


internal/operator-metrics/metrics.go
```diff
@@ -51,26 +54,35 @@ var _ manager.Runnable = &OperatorMetrics{}
...
+		log:        log,
 	}, nil
 }
 
 func (om OperatorMetrics) Start(ctx context.Context) error {
 	rawNamespace, err := os.ReadFile(namespaceFile)
 	if err != nil {
-		return fmt.Errorf("error reading namespace file: %w", err)
+		om.log.Error(err, "error reading namespace file. No ServiceMonitor will be created")
+		return nil
```
swiatekm, line 76: Is the idea here that this is an optional dependency - that is, if it fails, the operator should continue to run? In that case, I think we should return all the errors normally, and have a wrapper which only logs them. Would be easier to manage that way imo.

iblancasa: Oh thanks for the suggestion. Added

swiatekm (approved):

pavolloffay (approved):

vladst3f:
Hi @pavolloffay, can we get this in a patch release sometime soon ? we're forced to use a main build now.

swiatekm:
@vladst3f we're going to be releasing v0.114.0 shortly. We'll also discuss whether this warrants releasing a patch for 0.113.0 in addition. Apologies for the hassle!

vladst3f:
cheers, most likely rolling forward will go down smoothly. the quick reaction is appreciated!

Starefossen:
This has been broken for two weeks now and should have been a patch to 0.113 once this pr was landed and not waiting the full length for a new release imho

PR #3448, opened 2024-11-11
feat(nodejs-autoinstrumentation): enable overriding default histogram buckets

### LINKED ISSUE #3436: [AutoInstrumentation] NodeJS histogram buckets should be configurable
### Component(s)

auto-instrumentation

## Description

When injecting autoinstrumentation on nodejs apps, we use the [default histogram buckets](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#explicit-bucket-histogram-aggregation): `[ 0, 5, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500, 5000, 7500, 10000 ]`

While it makes sense for values expressed in `ms`, it does not work with metrics such as `http_server_duration_bucket` expressed in `seconds`.

It seems other SDKs face(d) the same issue:
- [.NET](https://github.com/open-telemetry/opentelemetry-dotnet/issues/4797)
- [Go](https://github.com/open-telemetry/opentelemetry-go/issues/5821)

I would expect to be able to configure the bucket list for metrics to adjust to the currently used unit, ideally through the `Instrumentation` CRD. As I understand it, we could, for example, pass a [view](https://github.com/open-telemetry/opentelemetry-js/tree/72c9af91983e4b7aade98c901bd45c6cefee0da4/experimental/packages/opentelemetry-sdk-node#views) to [the created NodeSDK](https://github.com/open-telemetry/opentelemetry-operator/blob/main/autoinstrumentation/nodejs/src/autoinstrumentation.ts#L53)

While this is not an issue for span metrics or service-graph metrics (buckets can be configured downstream), it makes other metrics like `http_server_duration_bucket` unusable.

### Kubernetes Version

1.30

### Operator version

0.111.0

### Collector version

0.111.0

### Environment information

_No response_

### Log output

_No response_

### Additional context

_No response_

CCOLLOT:
### Description
Fixes https://github.com/open-telemetry/opentelemetry-operator/issues/3436 (default metrics are in seconds but default buckets are in ms)

I implemented this patch to specify a new list of buckets (seconds) for the nodejs autoinstrumentation NodeSDK:

- Enable overriding the default SDK buckets using the `OTEL_METRICS_EXPLICIT_BUCKET_HISTOGRAM` environment variable (comma-separated list)
- Provide sane default values for the generated metrics (all expressed in seconds AFAIK), using the same set of buckets as the [.NET library](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4820) (0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10)

**Testing:**: tested with opentelemetry operator 0.112.0 and got the expected bucket list:
image

**Documentation:** I will add the necessary documentation if the change is accepted


iblancasa (approved):

swiatekm:
@open-telemetry/javascript-maintainers could you have a look? I'm reluctant to introduce non-standard environment variables for instrumentations that only the operator supports.

CCOLLOT:
It would be nice to get an SDK maintainer's opinion. 👍

Fixing the unit mismatch is one thing, but we also need to allow specifying a custom list of buckets.

The nodejs SDK already supports providing a list of views.
Either it's the autoinstrumentation's concern to use the SDK's current way to customize the buckets (passing the view), or else the SDK needs to provide another way to customize the buckets that the autoinstrumentation init-container can use.

As you suggested in the issue, we could alternatively specify it as an attribute on the Instrumentation CR instead of an environment variable 👍

pavolloffay (approved):

pavolloffay:
@CCOLLOT is `OTEL_METRICS_EXPLICIT_BUCKET_HISTOGRAM` non standard env var? Is it documented anywhere?

pavolloffay (changes requested):
Is the env var documented anywhere? 

If it is operator specific it makes more sense to expose the setting as CR option.

CCOLLOT:
> Is the env var documented anywhere? 
> 
> 
> 
> If it is operator specific it makes more sense to expose the setting as CR option. 

It is indeed operator-specific. In the case of nodejs we need autoinstrumentation.ts to pick up the custom bucket list to pass the view to the NodeSDK.


To be clear, are you suggesting specifying on the Instrumentation object a parameter like 'spec.nodejs.histogram_buckets' and then injecting the bucket list as en environment variable into the init-container? 
Did you have another approach in mind to pass the bucket list to autoinstrumentation.ts?

pavolloffay:
> To be clear, are you suggesting specifying on the Instrumentation object a parameter like 'spec.nodejs.histogram_buckets'

I don't like using `env` vars for operator specific functionality, this should be really exposed as a standard CR option.

The env vars are there for upstream auto-instrumentation features or SDK, not features that the operator implements.

CCOLLOT:
@pavolloffay 
Fair enough, how would you recommend passing the bucket list from the CR parameter to the autoinstrumentation image (init-container) to create the View and pass it to the NodeSDK then?

autoinstrumentation/nodejs/src/autoinstrumentation.ts
```diff
@@ -50,11 +50,35 @@ function getMetricReader() {
     }
 }
 
+function getView() {
+  const buckets = process.env.OTEL_METRICS_EXPLICIT_BUCKET_HISTOGRAM;
+  const defaultHistogramBuckets = [
+    0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10,
```
pavolloffay, line 56: Does this follow https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#http-server ?

pavolloffay:
Passing through env var is fine, but in the CR I would expect people using CR properties. 

Could you please as well ping someone from nodejs team to review?

pichlermarc:
As others on this PR have said before this is a non-standard env var. 

If the PR was opened today against the OTel JS (or contrib) repos I'd reject it for that reason. It'd have to go through the specification process first, only then we'd accept the PR.

It is somewhat common practice that implementation specific non-standard env vars end up with a naming like `OTEL_NODE_*`, `OTEL_PYTHON_*`, so this one would have to prefix `OTEL_OPERATOR_` and then you can go from there but IIRC env vars for this type of feature have been discussed in the spec in the past but have been rejected because of a lack of granularity.

FWIW, the next release of `@opentelemetry/instrumentation-http` will include a functional env var `OTEL_SEMCONV_STABILITY_OPT_IN` that can be set to `http`, which will use the new (stable) bucket layout for HTTP metrics and a new unit as well.

CCOLLOT:
> Another thought: changing all histogram buckets with a view is likely not what users would want, because it overrides the [instrument advisory parameter for each instrument](https://github.com/open-telemetry/opentelemetry-specification/blob/f599c40953da38769610742282f380e20781b6b0/specification/metrics/api.md#instrument-advisory-parameter-explicitbucketboundaries), so one will end up with a lot of non-nonsensical data for any histograms that specify non-default buckets, which I expect to become the norm now that the feature has become stable. Instrument advisory parameters fro this are also implemented in OTel JS already.
> 
> There's plenty of histogram data that one could write that's not seconds (request sizes for instance). Any option to override all of them (like the View in the PR does) I'd consider quite harmful, actually.

Indeed, after taking a closer look at the spec I agree that overriding all of them is radical, to say the least.
What do you think dropping the idea of trying to provide a default View and instead just allow users of the Instrumentation CR to specify buckets for specific instrumentation? We could add a parameter to the Instrumentation CR like `spec.nodejs.histogramBuckets` such as:

```yaml
spec:
  nodejs:
    histogramBuckets:
      - instrumentName: "http.server.duration.bucket"
         buckets: [0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10]
      - instrumentName: "http.server.*"
         buckets: [x,y,z]
```

We would then inject an OTEL_OPERATOR_NODEJS_HISTOGRAM_BUCKETS env var to pass it to autoinstrumentation.ts, and create a view for each element of this array and pass the views to the NodeSDK ? 

In that case, we would by default use the SDK's default values and it would be the end-user's responsibility and freedom to choose override what they need.

pichlermarc:
@CCOLLOT Hmm, is there any other instrument other than the `http.server*` and `http.client*` instruments that you're trying to override (for your specific use-case that you're trying to address)?  🤔 

If `http.*` metrics are the ones you're interested in and the bucket layout you're expecting is the one from your example then there's no action necessary as it'll be included in the next release of `@opentelemetry/instrumentation-http` via the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable. 

If that's not what you're looking for, [OpenTelemetry Declarative configuration](https://github.com/open-telemetry/opentelemetry-configuration) will eventually be the more sustainable solution to the issue at hand because it includes all that functionality and more.

CCOLLOT:
It's good news that the SDK will fix the unit mismatch with better bucket defaults. However, I still think that users of the auto instrumentation injection should be able to control the exact bucket list through the `Instrumentation CR`.

I updated the PR with the following changes:
- Remove the default view with default values overriding all instruments.
- Introduce `spec.nodejs.histogrambuckets` to the `Instrumentation` CR
- Marshal `spec.nodejs.histogrambuckets` into JSON and the inject it as a environment variable into the container.
- In `autoinstrumentation.ts`, unmarshal the environment variable and pass a view per item in the array.

As a result, I've achieved the desired behavior with the following Instrumentation CR:

```yaml
---
kind: Instrumentation
metadata:
  name: default
spec:
  env:
    - name: OTEL_METRIC_EXPORT_INTERVAL
      value: "30000"
  exporter:
    endpoint: http://otel-collector.opentelemetry.svc:4317
  propagators:
    - tracecontext
    - baggage
  nodejs:
    histogramBuckets:
      - instrumentName: http*
        buckets: [0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 8]
  sampler:
    argument: "1"
```

I believe this implementation is better than the one I proposed earlier. For default buckets, it's using the SDK's defaults, which is a better separation of concerns. 
Therefore it will not modify the current default behavior but still give full flexibility to end-users of auto-instrumentation injection to adjust the cardinality/resolution of any metrics introduced by the SDK either by providing the full name of the instrument or a pattern expression using a `*` symbol.

PR #3449, opened 2024-11-11
Fix invalid manifests in e2e tests

swiatekm:
We have some invalid resource manifests in e2e tests. Kubernetes doesn't care, and just ignores the invalid fields, and so does chainsaw 0.2.8, but chainsaw 0.2.11 errors. Fix these problems so we can upgrade.

### DIFF
--- a/tests/e2e-instrumentation/instrumentation-nginx-contnr-secctx/01-install-app.yaml
+++ b/tests/e2e-instrumentation/instrumentation-nginx-contnr-secctx/01-install-app.yaml
@@ -22,7 +22,6 @@ spec:
         securityContext:
           runAsUser: 1000
           runAsGroup: 3000
-          fsGroup: 3000
         ports:
         - containerPort: 8765
         env:
@@ -33,7 +32,6 @@ spec:
             mountPath: /etc/nginx/nginx.conf
             subPath: nginx.conf
             readOnly: true
-        imagePullPolicy: Always
         resources:
           limits:
             cpu: "1"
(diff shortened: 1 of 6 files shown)


tests/e2e-instrumentation/instrumentation-nginx-contnr-secctx/01-install-app.yaml
```diff
@@ -22,7 +22,6 @@ spec:
         securityContext:
           runAsUser: 1000
           runAsGroup: 3000
-          fsGroup: 3000
```
iblancasa, line 25: Why is this marked as non valid? https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

swiatekm: This is a container security context, which doesn't have a `fsGroup` field: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#securitycontext-v1-core. As opposed to a Pod security context, which does have one: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#podsecuritycontext-v1-core

iblancasa:
@IshwarKanse can you verify if this breaks something on our end?

IshwarKanse:
@iblancasa @swiatekm I checked on a OCP cluster with Chainsaw v0.2.11, all tests passed. We are good to merge this PR.

iblancasa (approved):

Issue #3078, opened 2024-06-27
Using the create-rbac-permissions flag but clusterrole/clusterrolebindings are not being created for generated collector pods


paebersold-tyro:
### Component(s)

collector

## Description

I am running the opentelemetry-operator with the `--create-rbac-permissions` flag set.  When a new OpenTelemetryCollector resource is created (eg mode: daemonset) new pods are created and a new serviceaccount is created as well.  However no new clusterroles or clusterrolebindings are created.  This results in prometheus scrape errors due to lack of permissions for example.  Eg

```
E0627 04:07:32.435836       1 reflector.go:147] k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229: Failed to watch *v1.Pod: failed to list *v1.Pod: pods is forbidden: User "system:serviceaccount:observability:collector-with-ta-collector" cannot list resource "pods" in API group "" in the namespace "app-platform-monitoring"
```

No logs are generated on the operator-manager pod.

The clusterole that the operator manager is using has the access to create clusterroles/clusterrolebindings (I am deploying via the helm chart opentelemetry-operator version 0.62.0 (https://open-telemetry.github.io/opentelemetry-helm-charts)

Based on other issues raised previously it seems this flag was optional but now may no longer be required with the permissions being automatically granted based on existing access - I would like clarification on this aspect too please. 

## Steps to Reproduce

Run the opentelementry-operator with the create-rbac-permissions flag. 

## Expected Result

Clusterroles/bindings would be create when the new collector pods are created

## Actual Result

No new roles/bindings created

### Kubernetes Version

1.29

### Operator version

0.102.0

### Collector version

0.102.0

### Environment information

Serviceaccount used by manager
```
% kubectl -n observability get pods otel-operator-opentelemetry-operator-dfb985c65-ngh9n -o yaml | grep serviceAccount   
  serviceAccount: opentelemetry-operator
```
Clusterrolebinding
```
% kubectl get clusterrolebinding -o wide | grep opentelemetry-operator
otel-operator-opentelemetry-operator-manager               ClusterRole/otel-operator-opentelemetry-operator-manager             6d 
```
clusterrole for the operator manager (generated via helm chart)
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels:
    app.kubernetes.io/component: controller-manager
    app.kubernetes.io/instance: otel-operator
    app.kubernetes.io/managed-by: Helm
[...]

PR #3459, opened 2024-11-14
Add automatic RBAC creation for prometheus receiver

iblancasa:

- Resolves: #3078

### DIFF
--- a/apis/v1beta1/config.go
+++ b/apis/v1beta1/config.go
@@ -153,7 +153,7 @@ type Config struct {
 }
 
 // getRbacRulesForComponentKinds gets the RBAC Rules for the given ComponentKind(s).
-func (c *Config) getRbacRulesForComponentKinds(logger logr.Logger, componentKinds ...ComponentKind) ([]rbacv1.PolicyRule, error) {
+func (c *Config) getClusterRoleRbacRulesForComponentKinds(logger logr.Logger, componentKinds ...ComponentKind) ([]rbacv1.PolicyRule, error) {
 	var rules []rbacv1.PolicyRule
 	enabledComponents := c.GetEnabledComponents()
 	for _, componentKind := range componentKinds {
@@ -179,7 +179,7 @@ func (c *Config) getRbacRulesForComponentKinds(logger logr.Logger, componentKind
 		for componentName := range enabledComponents[componentKind] {
 			// TODO: Clean up the naming here and make it simpler to use a retriever.
 			parser := retriever(componentName)
-			if parsedRules, err := parser.GetRBACRules(logger, cfg.Object[componentName]); err != nil {
+			if parsedRules, err := parser.GetClusterRoleRules(logger, cfg.Object[componentName]); err != nil {
 				return nil, err
 			} else {
 				rules = append(rules, parsedRules...)
@@ -187,6 +187,68 @@ func (c *Config) getRbacRulesForComponentKinds(logger logr.Logger, componentKind
 		}
 	}
 	return rules, nil
+}
+
+// getRbacRolesForComponentKinds gets the RBAC Roles for the given ComponentKind(s).
+func (c *Config) getRbacRolesForComponentKinds(logger logr.Logger, otelCollectorName string, componentKinds ...ComponentKind) ([]*rbacv1.Role, error) {
+	var roles []*rbacv1.Role
+	enabledComponents := c.GetEnabledComponents()
+	for _, componentKind := range componentKinds {
+		var retriever components.ParserRetriever
+		var cfg AnyConfig
+		switch componentKind {
+		case KindReceiver:
+			retriever = receivers.ReceiverFor
+			cfg = c.Receivers
+		case KindExporter:
+			continue
+		case KindProcessor:
+			continue
+		case KindExtension:
+			continue
+		}
+		for componentName := range enabledComponents[componentKind] {
+			// TODO: Clean up the naming here and make it simpler to use a retriever.
+			parser := retriever(componentName)
+			if parsedRoles, err := parser.GetRbacRoles(logger, otelCollectorName, cfg.Object[componentName]); err != nil {
+				return nil, err
+			} else {
+				roles = append(roles, parsedRoles...)
+			}
+		}
+	}
+	return roles, nil
+}
+
+// getRbacRoleBindingsForComponentKinds gets the RBAC RoleBindings for the given ComponentKind(s).
+func (c *Config) getRbacRoleBindingsForComponentKinds(logger logr.Logger, serviceAccountName string, otelCollectorName string, otelCollectorNamespace string, componentKinds ...ComponentKind) ([]*rbacv1.RoleBinding, error) {
+	var roleBindings []*rbacv1.RoleBinding
+	enabledComponents := c.GetEnabledComponents()
+	for _, componentKind := range componentKinds {
+		var retriever components.ParserRetriever
+		var cfg AnyConfig
+		switch componentKind {
+		case KindReceiver:
+			retriever = receivers.ReceiverFor
+			cfg = c.Receivers
+		case KindExporter:
+			continue
+		case KindProcessor:
+			continue
+		case KindExtension:
+			continue
+		}
+		for componentName := range enabledComponents[componentKind] {
+			// TODO: Clean up the naming here and make it simpler to use a retriever.
+			parser := retriever(componentName)
+			if parsedRoleBindings, err := parser.GetRbacRoleBindings(logger, otelCollectorName, cfg.Object[componentName], serviceAccountName, otelCollectorNamespace); err != nil {
+				return nil, err
+			} else {
+				roleBindings = append(roleBindings, parsedRoleBindings...)
+			}
+		}
+	}
+	return roleBindings, nil
 }
 
 // getPortsForComponentKinds gets the ports for the given ComponentKind(s).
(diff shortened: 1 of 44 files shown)


internal/components/receivers/prometheus.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func generatePrometheusReceiverRoles(logger logr.Logger, config prometheusReceiverConfig, componentName string, otelCollectorName string) ([]*rbacv1.Role, error) {
+	if config.Config == nil {
+		return nil, nil
+	}
+
+	if config.Config.ScrapeConfigs == nil {
+		return nil, nil
+	}
```
swiatekm, line 54: Should this cover the target allocator as well?

controllers/common.go
```diff
@@ -144,16 +144,22 @@ func reconcileDesiredObjects(ctx context.Context, kubeClient client.Client, logg
 			"object_kind", desired.GetObjectKind(),
 		)
 		if isNamespaceScoped(desired) {
-			if setErr := ctrl.SetControllerReference(owner, desired, scheme); setErr != nil {
-				l.Error(setErr, "failed to set controller owner reference to desired")
-				errs = append(errs, setErr)
-				continue
+			switch desired.(type) {
+			case *rbacv1.Role, *rbacv1.RoleBinding:
+				l.Info("skipping setting controller reference for role or rolebinding")
```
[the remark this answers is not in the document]
iblancasa: The problem is that those `Roles` and `Rolebindings` will be created from a `OpenTelemetryCollector` instance from another namespace.

internal/components/receivers/prometheus.go
iblancasa: For the TA we should do it in a different PR... otherwise this will become huge

controllers/common.go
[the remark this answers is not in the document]
swiatekm: Ah, because we actually check in the prometheus receiver configuration which namespaces it wants service discovery from, and then create a Role in each of them? Makes sense.

From what I've skimmed of this PR, it looks good, but I'd appreciate if you summarized in the PR description what kinds of changes you're making and why. The original issue has a description of what the user-facing behaviour should be, but this is a sufficiently complicated changes that I'd also like to see one for the implementation.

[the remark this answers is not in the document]
iblancasa: > From what I've skimmed of this PR, it looks good, but I'd appreciate if you summarized in the PR description what kinds of changes you're making and why. The original issue has a description of what the user-facing behaviour should be, but this is a sufficiently complicated changes that I'd also like to see one for the implementation.

Sorry for that. I tried to keep it simple but... I think the description should help much more now.

Thanks for your feedback! @swiatekm

internal/manifests/collector/collector.go
```diff
@@ -98,6 +98,24 @@ func Build(params manifests.Params) ([]client.Object, error) {
...
+	}
+
+	roleBindings, err := RoleBinding(params)
+	if err != nil {
+		return nil, err
+	}
+	// NOTE: we cannot just unpack the slice, the type checker doesn't coerce the type correctly.
+	for _, roleBinding := range roleBindings {
+		resourceManifests = append(resourceManifests, roleBinding)
+	}
```
swiatekm, line 117: I'd prefer to fold these into the manifest factory framework, but the idea of changing all the current functions to return slices doesn't seem appealing. Maybe we can instead do the conversion in `manifests.Factory`? WDYT @jaronoff97 ?

jaronoff97: Yeah i think modifying the manifest factory to have a method that allows returning a slice makes sense to me. You could accomplish that in a similar way we do the factory without error today.

iblancasa: I will try. Thanks!

zarbis:
Really anticipating this one!

iblancasa:
@jaronoff97 can you take a look?

apis/v1beta1/config.go
```diff
@@ -179,6 +179,68 @@ func (c *Config) getRbacRulesForComponentKinds(logger logr.Logger, componentKind
 	return rules, nil
 }
 
+// getRbacRolesForComponentKinds gets the RBAC Roles for the given ComponentKind(s).
+func (c *Config) getRbacRolesForComponentKinds(logger logr.Logger, otelCollectorName string, componentKinds ...ComponentKind) ([]*rbacv1.Role, error) {
```
jaronoff97, line 183: these methods are starting to really really bloat this file. I think it's time we figure out a better way to the delegation here. I'm thinking we could make a method that wraps the iteration logic and the enabled components logic and then is passed in the parser method that should be called. LMK if you want to pair on this.

internal/components/builder.go
```diff
@@ -15,19 +15,21 @@ import (
...
+	appProtocol         *string
+	targetPort          intstr.IntOrString
+	nodePort            int32
+	name                string
+	port                int32
+	defaultRecAddr      string
+	portParser          PortParser[ComponentConfigType]
+	clusterRoleRulesGen ClusterRoleRulesGenerator[ComponentConfigType]
+	roleGen             RoleGenerator[ComponentConfigType]
+	roleBindingGen      RoleBindingGenerator[ComponentConfigType]
```
jaronoff97, line 28: do we also need to have a cluster role binding gen?

internal/components/generic_parser.go
```diff
@@ -67,15 +69,38 @@ func (g *GenericParser[T]) GetReadinessProbe(logger logr.Logger, config interfac
...
-func (g *GenericParser[T]) GetRBACRules(logger logr.Logger, config interface{}) ([]rbacv1.PolicyRule, error) {
-	if g.rbacGen == nil {
+func (g *GenericParser[T]) GetClusterRoleRules(logger logr.Logger, config interface{}) ([]rbacv1.PolicyRule, error) {
+	if g.clusterRoleRulesGen == nil {
 		return nil, nil
 	}
 	var parsed T
 	if err := mapstructure.Decode(config, &parsed); err != nil {
 		return nil, err
 	}
```
jaronoff97, line 79: this is other logic that would probably be good to abstract...

internal/components/receivers/prometheus.go
```diff
@@ -0,0 +1,149 @@
...
+	Role       string     `mapstructure:"role"`
+}
+
+type scrapeConfig struct {
+	KubernetesSDConfigs []kubernetesSDConfig `mapstructure:"kubernetes_sd_configs"`
+	JobName             string               `mapstructure:"job_name"`
+}
+
+type prometheusConfig struct {
+	ScrapeConfigs *[]scrapeConfig `mapstructure:"scrape_configs"`
```
jaronoff97, line 29: does this need to be a pointer to a slice? an empty slice is nil already?

internal/manifests/collector/collector.go
```diff
@@ -74,6 +75,21 @@ func Build(params manifests.Params) ([]client.Object, error) {
 		}
 	}
 
+	manifestSliceFactories = append(
+		manifestSliceFactories,
+		manifests.FactorySlice(Role),
+		manifests.FactorySlice(RoleBinding),
+		manifests.FactorySlice(Routes),
+	)
```
jaronoff97, line 83: i like what you did here :)

internal/manifests/collector/route.go
```diff
@@ -86,5 +87,12 @@ func Routes(params manifests.Params) ([]*routev1.Route, error) {
 			},
 		}
 	}
-	return routes, nil
+
+	// Convert []*routev1.Route to []client.Object
+	result := make([]client.Object, len(routes))
+	for i, route := range routes {
+		result[i] = route
+	}
```
jaronoff97, line 95: Could this result in an improper copy because of shadowing?

internal/manifests/mutate.go
```diff
@@ -224,12 +224,18 @@ func mutateRole(existing, desired *rbacv1.Role) {
 	existing.Annotations = desired.Annotations
 	existing.Labels = desired.Labels
 	existing.Rules = desired.Rules
+	// This role can exists in a different namespace than the otel collector, so we need to remove the owner references
+	// since cross namespace owner references are not allowed
+	existing.SetOwnerReferences(nil)
```
jaronoff97, line 229: why does the role not exist in the same namespace?

jaronoff97: this isn't a cluster role right?

jaronoff97 (commented):
Sorry for the very delayed review here. Overall I think this is on the right track, but I think we need to rework the abstraction in the config.go to be simpler (this can be done in a follow up to this, or a side PR to this). I had a few questions / suggestions otherwise.

apis/v1beta1/config.go
iblancasa: It totally makes sense to me.

internal/manifests/mutate.go
iblancasa: The thing is that the resource can be created by an OpenTelemetry Collector from another namespace https://github.com/open-telemetry/opentelemetry-operator/pull/3459/files#r1865462426

internal/components/builder.go
iblancasa: It is not needed because it is not associated with a specific namespace.

internal/components/generic_parser.go
iblancasa: Yes

internal/components/receivers/prometheus.go
iblancasa: Sorry because I made this changes some time ago and I don't remember all the details. But I know there was something that didn't allow me to not use a pointer. Like an error while parsing unset `scrape_configs` or something.

jaronoff97: ah you probably need to set `omitempty` as an option on mapstructure?

internal/manifests/collector/route.go
iblancasa: I'm not sure. Let me check and improve the code.

iblancasa:
> Sorry for the very delayed review here.

Don't worry 😆!

> Overall I think this is on the right track, but I think we need to rework the abstraction in the config.go to be simpler (this can be done in a follow up to this, or a side PR to this).

Sure. I think it would make more sense to do as part of a separate PR because this one is becoming so big.

zarbis:
@iblancasa hey, I wonder if there is any work left to be done here? My setup would really benefit from this RP!

iblancasa:
> @iblancasa hey, I wonder if there is any work left to be done here? My setup would really benefit from this RP!

https://github.com/open-telemetry/opentelemetry-operator/pull/3459#discussion_r1983934390
https://github.com/open-telemetry/opentelemetry-operator/pull/3459#discussion_r1984153518

PR #3461, opened 2024-11-15
Update api.md

dominicqi:
Reference Prometheus operator Api spec, the pod/service monitorSelector nil will matches no objects.

Fix api.md,  PrometheusCR.ServiceMonitorSelector PodMonitorSelector nil will matches no objects

[Prometheus CR Document](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#monitoring.coreos.com/v1.Prometheus)


iblancasa (changes requested):
You need to do that change in the `apis/v1alpha1/opentelemetrycollector_types.go` file

PR #3473, opened 2024-11-19
Bump base memory requirements for python and go

jcreixell:
  - When auto-instrumenting applications, I have noticed that default memory limits are too tight for some languages. This leads to the following:
    - Intermitent OOMKilled events in init container when auto-instrumenting python applications. Eventually the pods are able to start.
    - OOMKilled events for sidecar containers in go applications. The pods are not able to start.
  - 64Mi seems to be enough to fix these issues. While some tweaking by users may still be necessary, the operator should work out-of-the-box for all supported languages.

https://github.com/open-telemetry/opentelemetry-operator/issues/3479

### DIFF
--- a/.chloggen/bump-base-instrumentation-mem-limit.yaml
+++ b/.chloggen/bump-base-instrumentation-mem-limit.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: auto-instrumentation
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Bump base memory requirements for python and go
+
+# One or more tracking issues related to the change
+issues: []
+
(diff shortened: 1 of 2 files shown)


iblancasa (commented):
Please, add a changelog

swiatekm:
Can you reproduce the OOMKills in a kind cluster? I'm not against bumping the requirements, but I don't understand why these containers would need it. The only thing they actually do is copy some files over to a shared volume.

jcreixell:
I am able to consistently reproduce the go OOMs with both k3d and kind:

kind

Interestingly, the python OOMs only happen with k3d, not kind.

Screenshot from 2024-11-20 18-33-10

Screenshot from 2024-11-20 18-33-55

I am using a standard setup following the README. For testing, I use the rolldice applications with their manual instrumentation code stripped out.

(I will move this to an issue and link it in the changelog)

iblancasa (approved):

swiatekm (approved):
This looks good to me, and I'm fine merging it as-is. But I'd like to understand why our e2e tests don't catch it, and ideally modify them to do so. @jcreixell could you open a follow-up issue to track that?

jcreixell:
@swiatekm  sure, there we go https://github.com/open-telemetry/opentelemetry-operator/issues/3484

PR #3482, opened 2024-11-21
Add allocation_fallback_strategy option as fallback strategy for per-node strategy

### LINKED ISSUE #3477: Add fallback for "per-node" allocation strategy
### Component(s)

target allocator

The operator does not allow me to deploy a daemonset collector with any other allocation strategy than per-node. This is very reasonable, but if I add a prometheus scrape job that scrapes a single non-node-specific target it will not be allocated as that is currently the expected functionality of the "per-node" allocation strategy.

See below example of failure to deploy as daemonset with "consistent-hashing" allocation strategy.

`Error: UPGRADE FAILED: an error occurred while rolling back the release. original upgrade error: cannot patch "opentelemetry" with kind OpenTelemetryCollector: admission webhook "vopentelemetrycollectorcreateupdatebeta.kb.io" denied the request: the OpenTelemetry Collector mode is set to daemonset, which must be used with target allocation strategy per-node`

I think there should be a fallback, ideally just using "consistent-hashing" or "least-weighted" for the leftover targets after "per-node" allocation is done.

_No response_

### Additional context

The possibility of adding a fallback was talked about in [this issue](https://github.com/open-telemetry/opentelemetry-operator/issues/1828), but never acted upon.

lhp-nemlig:

Adds allocation_fallback_strategy option as fallback strategy for per-node strategy, such that targets that are not related to a specific node can be assigned. 

The fallback strategy is configurable via the config option allocation_fallback_strategy

- Resolves: #3477 

Fixed per-node tests to expect non-node target to be assigned according to fallback strategy

### DIFF
--- a/.chloggen/add_fallback_strategy_for_per_node_strategy.yaml
+++ b/.chloggen/add_fallback_strategy_for_per_node_strategy.yaml
@@ -5,7 +5,7 @@ change_type: 'enhancement'
 component: target allocator
 
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Added allocation_fallback_strategy option as fallback strategy for per-node allocation strategy
+note: Added allocation_fallback_strategy option as fallback strategy for per-node allocation strategy, can be enabled with feature flag operator.targetallocator.fallbackstrategy
 
 # One or more tracking issues related to the change
 issues: [3477]
@@ -16,4 +16,6 @@ issues: [3477]
 subtext: |
   If using per-node allocation strategy, targets that are not attached to a node will not
   be allocated. As the per-node strategy is required when running as a daemonset, it is 
-  not possible to assign some targets under a daemonset deployment
+  not possible to assign some targets under a daemonset deployment.
+  Feature flag operator.targetallocator.fallbackstrategy has been added and results in consistent-hashing
+  being used as the fallback allocation strategy for "per-node" only at this time.
--- a/.chloggen/add_fallback_strategy_for_per_node_strategy.yaml
+++ b/.chloggen/add_fallback_strategy_for_per_node_strategy.yaml
@@ -1,0 +1,19 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: target allocator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Added least-weighted as fallback strategy for per-node allocation strategy
+
+# One or more tracking issues related to the change
+issues: [3477]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  If using per-node allocation strategy, targets that are not attached to a node will not
+  be allocated. As the per-node strategy is required when running as a daemonset, it is 
+  not possible to assign some targets under a daemonset deployment
--- a/.chloggen/add_fallback_strategy_for_per_node_strategy.yaml
+++ b/.chloggen/add_fallback_strategy_for_per_node_strategy.yaml
@@ -5,7 +5,7 @@ change_type: 'enhancement'
 component: target allocator
 
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Added consistent-hashing as fallback strategy for per-node allocation strategy
+note: Added allocation_fallback_strategy option as fallback strategy for per-node allocation strategy
 
 # One or more tracking issues related to the change
 issues: [3477]
--- a/.chloggen/add_fallback_strategy_for_per_node_strategy.yaml
+++ b/.chloggen/add_fallback_strategy_for_per_node_strategy.yaml
@@ -5,7 +5,7 @@ change_type: 'enhancement'
 component: target allocator
 
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Added least-weighted as fallback strategy for per-node allocation strategy
+note: Added consistent-hashing as fallback strategy for per-node allocation strategy
 
 # One or more tracking issues related to the change
 issues: [3477]
--- a/cmd/otel-allocator/allocation/per_node_test.go
+++ b/cmd/otel-allocator/allocation/per_node_test.go
@@ -154,7 +154,7 @@ func TestAllocationPerNodeUsingFallback(t *testing.T) {
 	// verify allocation to nodes
 	for targetHash, item := range targetList {
 		actualItem, found := actualItems[targetHash]
-		// if third target, should be skipped
+		// if third target, should be allocated by the fallback strategy
 		assert.True(t, found, "target with hash %s not found", item.Hash())
 
 		// only the first two targets should be allocated
--- a/cmd/otel-allocator/allocation/per_node_test.go
+++ b/cmd/otel-allocator/allocation/per_node_test.go
@@ -118,7 +118,7 @@ func TestAllocationPerNodeUsingFallback(t *testing.T) {
 		{Name: "test", Value: "test2"},
 		{Name: "__meta_kubernetes_node_name", Value: "node-1"},
 	}
-	// no label, should be skipped
+	// no label, should be allocated by the fallback strategy
 	thirdLabels := labels.Labels{
 		{Name: "test", Value: "test3"},
 	}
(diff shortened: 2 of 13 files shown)


cmd/otel-allocator/allocation/strategy.go
```diff
@@ -149,7 +149,7 @@ func init() {
 		panic(err)
 	}
 	err = Register(perNodeStrategyName, func(log logr.Logger, opts ...AllocationOption) Allocator {
-		return newAllocator(log, newPerNodeStrategy(), opts...)
+		return newAllocator(log, newPerNodeStrategy(newleastWeightedStrategy()), opts...)
```
swiatekm, line 152: I would much prefer the fallback strategy be `consistent-hashing`, and that should also be the default if we make this configurable.

The reason for this is that the `least-weighted` strategy isn't stable over time - the end result depends on what order changes to the set of targets arrive. To put it another way, with `least-weighted`, if you start multiple target allocator Pods, there's no guarantee that they give the same target allocation unless they're started at the exact same time. In comparison, `consistent-hashing` is stable, and you can scale the Target Allocator with it.

swiatekm (commented):
This is a breaking change, so we either have to mark it as such, or hide it behind a feature gate. I'd prefer the latter, but am willing to be convinced otherwise. We'll discuss it during today's SIG meeting, feel free to join if you'd like to contribute @lhp-nemlig.

On a more technical note, you'll need to sign the CLA and add a changelog entry before we can merge this PR.

cmd/otel-allocator/allocation/strategy.go
lhp-nemlig: I changed the strategy to consistent-hashing

lhp-nemlig:
> This is a breaking change, so we either have to mark it as such, or hide it behind a feature gate. I'd prefer the latter, but am willing to be convinced otherwise. We'll discuss it during today's SIG meeting, feel free to join if you'd like to contribute @lhp-nemlig.
> 
> On a more technical note, you'll need to sign the CLA and add a changelog entry before we can merge this PR.

I have changed the PR to add fallback strategy as a config option, making this a non-breaking change as the default is NO fallback. 
That makes the changes a little more extensive of course, but I think I have implemented this in a mostly reasonable way :)

swiatekm:
> > This is a breaking change, so we either have to mark it as such, or hide it behind a feature gate. I'd prefer the latter, but am willing to be convinced otherwise. We'll discuss it during today's SIG meeting, feel free to join if you'd like to contribute @lhp-nemlig.
> > On a more technical note, you'll need to sign the CLA and add a changelog entry before we can merge this PR.
> 
> @swiatekm I have changed the PR to add fallback strategy as a config option, making this a non-breaking change as the default is NO fallback. That makes the changes a little more extensive of course, but I think I have implemented this in a mostly reasonable way :)

I'm ok with this, but keep in mind that in order for this to actually be usable, you'll also need to add an attribute to the OpenTelemetryCollector CRD. That's not a problem, but since in that case you're changing the public API, it's going to invite more scrutiny during review. Just FYI.

lhp-nemlig:
> > > This is a breaking change, so we either have to mark it as such, or hide it behind a feature gate. I'd prefer the latter, but am willing to be convinced otherwise. We'll discuss it during today's SIG meeting, feel free to join if you'd like to contribute @lhp-nemlig.
> > > On a more technical note, you'll need to sign the CLA and add a changelog entry before we can merge this PR.
> > 
> > 
> > @swiatekm I have changed the PR to add fallback strategy as a config option, making this a non-breaking change as the default is NO fallback. That makes the changes a little more extensive of course, but I think I have implemented this in a mostly reasonable way :)
> 
> I'm ok with this, but keep in mind that in order for this to actually be usable, you'll also need to add an attribute to the OpenTelemetryCollector CRD. That's not a problem, but since in that case you're changing the public API, it's going to invite more scrutiny during review. Just FYI.

Ok @swiatekm, I expect to join the SIG meeting and you can tell me what I need to do in order to complete this

swiatekm:
@lhp-nemlig as per our discussion during the SIG meeting, the simplest way to proceed with this change without needing to add attributes to the CRDs, would be to:

1. Keep your current changes making it configurable on the target allocator.
2. Add a feature flag which, when enabled, would set the fallback to the `consistent-hashing` strategy.

You can add feature flags [here](https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/featuregate/featuregate.go) and the target allocator ConfigMap is generated [here](https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/manifests/targetallocator/configmap.go#L39). Feature flags are global, so you can just import the package, and add the config value for the fallback strategy if the flag is enabled. Does that make sense?

lhp-nemlig:
@swiatekm Thanks for the clear instructions, I believe everything should be in order now.

cmd/otel-allocator/allocation/per_node_test.go
```diff
@@ -93,6 +103,78 @@ func TestAllocationPerNode(t *testing.T) {
...
+	expectedTargetLen := len(targetList)
+	assert.Len(t, actualItems, expectedTargetLen)
+
+	// verify allocation to nodes
+	for targetHash, item := range targetList {
+		actualItem, found := actualItems[targetHash]
+		// if third target, should be skipped
+		assert.True(t, found, "target with hash %s not found", item.Hash())
+
+		// only the first two targets should be allocated
```
swiatekm, line 160: This comment seems unnecessary. In this test, we expect all targets to be allocated.

cmd/otel-allocator/main.go
```diff
@@ -81,7 +81,7 @@ func main() {
 	log := ctrl.Log.WithName("allocator")
 
 	allocatorPrehook = prehook.New(cfg.FilterStrategy, log)
-	allocator, err = allocation.New(cfg.AllocationStrategy, log, allocation.WithFilter(allocatorPrehook))
+	allocator, err = allocation.New(cfg.AllocationStrategy, log, allocation.WithFilter(allocatorPrehook), allocation.WithFallbackStrategy(cfg.AllocationFallbackStrategy))
```
swiatekm, line 84: I'd prefer if we only used allocation.WithFallbackStrategy here if it is actually set. Relying on the function to correctly deal with an empty strategy name feels unnecessarily error-prone.

internal/manifests/targetallocator/configmap.go
```diff
@@ -90,6 +90,11 @@ func ConfigMap(params Params) (*corev1.ConfigMap, error) {
 	} else {
 		taConfig["allocation_strategy"] = v1beta1.TargetAllocatorAllocationStrategyConsistentHashing
 	}
+
+	if featuregate.EnableTargetAllocatorFallbackStrategy.IsEnabled() {
+		taConfig["allocation_fallback_strategy"] = v1beta1.TargetAllocatorAllocationStrategyConsistentHashing
+	}
```
swiatekm, line 96: Could you check this in a unit test as well?

.chloggen/add_fallback_strategy_for_per_node_strategy.yaml
```diff
@@ -0,0 +1,19 @@
...
+# One or more tracking issues related to the change
+issues: [3477]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  If using per-node allocation strategy, targets that are not attached to a node will not
+  be allocated. As the per-node strategy is required when running as a daemonset, it is 
+  not possible to assign some targets under a daemonset deployment
```
swiatekm, line 19: Could you update the note and subtext with information about the feature flag?

swiatekm (commented):
Looks good to me overall, had some minor comments and nitpicks. I'm fine merging this PR as-is when they're addressed, but in an ideal world we'd also include an e2e test showing the fallback strategy assigning targets without Nodes when the flag is enabled. I'm ok with that happening in a follow-up though, as this is an off-by-default feature flag, and writing these kinds of tests in chainsaw isn't very straightforward.

cmd/otel-allocator/allocation/per_node_test.go
lhp-nemlig: Removed these 2 unnecessary comments

cmd/otel-allocator/main.go
lhp-nemlig: Added conditional here

internal/manifests/targetallocator/configmap.go
lhp-nemlig: Added unit test for feature flag toggle

.chloggen/add_fallback_strategy_for_per_node_strategy.yaml
lhp-nemlig: Updated note and subtext

swiatekm (approved):

jaronoff97 (approved):
Looks great! Thank you for your contribution :bow:

vidam-io:
Thanks for this feature! We're using it in production to scrape external VM metrics via ServiceMonitor with per-node allocation. Works great.

Issue #3454, opened 2024-11-13
Operator doesn't delete components when removed from configuration


sfc-gh-dguy:
### Component(s)

collector, target allocator

## Description
I have an otel operator that creates otel collectors as daemonset and an otel target allocator based on an OpenTelemetryCollector CR.
I removed the target allocator part from the OpenTelemetryCollector CR but the target allocator wasn't deleted.
Only when I manually delete it, it's removed and is not recreated as it was removed from the config.
I also deleted the whole OpenTelemetryCollector CR and both the daemonset and the target allocator weren't deleted.

## Steps to Reproduce
1. Create an OpenTelemetryCollector CR that defines collectors daemonset and atarget allocator deployment
2. Let the operator create the components
3. Delete the target allocator part from the CR

## Expected Result
Target allocator should be deleted when removed from the CR

## Actual Result
Target allocator wasn't deleted when removed from the CR

### Kubernetes Version

1.29.8

### Operator version

0.103.0

### Collector version

0.111.0

## Environment
OS: (e.g., "Ubuntu 20.04")
Compiler(if manually compiled): (e.g., "go 14.2")

### Log output

```shell
These are the only prints I see in the operator logs:
{"level":"INFO","timestamp":"2024-11-13T16:43:17.143921523Z","logger":"controllers.OpenTelemetryCollector","message":"pdb field is unset in Spec, skipping podDisruptionBudget creation"}
{"level":"INFO","timestamp":"2024-11-13T16:43:17.362086635Z","logger":"controllers.OpenTelemetryCollector","message":"pdb field is unset in Spec, skipping podDisruptionBudget creation"}
{"level":"INFO","timestamp":"2024-11-13T16:43:17.543764697Z","logger":"controllers.OpenTelemetryCollector","message":"pdb field is unset in Spec, skipping podDisruptionBudget creation"}
{"level":"INFO","timestamp":"2024-11-13T16:47:27.3789637Z","logger":"controllers.OpenTelemetryCollector","message":"pdb field is unset in Spec, skipping podDisruptionBudget creation"}
{"level":"INFO","timestamp":"2024-11-13T16:47:27.554856277Z","logger":"controllers.OpenTelemetryCollector","message":"pdb field is unset in Spec, skipping podDisruptionBudget creation"}
[...]

PR #3494, opened 2024-11-25
Fix optional resource deletion for collector CR

swiatekm:
**Description:**

I've refactored and cleaned up our owned resource tracking:

* I've added a method that returns all the tracked resources for the reconciler. During registration, we loop over that list.
* To be able to easily find the owned resources, I added an index with a custom field, similar to what the [kubebuilder docs](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/cronjob-tutorial/testdata/project/internal/controller/cronjob_controller.go) suggest. This makes test setup more annoying, as we need to make sure to use the caching client in controller tests, but makes the code much more robust.
* We now track ownership of all namespaced resources instead of just a subset of them. This way we actually delete a target allocator Deployment if it's disabled in the collector CR.

- Resolves: #3454

* Added a controller test for deleting optional resources.
* Added a e2e test for this as well.
* Added a controller test for versioned ConfigMaps.
* Added more unit tests for the controller functions.

### DIFF
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -21,8 +21,8 @@ import (
 
 	"github.com/go-logr/logr"
 	rbacv1 "k8s.io/api/rbac/v1"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
@@ -119,18 +119,32 @@ func BuildTargetAllocator(params targetallocator.Params) ([]client.Object, error
 // getList queries the Kubernetes API to list the requested resource, setting the list l of type T.
 func getList[T client.Object](ctx context.Context, cl client.Client, l T, options ...client.ListOption) (map[types.UID]client.Object, error) {
 	ownedObjects := map[types.UID]client.Object{}
-	list := &unstructured.UnstructuredList{}
 	gvk, err := apiutil.GVKForObject(l, cl.Scheme())
 	if err != nil {
 		return nil, err
 	}
-	list.SetGroupVersionKind(gvk)
-	err = cl.List(ctx, list, options...)
+	gvk.Kind = fmt.Sprintf("%sList", gvk.Kind)
+	list, err := cl.Scheme().New(gvk)
+	if err != nil {
+		return nil, fmt.Errorf("unable to list objects of type %s: %w", gvk.Kind, err)
+	}
+
+	objList := list.(client.ObjectList)
+
+	err = cl.List(ctx, objList, options...)
 	if err != nil {
 		return ownedObjects, fmt.Errorf("error listing %T: %w", l, err)
 	}
-	for i := range list.Items {
-		ownedObjects[list.Items[i].GetUID()] = &list.Items[i]
+	objs, err := apimeta.ExtractList(objList)
+	if err != nil {
+		return ownedObjects, fmt.Errorf("error listing %T: %w", l, err)
+	}
+	for i := range objs {
+		typedObj, ok := objs[i].(T)
+		if !ok {
+			return ownedObjects, fmt.Errorf("error listing %T: %w", l, err)
+		}
+		ownedObjects[typedObj.GetUID()] = typedObj
 	}
 	return ownedObjects, nil
 }
--- a/controllers/opentelemetrycollector_controller.go
+++ b/controllers/opentelemetrycollector_controller.go
@@ -47,7 +47,7 @@ import (
 	"github.com/open-telemetry/opentelemetry-operator/internal/config"
 	"github.com/open-telemetry/opentelemetry-operator/internal/manifests"
 	"github.com/open-telemetry/opentelemetry-operator/internal/manifests/collector"
-	"github.com/open-telemetry/opentelemetry-operator/internal/naming"
+	"github.com/open-telemetry/opentelemetry-operator/internal/manifests/manifestutils"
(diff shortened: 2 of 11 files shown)


controllers/common.go
```diff
@@ -119,18 +119,32 @@ func BuildTargetAllocator(params targetallocator.Params) ([]client.Object, error
...
 func getList[T client.Object](ctx context.Context, cl client.Client, l T, options ...client.ListOption) (map[types.UID]client.Object, error) {
 	ownedObjects := map[types.UID]client.Object{}
-	list := &unstructured.UnstructuredList{}
 	gvk, err := apiutil.GVKForObject(l, cl.Scheme())
 	if err != nil {
 		return nil, err
 	}
-	list.SetGroupVersionKind(gvk)
-	err = cl.List(ctx, list, options...)
+	gvk.Kind = fmt.Sprintf("%sList", gvk.Kind)
```
swiatekm, line 126: This feels really stupid, but I couldn't find an idiomatic way to get a `xxxList` instance given an `xxx` instance.

controllers/opentelemetrycollector_controller.go
```diff
@@ -152,25 +126,34 @@ func (r *OpenTelemetryCollectorReconciler) findClusterRoleObjects(ctx context.Co
 	return ownedObjects, nil
 }
 
-// getConfigMapsToRemove returns a list of ConfigMaps to remove based on the number of ConfigMaps to keep.
-// It keeps the newest ConfigMap, the `configVersionsToKeep` next newest ConfigMaps, and returns the remainder.
-func (r *OpenTelemetryCollectorReconciler) getConfigMapsToRemove(configVersionsToKeep int, configMapList *corev1.ConfigMapList) []corev1.ConfigMap {
+// removeKeptConfigMaps removes old ConfigMaps that we want to keep from the map of owned objects.
+// Normally the controller would delete them after determining they're not in the list of desired objects generated
+// from the OpenTelemetryCollector CR, but we want to keep them around.
+func removeKeptConfigMapVersions(configVersionsToKeep int, ownedObjects map[types.UID]client.Object) {
```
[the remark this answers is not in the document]
swiatekm: I think the best way would be to change the function to return the CMs that should be kept, then it could be named `getConfigMapsToKeep`, it would take a slice of configmaps and return one. Then the calling function could build the slice from the map, call the function, and delete the output from the map.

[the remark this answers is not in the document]
swiatekm: Refactored it to a very simple form and added some unit tests.

controllers/opentelemetrycollector_controller.go
```diff
@@ -138,7 +109,10 @@ func (r *OpenTelemetryCollectorReconciler) findClusterRoleObjects(ctx context.Co
 	// Remove cluster roles and bindings.
 	// Users might switch off the RBAC creation feature on the operator which should remove existing RBAC.
 	listOpsCluster := &client.ListOptions{
-		LabelSelector: labels.SelectorFromSet(manifestutils.SelectorLabels(params.OtelCol.ObjectMeta, collector.ComponentOpenTelemetryCollector)),
+		LabelSelector: labels.SelectorFromSet(map[string]string{
+			"app.kubernetes.io/managed-by": "opentelemetry-operator",
+			"app.kubernetes.io/instance":   naming.Truncate("%s.%s", 63, params.OtelCol.Namespace, params.OtelCol.Name),
```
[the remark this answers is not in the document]
swiatekm: Reverted this change completely, turned out not to be necessary.

controllers/common.go
swiatekm: FYI, if I go back to the previous version using `Unstructured`, then the cache doesn't work even if I explicitly enable unstructured caching in the client. I think I'd need to change the indexer as well.

controllers/opentelemetrycollector_controller.go
```diff
@@ -82,51 +85,40 @@ type Params struct {
...
-	}
-	if params.Config.CreateRBACPermissions() == rbac.Available {
-		objs, err := r.findClusterRoleObjects(ctx, params)
-		if err != nil {
-			return nil, err
-		}
-		for uid, object := range objs {
-			ownedObjects[uid] = object
+		// save Collector ConfigMaps into a separate slice, we need to do additional filtering on them
+		if _, ok := objectType.(*corev1.ConfigMap); ok {
```
jaronoff97, line 103: nit: let's change this to a switch, which IMO is more idiomatic and extensible in case we need to do anything special for other types.

controllers/opentelemetrycollector_controller.go
```diff
@@ -82,51 +85,40 @@ type Params struct {
...
+			for _, object := range objs {
+				if !featuregate.CollectorUsesTargetAllocatorCR.IsEnabled() && object.GetLabels()["app.kubernetes.io/component"] != "opentelemetry-collector" {
```
jaronoff97, line 105: should we check for something more clearly related to the operator? I'm worried this is too broad and will conflict with what a helm installation of the collector chart installs.

jaronoff97: yeah you're right, i was thinking that helm adds the same label and if someone did have the gate enabled this would continue, but the fact that we check the owner ref means this point is moot.

controllers/common.go
swiatekm: For what's it's worth, controller-runtime itself does these kinds of things as well: https://github.com/kubernetes-sigs/controller-runtime/blob/f529320b4c0fa5205067dec4eaeabc3c6cf62d25/pkg/client/client.go#L230.

And if the object doesn't support `List`, well, the test won't pass. We could even add a test with a fake client to verify.

I would like a solution without these kinds of hacks, but I'm losing hope that it's actually possible.

jaronoff97: yeah it's a huge bummer.

jaronoff97 (approved):
This LGTM, but I would like some other approvers / maintainers review on this too

iblancasa (approved):

Issue #3450, opened 2024-11-12
Automatically update tool versions


swiatekm:
### Describe the issue you're reporting

Our tool versions are pinned in the Makefile, and we don't have any process for updating them. Ideally this should happen the same way as all other dependencies. Same thing goes for the Golang version we use to build our binaries - it's currently pinned to an arbitrary version in the GHA workflow.

My proposal would be to use renovate for this. It's already used elsewhere in otel - notably the collector contrib repo uses it for [dependency management](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/20907), and it is capable of updating versions embedded in arbitrary files.

I successfully achieved this in my fork. With the [following configuration](https://github.com/swiatekm/opentelemetry-operator/blob/main/renovate.json):

- golangci-lint update: https://github.com/swiatekm/opentelemetry-operator/pull/68
- CI Go version update: https://github.com/swiatekm/opentelemetry-operator/pull/54

jaronoff97:
I would love to add in renovate here and it would be great to use renovate for more than just this too. Given the collector uses it for dependency management, i think we should too.

jaronoff97:
@secustor if you could help out that would be great :D

swiatekm:
The situation with versions.txt is a bit more complicated, because that file is a manifest of the defaults used by the operator, and isn't always in sync with the latest instrumentation images. I *would* like to automate instrumentation library upgrades in the images themselves, and for some of them this is already true, but I'm not fully comfortable doing that without having SDK maintainers as code owners for them.

secustor:
If auto merging is not enabled which is the default, you would still have to merge the auto instrumentation PRs manually.

swiatekm:
Consensus from the SIG meeting on 05.12.2024 is that we should do this.

PR #3527, opened 2024-12-09
Update tools using renovate

swiatekm:
**Description:**

Add a renovate configuration, enabling a single manager to update tool versions in the Makefile. The updates are enabled by adding a comment above the version declaration in the Makefile, specifying where to look for the dependency and how to update it.

I've ignored operator-sdk for now, because I'm not sure what the upgrade strategy should be there.

#3450 

**Testing:**

I have this configuration active in my [fork](https://github.com/swiatekm/opentelemetry-operator/blob/main/renovate.json). Here's an example of a tool update PR: https://github.com/swiatekm/opentelemetry-operator/pull/71.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -491,10 +491,15 @@ CHLOGGEN ?= $(LOCALBIN)/chloggen
 GOLANGCI_LINT ?= $(LOCALBIN)/golangci-lint
 CHAINSAW ?= $(LOCALBIN)/chainsaw
 
+# renovate: datasource=go depName=sigs.k8s.io/kustomize/kustomize/v5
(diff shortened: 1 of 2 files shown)


frzifus (approved):
> I've ignored operator-sdk for now, because I'm not sure what the upgrade strategy should be there.

Is there an option to only update minior versions automatically?

swiatekm:
> > I've ignored operator-sdk for now, because I'm not sure what the upgrade strategy should be there.
> 
> Is there an option to only update minior versions automatically?

Yeah, you can add a rule to disable specific update types for a package.

jaronoff97 (approved):
perfect. This is so cool!

Issue #3513, opened 2024-12-04
Environment variable expansion in `spec.config.service.telemetry.metrics.address` doesn't work


douglascamata:
### Component(s)

collector

## Description

Looks like the shell environment variable expansion breaks config validation when used in  `spec.config.service.telemetry.metrics.address`.

It doesn't break the config validation when used in other places, like `spec.config.receivers.otlp.protocols.grpc.endpoint`.

## Steps to Reproduce

1. Install cert manager operator (optional)
2. Install OTEL operator v0.114.0
3. Try to deploy a collector using the manifest below:

```yaml
apiVersion: opentelemetry.io/v1beta1
kind: OpenTelemetryCollector
metadata:
  name: simplest
spec:
  config:
    receivers:
      otlp:
        protocols:
          grpc:
            endpoint: ${env:POD_IP}:4317
          http:
            endpoint: ${env:POD_IP}:4318
    processors:
      memory_limiter:
        check_interval: 1s
        limit_percentage: 75
        spike_limit_percentage: 15
      batch:
        send_batch_size: 10000
        timeout: 10s
    exporters:
      debug: {}
    service:
      telemetry:
        metrics:
          address: ${env:POD_IP}:8888
      pipelines:
        traces:
          receivers: [otlp]
          processors: [memory_limiter, batch]
          exporters: [debug]
```

## Expected Result

The Collector is created.

## Actual Result

It returns the error `Error from server (Forbidden): error when creating "collector_crd.yaml": admission webhook "mopentelemetrycollectorbeta.kb.io" denied the request: address ${env:POD_IP}:8888: too many colons in address`

### Kubernetes Version

1.31.2

### Operator version

0.114.0

### Collector version

0.114.0

## Environment
OS: (e.g., "Ubuntu 20.04")
Compiler(if manually compiled): (e.g., "go 14.2")

### Additional context

According to https://github.com/open-telemetry/opentelemetry-helm-charts/blob/831603f1351049a65550ec3211d8e3a44a0c5a2a/charts/opentelemetry-collector/UPGRADING.md#0782-to-0783, it would be a good idea to update all the examples to use `${env:POD_IP}` instead of listening on `0.0.0.0`.

PePoDev:
mine workaround

```yaml
manager:
  env:
    ENABLE_WEBHOOKS: "false"
```

frzifus:
I didnt test it yet, but did you try:
```diff
[...]

PR #3531, opened 2024-12-10
Fix webhook error when metrics service address uses env var expansion

douglascamata:
**Description:** 

Use a naive port parsing in the operator's webhook to ensure it doesn't error out when the metrics service address uses environment variable expansion for the host part.

- Resolves: #3513 

**Testing:** 

Added a unit test covering the IPv4 case.

**Documentation:** 

None.


swiatekm:
This change looks good to me (minus the gosec warning), but it doesn't fully address #3513. What we should do in addition is emit a warning when failing to parse the port instead of returning an error.

douglascamata:
@swiatekm will add the warning in a few mins. Thanks!

internal/manifests/collector/service.go
```diff
@@ -84,10 +83,7 @@ func MonitoringService(params manifests.Params) (*corev1.Service, error) {
 		return nil, err
 	}
 
-	_, metricsPort, err := params.OtelCol.Spec.Config.Service.MetricsEndpoint()
-	if err != nil {
-		return nil, err
-	}
+	_, metricsPort := params.OtelCol.Spec.Config.Service.MetricsEndpoint(params.Log)
```
swiatekm, line 86: Looking at this code again, I'm not wholly sure what the right thing to do here is. I thought the webhook shouldn't reject Collector CRs just because it can't parse the port, but this is the monitoring Service which can be enabled or disabled via a CR attribute. So, to some extent, the user should *want* to get an error here if this won't actually work.

@pavolloffay @jaronoff97 @frzifus what do you think we should do here? The options are

1. Do what we do right now, return an error, and just make sure we handle the case where the host comes from an env variable gracefully, like what this PR originally did.
2. Emit a warning, and simply do not create the monitoring Service if we can't get the port, but otherwise continue with manifest generation.

douglascamata: FYI, trying to add IPv6 support to this makes it even more complicated. I gave up on making my patch IPv6 compatible pretty quickly. 😅

yuriolisa: I'd go with the first option suggested by @swiatekm, but also add a doc for the users clarifying how the monitoring Service configuration is handled.

douglascamata: I believe the simplest way forward is to log a warning when we cannot parse the host-port pair in any way to let people know that service creation will be skipped because of this and create the collector.

A slightly better, and yet still simple, approach could be:

- if env vars are detected through the presence of `${}`, skip service creation, warn user about it, and create the Collector anyway.
- if no env vars are detected it means we should 100% be able to parse the host and port, apply defaults, and creator the Collector.

WDYT, @yuriolisa and @swiatekm?

douglascamata: Just a small note: the port could also be an env var, which is totally valid Collector configuration and thus we should still accept it.

swiatekm: @douglascamata if the port value is an env var reference, then we can't know the value, and we can't create a monitoring Service for it. In that case, I think it's correct for us to return an error. If the user doesn't want that, they can disable the Service creation in the Collector CR.

yuriolisa: Due to that, I suggested following the first option mentioned above.

douglascamata: @swiatekm are you sure Service creation can be disabled by the user in the Collector CR? I couldn't find anything related there. 🤔

swiatekm: Yes, it can be controlled via the `observability.metrics.enableMetrics` attribute: https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/api.md#opentelemetrycollectorspecobservabilitymetrics.

douglascamata: Ah, that's why I didn't find it: the name of this configuration option is not closely related to what its description says it does.

Thanks for the reference. I'm mentioning this in the README. 👍

douglascamata: I mean, it says "enableMetrics" and disabling it in fact prevents the creation of Pod/Service Monitor and Service objects. But the Collector's metrics are still being exported.

douglascamata:
@swiatekm @yuriolisa: I just updated the PR with the changes you suggested:

- If we fail to parse the port (due to invalid port or env var usage), an error is returned.
- Env var for the host part is still accepted.
- Documented the points above in the README, close to where there was already some explanation regarding how the Operator examines the Collector's configuration.

On top of this, I could modify the code to support IPv6 and added a bunch of tests to ensure we don't have regressions. 


Thank you for your review. 🙇

README.md
```diff
@@ -72,12 +72,14 @@ This will create an OpenTelemetry Collector instance named `simplest`, exposing
...
+The Operator does examine the configuration file for a few purposes:
+
+- To discover configured receivers, services and their ports. If it finds receivers with ports, it creates a pair of kubernetes services, one headless, exposing those ports within the cluster. If the port is using environment variable expansion or cannot be parsed, an error will be returned. To ignore service creation set `observability.metrics.enableMetric` to `false`. You can create the Service objects yourself if you need them. The headless service contains a `service.beta.openshift.io/serving-cert-secret-name` annotation that will cause OpenShift to create a secret containing a certificate and key. This secret can be mounted as a volume and the certificate and key used in those receivers' TLS configurations.
```
yuriolisa, line 79: I believe you mixed the idea of enabling the Monitoring Service and how the Operator validates the configuration file's contents. Please try to clarify each of the topics mentioned in the doc.

apis/v1beta1/config.go
```diff
@@ -427,37 +427,55 @@ type Service struct {
...
+	defaultServicePort int32 = 8888
+	defaultServiceHost       = "0.0.0.0"
+)
+
+// MetricsEndpoint attempts gets the host and port number from the host address without doing any validation regarding the
+// address itself.
+// It works even before env var expansion happens, when a simple `net.SplitHostPort` would fail because of  the extra colon
+// from the env var, i.e. the address looks like "${env:POD_IP}:4317", "${env:POD_IP}", or "${POD_IP}".
+// In cases which the port itself is a variable, i.e. "${env:POD_IP}:${env:PORT}", this returns an error.
+// It should work with IPv4 and IPv6 addresses.
```
yuriolisa, line 440: I would put the support for IPv6 in another PR.

yuriolisa (changes requested):

README.md
douglascamata: Why mixed, could you explain a bit better please? There are two separate points there:

The first list item, which talks about the receivers and theirs ports stays almost the same. The only thing I modified there is that `configured receivers and their ports` became `configured receivers, services and their ports`, which I forgot to fix. I will revert back to the original text.

The second list item is where the explanation regarding the metrics service address is explained.

douglascamata: With the changes in [62e7f1e](https://github.com/open-telemetry/opentelemetry-operator/pull/3531/commits/62e7f1e2151e095f855831e514ec69314fce3f61), the 1st bullet point there has the exact same text as before.

apis/v1beta1/config.go
douglascamata: Will separate it.

douglascamata: Well, I don't any special logic for IPv6 support. The logic I use "just works" for it out of the box. 

Do you think it's still worth to remove the extra IPv6 tests and mention in the function's doc, @yuriolisa?

yuriolisa: You're right!! Please leave it as it is.

README.md
yuriolisa: My point is because this setence `.. To ignore service creation set observability.metrics.enableMetric to false.` is placed in the same bullet point of the description how the operator handles the receivers and their ports.

douglascamata: Ahhh, my bad. Another bad leftover there. I'll get rid of it.

douglascamata: Should be good now after [a4ef4f0](https://github.com/open-telemetry/opentelemetry-operator/pull/3531/commits/a4ef4f05a4b967f605ea056c3ebcd878c6c3610a)

.chloggen/fix-metrics-service-address-env-var.yaml
```diff
@@ -0,0 +1,18 @@
...
+note: Fix the admission webhook to when metrics service address host uses env var expansion
+
+# One or more tracking issues related to the change
+issues: [3513]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  This should allow the metrics service address to have the host portion expanded from an environment variable,
```
yuriolisa, line 17: Please inform that you enabled the IPv6.

douglascamata: @yuriolisa I didn’t necessarily enable ipv6. I only refactored that ‘Service.MetricsEndpoint’ method in a way that it also works with ipv6 addresses. I do not know if the project has everything else that is needed for full ipv6 support. Do you know about this?

swiatekm: I don't think there's a need to specifically call out IPv6. We only care about the port to begin with.

README.md
```diff
@@ -72,12 +72,14 @@ This will create an OpenTelemetry Collector instance named `simplest`, exposing
 
 The `config` node holds the `YAML` that should be passed down as-is to the underlying OpenTelemetry Collector instances. Refer to the [OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector) documentation for a reference of the possible entries.
 
-> 🚨 **NOTE:** At this point, the Operator does _not_ validate the contents of the configuration file: if the configuration is invalid, the instance will still be created but the underlying OpenTelemetry Collector might crash.
```
jaronoff97, line 75: this is still mostly true, why are we removing it?

jaronoff97: (right now we validate the base structure, but not much more)

apis/v1beta1/config.go
```diff
@@ -427,37 +427,55 @@ type Service struct {
...
-		// telemetry isn't set, use the default
-		return defaultAddr, 8888, nil
+const (
+	defaultServicePort int32 = 8888
+	defaultServiceHost       = "0.0.0.0"
+)
+
+// MetricsEndpoint attempts gets the host and port number from the host address without doing any validation regarding the
+// address itself.
+// It works even before env var expansion happens, when a simple `net.SplitHostPort` would fail because of  the extra colon
```
jaronoff97, line 437:
```suggestion
// It works even before env var expansion happens, when a simple `net.SplitHostPort` would fail because of the extra colon
```

apis/v1beta1/config.go
```diff
@@ -427,37 +427,55 @@ type Service struct {
...
+// from the env var, i.e. the address looks like "${env:POD_IP}:4317", "${env:POD_IP}", or "${POD_IP}".
+// In cases which the port itself is a variable, i.e. "${env:POD_IP}:${env:PORT}", this returns an error.
```
jaronoff97, line 439: it may be worth commenting why we need the port (i.e. to generate the service and mappings)

apis/v1beta1/config.go
```diff
@@ -427,37 +427,55 @@ type Service struct {
...
+	if telemetry == nil || telemetry.Metrics.Address == "" {
+		return defaultServiceHost, defaultServicePort, nil
 	}
-	host, port, netErr := net.SplitHostPort(s.GetTelemetry().Metrics.Address)
-	if netErr != nil && strings.Contains(netErr.Error(), "missing port in address") {
-		return defaultAddr, 8888, nil
-	} else if netErr != nil {
-		return "", 0, netErr
+
+	isPortEnvVar := regexp.MustCompile(`:\${[env:]?.*}$`).MatchString(telemetry.Metrics.Address)
```
jaronoff97, line 447: suggestion: can we make the regex string a const? and we should add a comment explaining how this works/matches.

apis/v1beta1/config.go
```diff
@@ -427,37 +427,55 @@ type Service struct {
...
+		errMsg := fmt.Sprintf("couldn't determine metrics port from configuration: %s",
+			telemetry.Metrics.Address)
+		logger.Info(errMsg)
+		return "", 0, fmt.Errorf(errMsg)
 	}
-	i64, err := strconv.ParseInt(port, 10, 32)
-	if err != nil {
-		return "", 0, err
+
+	explicitPortMatches := regexp.MustCompile(`:(\d+$)`).FindStringSubmatch(telemetry.Metrics.Address)
```
jaronoff97, line 455: should we include a starting terminator to ensure we're not matching something later? i.e. `^:(\d+$)`

jaronoff97 (approved):
This all makes sense to me and I appreciate the thought put in here :) I agree with Mikolaj's original comment (and thus this implementation).

README.md
douglascamata: Well, we are validating that ports are integers (since before the changes introduced by this PR). This note is misleading. Only a few lines below it we say that there's some validation. This might leave users confused about which statement is true.

apis/v1beta1/config.go
douglascamata: @jaronoff97 we cannot do this. In regexes, `^` means beginning of the string. So `^:(\d+$)` would never match on strings that do not start with `:`.

See https://go.dev/play/p/Q9iIJqotcGI

douglascamata: I'm adding an explanation for this and making the regex string a const too, btw.

jaronoff97: oh this was due to my misreading of what this was doing, i thought you were matching here for the case `:8000` and the next line was for when you have the start as well... i.e. disregard this suggestion :)

douglascamata:
@jaronoff97: I took care of your feedback (plus another small pending change) in a batch on a6744fd

yuriolisa (approved):

PR #3545, opened 2024-12-13
[chore] bump 1.x java version

TylerHelmuth:
**Description:** 
Bumps the 1.x java version to the 1.33.6 release.

Also adds a second tagging scheme for tagging using the [major version](https://github.com/docker/metadata-action?tab=readme-ov-file#typesemver).  This will allow us to keep a `1` or `2` tag on the latest major version, so that users can also stay on the latest of the major version they use.


.github/workflows/publish-autoinstrumentation-java.yaml
```diff
@@ -37,6 +37,7 @@ jobs:
             ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-java
           tags: |
             type=match,pattern=v(.*),group=1,value=v${{ env.VERSION }}
+            type=semver,pattern={{major}},value=v${{ env.VERSION }}
```
swiatekm, line 40: Shouldn't this be
```suggestion
            type=semver,pattern=v{{major}},value=v${{ env.VERSION }}
```
? Otherwise we'd get just `1`, and I think we want `v1`.

swiatekm: Or maybe it should be `1.x`?

TylerHelmuth: Ya it can be whatever we want

TylerHelmuth: since `v1` is defined in the readme's example I'd choose that over `1.x`.

pavolloffay (approved):

.github/workflows/publish-autoinstrumentation-java.yaml
pavolloffay: https://github.com/open-telemetry/opentelemetry-operator/pkgs/container/opentelemetry-operator%2Fautoinstrumentation-java

pavolloffay: at the moment we don't prefix with `v`

swiatekm (approved):

PR #3564, opened 2024-12-20
Fix flaky controller test for versioned ConfigMaps

swiatekm:
#3560 wasn't enough. I reverted those changes, as they didn't really help and just made the code more complex for no benefit.

Kubernetes creation timestamps only have a seconds-level precision. The ConfigMap versioning feature for collectors deleted the oldest ConfigMaps. For the test to work correctly, we need to wait at least 1 second between CR changes.

### DIFF
--- a/controllers/reconcile_test.go
+++ b/controllers/reconcile_test.go
@@ -858,6 +858,9 @@ func TestOpenTelemetryCollectorReconciler_VersionedConfigMaps(t *testing.T) {
 			Namespace: metav1.NamespaceDefault,
 		},
 		Spec: v1beta1.OpenTelemetryCollectorSpec{
+			OpenTelemetryCommonFields: v1beta1.OpenTelemetryCommonFields{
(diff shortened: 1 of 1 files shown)


controllers/reconcile_test.go
```diff
@@ -858,6 +858,9 @@ func TestOpenTelemetryCollectorReconciler_VersionedConfigMaps(t *testing.T) {
 			Namespace: metav1.NamespaceDefault,
 		},
 		Spec: v1beta1.OpenTelemetryCollectorSpec{
+			OpenTelemetryCommonFields: v1beta1.OpenTelemetryCommonFields{
+				PodDisruptionBudget: &v1beta1.PodDisruptionBudgetSpec{},
+			},
```
swiatekm, line 863: This is just to avoid logs about the default being set. We should really get rid of that log line imo.

jaronoff97 (approved):

PR #3576, opened 2024-12-28
[chore] Fix err variable

atoulme:
**Description:**
Fix the error check on creating the kubernetes client.


swiatekm (approved):

swiatekm:
The test failure is unrelated and should be fixed by #3577.

PR #3582, opened 2024-12-30
Fix timestamp conversion in opamp bridge

swiatekm:

Kubernetes (and Go in general) allow for signed unix timestamps representing dates before 01-01-1970. However, OpAMP only accepts unsigned timestamps. Until now, opamp bridge simply assumed the conversion could always be carried out. This change instead returns errors when either the bridge or the respective K8s resources have negative Unix timestamps.

Unblocks #3538 and building the operator with Go 1.23 by addressing gosec errors.

### DIFF
--- a/cmd/operator-opamp-bridge/agent/agent.go
+++ b/cmd/operator-opamp-bridge/agent/agent.go
@@ -17,6 +17,7 @@ package agent
 import (
 	"bytes"
 	"context"
+	"errors"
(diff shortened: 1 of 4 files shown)


swiatekm:
I didn't include a changelog because I doubt this really affects anyone, but I can be persuaded otherwise.

jaronoff97 (approved):

PR #3589, opened 2025-01-02
Include Go version, os and arch in golangci-lint cache key

swiatekm:
Just the `go.sum` value isn't enough.


swiatekm:
Actually, looking at the logic in https://github.com/golangci/golangci-lint-action/blob/master/src/cache.ts, it seems like the Go version shouldn't be necessary. Closing for now.

PR #3596, opened 2025-01-08
fix(deps): use runtime.NewScheme directly instead of kubectl/pkg/scheme

zchee:

Use the `runtime.NewScheme` function directly instead of the `kubectl/pkg/scheme` alias.


iblancasa (approved):

swiatekm (approved):
Thank you for your contribution!

PR #3615, opened 2025-01-15
feat: add auto-instrumentation support for Deno

### LINKED ISSUE #3613: Add auto-instrumentation support for Deno
### Component(s)

auto-instrumentation

[Deno](https://deno.com), a modern JavaScript and TypeScript runtime, version 2.1.5 and later, ships with OpenTelemetry support out of the box enabled by setting `OTEL_DENO=true`. Setting this env var will enable auto-instrumentation of various components of Deno (`Deno.serve` http server, `fetch` http client, and more), OTLP exporting for metrics, logs, and traces, and automatic exporting of telemetry created through the JS `@opentelemetry/api` package from NPM: https://docs.deno.com/runtime/fundamentals/open_telemetry/

Right now it is possible to use OpenTelemetry operator with Deno, by using `instrumentation.opentelemetry.io/inject-sdk: "true"` and manually setting `OTEL_DENO=1` as an env var. This is however not as nice as some other auto-instrumentation, where one just has to set, for example, `instrumentation.opentelemetry.io/inject-nginx: "true"`.

Add support for `instrumentation.opentelemetry.io/inject-deno: "true"`, which would be behaviourally equivalent to `instrumentation.opentelemetry.io/inject-sdk: "true"` and setting `OTEL_DENO=true` as an env var.

Unlike most (all) other auto-instrumentation, a separate container (image) with auto-instrumentation payload is unneeded, because Deno comes with auto-instrumentation out of the box.

_No response_

### Additional context

Support for OTEL is currently unstable in Deno, and requires that the user pass `--unstable-otel` when starting to enable. We expect this feature to be stabilized within the next three months.

I will open a PR with the code needed to support this, disabled by default (ie one needs to pass `--enable-deno-instrumentation`).

lucacasonato:
**Description:**

This commit adds support for auto-instrumentation in Deno, when using Deno 2.1.5 or higher and Deno was invoked with `--unstable-otel`.

Deno support is disabled by default and must be explicitly enabled with `--enable-deno-instrumentation`.

**Link to tracking Issue(s):** resolves #3613
 
**Testing:**

- Added unit tests for injection + detection
- Added e2e test

**Documentation:** Added to readme, will open a PR to the website if this is landed.


lucacasonato:
Auto-instrumenting Deno is simple enough that probably a docs change suffices - no need for a new annotation. See https://github.com/open-telemetry/opentelemetry-operator/issues/3613#issuecomment-2598005507

PR #3623, opened 2025-01-19
[chore] delete unused file

atoulme:
This file is not used and can be removed.


swiatekm (approved):
Good catch, this is a leftover from a time where the target allocator had its own linter config.

PR #3625, opened 2025-01-19
[chore] remove init function from prehook package

atoulme:
**Description:**
Remove the init function by initializing the map of providers with the initial provider.

Unexport the provider struct as it is not used outside the package.

Add tests to check the behavior of the Register function.


swiatekm (commented):
We may as well get rid of `Register`, it's not called anywhere outside of the init function.

jaronoff97 (approved):

atoulme:
OK, I have removed Register as well, thanks!

swiatekm (approved):

PR #3626, opened 2025-01-20
[chore] do not initialize empty slices

atoulme:
Apply go best practice to avoid assigning an empty slice.


swiatekm (approved):

swiatekm:
It's a bit annoying there's no linter that can enforce this best practice. Best I found was a recommendation to use ruleguard.

PR #3631, opened 2025-01-20
[chore] use errors.New unless there is interpolation

atoulme:


swiatekm (approved):

swiatekm:
@atoulme did you find this using a linter? If so, we should enable it in our config.

PR #3632, opened 2025-01-20
[chore] move all target-allocator code to internal

atoulme:
**Description:**
This is a proposed change and a test to see if we can move all the code under target allocator to internal, meaning its API is no longer public and exported.


swiatekm:
Looks good to me, broadly speaking. I would even consider moving all of it out of `cmd/` and into `internal/targetallocator`.

jaronoff97 (approved):

iblancasa (approved):

swiatekm (approved):

PR #3633, opened 2025-01-20
[chore] move controllers under internal

atoulme:
**Description:**
Move controllers under internal so their API is not exported.


swiatekm:
This placement of `controllers` is a kubebuilder standard. I'm not sure how much that matters to us at this point, but it would certainly make the migration to kubebuilder v4 more difficult. WDYT @open-telemetry/operator-approvers ?

atoulme:
I don't see an issue for moving to kubebuilder 4, is it ok to open one? Is anyone working on it?

frzifus:
I dont see a huge downside of exposing it or benefit of moving it into internal here. Thats why I would prefer to  just follow whatever kubebuilder standard is.

atoulme:
The problem is that you're currently exposing incomplete APIs. It is possible other downstream projects may start to depend on it and bind you to a specific go API, which doesn't seem to be the intent of this project. See 
https://pkg.go.dev/github.com/open-telemetry/opentelemetry-operator@v0.116.0/controllers

swiatekm:
It used to be at top level in v2: https://book-v2.book.kubebuilder.io/cronjob-tutorial/controller-implementation.

I guess if it's under internal in subsequent versions, then moving it right now won't really make the migration any more difficult.

atoulme:
Rebased.

atoulme:
Rebased.

swiatekm (approved):
I'm in favor of making this change, but I'd like more opinions before merging. @open-telemetry/operator-approvers please review if you care.

iblancasa (approved):

jaronoff97 (approved):

PR #3634, opened 2025-01-20
[chore] delete unused constant

atoulme:


swiatekm:
Not sure how I feel about this. These constants are not used in the code, but they represent values the CR attribute can take, so I think it's less confusing to have them defined. @open-telemetry/operator-approvers wdyt?

frzifus (commented):
In case we remove this variables, I would prefer to move the description of those  options into the CRD.

atoulme:
Closing for now.

PR #3645, opened 2025-01-20
[chore] simplify allocation Option and registry

atoulme:
Rename the AllocationOption type to Option, and simplify the handling of strategies and the registry in the allocation package.


swiatekm (approved):
LGTM, thanks for doing this cleanup. The current messy state was the remainder of a refactor I forgot to clean up after.

PR #3668, opened 2025-01-25
[chore] simplify code by using a utility method

atoulme:


pkg/instrumentation/sdk.go
```diff
@@ -661,6 +648,20 @@ func getIndexOfEnv(envs []corev1.EnvVar, name string) int {
...
+		keys[e.Name] = struct{}{}
+	}
+	for _, envVar := range newEnvVars {
+		if _, ok := keys[envVar.Name]; !ok {
+			envs = append(envs, envVar)
+			keys[envVar.Name] = struct{}{}
+		}
+	}
+	return envs
+}
```
swiatekm, line 663: Could you add a few simple tests for this function?

atoulme: Good idea, looks like I have a regression too according to CI. Will work on it more.

pavolloffay (approved):

Issue #3575, opened 2024-12-27
[instrumentation] Support setting image pull policy for agent images


atoulme:
### Component(s)

auto-instrumentation

### Is your feature request related to a problem? Please describe.

The current setup relies on the default behavior of kubernetes imagePullPolicy as it is not set explicitly.

### Describe the solution you'd like

Add a field to the instrumentation CRDs and config flags to support setting image pull policies globally or per instrumentation.

swiatekm:
:+1: I'm in favor of adding this.

yuriolisa:
Would the idea be like the following?

```yaml
kind: Instrumentation
metadata:
  name: my-instrumentation
spec:
  exporter:
    endpoint: http://otel-collector:4317
  propagators:
    - tracecontext
    - baggage
    - b3
  sampler:
    type: parentbased_traceidratio
    argument: "0.25"
  imagePullPolicy: Always
  python:
    imagePullPolicy: IfNotPresent
    env:
      # Required if endpoint is set to 4317.
      # Python autoinstrumentation uses http/proto by default
      # so data must be sent to 4318 instead of 4317.
      - name: OTEL_EXPORTER_OTLP_ENDPOINT
        value: http://otel-collector:4318
```

If so, would you have interest to send a PR or can I work on this one?

atoulme:
Yes, you can do it like this.

PR #3675, opened 2025-01-27
Add support to set image pull policy for agent images

yuriolisa:

Add support to set image pull policy for agent images

- Resolves: #3575 


swiatekm:
Can't we put this at the top level, instead of in each instrumentation individually? I doubt anyone would need to set a different pull policy in different instrumentations - and if they do, they can create a separate CR.

yuriolisa:
> Can't we put this at the top level, instead of in each instrumentation individually? I doubt anyone would need to set a different pull policy in different instrumentations - and if they do, they can create a separate CR.

Sure thing. I created both ways because the issue was requesting that, and I thought would be beneficial to the end-user.

yuriolisa:
> * [instrumentation] Support setting image pull policy for agent images #3575

@atoulme, do you have a use case for putting the pullPolicy for each instrumentation individually?

yuriolisa:
@swiatekm, I've removed the specific pullPolicy setting.

pkg/instrumentation/helper.go
```diff
@@ -171,3 +171,13 @@ func setContainersFromAnnotation(inst *instrumentationWithContainers, annotation
 	inst.Containers = append(inst.Containers, languageContainers...)
 	return nil
 }
+
+func setImagePullPolicy(instSpec corev1.PullPolicy) corev1.PullPolicy {
```
swiatekm, line 175: Is this necessary? Can't we just leave this empty if the user does?

iblancasa (approved):

pkg/instrumentation/helper.go
yuriolisa: I will have a look whether we could do it.

yuriolisa:
@swiatekm, I believe we're to get this merged.

swiatekm (approved):

yuriolisa:
@swiatekm, @jaronoff97, could you please check it out to see if we are good to get this merged?

PR #3678, opened 2025-01-28
[chore] update alpine to 3.21

atoulme:
This is not immediately needed. This PR updates the docker image Alpine to 3.21 to avoid running into a situation where the certs downloaded from Alpine end up out of date.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Get CA certificates from alpine package repo
-FROM alpine:3.20 as certificates
+FROM alpine:3.21 as certificates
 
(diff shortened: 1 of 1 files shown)


swiatekm:
I manually triggered dependabot and it created an update PR here: https://github.com/open-telemetry/opentelemetry-operator/pull/3680. Not sure why it didn't trigger on its own, we have it configured to check weekly. There used to be an issue with parsing kind configuration, but it looks like it's fixed now given that https://github.com/open-telemetry/opentelemetry-operator/pull/3679 was created successfully.

In any case, I think we can close this.

PR #3700, opened 2025-02-06
[docs] API reference copyedits

chalin:
- Thanks again for solving https://github.com/open-telemetry/opentelemetry.io/issues/6200 by refactoring the API docs.
- This PRs suggests a few copyedits to the API pages:
  - Avoids the use of "here" as link text.
  - Adds "API reference" back into the (new) page title.

/cc @open-telemetry/docs-maintainers

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -1,4 +1,4 @@
-# CRDs
+# API reference for CRDs
 
 - [Instrumentation](instrumentations.md)
 - [OpAMPBridge](opampbridges.md)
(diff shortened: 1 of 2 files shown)


docs/api/README.md
```diff
@@ -1,6 +1,6 @@
-# CRDs
+# API reference for CRDs
```
chalin, line 1: Should this be simply "API reference" as it was before?

iblancasa (approved):

docs/api/README.md
swiatekm: I think what you did here is better.

swiatekm (approved):

jaronoff97 (approved):

Issue #3726, opened 2025-02-19
ServiceMonitor/PodMonitor referenced even when Prometheus CRDs are not installed


a-asaad:
### Component(s)

target allocator

## Description

The controller attempts to watch Service and Pod Monitor resources even if Prometheus CRDs are not defined.

## Steps to Reproduce

Install Operator using the static manifests
```shell
kubectl apply -f https://github.com/open-telemetry/opentelemetry-operator/releases/download/v0.118.0/opentelemetry-operator.yaml
```

## Expected Result

Controller pod should start up successfully

## Actual Result

The controller pod enters a loop where it crashes every few minutes.

The logs are spammed with the following errors:
```json
{"level":"ERROR","timestamp":"2025-02-18T23:48:13.911351716Z","logger":"controller-runtime.source.EventHandler","message":"kind must be registered to the Scheme","error":"no kind is registered for the type v1.ServiceMonitor in scheme \"pkg/runtime/scheme.go:100\"","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/source.(*Kind[...]).Start.func1.1\n\t/opt/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.3/pkg/internal/source/kind.go:74\nk8s.io/apimachinery/pkg/util/wait.loopConditionUntilContext.func2\n\t/opt/go/pkg/mod/k8s.io/apimachinery@v0.31.3/pkg/util/wait/loop.go:87\nk8s.io/apimachinery/pkg/util/wait.loopConditionUntilContext\n\t/opt/go/pkg/mod/k8s.io/apimachinery@v0.31.3/pkg/util/wait/loop.go:88\nk8s.io/apimachinery/pkg/util/wait.PollUntilContextCancel\n\t/opt/go/pkg/mod/k8s.io/apimachinery@v0.31.3/pkg/util/wait/poll.go:33\nsigs.k8s.io/controller-runtime/pkg/internal/source.(*Kind[...]).Start.func1\n\t/opt/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.3/pkg/internal/source/kind.go:64"}
[...]

PR #3728, opened 2025-02-20
fix(target-allocator): check for Prometheus CRDs before watching Service/Pod Monitors

a-asaad:

Add an additional check for installed Prometheus CRDs before owning Service and Pod Monitor resources in the `TargetAllocator` controller.

- Resolves: #3726 

### DIFF
--- a/.chloggen/fix-3726.yaml
+++ b/.chloggen/fix-3726.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'bug_fix'
+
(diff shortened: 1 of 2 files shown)


a-asaad:
Didn't find an obvious spot to add a unit test, but I can work on one if you can point me in the right direction.

iblancasa:
> Didn't find an obvious spot to add a unit test, but I can work on one if you can point me in the right direction.

I think we actually need to implement an E2E test to run stuff without the Prometheus CRDs installed. I think we have it but... seems we are missing something.

iblancasa (approved):

pavolloffay (approved):

PR #3731, opened 2025-02-21
LD_PRELOAD-based injection

### LINKED ISSUE #2375: Use LD_PRELOAD for setting environment variables e.g. JAVA_TOOL_OPTIONS / OTEL_
### Component(s)

_No response_

At the moment the operator sets the env variables directly on the pod spec. The downside is if user defines the same variables in docker image or in config map. The operator does not know anything about those env vars. The solution might be to use LD_PRELOAD to set and override env vars. 

The same approach is used by Lumigo operator and probably other vendors. https://github.com/lumigo-io/lumigo-kubernetes-operator/blob/88a3f78e5a32b965210659db11ef97f2a0c6f6c6/controller/src/mutation/mutate.go#L50

This would also resolve https://github.com/open-telemetry/opentelemetry-operator/issues/1814 

Use LD_PRELOAD to set env vars for the instrumentation CR.

_No response_

### Additional context

_No response_

mmanciop:
**Description:**

This PR is a (very) early experiment of using LD_PRELOAD-based injection in the autoinstrumentation performed by the operator.

It can only inject Java for now, as a proof of concept. I am positive that all other instrumentations where there are no dependencies to libc flavors (i.e., Python and .NET) can be supported with little effort, especially if we agree on a way to reuse instrumentation images (see open questions below).

## Open questions

There are two main issues to deal with before proceeding further:

1. How can the LD_PRELOAD-based injector reuse configurations from other instrumentations, e.g., instrumentation image versions or Java Extensions?

2. How to deal with the languages (Python and .NET specifically) that cannot currently be injected reliably due to dependencies in the libc version used within the container, which cannot be determined at webhook time without user-provided configurations (which is a terminally bad UX imo). Also, the libc flavor can change between containers in the same pod, and therefore pod-based annotations are generally technically not good enough to provide full coverage to all containers.

## Known TODOs

* Introduce C linter for injector code
* Set default of `enable-injector-instrumentation` to `false` (this feature is as "experimental" as they come)
* Establish versioning scheme of injector image
* Injector Image CI build
* More documentation

- Resolves: #2375

**Testing:** Unit + e2e tests, but many more of the latter are needed

**Documentation:** Extended README.md to an extent comparable to other
  autoinstrumentation


mmanciop:
Hi, an update on my end, as I have not shown process since around KubeCon. We are actively working in Dash0 to make production-ready the Zig-based injector, covering also environment use-cases that are common in Python (which does not use getenv at runtime, but rather parses once the __environment on startup). When we sort that out, we'll swap the injector on this branch.

Issue #2884, opened 2024-04-21
OpenTelemetryCollector should have specific serviceAnnotations, podAnnotations, serviceLabels and podLabels settings


diranged:
### Component(s)

collector

### Is your feature request related to a problem? Please describe.

Today the operator _copies_ the `annotations` and `labels` from the `OpenTelemetryCollector` objects to all of the resources it creates. This causes problems with tools like ArgoCD which use Annotations for resource tracking, so then the [`--annotations-filter`](https://github.com/open-telemetry/opentelemetry-operator/blob/main/main.go#L158C265-L158C298) flag was created.  This is an unideal hack though because it's an all or nothing thing... and it means that we propagate labels and annotations to lots of resources where they don't make sense. For example, the [`service.kubernetes.io/topology-mode`](https://kubernetes.io/docs/concepts/services-networking/topology-aware-routing/) label has to be added to all the resources (clutster roles, pods, deployments, etc) when we really only want it on the service.

### Describe the solution you'd like

I don't think that the Operator should automatically copy these Annotations or Labels at all ... and instead they should be defined as parameters.

taer:
I see https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/api.md#opentelemetrycollectorspec-1 has an option for podAnnotations, but not podLabels

Rohlik:
We would also love to see support for `podLabels` for our use case, where we set a `team` label based on the owner's team and use that later on to route alerts.

shuheiktgw:
This totally makes sense to me. Let me see if I can help with this issue 🙂

shuheiktgw:
I've created a PR to support pod labels. 
- https://github.com/open-telemetry/opentelemetry-operator/pull/3733

Once it's been merged, I'm going to check service[Annotations|Labels] as well 🙂

shuheiktgw:
@iblancasa Let me move [our discussion](https://github.com/open-telemetry/opentelemetry-operator/pull/3733#issuecomment-2681200826) here! Thanks for the clarification. I feel that maintaining a list of annotations and labels that shouldn't be propagated to child resources within the operator may not be flexible enough, as some users might want to prevent the propagation of service- or environment-specific labels/annotations, while others may actually want to propagate those prohibited ones.

If you have any input, please let me know! @diranged @taer @Rohlik

iblancasa:
[...]

PR #3733, opened 2025-02-22
Add support to set additional pod labels

shuheiktgw:
I've added support to set additional pod labels like pod annotations. Thank you for your review!

- Relates to: https://github.com/open-telemetry/opentelemetry-operator/issues/2884

### DIFF
--- a/internal/manifests/targetallocator/deployment.go
+++ b/internal/manifests/targetallocator/deployment.go
@@ -15,14 +15,24 @@ import (
 // Deployment builds the deployment for the given instance.
 func Deployment(params Params) (*appsv1.Deployment, error) {
 	name := naming.TargetAllocator(params.TargetAllocator.Name)
-	labels := manifestutils.Labels(params.TargetAllocator.ObjectMeta, name, params.TargetAllocator.Spec.Image, ComponentOpenTelemetryTargetAllocator, nil)
+	labels := manifestutils.Labels(params.TargetAllocator.ObjectMeta, name, params.TargetAllocator.Spec.Image, ComponentOpenTelemetryTargetAllocator, manifestutils.WithFilterLabels(params.Config.LabelsFilter()))
 
 	configMap, err := ConfigMap(params)
 	if err != nil {
 		params.Log.Info("failed to construct target allocator config map for annotations")
 		configMap = nil
 	}
-	annotations := Annotations(params.TargetAllocator, configMap, params.Config.AnnotationsFilter())
+
+	podLabels := manifestutils.Labels(
+		params.TargetAllocator.ObjectMeta,
+		name,
+		params.TargetAllocator.Spec.Image,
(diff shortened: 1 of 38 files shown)


internal/manifests/targetallocator/deployment.go
```diff
@@ -15,14 +15,24 @@ import (
 // Deployment builds the deployment for the given instance.
 func Deployment(params Params) (*appsv1.Deployment, error) {
 	name := naming.TargetAllocator(params.TargetAllocator.Name)
-	labels := manifestutils.Labels(params.TargetAllocator.ObjectMeta, name, params.TargetAllocator.Spec.Image, ComponentOpenTelemetryTargetAllocator, nil)
+	labels := manifestutils.Labels(params.TargetAllocator.ObjectMeta, name, params.TargetAllocator.Spec.Image, ComponentOpenTelemetryTargetAllocator, manifestutils.WithFilterLabels(params.Config.LabelsFilter()))
```
shuheiktgw, line 18: I believe we need to support label filter here.

internal/manifests/targetallocator/issuer.go
```diff
@@ -14,7 +14,7 @@ import (
 // SelfSignedIssuer returns a self-signed issuer for the given instance.
 func SelfSignedIssuer(params Params) *cmv1.Issuer {
 	name := naming.SelfSignedIssuer(params.TargetAllocator.Name)
-	labels := manifestutils.Labels(params.TargetAllocator.ObjectMeta, name, params.TargetAllocator.Spec.Image, ComponentOpenTelemetryTargetAllocator, nil)
+	labels := manifestutils.Labels(params.TargetAllocator.ObjectMeta, name, params.TargetAllocator.Spec.Image, ComponentOpenTelemetryTargetAllocator)
```
shuheiktgw, line 17: https://github.com/open-telemetry/opentelemetry-operator/pull/3733/files#r1966526847

iblancasa:
I'm not sure if this PR just covers #2884 or more things. My feeling is that we should filter some of the values in the created resources, not adding a new field (that can be interesting but I don't feel is what the issue is about).

shuheiktgw:
@iblancasa Thank you for your feedback! Now that I’ve reread the issue, I realize I completely misunderstood it—this PR actually does propagate annotations and labels to child resources 😂 Let me make sure I understand correctly. Are you suggesting that instead of adding new fields to the CRD or flags to the controllers, we should create a list of annotations and labels that should not be propagated to child resources?

iblancasa:
>  Are you suggesting that instead of adding new fields to the CRD or flags to the controllers, we should create a list of annotations and labels that should not be propagated to child resources?

Yes.

shuheiktgw:
Let me move our discussion to [the original issue](https://github.com/open-telemetry/opentelemetry-operator/issues/2884#issuecomment-2684303232) 🙏

PR #3742, opened 2025-02-23
Fix the potential issue of duplicate targets

### LINKED ISSUE #3617: Targets in the same group are assigned to the same OTel
### Component(s)

target allocator

otel-allocator allocates multiple container endpoints of the same pod to different OTel. After the relabel operation, different OTel will collect the same metrics. Since I used `cumulativeToDelta`, the final metric data will be doubled.

For example, the same pod detects the following targets:

- 10.22.14.119            --> assigned to OTel-0

- 10.22.14.119:8000  --> assigned to OTel-1

- 10.22.14.119:8080  --> assigned to OTel-2

The address will be changed to 10.22.14.119:8000 after the `relabel` operation, and all the above OTel will scrape the metric of 10.22.14.119:8000, which causes metric data errors.

After the otel - allocator performs the relabel operation, it unifies and deduplicates the items. 
The selection of labels for deduplication is as follows:

- Select the label of the target where the original `__address__` completely matches the `__address__` after relabeling.

- If no such target exists, the labels of the first target will be used by default, while the metrics with the `_meta` prefix from the last target will overwrite the labels. This ensures consistency with the behavior of the OTel Prometheus receiver, for detail, see [prometheus](https://github.com/prometheus/prometheus/blob/e6cfa720fbe6280153fab13090a483dbd40bece3/scrape/target.go#L534) and [prometheus receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/0788185420201075702f23530b0dc49d26598ad0/receiver/prometheusreceiver/internal/transaction.go#L403)

The following are examples.

Image

### Additional context

_No response_

mike9421:

There may be a problem of unexpected target duplication -- for example, users rewrite the address during relabeling.

- Resolves: #3617 

add `TestDiscovery_NewItem` unit test function and adjust the relabel_test.go file to accommodate the new test


mike9421:
@swiatekm Hello, I've submitted this pull request. Could you please review it at your convenience? Thank you!

swiatekm:
This approach seems wasteful, as you're applying the relabeling step twice. It also circumvents current component responsibility boundaries, as we'd like all relabeling to happen in the prehook.

Instead, I think it'd be much simpler to defer calculating the target url. We only need the url during target allocation, at which point relabeling is guaranteed to be done. What I'd do:

1. Make the `TargetURL` attribute private and start empty.
2. Add a `GetTargetURL` method that returns the attribute value if set, otherwise checks the label and sets it.
3. Since this attribute is serialized into JSON by the web server, you might need to explicitly set a tag on it.

mike9421:
The general processing flow of  "Targets" is as follows:
[Discovery](https://github.com/open-telemetry/opentelemetry-operator/blob/9f152fb018e25eb407e535c5ad8b02f8fcca9f8b/cmd/otel-allocator/internal/target/discovery.go#L189)  -> [NewItem](https://github.com/open-telemetry/opentelemetry-operator/blob/9f152fb018e25eb407e535c5ad8b02f8fcca9f8b/cmd/otel-allocator/internal/target/discovery.go#L207) -> [SetTargets](https://github.com/open-telemetry/opentelemetry-operator/blob/9f152fb018e25eb407e535c5ad8b02f8fcca9f8b/cmd/otel-allocator/internal/allocation/allocator.go#L76) -> [prehook](https://github.com/open-telemetry/opentelemetry-operator/blob/9f152fb018e25eb407e535c5ad8b02f8fcca9f8b/cmd/otel-allocator/internal/prehook/relabel.go#L29) -> [assign](https://github.com/open-telemetry/opentelemetry-operator/blob/9f152fb018e25eb407e535c5ad8b02f8fcca9f8b/cmd/otel-allocator/internal/allocation/allocator.go#L201)

My considerations are as follows:
- If deduplication is not performed before assigning the target, it may still lead to duplicate targets being scraped. [consistent-hashing](https://github.com/open-telemetry/opentelemetry-operator/blob/9f152fb018e25eb407e535c5ad8b02f8fcca9f8b/cmd/otel-allocator/internal/allocation/consistent_hashing.go#L49) can ensure that targets are assigned to the same OTel instance by hashing the `TargetURL` after relabeling to avoid duplication, but [least-weighted](https://github.com/open-telemetry/opentelemetry-operator/blob/9f152fb018e25eb407e535c5ad8b02f8fcca9f8b/cmd/otel-allocator/internal/allocation/least_weighted.go#L24) cannot ensure assignment to the same OTel instance.
- The prehook is not currently required to be enabled, so in the above PR, all relabel operations have been moved to `NewItem`, and the `RelabeledKeep` field information is saved for [relabelConfigTargetFilter](https://github.com/open-telemetry/opentelemetry-operator/blob/9f152fb018e25eb407e535c5ad8b02f8fcca9f8b/cmd/otel-allocator/internal/prehook/relabel.go#L29).
- After deduplication, the `TargetURL` used by `consistent-hashing` could have remained unchanged. However, I later found that the targets returned by Prometheus are unordered (The targets are obtained by [allGroups](https://github.com/prometheus/prometheus/blob/3f0de72da7672313b0a53c81dc908279609799bb/discovery/manager.go#L427) traversing [map](https://github.com/prometheus/prometheus/blob/3f0de72da7672313b0a53c81dc908279609799bb/discovery/manager.go#L152)), which means the target selected after deduplication may vary each time. Therefore, I added the `TargetURLRelabeled` field for `consistent-hashing` to perform hashing. This ensures that even if different targets are selected, they will still be assigned to the same OTel instance.


> This approach seems wasteful, as you're applying the relabeling step twice. It also circumvents current component responsibility boundaries, as we'd like all relabeling to happen in the prehook.
> 
> Instead, I think it'd be much simpler to defer calculating the target url. We only need the url during target allocation, at which point relabeling is guaranteed to be done. What I'd do:
> 
> 1. Make the `TargetURL` attribute private and start empty.
> 2. Add a `GetTargetURL` method that returns the attribute value if set, otherwise checks the label and sets it.
> 3. Since this attribute is serialized into JSON by the web server, you might need to explicitly set a tag on it.

@swiatekm Modifying the `TargetURL` alone cannot solve the potential issue of `least-weighted` strategy scraping duplicate targets. Additionally, since `prometheusreceiver` will perform relabeling again, the `TargetURL` provided to OTel must be the original `TargetURL` (i.e., original `__address__`).

However, the above PR does circumvent the component responsibility boundaries, and placing the relabel operation in the prehook is indeed more reasonable. Since the deduplication operation is very similar to the target discarding operation, I think the deduplication operation can be placed in the default filter (i.e., `relabelConfigTargetFilter`) first. Afterward, we may need to consider whether multiple prehooks are necessary or if there should be a default prehook. I'd love to hear your thoughts on this.

swiatekm:
Allright, I see what the problem is. What we need to defer here is not just determining the target url, but also the hash. Is that right @mike9421?

If that's the case, could you hold off on this change for a bit. I have a WIP change that accomplishes the same thing, albeit for different reasons. It is *much* simpler than what you have here, so I'll submit it and we can see if it addresses your issue. Does that sound ok?

mike9421:
> Allright, I see what the problem is. What we need to defer here is not just determining the target url, but also the hash. Is that right @mike9421?
> 
> If that's the case, could you hold off on this change for a bit. I have a WIP change that accomplishes the same thing, albeit for different reasons. It is _much_ simpler than what you have here, so I'll submit it and we can see if it addresses your issue. Does that sound ok?

Alright, this PR will be put on hold until your WIP is completed. Which PR corresponds to the WIP? I'd like to see if it resolves the issue.

swiatekm:
https://github.com/open-telemetry/opentelemetry-operator/pull/3777 is the PR. It doesn't resolve the issue, but it will make it much simpler to do so. I think just calculating the Hash and TargetUrl forcefully in the relabel prehook should solve it.

mike9421:
>I think just calculating the Hash and TargetUrl forcefully in the relabel prehook should solve it.

Yes, a forced relabel operation is required. As I mentioned above, since the prehook is not mandatory (in the current design), the following issue may need to be considered: 

> Afterward, we may need to consider whether multiple prehooks are necessary or if there should be a default prehook. I'd love to hear your thoughts on this.

swiatekm:
> Yes, a forced relabel operation is required. As I mentioned above, since the prehook is not mandatory (in the current design), the following issue may need to be considered:

Yes, we might need to force it, or at least document that target duplication is possible when it's disabled. Right now it's enabled by default, so it wouldn't be too drastic of a change at least.

More broadly, we do want to look into doing all target relabeling in the target allocator (right now we just drop targets, but don't change the labels themselves). This is more complicated, but it would make us more compatible with Prometheus and resolve these duplication issues.

mike9421:
The updated PR enables passing the required hash value and enhances extensibility, allowing different deduplication strategies through the use of different hash calculation methods for different `FilterStrategy`. For now, I haven’t found a better approach to pass the relabeled data to `func (t *Item) Hash() ItemHash`.

cmd/otel-allocator/internal/prehook/relabel.go
```diff
@@ -48,6 +49,14 @@ func (tf *relabelConfigTargetFilter) Apply(targets []*target.Item) []*target.Ite
...
 		if keepTarget {
+			// Only if the key model.AddressLabel remains after relabeling is the value considered valid.
+			// For detail, see https://github.com/prometheus/prometheus/blob/e6cfa720fbe6280153fab13090a483dbd40bece3/scrape/target.go#L457
+			if address := lset.Get(model.AddressLabel); len(address) != 0 {
+				hash := lset.Hash()
+				tItem.HashFunc = func() uint64 {
+					return hash
+				}
+			}
 			targets[writeIndex] = tItem
```
jaronoff97, line 60: What would happen here if we created a new target item with the new label set? I think only recreating the hash may not be enough, and we may as well save the relabel... on the other hand I wonder if we would get into a scenario where we drop the target from the relabel. So if we were to save the new target we could just remove the relabel configs from the job, given we've already done the work.

jaronoff97 (commented):
I don't love needing to set a function like this, I propose another idea that may be worth testing

cmd/otel-allocator/internal/prehook/relabel.go
jaronoff97: This is the approach Mikolaj says is a bit more challenging but I think it's the more correct thing to do.

swiatekm: We did talk about doing all the target relabelling in the target allocator, which would solve this. However, that's also not trivial, and I'm not sufficiently confident in our test coverage for target relabelling to just drop this kind of change on users without the ability to turn it off.

mike9421: > What would happen here if we created a new target item with the new label set? I think only recreating the hash may not be enough, and we may as well save the relabel... on the other hand I wonder if we would get into a scenario where we drop the target from the relabel. So if we were to save the new target we could just remove the relabel configs from the job, given we've already done the work.

Currently in OTel processing, Prometheus saves the original labels (for detail, see [NewTarget](https://github.com/prometheus/prometheus/blob/e6cfa720fbe6280153fab13090a483dbd40bece3/scrape/target.go#L581)  and [scrape.go](https://github.com/prometheus/prometheus/blob/e6cfa720fbe6280153fab13090a483dbd40bece3/scrape/scrape.go#L505)), and the prometheusreceiver reads certain values from these original labels as resource attributes (for detail, see [transaction.go](https://github.com/mike9421/opentelemetry-collector-contrib/blob/92468af14dd4ccee463f370bc7103be08a1ed7e3/receiver/prometheusreceiver/internal/transaction.go#L412)). If the `FilterStrategy` creates a new item using relabeled labels, it may lead to incorrect resource information.

mike9421: I think doing all the target relabelling in the target allocator might require the prometheusreceiver to implement custom handling logic for data coming from the target allocator, or to reimplement the prometheusreceiver’s metric processing logic.

I initially considered adding `RelabeledLabels` field to the item struct, but ultimately decided to use only a hash for the following reasons:
1. There is currently only one `FilterStrategy`, so the utility of this field is limited;
2. At present, `RelabeledLabels` is only used for dropping items and deduplication;
3. Adding the field would increase memory usage;
4. The data fields of item are immutable after creation.
https://github.com/open-telemetry/opentelemetry-operator/blob/65af20a9d627b053728bf6e53666f44324584c9c/cmd/otel-allocator/internal/target/target.go#L58

Let me think further about whether there's a better way to handle this.

swiatekm: For the record, doing target relabelling in the target allocator doesn't require any changes in prometheus receiver. Since the target allocator exposes scrape config definitions to the receiver, we can simply clear the respective sections of those definitions. Then, from the perspective of the prometheus receiver, there's no target relabelling to be done.

When I last tried to implement this, it was mimicking what Prometheus does for target relabelling that was actually more difficult.

mike9421: > For the record, doing target relabelling in the target allocator doesn't require any changes in prometheus receiver. Since the target allocator exposes scrape config definitions to the receiver, we can simply clear the respective sections of those definitions. Then, from the perspective of the prometheus receiver, there's no target relabelling to be done.

Doing all the relabeling in the target allocator and removing the `relabel_configs` might cause issues when labels are dropped—for example, if a user uses the `labeldrop` action to remove information needed for resource attributes (for detail, see [transaction.go](https://github.com/mike9421/opentelemetry-collector-contrib/blob/92468af14dd4ccee463f370bc7103be08a1ed7e3/receiver/prometheusreceiver/internal/transaction.go#L412)).
But this case might be solvable by keeping labels with the `__meta` prefix (for detail, see [del label](https://github.com/prometheus/prometheus/blob/e6cfa720fbe6280153fab13090a483dbd40bece3/scrape/target.go#L534)) in the target allocator. Doing it this way should be totally fine for now. You're right.

mike9421:
@swiatekm @jaronoff97  Hi, could you please help review this PR when you have time and check if any updates are needed?

This PR:

- Updates the scrape target labels to align with Prometheus-processed labels
- Removes relabel_configs
- Adds unit tests comparing the behavior with Prometheus

It’s a bit challenging to fully decouple these unit tests from Prometheus.
Thanks in advance for your review! 🙏

eenchev:
@swiatekm @jaronoff97 
This issue is rendering Target Allocator unusable for us as well, could you please review this MR or propose a workaround in the meantime?

jaronoff97 (approved):
I _think_ this should work. Do we have any existing end to end tests that we can update or view to verify this is working as expected? Thank you for your patience :bow:

mike9421:
Hi @jaronoff97  , could you please help review this PR when you have time? Here's a summary of the changes:

- Fixed golint issues.

- Removed incorrect assert statements.

- Added end-to-end (e2e) tests.

pandoralink:
> Hi @jaronoff97 @swiatekm , could you please help review this PR when you have time? Here's a summary of the changes:
> 
> * Fixed golint issues.
> * Removed incorrect assert statements.
> * Added end-to-end (e2e) tests.

@jaronoff97 @swiatekm
This issue affects our Target Allocator usage too. Could you please review this MR or propose a workaround?

tests/e2e-targetallocator/targetallocator-consistency/02-install.yaml
```diff
@@ -0,0 +1,94 @@
...
+            - source_labels: [job_sign]
+              action: replace
+              regex: ".*"
+              replacement: single_job
+              target_label: job_sign 
+    processors:
+      # Ensure metrics are flushed to the exporter at least once
+      batch:
+        send_batch_size: 1000
+        timeout: 20s
```
swiatekm, line 82: If we want to guarantee a flush, isn't it better to set a very low timeout? Or just not use the batch processor at all.

tests/e2e-targetallocator/targetallocator-consistency/03-install.yaml
```diff
@@ -0,0 +1,148 @@
+# Metrics consistency verification Pod 
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: metrics-consistency-script
+data:
+  main.py: |
```
swiatekm, line 7: I wish this was achievable using only chainsaw's Prometheus metric parsing function. Or, if we do need the script, I'd prefer it live in a .py file in the tests directory and be imported here somehow. Python code as text inside a ConfigMap definition is highly unmaintainable.

cmd/otel-allocator/internal/prehook/relabel.go
```diff
@@ -41,14 +42,30 @@ func (tf *relabelConfigTargetFilter) Apply(targets []*target.Item) []*target.Ite
...
 		for _, cfg := range tf.relabelCfg[tItem.JobName] {
+			// These labels are typically required for correct scraping behavior and are expected to be retained after relabeling.:
+			//   - job
+			//   - __scrape_interval__
+			//   - __scrape_timeout__
+			//   - __scheme__
+			//   - __metrics_path__
+			// Prometheus adds these labels by default. Removing them via relabel_configs is considered invalid and is therefore ignored.
+			// For details, see:
+			// https://github.com/prometheus/prometheus/blob/e6cfa720fbe6280153fab13090a483dbd40bece3/scrape/target.go#L429
```
swiatekm, line 53: Is the intent of this comment to explain what `relabel.Process` does? It doesn't have anything to do with the code in this package.

cmd/otel-allocator/internal/prehook/testutils.go
```diff
@@ -0,0 +1,168 @@
+// Copyright The OpenTelemetry Authors
```
swiatekm, line 1: Why can't this code live in relabel_test.go?

swiatekm (commented):
This change looks broadly correct to me. My main issue with it is that we'd be changing how relabeling works in the target allocator for all users, without the ability for them to change it back in the event that the change has unforeseen consequences or bugs. For me to approve this, it needs to be configurable, either via a feature flag or configuration option.

In the interest of moving things along, how about we do the following:

1. You extract the hash calculation changes into a separate PR. These are a straightforward bug fix, so we can merge them quickly, and removing them from this PR will make it easier to review.
2. In this PR, we remove the e2e tests, and make the new behavior contingent on a new config option which will remain hidden and only used in unit tests.
3. In a separate PR, we expose the new config option in the TargetAllocator CRD and add the e2e tests using it.

How does that sound? I know I'm asking for additional work, but I really don't want to make this change haphazardly.

tests/e2e-targetallocator/targetallocator-consistency/02-install.yaml
mike9421: In this test, a single collector needs to scrape metrics from multiple prometheus targets, and different targets have varying scrape completion times. The verification script needs to wait until all target metrics are collected before performing consistency validation, so we use the batch processor to ensure data completeness. Setting a very low timeout could result in the verification only getting partial target data.
>
> P.S. My previous comment was indeed not clear enough 😅

tests/e2e-targetallocator/targetallocator-consistency/03-install.yaml
mike9421: You're right. I will split it into separate files and look for simpler approaches than the current validation script.

cmd/otel-allocator/internal/prehook/relabel.go
mike9421: Thanks for the feedback. You're right that this comment is confusing in its current placement. 

The intent of this comment was to document a behavioral difference between Prometheus and the target allocator regarding relabeling of these specific labels. In Prometheus, attempting to remove these labels via relabel_configs would cause an error, while the target allocator allows it and can still scrape normally. 

However, since this behavioral difference doesn't actually impact functionality (the target allocator works fine either way), and the comment is indeed misleading about what the local code does, I'll remove it in a follow-up change.

The key point was just that these labels shouldn't be dropped during relabeling to maintain consistency with Prometheus behavior, but since it doesn't break anything when they are dropped, the comment adds more confusion than value.

cmd/otel-allocator/internal/prehook/testutils.go
mike9421: Good point! You're right - since these functions are only used by relabel_test.go, they should live there instead of in a separate utility file. I'll move them over in the next update.

tests/e2e-targetallocator/targetallocator-consistency/02-install.yaml
swiatekm: I think it'd be simpler to just have the verification script poll periodically and check if all the data is available. You can't really guarantee completeness with what you're doing now, and the test will likely be flaky on slow GHA runners.

mike9421:
This PR went stale while I was tied up with work and personal matters, but I'm now planning to continue the split we discussed earlier — one part was already merged, and I'll work on submitting the remaining PRs. I'll review the current state of the repo and push the work soon.

mike9421:
> This PR was marked stale due to lack of activity. It will be closed in 14 days.

I've reimplemented the first of the remaining split PRs locally on top of the current main branch. Tests pass, but preliminary benchmarks show a performance regression. I'm still actively working on optimizations.

swiatekm:
@mike9421 I'd verify if your original issue is still a problem. We've added a lot of conformance tests to the target allocation and relabeling packages, alongside some architectural changes. You can also check out https://github.com/open-telemetry/opentelemetry-operator/pull/5018, which would bring us very close to what Prometheus itself does.

mike9421:
> @mike9421 I'd verify if your original issue is still a problem. We've added a lot of conformance tests to the target allocation and relabeling packages, alongside some architectural changes. You can also check out #5018, which would bring us very close to what Prometheus itself does.

@swiatekm  Thanks. My understanding is that #4381 already addressed the original duplicate-target issue and was the first part of the split. The remaining work is a separate performance optimization: avoiding the same target relabeling in both the TA and Collectors, with a default-off option added in a follow-up. Does #5018 change your view on whether that remaining optimization is still worth pursuing?

swiatekm:
> > @mike9421 I'd verify if your original issue is still a problem. We've added a lot of conformance tests to the target allocation and relabeling packages, alongside some architectural changes. You can also check out #5018, which would bring us very close to what Prometheus itself does.
> 
> @swiatekm Thanks. My understanding is that #4381 already addressed the original duplicate-target issue and was the first part of the split. The remaining work is a separate performance optimization: avoiding the same target relabeling in both the TA and Collectors, with a default-off option added in a follow-up. Does #5018 change your view on whether that remaining optimization is still worth pursuing?

I think it's worth pursuing, but the change from #5018 should go in first. Like I said, we now have proper e2e tests verifying that we output the same data as vanilla Prometheus, so the optimization is much safer to pursue now.

mike9421:
> I think it's worth pursuing, but the change from #5018 should go in first. Like I said, we now have proper e2e tests verifying that we output the same data as vanilla Prometheus, so the optimization is much safer to pursue now.

Got it, thanks for clarifying. I’ll base the optimization on #5018 and use the Prometheus-equivalence e2e tests to verify that behavior remains unchanged. The optimization PR will remain dependent on #5018, and I’ll keep the default-off configuration as a separate follow-up.

Issue #3730, opened 2025-02-21
Operator does not respect new metrics readers, favors deprecated address


csatib02:
### Component(s)

collector

## Description

Although `service::telemetry::metrics::address` field was deprecated in https://github.com/open-telemetry/opentelemetry-collector/pull/11205 starting from `v0.111.0`, an address defaulting was implemented in https://github.com/open-telemetry/opentelemetry-operator/pull/3361.

This can lead to issues:

* When using opentelemetry-operator, with a contrib distro collector, setting the telemetry like this:

```yaml
  telemetry:
    metrics:
      level: detailed
      readers:
        - pull:
            exporter:
              prometheus:
                host: ""
                port: 8888
```

gets omitted completly:

```yaml
      Telemetry:
        Metrics:
          Address:  0.0.0.0:8888
``` 
 
* On our custom otel-collector distribution, where we set readers through our operator, it also sets the deprecated address:

### Before deploying

```yaml
  telemetry:
    metrics:
      level: detailed
      readers:
        - pull:
            exporter:
              prometheus:
                host: ""
                port: 8888
```

### After deploying

```yaml
  telemetry:
    metrics:
      address: 0.0.0.0:8888
      level: detailed
      readers:
        - pull:
            exporter:
              prometheus:
                host: ""
                port: 8888
```

Which can cause the collector to crash:

```sh
Error: failed to create SDK: binding address 0.0.0.0:8888 for Prometheus exporter: listen tcp 0.0.0.0:8888: bind: address already in use
2025/02/20 10:38:05 collector server run finished with error: failed to create SDK: binding address 0.0.0.0:8888 for Prometheus exporter: listen tcp 0.0.0.0:8888: bind: address already in use
stream closed EOF for collector/otelcollector-example-collector-dh7qv (otc-container)
```

## Steps to Reproduce

1. Set a reader in a collector's config as suggested in the deprecation notice: https://github.com/open-telemetry/opentelemetry-collector/pull/11205
2. Deploy the collector.
3. Eventually it will omit the set telemetry configuration or the collector crashes.

## Expected Result

The operator should respect the deprecation and not set a default address, instead it should support and default to setting a reader.

## Actual Result

The operator omits the telemetry config and sets the deprecated default, or it crashes.

### Kubernetes Version

v1.31.2

### Operator version

v0.117.0

### Collector version

v0.120.0

## Environment
OS: Ubuntu 20.04

### Log output

```shell
[...]

PR #3744, opened 2025-02-25
Fix default internal telemetry endpoint for collector

shuheiktgw:

I've fixed the default internal telemetry endpoint for collector to use the new configuration.

- Resolves: https://github.com/open-telemetry/opentelemetry-operator/issues/3730

### DIFF
--- a/apis/v1beta1/collector_webhook_test.go
+++ b/apis/v1beta1/collector_webhook_test.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 
 	"github.com/go-logr/logr"
+	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"gopkg.in/yaml.v3"
@@ -158,7 +159,7 @@ func TestCollectorDefaultingWebhook(t *testing.T) {
 					Mode:            v1beta1.ModeDeployment,
 					UpgradeStrategy: v1beta1.UpgradeStrategyAutomatic,
 					Config: func() v1beta1.Config {
-						const input = `{"receivers":{"otlp":{"protocols":{"grpc":{"endpoint":"0.0.0.0:4317"},"http":{"endpoint":"0.0.0.0:4318"}}}},"exporters":{"debug":null},"service":{"telemetry":{"metrics":{"address":"0.0.0.0:8888"}},"pipelines":{"traces":{"receivers":["otlp"],"exporters":["debug"]}}}}`
+						const input = `{"receivers":{"otlp":{"protocols":{"grpc":{"endpoint":"0.0.0.0:4317"},"http":{"endpoint":"0.0.0.0:4318"}}}},"exporters":{"debug":null},"service":{"telemetry":{"metrics":{"readers":[{"pull":{"exporter":{"prometheus":{"host":"0.0.0.0","port":8888}}}}]}},"pipelines":{"traces":{"receivers":["otlp"],"exporters":["debug"]}}}}`
 						var cfg v1beta1.Config
 						require.NoError(t, yaml.Unmarshal([]byte(input), &cfg))
 						return cfg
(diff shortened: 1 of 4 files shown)


apis/v1beta1/config.go
```diff
@@ -474,23 +474,81 @@ func (s *Service) ApplyDefaults(logger logr.Logger) error {
...
+		metricsMap["readers"] = readers
+	}
+	readersSlice, ok := readers.([]any)
+	if !ok {
+		logger.Info("readers field is not a slice")
 		return nil
 	}
-	// NOTE: Merge without overwrite. If a telemetry endpoint is specified, the defaulting
-	// respects the configuration and returns an equal value.
-	if err := mergo.Merge(s.Telemetry, tm); err != nil {
```
shuheiktgw, line 491: I couldn't just use mergo.Merge here because it doesn't work when an element exists in the readers. It simply doesn't add the default endpoint in that case.

apis/v1beta1/collector_webhook_test.go
```diff
@@ -547,7 +548,9 @@ func TestCollectorDefaultingWebhook(t *testing.T) {
 				assert.NoError(t, test.expected.Spec.Config.Service.ApplyDefaults(logr.Discard()), "could not apply defaults")
 			}
 			assert.NoError(t, err)
-			assert.Equal(t, test.expected, test.otelcol)
+			if diff := cmp.Diff(test.expected, test.otelcol); diff != "" {
```
shuheiktgw, line 551: I’d like to use cmp.Diff instead of assert.Equal because debugging was difficult. assert.Equal doesn’t show a detailed enough diff and omits some parts if the object is too deep.

iblancasa: Can you add this as a comment to the code? I think it makes sense for future readers.

apis/v1beta1/config.go
```diff
@@ -474,23 +474,81 @@ func (s *Service) ApplyDefaults(logger logr.Logger) error {
...
-		return fmt.Errorf("telemetry config merge failed: %w", err)
+
+	var found bool
+	for _, reader := range readersSlice {
```
jaronoff97, line 506: I've been preferring approaches where you use mapstructure to deserialize into a known struct instead of this map check / type assertion stuff. [example](https://github.com/open-telemetry/opentelemetry-operator/blob/17e463a5fd12ff52b7fd405fa734b8948584ef4a/internal/components/multi_endpoint.go#L37-L56)

I think we should do a similar approach for internal telemetry especially as setting will only get more advanced going forward.

jaronoff97 (changes requested):

apis/v1beta1/config.go
shuheiktgw: Thank you for your feedback! I'm not entirely sure I understand your suggestion, so let me check if I got it right.

Are you suggesting defining a complete struct for the Telemetry field and decoding it using mapstructure.Decode? If so, I personally feel that instead of decoding it with mapstructure, we should change the type of the Telemetry field from AnyConfig to the newly defined struct.

Or are you suggesting that we use the newly defined struct to read the internal telemetry configuration and set default values dynamically if they are missing? In that case, if I understand correctly, we would still need to perform type assertions like we do now, since we can't set default values without knowing the types.

Please let me know if I'm missing anything! 🙇

iblancasa: I'm fine with this:
```
Are you suggesting defining a complete struct for the Telemetry field and decoding it using mapstructure.Decode? If so, I personally feel that instead of decoding it with mapstructure, we should change the type of the Telemetry field from AnyConfig to the newly defined struct.
```

jaronoff97: Initially we should stick with anyconfig, but we should be able to convert that anyconfig into the defined struct from go-contrib i.e. [this file](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/otelconf/v0.3.0/generated_config.go). If not, we can define a smaller intermediate type and convert to that. I want to avoid any `_, ok := ...` as much as we can.

swiatekm:
@shuheiktgw could we close this PR, given that #3874 was merged?

Issue #3086, opened 2024-07-01
Unable to run the targetAllocator in namespace mode


alita1991:
### Component(s)

target allocator

### Describe the issue you're reporting

Hi,

I'm trying to run the targetAllocator in namespace mode, but I encountered the following errors:

```
{"level":"error","ts":"2024-07-01T09:59:11Z","logger":"setup.prometheus-cr-watcher","msg":"Failed to create namespace informer in promOperator CRD watcher","error":"missing list/watch permissions on the 'namespaces' resource: missing \"list\" permission on resource \"namespaces\" (group: \"\") for all namespaces: missing \"watch\" permission on resource \"namespaces\" (group: \"\") for all namespaces","stacktrace":"[github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/watcher.NewPrometheusCRWatcher\n\t/home/runner/work/opentelemetry-operator/opentelemetry-operator/cmd/otel-allocator/watcher/promOperator.go:99\nmain.main\n\t/home/runner/work/opentelemetry-operator/opentelemetry-operator/cmd/otel-allocator/main.go:119\nruntime.main\n\t/opt/hostedtoolcache/go/1.22.4/x64/src/runtime/proc.go:271](http://github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/watcher.NewPrometheusCRWatcher/n/t/home/runner/work/opentelemetry-operator/opentelemetry-operator/cmd/otel-allocator/watcher/promOperator.go:99/nmain.main/n/t/home/runner/work/opentelemetry-operator/opentelemetry-operator/cmd/otel-allocator/main.go:119/nruntime.main/n/t/opt/hostedtoolcache/go/1.22.4/x64/src/runtime/proc.go:271)"}
[...]

PR #3573, opened 2024-12-23
feat(TargetAllocator): allow configuration of PrometheusCR namespaceSelectors

dexter0195:
**Description:**
Adding the possibility to set values for PrometheusCR: 
- `probeNamespaceSelector`
- `serviceMonitorNamespaceSelector`
- `scrapeConfigNamespaceSelector`
- `probeNamespaceSelector`

**Link to tracking Issue(s):** https://github.com/open-telemetry/opentelemetry-operator/issues/3086

- Resolves: #3086 

**Testing:** Added tests for configmap generation

**Documentation:** Updated README.md

PR #3763, opened 2025-03-03
Allow for scoping TA to watch namespace

CharlieTLe:
When the target allocator is configured to watch Prometheus custom
resources in the cluster to discover targets, it is currently hard-coded
to require a ClusterRole with a policy rule of listing namespaces. This
prevents usage in environments where configuring ClusterRoles is not
permitted i.e. in namespace-as-a-service setups where only Roles can be
created.

This change introduces two fields in the prometheusCR specification to
allow configuring the namespaces that can be interacted with by the
target allocator.

- allowNamespaces is a list of namespaces for the target allocator to
  watch. If set to an empty list, it will list all namespaces in the
  cluster. This is mutually exclusive with denyNamespaces.  Defaults to
  an empty list.
- denyNamespaces is a list of namespaces for the target allocator to not
  watch. If set to an empty list, it will not exclude any namespaces.
  This is mutually exclusive with allowNamespaces.  Defaults to an empty
  list.

Validation is performed at the admission webhook and TA startup to
validate that both fields are not set at the same time.

A new end to end test was created,
e2e-targetallocator/targetallocator-namespace, to test setting
allowNamespaces and denyNamespaces.

The first part of the test sets up a a collector and TA in a namespace suffixed with
'-top-secret'. The TA in this namespace is configured with the
allowNamespaces to the '-top-secret' namespace. Then a check is
performed to make sure that it can discover the service monitor that is
created by the otel-operator. This setup also doesn't make use of any
ClusterRole as it only creates a Role to perform its task of discovering
targets in its own namespace.

The second part of the test sets up another collector and TA in a
separate namespace without the '-top-secret' namespace with a
ClusterRole. The denyNamespaces is set to the namespace suffixed with
the '-top-secret' namespace, and later we check that we can't discover
the ServiceMonitor that is in the 'top-secret' namespace. We also check
that it can still discover the ServiceMonitor that is in its own
namespace.

Documentation was added to the otel-allocator README on how to set
allowNamespaces to configure the scope that the TA is allowed to
interact with.

Fixes: #3086


cmd/otel-allocator/internal/watcher/promOperator.go
```diff
@@ -53,7 +54,21 @@ func NewPrometheusCRWatcher(ctx context.Context, logger logr.Logger, cfg allocat
 		return nil, err
 	}
 
-	factory := informers.NewMonitoringInformerFactories(map[string]struct{}{v1.NamespaceAll: {}}, map[string]struct{}{}, mClient, allocatorconfig.DefaultResyncTime, nil) //TODO decide what strategy to use regarding namespaces
+	// Check env var for WATCH_NAMESPACE and use it if its set, else use v1.NamespaceAll
+	// This is to allow the operator to watch only a specific namespace
+	watchNamespace, found := os.LookupEnv("WATCH_NAMESPACE")
```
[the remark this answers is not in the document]
jaronoff97: I also wonder if we should follow the namespace selector pattern prometheus uses?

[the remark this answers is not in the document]
swiatekm: That is also true. @CharlieTLe could you check out https://github.com/open-telemetry/opentelemetry-operator/pull/3573 and see if it would address your need?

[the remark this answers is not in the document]
CharlieTLe: I don’t think it would, I was trying to get the TA to run without the ability to list namespaces so that it wouldn’t require a ClusterRole.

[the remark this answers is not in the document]
CharlieTLe: When 

https://github.com/open-telemetry/opentelemetry-operator/pull/3763/files#diff-c946a35cc8b11bb5df8b703eef3fb16c8676aa37948d54e63bf2bd5b04859f69R65

is used here, there’s an assumption that the service account that the TA uses has a cluster role with the permission to list namespaces in the cluster.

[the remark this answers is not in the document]
CharlieTLe: The error that gets printed when the TA attempts to list all namespace without the ClusterRole.


> missing list/watch permissions on the 'namespaces' resource: missing "list" permission on resource "namespaces" (group: "") for all namespaces: missing "watch" permission on resource "namespaces" (group: "") for all namespaces


The needed cluster role would look something like this

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: ta
rules:
- apiGroups:
  - ""
  resources:
  - namespaces
  verbs:
  - watch
  - list
```

So the main change in this PR is to not use `v1.NamespaceAll` so that the TA can run in a namespace-scoped Role when the WATCH_NAMESPACE env var is set.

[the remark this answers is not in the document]
swiatekm: Allright, I can see this being useful. We like to align these kinds of options with how prometheus-operator does it. There, this is a [command-line argument ](https://github.com/prometheus-operator/prometheus-operator/blob/5eb6568d3e5dd8bc59fefc354124d1b0528a1b9a/cmd/operator/main.go#L165)for the operator, which won't work in our case, but we can use the same terminology. So this would be named `AllowNamespaces`, and we could also add `DenyNamespaces` to let this set be constrained from the other side. WDYT @jaronoff97 ?

swiatekm:
>     * allowNamespaces is a comma-separated list of namespaces for the target
>       allocator to watch. If set to an empty string, it will list all
>       list all namespaces in the cluster. This is mutually exclusive with
>       denyNamespaces. Defaults to an empty string.
> 
>     * denyNamespaces is a comma-separated list of namespaces for the target
>       allocator to not watch. If set to an empty string, it will not exclude
>       any namespaces. This is mutually exclusive with allowNamespaces.
>       Defaults to an empty string.

Why do these need to be comma-separated lists represented by strings? This made sense when we were going to use an environment variable, but both in CRDs and YAML config files, we can just use a list of strings.

Also, if we want these options to be mutually-exclusive, we need to implement validation for that logic in the admission webhook.

internal/manifests/targetallocator/configmap.go
```diff
@@ -94,6 +94,9 @@ func ConfigMap(params Params) (*corev1.ConfigMap, error) {
 			prometheusCRConfig["scrape_interval"] = taSpec.PrometheusCR.ScrapeInterval.Duration
 		}
 
+		prometheusCRConfig["allow_namespaces"] = taSpec.PrometheusCR.AllowNamespaces
+		prometheusCRConfig["deny_namespaces"] = taSpec.PrometheusCR.DenyNamespaces
```
swiatekm, line 98: It'd make things easier if we only set these when they were non-empty. Then we wouldn't need to update all the tests.

cmd/otel-allocator/internal/watcher/promOperator.go
```diff
@@ -53,7 +53,30 @@ func NewPrometheusCRWatcher(ctx context.Context, logger logr.Logger, cfg allocat
 		return nil, err
 	}
 
-	factory := informers.NewMonitoringInformerFactories(map[string]struct{}{v1.NamespaceAll: {}}, map[string]struct{}{}, mClient, allocatorconfig.DefaultResyncTime, nil) //TODO decide what strategy to use regarding namespaces
+	if len(cfg.PrometheusCR.AllowNamespaces) != 0 && len(cfg.PrometheusCR.DenyNamespaces) != 0 {
+		return nil, fmt.Errorf("both allow and deny namespaces are set, only one should be set")
+	}
```
swiatekm, line 58: This check should happen during config validation in config.go. Here, we can assume it's valid.

tests/e2e-targetallocator/targetallocator-namespace/assert-job-failed.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: check-metrics
+  namespace: ($namespace)
```
swiatekm, line 5: The namespace is assumed to be the test namespace if you don't specify it, so this isn't necessary here, nor for the other resources.

swiatekm (commented):
Looks good to me overall, thank you for working on this! I like the E2E test in particular, nice use of chainsaw.

I have a few remaining comments, afterwards we should be good to go.

cmd/otel-allocator/internal/watcher/promOperator.go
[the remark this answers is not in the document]
jaronoff97: Yeah i definitely like this approach – if we can align to what prom does i think it provides a nice consistent UX

cmd/otel-allocator/README.md
```diff
@@ -263,8 +265,82 @@ rules:
...
+    enabled: true
+    serviceAccount: opentelemetry-targetallocator-sa
+    prometheusCR:
+      enabled: true
+      allowNamespaces: 
+      - foo
+```
+
+In this case, you will need to create a Role and RoleBinding instead of a ClusterRole and ClusterRoleBinding. The Role
+and RoleBinding should be created in the namespace specified by the allowNamespaces field.
```
jaronoff97, line 287: Would you mind writing an issue for a follow up to use our auto-rbac system to make these for users if they've enabled the operator to do so?

cmd/otel-allocator/internal/watcher/promOperator.go
jaronoff97: I think it's good to do this here too IMO, we do have some users who are asking to be able to self-manage the TA outside of the operator and doing this validation here wouldn't hurt.

cmd/otel-allocator/internal/watcher/promOperator.go
```diff
@@ -53,7 +53,30 @@ func NewPrometheusCRWatcher(ctx context.Context, logger logr.Logger, cfg allocat
...
+
+	allowList := map[string]struct{}{}
```
jaronoff97, line 60: minor suggestion, take it or leave it... Can we move the construction of these lists to a method on the PrometheusCR object?

jaronoff97 (approved):
Just a few minor things, otherwise this looks great. Thank you for the work here! This solves a highly requested user issue so this is great to see.

cmd/otel-allocator/internal/watcher/promOperator.go
swiatekm: I agree it should be done in the target allocator application. I just think it should be done when validation configuration rather than when instantiating this struct.

jaronoff97: ah i misread, apologies :bow:

tests/e2e-targetallocator/targetallocator-namespace/assert-job-failed.yaml
CharlieTLe: It's specified to allow the job to run in a separate namespace than the one originally setup by chainsaw.

For example, if the namespace chainsaw uses in its test is chainsaw-happy-dog, then I want to set up the Collector + TA in chainsaw-happy-dog-top-secret to test setting allowNamespaces to just chainsaw-happy-dog-top-secret so that it isn't discovering any other namespaces in the cluster, i.e. chainsaw-happy-dog.

Setting the built-in binding `namespace` wasn't enough for chainsaw to deploy the resources into that namespace; the resource must explicitly specify the namespace field and `($namespace)` for it to work. Otherwise, chainsaw would have deployed the resource into chainsaw-happy-dog instead of chainsaw-happy-dog-top-secret.

cmd/otel-allocator/README.md
CharlieTLe: Created the issue here: https://github.com/open-telemetry/opentelemetry-operator/issues/3816

swiatekm (approved):
Nice!

PR #3780, opened 2025-03-06
Fix auto-instrumentation with multi annotation

### LINKED ISSUE #3407: Auto instrumentation fails when the annotation set on both the namespace and the pod with different values
### Component(s)

_No response_

## Description
I have a use case where most of the apps inside a namespace are of one language (python apps), with some being a different language (go applications) . I set the annotation for to auto inject python on the namespace level and the two annotations to inject go ( the inject and auto target )on the pod level. 

## Steps to Reproduce
put any annotation related to injection on the namespace equal to true
put a different annotation on the pod level (other than go, tried also with nodejs)

## Expected Result
the annotation on the pod level to take precedence and instrumentation happening

## Actual Result
the instrumentation doesn't happen it with the operator spewing out an error message saying I need to provide container names for all instrumentations.

### Kubernetes Version

1.30.4+k3s1

### Operator version

0.111.0

### Collector version

NA

## Environment
OS: arch-linux

### Log output

```shell
{"level":"ERROR","timestamp":"2024-10-29T11:05:03Z","message":"skipping instrumentation injection","namespace":"btech-apps","name":"trace-generator-pod","error":"incorrect instrumentation configuration - please provide container names for all instrumentations","stacktrace":"github.com/open-telemetry/opentelemetry-operator/pkg/instrumentation.(*instPodMutator).Mutate\n\t/home/runner/work/opentelemetry-operator/opentelemetry-operator/pkg/instrumentation/podmutator.go:408\ngithub.com/open-telemetry/opentelemetry-operator/internal/webhook/podmutation.(*podMutationWebhook).Handle\n\t/home/runner/work/opentelemetry-operator/opentelemetry-operator/internal/webhook/podmutation/webhookhandler.go:93\nsigs.k8s.io/controller-runtime/pkg/webhook/admission.(*Webhook).Handle\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/webhook/admission/webhook.go:181\nsigs.k8s.io/controller-runtime/pkg/webhook/admission.(*Webhook).ServeHTTP\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/webhook/admission/http.go:119\nsigs.k8s.io/controller-runtime/pkg/webhook/internal/metrics.InstrumentedHook.InstrumentHandlerInFlight.func1\n\t/home/runner/go/pkg/mod/github.com/prometheus/client_golang@v1.20.4/prometheus/promhttp/instrument_server.go:60\nnet/http.HandlerFunc.ServeHTTP\n\t/opt/hostedtoolcache/go/1.22.8/x64/src/net/http/server.go:2171\ngithub.com/prometheus/client_golang/prometheus/promhttp.InstrumentHandlerCounter.func1\n\t/home/runner/go/pkg/mod/github.com/prometheus/client_golang@v1.20.4/prometheus/promhttp/instrument_server.go:147\nnet/http.HandlerFunc.ServeHTTP\n\t/opt/hostedtoolcache/go/1.22.8/x64/src/net/http/server.go:2171\ngithub.com/prometheus/client_golang/prometheus/promhttp.InstrumentHandlerDuration.func2\n\t/home/runner/go/pkg/mod/github.com/prometheus/client_golang@v1.20.4/prometheus/promhttp/instrument_server.go:109\nnet/http.HandlerFunc.ServeHTTP\n\t/opt/hostedtoolcache/go/1.22.8/x64/src/net/http/server.go:2171\nnet/http.(*ServeMux).ServeHTTP\n\t/opt/hostedtoolcache/go/1.22.8/x64/src/net/http/server.go:2688\nnet/http.serverHandler.ServeHTTP\n\t/opt/hostedtoolcache/go/1.22.8/x64/src/net/http/server.go:3142\nnet/http.(*conn).serve\n\t/opt/hostedtoolcache/go/1.22.8/x64/src/net/http/server.go:2044"}
```

### Additional context

If this is the default behavior and I have to put the same annotation from the namespace on the pod but set it to false in order for the instrumentation to take place, please consider having the new behavior I'm describing as the default especially since this is happening with single container pods

yuriolisa:

It fix a case when the auto instrumentation fails with annotations set on both namespace and pod with different valid values.

- Resolves: #3407 

### DIFF
--- a/internal/instrumentation/annotation.go
+++ b/internal/instrumentation/annotation.go
@@ -4,8 +4,6 @@
 package instrumentation
 
 import (
-	"strings"
-
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -34,34 +32,91 @@ const (
 	annotationInjectNginxContainersName       = "instrumentation.opentelemetry.io/inject-nginx-container-names"
 )
 
-// annotationValue returns the effective annotationInjectJava value, based on the annotations from the pod and namespace.
-func annotationValue(ns metav1.ObjectMeta, pod metav1.ObjectMeta, annotation string) string {
-	// is the pod annotated with instructions to inject sidecars? is the namespace annotated?
-	// if any of those is true, a sidecar might be desired.
-	podAnnValue := pod.Annotations[annotation]
-	nsAnnValue := ns.Annotations[annotation]
-
-	// if the namespace value is empty, the pod annotation should be used, whatever it is
-	if len(nsAnnValue) == 0 {
-		return podAnnValue
+func hasMatchingInjectAnnotation(annotations map[string]string, containerAnnotation string) (string, bool) {
+	annotationPairs := map[string]string{
+		annotationInjectJavaContainersName:        annotationInjectJava,
+		annotationInjectNodeJSContainersName:      annotationInjectNodeJS,
+		annotationInjectPythonContainersName:      annotationInjectPython,
+		annotationInjectDotnetContainersName:      annotationInjectDotNet,
+		annotationInjectGoContainersName:          annotationInjectGo,
+		annotationInjectSdkContainersName:         annotationInjectSdk,
+		annotationInjectApacheHttpdContainersName: annotationInjectApacheHttpd,
+		annotationInjectNginxContainersName:       annotationInjectNginx,
 	}
 
-	// if the pod value is empty, the annotation should be used (true, false, instance)
-	if len(podAnnValue) == 0 {
+	injectAnnotation, exists := annotationPairs[containerAnnotation]
+	if !exists {
+		return "", false
+	}
+
+	_, found := annotations[injectAnnotation]
+	return injectAnnotation, found
+}
+
+func allInstrumentationAnnotations() []string {
+	return []string{
+		annotationInjectJava,
+		annotationInjectNodeJS,
+		annotationInjectPython,
+		annotationInjectDotNet,
+		annotationInjectGo,
+		annotationInjectSdk,
+		annotationInjectApacheHttpd,
+		annotationInjectNginx,
+	}
+}
+
+func hasAnyInstrumentationAnnotation(annotations map[string]string) bool {
+	if annotations == nil {
+		return false
+	}
+	for _, annotation := range allInstrumentationAnnotations() {
+		if val, exists := annotations[annotation]; exists && len(val) > 0 {
+			return true
+		}
+	}
+	return false
+}
+
+// annotationValue returns the effective annotation value, based on the annotations from the pod and namespace.
+// Implementation of the unified precedence: if pod has any instrumentation annotation,
+// namespace instrumentation annotations are ignored completely.
+func annotationValue(ns metav1.ObjectMeta, pod metav1.ObjectMeta, annotation string) string {
+	if injectAnnotation, isContainerNames := hasMatchingInjectAnnotation(pod.Annotations, annotation); isContainerNames {
+		if _, injectExists := pod.Annotations[injectAnnotation]; injectExists {
+			if val, exists := pod.Annotations[annotation]; exists && len(val) > 0 {
+				return val
+			}
+		}
+	}
+
+	if injectAnnotation, isContainerNames := hasMatchingInjectAnnotation(ns.Annotations, annotation); isContainerNames {
+		if _, injectExists := ns.Annotations[injectAnnotation]; injectExists {
+			if val, exists := ns.Annotations[annotation]; exists && len(val) > 0 {
+				return val
+			}
+		}
+	}
+
+	// Check if pod has any instrumentation annotations
+	podHasInstrumentationAnnotations := hasAnyInstrumentationAnnotation(pod.Annotations)
+
+	podAnnValue, podExists := pod.Annotations[annotation]
+	nsAnnValue, nsExists := ns.Annotations[annotation]
+
+	if podHasInstrumentationAnnotations {
+		if podExists && len(podAnnValue) > 0 {
+			return podAnnValue
+		}
+		return ""
+	}
+
+	if podExists && len(podAnnValue) > 0 {
+		return podAnnValue
+	}
+	if nsExists && len(nsAnnValue) > 0 {
 		return nsAnnValue
 	}
 
-	// the pod annotation isn't empty -- if it's an instance name, or false, that's the decision
-	if !strings.EqualFold(podAnnValue, "true") {
-		return podAnnValue
-	}
-
-	// pod annotation is 'true', and if the namespace annotation is false, we just return 'true'
-	if strings.EqualFold(nsAnnValue, "false") {
-		return podAnnValue
-	}
-
-	// by now, the pod annotation is 'true', and the namespace annotation is either true or an instance name
-	// so, the namespace annotation can be used
-	return nsAnnValue
+	return ""
 }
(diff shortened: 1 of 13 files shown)


tests/e2e-instrumentation/instrumentation-dotnet-multiannotation/00-install-collector.yaml
```diff
@@ -0,0 +1,31 @@
...
+  annotations:
+    instrumentation.opentelemetry.io/inject-sdk: "true"
+---
+apiVersion: opentelemetry.io/v1alpha1
+kind: OpenTelemetryCollector
+metadata:
+  name: sidecar
+  namespace: my-other-dotnet-ns
+spec:
+  config: |
```
jaronoff97, line 15: let's use the new collector version here :)

jaronoff97 (approved):
minor comment :)

tests/e2e-instrumentation/instrumentation-dotnet-multiannotation/00-install-instrumentation.yaml
```diff
@@ -0,0 +1,28 @@
...
+      value: otlp
+    - name: OTEL_EXPORTER_OTLP_ENDPOINT
+      value: http://localhost:4317
+    - name: OTEL_EXPORTER_OTLP_TIMEOUT
+      value: "20"
+    - name: OTEL_TRACES_SAMPLER
+      value: parentbased_traceidratio
+    - name: OTEL_TRACES_SAMPLER_ARG
+      value: "0.85"
+    - name: SPLUNK_TRACE_RESPONSE_HEADER_ENABLED
```
iblancasa, line 19: Is this really needed?

yuriolisa: No, I can remove it.

tests/e2e-instrumentation/instrumentation-dotnet-multiannotation/00-install-collector.yaml
yuriolisa: Just fixed.

iblancasa (approved):

pavolloffay:
@yuriolisa the e2e tests are failing

yuriolisa:
> @yuriolisa the e2e tests are failing

@pavolloffay, I was working on it today after talking with @iblancasa about the best approach to the pod's precedence. You can see through this slack [thread](https://cloud-native.slack.com/archives/C056QUS3RPV/p1742327882874819).

internal/instrumentation/annotation.go
```diff
@@ -34,34 +32,91 @@ const (
 	annotationInjectNginxContainersName       = "instrumentation.opentelemetry.io/inject-nginx-container-names"
 )
 
-// annotationValue returns the effective annotationInjectJava value, based on the annotations from the pod and namespace.
-func annotationValue(ns metav1.ObjectMeta, pod metav1.ObjectMeta, annotation string) string {
-	// is the pod annotated with instructions to inject sidecars? is the namespace annotated?
-	// if any of those is true, a sidecar might be desired.
-	podAnnValue := pod.Annotations[annotation]
-	nsAnnValue := ns.Annotations[annotation]
+func hasMatchingInjectAnnotation(annotations map[string]string, containerAnnotation string) (string, bool) {
```
pavolloffay, line 35: What is the container annotation? Should it be pod annotation?

pavolloffay (approved):

internal/instrumentation/annotation.go
yuriolisa: I've used the containerAnnotation because on the function annotationValue I evaluate the precedence pod > pod-with-containers-name > namespace.

internal/instrumentation/podmutator.go
```diff
@@ -131,26 +105,10 @@ func (langInsts languageInstrumentations) areInstrumentedContainersCorrect() (bo
...
-	// Look for mixed multiple instrumentations with and without container names.
-	if instrWithoutContainers > 0 && instrWithContainers > 0 {
-		return false, fmt.Errorf("incorrect instrumentation configuration - please provide container names for all instrumentations")
-	}
-
-	// Look for multiple instrumentations without container names.
-	if instrWithoutContainers > 1 && instrWithContainers == 0 {
-		return false, fmt.Errorf("incorrect instrumentation configuration - please provide container names for all instrumentations")
-	}
-
```
jaronoff97, line 143: still wrapping my head around the code, but why do we not need this anymore?

internal/instrumentation/sdk.go
```diff
@@ -228,6 +228,52 @@ func (i *sdkInjector) setInitContainerSecurityContext(pod corev1.Pod, securityCo
 	return pod
 }
 
+func containersByLanguage(pod corev1.Pod, language string) []string {
+	var detectedContainers []string
+
+	languagePatterns := map[string][]string{
```
jaronoff97: should this be a const?

jaronoff97 (commented):
one question, one suggestion. otherwise i think this makes sense.

iblancasa (approved):

internal/instrumentation/podmutator.go
yuriolisa: Removed this validation to avoid the combination `sdk` and `container` instrumentation, like: having `inject-sdk: "true"` which applies to all containers and at the same time the `inject-python: "true" + python-container-names: "pythonapp"`.

internal/instrumentation/sdk.go
yuriolisa: I could change to package-level var.

jaronoff97 (approved):

internal/instrumentation/sdk.go
```diff
@@ -228,6 +239,41 @@ func (i *sdkInjector) setInitContainerSecurityContext(pod corev1.Pod, securityCo
 	return pod
 }
 
+func containersByLanguage(pod corev1.Pod, language string) []string {
```
swiatekm: Is this necessary to fix the bug? It looks like a new feature where we try to guess which container to inject into using a heuristic. I don't mind this feature, but I'd rather it be submitted in a separate PR.

swiatekm (commented):
The fix itself looks good to me. I'd rather see the feature added in a separate PR, unless it's necessary here in a way I don't understand.

internal/instrumentation/sdk.go
yuriolisa: Good question, but it's necessary to take the evaluation of the precedence working, like: pod > pod-with-containers-name > namespace. Otherwise will fail if we mix up, sdk, containers, and automatic instrumentation based on the containers name.

yuriolisa: @swiatekm, could you please check it out again before I submit the branch conflict fix?

swiatekm: I'll admit I don't really get why it's necessary. The original issue only involves annotation precedence. Why can't we fix only that problem? @jaronoff97 @iblancasa you approved this PR, does this make sense to you? I don't want to hold this fix up through my ignorance, but I want to make sure we know what kind of change we're making here.

iblancasa: After a second review I agree that maybe the change may not be 100% needed even when it brings benefits...
Reading the code, this fixes something related to the main issue but not needed by the main issue.

For me it is fine to let this thing here but add it to the changelog.

PR #3797, opened 2025-03-12
Recommeded resource attributes

zeitlinger:

Updates the recommended resource attributes to comply with the semantic conventions.

https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/k8s-attributes.md

Technically, it's a breaking change - looking for feedback if it's OK to change as is.

- Resolves: https://github.com/open-telemetry/semantic-conventions/pull/1756

Adapted existing tests.

Adapted existing docs.


iblancasa (approved):
LGTM. Please fix the CI

iblancasa (commented):
We discussed about this PR during the SIG call. Some comments:
1. Could you add a feature gate to avoid the breaking change right now and give some time to users to upgrade?
2. Are the semantic convention changes already released?
3. Is there something we can do to avoid the breaking change?

.chloggen/recommedded-resource-attributes.yaml
```diff
@@ -0,0 +1,22 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: auto-instrumentation
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Update recommended resource attributes to match the [semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/k8s-attributes.md)
```
iblancasa, line 8: Please add more info about the exact changes that need to be done for the operator.

Also, provide some kind of recommendation regarding upgrading.

zeitlinger: the subtext has all changes - is that sufficient?

The subtext should also cover what you need to consider for upgrading.

zeitlinger:
> We discussed about this PR during the SIG call. Some comments:
> 
> 1. Could you add a feature gate to avoid the breaking change right now and give some time to users to upgrade?

yes, we could - but I'm not sure if that's really warranted in this case:

1. the exact format of the version is unlikely to affect anyone
2. the well-known labels need an extra opt-in already

See

https://github.com/open-telemetry/opentelemetry-operator/pull/3797/files#diff-5fd4df0f3bace6e5f00b361ea2060d8ddf95cf8136445af1a530c2b81afe2c2bR17-R21

Having said that, I'm happy to add a flag if we decide to do so

> 2. Are the semantic convention changes already released?

yes, see https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/k8s-attributes.md

> 3. Is there something we can do to avoid the breaking change?

we could add a different flag for the well-known attributes and deprecate the old flag
that would only leave the version format - which is less likely to affect anyone

zeitlinger:
just found a missing piece: `k8s.namespace.name` should also be translated to `service.namespace`

zeitlinger:
@iblancasa now it should be complete :smile:

pkg/instrumentation/sdk.go
```diff
@@ -431,48 +430,88 @@ func chooseServiceName(pod corev1.Pod, useLabelsForResourceAttributes bool, reso
...
+		return namespace
+	}
+	return namespaceName
+}
+
+var cannotRetrieveImage = errors.New("cannot retrieve image name")
+
+// parseServiceVersionFromImage parses the service version for differently-formatted image names
+// according to https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/k8s-attributes.md#how-serviceversion-should-be-calculated
+func parseServiceVersionFromImage(image string) (string, error) {
```
swiatekm, line 476: Could you add some unit tests for these new functions?

zeitlinger: it's here - and many tests:

image:

https://github.com/open-telemetry/opentelemetry-operator/blob/8d35a59c9f1ab26df4a6a6765cc2daac6763fbb8/pkg/instrumentation/sdk_test.go#L2645-L2683

namespace:

https://github.com/open-telemetry/opentelemetry-operator/blob/8d35a59c9f1ab26df4a6a6765cc2daac6763fbb8/pkg/instrumentation/podmutator_test.go#L853

swiatekm: I don't see `service.version` checked anywhere in there. You've changed the behaviour for that attribute, so I'd like to see that reflected in a unit test. Plus. you've added a new function which doesn't look trivial to me, so it should have some basic tests to show that it actually does what the docstring claims.

zeitlinger: > I don't see `service.version` checked anywhere in there. You've changed the behaviour for that attribute, so I'd like to see that reflected in a unit test. 

here's one: https://github.com/open-telemetry/opentelemetry-operator/blob/8d35a59c9f1ab26df4a6a6765cc2daac6763fbb8/pkg/instrumentation/sdk_test.go#L229

It also shows in the diff when you open that file.

> Plus. you've added a new function which doesn't look trivial to me, so it should have some basic tests to show that it actually does what the docstring claims.

this is the test https://github.com/open-telemetry/opentelemetry-operator/blob/8d35a59c9f1ab26df4a6a6765cc2daac6763fbb8/pkg/instrumentation/sdk_test.go#L2645-L2683

or do you mean another function?

swiatekm: Ah, I see it, my bad. Completely missed that file was being changed at all during review.

swiatekm (approved):
I am ok with this going in without a feature flag, but I'd like opinions from other maintainers before we merge.

iblancasa:
> I am ok with this going in without a feature flag, but I'd like opinions from other maintainers before we merge.

I thing we should include this using a feature gate to give users some time to migrate. But not strong opinion.

swiatekm:
> > I am ok with this going in without a feature flag, but I'd like opinions from other maintainers before we merge.
> 
> I thing we should include this using a feature gate to give users some time to migrate. But not strong opinion.

Rather than a feature gate, I think I'd prefer some instructions on how to get the old behaviour by changing resource labels.

zeitlinger:
> > > I am ok with this going in without a feature flag, but I'd like opinions from other maintainers before we merge.
> > 
> > 
> > I thing we should include this using a feature gate to give users some time to migrate. But not strong opinion.
> 
> Rather than a feature gate, I think I'd prefer some instructions on how to get the old behaviour by changing resource labels.

I've added a section in the changelog

jaronoff97 (approved):
looks great, thanks for your work here @zeitlinger !

Issue #3781, opened 2025-03-07
[target-allocator] TargetAllocator assigns Targets to an unhealthy collector?


rice-junhaoyu:
### Component(s)

target allocator

## Description
I am wondering if it's expected that TargetAllocator still assign Targets to an unhealthy collector?

From the code [[link]](https://github.com/open-telemetry/opentelemetry-operator/blob/main/cmd/otel-allocator/internal/collector/collector.go#L116-L129), it seems there's no pod status check at all.

Can someone help double check if this is true?

## Steps to Reproduce
1. Create regular TargetAllocator and OTEL Collectors.
2. Specify the OTEL Collector StatefulSet to use a bad image or etc, this will cause a OTEL Collector pod to be in a non-Running state.
3. Port forward to the TargetAllocator, curl `http://localhost:8080/jobs/kube-state-metrics/targets`, and you can see the unhealthy OTEL Collector pod is still a candidate to assign targets to.

## Expected Result
We shouldn't assign targets to an unhealthy OTEL Collector? Or we should have a configuration that enables such a feature?

## Actual Result
I was able to make a collector unhealthy but still seeing the unhealthy collector on `http://localhost:8080/jobs/kube-state-metrics/targets`.

### Kubernetes Version

1.31

### Operator version

v0.120.0

### Collector version

v0.120.0

swiatekm:
The problem here is that it's not clear what the behaviour should be. Ideally, we shouldn't assign targets to collectors which aren't healthy, but we also don't want to reallocate all targets just because a single collector was temporarily unhealthy. I had a basic idea on how this could be done in https://github.com/open-telemetry/opentelemetry-operator/issues/2201, but I'm not sure what the possible consequences could be, or how to test it exhaustively.

Maybe the way to go is to add a configurable grace period, and let users turn the behaviour off completely by setting that to 0. If anyone wants to submit a change like that, I'd be willing to review it. The other issue is assigned to me, but I've clearly not made much progress on it since then.

rice-junhaoyu:
Hey Mikolaj,

I agree that always trigger a target reallocation for a bad collector is not the ideal solution, that could be too much.

Just one more question, I am still confused about `Maybe the way to go is to add a configurable grace period, and let users turn the behaviour off completely by setting that to 0.`. Can you elaborate a bit more and maybe point me to somewhere? I am kinda new to OTEL and not aware of such a `grace period`.

[...]

PR #3807, opened 2025-03-15
Do not assign targets to a collector pod that are not Ready for accepting traffic more than a while

rice-junhaoyu:
**Description:**

Ideally, TargetAllocator should not assign targets to an unhealthy pod that is not ready for accepting traffic.

By discussing with Mikolaj in [#3781](https://github.com/open-telemetry/opentelemetry-operator/issues/3781), we believe that it will be a good idea to NOT assign targets to an unhealthy collector pod that has been staying in the bad state for a grace period of `30s` or something.

Note that the `unhealthy` here not only necessarily means `Pod Phase is non-Running`, but more of saying the pod's `PodCondition Ready is non-True`.
- `Running` is a Pod Phase that is defined as: "The Pod has been bound to a node, and all of the containers have been created. At least one container is still running, or is in the process of starting or restarting."
- `Ready` is a Pod Condition that is defined as: "The Pod is able to serve requests and should be added to the load balancing pools of all matching Services."

Although I believe the check on `Ready` PodCondition is already covering  the check on `Running` PodPhase, but I am fine to keep both of them as of now. Let me know your opinion if you have a strong one.

Resolves: [#3781](https://github.com/open-telemetry/opentelemetry-operator/issues/3781)

**Testing:**

Unit Test:
- I added two unit tests that
   - `given three pods (1) a running pod (2) a non-running pod but is still within grace period (3) a non-running pod and is already over grace period`, the test passed.
   - `given three pods (1) a ready pod (2) a non-ready pod but is still within grace period (3) a non-ready pod and is already over grace period`, the test passed.
- Also I want to call out that existing unit tests are not impacted.

Local Test:
- I built a docker image with this change and verified that if a pod is non-running for too long, TargetAllocator will not assign any more targets to it.

Local Test Step By Step:
1. Initial State was 10 available collector pods and two target-allocator pods with the change in this PR
```
otel-collector-with-ta-kube-state-metrics-collector-0             1/1     Running       0            31h
otel-collector-with-ta-kube-state-metrics-collector-1             1/1     Running       0            31h
otel-collector-with-ta-kube-state-metrics-collector-2             1/1     Running       0            31h
otel-collector-with-ta-kube-state-metrics-collector-3             1/1     Running       0            31h
otel-collector-with-ta-kube-state-metrics-collector-4             1/1     Running       0            31h
otel-collector-with-ta-kube-state-metrics-collector-5             1/1     Running       0            31h
otel-collector-with-ta-kube-state-metrics-collector-6             1/1     Running       0            31h
otel-collector-with-ta-kube-state-metrics-collector-7             1/1     Running       0            31h
otel-collector-with-ta-kube-state-metrics-collector-8             1/1     Running       0            31h
otel-collector-with-ta-kube-state-metrics-collector-9             1/1     Running       0            61s
otel-collector-with-ta-kube-state-metrics-targetallocator-kp5xm   1/1     Running       0            5m15s
otel-collector-with-ta-kube-state-metrics-targetallocator-p9kvl   1/1     Running       0            3m50s
```

By port-forwarding to `http://localhost:8080/jobs/kube-state-metrics/targets`, I could see 10 collectors are candidates.

Screenshot 2025-03-14 at 7 04 47 PM

2. I changed collectors to use a bad image and then `collector-9` went into non-Running state
```
otel-collector-with-ta-kube-state-metrics-collector-0             1/1     Running        0            31h
otel-collector-with-ta-kube-state-metrics-collector-1             1/1     Running        0            31h
otel-collector-with-ta-kube-state-metrics-collector-2             1/1     Running        0            31h
otel-collector-with-ta-kube-state-metrics-collector-3             1/1     Running        0            31h
otel-collector-with-ta-kube-state-metrics-collector-4             1/1     Running        0            31h
otel-collector-with-ta-kube-state-metrics-collector-5             1/1     Running        0            31h
otel-collector-with-ta-kube-state-metrics-collector-6             1/1     Running        0            31h
otel-collector-with-ta-kube-state-metrics-collector-7             1/1     Running        0            31h
otel-collector-with-ta-kube-state-metrics-collector-8             1/1     Running        0            31h
otel-collector-with-ta-kube-state-metrics-collector-9             0/1     ErrImagePull   0            37s
otel-collector-with-ta-kube-state-metrics-targetallocator-kp5xm   1/1     Running        0            5m55s
otel-collector-with-ta-kube-state-metrics-targetallocator-p9kvl   1/1     Running        0            4m30s
```

Pod Condition of `Ready` also went `False` in a bit.

image

By port-forwarding to `http://localhost:8080/jobs/kube-state-metrics/targets`, I could only see 9 collectors are candidates.

Screenshot 2025-03-14 at 7 08 53 PM

### DIFF
--- a/.chloggen/ta_not_assign_targets_to_bad_collectors.yaml
+++ b/.chloggen/ta_not_assign_targets_to_bad_collectors.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: target allocator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Do not assign targets to a collector pod that is not Ready for longer than a non-zero grace period
+
+# One or more tracking issues related to the change
+issues: [3781]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/cmd/otel-allocator/internal/collector/collector.go
+++ b/cmd/otel-allocator/internal/collector/collector.go
@@ -31,23 +31,25 @@ var (
 )
 
 type Watcher struct {
-	log               logr.Logger
-	k8sClient         kubernetes.Interface
-	close             chan struct{}
-	minUpdateInterval time.Duration
+	log                          logr.Logger
+	k8sClient                    kubernetes.Interface
+	close                        chan struct{}
+	minUpdateInterval            time.Duration
+	collectorNotReadyGracePeriod time.Duration
 }
 
-func NewCollectorWatcher(logger logr.Logger, kubeConfig *rest.Config) (*Watcher, error) {
+func NewCollectorWatcher(logger logr.Logger, kubeConfig *rest.Config, collectorNotReadyGracePeriod time.Duration) (*Watcher, error) {
 	clientset, err := kubernetes.NewForConfig(kubeConfig)
 	if err != nil {
 		return &Watcher{}, err
 	}
 
 	return &Watcher{
-		log:               logger.WithValues("component", "opentelemetry-targetallocator"),
-		k8sClient:         clientset,
-		close:             make(chan struct{}),
-		minUpdateInterval: defaultMinUpdateInterval,
+		log:                          logger.WithValues("component", "opentelemetry-targetallocator"),
+		k8sClient:                    clientset,
+		close:                        make(chan struct{}),
+		minUpdateInterval:            defaultMinUpdateInterval,
+		collectorNotReadyGracePeriod: collectorNotReadyGracePeriod,
 	}, nil
 }
 
@@ -124,6 +126,12 @@ func (k *Watcher) runOnCollectors(store cache.Store, fn func(collectors map[stri
 		if pod.Spec.NodeName == "" {
 			continue
 		}
+
+		// pod healthiness check will always be disabled if CollectorNotReadyGracePeriod is set to 0 * time.Second
+		if k.isPodUnhealthy(pod, k.collectorNotReadyGracePeriod) {
+			continue
+		}
+
 		collectorMap[pod.Name] = allocation.NewCollector(pod.Name, pod.Spec.NodeName)
 	}
 	collectorsDiscovered.Set(float64(len(collectorMap)))
@@ -132,4 +140,26 @@ func (k *Watcher) runOnCollectors(store cache.Store, fn func(collectors map[stri
 
 func (k *Watcher) Close() {
 	close(k.close)
+}
+
+func (k *Watcher) isPodUnhealthy(pod *v1.Pod, collectorNotReadyGracePeriod time.Duration) bool {
+	if collectorNotReadyGracePeriod == 0*time.Second {
+		return false
+	}
+
+	isPodUnhealthy := false
+	timeNow := time.Now()
+	// stop assigning targets to a non-Running pod that has lasted for a specific period
+	if pod.Status.Phase != v1.PodRunning &&
+		pod.Status.StartTime != nil &&
+		(timeNow.Sub(pod.Status.StartTime.Time) > collectorNotReadyGracePeriod) {
+		isPodUnhealthy = true
+	}
+	// stop assigning targets to a non-Ready pod that has lasted for a specific period
+	for _, condition := range pod.Status.Conditions {
+		if condition.Type == v1.PodReady && condition.Status != v1.ConditionTrue && (timeNow.Sub(condition.LastTransitionTime.Time) > collectorNotReadyGracePeriod) {
+			isPodUnhealthy = true
+		}
+	}
+	return isPodUnhealthy
 }
--- a/cmd/otel-allocator/internal/collector/collector_test.go
+++ b/cmd/otel-allocator/internal/collector/collector_test.go
@@ -29,12 +29,13 @@ var labelSelector = metav1.LabelSelector{
 	MatchLabels: labelMap,
 }
 
-func getTestPodWatcher() Watcher {
+func getTestPodWatcher(collectorNotReadyGracePeriod time.Duration) Watcher {
 	podWatcher := Watcher{
-		k8sClient:         fake.NewSimpleClientset(),
-		close:             make(chan struct{}),
-		log:               logger,
-		minUpdateInterval: time.Millisecond,
+		k8sClient:                    fake.NewSimpleClientset(),
+		close:                        make(chan struct{}),
+		log:                          logger,
+		minUpdateInterval:            time.Millisecond,
+		collectorNotReadyGracePeriod: collectorNotReadyGracePeriod,
 	}
 	return podWatcher
 }
@@ -49,14 +50,51 @@ func pod(name string) *v1.Pod {
 		Spec: v1.PodSpec{
 			NodeName: "test-node",
 		},
+		Status: v1.PodStatus{Phase: v1.PodRunning, Conditions: []v1.PodCondition{{
+			Type:   v1.PodReady,
+			Status: v1.ConditionTrue,
+		}}},
+	}
+}
+
+func podWithPodPhaseAndStartTime(name string, podPhase v1.PodPhase, startTime time.Time) *v1.Pod {
+	return &v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: "test-ns",
+			Labels:    labelMap,
+		},
+		Spec: v1.PodSpec{
+			NodeName: "test-node",
+		},
+		Status: v1.PodStatus{Phase: podPhase, StartTime: &metav1.Time{Time: startTime}},
+	}
+}
+
+func podWithPodReadyConditionStatusAndLastTransitionTime(name string, podConditionStatus v1.ConditionStatus, lastTransitionTime time.Time) *v1.Pod {
+	return &v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: "test-ns",
+			Labels:    labelMap,
+		},
+		Spec: v1.PodSpec{
+			NodeName: "test-node",
+		},
+		Status: v1.PodStatus{Conditions: []v1.PodCondition{{
+			Type:               v1.PodReady,
+			Status:             podConditionStatus,
+			LastTransitionTime: metav1.Time{Time: lastTransitionTime},
+		}}},
 	}
 }
 
 func Test_runWatch(t *testing.T) {
 	namespace := "test-ns"
 	type args struct {
-		kubeFn       func(t *testing.T, podWatcher Watcher)
-		collectorMap map[string]*allocation.Collector
+		collectorNotReadyGracePeriod time.Duration
+		kubeFn                       func(t *testing.T, podWatcher Watcher)
+		collectorMap                 map[string]*allocation.Collector
 	}
 	tests := []struct {
 		name string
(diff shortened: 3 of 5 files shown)


cmd/otel-allocator/internal/collector/collector.go
```diff
@@ -124,6 +127,25 @@ func (k *Watcher) runOnCollectors(store cache.Store, fn func(collectors map[stri
 		if pod.Spec.NodeName == "" {
 			continue
 		}
+
+		isPodUnhealthy := false
```
swiatekm, line 131: Can we move this check to a helper function? It's a bit more complicated now.

cmd/otel-allocator/internal/collector/collector.go
```diff
@@ -124,6 +127,25 @@ func (k *Watcher) runOnCollectors(store cache.Store, fn func(collectors map[stri
...
+		timeNow := time.Now()
+		// stop assigning targets to a non-Running pod that has lasted for a specific period
+		if pod.Status.Phase != v1.PodRunning &&
+			pod.Status.StartTime != nil &&
+			(timeNow.Sub(pod.Status.StartTime.Time) > defaultGracePeriodBeforeSkipBadCollector) {
```
swiatekm, line 136: I think you want `k.gracePeriodBeforeSkipBadCollector` here instead of the default.

cmd/otel-allocator/internal/collector/collector.go
```diff
@@ -124,6 +127,25 @@ func (k *Watcher) runOnCollectors(store cache.Store, fn func(collectors map[stri
...
+			isPodUnhealthy = true
+		}
+		// stop assigning targets to a non-Ready pod that has lasted for a specific period
+		for _, condition := range pod.Status.Conditions {
+			if condition.Type == v1.PodReady && condition.Status != v1.ConditionTrue && (timeNow.Sub(condition.LastTransitionTime.Time) > defaultGracePeriodBeforeSkipBadCollector) {
```
swiatekm, line 141: I think you want `k.gracePeriodBeforeSkipBadCollector` here instead of the default.

cmd/otel-allocator/internal/collector/collector_test.go
```diff
@@ -157,6 +194,178 @@ func Test_runWatch(t *testing.T) {
...
+	type args struct {
+		kubeFn       func(t *testing.T, podWatcher Watcher)
+		collectorMap map[string]*allocation.Collector
+	}
+	tests := []struct {
+		name string
+		args args
+		want map[string]*allocation.Collector
+	}{
+		{
```
swiatekm, line 294: We only have one test case, so we can skip the whole table-based test setup.

cmd/otel-allocator/internal/collector/collector_test.go
```diff
@@ -157,6 +194,178 @@ func Test_runWatch(t *testing.T) {
...
+			name: "pod add",
```
swiatekm, line 209: We only have one test case here too.

cmd/otel-allocator/internal/collector/collector.go
```diff
@@ -20,7 +20,8 @@ import (
 )
 
 const (
-	defaultMinUpdateInterval = time.Second * 5
+	defaultMinUpdateInterval                 = time.Second * 5
+	defaultGracePeriodBeforeSkipBadCollector = time.Second * 30
```
swiatekm, line 24: That's a wordy name. How about `defaultNotReadyGracePeriod`, and add a comment explaining what this is?

cmd/otel-allocator/internal/collector/collector.go
rice-junhaoyu: Done! Good call out!

cmd/otel-allocator/internal/collector/collector.go
rice-junhaoyu: Exactly!! A miss from me.

cmd/otel-allocator/internal/collector/collector_test.go
rice-junhaoyu: Exactly!! A miss from me.

cmd/otel-allocator/internal/collector/collector_test.go
rice-junhaoyu: Makes sense. Thanks.

cmd/otel-allocator/internal/collector/collector.go
rice-junhaoyu: Changed! Appreciate the suggestion.

.chloggen/ta_not_assign_targets_to_bad_collectors.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
```
swiatekm, line 2: I think it should be an enhancement. The current behaviour, while not ideal, isn't unintended.

cmd/otel-allocator/internal/collector/collector_test.go
```diff
@@ -157,6 +194,168 @@ func Test_runWatch(t *testing.T) {
 	}
 }
 
+func Test_gracePeriodWithNonRunningPodPhase(t *testing.T) {
+	namespace := "test-ns"
+	type args struct {
+		collectorMap map[string]*allocation.Collector
+	}
+	test := struct {
```
swiatekm, line 202: I don't understand why this slice needs to exist. If there's only one test case, we can just run it directly, without test case structs or subtests.

swiatekm (commented):
I have some minor comments, but overall this looks good. 

@jaronoff97 do you think we can just merge this as is, or do we need a gating mechanism for users to turn this change off if it causes issues in their deployments?

jaronoff97:
hm... I think the interval should be configurable and if it's set to 0 or not set this feature should be disabled. I definitely know of cases where we don't want this at all (for users where some amount of downtime is acceptable, but target shifting in sub minute can cause issues). I don't think we need a feature gate though and i think it's probably fine to default it to 30s, I just want a way to disable it in case we have to.

.chloggen/ta_not_assign_targets_to_bad_collectors.yaml
rice-junhaoyu: Will fix in next revision.

cmd/otel-allocator/internal/collector/collector_test.go
rice-junhaoyu: Sorry I am a bit lost here.... I think I've removed the slice? This is just a single construct?

rice-junhaoyu: Ah, ok... I think I got you now. I can fix.

rice-junhaoyu:
@jaronoff97 I do agree that a user should have a workaround to disable. But without a feature gate, is it feasible?

Maybe a trade-off we can do is to increase from 30s to 1min or some.

Or maybe we should have a rate limiter for the `number of times of target reallocation` that can be performed within a specific time range.

jaronoff97:
Yes, if you add this to the configuration of the target allocator and then set it / disable in the CRD then it will do what we expect. Im on mobile but can explain more if that's unclear.

cmd/otel-allocator/internal/collector/collector_test.go
rice-junhaoyu: Hey Mikolaj, since we are adding a `new Config: CollectorNotReadyGracePeriod` on TA, I will have more than 2 test cases per scenario. So I end up adding the test case struct and slice back!

swiatekm:
FYI, I think it's fine to merge this PR as is if you:

1. Add the value to target allocator configuration.
2. Set it to 0 by default.
3. Change the logic so it doesn't do anything if the value is 0.

Then we can take care of the CRD changes in a separate PR.

rice-junhaoyu:
Did I get you correctly?

> Add the value to target allocator configuration.

I think I added `CollectorNotReadyGracePeriod` to `/internal/config/config.go` already

> Set it to 0 by default.

Yes! In `CreateDefaultConfig()`!

> Change the logic so it doesn't do anything if the value is 0.

In `isPodUnhealthy()` in `/internal/collector/collector.go`, I did make it a no-op if CollectorNotReadyGracePeriod is set to 0 (the default).

swiatekm:
@rice-junhaoyu that's my bad, I hadn't noticed that you'd updated the PR while I was writing the comment.

swiatekm (approved):

cmd/otel-allocator/internal/collector/collector.go
```diff
@@ -124,6 +127,12 @@ func (k *Watcher) runOnCollectors(store cache.Store, fn func(collectors map[stri
 		if pod.Spec.NodeName == "" {
 			continue
 		}
+
+		// pod healthiness check will always be disabled if CollectorNotReadyGracePeriod is set to 0 * time.Second
+		if isPodUnhealthy(pod, k.collectorNotReadyGracePeriod) {
+			continue
+		}
```
jaronoff97, line 134: let's change this to be a method of the watcher i.e. `(k *Watcher) func isPodUnhealthy` ...

cmd/otel-allocator/internal/collector/collector.go
```diff
@@ -133,3 +142,25 @@ func (k *Watcher) runOnCollectors(store cache.Store, fn func(collectors map[stri
 func (k *Watcher) Close() {
 	close(k.close)
 }
+
+func isPodUnhealthy(pod *v1.Pod, collectorNotReadyGracePeriod time.Duration) bool {
+	if collectorNotReadyGracePeriod == config.DefaultCollectorNotReadyGracePeriod {
```
jaronoff97, line 147: Suggestion: check that `collectorNotReadyGracePeriod` is equal to 0 here rather than relying on the default value. In the future we may want to change the default in which case we'd need to change this condition.

cmd/otel-allocator/internal/config/config.go
```diff
@@ -29,26 +29,28 @@ import (
...
+	RootLogger                   logr.Logger           `yaml:"-"`
+	CollectorSelector            *metav1.LabelSelector `yaml:"collector_selector,omitempty"`
+	CollectorNamespace           string                `yaml:"collector_namespace,omitempty"`
+	PromConfig                   *promconfig.Config    `yaml:"config"`
+	AllocationStrategy           string                `yaml:"allocation_strategy,omitempty"`
+	AllocationFallbackStrategy   string                `yaml:"allocation_fallback_strategy,omitempty"`
+	FilterStrategy               string                `yaml:"filter_strategy,omitempty"`
+	PrometheusCR                 PrometheusCRConfig    `yaml:"prometheus_cr,omitempty"`
+	HTTPS                        HTTPSServerConfig     `yaml:"https,omitempty"`
+	CollectorNotReadyGracePeriod time.Duration         `yaml:"collector_not_ready_grace_period,omitempty"`
```
jaronoff97, line 53: i think this name is a bit too long IMO, how about `CollectorUnhealthyTimeout` or something like that? cc @swiatekm

jaronoff97 (commented):
A few thoughts, but this is looking great! Thank you!

cmd/otel-allocator/internal/config/config.go
jaronoff97: `CollectorUnhealthyPeriod`?

cmd/otel-allocator/internal/collector/collector.go
rice-junhaoyu: Sure thing, will fix in next revision. I was thinking I shouldn't add such a `common method` to the class `Watcher`.

cmd/otel-allocator/internal/config/config.go
rice-junhaoyu: No strong preference. But please let me know the final decision!

jaronoff97: mm okay, lets leave it as is then.

rice-junhaoyu:
Comments all fixed and published a new revision.

I will follow-up and start a new pull request for the CRD change once this is complete.

Appreciate all the help!

jaronoff97 (approved):

PR #3812, opened 2025-03-17
Comment out autoinstrumentation-php in versions.txt

jvoravong:
PHP was recently added to [versions.txt](https://github.com/open-telemetry/opentelemetry-operator/blob/main/versions.txt#L37), which is great for progressing towards expanding language support. However, I’d like to keep versions.txt focused on features that are currently available in the operator.

Our downstream project relies on versions.txt for CI/CD, ensuring that only functional and accessible Docker images are referenced. Since PHP support isn’t ready yet, this is causing CI/CD failures due to the lack of a defined PHP image and the operator not having support for PHP implemented.

To avoid breaking CI/CD, this PR comments out the PHP version until official support is available.

### DIFF
--- a/versions.txt
+++ b/versions.txt
@@ -34,7 +34,7 @@ autoinstrumentation-go=v0.19.0-alpha
 
 # Represents the current release of PHP instrumentation.
 # Should match autoinstrumentation/php/version.txt
-autoinstrumentation-PHP=1.1.0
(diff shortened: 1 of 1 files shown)


jvoravong:
@SergeyKleyman would you see any issues with this? I don't want to break anything you might know about.

jaronoff97 (approved):

Issue #3819, opened 2025-03-19
Define a managed CRD skeleton


atoulme:
### Describe the issue you're reporting

Create a new CRD skeleton to be placed under the operator.

PR #3830, opened 2025-03-20
[docs] Add RFC for managed CR

atoulme:
**Description:**
RFC for managed operator custom resource.

See #3819

### DIFF
--- a/docs/rfcs/managed.md
+++ b/docs/rfcs/managed.md
@@ -80,7 +80,7 @@ All standard CRDs for instrumentations are all enabled by default for all namesp
 
 The customer upon installation will receive all logs from Kubernetes, scraped with the filelog receiver from the file system.
 
-## configuration
+## Configuration
 
 The customer can enable exporting signals individually: metrics, logs, traces and profiles.
 
--- a/docs/rfcs/managed.md
+++ b/docs/rfcs/managed.md
@@ -89,7 +89,7 @@ The user can enable exporting signals individually: metrics, logs, traces and pr
 ```go
 import "go.opentelemetry.io/collector/exporter/otlpexporter"
 
-type ManagedCustomResource struct {
+type ClusterObservability struct {
 	# List of signals supported: `logs`, `metrics`, `traces`, `profiles`
 	Signals        []string `yaml:"signals"`
 	# OTLP exporter configuration
@@ -100,18 +100,18 @@ type ManagedCustomResource struct {
 ```yaml
 ---
 apiVersion: apiextensions.k8s.io/v1
-kind: CustomResourceDefinition
+kind: ClusterObservability
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.32.0
-  name: managed.opentelemetry.io
+  name: clusterobservability.opentelemetry.io
 spec:
   group: opentelemetry.io
   names:
-    kind: Managed
-    listKind: ManagedList
-    plural: manageds
-    singular: managed
+    kind: ClusterObservability
+    listKind: ClusterObservabilityList
+    plural: co11ys
+    singular: co11y
   scope: Cluster
   versions:
   - additionalPrinterColumns:
--- a/docs/rfcs/managed.md
+++ b/docs/rfcs/managed.md
@@ -40,6 +40,12 @@ follow the traditional route of setting up collectors, target allocators, instru
 The solution exports data to a single endpoint using the OTLP gRPC protocol. Customers may choose to deploy a gateway
 to transform, filter, redact data before sending it to storage.
 
+### Context-aware
+
+Best practices differ between platforms, cloud providers, available endpoints, node operating system and so on.
+
+The solution will enable features based on those parameters. For example, it will deploy Openshift-specific features when deployed on the Openshift Cloud Platform.
+
 ## Use cases for proposal
 
 ### Initial installation
--- a/docs/rfcs/managed.md
+++ b/docs/rfcs/managed.md
@@ -78,7 +78,7 @@ All standard CRDs for instrumentations are enabled by default for all namespaces
 
 ### Node logs
 
-The customer upon installation will receive logs from the node operating system, as well as pod logs from Kubernetes, scraped with the filelog receiver from the file system.
+The user upon installation will receive logs from the node operating system, as well as pod logs from Kubernetes, scraped with the filelog receiver from the file system.
 
 ## Configuration
 
--- a/docs/rfcs/managed.md
+++ b/docs/rfcs/managed.md
@@ -37,7 +37,7 @@ follow the traditional route of setting up collectors, target allocators, instru
 
 ### OTLP, single endpoint support
 
-The solution exports data to a single endpoint using the OTLP gRPC protocol. Customers may choose to deploy a gateway
+The solution exports data to a single endpoint using the OTLP protocol. Customers may choose to deploy a gateway
 to transform, filter, redact data before sending it to storage.
 
 ### Context-aware
--- a/docs/rfcs/managed.md
+++ b/docs/rfcs/managed.md
@@ -35,6 +35,11 @@ We don't allow to toggle features in and out specifically because we want to avo
 If customers want to increase their level of control of the feature set, they can remove the managed custom resource and instead
 follow the traditional route of setting up collectors, target allocators, instrumentation custom resources as is possible today.
 
+### OTLP, single endpoint support
+
+The solution exports data to a single endpoint using the OTLP gRPC protocol. Customers may choose to deploy a gateway
+to transform, filter, redact data before sending it to storage.
+
 ## Use cases for proposal
 
 ### Initial installation
@@ -53,31 +58,25 @@ and are guaranteed to work through functional test coverage.
 The customer can cleanly delete the custom resource, which triggers the deletion of all OpenTelemetry assets under
 management by the operator. No stragglers are left.
 
-### Kubernetes metrics
+### Infrastructure monitoring
 
-The customer upon installation of the custom resource can see Kubernetes cluster metrics as reported by the k8sclusterreceiver with default configuration.
+The customer upon installation of the custom resource can see Kubernetes cluster metrics, and kubeletstats metrics.
 
-### Kubeletstats metrics
+The customer can follow events and receive Kubernetes entity data.
 
-The customer upon installation receives metrics from all Kubeletstats metrics from the default configuration of the kubeletstats receiver.
+The customer upon installation receives metrics from hostmetrics receiver with its default configuration.  Note the metrics map to the host, meaning the collector will have access to host volumes to scrape their utilization.
 
-### Automatic instrumentation
+### Workload monitoring
 
 All standard CRDs for instrumentations are all enabled by default for all namespaces.
-
-### Node host metrics
-
-The customer upon installation receives metrics from hostmetrics receiver with its default configuration.
-
-Note the metrics map to the host, meaning the collector will have access to host volumes to scrape their utilization.
 
 ### Node logs
 
 The customer upon installation will receive all logs from Kubernetes, scraped with the filelog receiver from the file system.
 
-### Kubernetes entities
+## configuration
 
-When entities become viable, the customer will receive Kubernetes entities as reported by the k8sclusterreceiver.
+The customer can enable exporting signals individually: metrics, logs, traces and profiles.
 
 ## Struct Design
 
@@ -85,6 +84,9 @@ When entities become viable, the customer will receive Kubernetes entities as re
 import "go.opentelemetry.io/collector/exporter/otlpexporter"
 
 type ManagedCustomResource struct {
+	# List of signals supported: `logs`, `metrics`, `traces`, `profiles`
+	Signals        []string `yaml:"signals"`
+	# OTLP exporter configuration
 	ExporterConfig otlpexporter.Config `yaml:"exporter"`
 }
 ```
--- a/docs/rfcs/managed.md
+++ b/docs/rfcs/managed.md
@@ -82,7 +82,7 @@ The user upon installation will receive logs from the node operating system, as 
 
 ## Configuration
 
-The customer can enable exporting signals individually: metrics, logs, traces and profiles.
+The user can enable exporting signals individually: metrics, logs, traces and profiles.
 
 ## Struct Design
 
--- a/docs/rfcs/managed.md
+++ b/docs/rfcs/managed.md
@@ -18,7 +18,7 @@ on how to best instrument all resources in the cluster on behalf of the administ
 OpenTelemetry standards and semantic conventions.
 
 The operator project offers custom resources (CR) to allow Kubernetes users/administrators to define OpenTelemetry components
-in their Kubernetes environments. Customers however do not want to manage components themselves, especially if they need
+in their Kubernetes environments. Users however do not want to manage components themselves, especially if they need
 to think about component versioning and upgrade patterns, and would rather have an out-of-the-box experience that 
 requires little insight and gets them to a working state effortlessly.
 
@@ -32,12 +32,12 @@ We mean to offer a specific functional coverage of a specific use case that offe
 
 We don't allow toggling features specifically because we want to avoid increasing the need for functional test coverage.
 
-If customers want to increase their level of control of the feature set, they can remove the managed custom resource and instead
+If Users want to increase their level of control of the feature set, they can remove the managed custom resource and instead
 follow the traditional route of setting up collectors, target allocators, instrumentation custom resources as is possible today.
 
 ### OTLP, single endpoint support
 
-The solution exports data to a single endpoint using the OTLP protocol. Customers may choose to deploy a gateway
+The solution exports data to a single endpoint using the OTLP protocol. Users may choose to deploy a gateway
 to transform, filter, redact data before sending it to storage.
 
 ### Context-aware
@@ -61,14 +61,14 @@ and are guaranteed to work through functional test coverage.
 
 ### Deletion and reinstallation
 
-The customer can cleanly delete the custom resource, which triggers the deletion of all OpenTelemetry assets under
+The user can cleanly delete the custom resource, which triggers the deletion of all OpenTelemetry assets under
 management by the operator. No stragglers are left.
 
 ### Infrastructure monitoring
 
-The customer upon installation of the custom resource can see Kubernetes cluster metrics, and kubeletstats metrics.
+The user upon installation of the custom resource can see Kubernetes cluster metrics, and kubeletstats metrics.
 
-The customer can follow events and receive Kubernetes entity data.
+The user can follow events and receive Kubernetes entity data.
 
 The user upon installation receives metrics from hostmetrics receiver with its default configuration. Note the metrics map to the host, meaning the collector will have access to host volumes to scrape their utilization.
 
--- a/docs/rfcs/managed.md
+++ b/docs/rfcs/managed.md
@@ -1,0 +1,142 @@
+# TITLE
+
+**Status:** *Draft*
+
+**Author:** Antoine Toulme (@atoulme)
+
+**Date:** 2025-03-20
+
+## Objective
+
+Condense all best practices and deployment patterns of OpenTelemetry components into a single custom resource
+that covers the feature set of the OpenTelemetry Helm chart.
+
+## Summary
+This request for comment aims to establish a new approach to management of resources in Kubernetes clusters,
+specifically in Openshift environments, with the explicit goal of taking over the whole cluster, and making decisions
+on how to best instrument all resources in the cluster on behalf of the administrator based on best practices,
+OpenTelemetry standards and semantic conventions.
+
+The operator project offers custom resources (CR) to allow Kubernetes operators to define OpenTelemetry components
+in their Kubernetes environments. Customers however do not want to manage components themselves, especially if they need
+to think about component versioning and upgrade patterns, and would rather have an out-of-the-box experience that 
+requires little insight and gets them to a working state effortlessly.
+
+## Goals and non-goals
+
+### Give a comprehensive, out-of the box experience of OpenTelemetry for a first-time user
+The goal is to delight first-time customers by giving them a complete view of what is possible with OpenTelemetry.
+
+### Do not offer a way to turn on and off specific features
+We mean to offer a specific functional coverage of a specific use case that offers best practices of OpenTelemetry.
+
+We don't allow to toggle features in and out specifically because we want to avoid increasing the need for functional test coverage.
+
+If customers want to increase their level of control of the feature set, they can remove the managed custom resource and instead
+follow the traditional route of setting up collectors, target allocators, instrumentation custom resources as is possible today.
+
+## Use cases for proposal
+
+### Initial installation
+
+The customer wants to install the operator and immediately get value from the installation.
+
+Upon installation, they install the custom resource with information on an OTLP endpoint where they can direct all data.
+
+### Upgrade
+
+The customer wants to upgrade their operator. All assets are upgraded out of the box in concert as part of the upgrade,
+and are guaranteed to work through functional test coverage.
+
+### Deletion and reinstallation
+
+The customer can cleanly delete the custom resource, which triggers the deletion of all OpenTelemetry assets under
+management by the operator. No stragglers are left.
+
+### Kubernetes metrics
+
+The customer upon installation of the custom resource can see Kubernetes cluster metrics as reported by the k8sclusterreceiver with default configuration.
+
+### Kubeletstats metrics
+
+The customer upon installation receives metrics from all Kubeletstats metrics from the default configuration of the kubeletstats receiver.
+
+### Automatic instrumentation
+
+All standard CRDs for instrumentations are all enabled by default for all namespaces.
+
+### Node host metrics
+
+The customer upon installation receives metrics from hostmetrics receiver with its default configuration.
+
+Note the metrics map to the host, meaning the collector will have access to host volumes to scrape their utilization.
+
+### Node logs
+
+The customer upon installation will receive all logs from Kubernetes, scraped with the filelog receiver from the file system.
+
+### Kubernetes entities
+
+When entities become viable, the customer will receive Kubernetes entities as reported by the k8sclusterreceiver.
+
+## Struct Design
+
+```go
+import "go.opentelemetry.io/collector/exporter/otlpexporter"
+
+type ManagedCustomResource struct {
+	ExporterConfig otlpexporter.Config `yaml:"exporter"`
+}
+```
+
+```yaml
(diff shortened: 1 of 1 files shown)


docs/rfcs/managed.md
```diff
@@ -0,0 +1,142 @@
...
+**Date:** 2025-03-20
+
+## Objective
+
+Condense all best practices and deployment patterns of OpenTelemetry components into a single custom resource
+that covers the feature set of the OpenTelemetry Helm chart.
+
+## Summary
+This request for comment aims to establish a new approach to management of resources in Kubernetes clusters,
+specifically in Openshift environments, with the explicit goal of taking over the whole cluster, and making decisions
```
iblancasa, line 16: Why specifically OpenShift?

docs/rfcs/managed.md
```diff
@@ -0,0 +1,142 @@
...
+
+Note the metrics map to the host, meaning the collector will have access to host volumes to scrape their utilization.
+
+### Node logs
+
+The customer upon installation will receive all logs from Kubernetes, scraped with the filelog receiver from the file system.
+
+### Kubernetes entities
+
+When entities become viable, the customer will receive Kubernetes entities as reported by the k8sclusterreceiver.
```
iblancasa, line 80: I'm not sure about this. It makes sense to me to create a CRD that manages OTEL "common patterns" for you". And those good practices can include different features (depending on the pattern). This part of the proposal sounds too specific for me in those terms.

docs/rfcs/managed.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Struct Design
+
+```go
+import "go.opentelemetry.io/collector/exporter/otlpexporter"
+
+type ManagedCustomResource struct {
+	ExporterConfig otlpexporter.Config `yaml:"exporter"`
```
iblancasa, line 88: Would it make sense to make this a list?

docs/rfcs/managed.md
```diff
@@ -0,0 +1,142 @@
+# TITLE
```
iblancasa, line 1:
```suggestion
# Managed CRD
```

docs/rfcs/managed.md
atoulme: that is the population we specifically target here. Openshift has heightened security requirements and offers operators as first class experience. For users using the vanilla Kubernetes experience, the Helm chart might be a better fit.

docs/rfcs/managed.md
atoulme: OK.

docs/rfcs/managed.md
atoulme: I don't think that would make sense without additional context. Is that a parallel ingestion pattern? A load-balancer configuration with round robbin or failover? It's better to have one output, and if customers have needs to route, they can point to a collector they manage out of band to send to their backends.

docs/rfcs/managed.md
iblancasa: I would say that was not the philosophy we tried to follow for the operator. We created some OpenShift specific features but just because vanilla Kubernetes could not do those specific stuff. In this case, I don't see why the non-OpenShift environments should be excluded.

docs/rfcs/managed.md
```diff
@@ -0,0 +1,142 @@
...
+with the explicit goal of taking over the whole cluster, and making decisions
+on how to best instrument all resources in the cluster on behalf of the administrator based on best practices,
+OpenTelemetry standards and semantic conventions.
+
+The operator project offers custom resources (CR) to allow Kubernetes operators to define OpenTelemetry components
+in their Kubernetes environments. Customers however do not want to manage components themselves, especially if they need
+to think about component versioning and upgrade patterns, and would rather have an out-of-the-box experience that 
+requires little insight and gets them to a working state effortlessly.
+
+## Goals and non-goals
```
pavolloffay, line 25: is the scope of the CR to enable infrastructure monitoring or workload monitoring?

docs/rfcs/managed.md
```diff
@@ -0,0 +1,142 @@
...
+### Initial installation
+
+The customer wants to install the operator and immediately get value from the installation.
+
+Upon installation, they install the custom resource with information on an OTLP endpoint where they can direct all data.
+
+### Upgrade
+
+The customer wants to upgrade their operator. All assets are upgraded out of the box in concert as part of the upgrade,
+and are guaranteed to work through functional test coverage.
```
jaronoff97, line 49: We'll probably need to care more about using the migration functionality which has some dragons... is that the expectation?

docs/rfcs/managed.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Deletion and reinstallation
+
+The customer can cleanly delete the custom resource, which triggers the deletion of all OpenTelemetry assets under
+management by the operator. No stragglers are left.
+
+### Kubernetes metrics
+
+The customer upon installation of the custom resource can see Kubernetes cluster metrics as reported by the k8sclusterreceiver with default configuration.
```
jaronoff97, line 58: Will we also want to load in the k8sobjectsreceiver to get events?

docs/rfcs/managed.md
```diff
@@ -0,0 +1,142 @@
...
+to think about component versioning and upgrade patterns, and would rather have an out-of-the-box experience that 
+requires little insight and gets them to a working state effortlessly.
+
+## Goals and non-goals
+
+### Give a comprehensive, out-of the box experience of OpenTelemetry for a first-time user
+The goal is to delight first-time customers by giving them a complete view of what is possible with OpenTelemetry.
+
+### Do not offer a way to turn on and off specific features
+We mean to offer a specific functional coverage of a specific use case that offers best practices of OpenTelemetry.
```
jaronoff97, line 31: I do think it would be worth exposing something so people can declare resourcing limits, even if it's just a "small, medium, large"

docs/rfcs/managed.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Kubernetes metrics
+
+The customer upon installation of the custom resource can see Kubernetes cluster metrics as reported by the k8sclusterreceiver with default configuration.
+
+### Kubeletstats metrics
+
+The customer upon installation receives metrics from all Kubeletstats metrics from the default configuration of the kubeletstats receiver.
+
+### Automatic instrumentation
```
jaronoff97, line 64: for instrumentation, how could we best communicate to users how to instrument their workloads post installation?

docs/rfcs/managed.md
```diff
@@ -0,0 +1,142 @@
...
+
+When entities become viable, the customer will receive Kubernetes entities as reported by the k8sclusterreceiver.
+
+## Struct Design
+
+```go
+import "go.opentelemetry.io/collector/exporter/otlpexporter"
+
+type ManagedCustomResource struct {
+	ExporterConfig otlpexporter.Config `yaml:"exporter"`
```
jaronoff97, line 88: I really dig the idea of pulling in the v1 otlpexporter config and would love for us to do more of this now.

docs/rfcs/managed.md
atoulme: both.

docs/rfcs/managed.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Do not offer a way to turn on and off specific features
+We mean to offer a specific functional coverage of a specific use case that offers best practices of OpenTelemetry.
+
+We don't allow to toggle features in and out specifically because we want to avoid increasing the need for functional test coverage.
+
+If customers want to increase their level of control of the feature set, they can remove the managed custom resource and instead
+follow the traditional route of setting up collectors, target allocators, instrumentation custom resources as is possible today.
+
+## Use cases for proposal
```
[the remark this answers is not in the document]
atoulme: I mean, yes, that's likely the result, but you're jumping into the implementation.

docs/rfcs/managed.md
```diff
@@ -0,0 +1,142 @@
...
+### Give a comprehensive, out-of the box experience of OpenTelemetry for a first-time user
+The goal is to delight first-time customers by giving them a complete view of what is possible with OpenTelemetry.
+
+### Do not offer a way to turn on and off specific features
+We mean to offer a specific functional coverage of a specific use case that offers best practices of OpenTelemetry.
+
+We don't allow to toggle features in and out specifically because we want to avoid increasing the need for functional test coverage.
+
+If customers want to increase their level of control of the feature set, they can remove the managed custom resource and instead
+follow the traditional route of setting up collectors, target allocators, instrumentation custom resources as is possible today.
```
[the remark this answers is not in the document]
atoulme: Sounds like a nice enhancement, but I'm not sure I would put it in scope. `kubectl describe` all your artifacts if you like to save them as k8s manifests and tweak them by hand if you have to. Anything fancier can wait a bit imo.

docs/rfcs/managed.md
atoulme: yes, we can add that. both k8sobjects and k8sevents receivers.

docs/rfcs/managed.md
atoulme: start with one size, and expand imo, but yes, that might be a way to go.

docs/rfcs/managed.md
atoulme: current approach still applies, and we still need to have users to annotate their workloads so they get picked up.

docs/rfcs/managed.md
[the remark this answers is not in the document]
jaronoff97: fair, i'm excited and curious :D

docs/rfcs/managed.md
jaronoff97: makes sense, im just wondering if that should be in the Status section of the CRD or maybe just in the docs.

docs/rfcs/managed.md
```diff
@@ -0,0 +1,142 @@
...
+Condense all best practices and deployment patterns of OpenTelemetry components into a single custom resource
+that covers the feature set of the OpenTelemetry Helm chart.
+
+## Summary
+This request for comment aims to establish a new approach to management of resources in Kubernetes clusters,
+with the explicit goal of taking over the whole cluster, and making decisions
+on how to best instrument all resources in the cluster on behalf of the administrator based on best practices,
+OpenTelemetry standards and semantic conventions.
+
+The operator project offers custom resources (CR) to allow Kubernetes operators to define OpenTelemetry components
```
yuriolisa, line 20: I'd change from Kubernetes operators to Kubernetes users/administrators to avoid confusion with the Kubernetes operators concept.

docs/rfcs/managed.md
yuriolisa: I'd suggest that you cover an eventual API version upgrade here. Do you think it makes sense?

docs/rfcs/managed.md
atoulme: Interesting, we can discuss this in more detail. I guess I am not doing a good job here because I went into how we get the data, not _why_. I will work on reformatting as a use case - and then based of that, we can have an implementation discussion on what receivers to bring in.

docs/rfcs/managed.md
atoulme:
```suggestion
The operator project offers custom resources (CR) to allow Kubernetes users/administrators to define OpenTelemetry components
```

atoulme:
@yuriolisa 
> I'd suggest that you cover an eventual API version upgrade here. Do you think it makes sense?

Please expand, do you mean you want semver of upgrades? With explicit version changes on new features?

docs/rfcs/managed.md
atoulme: heh. I would apply a strict versioning here where an update means uninstalling the old CR and installing the new CR. No sharing of resources. No easy path. The only easy button is that the uninstall wipes the slate clean and allows to install the new setup. I would even encourage running multiple CRs of different versions concurrently rather than do a smooth migration.

In any case, I would first obsess with landing the user experience we're going for here rather than talk about upgrades.

docs/rfcs/managed.md
```diff
@@ -0,0 +1,142 @@
...
+
+**Status:** *Draft*
+
+**Author:** Antoine Toulme (@atoulme)
+
+**Date:** 2025-03-20
+
+## Objective
+
+Condense all best practices and deployment patterns of OpenTelemetry components into a single custom resource
```
pavolloffay, line 11: How does the proposal handle customizations done by the operator providers? For instance on OpenShift most likely there will be need to add a custom (OCP) related components.

docs/rfcs/managed.md
```diff
@@ -0,0 +1,142 @@
...
+
+```yaml
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.32.0
+  name: managed.opentelemetry.io
+spec:
```
pavolloffay, line 100: Can users enable/disable log or metrics collections if they are not interested in it?

atoulme: sure, sounds good.

docs/rfcs/managed.md
atoulme: done

docs/rfcs/managed.md
atoulme: done, added some config

docs/rfcs/managed.md
atoulme: It should pick the best approach based on the environment. So on Openshift it would behave differently out of the box, installing those additional components.

atoulme: added something to that effect.

docs/rfcs/managed.md
```diff
@@ -0,0 +1,150 @@
...
+We mean to offer a specific functional coverage of a specific use case that offers best practices of OpenTelemetry.
+
+We don't allow to toggle features in and out specifically because we want to avoid increasing the need for functional test coverage.
+
+If customers want to increase their level of control of the feature set, they can remove the managed custom resource and instead
+follow the traditional route of setting up collectors, target allocators, instrumentation custom resources as is possible today.
+
+### OTLP, single endpoint support
+
+The solution exports data to a single endpoint using the OTLP gRPC protocol. Customers may choose to deploy a gateway
```
pavolloffay, line 40: Shall we be more relaxed here? Not all vendors support gRPC. Based on my old notes more vendors supported HTTP (probbaly due to better ingress/LB support).

docs/rfcs/managed.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Workload monitoring
+
+All standard CRDs for instrumentations are all enabled by default for all namespaces.
+
+### Node logs
+
+The customer upon installation will receive all logs from Kubernetes, scraped with the filelog receiver from the file system.
+
+## configuration
```
pavolloffay, line 83: Configuration

docs/rfcs/managed.md
```diff
@@ -0,0 +1,150 @@
...
+**Status:** *Draft*
+
+**Author:** Antoine Toulme (@atoulme)
+
+**Date:** 2025-03-20
+
+## Objective
+
+Condense all best practices and deployment patterns of OpenTelemetry components into a single custom resource
+that covers the feature set of the OpenTelemetry Helm chart.
```
pavolloffay, line 12: I would rephrase this sentence, I am not sure what the helm chart supports and on top of that there are multiple helm charts https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts

pavolloffay (approved):

docs/rfcs/managed.md
atoulme:
```suggestion
## Configuration
```

docs/rfcs/managed.md
atoulme:
```suggestion
The solution exports data to a single endpoint using the OTLP protocol. Customers may choose to deploy a gateway
```

docs/rfcs/managed.md
atoulme:
```suggestion
that covers the use cases fulfilled by the OpenTelemetry helm charts.
```

atoulme: done :)

docs/rfcs/managed.md
atoulme: I don't know for sure.

iblancasa (approved):

docs/rfcs/managed.md
```diff
@@ -0,0 +1,150 @@
...
+
+The customer upon installation receives metrics from hostmetrics receiver with its default configuration.  Note the metrics map to the host, meaning the collector will have access to host volumes to scrape their utilization.
+
+### Workload monitoring
+
+All standard CRDs for instrumentations are all enabled by default for all namespaces.
+
+### Node logs
+
+The customer upon installation will receive all logs from Kubernetes, scraped with the filelog receiver from the file system.
```
swiatekm, line 81: Do you actually mean Pod container logs here? "Node logs" makes me think of logs from the Node host. Or is it both?

swiatekm (commented):
LGTM, modulo one question.

docs/rfcs/managed.md
atoulme: I meant both. Typically we scrape logs of pods from the node host system. I can clarify a bit.

atoulme:
```suggestion
### Node logs

The customer upon installation will receive logs from the node operating system, as well as pod logs from Kubernetes, scraped with the filelog receiver from the file system.
```

docs/rfcs/managed.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Goals and non-goals
+
+### Give a comprehensive, out-of the box experience of OpenTelemetry for a first-time user
+The goal is to delight first-time customers by giving them a complete view of what is possible with OpenTelemetry.
+
+### Do not offer a way to turn on and off specific features
+We mean to offer a specific functional coverage of a specific use case that offers best practices of OpenTelemetry.
+
+We don't allow to toggle features in and out specifically because we want to avoid increasing the need for functional test coverage.
```
jaronoff97, line 33:
```suggestion
We don't allow toggling features specifically because we want to avoid increasing the need for functional test coverage.
```

jaronoff97: I'm not sure I love this reason to not allow toggling. I think it would be more accurate to say "this CR is meant as an introduction to the features OpenTelemetry has and is not meant to fulfill every use case." This then flows much better into the next paragraph.

docs/rfcs/managed.md
```diff
@@ -0,0 +1,150 @@
...
+### OTLP, single endpoint support
+
+The solution exports data to a single endpoint using the OTLP protocol. Customers may choose to deploy a gateway
+to transform, filter, redact data before sending it to storage.
+
+### Context-aware
+
+Best practices differ between platforms, cloud providers, available endpoints, node operating system and so on.
+
+The solution will enable features based on those parameters. For example, it will deploy Openshift-specific features when deployed on the Openshift Cloud Platform.
```
jaronoff97, line 47: i really like this :bow:

docs/rfcs/managed.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Use cases for proposal
+
+### Initial installation
+
+The customer wants to install the operator and immediately get value from the installation.
```
jaronoff97, line 53:
```suggestion
The user wants to install the operator and immediately get value from the installation.
```

docs/rfcs/managed.md
```diff
@@ -0,0 +1,150 @@
...
+The customer can cleanly delete the custom resource, which triggers the deletion of all OpenTelemetry assets under
+management by the operator. No stragglers are left.
+
+### Infrastructure monitoring
+
+The customer upon installation of the custom resource can see Kubernetes cluster metrics, and kubeletstats metrics.
+
+The customer can follow events and receive Kubernetes entity data.
+
+The customer upon installation receives metrics from hostmetrics receiver with its default configuration.  Note the metrics map to the host, meaning the collector will have access to host volumes to scrape their utilization.
```
jaronoff97, line 73:
```suggestion
The user upon installation receives metrics from hostmetrics receiver with its default configuration. Note the metrics map to the host, meaning the collector will have access to host volumes to scrape their utilization.
```

docs/rfcs/managed.md
```diff
@@ -0,0 +1,150 @@
...
+### Deletion and reinstallation
+
+The customer can cleanly delete the custom resource, which triggers the deletion of all OpenTelemetry assets under
+management by the operator. No stragglers are left.
+
+### Infrastructure monitoring
+
+The customer upon installation of the custom resource can see Kubernetes cluster metrics, and kubeletstats metrics.
+
+The customer can follow events and receive Kubernetes entity data.
```
jaronoff97, line 71:
```suggestion
The user can follow events and receive Kubernetes entity data.
```

docs/rfcs/managed.md
```diff
@@ -0,0 +1,150 @@
...
+and are guaranteed to work through functional test coverage.
+
+### Deletion and reinstallation
+
+The customer can cleanly delete the custom resource, which triggers the deletion of all OpenTelemetry assets under
+management by the operator. No stragglers are left.
+
+### Infrastructure monitoring
+
+The customer upon installation of the custom resource can see Kubernetes cluster metrics, and kubeletstats metrics.
```
jaronoff97, line 69:
```suggestion
The user upon installation of the custom resource can see Kubernetes cluster metrics, and kubeletstats metrics.
```

docs/rfcs/managed.md
```diff
@@ -0,0 +1,150 @@
...
+Upon installation, they install the custom resource with information on an OTLP endpoint where they can direct all data.
+
+### Upgrade
+
+The customer wants to upgrade their operator. All assets are upgraded out of the box in concert as part of the upgrade,
+and are guaranteed to work through functional test coverage.
+
+### Deletion and reinstallation
+
+The customer can cleanly delete the custom resource, which triggers the deletion of all OpenTelemetry assets under
```
jaronoff97, line 64:
```suggestion
The user can cleanly delete the custom resource, which triggers the deletion of all OpenTelemetry assets under
```

docs/rfcs/managed.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Initial installation
+
+The customer wants to install the operator and immediately get value from the installation.
+
+Upon installation, they install the custom resource with information on an OTLP endpoint where they can direct all data.
+
+### Upgrade
+
+The customer wants to upgrade their operator. All assets are upgraded out of the box in concert as part of the upgrade,
```
jaronoff97, line 59:
```suggestion
The user wants to upgrade their operator. All assets are upgraded out of the box in concert as part of the upgrade,
```

docs/rfcs/managed.md
```diff
@@ -0,0 +1,150 @@
...
+
+The customer upon installation receives metrics from hostmetrics receiver with its default configuration.  Note the metrics map to the host, meaning the collector will have access to host volumes to scrape their utilization.
+
+### Workload monitoring
+
+All standard CRDs for instrumentations are all enabled by default for all namespaces.
+
+### Node logs
+
+The customer upon installation will receive logs from the node operating system, as well as pod logs from Kubernetes, scraped with the filelog receiver from the file system.
```
jaronoff97, line 81:
```suggestion
The user upon installation will receive logs from the node operating system, as well as pod logs from Kubernetes, scraped with the filelog receiver from the file system.
```

docs/rfcs/managed.md
```diff
@@ -0,0 +1,150 @@
...
+
+The customer upon installation of the custom resource can see Kubernetes cluster metrics, and kubeletstats metrics.
+
+The customer can follow events and receive Kubernetes entity data.
+
+The customer upon installation receives metrics from hostmetrics receiver with its default configuration.  Note the metrics map to the host, meaning the collector will have access to host volumes to scrape their utilization.
+
+### Workload monitoring
+
+All standard CRDs for instrumentations are all enabled by default for all namespaces.
```
jaronoff97, line 77:
```suggestion
All standard CRDs for instrumentations are enabled by default for all namespaces.
```

docs/rfcs/managed.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Node logs
+
+The customer upon installation will receive logs from the node operating system, as well as pod logs from Kubernetes, scraped with the filelog receiver from the file system.
+
+## Configuration
+
+The customer can enable exporting signals individually: metrics, logs, traces and profiles.
```
jaronoff97, line 85:
```suggestion
The user can enable exporting signals individually: metrics, logs, traces and profiles.
```

docs/rfcs/managed.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Struct Design
+
+```go
+import "go.opentelemetry.io/collector/exporter/otlpexporter"
+
+type ManagedCustomResource struct {
```
jaronoff97, line 92: I don't love the name Managed here because we have that as a field on the collector already. I would prefer something like `MonitoringStack` or `ClusterMonitor` or `ClusterObservability`. Something that makes it clear that this resource is deploying a way to monitor the entirety of the user's cluster.

atoulme: Naming is hard. I thought it over, `ClusterObservability` I guess works

docs/rfcs/managed.md
jaronoff97: @atoulme missed this one, i may have missed more, i would do a find and replace to be sure you got everything.

jaronoff97 (commented):
I think all that's left from my review is s/customer/user/g and change the name to use ClusterObservability, otherwise this looks g2g

docs/rfcs/managed.md
jaronoff97: @atoulme still some instances of customer, can you change them to user?

docs/rfcs/managed.md
```diff
@@ -0,0 +1,150 @@
...
+We mean to offer a specific functional coverage of a specific use case that offers best practices of OpenTelemetry.
+
+We don't allow toggling features specifically because we want to avoid increasing the need for functional test coverage.
+
+If customers want to increase their level of control of the feature set, they can remove the managed custom resource and instead
+follow the traditional route of setting up collectors, target allocators, instrumentation custom resources as is possible today.
+
+### OTLP, single endpoint support
+
+The solution exports data to a single endpoint using the OTLP protocol. Customers may choose to deploy a gateway
```
jaronoff97, line 40:
```suggestion
The solution exports data to a single endpoint using the OTLP protocol. Users may choose to deploy a gateway
```

by my count there are still ten instances of customer

atoulme: oops fixed.

jaronoff97 (approved):

jaronoff97:
auto merged enabled, thank you for your patience and contribution! :bow:

Issue #3782, opened 2025-03-07
Upgrade prometheus and prometheus-operator dependencies to latest


swiatekm:
### Component(s)

target allocator

### Describe the issue you're reporting

There are currently several issues blocking us from upgrading to the latest version of prometheus dependencies:

- [x] Some changes to logging that we should be able to easily make.
- [x] Prometheus-operator now generates Prometheus config a bit differently, and we'll need to clean up our handling of namespace selectors in target allocator to be compatible.
- [ ] Prometheus 0.300.0 includes a breaking change to default scrape protocols, and this change is supported by the Otel collector starting in version 0.120.0. We need to account for this in the configurations generated by target allocator and think about when and how we want to introduce the breaking change.

oldium:
Is it planned to fix the following config or should I create a separate issue for it?

```
    service:
      telemetry:
        metrics:
          level: detailed
          readers:
            - pull:
                exporter:
                  prometheus:
                    host: '0.0.0.0'
                    port: 8888
 ```

This creates an “address already in use” Collector startup error, because OTEL K8s Operator adds a default `service.telemetry.metrics.address=0.0.0.0:8888`, which conflicts with custom `service.telemetry.metrics.readers` config.

iblancasa:
> Is it planned to fix the following config or should I create a separate issue for it?
> 
> ```
>     service:
>       telemetry:
>         metrics:
>           level: detailed
>           readers:
>             - pull:
>                 exporter:
>                   prometheus:
>                     host: '0.0.0.0'
>                     port: 8888
> ```
> 
> This creates an “address already in use” Collector startup error, because OTEL K8s Operator adds a default `service.telemetry.metrics.address=0.0.0.0:8888`, which conflicts with custom `service.telemetry.metrics.readers` config.

Please, create something separate. That is not related to this ticket.

PR #3831, opened 2025-03-20
 Bump prometheus-operator to v0.80.1

swiatekm:
**Description:**

Bump prometheus-operator to 0.80.1. This also bumps the Prometheus libraries past 3.0. Aside from a migration from gokitlog to slog, this required addressing some breaking changes in Prometheus itself. With Prometheus 3.0 config unmarshaling, the default scrape protocols include `PrometheusText1_0_0`, which is only supported by Prometheus 3.0, and prometheus receiver starting at 0.120.0.

I'd rather avoid breaking compatibility with prometheus receiver <0.120.x, so I set our own default. This required hacking the config unmarshalling logic in target allocator, because we need to set the default before Prometheus' own custom unmarshalling logic gets to it. 

- Resolves: #3782


iblancasa (approved):

cmd/otel-allocator/internal/config/config.go
```diff
@@ -38,6 +40,17 @@ const (
 	DefaultCollectorNotReadyGracePeriod                = 0 * time.Second
 )
 
+// By default, scrape protocols include PrometheusText1_0_0, which only Prometheus >=3.0 supports.
+// Manually exclude this protocol until several versions of the Otel Collector support it.
```
jaronoff97, line 44: should we specify a version here and make an issue for updating this?

cmd/otel-allocator/main.go
```diff
@@ -93,7 +92,7 @@ func main() {
 		setupLog.Error(err, "Unable to register metrics for Prometheus service discovery")
 		os.Exit(1)
 	}
-	discoveryManager = discovery.NewManager(discoveryCtx, gokitlog.NewNopLogger(), prometheus.DefaultRegisterer, sdMetrics)
+	discoveryManager = discovery.NewManager(discoveryCtx, config.NopLogger, prometheus.DefaultRegisterer, sdMetrics)
```
jaronoff97, line 95: IMO we should capture these logs instead of using the Nop logger, maybe put them under debug verbosity?

jaronoff97 (commented):
two minor thoughts, otherwise looks good.

cmd/otel-allocator/main.go
swiatekm: We could, but I'd do that in a separate PR. This change simply preserves current behaviour.

jaronoff97: sounds good to me.

zarbis:
I'm curious whether this PR solves the problem of being unable to set default scrape protocol for target allocator-generated configs?

cmd/otel-allocator/internal/config/config.go
swiatekm: https://github.com/open-telemetry/opentelemetry-operator/issues/3872

swiatekm:
> I'm curious whether this PR solves the problem of being unable to set default scrape protocol for target allocator-generated configs?

It doesn't. This PR doesn't touch any public API. Have you opened an issue for this problem?

jaronoff97 (approved):

Issue #3823, opened 2025-03-19
OpAMP Bridge Proxy OpAMP Traffic


jaronoff97:
### Component(s)

opamp

### Is your feature request related to a problem? Please describe.

It can be cumbersome and overbearing to have multiple OpAMP socket connections from N collectors in a single cluster. It's also annoying that for a user that wants to use OpAMP for their collectors they need to manually configure the extension.

### Describe the solution you'd like

The bridge should be able to serve as an OpAMP server for the opamp extension of the collectors it creates. this would make the bridge work much more like the supervisor does today. The bridge should be able to:
- [x] act as an OpAMP server and aggregate messages received
- [ ] #3837
- [ ] #3838

### Additional context

I was initially worried about scalability concerns, but with #3822 I think we'll be okay.

PR #3836, opened 2025-03-21
Introduce ability for bridge to act as an in-cluster opamp server

jaronoff97:
**Description:** Introduces a proxy server to allow the bridge to aggregate health and effective config messages from collectors inside the cluster.

- Resolves: #3823 

**Testing:** 
Unit testing

**Documentation:** n/a


cmd/operator-opamp-bridge/internal/agent/agent.go
```diff
@@ -138,6 +141,10 @@ func (agent *Agent) generateCollectorPoolHealth() (map[string]*protobufs.Compone
 			Healthy:            isPoolHealthy,
 		}
 	}
+	// TODO: Figure out how to tie this to the agent health from the proxy.
```
jaronoff97, line 144: this will be done in a follow up.

cmd/operator-opamp-bridge/internal/proxy/agent.go
```diff
@@ -0,0 +1,170 @@
...
+func (a *Agent) updateAgentDescription(newStatus *protobufs.AgentToServer) (agentDescrChanged bool) {
+	prevStatus := a.Status
+
+	if a.Status == nil {
+		// First time this Agent reports a status, remember it.
+		a.Status = newStatus
+		agentDescrChanged = true
+	} else {
+		// Not a new Agent. Update the Status.
+		a.Status.SequenceNum = newStatus.SequenceNum
```
swiatekm, line 57: This feels like it should be outside this function. It doesn't have anything to do with the description.

cmd/operator-opamp-bridge/internal/proxy/agent.go
```diff
@@ -0,0 +1,170 @@
...
+		} else {
+			// AgentDescription field is not set, which means description didn't change.
+			agentDescrChanged = false
+		}
+
+		// Update remote config status if it is included and is different from what we have.
+		if newStatus.RemoteConfigStatus != nil &&
+			!proto.Equal(a.Status.RemoteConfigStatus, newStatus.RemoteConfigStatus) {
+			a.Status.RemoteConfigStatus = newStatus.RemoteConfigStatus
+		}
```
swiatekm, line 83: This also doesn't feel like it belongs in a function which is only supposed to update the description.

cmd/operator-opamp-bridge/internal/proxy/server.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		s.logger.Error(err, "failed to get instance id")
+		response.ErrorResponse = &protobufs.ServerErrorResponse{
+			ErrorMessage: err.Error(),
+		}
+		return response
+	}
+	s.logger.V(5).Info("received message", "instance ID", instanceId)
+
+	agentUpdated := false
```
swiatekm, line 121: This should be called `agentAdded` if that's the only condition it actually tracks.

cmd/operator-opamp-bridge/internal/proxy/server.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+func (s *OpAMPProxy) onDisconnect(conn types.Connection) {
+	s.mux.Lock()
+	defer s.mux.Unlock()
+
+	for instanceId := range s.connections[conn] {
+		delete(s.agentsById, instanceId)
+	}
+	delete(s.connections, conn)
```
swiatekm, line 104: Shouldn't we also emit a notification here?

cmd/operator-opamp-bridge/internal/agent/agent.go
jaronoff97: yes and no, it's really just more in depth information for the collector pods, you can see how i'll do this in [this branch](https://github.com/open-telemetry/opentelemetry-operator/compare/main...jaronoff97:opentelemetry-operator:instance-id-correlation?expand=1#diff-77162bc58930af5dede9b3b691b114de2790b1ac42e5ab21a0f6a0528630879bR116-R157).

cmd/operator-opamp-bridge/internal/agent/agent.go
```diff
@@ -260,12 +267,35 @@ func (agent *Agent) Start() error {
...
 	agent.logger.V(3).Info("OpAMP Client started.")
 
 	return nil
 }
 
+func (agent *Agent) checkForProxyUpdates() {
+	for {
+		select {
+		case <-agent.proxy.HasUpdates():
+			agent.logger.Info("new agent connected to proxy")
```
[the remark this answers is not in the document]
jaronoff97: no, let me change this.

cmd/operator-opamp-bridge/internal/agent/agent.go
```diff
@@ -320,6 +350,11 @@ func (agent *Agent) getEffectiveConfig(ctx context.Context) (*protobufs.Effectiv
 			ContentType: "yaml",
 		}
 	}
+	for id, instance := range agent.proxy.GetConfigurations() {
+		if cfg, ok := instance.GetConfigMap().GetConfigMap()[""]; ok {
```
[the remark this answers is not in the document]
jaronoff97: unfortunately yes, this is just the signature of the OpAMP Message. https://github.com/open-telemetry/opamp-spec/blob/main/proto/opamp.proto#L748

cmd/operator-opamp-bridge/internal/proxy/agent.go
jaronoff97: yeah a lot of this is boilerplate from other server implementations. Re-organizing now.

jaronoff97: i can actually get rid of this, because i'm not supporting remote config through the bridge proxy right now. That'll be future improvements.

cmd/operator-opamp-bridge/internal/proxy/server.go
jaronoff97: this actually does track updates from the agent itself. the UpdateStatus checks if things like the description changed.

swiatekm (approved):
LGTM as far as I can tell, but this would benefit from review from someone who knows opamp better.

evan-bradley (approved):
I wasn't able to take a super close look, but from the OpAMP protocol standpoint, nothing really jumped out at me.

PR #3840, opened 2025-03-22
Add unit tests to instrumentation webhook

shuheiktgw:

I've added missing unit tests to apis/v1alpha1/instrumentation_webhook_test.go. Thanks for the review!

N/A

Do I need to create a changelog for this change? 🤔

### DIFF
--- a/apis/v1alpha1/instrumentation_webhook_test.go
+++ b/apis/v1alpha1/instrumentation_webhook_test.go
@@ -7,9 +7,11 @@ import (
 	"context"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	"github.com/open-telemetry/opentelemetry-operator/internal/config"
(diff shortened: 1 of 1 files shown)


iblancasa (approved):

apis/v1alpha1/instrumentation_webhook_test.go
```diff
@@ -18,24 +20,525 @@ import (
...
-		),
-	}.Default(context.Background(), inst)
-	assert.NoError(t, err)
-	assert.Equal(t, "java-img:1", inst.Spec.Java.Image)
-	assert.Equal(t, "nodejs-img:1", inst.Spec.NodeJS.Image)
-	assert.Equal(t, "python-img:1", inst.Spec.Python.Image)
-	assert.Equal(t, "dotnet-img:1", inst.Spec.DotNet.Image)
-	assert.Equal(t, "apache-httpd-img:1", inst.Spec.ApacheHttpd.Image)
-	assert.Equal(t, "nginx-img:1", inst.Spec.Nginx.Image)
+	tests := []struct {
```
swiatekm, line 23: I think I preferred the way this test was written previously. You're right to want to add tests for other cases, but I'd prefer these to be plain subtests with unique assertions. Then you can have a test to check the resource override, and only assert that the override applies, instead of needing to include the whole instrumentation spec as an expected return value. Does that make sense?

shuheiktgw: @swiatekm Thank you for your feedback! I've fixed the PR so could you take a look at it again when you have time?

swiatekm (approved):

PR #3696, opened 2025-02-05
Split CRD documentation into separate files

swiatekm:
**Description:**
The CRD API doc became unmanagably large with the addition of TargetAllocators, and can't even be rendered by Github's markdown viewer. Split it into separate files per CRD.

The shell code to produce this is a bit hacky, but I'd like to merge it to resolve the problem first, and refine it a bit later.

**Link to tracking Issue(s):**

- Resolves: #3681

PR #3852, opened 2025-03-25
Correct some broken links in documentation

ChrisMcD1:
**Description:**

While reading the README I noticed that several of these links were outdated. 
Some links broken  due to the fallout from https://github.com/open-telemetry/opentelemetry-operator/pull/3696, and some due to changes in `prometheus-operator` docs. In the prometheus-operator docs I chose to link to the docs site, instead of docs source. I think either could be fine, and would be fine with a pushed commit that instead targets the raw docs on Github.

- Resolves: I didn't make an issue for this

**Testing:** No testing

**Documentation:** Just updated the links

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ The operator manages:
 ## Documentation
 
 - [Compatibility & Support docs](./docs/compatibility.md)
-- [API docs](./docs/api.md)
+- [API docs](./docs/api/README.md)
 - [Offical Telemetry Operator page](https://opentelemetry.io/docs/kubernetes/operator/)
 
 ## Helm Charts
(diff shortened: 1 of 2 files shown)


jaronoff97 (approved):

ChrisMcD1:
Added [chore] label for the changelog checks. I don't think this needs a reference in the changelog, but I'm happy to be wrong on that point.

jaronoff97:
@ChrisMcD1 yep, i thought i had added the skip-changelog label, but looks like i didn't. i did now though :)

iblancasa (approved):

swiatekm (approved):

PR #3874, opened 2025-04-04
Fix default internal telemetry endpoint for collector

iblancasa:
**Description:** 

Continuation of #3744.

- Resolves: #3730

### DIFF
--- a/apis/v1beta1/collector_webhook_test.go
+++ b/apis/v1beta1/collector_webhook_test.go
@@ -162,16 +162,8 @@ func TestCollectorDefaultingWebhook(t *testing.T) {
 						const input = `{"receivers":{"otlp":{"protocols":{"grpc":{"endpoint":"0.0.0.0:4317"},"http":{"endpoint":"0.0.0.0:4318"}}}},"exporters":{"debug":null},"service":{"telemetry":{"metrics":{"readers":[{"pull":{"exporter":{"prometheus":{"host":"0.0.0.0","port":8888}}}}]}},"pipelines":{"traces":{"receivers":["otlp"],"exporters":["debug"]}}}}`
 						var cfg v1beta1.Config
 						require.NoError(t, yaml.Unmarshal([]byte(input), &cfg))
-
-						r := cfg.Service.Telemetry.Object["metrics"].(map[string]interface{})["readers"]
-						readers := r.([]interface{})
-						firstReader := readers[0].(map[string]interface{})
-						exporter := firstReader["pull"].(map[string]interface{})["exporter"].(map[string]interface{})
-						// Convert port to float64 to avoid type mismatch because how go-yaml unmarshals
-						exporter["prometheus"].(map[string]interface{})["port"] = float64(8888)
-
-						// This is a workaround to avoid the type mismatch
-						exporter["AdditionalProperties"] = nil
+						// This is a workaround to avoid the type mismatch because how go-yaml unmarshals
+						cfg.Service.Telemetry.Object["metrics"].(map[string]interface{})["readers"].([]interface{})[0].(map[string]interface{})["pull"].(map[string]interface{})["exporter"].(map[string]interface{})["prometheus"].(map[string]interface{})["port"] = int32(8888)
 						return cfg
 					}(),
 				},
@@ -204,16 +196,6 @@ func TestCollectorDefaultingWebhook(t *testing.T) {
 						const input = `{"receivers":{"otlp":{"protocols":{"grpc":{"endpoint":"0.0.0.0:4317","headers":{"example":"another"}},"http":{"endpoint":"0.0.0.0:4000"}}}},"exporters":{"debug":null},"service":{"telemetry":{"metrics":{"readers":[{"pull":{"exporter":{"prometheus":{"host":"localhost","port":9999}}}}]}},"pipelines":{"traces":{"receivers":["otlp"],"exporters":["debug"]}}}}`
 						var cfg v1beta1.Config
 						require.NoError(t, yaml.Unmarshal([]byte(input), &cfg))
-
-						r := cfg.Service.Telemetry.Object["metrics"].(map[string]interface{})["readers"]
-						readers := r.([]interface{})
-						firstReader := readers[0].(map[string]interface{})
-						exporter := firstReader["pull"].(map[string]interface{})["exporter"].(map[string]interface{})
-						// Convert port to float64 to avoid type mismatch because how go-yaml unmarshals
-						exporter["prometheus"].(map[string]interface{})["port"] = float64(9999)
-
-						// This is a workaround to avoid the type mismatch
-						exporter["AdditionalProperties"] = nil
 						return cfg
 					}(),
 				},
--- a/apis/v1beta1/config.go
+++ b/apis/v1beta1/config.go
@@ -6,14 +6,7 @@ package v1beta1
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
-	"fmt"
-	"math"
-	"reflect"
-	"regexp"
 	"sort"
-	"strconv"
-	"strings"
 
 	"dario.cat/mergo"
 	"github.com/go-logr/logr"
@@ -400,24 +393,24 @@ func (c *Config) Yaml() (string, error) {
 // Returns null objects in the config.
 func (c *Config) nullObjects() []string {
 	var nullKeys []string
-	if nulls := hasNullValue(c.Receivers.Object); len(nulls) > 0 {
+	if nulls := getNullValuedKeys(c.Receivers.Object); len(nulls) > 0 {
 		nullKeys = append(nullKeys, addPrefix("receivers.", nulls)...)
 	}
-	if nulls := hasNullValue(c.Exporters.Object); len(nulls) > 0 {
+	if nulls := getNullValuedKeys(c.Exporters.Object); len(nulls) > 0 {
 		nullKeys = append(nullKeys, addPrefix("exporters.", nulls)...)
 	}
 	if c.Processors != nil {
-		if nulls := hasNullValue(c.Processors.Object); len(nulls) > 0 {
+		if nulls := getNullValuedKeys(c.Processors.Object); len(nulls) > 0 {
 			nullKeys = append(nullKeys, addPrefix("processors.", nulls)...)
 		}
 	}
 	if c.Extensions != nil {
-		if nulls := hasNullValue(c.Extensions.Object); len(nulls) > 0 {
+		if nulls := getNullValuedKeys(c.Extensions.Object); len(nulls) > 0 {
 			nullKeys = append(nullKeys, addPrefix("extensions.", nulls)...)
 		}
 	}
 	if c.Connectors != nil {
-		if nulls := hasNullValue(c.Connectors.Object); len(nulls) > 0 {
+		if nulls := getNullValuedKeys(c.Connectors.Object); len(nulls) > 0 {
 			nullKeys = append(nullKeys, addPrefix("connectors.", nulls)...)
 		}
 	}
@@ -452,52 +445,15 @@ func (s *Service) MetricsEndpoint(logger logr.Logger) (string, int32, error) {
 	}
 
 	if telemetry.Metrics.Address != "" && len(telemetry.Metrics.Readers) == 0 {
-		// The regex below matches on strings that end with a colon followed by the environment variable expansion syntax.
-		// So it should match on strings ending with: ":${env:POD_IP}" or ":${POD_IP}".
-		const portEnvVarRegex = `:\${[env:]?.*}$`
-		isPortEnvVar := regexp.MustCompile(portEnvVarRegex).MatchString(telemetry.Metrics.Address)
-		if isPortEnvVar {
-			errMsg := fmt.Sprintf("couldn't determine metrics port from configuration: %s",
-				telemetry.Metrics.Address)
-			logger.Info(errMsg)
-			return "", 0, errors.New(errMsg)
-		}
-
-		// The regex below matches on strings that end with a colon followed by 1 or more numbers (representing the port).
-		const explicitPortRegex = `:(\d+$)`
-		explicitPortMatches := regexp.MustCompile(explicitPortRegex).FindStringSubmatch(telemetry.Metrics.Address)
-		if len(explicitPortMatches) <= 1 {
-			return telemetry.Metrics.Address, defaultServicePort, nil
-		}
-
-		port, err := strconv.ParseInt(explicitPortMatches[1], 10, 32)
+		host, port, err := parseAddressEndpoint(telemetry.Metrics.Address)
 		if err != nil {
-			errMsg := fmt.Sprintf("couldn't determine metrics port from configuration: %s",
-				telemetry.Metrics.Address)
-			logger.Info(errMsg, "error", err)
 			return "", 0, err
 		}
 
-		host, _, _ := strings.Cut(telemetry.Metrics.Address, explicitPortMatches[0])
-		return host, intToInt32Safe(int(port)), nil
+		return host, port, nil
 	}
 
-	port := defaultServicePort
-	host := defaultServiceHost
-	for _, reader := range telemetry.Metrics.Readers {
-		if reader.Pull != nil {
-			if reader.Pull.Exporter.Prometheus != nil {
-				if reader.Pull.Exporter.Prometheus.Host != nil && *reader.Pull.Exporter.Prometheus.Host != "" {
-					host = *reader.Pull.Exporter.Prometheus.Host
-				}
-				if reader.Pull.Exporter.Prometheus.Port != nil && *reader.Pull.Exporter.Prometheus.Port != 0 {
-					port = intToInt32Safe(int(*reader.Pull.Exporter.Prometheus.Port))
-				}
-				break
-			}
-		}
-	}
-	return host, port, nil
+	return defaultServiceHost, defaultServicePort, nil
 }
 
 // ApplyDefaults inserts configuration defaults if it has not been set.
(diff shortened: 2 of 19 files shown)


iblancasa:
I would expect some E2E test fail but we can start the review.

apis/v1beta1/config.go
```diff
@@ -441,67 +441,119 @@ const (
...
-	tm := &AnyConfig{
+	reader := s.getPrometheusMetricReader(logger)
+	tm := AnyConfig{
 		Object: map[string]interface{}{
 			"metrics": map[string]interface{}{
-				"address": fmt.Sprintf("%s:%d", telemetryAddr, telemetryPort),
+				"readers": []any{reader},
 			},
 		},
 	}
```
jaronoff97, line 509: I do think eventually we should eventually just import and use the config struct directly.

apis/v1beta1/config.go
```diff
@@ -441,67 +441,119 @@ const (
...
+
 	if err := mergo.Merge(s.Telemetry, tm); err != nil {
 		return fmt.Errorf("telemetry config merge failed: %w", err)
 	}
+
 	return nil
 }
 
+func (s *Service) getPrometheusMetricReader(logger logr.Logger) any {
+	getReader := func(addr string, port int32) any {
```
jaronoff97, line 535: why does this need to be a function?

apis/v1beta1/config.go
```diff
@@ -513,6 +565,32 @@ type MetricsConfig struct {
 
 	// Address is the [address]:port that metrics exposition should be bound to.
 	Address string `json:"address,omitempty" yaml:"address,omitempty"`
+
+	// Readers is the list of metric readers.
+	Readers []MetricReader `json:"readers,omitempty" yaml:"readers,omitempty"`
```
jaronoff97, line 570: i don't think this approach will work if we want to support users that set readers other than the prometheus one, can we not import the config package directly?

iblancasa: > i don't think this approach will work if we want to support users that set readers other than the prometheus one, 

Why?

> can we not import the config package directly?

Why don't we do that for the full configuration we have here?

apis/v1beta1/config.go
iblancasa: Before I was doing something more complicated and made more sense... I will simplify.

apis/v1beta1/config.go
jaronoff97: if a user [sets this](https://github.com/open-telemetry/opentelemetry-configuration/blob/2e76e05faf862650b23363d2a0f4d6b34df9e646/examples/kitchen-sink.yaml#L259-L275), won't we just drop the config?

iblancasa: I see.

iblancasa:
@jaronoff97 can you take another look? I still have to fix some E2E but I prefer to have a consensus regarding the solution before fixing those.

apis/v1beta1/collector_webhook_test.go
```diff
@@ -158,9 +159,19 @@ func TestCollectorDefaultingWebhook(t *testing.T) {
...
+
+						r := cfg.Service.Telemetry.Object["metrics"].(map[string]interface{})["readers"]
+						readers := r.([]interface{})
+						firstReader := readers[0].(map[string]interface{})
+						exporter := firstReader["pull"].(map[string]interface{})["exporter"].(map[string]interface{})
+						// Convert port to float64 to avoid type mismatch because how go-yaml unmarshals
+						exporter["prometheus"].(map[string]interface{})["port"] = float64(8888)
+
+						// This is a workaround to avoid the type mismatch
+						exporter["AdditionalProperties"] = nil
```
swiatekm, line 174: Can you create an issue to migrate to https://github.com/goccy/go-yaml and fix these by using [custom unmarshalers](https://pkg.go.dev/github.com/goccy/go-yaml#CustomUnmarshaler)?

swiatekm (commented):
LGTM overall. I'm a bit confused where the new dependencies are coming from, but the logic looks sound to me.

go.mod
```diff
@@ -233,4 +233,29 @@ require (
...
+	github.com/knadh/koanf/providers/confmap v0.1.0 // indirect
+	github.com/knadh/koanf/v2 v2.1.2 // indirect
+	github.com/mitchellh/copystructure v1.2.0 // indirect
+	github.com/mitchellh/reflectwalk v1.0.2 // indirect
+	go.opentelemetry.io/collector/component v1.29.0 // indirect
+	go.opentelemetry.io/collector/config/configtelemetry v0.123.0 // indirect
+	go.opentelemetry.io/collector/confmap v1.29.0 // indirect
+	go.opentelemetry.io/collector/internal/telemetry v0.123.0 // indirect
+	go.opentelemetry.io/collector/pdata v1.29.0 // indirect
+	go.opentelemetry.io/collector/service v0.123.0 // indirect
```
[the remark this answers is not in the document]
iblancasa: I faced the same question while creating the PR. It seems the cause is `otelConfig "go.opentelemetry.io/contrib/otelconf/v0.3.0"`

apis/v1beta1/collector_webhook_test.go
iblancasa: I think I will need to implement this right now. The thing is the marshaling in the collector is producing errors because of this additional field...

Using the `config` package from the collector is producing some headaches...

iblancasa: That's the thing.. it didn't and is producing error when the config is provided to the collector. I'm looking for a good way to do it.

go.mod
```diff
@@ -227,8 +227,22 @@ require (
...
 )
 
+require go.opentelemetry.io/contrib/otelconf v0.15.0
+
 require (
 	github.com/envoyproxy/go-control-plane/envoy v1.32.4 // indirect
+	github.com/golang/snappy v1.0.0 // indirect
 	github.com/google/btree v1.1.3 // indirect
 	github.com/prometheus/sigv4 v0.1.0 // indirect
+	go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploggrpc v0.11.0 // indirect
```
jaronoff97, line 237: @swiatekm i think these are coming from the otelconf new dependency

apis/v1beta1/config.go
```diff
@@ -445,67 +447,133 @@ const (
...
+		// delete(metrics, "address") does not work because Kubernetes treats missing fields
+		// differently from explicitly empty ones. By assigning "", we ensure the configuration
+		// is updated correctly when the resource is persisted.
+		tel.Metrics.Address = ""
 	}
 
-	if s.Telemetry == nil {
-		s.Telemetry = tm
-		return nil
+	yamlConfig, err := tel.toMap()
```
jaronoff97, line 522: maybe a dumb question... how hard would it be for us to just set the telemetry struct's config to be the otelconf struct? Actually, they're still working through the config and should probs wait on a v1, disregard...

iblancasa: Do you mean the full `Telemetry` struct?

apis/v1beta1/config.go
```diff
@@ -517,6 +585,23 @@ type MetricsConfig struct {
 
 	// Address is the [address]:port that metrics exposition should be bound to.
 	Address string `json:"address,omitempty" yaml:"address,omitempty"`
+
+	otelConfig.MeterProvider `mapstructure:",squash"`
```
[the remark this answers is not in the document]
iblancasa: I think we should try to solve the issue from this PR instead of continuing making it more difficult to handle... And create follow up PRs for improvements/refactors.

go.mod
iblancasa: They do. I double checked.

apis/v1beta1/config.go
jaronoff97: yes, but i don't think we should do this yet. i was thinking we could have [this](https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1beta1/config.go#L430) link to the config struct rather than anyconfig

apis/v1beta1/config.go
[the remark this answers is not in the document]
jaronoff97: yeah i just mean it's tough to review this with all the changes in config.go, im struggling to link them together effectively. I maybe need to spend some time jumping through the code.

apis/v1beta1/config.go
```diff
@@ -445,67 +447,133 @@ const (
...
-	if isPortEnvVar {
-		errMsg := fmt.Sprintf("couldn't determine metrics port from configuration: %s",
-			telemetry.Metrics.Address)
-		logger.Info(errMsg)
-		return "", 0, errors.New(errMsg)
-	}
+	if telemetry.Metrics.Address != "" && len(telemetry.Metrics.Readers) == 0 {
+		// The regex below matches on strings that end with a colon followed by the environment variable expansion syntax.
+		// So it should match on strings ending with: ":${env:POD_IP}" or ":${POD_IP}".
+		const portEnvVarRegex = `:\${[env:]?.*}$`
```
jaronoff97, line 457: can we move this and the following regex stuff to a helper function?

apis/v1beta1/config.go
```diff
@@ -445,67 +447,133 @@ const (
...
+	s.configureMetricReaders(logger, tel)
 
-	tm := &AnyConfig{
-		Object: map[string]interface{}{
-			"metrics": map[string]interface{}{
-				"address": fmt.Sprintf("%s:%d", telemetryAddr, telemetryPort),
-			},
-		},
+	// This field is deprecated and should not be used anymore.
+	if tel.Metrics.Address != "" {
```
jaronoff97, line 513: should this logic be in the upgrade package? Given we don't want to block this for older collectors being reconciled?

apis/v1beta1/config.go
```diff
@@ -445,67 +447,133 @@ const (
...
-		return fmt.Errorf("telemetry config merge failed: %w", err)
+
+	s.Telemetry = &AnyConfig{
+		Object: yamlConfig,
 	}
+
 	return nil
 }
 
+func (s *Service) configureMetricReaders(logger logr.Logger, tel *Telemetry) {
```
jaronoff97, line 534: suggestion: can we make this function idempotent? Should be easier to test this and then also test in the future when we expand our own capabilities.

jaronoff97: also we could just return a raw map here which would mean we could remove the toMap function. I don't _love_ using a map here instead of real types, but i'd rather this than needing that marshal/unmarshal below. @swiatekm thoughts?

apis/v1beta1/config.go
```diff
@@ -445,67 +447,133 @@ const (
...
+					},
+				},
+			},
+		}
+		return
+	}
+
+	for _, reader := range tel.Metrics.Readers {
+		if reader.Pull != nil {
+			if reader.Pull.Exporter.Prometheus != nil {
```
jaronoff97, line 565: is this a valid case? I guess i'm not sure if we need to do anything in the case someone has the pull exporter set to prometheus but no host/port. I think in this case, the host/port becomes the defaults [here](https://github.com/open-telemetry/opentelemetry-configuration/blob/2e76e05faf862650b23363d2a0f4d6b34df9e646/examples/kitchen-sink.yaml#L199-L202) and we wouldn't need to set this.

pkg/collector/upgrade/v0_111_0_test.go
```diff
@@ -34,9 +34,20 @@ func Test0_111_0Upgrade(t *testing.T) {
 	defaultCollectorWithConfig := defaultCollector.DeepCopy()
 
 	defaultCollectorWithConfig.Spec.Config.Service.Telemetry = &v1beta1.AnyConfig{
-		Object: map[string]interface{}{
-			"metrics": map[string]interface{}{
-				"address": "1.2.3.4:8888",
+		Object: map[string]any{
```
jaronoff97, line 37: technically this change is available in [v0.111.0](https://github.com/open-telemetry/opentelemetry-collector/blob/564818fd7f6291e18c1e1881d1596b72782199bd/service/telemetry/config.go#L25), but this uses v0.10.0 of the [config package](https://github.com/open-telemetry/opentelemetry-collector/blob/v0.111.0/service/go.mod#L40). And i _believe_ there were breaking changes between there, but those may be breaking changes for the API... either way I think this should be a forward upgrade and not modify previous installations.

jaronoff97 (commented):
okay i think i parsed through it some more. I definitely want more reviews on this prior to merge. Let's try to get this out this week to keep up with collector releases. cc @open-telemetry/operator-maintainers @open-telemetry/operator-triagers @open-telemetry/operator-approvers

apis/v1beta1/config.go
swiatekm: On second thought, I agree. The only thing the defaulter should do is add the reader configuration if the user hasn't configured anything. If they have, we just create Service entries for it and let the upgrade take care of migrating.

apis/v1beta1/config.go
```diff
@@ -445,67 +447,133 @@ const (
...
+				if reader.Pull.Exporter.Prometheus.Host != nil && *reader.Pull.Exporter.Prometheus.Host != "" {
+					host = *reader.Pull.Exporter.Prometheus.Host
+				}
+				if reader.Pull.Exporter.Prometheus.Port != nil && *reader.Pull.Exporter.Prometheus.Port != 0 {
+					port = intToInt32Safe(int(*reader.Pull.Exporter.Prometheus.Port))
+				}
+				break
+			}
+		}
+	}
```
swiatekm, line 499: I think this should happen in a separate PR. In this one, we want to change the defaulting logic to use non-deprecated configuration. Right now we aren't adding Service port entries for readers configured here, and we don't need to change that here.

apis/v1beta1/config.go
```diff
@@ -530,6 +615,48 @@ type Telemetry struct {
...
+					normalizeConfig(sub)
+				}
+			}
+		case float64:
+			// All numbers (even int32, int64, etc.) are parsed as float64 by default
+			// We need to convert them to the correct type
+			if k == "port" {
+				m[k] = int32(val)
+			}
+		}
```
swiatekm, line 656: Should have a default case here as well.

apis/v1beta1/config.go
```diff
@@ -530,6 +615,48 @@ type Telemetry struct {
...
+	if err := json.Unmarshal(data, &result); err != nil {
+		return nil, err
+	}
+
+	normalizeConfig(result)
+
+	return result, nil
+}
+
+func normalizeConfig(m map[string]interface{}) {
```
swiatekm, line 633: This function needs a docstring explaining why it exists.

apis/v1beta1/config.go
swiatekm: In general, I don't think we should do anything to existing readers. This whole for loop is unnecessary. If the user set their own metric readers, we don't need to add any defaults ourselves.

apis/v1beta1/config.go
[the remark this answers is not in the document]
iblancasa: I will move some stuff a to a new file.

apis/v1beta1/config.go
iblancasa: Ok, I will simplify this.

apis/v1beta1/config.go
iblancasa: This is removed so no action items.

apis/v1beta1/config.go
```diff
@@ -445,64 +441,61 @@ const (
...
+			if reader.Pull.Exporter.Prometheus != nil {
+				if reader.Pull.Exporter.Prometheus.Host != nil && *reader.Pull.Exporter.Prometheus.Host != "" {
+					host = *reader.Pull.Exporter.Prometheus.Host
+				}
+				if reader.Pull.Exporter.Prometheus.Port != nil && *reader.Pull.Exporter.Prometheus.Port != 0 {
+					port = intToInt32Safe(int(*reader.Pull.Exporter.Prometheus.Port))
+				}
+				break
+			}
+		}
```
swiatekm, line 470: In my opinion we should do this in a follow-up. It's a separate feature to generate Service port entries for all Pull readers. This implementation also only takes the first host/port pair, which is a bit arbitrary. I think we should drop this logic and open a separate issue to add it later.

swiatekm (commented):
Some final comments, but overall this looks good to me. Thanks for being so patient with the review process @iblancasa !

apis/v1beta1/config.go
swiatekm: Nothing. I think it's a good practice to always specify what happens if none of the cases match, but I don't care that much if you don't want to change it.

iblancasa: Didn't know about that. Thanks!

pkg/collector/upgrade/v_0_122_0.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1beta1"
+)
+
+func upgrade0_122_0(u VersionUpgrade, otelcol *v1beta1.OpenTelemetryCollector) (*v1beta1.OpenTelemetryCollector, error) { //nolint:unparam
+	tel := otelcol.Spec.Config.Service.GetTelemetry()
+
+	// if tel == nil || tel.Metrics.Address == "" {
+	// 	err := otelcol.Spec.Config.Service.ApplyDefaults(u.Log)
+	// 	return otelcol, err
+	// }
```
swiatekm, line 16: Here, I think we should just return.

```suggestion
	if tel == nil || tel.Metrics.Address == "" {
	 	return otelcol, err
	 }
```

apis/v1beta1/helper.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	if v < math.MinInt32 {
+		return math.MinInt32
+	}
+	return int32(v)
+}
+
+// getNullValuedKeys returns keys from the input map whose values are nil. Keys from nested maps are prefixed
+// by the name of the parent key, using a dot notation.
+func hasNullValue(cfg map[string]interface{}) []string {
```
swiatekm, line 72:
```suggestion
func getNullValuedKeys(cfg map[string]interface{}) []string {
```

apis/v1beta1/config.go
```diff
@@ -445,64 +441,46 @@ const (
...
 	}
 
-	tm := &AnyConfig{
-		Object: map[string]interface{}{
-			"metrics": map[string]interface{}{
-				"address": fmt.Sprintf("%s:%d", telemetryAddr, telemetryPort),
-			},
-		},
+	if tel.Metrics.Address != "" {
+		tel.Metrics.Address = fmt.Sprintf("%s:%d", host, port)
```
swiatekm, line 474: Is there a reason to change the address here?

pkg/collector/upgrade/v_0_122_0.go
iblancasa: sorry. commented to test and forget to remove

apis/v1beta1/config.go
iblancasa: Yes.
1. If you check the code before this PR, `ApplyDefaults` is setting a correct value for `address`. 
2. In this PR, I was removing the value from `address` to ensure that when we set a reader, there is not an issue due to two Prometheus servers listening on the same port.
3. I was requested to remove point 2.
4. I reverted to set a proper value for `address` when the user sets `address` and not add any readers. Which makes sense because there can be users that are still creating `OpenTelemetryCollector` instances using the deprecated field. Some of our tests are doing that.
5. If the field `address` was set in an `OpenTelemetryCollector` instance running in a previous version of the operator, the upgrade mechanism will remove the usage of `address`.

swiatekm: But if address is already set, why do we need to modify it? My naive view of this logic should be something like "If neither `address` nor `readers` are set, add a default reader. Otherwise, do nothing". Is there an issue with that?

iblancasa: Actually not. Sorry this comes from the mindset of the previous solution.

iblancasa: > But if address is already set, why do we need to modify it? My naive view of this logic should be something like "If neither `address` nor `readers` are set, add a default reader. Otherwise, do nothing". Is there an issue with that?

I posted a comment, removed, replied something different and later you posted again. 
Screenshot 2025-04-09 at 13 37 08

I fixed it.

swiatekm (approved):
:+1:

apis/v1beta1/helper.go
jaronoff97, line 1: can we rename this and the test to be helpers and helpers_test ?

apis/v1beta1/config.go
```diff
@@ -530,6 +525,38 @@ type Telemetry struct {
...
+	}
+
+	normalizeConfig(result)
+
+	return &AnyConfig{
+		Object: result,
+	}, nil
+}
+
+func (t *Telemetry) AddPrometheusMetricsEndpoint(host string, port int32) {
```
jaronoff97, line 546: IMO this method should be a helper function that returns an otelConfig.MetricReader and then we append it to t.Metrics.Readers rather than putting a method on Telemetry

pkg/collector/upgrade/v0_111_0_test.go
```diff
@@ -34,9 +34,20 @@ func Test0_111_0Upgrade(t *testing.T) {
...
+			"metrics": map[string]any{
```
jaronoff97, line 38: is this still supposed to be happening? I thought only the upgrade step would be making this change?

pkg/collector/upgrade/v_0_122_0.go
jaronoff97, line 1: filename should be v0_122_0 (same for the test)

pkg/collector/upgrade/v_0_122_0_test.go
```diff
@@ -0,0 +1,201 @@
...
+)
+
+var logger = logf.Log.WithName("unit-tests")
+
+func TestUpgrade0_122_0(t *testing.T) {
+	tests := []struct {
+		name           string
+		initialConfig  *v1beta1.OpenTelemetryCollector
+		expectedConfig *v1beta1.OpenTelemetryCollector
+	}{
```
jaronoff97, line 24: can we add tests for the following:
* someone sets a prom reader already, we should do nothing
* someone sets sending otlp instead of prom, we should do nothing

jaronoff97 (commented):
just a few final thoughts/questions, looking really good. much cleaner with the new helpers file too. thank you!

pkg/collector/upgrade/v0_111_0_test.go
iblancasa: Sorry. I thought I reverted this. Since the method `ApplyDefaults` has change and this test will not pass otherwise, I will port the logic to the `upgrade0_111_0.go` file.

pkg/collector/upgrade/v0_122_0_test.go
```diff
@@ -0,0 +1,341 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-collector",
+					Namespace: "default",
+				},
+				Spec: v1beta1.OpenTelemetryCollectorSpec{
+					Config: v1beta1.Config{
+						Service: v1beta1.Service{
+							Telemetry: &v1beta1.AnyConfig{
+								Object: map[string]interface{}{
+									"metrics": map[string]interface{}{
```
jaronoff97, line 221: OTLP looks a bit different:
```
          readers:
            - periodic:
                interval: 30000
                exporter:
                  otlp:
                    protocol: http/protobuf
                    endpoint: http://${env:MY_POD_IP}:4318
```

iblancasa: You are right. I copied/pasted and later not commit. Sorry again.

jaronoff97: np np! thanks for all the diligence here.

jaronoff97 (approved):

PR #3884, opened 2025-04-07
[chore]: update golangci-lint to v2

mmorel-35:
#### Description

Update golangci-lint to v2 and associated configurations


swiatekm (approved):

PR #3888, opened 2025-04-08
[Chore] Update Java instrumentation tests to assert telemetry data

IshwarKanse:
The PR updates Java instrumentation tests to assert telemetry data exported to the OTEL collector.


IshwarKanse:
The PR extends work done in https://github.com/open-telemetry/opentelemetry-operator/pull/3381

IshwarKanse:
cc @pavolloffay @jaronoff97 If this approach is reasonable, I will implement the same for other instrumentation tests.

jaronoff97:
@IshwarKanse i can get behind this method initially, but post kubecon we talked with the folks on the [weaver project](https://github.com/open-telemetry/weaver) and theyre working on doing telemetry schema validation at ingestion for testing which is exactly what we need. it would def be valuable to sync with them to see when that lands. If you'd like to lead this effort, lmk otherwise I can follow up with them.

IshwarKanse:
@jaronoff97 Ok, I was looking for a lightweight storage for all the telemetry data which can be used with OTEL collector, from which we can assert the exported data from OTEL. However I couldn't find an allinone storage. This PR uses the debug exporter and the tests check the logs for the presence of the telemetry data in the collector logs. I am interested in what the folks from Weaver project are doing for the telemetry schema validation and whether we can use it in our end to end tests.

jaronoff97:
IIRC they're working on something super lightweight that does the validation via OTLP ingestion and just stores in memory.

IshwarKanse:
@jaronoff97 I think this is the one they are working on. https://github.com/open-telemetry/weaver/pull/630/files#diff-06dbb93928a7cae1e6796a895e12491a357866cf4d76123b1b31c61c87e4f359 https://github.com/open-telemetry/weaver/pull/630

swiatekm:
Thanks for taking this up @IshwarKanse, it's one of the major blockers for stabilizing the Instrumentation CRD and a bunch of other improvements.

I like the approach of just using collector logs to get the data out of the cluster. Debugexporter seems like an inconvenient tool for this, though. Why not use fileexporter instead? It can serialize the data to the OTLP JSON format, which should make it easier to handle.

More generally, could we write the check scripts in Go instead of bash? It'd be even better if we could write them as proper Go tests, where we load the logs into pdata and assert on attributes instead of doing text search. Does that sound doable to you?

IshwarKanse:
@swiatekm Thats a good suggestion to use filelog exporter and doing validation using go tests. @jaronoff97 @pavolloffay Let me know what you guys think. I can update the PR by next week with the required changes.

jaronoff97:
I like Mikolaj's proposal here, i think once that weaver PR is merged we can move to that, but this is a great start and i don't think we should block on it.

swiatekm:
For reference, as a first iteration, I'm perfectly ok with just asserting that we got *some* data from instrumentation. We can add more detailed assertions about the shape of the data is subsequent PRs.

pavolloffay:
I also like the approach with the filelogexporter and golang. I would prefer asserts on entire sanitized   trace/metric JSON (e.g. removed timestamps, Ids)

tests/test-e2e-apps/otel-telemetry-validator/otel_telemetry_validator_test.go
IshwarKanse, line 1: @jaronoff97 @swiatekm @pavolloffay Updated the PR and added this GO test to validate the telemetry data exported by file exporter. We can expand the test later to include more comprehensive checks.

pavolloffay: I talked to @IshwarKanse privatelly and we have agreed to revert data checking for now. 

There are two objectives with e2e instrumentation tests:
1. make sure the instrumentation was loaded correctly and telemetry data is reported
2. make sure the reported data is consistent - e.g. the attributes do not change

First, we will implement 1. by checking the collector internal metrics (e.g. OTLP receiver received # of spans/metrics/logs). Then we will look at 2. by looking at the reported data, probably using https://github.com/open-telemetry/weaver/pull/630, or filelog exporter with declarative asserts.

IshwarKanse: @pavolloffay I have updated the PR to implement the first objective. After the weaver project PR is merged, I'll update for the comprehensive telemetry data validation. If this looks good, we can probably merge this PR. I'll update the rest of the instrumentation tests with same changes.

tests/e2e-instrumentation/instrumentation-java-multicontainer/chainsaw-test.yaml
```diff
@@ -40,12 +40,203 @@ spec:
...
+          outputs:
+          - name: podName
+            value: ($stdout)
+      - proxy:
+          apiVersion: v1
+          kind: Pod
+          name: ($podName)
+          namespace: ($namespace)
+          port: "8080"
+  - name: Wait for telemetry data to be collected
```
pavolloffay, line 68: Do we need this step and especially the script?

Could the next step retry the metrics query od fail if it does not pass after some time?

IshwarKanse: The script and the step is required for the time being, until this is implemented natively in Chainsaw. There is a feature request for that. https://github.com/kyverno/chainsaw/issues/1896 cc @eddycharly The other option is to add a sleep but I would prefer the script as we go onto the next step as soon as the strings are found in the pod logs. Otherwise we have to wait for whatever sleep duration that we have added. 

As for the metrics query step, this cannot be run in a loop for the required metrics to be found. The kubectl proxy is a single operation, the metrics assert works on the data fetched from the kubectl proxy. There is no retry at the try test step spec.

pavolloffay (approved):

swiatekm:
@IshwarKanse rebasing on `main` should resolve the CI failures.

pavolloffay (approved):

Issue #3515, opened 2024-12-05
Operator does not re-try failed upgrades


andreasgerstmayr:
### Component(s)

collector

## Description
The operator does not re-try failed upgrades of managed instances. In case an upgrade fails here: https://github.com/open-telemetry/opentelemetry-operator/blob/42a689e93bdeabded2ef4af08a9b5edc9b6e6486/pkg/collector/upgrade/upgrade.go#L86 (for example the Kubernetes API server is temporarily unreachable), an error is printed to the log, and the `status.version` field of the instance is updated  in the reconcile loop here: https://github.com/open-telemetry/opentelemetry-operator/blob/42a689e93bdeabded2ef4af08a9b5edc9b6e6486/internal/status/collector/handle.go#L55-L69 to the latest version regardless (note, the `spec` is not updated, only the `status` subresource). Therefore, any future re-starts of the operator also won't attempt to upgrade this instance.

Related, if the collector instance is moved from `unmanaged` to `managed` state, the upgrade process also doesn't run.

## Expected Result
The upgrade is re-tried.

## Actual Result
The `status.version` field of the instance is updated as part of the reconcile loop, however the `spec` field didn't get upgraded.

## Possible Solutions
Perform the upgrade process in the reconcile loop instead of the operator startup. This resolves the issue of re-trying failed upgrades, and also upgrading instances when they are moved from `unmanaged` to `managed` state.

### Kubernetes Version

1.31.0

### Operator version

0.113.0

### Collector version

0.113.0

jaronoff97:
cc @pavolloffay who wrote:
> Proposal is: run upgrade only as part of the reconciliation

I'll let him follow up with more info.

PR #3893, opened 2025-04-10
Re-try failed upgrades

andreasgerstmayr:
**Description:**
Currently, the operator upgrades all OpenTelemetryCollector CRs on startup of the operator. If any upgrade fails (e.g. intermittent errors of the Kubernetes API server), it won't be re-tried until the operator is restarted.

This commit moves the upgrade step to the reconcile loop. The operator reconciles all managed instances on startup, and in case of an error, re-tries the upgrade with exponential backoff.

Additionally, I changed the event type from Error to Warning, because "Error" is not a valid event type.

- Resolves: #3515

Manual testing, existing upgrade tests.

Not required


main.go
```diff
@@ -509,23 +508,9 @@ func main() {
 	}
 }
 
-func addDependencies(_ context.Context, mgr ctrl.Manager, cfg config.Config, v version.Version) error {
+func addDependencies(_ context.Context, mgr ctrl.Manager, cfg config.Config) error {
 	// adds the upgrade mechanism to be executed once the manager is ready
```
iblancasa, line 512: I think this line can be removed now, right?

andreasgerstmayr: I didn't change the upgrade process for the `Instrumentation` CR.
The `Instrumentation` CR upgrade is a bit different, it doesn't use versions (I can't skip the upgrade because I don't know if it's up-to-date), therefore I left it as-is for now.

swiatekm (commented):
The logic looks good to me, but I'm not fully convinced doing this during reconciliation is a good idea. Right now, we can make the very strong assumption that reconciliation does not modify the source CR, which this PR removes. Not sure that's worth it. @open-telemetry/operator-approvers I'd like to get more opinions about this change.

On a separate note, we definitely need more tests for this change.

controllers/opentelemetrycollector_controller.go
```diff
@@ -263,6 +265,26 @@ func (r *OpenTelemetryCollectorReconciler) Reconcile(ctx context.Context, req ct
 		return ctrl.Result{}, nil
 	}
 
+	// Perform an upgrade of this CR if required.
```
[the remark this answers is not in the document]
andreasgerstmayr: done

controllers/opentelemetrycollector_controller.go
```diff
@@ -263,6 +265,26 @@ func (r *OpenTelemetryCollectorReconciler) Reconcile(ctx context.Context, req ct
...
+			Client:   r,
+			Recorder: r.recorder,
+		}
+
+		var upgraded v1beta1.OpenTelemetryCollector
+		upgraded, err = up.Upgrade(ctx, instance)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		instance = upgraded
```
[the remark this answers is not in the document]
andreasgerstmayr: Good idea :+1:
I extracted the logic to `upgrade.NeedsUpgrade()` and `upgrade()` methods, and return and re-queue the reconcile event on upgrade.

I also added the `Spec.UpgradeStrategy` and `Spec.ManagementState` check back to the `upgrade.Upgrade` func.

andreasgerstmayr:
> The logic looks good to me, but I'm not fully convinced doing this during reconciliation is a good idea. Right now, we can make the very strong assumption that reconciliation does not modify the source CR, which this PR removes. Not sure that's worth it. @open-telemetry/operator-approvers I'd like to get more opinions about this change.

#3515 has more context why it's required in the reconcile loop. tl;dr
* outdated instances should be upgraded when the management state is switched from unmanaged back to managed
* re-trying failed upgrades with exponential backoff comes for free when it's in the reconcile loop
* I'd argue the reconcile should never run before the CR is up-to-date, because the manifest generation code should be able to assume that the CR is up-to-date.

> On a separate note, we definitely need more tests for this change.

Ok, I'll add some next week.

.chloggen/retry_upgrades.yaml
```diff
@@ -0,0 +1,18 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: operator
```
pavolloffay, line 5: It seems this changes the upgrade only for the collector. I would change the component to `collector`

pkg/collector/upgrade/upgrade.go
```diff
@@ -39,55 +39,46 @@ func (u VersionUpgrade) semVer() *semver.Version {
...
-	list := &v1beta1.OpenTelemetryCollectorList{}
-	if err := u.Client.List(ctx, list); err != nil {
-		return fmt.Errorf("failed to list: %w", err)
-	}
-
-	for i := range list.Items {
-		original := list.Items[i]
-		itemLogger := u.Log.WithValues("name", original.Name, "namespace", original.Namespace)
+// NeedsUpgrade checks if this CR needs to be upgraded.
+func NeedsUpgrade(instance v1beta1.OpenTelemetryCollector) bool {
```
pavolloffay, line 43: A test for this would be nice

.chloggen/retry_upgrades.yaml
andreasgerstmayr: done

pkg/collector/upgrade/upgrade.go
andreasgerstmayr: done

andreasgerstmayr:
> When it comes to testing, we do have some reconciliation tests running against a real API Server via envtest, but they're painful to write due to needing to manually wait for each condition to be reflected in controller-runtime's caching K8s client. So I won't blame you if you only add a chainsaw e2e test.

I gave it a try with envtest, and hit one error: `the object has been modified; please apply your changes to the latest version and try again` thrown at line https://github.com/open-telemetry/opentelemetry-operator/blob/2a075011e3ce5403baf2da1b84f1aaa1635154d6/controllers/opentelemetrycollector_controller.go#L269 after the first reconciliation.

I think it has to do with caching, as you mentioned. Fetching the object again resolved the (intermittent) issue.

I added three tests, one for upgrade, one already-up-to-date and one with an empty version in the status field.
Let me know if I should add more test cases.

internal/status/collector/handle.go
```diff
@@ -30,32 +26,21 @@ const (
...
 	}
-	changed := otelcol.DeepCopy()
 
-	up := &collectorupgrade.VersionUpgrade{
-		Log:      params.Log,
-		Version:  version.Get(),
-		Client:   params.Client,
-		Recorder: params.Recorder,
-	}
-	upgraded, upgradeErr := up.ManagedInstance(ctx, *changed)
```
andreasgerstmayr, line 44: Not sure why the upgrade was called in the `HandleReconcileStatus()` function?

My best guess is to update the `Status.Version` field? But that's already set in `updateCollectorStatus()`, which is called in `HandleReconcileStatus()`.
Also, the upgrade ran here, but it did not update the (modified) CR in the cluster.

swiatekm: I think it was to set the status field, yes. @jaronoff97 you refactored this back in 2023, do you recall if this is correct?

jaronoff97: yes, this was to set the status. I believe this was also to keep the functionality the same during my refactor.

swiatekm (approved):
LGTM. This is a pretty fundamental change to this feature though, so I'd like more reviews before we merge. @open-telemetry/operator-approvers please have a look, especially if you worked on upgrades in the past.

jaronoff97 (approved):
this makes sense to me, thanks for updating the logic :bow: Were you seeing any of the "object modified" messages after this change?

andreasgerstmayr:
> this makes sense to me, thanks for updating the logic 🙇 Were you seeing any of the `the object has been modified; please apply your changes to the latest version and try again` messages after this change?

I only saw this error message in the reconcile unit test, I didn't see it when I did manual testing using minikube.

PR #3914, opened 2025-04-16
Update the OpenTelemetry python agent version to 0.53b1

opentelemetrybot:
Update the OpenTelemetry python agent version to `0.53b1`.


autoinstrumentation/python/requirements.txt
```diff
@@ -1,4 +1,4 @@
-opentelemetry-distro==0.51b0
```
pavolloffay, line 1: all the versons below should be updated as well

swiatekm: I updated them manually.

swiatekm (approved):

iblancasa (approved):

Issue #3928, opened 2025-04-25
Some file config of OpAMPBridge ALWAYS overriden by CLI


OhJuhun:
### Component(s)

OpAMP Bridge

### Describe the issue you're reporting

because of this [function](https://github.com/open-telemetry/opentelemetry-operator/blob/main/cmd/operator-opamp-bridge/internal/config/config.go#L254-L292), some values written in config.yaml are always overriden by default values.

listen-addr
kubeconfig-path
heartbeat-interval
name

Since these fields have YAML tags in the Config struct, I believe the values should be injectable via a configuration file.
Is this intentional?

jaronoff97:
cc @swiatekm this is a bug in the bridge i believe, but this is also a bug in the target allocator which does the same thing.https://github.com/open-telemetry/opentelemetry-operator/blob/main/cmd/otel-allocator/internal/config/config.go#L248

@OhJuhun would you be willing to work on a fix? If not, I can ask around.

swiatekm:
Closed by accident by Github being overly eager.

OhJuhun:
this issue fixed by #3962 #3934 
cc @swiatekm @jaronoff97

PR #3934, opened 2025-04-26
fix bug: some OpAMPBridge values are always overriden by default

OhJuhun:

I fixed a bug where certain variables that could be set via both the config file and CLI were always being overridden by their default values, even if a config file was present. Now, when a config file exists, its values take precedence over the defaults, and if both a config file and CLI input are provided, the CLI input takes precedence.

@jaronoff97 

- Resolves: #https://github.com/open-telemetry/opentelemetry-operator/issues/3928


OhJuhun:
@jaronoff97 Can you review this plz?

swiatekm:
Instead of trying to detect if a value was loaded from file, I'd prefer if we took an approach similar to https://github.com/open-telemetry/opentelemetry-operator/pull/3962. That is:

1. Set all the defaults in a dedicated function.
2. Load values from the config value on top of that.
3. Apply flag values on top of the above, but only if they're actually set (by using flag.Changed).

WDYT @jaronoff97 ?

jaronoff97:
@swiatekm i think that approach is best. @OhJuhun would you be able to do a similar approach to what Mikolaj has done in his pr?

cmd/operator-opamp-bridge/internal/config/config_test.go
```diff
@@ -212,10 +273,9 @@ func TestLoad(t *testing.T) {
 			}
 			got := NewConfig(logr.Discard())
 			err := LoadFromFile(got, tt.args.file)
-			if !tt.wantErr(t, err, fmt.Sprintf("Load(%v)", tt.args.file)) {
+			if err != nil && tt.wantErr(t, err, fmt.Sprintf("Load(%v)", tt.args.file)) {
```
OhJuhun, line 276: @jaronoff97 
this line didn't work before.
I understood that if err is not nil and wantErr is satisfied, the process should not continue any further.
That's why I changed it this way. Isn't that correct?

jaronoff97: how did it not work before? We want to check both cases here, i.e. if we expect an error (where wantErr = assert.Error) and there is no error, then we should fail. And if we don't expect an error and get one we should fail too.

jaronoff97 (commented):
just one thought about the test, otherwise looks good. thank you! :bow:

cmd/operator-opamp-bridge/internal/config/config_test.go
OhJuhun: @jaronoff97
I understood that in cases where wantErr is not NoError, meaning when an error should occur, it was intended to terminate at this return statement and not execute the following code. However, in reality, even in cases where an error should occur, it wasn't returning from the if statement and was continuing to execute the next lines.
It seems to be fixed now.

jaronoff97: I see, regardless I think we should modify this line to serve its original purpose. I think what you could do to modify this would be change wantErr to be a boolean and then check if wantErr is true, run assert.Error(t, err)

OhJuhun: @jaronoff97
I fixed it.

OhJuhun: @jaronoff97 hello can you review this please?

jaronoff97 (approved):

swiatekm (approved):

PR #3944, opened 2025-04-28
[Chore] Update Python instrumenation tests to assert telemetry data

IshwarKanse:
**Description:** <Describe what has changed.>
The PR updates the Python auto-instrumentation tests to generate and assert telemetry data. 

The CI test jobs need to be rerun after PR https://github.com/open-telemetry/opentelemetry-operator/pull/3945 is merged.

PR #3945, opened 2025-04-28
[Chore] Update Python app Dockerfile used for testing auto-instrumentation

IshwarKanse:
Set the listen address for Python app to use --host=0.0.0.0 so that we can make a request to the app, required generating telemetry data from the instrumented app. 

Required by PR https://github.com/open-telemetry/opentelemetry-operator/pull/3944


tests/test-e2e-apps/python/Dockerfile
```diff
@@ -1,7 +1,16 @@
...
+RUN pip install --no-cache-dir -r requirements.txt
+
 COPY app.py .
 
-ENTRYPOINT ["flask", "run", "-p", "8080"]
+# Set the FLASK_APP environment variable
+ENV FLASK_APP=app.py
+
+# Expose the port the app runs on
+EXPOSE 8080
```
swiatekm, line 14: I don't think this is necessary. It only affects docker's `--publish-all` flag, which is not something we care about. In E2E tests, we always explicitly set the ports.

IshwarKanse: Yeah, I removed it in the updated commit.

swiatekm (approved):

PR #3946, opened 2025-04-29
[Chore] Update Dotnet test app to generate telemetry data.

IshwarKanse:
The PR updates the Dotnet test app with DiceRoller app for generating telemetry data.

### DIFF
--- a/tests/test-e2e-apps/dotnet/DiceRoller/DiceRoller.csproj
+++ b/tests/test-e2e-apps/dotnet/DiceRoller/DiceRoller.csproj
@@ -1,0 +1,9 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>net7.0</TargetFramework>
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
+    </PropertyGroup>
+
+</Project>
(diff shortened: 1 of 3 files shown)


swiatekm:
Is there a reason we're not making all the test apps do the same thing? It would be easier to write tests if that were the case imo.

iblancasa:
> Is there a reason we're not making all the test apps do the same thing? It would be easier to write tests if that were the case imo.

Agree

IshwarKanse:
@swiatekm Yes, I'm going through each instrumentation language tests one-by-one. I see many apps are instrumented but not generating any telemetry like this one with Dotnet. Seeing similar issue with Nodejs, the app is instrumented but not generating any telemetry. I'll check and update the tests accordingly to resolve any config issues.

swiatekm:
> @swiatekm Yes, I'm going through each instrumentation language tests one-by-one. I see many apps are instrumented but not generating any telemetry like this one with Dotnet. Seeing similar issue with Nodejs, the app is instrumented but not generating any telemetry. I'll check and update the tests accordingly to resolve any config issues.

I really appreciate your work here! And I don't have any objections to the content of this PR. I noticed that we have a dice rolling application here, while the Python application just returns a 200. Is the plan for all of these apps to do the dice rolling?

IshwarKanse:
@swiatekm Yes, we can do that to make it more consistent. :D  I'll update the rest of the apps in subsequent PRs.

swiatekm (approved):

iblancasa (approved):

jaronoff97 (approved):

Issue #3568, opened 2024-12-21
Ability to disable watching for OpenTelemetryCollector CRDs


atoulme:
### Component(s)

auto-instrumentation

### Is your feature request related to a problem? Please describe.

The manager currently registers to watch for opentelemetrycollector CRDs and fails if those are not defined or if its permissions are lacking.

### Describe the solution you'd like

This watcher should be optional.

jaronoff97:
@atoulme IIRC this was a design choice. If the operator cannot watch for the collector CRDs, it wouldn't be able to do it's purpose?

atoulme:
My purpose is to use the operator only to instrument applications, and I run my collector deployment in a different manner.

jaronoff97:
i understand that, but despite not running your collectors this way you can install the CRDs and do nothing with them. For optional CRDs that are not necessarily part of the project we can enable and disable some amount of this, but I'm not sure we would want to do this for the core collector functionality.

atoulme:
Why is that?

atoulme:
Here is an example showing how to execute on this approach, with a much smaller footprint:
https://github.com/otel-warez/whitegloves-operator

JaredTan95:
In my case, we are not using opentelemetrycollector CR to create the instance of opentelemetrycollector, just autoinstrumentation for java/python. I hope there is an optional to disable opentelemetry-collector controller. 

more https://github.com/open-telemetry/opentelemetry-operator/issues/3804

JaredTan95:
> > In my case, we are not using opentelemetrycollector CR to create the instance of opentelemetrycollector, just autoinstrumentation for java/python. I hope there is an optional to disable opentelemetry-collector controller.
> > more [#3804](https://github.com/open-telemetry/opentelemetry-operator/issues/3804)
> 
> This issue takes that into account since the controller will not "control" the OpenTelemetryCollectors instances.

Yes, I'm just following the issue and upvote here

PR #3955, opened 2025-04-30
[webhook] do not crash if collector CRD is not present in the cluster

atoulme:
**Description:** 
Skip registering the webhook and keep the operator working.

Related to #3568 - after merging this, adding an option to disable watching with the collector CRD is easy.

### DIFF
--- a/.chloggen/keep_going_if_collector_crd_is_not_present.yaml
+++ b/.chloggen/keep_going_if_collector_crd_is_not_present.yaml
@@ -1,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: webhook
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Do not exit if the collector CRD is missing.
+
+# One or more tracking issues related to the change
+issues: [3568]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Skip registering the webhook and keep the operator working.
--- a/.chloggen/keep_going_if_collector_crd_is_not_present.yaml
+++ b/.chloggen/keep_going_if_collector_crd_is_not_present.yaml
@@ -1,11 +1,11 @@
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: bug_fix
+change_type: enhancement
 
 # The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
 component: webhook
 
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Do not exit if the collector CRD is missing.
+note: Allow to run the operator without the OpenTelemetry CRDs present
 
 # One or more tracking issues related to the change
 issues: [3568]
@@ -14,4 +14,4 @@ issues: [3568]
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
 subtext: |
-  Skip registering the webhook and keep the operator working.
+  Skip registering the webhook and keep the operator working in case the OpenTelemetryCollector CRDs are not deployed.
--- a/internal/autodetect/main.go
+++ b/internal/autodetect/main.go
@@ -14,6 +14,7 @@ import (
 	"k8s.io/client-go/rest"
 
 	"github.com/open-telemetry/opentelemetry-operator/internal/autodetect/certmanager"
+	"github.com/open-telemetry/opentelemetry-operator/internal/autodetect/collector"
 	"github.com/open-telemetry/opentelemetry-operator/internal/autodetect/fips"
 	"github.com/open-telemetry/opentelemetry-operator/internal/autodetect/openshift"
 	"github.com/open-telemetry/opentelemetry-operator/internal/autodetect/prometheus"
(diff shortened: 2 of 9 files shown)


.chloggen/keep_going_if_collector_crd_is_not_present.yaml
```diff
@@ -0,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: webhook
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Do not exit if the collector CRD is missing.
```
iblancasa, line 8: I would prefer to add more information about this because right now, this is done by design, not by mistake.

.chloggen/keep_going_if_collector_crd_is_not_present.yaml
```diff
@@ -0,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
```
iblancasa, line 2: I would say it is a `enhancement`.

controllers/suite_test.go
```diff
@@ -92,6 +93,7 @@ type mockAutoDetect struct {
 	RBACPermissionsFunc             func(ctx context.Context) (autoRBAC.Availability, error)
 	CertManagerAvailabilityFunc     func(ctx context.Context) (certmanager.Availability, error)
 	TargetAllocatorAvailabilityFunc func() (targetallocator.Availability, error)
+	CollectorAvailabilityFunc       func() (collector.Availability, error)
```
iblancasa, line 96:
```suggestion
	CollectorCRDAvailabilityFunc       func() (collector.Availability, error)
```

iblancasa (commented):
I think we need to add E2E tests for this.

.chloggen/keep_going_if_collector_crd_is_not_present.yaml
atoulme: added more to the changelog

controllers/suite_test.go
atoulme: changed

main.go
```diff
@@ -371,19 +372,22 @@ func main() {
...
-		Client:   mgr.GetClient(),
-		Log:      ctrl.Log.WithName("controllers").WithName("OpenTelemetryCollector"),
-		Scheme:   mgr.GetScheme(),
-		Config:   cfg,
-		Recorder: mgr.GetEventRecorderFor("opentelemetry-operator"),
-		Reviewer: reviewer,
-		Version:  v,
-	})
+	var collectorReconciler *controllers.OpenTelemetryCollectorReconciler
+	if cfg.CollectorAvailability() == collector.Available {
```
swiatekm, line 376: This changes current behavior in a way I don't think is desirable. Crashing isn't good, so this is a useful detection mechanism, but the operator should still exit with an error if it doesn't find the CRDs it expects. For the vast majority of users, this will be a misconfiguration, and it will be confusing for them if the operator just keeps running without doing the work.

I'm fine with adding a flag later to change this behaviour, but we shouldn't change it by default.

atoulme: ok, I see what you're saying. I can make this consistent.

swiatekm (approved):

swiatekm:
It doesn't need to happen in this PR, but we should really pull all this logic out of `main.go` and make it independently testable. As is, we can only really check this in an E2E test, and even for those it's quite awkward to change operator command-line flags.

internal/autodetect/main.go
```diff
@@ -182,6 +184,38 @@ func (a *autoDetect) TargetAllocatorAvailability() (targetallocator.Availability
 	return targetallocator.NotAvailable, nil
 }
 
+// CollectorAvailability checks if OpenTelemetryCollector CR are available.
+func (a *autoDetect) CollectorAvailability() (collector.Availability, error) {
```
jaronoff97, line 188: should we merge this with the target allocator availability option and cover the other CRDs too?

jaronoff97 (approved):
one thought, but not blocking. Do we want to merge this before or after #3958

atoulme:
> It doesn't need to happen in this PR, but we should really pull all this logic out of `main.go` and make it independently testable. As is, we can only really check this in an E2E test, and even for those it's quite awkward to change operator command-line flags.

this is what I am working towards. I have been stripping main.go of logic and config. Eventually we can move cli arg parsing and env vars to something that reads them.

internal/autodetect/main.go
atoulme: I don't follow, sorry.

jaronoff97: should we also do this for the other CRDs (target allocator, instrumentation, opampbridge) so that we don't crash if those aren't present?

main.go
```diff
@@ -347,6 +353,15 @@ func main() {
 	} else {
 		setupLog.Info("Cert-Manager is not available to the operator, skipping adding to scheme.")
 	}
+	if cfg.CollectorAvailability() == collector.Available {
+		setupLog.Info("OpenTelemetryCollectorCRDSs is available to the operator")
+	} else {
+		setupLog.Info("OpenTelemetryCollectorCRDSs are not available to the operator")
+		if !cfg.IgnoreMissingCollectorCRDs() {
+			setupLog.Error(errors.New("missing OpenTelemetryCRDs"), "The OpenTelemetryCollector CRDs are not present in the cluster. Set ignore_missing_collector_crds to true or install the CRDs in the cluster.")
```
frzifus, line 361: Should we name the specific  CRD or CRDs that are missing?

atoulme:
```suggestion
			setupLog.Error(errors.New("missing OpenTelemetryCollector CRDs"), "The OpenTelemetryCollector CRDs are not present in the cluster. Set ignore_missing_collector_crds to true or install the CRDs in the cluster.")
```

internal/autodetect/main.go
atoulme: target allocator already has the logic: https://github.com/open-telemetry/opentelemetry-operator/blob/main/main.go#L389

If you don't mind, I'd leave this for the next PR.

frzifus (approved):

PR #3958, opened 2025-04-30
[chore] make Config a struct with config fields, but no logic

atoulme:
By moving the logic out of Config, we make its lifecycle simple as a struct. We can then start to simplify how it is initialized and load it from a file, CLI flags, env vars and so on.


internal/manifests/targetallocator/configmap.go
```diff
@@ -113,7 +113,7 @@ func ConfigMap(params Params) (*corev1.ConfigMap, error) {
 		taConfig["prometheus_cr"] = prometheusCRConfig
 	}
 
-	if params.Config.CertManagerAvailability() == certmanager.Available && featuregate.EnableTargetAllocatorMTLS.IsEnabled() {
+	if params.Config.CertManagerAvailability == certmanager.Available && featuregate.EnableTargetAllocatorMTLS.IsEnabled() {
```
iblancasa, line 116: Note this is a change in the behaviour: before, we were checking if cert-manager was available in each reconciliation. So, if it was installed/removed during two different reconciliations, the result would be different.

This applies to other changes introduced as part of this PR too.

I'm not against the change but we should, at least, add a changelog entry.

iblancasa (changes requested):

internal/manifests/targetallocator/configmap.go
atoulme: I don't see the functional change. The difference here is we directly access the field, but the method just returns that field:
```
// CertManagerAvailability represents the availability of the Cert-Manager.
func (c *Config) CertManagerAvailability() certmanager.Availability {
	return c.certManagerAvailability
}
```

iblancasa: Oh sorry. You are right. I thought this was calling the `autodetect` method (it was at some point).

iblancasa (approved):

swiatekm (approved):

internal/autodetect/main.go
```diff
@@ -185,3 +187,45 @@ func (a *autoDetect) TargetAllocatorAvailability() (targetallocator.Availability
 func (a *autoDetect) FIPSEnabled(_ context.Context) bool {
 	return fips.IsFipsEnabled()
 }
+
+// ApplyAutoDetect attempts to automatically detect relevant information for this operator.
+func ApplyAutoDetect(autoDetect AutoDetect, c *config.Config, logger logr.Logger) error {
```
swiatekm, line 192: Is there any reason not to make this a method on `AutoDetect`?

atoulme: Yes, though this is a more impactful change since we need all mocks to implement it. That might also change the way the interface works.

swiatekm: Allright, I'm fine leaving it as-is in this PR.

internal/autodetect/main.go
```diff
@@ -185,3 +187,45 @@ func (a *autoDetect) TargetAllocatorAvailability() (targetallocator.Availability
...
+	logger.V(2).Info("auto-detecting the configuration based on the environment")
+
+	ora, err := autoDetect.OpenShiftRoutesAvailability()
```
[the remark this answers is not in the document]
jaronoff97: i'm more thinking about how we can make autodetect a bit more extensible / generic. I was thinking it would be better to have a config options struct with autodetect supplying a set of options such that we can just provide a list of options from autodetect rather than going one by one here.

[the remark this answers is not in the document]
atoulme: as a layer that works. Definitely worth looking into, and maybe we can borrow from the collector libs on configsources.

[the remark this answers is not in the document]
jaronoff97: @atoulme do we want to do this in a follow up or in this pr?

jaronoff97 (approved):
one outstanding comment, otherwise lgtm

internal/autodetect/main.go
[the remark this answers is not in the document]
atoulme: The way I see it, I want to copy the collector approach of starting with a config struct initialized with default values, which then goes through mutators. First apply autodetect, then env vars, then cli args.

jaronoff97:
@atoulme tests are failing, otherwise i'd merge this

pavolloffay (approved):

Issue #3963, opened 2025-04-30
Inferred environment variables are duplicated if the user provided their own


iblancasa:
### Component(s)

collector

## Description
Some components (currently only `kubeletstats` receiver) provide some environment variables to set (`K8S_NODE_NAME` for instance). If a user provides the same variables, both are added to the containers.

This is not producing any error (at least I didn't see any error) but will produce some warnings in the OpenTelemetry Operator logs.

### Kubernetes Version

.

### Operator version

.

### Collector version

.

## Environment
OS: (e.g., "Ubuntu 20.04")
Compiler(if manually compiled): (e.g., "go 14.2")

### Log output

```shell

```

PR #3964, opened 2025-04-30
Use the env vars provided by the user instead of the ones inferred by the operator

iblancasa:
- Resolves: #3963


internal/manifests/collector/container.go
```diff
@@ -264,3 +209,78 @@ func defaultProbeSettings(probe *corev1.Probe, probeConfig *v1beta1.Probe) {
 		probe.TerminationGracePeriodSeconds = probeConfig.TerminationGracePeriodSeconds
 	}
 }
+
+func getEnvironmentVariables(otelcol v1beta1.OpenTelemetryCollector, logger logr.Logger) []corev1.EnvVar {
```
swiatekm, line 213: For this implementation, instead of going name by name, wouldn't it make more sense to add all the inferred env variables, and then deduplicate?

swiatekm: This way, we mix the deduplication logic with the logic for adding the environment variables in the first place. They aren't really related to each other and should ideally be separated in the code. For example, the following line:

```go
if _, ok := userDefined["SHARD"]; !ok && otelcol.Spec.TargetAllocator.Enabled {
```

is less readable to me than what we had before.

.chloggen/3963.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: collector
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Use the env vars provided by the user instead of the ones inferred by the operator"
```
pavolloffay, line 8: Could we be more specific on what is being fixed? 

e.g. Prevent the operator from overriding the env vars defined in the collector CR?

internal/manifests/collector/container.go
pavolloffay: Shouldn't be `SHARD` enforced by the operator to 0 anyways?

internal/manifests/collector/container.go
```diff
@@ -264,3 +209,78 @@ func defaultProbeSettings(probe *corev1.Probe, probeConfig *v1beta1.Probe) {
...
+		for _, envVar := range configEnvVars {
+			if _, ok := userDefined[envVar.Name]; !ok {
+				envVars = append(envVars, envVar)
+			}
+		}
+	}
+
+	envVars = append(envVars, proxy.ReadProxyVarsFromEnv()...)
+
+	return envVars
```
pavolloffay, line 285: `return append(envVars, proxy.ReadProxyVarsFromEnv()...)`

internal/manifests/collector/container.go
iblancasa: > Shouldn't be `SHARD` enforced by the operator to 0 anyways?

I'm not sure if there is any situation where users will want to set their own value. I will enforce 0 anyway since this was the previous behavior.

swiatekm (approved):
LGTM

jaronoff97 (approved):

PR #3965, opened 2025-05-01
Promote the `operator.collector.targetallocatorcr` feature flag to Beta

swiatekm:
**Description:**
As a result of this change, when the target allocator section is enabled in the Collector CR, this now creates a TargetAllocator CR instead of generating the manifests directly. Behavior should otherwise be unchanged.

#2422 

**Testing:**
Tests already exists for this behaviour. We simply stop running them with it disabled.

### DIFF
--- a/.chloggen/feat_enable-collector-tacr.yaml
+++ b/.chloggen/feat_enable-collector-tacr.yaml
@@ -1,0 +1,20 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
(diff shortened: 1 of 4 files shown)


jaronoff97 (approved):
our e2e tests cover the result of this already right? im 99% sure the answer is yes but just want to confirm my understanding

iblancasa (approved):

swiatekm:
> our e2e tests cover the result of this already right? im 99% sure the answer is yes but just want to confirm my understanding

Yes, this PR simply stops running them with the flag off. We were running them with it on for some time now, and continue to do so.

PR #3968, opened 2025-05-03
[validations]: use common expression language to validate sidecar checks

VihasMakwana:
Related https://github.com/open-telemetry/opentelemetry-operator/issues/3319

This PR moves some validations to CRD.

### User Impact:

If an invalid config is provide, then `kubectl` will exit if any of the validations fail. Previously, admission controller used to reject the request if CRD vas invalid.

### Tests

Added chainsaw e2e testing.


VihasMakwana:
Moving it to draft. Will mark it for review once test cases are added.

swiatekm:
You can't delete validations from the webhook, as we support K8s versions where CEL isn't available.

Another note is that we can only really test CEL validations using a full API Server, either in E2E tests using chainsaw, or in integration tests with envtest.

Finally, it'd be useful to leave a comment in the webhook code, noting which validations are covered in CEL.

VihasMakwana:
> You can't delete validations from the webhook, as we support K8s versions where CEL isn't available.

Oh okay. Thanks for heads up.

> Another note is that we can only really test CEL validations using a full API Server, either in E2E tests using chainsaw, or in integration tests with envtest.

Yeah, I kinda figured that out. I'm still playing around with it.

VihasMakwana:
@swiatekm This should be good now. Can you take a look?

apis/v1beta1/opentelemetrycollector_types.go
```diff
@@ -65,6 +65,11 @@ type OpenTelemetryCollectorStatus struct {
 	Image string `json:"image,omitempty"`
 }
 
+// +kubebuilder:validation:XValidation:rule="!(self.mode == 'sidecar' && size(self.tolerations) > 0) || !has(self.tolerations)",message="the OpenTelemetry Collector mode is set to sidecar, which does not support the attribute 'tolerations'"
```
swiatekm, line 68: Why is `|| !has(self.tolerations)` necessary? Doesn't the first clause cover it?

tests/e2e/validations/chainsaw-test.yaml
```diff
@@ -0,0 +1,40 @@
...
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  creationTimestamp: null
+  name: statefulset-features
+spec:
+  steps:
+  - name: step-00
+    try:
+    - script:
```
swiatekm, line 11: Instead of using `script` here, I think you can just apply the resource normally, as in https://kyverno.github.io/chainsaw/0.2.3/examples/negative-testing/#working-with-resources.

swiatekm (commented):
Could you also add a comment to the same checks in the webhook code, indicating that they're implemented using CEL validations as well.

tests/e2e/validations/chainsaw-test.yaml
VihasMakwana: yeah, but I also want to compare the error contents. When I used [standard way](https://kyverno.github.io/chainsaw/0.2.3/examples/negative-testing/#working-with-resources), I get `Internal error: variable not defined: $stderr`.

apis/v1beta1/opentelemetrycollector_types.go
VihasMakwana: Surprisingly, it doesn't. As these fields are optional, we need to check for their presence as well. https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/#optional-parameters

I initially assumed that priorityClassName, being an optional string field, would default to an empty string (''). However, validation fails if `priorityClassName` is not explicitly defined in the spec. Adding `has(..)` check fixes this.

VihasMakwana:
> Could you also add a comment to the same checks in the webhook code, indicating that they're implemented using CEL validations as well.

Sure. I'll do it

tests/e2e/validations/chainsaw-test.yaml
swiatekm: I thought `$error` contained the error message in that case. Doesn't it?

VihasMakwana: Let me verify that.

VihasMakwana: I've applied resource normally now. It should be good.

swiatekm:
Run `make update` and commit the resulting changes to pass unit tests.

VihasMakwana:
@swiatekm Now the [prometheus tests](https://github.com/open-telemetry/opentelemetry-operator/actions/runs/15101206181/job/42442403624?pr=3968) are failing and they look unrelated to my changes 😅

Makefile
```diff
@@ -369,6 +369,14 @@ e2e-upgrade: undeploy chainsaw
 	go run hack/check-operator-ready.go
 	$(CHAINSAW) test --test-dir ./tests/e2e-upgrade
 
+# end-to-end tests to test crd validations
+# skip testing for k8s versions < 1.25, as crd validations are still in alpha stage in those versions
```
VihasMakwana, line 373: I couldn't find a better way to achieve this. It seems like we can’t conditionally skip chaimsaw tests, though I might have missed something since I’m still new to this.

swiatekm: You can do this, but it's currently a bit awkward. You can't skip the step or match the check based on bindings, but you can incorporate the K8s server version into the value itself. The following worked for me:

```yaml
apiVersion: chainsaw.kyverno.io/v1alpha1
kind: Test
metadata:
  name: cel-validations-test
spec:
  bindings:
  - name: version
    value: (x_k8s_server_version($config))
  - name: minorVersion
    value: (to_number($version.minor))
  - name: priorityClassErrorMessageCEL
    value: |-
      OpenTelemetryCollector.opentelemetry.io "sidecar-priorityclass" is invalid: spec: Invalid value: "object": the OpenTelemetry Collector mode is set to sidecar, which does not support the attribute 'priorityClassName'
  - name: priorityClassErrorMessageWebhook
    value: |-
      admission webhook "vopentelemetrycollectorcreateupdatebeta.kb.io" denied the request: the OpenTelemetry Collector mode is set to sidecar, which does not support the attribute 'priorityClassName'
  steps:
  - name: step-00
    try:
    - apply:
        file: 00-install-priority-class.yaml
        expect:
         - check:
            ($error != null): true
            ($error): (($minorVersion > `24` && $priorityClassErrorMessageCEL) || $priorityClassErrorMessageWebhook)
```

It's a bit verbose, but it can run on both versions, and verify the error message from the webhook as well.

VihasMakwana: This looks neat. Thank you!

swiatekm (approved):
Nice! Thank you for the contribution!

VihasMakwana:
> Nice! Thank you for the contribution!

🥳 

Is this ready to be merged yet?

swiatekm:
> > Nice! Thank you for the contribution!
> 
> 🥳
> 
> Is this ready to be merged yet?

Waiting for a second approval before merging.

iblancasa (approved):

PR #3973, opened 2025-05-06
[Chore] Fix smoke-daemonset test assert

IshwarKanse:
The PR fixes the assert for smoke-daemonset test when running on Kubernetes cluster with more than 1 worker nodes and some other minor fixes.

### DIFF
--- a/tests/e2e-instrumentation/instrumentation-apache-multicontainer/chainsaw-test.yaml
+++ b/tests/e2e-instrumentation/instrumentation-apache-multicontainer/chainsaw-test.yaml
@@ -111,8 +111,8 @@ spec:
             ($error == null): true
       - assert:
           resource:
-            ($metrics[?as_string(metric."__name__") == 'otelcol_exporter_sent_spans_total'].value | [0] > `1`): true
-            ($metrics[?as_string(metric."__name__") == 'otelcol_receiver_accepted_spans_total'].value | [0] > `1`): true
+            ($metrics[?as_string(metric."__name__") == 'otelcol_exporter_sent_spans_total'].value | [0] > `0`): true
+            ($metrics[?as_string(metric."__name__") == 'otelcol_receiver_accepted_spans_total'].value | [0] > `0`): true
     catch:
       - podLogs:
           selector: app=my-apache-multi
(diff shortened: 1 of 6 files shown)


iblancasa (approved):

tests/e2e/smoke-daemonset/00-assert.yaml
```diff
@@ -19,6 +34,3 @@ metadata:
 status:
   (starts_with(image, 'ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector')): true
   (version != ''): true
-  scale:
-    replicas: 1
-    statusReplicas: "1/1"
```
swiatekm, line 24: We shouldn't remove this assertion, it checks if our handling of Collector CR status is correct. I'd rather you made it work in the multi-node case instead.

swiatekm, line 24: We shouldn't remove this assertion, it checks if our handling of Collector CR status is correct. I'd rather you made it work in the multi-node case instead.

swiatekm (changes requested):

tests/e2e-instrumentation/instrumentation-apache-multicontainer/chainsaw-test.yaml
```diff
@@ -222,8 +219,8 @@ spec:
             ($error == null): true
       - assert:
           resource:
-            ($metrics[?as_string(metric."__name__") == 'otelcol_exporter_sent_spans_total'].value | [0] > `1`): true
-            ($metrics[?as_string(metric."__name__") == 'otelcol_receiver_accepted_spans_total'].value | [0] > `1`): true
```
swiatekm, line 226: Shouldn't this be at least 2 given that it's the second test we're running against this collector?

IshwarKanse: A new pod is created for the second test as we apply a changed CR which removes the instrumentation annotation for the second container.

swiatekm: Ah, I missed that we're running the collector as a sidecar in this test. Everything's fine then. :+1:

PR #3986, opened 2025-05-12
feat: run autoinstrumentation init container as non root user 1000

### LINKED ISSUE #2272: [auto-instrumentation] Init containers image run as root prevents usage of auto-injection where OPA Policies enforce runAsNonRoot
### Component(s)

instrumentation

## Description

With the recent release of operators, the init-containers takes the securityContext of the container where they are injecting the instrumentation libraries . However, several enterprises enforce runAsNonRoot to be true in the application containers through OPA policies. With this update, when the initContainers inherit runAsNonRoot, they are erroring out with the following error
"Error: container has runAsNonRoot and image will run as root (pod: xxx, container: opentelemetry-auto-instrumentation-java)

## Steps to Reproduce

1. Created a pod with a container running a basic java application
2. Set the container securityContext to have runAsNonRoot: true
3. Create an instrumentation object 
4. Add the auto-instrumentation annotation to the container where instrumentation is required - instrumentation.opentelemetry.io/inject-java: true

## Expected Result

The initContainer should come up successfully and inject the instrumentation libraries

## Actual Result

The initContainer fails to come up with the error mentioned above.

### Kubernetes Version

1.23.0

### Operator version

0.87.0

### Collector version

0.87.0

### Environment information

_No response_

### Log output

_No response_

### Additional context

_No response_

patst:
Refs: #2272

We are rebuilding currently all auto-instrumentation images to set a non-root user with id.
This makes sure it passes our kyverno policies, which prevent running containers as root user.

**Link to tracking Issue(s):** https://github.com/open-telemetry/opentelemetry-operator/issues/2272

- Resolves: #2272

**Testing:** We are using this since over a year in our environment for nodejs, dotnet and java. Unfortunately I cannot test it for python and apache-http, but it should work similiar

**Documentation:** none


iblancasa (approved):

jaronoff97 (approved):
I think this should be okay, but would like some other maintainers to approve prior to merging.

swiatekm (changes requested):
I'm not sure what the consequences of this kind of change could be, especially given that E2E tests don't run on these changed images as part of PR checks. Could we instead start by setting `runAsUser` in the container security context for these containers? Even better if we make it configurable.

patst:
> I'm not sure what the consequences of this kind of change could be, especially given that E2E tests don't run on these changed images as part of PR checks. Could we instead start by setting `runAsUser` in the container security context for these containers? Even better if we make it configurable.

@swiatekm  Thanks for your feedback, I understand your concerns and share them to a certain extend. Would be great to run the integration tests for this change.

In general it makes sense to be secure by default and make sure the workloads only have the permissions they really require, without requiring the user to configure it. If that is achieved another way than the one in the PR I am very happy to use the solution as well!

If this change is not the way to go, I gonna close the PR. Unfortunately, I have to the time to look into the other solution at the moment :-(

iblancasa:
> I'm not sure what the consequences of this kind of change could be, especially given that E2E tests don't run on these changed images as part of PR checks.

How do you think we can increase the confidence in PRs like this one? To be honest, I think this is a change we need to do at some point. Is it a breaking change? I'm not sure (I guess it is not) but I think this is something we need to do at some point.

aalexiev42:
When can we expect this to get merged? I'm also facing the problem with runAsNonRoot in my envrironment.

pavolloffay (approved):

iblancasa:
> We need to start running e2e tests on instrumentation images built from the PR tip. To expedite this specific case, I'd also accept if someone does that manually and confirms that all the e2e tests pass.
> 

+1

Issue #3988, opened 2025-05-12
Replace instances of `gopkg.in/yaml.v3`


jaronoff97:
the package is fully unmaintained and we shouldn't use it anymore.

LucaLanziani:
hello @jaronoff97 replace it with https://pkg.go.dev/sigs.k8s.io/yaml or something else?

jaronoff97:
That one is fine probably, yes

jaronoff97:
iirc sigs.k8s.io is just a wrapper around go yaml, but i hope the k8s folks are maintaining it haha. I think it should be fine to do this move initially, and if we need to do it again we can do it again.

LucaLanziani:
Happy to take this then.

jaronoff97:
awesome, thank you!

LucaLanziani:
With the k8s library the TestConfigYaml result goes from:

```
receivers:
  otlp: null
exporters:
  otlp/exporter: null
processors:
  modify_2000: enabled
connectors:
  con: magic
extensions:
  addon: option1
service:
  extensions:
    - addon
  telemetry:
    insights: yeah!
  pipelines:
    traces:
      exporters:
        - otlp/exporter
        - con
      processors:
        - modify_2000
      receivers:
        - otlp
```

to:

```
connectors:
  con: magic
exporters:
  otlp/exporter: null
extensions:
  addon: option1
processors:
  modify_2000: enabled
receivers:
  otlp: null
service:
  extensions:
  - addon
  pipelines:
    traces:
      exporters:
      - otlp/exporter
      - con
      processors:
      - modify_2000
      receivers:
      - otlp
  telemetry:
    insights: yeah!
```

You can see how lists are not indented while they were with the unsupported library, would you be ok with this new format?

jaronoff97:
that _should_ be okay, i'm not opposed. you may be able to do a setting to change this such that it's a no-op though

jaronoff97:
errant closure oops!

LucaLanziani:
so, I tried moving to  https://pkg.go.dev/sigs.k8s.io/yaml but:
1. as you said that is a wrapper of go yaml
2. it doesn't offer any option to marshal the yaml in the format currently used

I tested https://github.com/goccy/go-yaml and it seems to me it could be a better option.

I have the changes ready with all test passing I'll open a PR soon

LucaLanziani:
On a second note, I've found the following in the code:
```
[...]

PR #3997, opened 2025-05-15
feat: migrate from sigs.k8s.io/yaml to goccy/go-yaml

LucaLanziani:
**Description:**

Replace unmaintenance yaml library sigs.k8s.io/yaml with https://github.com/goccy/go-yaml

- Resolves: #3988 

**Testing:**

```
make test
```


iblancasa (approved):

LucaLanziani:
I'll check these in a moment

LucaLanziani:
@iblancasa sure, I'll fix the remaining test this evening, I assumed that `make e2e` was running all tests but apparently it doesn't.

LucaLanziani:
@iblancasa I don't know if you have seen my comment on the issue
There other instances of the package `yaml.v2`

I believe we should target these as part of this PR

```
cmd/operator-opamp-bridge/internal/config/config.go:    "gopkg.in/yaml.v2"
cmd/otel-allocator/internal/config/config.go:   "gopkg.in/yaml.v2"
cmd/otel-allocator/internal/server/server.go:   "gopkg.in/yaml.v2"
cmd/otel-allocator/internal/server/server_test.go:      "gopkg.in/yaml.v2"
cmd/otel-allocator/internal/watcher/promOperator.go:    "gopkg.in/yaml.v2"
internal/manifests/collector/adapters/config_from.go:   "gopkg.in/yaml.v2"
internal/manifests/collector/config_replace_test.go:    "gopkg.in/yaml.v2"
internal/manifests/opampbridge/configmap.go:    "gopkg.in/yaml.v2"
internal/manifests/targetallocator/configmap.go:        "gopkg.in/yaml.v2"
pkg/collector/upgrade/v0_19_0.go:       "gopkg.in/yaml.v2"
pkg/collector/upgrade/v0_24_0.go:       "gopkg.in/yaml.v2"
pkg/collector/upgrade/v0_31_0.go:       "gopkg.in/yaml.v2"
pkg/collector/upgrade/v0_36_0.go:       "gopkg.in/yaml.v2"
pkg/collector/upgrade/v0_38_0.go:       "gopkg.in/yaml.v2"
pkg/collector/upgrade/v0_39_0.go:       "gopkg.in/yaml.v2"
pkg/collector/upgrade/v0_43_0.go:       "gopkg.in/yaml.v2"
pkg/collector/upgrade/v0_57_2.go:       "gopkg.in/yaml.v2"
pkg/collector/upgrade/v0_9_0.go:        "gopkg.in/yaml.v2"
```

LucaLanziani:
and I found more.

https://github.com/open-telemetry/opentelemetry-operator/blob/main/cmd/otel-allocator/internal/server/server.go#L17

apis/v1alpha1/convert.go
```diff
@@ -7,7 +7,7 @@ import (
 	"errors"
 	"fmt"
 
-	"gopkg.in/yaml.v3"
+	go_yaml "github.com/goccy/go-yaml"
```
frzifus, line 10: (nit)
```suggestion
	yaml "github.com/goccy/go-yaml"
```

frzifus (commented):
I guess we need to adapt all the e2e tests in that case :see_no_evil:

LucaLanziani:
After looking at the indentation here once more https://github.com/open-telemetry/opentelemetry-operator/actions/runs/15075431358/job/42467505446?pr=3997#step:8:430

I realized that the "github.com/goccy/go-yaml" library is right and the format in the test [here](https://github.com/open-telemetry/opentelemetry-operator/blob/main/tests/e2e-ta-collector-mtls/ta-collector-mtls/00-assert.yaml#L66) is wrong.

If we use two spaces for the following:

```
{"readers":[{"pull":{"exporter":{"prometheus":{"host":"0.0.0.0","port":8888}}}}]}
```

looks like:

```
readers:
  - pull:
      exporter:
        prometheus:
          host: 0.0.0.0
          port: 8888
```

so indenting by four can only be:

```
readers:
    - pull:
          exporter:
              prometheus:
                  host: 0.0.0.0
                  port: 8888
```

but in the test you see we have a mix of the two.

You can see the difference between the two libraries here: https://go.dev/play/p/fo6MEYZyIuy

So ideally we should fix the indentation in the tests but I believe @swiatekm you said that doing that will cause an unnecessary reconciliation for users.

swiatekm:
> So ideally we should fix the indentation in the tests but I believe @swiatekm you said that doing that will cause an unnecessary reconciliation for users.

If the current output is incorrect and difficult to reproduce, then we should bite the bullet and change it.

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,15 @@
...
+- name: controller
+  newName: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
+  newTag: 0.124.0-42-ge40b341e
+patches:
+- patch: '[{"op":"add","path":"/spec/template/spec/containers/0/args/-","value":"--target-allocator-image=ghcr.io/open-telemetry/opentelemetry-operator/target-allocator:v0.124.0-42-ge40b341e"}]'
+  target:
+    kind: Deployment
+- patch: '[{"op":"add","path":"/spec/template/spec/containers/0/args/-","value":"--operator-opamp-bridge-image=ghcr.io/open-telemetry/opentelemetry-operator/operator-opamp-bridge:v0.124.0-42-ge40b341e"}]'
+  target:
+    kind: Deployment
```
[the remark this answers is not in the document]
LucaLanziani: FYI: I'm not sure if I'm reading this right but I believe the command in the `make reset` target should be checking out from `main`:

```bash
git checkout main config/manager/kustomization.yaml
```

LucaLanziani:
I'll create a second issue to replace all other yaml libraries.

LucaLanziani:
I might be doing something wrong but I'm not able to reproduce [this error](https://github.com/open-telemetry/opentelemetry-operator/actions/runs/15210517242/job/42796513455?pr=3997#step:6:843) on my machine.

swiatekm:
> I might be doing something wrong but I'm not able to reproduce [this error](https://github.com/open-telemetry/opentelemetry-operator/actions/runs/15210517242/job/42796513455?pr=3997#step:6:843) on my machine.

It's flaky, I restarted it.

apis/v1alpha1/convert.go
swiatekm: I think it's better for this to consistently be `go_yaml` everywhere. When I see `yaml`, I assume it's `gopkg.in/yaml`, so a different alias seems useful to indicate that it's not.

LucaLanziani:
New issue created github.com/open-telemetry/opentelemetry-operator/issues/4028

PR #3998, opened 2025-05-15
Remove overwrite of global config SraperProtocols

bogdandrutu:
**Description:** <Describe what has changed.>

**Link to tracking Issue(s):** <Issue number if applicable>

- Resolves: #3996 
- Resolves: #3872 

**Testing:** <Describe what testing was performed and which tests were added.>

**Documentation:** <Describe the documentation added.>

PR #4000, opened 2025-05-15
Allow to configure ScraperProtocols in prometheus common config

bogdandrutu:


iblancasa (approved):

cmd/otel-allocator/internal/watcher/promOperator.go
```diff
@@ -89,7 +83,7 @@ func NewPrometheusCRWatcher(ctx context.Context, logger logr.Logger, cfg allocat
 				ProbeNamespaceSelector:          cfg.PrometheusCR.ProbeNamespaceSelector,
 				ServiceDiscoveryRole:            &serviceDiscoveryRole,
 				Version:                         "2.55.1", // fix Prometheus version 2 to avoid generating incompatible config
-				ScrapeProtocols:                 DefaultScrapeProtocols,
+				ScrapeProtocols:                 cfg.PrometheusCR.ScrapeProtocols,
```
bogdandrutu, line 86: Alternative, instead of having ScrapeProtocols in PrometheusCR, we can use the ScrapeProtocols from the `PromConfig.GlobalConfig.ScrapeProtocols`, do you think that is cleaner?

cmd/otel-allocator/internal/config/config.go
```diff
@@ -48,6 +49,12 @@ var (
 	DefaultKubeConfigFilePath string = filepath.Join(homedir.HomeDir(), ".kube", "config")
 )
 
+var defaultScrapeProtocolsCR = []monitoringv1.ScrapeProtocol{
```
jaronoff97, line 52: do we want this to land after #3998 ?

bogdandrutu: If we do, I will update :)

bogdandrutu: Alternative, can land this first (since this not change defaults) and update the defaults for this path as well in the other PR.

jaronoff97: lets do the other one first and then this one, i think it'll make this one even more clear

jaronoff97 (approved):

cmd/otel-allocator/internal/watcher/promOperator.go
bogdandrutu: I am ok with that :)

cmd/otel-allocator/internal/config/config.go
bogdandrutu: Ready to merge.

bogdandrutu:
@swiatekm @iblancasa this PR is now ready to final review and merge, thank you.

swiatekm (approved):

swiatekm:
@ChristianCiach I think Prometheus's documentation is wrong here. See the actual defaults in their [config code.](https://github.com/prometheus/prometheus/blob/9a0bbb60bc3eb68d045aae7535d34f4d02b959f1/config/config.go#L544).

The Prometheus 2 to 3 migration was quite painful for both prometheus receiver and target allocator, resulting in various confusing issues like this one. What version of target allocator gives you those warnings? This change was released in 0.126.0, which uses Prometheus 3.1.0, not 2.55.0 as the log insists.

ChristianCiach:
@swiatekm You may be right that the prometheus documentation is probably outdated.

I am using the current Target Allocator 0.129.1. The prometheus-operator code thinks we are using Prometheus 2.55 due to the hardcoded version string in https://github.com/open-telemetry/opentelemetry-operator/blob/2e23ad4ed4ea419fae344cc4c394e03100921da0/cmd/otel-allocator/internal/watcher/promOperator.go#L85

Maybe this should be removed? I understand the code well enough to judge this.

swiatekm:
Yes, that should definitely be removed.

Issue #4025, opened 2025-05-23
Promote the `operator.collector.targetallocatorcr` feature gate to Stable


swiatekm:
This feature gate controls whether the OpenTelemetryCollector CR with an emebedded target allocator configuration internally creates a TargetAllocator CR resource instead of managing the manifests directly.

- [x] Promoted to Alpha
- [ ] Promoted to Beta, enabled by default
- [ ] Promoted to Stable
- [ ] Removed

swiatekm:
For reference, the flag was moved to Beta in https://github.com/open-telemetry/opentelemetry-operator/pull/3965.

PR #4026, opened 2025-05-23
Fix embedded TargetAllocator CR labels

swiatekm:
**Description:**
Labels on the TargetAllocator CR created by the OpenTelemetryCollector CR are a bit tricky. If we inherit them the same way we do for other resources, the values will be incompatible with the current ones. They're also not semantically correct, as it doesn't make much sense that, for example, the standard name label would be the same for the TA CR.

Instead, I decided to add the `managed-by` label manually, but skip the rest. This doesn't affect any current functionality, and also gives the same result if the `operator.collector.targetallocatorcr` feature gate is enabled.

- Related https://github.com/open-telemetry/opentelemetry-operator/issues/4025

**Testing:**
Modified existing tests.

### DIFF
--- a/internal/controllers/builder_test.go
+++ b/internal/controllers/builder_test.go
@@ -3136,7 +3136,9 @@ service:
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      "test",
 						Namespace: "test",
-						Labels:    nil,
+						Labels: map[string]string{
+							"app.kubernetes.io/managed-by": "opentelemetry-operator",
+						},
 					},
 					Spec: v1alpha1.TargetAllocatorSpec{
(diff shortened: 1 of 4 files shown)


iblancasa (approved):
> Instead, I decided to add the `managed-by` label manually, but skip the rest. This doesn't affect any current functionality, and also gives the same result if the `operator.collector.targetallocatorcr` feature gate is enabled.

Can this affect users in any other way? I mean: let's say they are using some kind of dashboard or other tool to do something based on labels... would this break them? If yes, I'm not against the change but I think we might want to add a changelog entry just to have some kind of "warning" about this behavior change.

Issue #4029, opened 2025-05-24
Add serviceName Field to OpenTelemetryCollector CRD for StatefulSet Deployments


rewt:
### Component(s)

collector

### Is your feature request related to a problem? Please describe.

When deploying the OpenTelemetry Collector as a StatefulSet using the OpenTelemetry Operator in a GKE cluster, the OpenTelemetryCollector CRD does not allow setting the serviceName to match the headless Service name, causing a mismatch with statefulset.serviceName. Kubernetes requires statefulset.serviceName to match the headless Service name for pods to register DNS hostnames (e.g., <pod-name>.<serviceName>.<namespace>.svc.cluster.local). Without this, the loadbalancing exporter’s kubernetes resolver (return_hostnames: true) fails to resolve pod hostnames, resulting in:
```
error: couldn't find the exporter for the endpoint ""
```
when configured as

```
exporters:
  loadbalancing:
    routing_key: "traceID"
    protocol:
      otlp:
        timeout: 1s
        tls:
          insecure: true
    resolver:
      k8s:
        service: opentelemetry-backend-collector-headless.plat-observe-dev
        timeout: 3s
        return_hostnames: true
```
For example, my headless Service (opentelemetry-backend-collector-headless.plat-observe-dev) doesn’t match the StatefulSet’s default serviceName (opentelemetry-backend-collector), preventing hostname resolution:

```
kubectl get statefulset opentelemetry-backend-collector -n plat-observe-dev -o jsonpath='{.spec.serviceName}'
opentelemetry-backend-collector

nslookup opentelemetry-backend-collector-0.plat-observe-dev.svc.cluster.local
** server can't find ...: NXDOMAIN
```
This breaks traceID routing, critical for tail-based sampling in my setup.

### Describe the solution you'd like

Add a serviceName field to the OpenTelemetryCollector CRD to specify the headless Service name when mode: statefulset. The Operator should set statefulset.serviceName to this value, ensuring DNS hostname registration.

```yaml
apiVersion: opentelemetry.io/v1alpha1
kind: OpenTelemetryCollector
metadata:
  name: otel-backend
  namespace: plat-observe-dev
spec:
  name: "opentelemetry-backend-collector"
  serviceName: "opentelemetry-backend-collector-headless"
  mode: statefulset
  replicas: 9
  autoscaling:
    enabled: true
    minReplicas: 3
    maxReplicas: 10
  config:
    receivers:
      otlp:
        protocols:
          grpc:
[...]

PR #4041, opened 2025-05-27
Add serviceName Field to OpenTelemetryCollector CRD for StatefulSet Deployments

vignesh-codes:
**Description:** 
Add support for custom service names in StatefulSet headless services

This PR fixes the handling of service names for StatefulSet mode in the OpenTelemetry Collector operator:
- Adds ability to specify custom service names via `spec.serviceName` field
- When running in StatefulSet mode and no service name is specified, defaults to `<name>-headless`
- When creating HeadlessService and no service name is specified, defaults to `<name>-headless`
- Maintains existing behavior for non-StatefulSet modes
- Ensures TLS secret names follow the service naming pattern

**Link to tracking Issue(s):** [4029](https://github.com/open-telemetry/opentelemetry-operator/issues/4029)

**Testing:** 
Added new test cases in `service_test.go` for headlessService creation and in `statefulset_test.go` for statefulset creation with custom and default service names:
- Test default headless service name generation for StatefulSet mode
- Test custom service name handling
- Test TLS secret name generation
- Verify ClusterIP is set to "None" for headless services

**Documentation:**

### DIFF
--- a/internal/manifests/collector/service.go
+++ b/internal/manifests/collector/service.go
@@ -39,12 +39,17 @@ func (s ServiceType) String() string {
 }
 
 func HeadlessService(params manifests.Params) (*corev1.Service, error) {
+	// Use serviceName from spec if provided, otherwise default to <name>-headless for StatefulSets
+	serviceName := params.OtelCol.Spec.ServiceName
+	if serviceName == "" {
+		serviceName = naming.HeadlessService(params.OtelCol.Name)
+	}
 	h, err := Service(params)
 	if h == nil || err != nil {
 		return h, err
 	}
 
-	h.Name = naming.HeadlessService(params.OtelCol.Name)
+	h.Name = serviceName
 	h.Labels[headlessLabel] = valueExists
 	h.Labels[serviceTypeLabel] = HeadlessServiceType.String()
 
--- a/internal/manifests/collector/statefulset.go
+++ b/internal/manifests/collector/statefulset.go
@@ -27,6 +27,12 @@ func StatefulSet(params manifests.Params) (*appsv1.StatefulSet, error) {
 	if err != nil {
 		return nil, err
 	}
+	serviceName := naming.Service(params.OtelCol.Name)
+	if params.OtelCol.Spec.ServiceName != "" {
+		serviceName = params.OtelCol.Spec.ServiceName
+	} else {
+		serviceName = naming.HeadlessService(params.OtelCol.Name)
+	}
 
 	return &appsv1.StatefulSet{
 		ObjectMeta: metav1.ObjectMeta{
@@ -36,7 +42,7 @@ func StatefulSet(params manifests.Params) (*appsv1.StatefulSet, error) {
 			Annotations: annotations,
 		},
 		Spec: appsv1.StatefulSetSpec{
-			ServiceName: naming.Service(params.OtelCol.Name),
+			ServiceName: serviceName,
 			Selector: &metav1.LabelSelector{
 				MatchLabels: manifestutils.SelectorLabels(params.OtelCol.ObjectMeta, ComponentOpenTelemetryCollector),
 			},
--- a/.chloggen/sts-headless-svc-naming.yaml
+++ b/.chloggen/sts-headless-svc-naming.yaml
@@ -1,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern
+component: collector
+
+# A brief description of the change
+note: "Fix the headless service name in StatefulSet mode. When not specified, defaults to '<name>-headless' for StatefulSet headless services."
+
+# One or more tracking issues related to the change
+issues: [4029]
+
+# Optional additional information
+subtext: |
+  The ServiceName field in the OpenTelemetryCollector spec can now be used to customize 
+  the headless service name when running in StatefulSet mode. This provides more 
+  flexibility in service naming while maintaining default behavior.
--- a/apis/v1alpha1/opentelemetrycollector_types.go
+++ b/apis/v1alpha1/opentelemetrycollector_types.go
@@ -233,6 +233,11 @@ type OpenTelemetryCollectorSpec struct {
 	// +optional
 	InitContainers []v1.Container `json:"initContainers,omitempty"`
 
+	// ServiceName is the name of the Service to be used.
+	// If not specified, it will default to "<name>-headless".
+	// +optional
+	ServiceName string `json:"serviceName,omitempty"`
+
 	// AdditionalContainers allows injecting additional containers into the Collector's pod definition.
 	// These sidecar containers can be used for authentication proxies, log shipping sidecars, agents for shipping
 	// metrics to their cloud, or in general sidecars that do not support automatic injection. This option only
--- a/apis/v1beta1/opentelemetrycollector_types.go
+++ b/apis/v1beta1/opentelemetrycollector_types.go
@@ -130,6 +130,11 @@ type OpenTelemetryCollectorSpec struct {
 	// This is only applicable to Deployment mode.
 	// +optional
 	DeploymentUpdateStrategy appsv1.DeploymentStrategy `json:"deploymentUpdateStrategy,omitempty"`
+
+	// ServiceName is the name of the Service to be used.
+	// If not specified, it will default to "<name>-headless".
+	// +optional
+	ServiceName string `json:"serviceName,omitempty"`
 }
 
 // TargetAllocatorEmbedded defines the configuration for the Prometheus target allocator, embedded in the
--- a/internal/manifests/collector/service_test.go
+++ b/internal/manifests/collector/service_test.go
@@ -270,6 +270,27 @@ func TestHeadlessService(t *testing.T) {
 	})
 }
 
+func TestHeadlessServiceWithCustomName(t *testing.T) {
+	t.Run("should return headless service with default name", func(t *testing.T) {
+		param := statefulsetParams()
+		actual, err := HeadlessService(param)
+		assert.NoError(t, err)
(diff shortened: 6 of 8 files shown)


internal/manifests/collector/service.go
```diff
@@ -39,12 +39,17 @@ func (s ServiceType) String() string {
 }
 
 func HeadlessService(params manifests.Params) (*corev1.Service, error) {
+	// Use serviceName from spec if provided, otherwise default to <name>-headless for StatefulSets
```
iblancasa, line 42: I'm wondering if this change can break somehow the `ServiceMonitor` integration.

pavolloffay:
Could you please add assert here https://github.com/open-telemetry/opentelemetry-operator/blob/6d2f18b0ac0303aff2b904c2de76296cea60fbf9/tests/e2e/smoke-statefulset/00-assert.yaml

vignesh-codes:
> Could you please add assert here https://github.com/open-telemetry/opentelemetry-operator/blob/6d2f18b0ac0303aff2b904c2de76296cea60fbf9/tests/e2e/smoke-statefulset/00-assert.yaml

@pavolloffay I have added them

vignesh-codes:
@pavolloffay @iblancasa requesting for workflow approvals and reviews

internal/manifests/collector/service.go
swiatekm: As far as I'm aware, this only affects DNS for StatefulSet Pods. @iblancasa don't we have an e2e test for this, in any case?

iblancasa: > @iblancasa don't we have an e2e test for this, in any case?
I don't think so.

I'm fine with the change because -after checking the code- it seems there will not be any kind of interactions...

swiatekm: I created an issue to add this E2E test: https://github.com/open-telemetry/opentelemetry-operator/issues/4064.

swiatekm:
@vignesh-codes running `make update` should fix the CI failure.

tests/e2e/smoke-statefulset/00-assert.yaml
```diff
@@ -16,3 +16,24 @@ status:
   scale:
     replicas: 1
     statusReplicas: "1/1"
+---
```
swiatekm, line 21: We should also assert that `serviceName` is set to `stateful-collector-headless` in the StatefulSet definition above.

swiatekm (changes requested):
I imagined this feature working a bit differently. Rather than letting the user rename the Service (which in and out of itself isn't that useful), it should let them provide the name of a Service they created themselves. This how setting [`serviceAccount`](https://github.com/open-telemetry/opentelemetry-operator/blob/e8f834acb8643c3e68fa6b268819c931ffbceace/internal/manifests/collector/serviceaccount.go#L27) currently works, but we can always create the headless Service regardless. @open-telemetry/operator-approvers what do you think we should do here?

I don't have any complaints about the fix setting the default as the headless Service, so maybe we can make that a separate PR?

vignesh-codes:
> I imagined this feature working a bit differently. Rather than letting the user rename the Service (which in and out of itself isn't that useful), it should let them provide the name of a Service they created themselves. This how setting [`serviceAccount`](https://github.com/open-telemetry/opentelemetry-operator/blob/e8f834acb8643c3e68fa6b268819c931ffbceace/internal/manifests/collector/serviceaccount.go#L27) currently works, but we can always create the headless Service regardless. @open-telemetry/operator-approvers what do you think we should do here?
> 
> I don't have any complaints about the fix setting the default as the headless Service, so maybe we can make that a separate PR?

Cool! i made the `make update` command and pushed the commit. Let me know

internal/manifests/collector/service.go
```diff
@@ -39,12 +39,17 @@ func (s ServiceType) String() string {
...
+	serviceName := params.OtelCol.Spec.ServiceName
+	if serviceName == "" {
+		serviceName = naming.HeadlessService(params.OtelCol.Name)
+	}
```
swiatekm: In my opinion we shouldn't do this at all. The headless Service name should always be the same and `Spec.ServiceName` should only affect the setting on StatefulSets.

rewt: Kubernetes requires headless service name to 100% match the statefulset.servicename to create dns host records.

relevant post https://discuss.kubernetes.io/t/how-to-figure-out-pod-dns-name-for-statefulset-and-headless-service-mongodb/26221/2

swiatekm: We're adding a feature allowing users to use their own headless Service for this. Changing the name of the Service generated by the operator is unnecessary.

vignesh-codes: @swiatekm So we just add the feature to accept serviceName if given by the user and not update the auto generated service name to have <name>-headless to complete this ?
Let me know so i can wrap this up

swiatekm: Yes. We let the user set it, and make the auto generated headless Service name the default. I don't think anything else is needed.

vignesh-codes: @swiatekm i made the changes to go with default existing flow if no servicenames were given by the user

internal/manifests/collector/statefulset.go
```diff
@@ -28,6 +28,11 @@ func StatefulSet(params manifests.Params) (*appsv1.StatefulSet, error) {
 		return nil, err
 	}
 
+	serviceName := params.OtelCol.Spec.ServiceName
+	if serviceName == "" {
+		serviceName = naming.Service(params.OtelCol.Name)
```
swiatekm: This should be `naming.HeadlessService(params.OtelCol.Name)`.

internal/manifests/collector/service.go
swiatekm: I thought we agreed not to make any changes to the headless Service itself.

swiatekm (commented):
Could you also address my comment in the E2E test?

internal/manifests/collector/service.go
vignesh-codes: @swiatekm wait i thought we were supposed to take the name if given by the user right? 
Also, in previous version we set h.Name to naming.HeadlessService(params.OtelCol.Name) https://github.com/open-telemetry/opentelemetry-operator/pull/4041/files#diff-b76db9a2bef9a36d7b94a8922fc1f70b2b646fa6b2d0bbfb8ed827bc5f722451L47.

 
I just updated it to check if user gave their own svc name.

Its not the thing we intended to do?

internal/manifests/collector/statefulset.go
vignesh-codes: @swiatekm we sure about that?
We are currently using naming.Service(params.OtelCol.Name) as given here https://github.com/open-telemetry/opentelemetry-operator/pull/4041/files#diff-3536f4bc156bfb0fc500698bcc3ff96a3ab9f0ed0731804371d59482687fe917L39

I am just checking if user gave their own svc name - if not, then go with existing default flow.

Let me know if my understandings are wrong about the requirements.

internal/manifests/collector/service.go
swiatekm: We don't want to make any changes to the headless Service created by the operator. Instead, we want to let users set their own serviceName on the StatefulSet, and also make the default the headless Service, rather than the normal one. Does that make sense?

rewt:
this is my current working config from the [collector helm chart](https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-collector)
```opentelemetry-collector:
  nameOverride: "dev-opentelemetry-backend-collector"
  fullnameOverride: "dev-opentelemetry-backend-collector-headless"
```
here is how the collector chart currently functions https://github.com/open-telemetry/opentelemetry-helm-charts/blob/09a02cb2dc1fe2c0e7e0ee51c2fe9c6a95ddc807/charts/opentelemetry-collector/templates/statefulset.yaml#L16

It'd be nice for a similar option with the OpenTelemetryCollector CR when `mode: statefulset`

internal/manifests/collector/service.go
vignesh-codes: @swiatekm I have made the following changes
- reverted the changes in headless service creation in service.go (func HeadlessService) and removed the new test case i added for it.
- added functionality to have custom servicename for sts creation in statefulset.go (func StatefulSet) and changed the default naming to `serviceName = naming.HeadlessService(params.OtelCol.Name)` from `serviceName = naming.Service(params.OtelCol.Name)` if not given custom names.
- added test case in statefulset_test.go to validate that change.
- made changes in builder_test.go to expect for `test-collector-headless` in sts servicename creation.

Let me know if its fine. Thanks :D

swiatekm:
@vignesh-codes looks good to me now! Could you resolve the lint failure? Then we should be good to merge.

bundle/openshift/manifests/opentelemetry.io_opentelemetrycollectors.yaml
```diff
@@ -3,7 +3,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     cert-manager.io/inject-ca-from: opentelemetry-operator-system/opentelemetry-operator-serving-cert
-    controller-gen.kubebuilder.io/version: v0.18.0
+    controller-gen.kubebuilder.io/version: v0.17.1
```
vignesh-codes: @swiatekm I have fixed the lint issue and did make update

why this version got downgraded after i did make update? is it expected?

swiatekm: @vignesh-codes you'll need to delete your local `./bin` directory and run `make update` again. The controller-tools version on main was updated since you opened your PR, hence the change.

vignesh-codes: @swiatekm i have updated the pr with make update as u suggested. Requesting for your review

Thanks :)

swiatekm (approved):

vignesh-codes:
@iblancasa Waiting for your review and approval. Thanks

iblancasa (approved):

PR #4050, opened 2025-05-30
Invert autodetect

atoulme:
To be merged after #4046 

This change recreates #3958

### DIFF
--- a/internal/autodetect/main.go
+++ b/internal/autodetect/main.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"slices"
 
+	"github.com/go-logr/logr"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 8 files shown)


jaronoff97:
i like the approach here, seems CI isn't happy with you though :P

atoulme:
OK, it should be passing now.

internal/autodetect/main.go
```diff
@@ -219,3 +221,52 @@ func (a *autoDetect) CollectorAvailability() (collector.Availability, error) {
 func (a *autoDetect) FIPSEnabled(_ context.Context) bool {
 	return fips.IsFipsEnabled()
 }
+
+// ApplyAutoDetect attempts to automatically detect relevant information for this operator.
+func ApplyAutoDetect(autoDetect AutoDetect, c *config.Config, logger logr.Logger) error {
```
jaronoff97, line 226: should we make this a method on the autodetect struct? Or is the idea that doing it this way makes it more testable?

atoulme: more testable, and quite a few times you pass in autodetect mocks.

PR #4052, opened 2025-05-31
No options

atoulme:
This removes the intermediate struct used to configure the Config struct and instead gives direct access to the Config struct for configuration, removing the intermediate Options struct. This removes all the `With` helper methods used.

This PR also introduces 2 layers of config setting appliers, one for env vars and one for CLI args, to be assigned sequentially.

This PR is build on top of #4050 and should be reviewed after.


pkg/instrumentation/podmutator_test.go
```diff
@@ -863,7 +867,11 @@ func TestMutatePod(t *testing.T) {
 					},
 				},
 			},
-			config: config.New(config.WithEnableNodeJSInstrumentation(true)),
+			config: func() config.Config {
+				cfg := config.New()
+				cfg.EnableNodeJSAutoInstrumentation = true
+				return cfg
+			}(),
```
jaronoff97: i don't _love_ needing to do this in all of the tests, it's definitely where the options pattern is more elegant. Is there a plan for how this will work after this migration, or is this the desired final state?

atoulme: oh that's about it. Yes, maximum pain principle is applied here.

jaronoff97: argh yeah that's the one thing about doing this that i really don't love. I have to consider this strategy a bit more and think if there'a good balance we can strike here...

atoulme: another fun way to look at this - we create the config with opionated defaults in all our tests today. That means every time we change our defaults, we eat having to update one by one all our tests.
We might want to start config as a struct and set its value disregarding defaults for the purpose of testing a specific scenario, rather than patching the default config.

swiatekm: In the interest of keeping the more ergonomic option pattern, could we change options to be functions that operate directly on the Config struct? Then we could avoid all the test changes while still removing the options struct.

atoulme: This API is not external. I don't understand what this adds?

swiatekm: Config creation is nicer with the options pattern.

```go
cfg := config.New(
    config.WithCollectorImage("default-collector"),
    config.WithTargetAllocatorImage("default-ta-allocator"),
)
```

is preferable to this:

```go
cfg := config.New()
cfg.CollectorImage = "default-collector"
cfg.TargetAllocatorImage = "default-ta-allocator"
```

Or at least that's how I understood Jacob's comment. I don't have very strong feelings about it, but I'd like this PR a bit more if it didn't need to modify all these tests.

atoulme: It's not actually nicer. It's more code that no one looks at since the only users are us. It's a hop of indirection that makes you unable to check on struct field usage. And it's the reason why I had to modify all this code in the first place.

atoulme: discussed in SIG - making config immutable is the goal, so rewriting tests to generate the Config struct with fields set.

swiatekm (approved):

Issue #3565, opened 2024-12-21
Document all flags associated with manager


atoulme:
### Component(s)

auto-instrumentation

### Is your feature request related to a problem? Please describe.

The manager executable seems to accept a number of ways of configuration:
* A JSON file
* CLI flags
* Env vars

There is no uniformity between those arguments - some env vars for example like ENABLE_WEBHOOKS have no equivalent CLI flag.

### Describe the solution you'd like

* Create a documentation for the existing
* Create tests to uniformly enforce how the config is set so each key is sure to have an equivalent in JSON/CLI Flag/Env var
* Fix inconsistencies

jaronoff97:
Agreed, this is in a few places currently with most of documentation being the code itself. This would be a good one to take.

shuheiktgw:
Let me work on this issue as my first contribution to the project, if I may 🙏

swiatekm:
After @atoulme's recent refactor of configuration loading, this should now be possible looking solely at the [config package.](https://github.com/open-telemetry/opentelemetry-operator/tree/main/internal/config). I'd create a `docs/configuration.md`, then remove all configuration references from the README and replace them with links to the new document.

siddhant-khisty:
Hey, I'd like to work on this

gaurav337:
Hi, I’d like to work on this. I’ll create a new PR updating docs/configuration.md to include CLI flags, JSON options, env vars, and defaults, addressing the feedback from #4309.

swiatekm:
@gaurav337 could you base your change on https://github.com/open-telemetry/opentelemetry-operator/pull/5259? Part of a recently merged docs RFC is also to generate this documentation automatically.

gaurav337:
Hi @swiatekm,
I’m ready to tackle the auto-generated config docs for #3565.
Since #5259 introduces the new reference/ structure, I want to ensure my PR aligns with it. Should I base my branch on swiatekm:docs/reorganize to get the correct folder structure now, or would you prefer I wait until #5259 merges into main?
Thanks!

swiatekm:
@gaurav337 it's simpler if you base you branch on mine, I think. It's unlikely there's going to be sweeping changes in it, so once it's merged, your PR should apply cleanly to main.

amarkdotdev:
[...]

PR #4054, opened 2025-06-01
Harmonize config

atoulme:
**Description:**
Harmonize the configuration to expose both CLI and env vars.

Relates to #3565


iblancasa (approved):

internal/config/cli.go
```diff
@@ -8,35 +8,61 @@ import (
...
+	f.Bool(constants.FlagApacheHttpd, cfg.EnableApacheHttpdInstrumentation, "Controls whether the operator supports Apache HTTPD auto-instrumentation")
+	f.Bool(constants.FlagDotNet, cfg.EnableDotNetAutoInstrumentation, "Controls whether the operator supports dotnet auto-instrumentation")
+	f.Bool(constants.FlagGo, cfg.EnableGoAutoInstrumentation, "Controls whether the operator supports Go auto-instrumentation")
+	f.Bool(constants.FlagPython, cfg.EnablePythonAutoInstrumentation, "Controls whether the operator supports python auto-instrumentation")
+	f.Bool(constants.FlagNginx, cfg.EnableNginxAutoInstrumentation, "Controls whether the operator supports nginx auto-instrumentation")
+	f.Bool(constants.FlagNodeJS, cfg.EnableNodeJSAutoInstrumentation, "Controls whether the operator supports nodejs auto-instrumentation")
+	f.Bool(constants.FlagJava, cfg.EnableJavaAutoInstrumentation, "Controls whether the operator supports java auto-instrumentation")
+	f.Bool(constants.FlagCRMetrics, cfg.EnableCRMetrics, "Controls whether exposing the CR metrics is enabled")
```
swiatekm: While we're changing the config initialization, let's get rid of these constants.

atoulme: but that's like more work! I don't like work

atoulme: OK fine - inlined the constants.

internal/config/env.go
```diff
@@ -38,4 +40,73 @@ func ApplyEnvVars(cfg *Config) {
 	if v, ok := os.LookupEnv("RELATED_IMAGE_AUTO_INSTRUMENTATION_NGINX"); ok {
 		cfg.AutoInstrumentationNginxImage = v
 	}
+	if v, ok := os.LookupEnv("OPENSHIFT_CREATE_DASHBOARD"); ok {
+		cfg.OpenshiftCreateDashboard, _ = strconv.ParseBool(v)
+	}
+	if v, ok := os.LookupEnv("METRICS_ADDR"); ok {
+		cfg.MetricsAddr = v
+	}
+	if v, ok := os.LookupEnv("PROBE_ADDR"); ok {
```
swiatekm: Can we make the naming conversion consistent with the rest of the env variables here? It'll make it easier to eventually migrate to koanf if we decide to.

```suggestion
	if v, ok := os.LookupEnv("HEALTH_PROBE_ADDR"); ok {
```

atoulme: Just this one or is there more? I don’t know what all the variables do. I will take a look.

swiatekm: I looked them over and this was the only inconsistency I found. For clarity, what I mean is that the command-line option is named `health-probe-addr`, so if we're converting dashes into underscores and uppercasing, it should be `HEALTH_PROBE_ADDRESS`.

atoulme: that makes sense, thanks

atoulme: done

PR #4055, opened 2025-06-01
Support config file

atoulme:
**Description:**
Support a yaml config file to set all the configuration options exposed by controllers

#3565

**Testing:**
Unit tests.


.chloggen/support_config_file.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: controller
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Support a yaml config file to set all the configuration options exposed by controllers
```
iblancasa, line 8: Users will need to add a new volume or something with the file or just add it to their container,  right? Maybe we can document how to use this mechanism

atoulme: a ConfigMap

atoulme: I mean, you're speedrunning to where I am going - give us a chance to slowly add to the featureset and then make it possible to configure the manager this way. Possibly a helm chart down the road.

iblancasa: > you're speedrunning to where I am going

Sorry. I'm just trying to understand the changes you are implementing in order to do a proper review because some of the PRs are lacking of context.

atoulme: no worries

main.go
```diff
@@ -105,6 +108,13 @@ func main() {
 	ctrl.SetLogger(logger)
 
 	configLog := ctrl.Log.WithName("config")
+
+	if configFile != "" {
+		if err := config.ApplyConfigFile(configFile, &cfg, configLog); err != nil {
+			panic(err)
```
jaronoff97: is there a test that shows how this will work with setting things via the CLI? i.e. verifying the precedence of variable setting.

atoulme: both you and @swiatekm are asking for this iiuc ?

internal/config/config_test.go
```diff
@@ -1,142 +1,19 @@
...
-	"github.com/open-telemetry/opentelemetry-operator/internal/autodetect/collector"
-	"github.com/open-telemetry/opentelemetry-operator/internal/autodetect/opampbridge"
 	"github.com/open-telemetry/opentelemetry-operator/internal/autodetect/openshift"
-	"github.com/open-telemetry/opentelemetry-operator/internal/autodetect/prometheus"
-	"github.com/open-telemetry/opentelemetry-operator/internal/autodetect/rbac"
-	"github.com/open-telemetry/opentelemetry-operator/internal/autodetect/targetallocator"
-	"github.com/open-telemetry/opentelemetry-operator/internal/config"
 )
 
-func TestConfigChangesOnAutoDetect(t *testing.T) {
```
atoulme, line 25: this test was copied to autodetect https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/autodetect/main_test.go#L358 as part of https://github.com/open-telemetry/opentelemetry-operator/pull/4050 

Deleting the duplicate is needed to move to the config package.

jaronoff97 (approved):

PR #4056, opened 2025-06-02
Enable webhooks

atoulme:
Turn the undocumented ENABLE_WEBHOOKS env var into part of configuration so it is exposed just like any other setting.


iblancasa (commented):
Can you add a description to the PR? Also, something is failing in the CI.

swiatekm (approved):

PR #4062, opened 2025-06-02
[chore] migrate to OpenTelemetry metrics

atoulme:


cmd/otel-allocator/internal/collector/collector.go
```diff
@@ -23,33 +23,32 @@ const (
...
 }
 
-func NewCollectorWatcher(logger logr.Logger, kubeConfig *rest.Config, collectorNotReadyGracePeriod time.Duration) (*Watcher, error) {
+func NewCollectorWatcher(logger logr.Logger, meter metric.Meter, kubeConfig *rest.Config, collectorNotReadyGracePeriod time.Duration) (*Watcher, error) {
 	clientset, err := kubernetes.NewForConfig(kubeConfig)
 	if err != nil {
 		return &Watcher{}, err
 	}
 
+	collectorsDiscovered, err := meter.Int64Gauge("opentelemetry_allocator_collectors_discovered", metric.WithDescription("Number of collectors discovered."))
```
jaronoff97, line 43: i forget the recommendation, but should we just use the global meter here (and the other constructors)?

jaronoff97 (approved):
seems all good to me, just one question and approach

cmd/otel-allocator/internal/collector/collector.go
atoulme: well, we create the meter provider in main, we pass it down in the objects - maybe we are creating the gauge multiple times? I don't know.

jaronoff97: yeah im more wondering if we need to pass it down at all or just use the global provider

atoulme: there's a global provider?

cmd/otel-allocator/main.go
```diff
@@ -68,10 +66,17 @@ func main() {
 	ctx := context.Background()
 	log := ctrl.Log.WithName("allocator")
 
+	metricExporter, promErr := otelprom.New()
+	if promErr != nil {
+		panic(promErr)
+	}
+	meterProvider := sdkmetric.NewMeterProvider(sdkmetric.WithReader(metricExporter))
+	meter := meterProvider.Meter("otelallocator")
```
atoulme: @open-telemetry/go-approvers would you please help me? I am taking out prometheus gauges and introducing the metrics SDK. Is this pattern correct? Is a global provider a better approach?

cmd/otel-allocator/internal/collector/collector.go
jaronoff97: https://github.com/open-telemetry/opentelemetry-go/blob/main/metric.go#L35-L37

atoulme: ok, will use that instead!

atoulme: done.

cmd/otel-allocator/main.go
atoulme: going with a global provider per @jaronoff97 review.

iblancasa (commented):
Will this have any impact on how the metrics are provided to users? I mean: things like the name or any other conventions.

Just because, in that case, we should provide a changeling.

cmd/otel-allocator/benchmark_test.go
```diff
@@ -174,14 +174,20 @@ func createTestDiscoverer(allocationStrategy string, prehookConfig map[string][]
...
+		panic(err)
+	}
 	registry := prometheus.NewRegistry()
 	sdMetrics, _ := discovery.CreateAndRegisterSDMetrics(registry)
 	discoveryManager := discovery.NewManager(ctx, config.NopLogger, registry, sdMetrics)
-	targetDiscoverer := target.NewDiscoverer(logger, discoveryManager, allocatorPrehook, srv, allocator.SetTargets)
+	targetDiscoverer, err := target.NewDiscoverer(logger, discoveryManager, allocatorPrehook, srv, allocator.SetTargets)
+	if err != nil {
+		panic(err)
+	}
```
swiatekm: I think we should either turn this into a helper function and `require.NoError` here, or (preferable imo) return the error and assert on it in the test function itself.

cmd/otel-allocator/internal/allocation/allocator.go
```diff
@@ -25,19 +29,48 @@ import (
...
 		strategy:                      strategy,
 		collectors:                    make(map[string]*Collector),
 		targetItems:                   make(map[target.ItemHash]*target.Item),
 		targetItemsPerJobPerCollector: make(map[string]map[string]map[target.ItemHash]bool),
 		log:                           log,
+		targetsPerCollector:           targetsPerCollector,
+		collectorsAllocatable:         collectorsAllocatable,
+		timeToAssign:                  timeToAssign,
+		targetsRemaining:              targetsRemaining,
+		targetsUnassigned:             targetsUnassigned,
```
swiatekm, line 67: Is storing metrics as fields idiomatic with the Go Otel SDK?

pellared: Yes. E.g. https://github.com/open-telemetry/opentelemetry-go-contrib/blob/aee53e882ff210eb35d9fd561b3a8bc286017f91/instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go#L36-L46

PS. It is OTel Go **API** :wink:

atoulme:
> Will this have any impact on how the metrics are provided to users? I mean: things like the name or any other conventions.
> 
> Just because, in that case, we should provide a changeling.

That's not my expectation, especially as we don't change names. Don't we have tests to make sure we emit the same metrics?

cmd/otel-allocator/internal/allocation/allocator.go
atoulme: I don't know.

cmd/otel-allocator/benchmark_test.go
atoulme: can do

cmd/otel-allocator/internal/allocation/allocator.go
atoulme: should be good then :D resolving

swiatekm:
> > Will this have any impact on how the metrics are provided to users? I mean: things like the name or any other conventions.
> > Just because, in that case, we should provide a changeling.
> 
> That's not my expectation, especially as we don't change names. Don't we have tests to make sure we emit the same metrics?

We definitely have them for the operator, but I don't think we do for the target allocator.

atoulme:
Discussed in SIG - @swiatekm will file an issue requesting we first have a way to check the TA metrics, and then we can apply this test to this PR. The PR will wait until we have test coverage.

swiatekm:
> Discussed in SIG - @swiatekm will file an issue requesting we first have a way to check the TA metrics, and then we can apply this test to this PR. The PR will wait until we have test coverage.

Created the issue here: https://github.com/open-telemetry/opentelemetry-operator/issues/4064.

swiatekm:
@atoulme with https://github.com/open-telemetry/opentelemetry-operator/pull/4106, we now have E2E tests for target-allocator metrics. So if you could rebase on main and have the tests pass, I'd be happy to merge this PR.

swiatekm:
@atoulme it looks like one of the tests panics on a metric update. Can you have a look?

swiatekm (approved):

Issue #4044, opened 2025-05-28
Same target from two different jobs missing after targetallocator upgrade 0.121.0+


gracewehner:
### Component(s)

target allocator

## Description
In a practical case, this happens when two different jobs point to the same target but each job has a different metrics_path. Now the target is only scraped for one of the jobs, whereas before it was scraped for both.

I was able to change of the unit tests to reproduce this and attach a debugger. I have narrowed it down to this PR: https://github.com/open-telemetry/opentelemetry-operator/pull/3832 that changes the hash for a target from `t.JobName + t.TargetURL + strconv.FormatUint(t.Labels.Hash(), 10)` to `t.Labels.Hash()`.

The `targetgroup` map coming from the Prometheus discovery manager does not contain job as a label when the targets are being processed by the Target Allocator and stored in the target list:

<img width="417" alt="Image" src="https://github.com/user-attachments/assets/ed0deca3-e76e-41c4-8ae3-da6fec6c8e65" />

`Job` is the key to a `targetgroup` object in the map, which does have the `address` as a label per target. This is also not added a label when processing the targets right before hashing them: https://github.com/swiatekm/opentelemetry-operator/blob/main/cmd/otel-allocator/internal/target/discovery.go#L195. The metrics path is also not a label at the time.

## Steps to Reproduce
Use a scrape config pointing to the same target from two different jobs for Target Allocator version 0.121.0 and up:

```yaml
config:
  scrape_configs:
  - job_name: prometheus
    static_configs:
    - targets: ["prom.domain:9001", "prom.domain:9002", "prom.domain:9003"]
  - job_name: prometheus2
    static_configs:
    - targets: ["prom.domain:9001"]
```

## Expected Result
Target `prom.domain:9001` will be allocated to a collector for both jobs `prometheus` or `prometheus2`.

## Actual Result
Target `prom.domain:9001` will be allocated to a collector for only one job of `prometheus` or `prometheus2`, but not both.

### Kubernetes Version

1.31.8

### Operator version

v0.121.0+

### Collector version

v0.121.0+

## Environment
OS: "Ubuntu 20.04"

### Log output

```shell

```

gracewehner:
Thanks @swiatekm for the quick reply. The above debugger screenshot is for the config:
```
config:
  scrape_configs:
  - job_name: prometheus
    metrics_path: /metrics1
    static_configs:
[...]

PR #4066, opened 2025-06-03
fix: same target from two different jobs missing in targetallocator

gracewehner:
Use the Prometheus labels hash logic, but add in the job name to the hash as well. As discussed in the issue below, this is a temporary solution to have the same behavior by re-using the Prometheus code until we decide on a final implementation.

- Resolves: #4044

**Testing:** Added a unit test for this scenario. Added a test that fails with the current code and passes with the changes in this PR.

**Documentation:** None, this is returning to previous behavior.

### DIFF
--- a/cmd/otel-allocator/internal/target/discovery_test.go
+++ b/cmd/otel-allocator/internal/target/discovery_test.go
@@ -6,7 +6,6 @@ package target
 import (
 	"context"
 	"errors"
-	"fmt"
 	"hash"
 	"sort"
 	"testing"
@@ -401,10 +400,6 @@ func TestDiscoveryTargetHashing(t *testing.T) {
 			assert.NoError(t, err)
 
 			gotTargets := <-results
-			fmt.Printf("got %d targets\n", len(gotTargets))
-			for i, target := range gotTargets {
-				fmt.Printf("target %d: %s %s %s %d\n", i, target.JobName, target.TargetURL, target.Labels, target.Hash())
-			}
 
 			// Verify that all targets have different hashes
 			targetHashes := make(map[ItemHash]bool)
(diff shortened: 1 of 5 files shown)


cmd/otel-allocator/internal/target/discovery_test.go
```diff
@@ -341,6 +348,73 @@ func TestDiscovery_ScrapeConfigHashing(t *testing.T) {
...
+	go func() {
+		err := manager.Run()
+		assert.NoError(t, err)
+	}()
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			cfg := config.CreateDefaultConfig()
+			err := config.LoadFromFile(tt.args.file, &cfg)
+			assert.NoError(t, err)
+			assert.True(t, len(cfg.PromConfig.ScrapeConfigs) > 0)
```
swiatekm: Is there a need to load the full config file here vs defining the scrape configs as Go structs in the test body?

go.mod
```diff
@@ -227,6 +227,7 @@ require (
 )
 
 require (
+	github.com/cespare/xxhash v1.1.0
```
swiatekm: Prometheus uses `github.com/cespare/xxhash/v2 v2.3.0`, so we should do the same. Ideally, this change shouldn't add any new dependencies to go.sum.

swiatekm (commented):
Looks good to me, thank you for the contribution! I left two comments, but they're relatively minor.

You'll also need to add a changelog entry. `make chlog-new` will generate a template for you.

swiatekm (approved):

swiatekm:
@gracewehner we wanted to get this change into the next release, so I'd fixed the remaining lint errors myself. Sorry for that! Thank you for the fix, it's very appreciated.

PR #4070, opened 2025-06-03
[manager] Do not register the opampbridge webhook if the CRD is not present

atoulme:
**Description:**
Check for CRD presence ; if absent, do not register the webhook.


iblancasa (approved):

swiatekm (approved):
This change looks good to me, but I feel like we have some opportunity to factor out common logic for checking if a CRD exists and starting the webhook and manager conditionally.

jaronoff97 (approved):

Issue #4071, opened 2025-06-04
`spec.managementState` fields should not be "required"


ordovicia:
## Description

`spec.managementState` fields in CRDs are marked as required, but also has a default value.

https://github.com/open-telemetry/opentelemetry-operator/blob/63322cf3ee946045c12d05e01cf587cdc83ca0cf/apis/v1beta1/common.go#L88-L94

I think these "required" markers should be removed because:

- Not practically effective. Users can create custom resources with `spec.managementState` omitted. Kubernetes API server injects `spec.managementState: managed` on creation because the field has the default value.
- Incompatible with linter tools like [kubeconform](https://github.com/yannh/kubeconform). Since `spec.managementState` field is marked as required, kubeconform considers a custom resource without `spec.managementState` field to be invalid. However, users can still create these resources with no errors.

## Steps to Reproduce

N/A

## Expected Result

`spec.managementState` fields only have "default" markers.

## Actual Result

`spec.managementState` fields have "required" markers in addition to "default" ones.

### Kubernetes Version

1.32.5

### Operator version

0.126.0

### Collector version

N/A

## Environment
OS: (e.g., "Ubuntu 20.04")
Compiler(if manually compiled): (e.g., "go 14.2")

### Log output

```shell

```

LucaLanziani:
@iblancasa I'll take this one if you don't mind

PR #4076, opened 2025-06-04
remove required annotation from spec.managementState

LucaLanziani:
**Description:** 

Remove required annotation since default is present.

- Resolves: #4071

**Testing:**

```
make test
```

### DIFF
--- a/.chloggen/issue_4071.yaml
+++ b/.chloggen/issue_4071.yaml
@@ -1,5 +1,5 @@
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: 'bug_fix'
+change_type: 'enhancement'
 
 # The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
 component: managementState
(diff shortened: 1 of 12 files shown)


iblancasa (approved):

apis/v1beta1/common.go
```diff
@@ -88,7 +88,6 @@ type OpenTelemetryCommonFields struct {
 	// ManagementState defines if the CR should be managed by the operator or not.
 	// Default is managed.
 	//
-	// +required
 	// +kubebuilder:validation:Required
```
swiatekm: You need to delete this line as well, and then run `make update`. Right now, your change doesn't do anything other than removing a comment.

swiatekm (changes requested):

apis/v1beta1/common.go
LucaLanziani: ah, you are right @swiatekm let me do that

.chloggen/issue_4071.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'bug_fix'
```
LucaLanziani: @swiatekm not sure if this should be a bug_fix or an enhancement tbf

swiatekm: I'd say enhancement. The CRD is now a bit nicer to work with, even if the logic didn't change.

swiatekm (approved):

PR #4080, opened 2025-06-05
Do not register reconciler if CRD is not present

atoulme:
A continuation of #4070

This controller eventually times out when running without the CRD present and kills the manager.


atoulme:
skipping changelog as this would be applied under the same changelog entry as #4070

Issue #4064, opened 2025-06-02
Add E2E test for otel collector and target allocator observability


swiatekm:
We currently don't have an E2E test for checking if [`spec.observability.metrics.enableMetrics`](https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/api/opentelemetrycollectors.md#opentelemetrycollectorspecobservabilitymetrics) creates ServiceMonitor and PodMonitor definitions which actually allow the metrics to be scraped correctly. Add a test that verifies this. This will require enabling the target allocator so the otel collector can scrape its own metrics, so it may as well check the same setting for target allocator itself.

See https://github.com/open-telemetry/opentelemetry-operator/issues/3990 for how this can be done using chainsaw.

carsontham:
i will work on this :)

swiatekm:
@carsontham thank you for taking this up. To give a bit more detail, for collector metrics we just want to check if we can get any metrics at all - it's not a project we own, so we don't have to verify what those are exactly. For the target allocator, we want to check exhaustively, same as we did for the operator. It's fine to do these in separate PRs, but I'll leave that decision up to you. Does that make sense?

carsontham:
hey @swiatekm ! i have created a PR [here](https://github.com/open-telemetry/opentelemetry-operator/pull/4106) for asserting target allocator metrics. Will follow up with another one to test the collector after this PR is approved :)

PR #4106, opened 2025-06-15
[chore] Added e2e tests for target allocator metrics

carsontham:

Added e2e tests to assert target allocator metrics

- Resolves: #4064

Added e2e tests to assert target allocator metrics


tests/e2e/targetallocator-metrics/00-install.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: opentelemetry.io/v1beta1
+kind: OpenTelemetryCollector
+metadata:
+  name: simplest
+  namespace: opentelemetry-operator-system
```
swiatekm: We shouldn't hardcode this namespace. Chainsaw will take care of creating a temporary one.

tests/e2e/targetallocator-metrics/00-install.yaml
```diff
@@ -0,0 +1,32 @@
...
+  labels:
+    app.kubernetes.io/name: collector-with-target-allocator
```
swiatekm: Is this label necessary?

tests/e2e/targetallocator-metrics/00-install.yaml
```diff
@@ -0,0 +1,32 @@
...
+spec:
+  mode: statefulset
+  targetAllocator:
+    enabled: true
+  observability:
+    metrics:
+      enableMetrics: true
```
swiatekm: If we want to test the ServiceMonitor automatically generated for the target allocator, then this block should be under `targetAllocator:`.

tests/e2e/targetallocator-metrics/00-install.yaml
```diff
@@ -0,0 +1,32 @@
...
+  config:
+    receivers:
+      prometheus:
+        config:
+          scrape_configs:
+            - job_name: "otel-collector"
+              scrape_interval: 10s
+              static_configs:
+                - targets: ["0.0.0.0:8888"]
```
swiatekm: Instead of adding this fixed scrape config, we should enable metrics and Prometheus CRs for the target allocator. Then the target allocator will find itself as the target based on the ServiceMonitor we create, and the otel collector can scrape the metrics.

tests/e2e/targetallocator-metrics/00-install.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+    exporters:
+      debug: {}
```
swiatekm: We should instead have a prometheus exporter here, and the test should use its endpoint instead of the target allocator's endpoint directly. Then we're not just testing whether target allocator exposes the right metrics, but also `targetAllocator.observability.metrics.enableMetrics`.

swiatekm (changes requested):
I intended my previous review to be blocking, adding this one to ensure this PR is not merged as-is.

tests/e2e/targetallocator-metrics/00-install.yaml
carsontham: fixed!

tests/e2e/targetallocator-metrics/00-install.yaml
carsontham: removed :)

tests/e2e/targetallocator-metrics/00-install.yaml
carsontham: fixed!

tests/e2e/targetallocator-metrics/00-install.yaml
carsontham: fixed and enabled PrometheusCR

tests/e2e/targetallocator-metrics/00-install.yaml
carsontham: fixed!

tests/e2e/targetallocator-metrics/chainsaw-test.yaml
```diff
@@ -0,0 +1,159 @@
...
+                name: ready
+                value: 'true'
+        - assert:
+            file: 00-assert.yaml
+    - name: Assert Target Allocator metrics
+      try:
+        - command:
+            env:
+              - name: namespace
+                value: ($namespace)
```
swiatekm: Is this necessary? I thought the built-in variables were always available.

swiatekm (commented):
Looks good to me! One final request I have is that you move this test from `tests/e2e` to `tests/e2e-targetallocator`.

tests/e2e/targetallocator-metrics/chainsaw-test.yaml
carsontham: fixed!

swiatekm:
@iblancasa can you have a second look?

Issue #4124, opened 2025-06-20
Release the operator v0.128.0


bogdandrutu:
See https://github.com/open-telemetry/opentelemetry-collector/issues/13179

Issue #3911, opened 2025-04-15
OpenTelemetry Operator v0.120 -> v0.122: Upgrade retains deprecated telemetry.metrics.address field causing port conflicts and collector crashes


dschaaff:
## Description

When upgrading OpenTelemetry Operator from v0.120 to v0.122, existing OpenTelemetryCollector resources maintain the deprecated `telemetry.metrics.address` field while also adding the new `telemetry.metrics.readers` configuration, resulting in both configurations being present simultaneously. This causes collectors to crash with "address already in use" errors as they attempt to bind the same port twice.

## Steps to Reproduce

1. Create an opentelemetrycollector manfiest on version 0.120

```yaml
apiVersion: opentelemetry.io/v1beta1
kind: OpenTelemetryCollector
metadata:
  name: prom-exporter-sidecar
  namespace: opentelemetry-operator
spec:
  mode: sidecar
  resources:
    requests:
      memory: 50Mi
      cpu: 5m
    limits:
      memory: 150Mi
  config:
    receivers:
      otlp:
        protocols:
          grpc: {}
          http: {}

    processors:
      memory_limiter:
        check_interval: 5s
        limit_percentage: 80
        spike_limit_percentage: 25

      resourcedetection/env: {}

      batch: {}

    exporters:
      prometheus:
        endpoint: "0.0.0.0:9317"
        resource_to_telemetry_conversion:
          enabled: true
        metric_expiration: 10m

    extensions:
      health_check: {}

    service:
      pipelines:
        metrics:
          receivers: [otlp]
          processors: [memory_limiter, resourcedetection/env, batch]
          exporters: [prometheus]

```
2. Apply the manifest
3. Upgrade the operator to 0.122.0 (helm chart 0.86)
4. Observe the collector now contains a merge of the old and new formats 

```
service:
     telemetry:
       metrics:
         address: 0.0.0.0:8888
         readers:
         - pull:
             exporter:
               prometheus:
                 host: 0.0.0.0
                 port: 8888
```

## Expected Result

The operator should have remove service.telemetry.metrics.address from the generated config.

## Actual Result
service.telemetry.metrics.address is merged with the new metrics reader config.

Here are the managed fields from the object in k8s for reference 

```yaml
 managedFields:
  - apiVersion: opentelemetry.io/v1beta1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:labels:
[...]

PR #4134, opened 2025-06-25
Prepare release 0.129.0

yuriolisa:

Prepare release 0.128.0

- Resolves: #4124 
- Resolves https://github.com/open-telemetry/opentelemetry-operator/issues/4153 
- Resolves https://github.com/open-telemetry/opentelemetry-operator/issues/3911# 


iblancasa:
> @open-telemetry/operator-approvers how do we want to deal with this? We can either add a workaround and keep it for some number of versions, or simply not release 0.128.0 and wait for 0.129.0.

I would say the workaround makes sense.

pavolloffay:
> @open-telemetry/operator-approvers how do we want to deal with this? We can either add a workaround and keep it for some number of versions, or simply not release 0.128.0 and wait for 0.129.0.

We should have asked for 0.128.1 of the collector

swiatekm:
> > @open-telemetry/operator-approvers how do we want to deal with this? We can either add a workaround and keep it for some number of versions, or simply not release 0.128.0 and wait for 0.129.0.
> 
> We should have asked for 0.128.1 of the collector

That ship has sailed at this point I'm afraid. We should make an effort to catch this kind of problem early. Bumping the collector version in versions.txt as soon as it's available should be straightforward. We can also run e2e tests daily with the dev version of contrib.

But for now, we do need to decide how to deal with this problem. A workaround would involve setting the missing attribute on any scrape config emitted by the target allocator, which shouldn't be too onerous.

swiatekm:
I'm not even sure if it's *possible* for us to work around this problem right now in a non-hacky way. Even after setting `metric_name_validation_scheme`, the receiver still complains about `metric_name_escaping_scheme`:

```
2025-06-30T15:59:41.642Z    error    error creating new scrape pool    {"resource": {"service.instance.id": "ed44120e-6c01-48b3-857e-37cc8fd1d7ab", "service.name": "otelcol", "service.version": "0.128.0"}, "otelcol.component
.id": "prometheus", "otelcol.component.kind": "receiver", "otelcol.signal": "metrics", "err": "invalid metric name escaping scheme, got empty string instead of escaping scheme", "scrape_pool": "serviceMonitor/chainsaw-patien
t-minnow/prometheus-cr/0"}  
```

This is a value that only exists in prometheus config as of 3.4.0, and we can't upgrade to that version because the respective prometheus-operator requires Go 1.24. The only way we can do it is by intercepting the values during serialization in the target allocator's HTTP server. There's already a similar hack we do there for regexes, but it's pretty dirty. I'm not sure if it isn't better to just skip 0.128.0.

swiatekm:
Here's how the hacky workaround looks: https://github.com/open-telemetry/opentelemetry-operator/pull/4154.

yuriolisa:
> The collector 0.129.0 is out I am fine skipping the 0.128.0 release and don't introduce the workaround.

So, I will release the 0.129.0 today.

RELEASE.md
```diff
@@ -43,11 +43,11 @@ In order to have more people comfortable with the release process, and in order
 The operator should be released within a week after the [OpenTelemetry collector release](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/release.md#release-schedule).
 
 | Version  | Release manager |
-|----------|-----------------|
-| v0.128.0 | @yuriolisa      |  
+|----------|-----------------|
```
pavolloffay: The table should also be updated to reflect skipped 0.128.0 release

pavolloffay:
@yuriolisa there are merge conflicts

yuriolisa:
> @yuriolisa there are merge conflicts

Just fixed.

yuriolisa:
> @yuriolisa the 0.129.0 release failed, you'll need to update to 0.129.1.
> 
> Could you also regenerate the changelog to include #4041? That's my bad, I merged I saw you recreated this PR - sorry about that!

@swiatekm, no problem!! Just submitted the update to 0.129.1. And the #4041 is there, but pointing to the issue instead.

swiatekm:
> > @yuriolisa the 0.129.0 release failed, you'll need to update to 0.129.1.
> > Could you also regenerate the changelog to include #4041? That's my bad, I merged I saw you recreated this PR - sorry about that!
> 
> @swiatekm, no problem!! Just submitted the update to 0.129.1. And the #4041 is there, but pointing to the issue instead.

You've still got some e2e test failures due to deprecated configuration. I fixed them [here](https://github.com/swiatekm/opentelemetry-operator/commit/b86082e4c4b0390d0362db57b3a6711757a0d9e9) if you want to just pull in the changes.

swiatekm (approved):

pavolloffay:
@yuriolisa could you please rebase and re-generate the changelog? I have merged https://github.com/open-telemetry/opentelemetry-operator/pull/4158

pavolloffay (approved):

PR #4168, opened 2025-07-07
[chore] Move k8s client init to target allocator main

swiatekm:
Initialize the k8s clients in the main function and pass them to objects which need them. This way we can reuse the clients and test more easily.


jaronoff97 (approved):
other than the GCI, this is fine

iblancasa (approved):

Issue #4138, opened 2025-06-27
Add `extraLabels` to `spec.observability.metrics`


yuzhouliu9:
### Component(s)

collector

### Is your feature request related to a problem? Please describe.

When an OpenTelemetryCollector is configured with the following
```yaml
spec:
  observability:
    metrics:
      enableMetrics: true
```

It will create a ServiceMonitor. However there is no configuration to add [extraLabels](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/values.yaml#L133) like in the case of the opentelemetry-operator.

Our prometheus instance is configured to only pick up ServiceMonitors and PodMonitors with a specific label. So the ServiceMonitors created from OpenTelemetryCollector are not being picked up as a target by prometheus right now.

Could this be added as a configuration parameter?

Thanks!

### Describe the solution you'd like

New field under metrics named `extraLabels` that is a map[string]string
```yaml
spec:
  observability:
    metrics:
      enableMetrics: true
      serviceMonitor:
        extraLabels:
          prometheus: kube-prometheus
```

### Describe alternatives you've considered

Create the ServiceMonitor object manually instead of relying on OpenTelemetryCollector

yuzhouliu9:
Thank you so much @Horiodino !

PR #4174, opened 2025-07-09
[chore] Add extra labels to  spec observability metrics

Horiodino:
**Description:**  
Added support for `extraLabels` in `spec.observability.metrics` to allow users to attach additional labels to the generated ServiceMonitor.  

Fixes: https://github.com/open-telemetry/opentelemetry-operator/issues/4138#event-18534809348

- Resolves: #4138

**Testing:**  
- Added unit tests to cover cases:
  - With empty `extraLabels`
  - With custom `extraLabels` map containing multiple labels
- Verified generated ServiceMonitor contains the expected labels.

**Documentation:**  
- Updated API docs to document the new `extraLabels` field under `spec.observability.metrics`.
- Regenerated CRDs and bundle manifests to include the new field.


internal/manifests/collector/servicemonitor.go
```diff
@@ -111,3 +113,10 @@ func endpointsFromConfig(logger logr.Logger, otelcol v1beta1.OpenTelemetryCollec
 	}
 	return endpoints
 }
+
+// addExtraLabels adds extra labels to the ServiceMonitor labels map.
+func addExtraLabels(labels *map[string]string, extraLabels map[string]string) {
+	for k, v := range extraLabels {
```
iblancasa: Maybe we should log something if the label already exists. Because this will overwrite the value.

iblancasa (dismissed):
Please add:
- a changelog
- improve the e2e test to check this new feature

Horiodino:
@iblancasa could you please cross-check https://github.com/open-telemetry/opentelemetry-operator/issues/4184?
I thought it might be related to the changes from this PR, but I already tested it locally and didn’t get any errors:

```make test

✓  internal/instrumentation (21.838s)
✓  internal/controllers (40.73s)

=== Skipped
=== SKIP: internal/controllers TestRegisterWithManager_OpAMPBridge (0.00s)
    opampbridge_controller_test.go:159: this test requires a real cluster, otherwise the GetConfigOrDie will die

=== SKIP: internal/controllers TestRegisterWithManager (0.00s)
    reconcile_test.go:1188: this test requires a real cluster, otherwise the GetConfigOrDie will die

DONE 1795 tests, 2 skipped in 55.687s
```

internal/manifests/collector/servicemonitor.go
```diff
@@ -49,6 +49,8 @@ func createServiceMonitor(name string, params manifests.Params, serviceType Serv
 	var sm monitoringv1.ServiceMonitor
 
 	labels := manifestutils.Labels(params.OtelCol.ObjectMeta, name, params.OtelCol.Spec.Image, ComponentOpenTelemetryCollector, []string{})
+	// Add extra labels to the ServiceMonitor
+	addExtraLabels(&params.Log, &labels, params.OtelCol.Spec.Observability.Metrics.ExtraLabels)
```
swiatekm: You've added the new `extraLabels` field to a common struct which affects both the OpenTelemetryCollector and TargetAllocator CRDs. This means you also need to implement this change for ServiceMonitors created by the latter [here](https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/manifests/targetallocator/servicemonitor.go).

This will also require moving your helper function to the `manifestutils` package and updating E2E tests.

tests/e2e-prometheuscr/create-sm-prometheus-exporters/09-assert.yaml
```diff
@@ -0,0 +1,122 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
```
swiatekm, line 4: For all of these asserts, do we need to check the whole ServiceMonitor? It looks like checking the labels should be sufficient.

Horiodino: I added it just to be safe in case it didn’t break anything , no problem, I’ll minimize the assert.

internal/manifests/manifestutils/labels.go
```diff
@@ -88,3 +89,15 @@ func TASelectorLabels(instance v1alpha1.TargetAllocator, component string) map[s
 	}
 	return selectorLabels
 }
+
+// AddExtraLabels adds extra labels to the ServiceMonitor labels map.
+func AddExtraLabels(logger *logr.Logger, labels *map[string]string, extraLabels map[string]string) {
```
swiatekm:
```suggestion
func AddExtraLabels(logger *logr.Logger, labels map[string]string, extraLabels map[string]string) {
```

Maps are already reference types, you don't need a pointer here.

internal/manifests/manifestutils/labels.go
```diff
@@ -88,3 +89,15 @@ func TASelectorLabels(instance v1alpha1.TargetAllocator, component string) map[s
 	}
 	return selectorLabels
 }
+
+// AddExtraLabels adds extra labels to the ServiceMonitor labels map.
+func AddExtraLabels(logger *logr.Logger, labels map[string]string, extraLabels map[string]string) {
+	for k, v := range extraLabels {
+		if existingValue, exists := (labels)[k]; exists {
+			if existingValue != v {
+				logger.Info("Overwriting label", "label", k, "old value", existingValue, "new value", v)
```
frzifus, line 98: Should that maybe be a warning? That would highlight it in any case.

Horiodino: actually its interface from the github.com/go-logr/logr package does NOT provide a Warn method.

Horiodino:
Thank you, I appreciate your guidance! If there are any other issues you’d like to assign , that would be wonderful

PR #4179, opened 2025-07-10
Ensure correct CodeQL workflow permissions

trask:
Related to https://github.com/open-telemetry/sig-security/issues/148#issuecomment-3051569200.

cc @trask

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -76,7 +76,7 @@ jobs:
 
   security:
     permissions:
-      security-events: write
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yaml
```diff
@@ -75,6 +75,8 @@ jobs:
       run: make lint
 
   security:
+    permissions:
+      security-events: write
```
trask:
```suggestion
      security-events: write # for github/codeql-action/analyze to upload SARIF results
```

swiatekm (approved):

PR #4185, opened 2025-07-11
fix(config): enable metrics in controller-manager

sergeysedoy97:
**Problem:** For now there is no way to collect manager's metrics


iblancasa (changes requested):
Missing changeling. Also, can you add a test to detect regressions in the future?

PR #4197, opened 2025-07-16
[chore]: prepare v0.130.0 release

TylerHelmuth:


bundle/community/manifests/opentelemetry.io_instrumentations.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.18.0
+    controller-gen.kubebuilder.io/version: v0.17.1
```
swiatekm: @TylerHelmuth you should delete your local bin/controller-tools and run `make update`. Unfortunately our tooling doesn't do this automatically.

TylerHelmuth:
I've identified the issue is something in v0.130.0.

The extra `_span_` is the [unit](https://github.com/open-telemetry/opentelemetry-collector/blob/8183eb10d1a220f83bdee4c8de72f72b272f8fa1/exporter/exporterhelper/metadata.yaml#L21) of the metric. (I confirmed this by running a collector locally with a modified unit and seeing it show up)

On v0.130.0 if you ever set the a prometheus exporter yourself you need to disable units:

```yaml
  telemetry:
    metrics:
      readers:
        - pull:
            exporter:
              prometheus:
                host: 0.0.0.0
                port: 8888
                without_units: true
```

If you're using the collector's default prometheus exporter things work as expected.

This is a change from v0.129.1. In v0.129.1 if I use the config

 ```yaml 
  telemetry:
    metrics:
      readers:
        - pull:
            exporter:
              prometheus:
                host: 0.0.0.0
                port: 8888
```

I see `otelcol_exporter_send_failed_spans_total`, but if I use the same config in v0.130.0 I see `otelcol_exporter_send_failed_spans__{unit}__total`.

I am still investigating why this changed.

TylerHelmuth:
The issue is https://github.com/open-telemetry/opentelemetry-go/issues/7039. For now I'll update our tests to manually set `without_units: true`

swiatekm:
Another reason to run E2E tests on the latest contrib image daily. Would've caught it before the release.

TylerHelmuth:
Should we solve this for our users with an upgrade script that adds `without_units: true` to prometheus exporters that dont already have it set to `false`?

swiatekm:
> Should we solve this for our users with an upgrade script that adds `without_units: true` to prometheus exporters that dont already have it set to `false`?

That would be ideal, yeah. Probably should be a separate PR though.

TylerHelmuth:
@swiatekm I expected the upgrade script to handle inserting the `without_units` field on the e2e tests. Was the a bad assumption?

TylerHelmuth:
Ok I see another issue, we also need to update AddPrometheusMetricsEndpoint to add the field for the default prometheus exporter we're setting.

TylerHelmuth:
The failing tests seem related to the target allocator, but I cant see why yet.

swiatekm:
> The failing tests seem related to the target allocator, but I cant see why yet.

I'll have a look later today.

PR #4198, opened 2025-07-17
[fix]: Add update script for v0.130.0

TylerHelmuth:
**Description:**
Add update script for v0.130.0 of the collector.

Due to https://github.com/open-telemetry/opentelemetry-go/issues/7039, the collector will emit new metric names for some of its internal metrics if the user has configured their own prometheus exporter. The workaround is to set `without_units` on the configured prometheus exporter.

To ensure users don't experience a breaking change in the telemetry, the upgrade script sets `without_units` on any configured prometheus exporter that has not already set `without_units`. This ensures that the user gets the same experience as the v0.129.1 collector.

- Related to https://github.com/open-telemetry/opentelemetry-go/issues/7039

**Testing:** 

Added tests

### DIFF
--- a/pkg/collector/upgrade/v0_130_0_test.go
+++ b/pkg/collector/upgrade/v0_130_0_test.go
@@ -6,10 +6,11 @@ package upgrade
 import (
 	"testing"
 
-	"github.com/open-telemetry/opentelemetry-operator/apis/v1beta1"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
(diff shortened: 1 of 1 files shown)


pkg/collector/upgrade/v0_130_0.go
```diff
@@ -0,0 +1,32 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package upgrade
+
+import (
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1beta1"
+)
```
TylerHelmuth, line 8: Why is the linter upset about this?

TylerHelmuth: got my files mixed up

swiatekm (approved):

PR #4201, opened 2025-07-18
[manager] Run the mutating webhook with configuration, disabling instrumentation CRDs.

atoulme:
With this PR, we are able to deploy the mutating webhook using exclusively static configuration rather than instrumentation CRDs.

There are 3 parts to it:
* Add the instrumentation CR to the config struct, so it can be fulfilled.
* Add an option to disable looking up CRDs for instrumentations, instead relying on what the config says.
* Add an integration test showing this works end to end - this test runs outside of the current e2e tests and performs the installation step explicitly, which is not possible right now with the existing integration tests.

### DIFF
--- a/.chloggen/crds_wip.yaml
+++ b/.chloggen/crds_wip.yaml
@@ -1,0 +1,35 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: auto-instrumentation
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Allow to run the mutating webhook using static configuration, without the need for CRDs.
+
+# One or more tracking issues related to the change
+issues: [4201]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  With this change, you can deploy the manager as a mutating webhook without setting up a v1alpha1.Instrumentation custom resource
+  or the v1alpha1.Instrumentation CRD.
+  
+  Instead, you can now set up instrumentation by configuring the manager via its config file with these settings:
+  ```yaml
+    ignore-missing-collector-crds: true 
+    enable-instrumentation-crds: false # Ignore that the CRD is not registered.
+    enable-multi-instrumentation: false
+    instrumentations: # Static configuration for our instrumentation
+      spec:
+        exporter:
+          endpoint: http://collector.default.svc:4318
+        propagators:
+          - tracecontext
+          - baggage
+          - b3
+        java:
+          image: "java-autoinstrumentation:dev"
+  ```
--- a/config/no-crds/kustomization.yaml
+++ b/config/no-crds/kustomization.yaml
@@ -1,0 +1,61 @@
+# Adds namespace to all resources.
+namespace: opentelemetry-operator-system
+
+# Value of this field is prepended to the
+# names of all resources, e.g. a deployment named
+# "wordpress" becomes "alices-wordpress".
+# Note that it should also match with the prefix (text before '-') of the namespace
+# field above.
+namePrefix: opentelemetry-operator-
+
+# Labels to add to all resources and selectors.
+
+# The /metrics endpoint is protected by controller-runtime's built-in
+# authentication and authorization using the FilterProvider option.
+
+
+# the following config is for teaching kustomize how to do var substitution
+vars:
+- fieldref:
+    fieldPath: metadata.namespace
+  name: CERTIFICATE_NAMESPACE
+  objref:
+    group: cert-manager.io
+    kind: Certificate
+    name: serving-cert
+    version: v1
+- fieldref: {}
+  name: CERTIFICATE_NAME
+  objref:
+    group: cert-manager.io
+    kind: Certificate
+    name: serving-cert
+    version: v1
+- fieldref:
+    fieldPath: metadata.namespace
+  name: SERVICE_NAMESPACE
+  objref:
+    kind: Service
+    name: webhook-service
+    version: v1
+- fieldref: {}
+  name: SERVICE_NAME
+  objref:
+    kind: Service
+    name: webhook-service
+    version: v1
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../rbac
+- ../manager
+- ../webhook
+- ../certmanager
+- config.yaml
+labels:
+- includeSelectors: true
+  pairs:
+    app.kubernetes.io/name: opentelemetry-operator
+patches:
+- path: manager_webhook_patch.yaml
+- path: webhookcainjection_patch.yaml
--- a/main.go
+++ b/main.go
@@ -369,10 +369,12 @@ func main() {
 		}
 	}
 
-	err = addDependencies(ctx, mgr, cfg)
-	if err != nil {
-		setupLog.Error(err, "failed to add/run bootstrap dependencies to the controller manager")
-		os.Exit(1)
+	if cfg.EnableInstrumentationCRDs {
+		err = addInstrumentationUpgrader(ctx, mgr, cfg)
+		if err != nil {
+			setupLog.Error(err, "failed to add/run bootstrap dependencies to the controller manager")
+			os.Exit(1)
+		}
 	}
 
 	var collectorReconciler *controllers.OpenTelemetryCollectorReconciler
@@ -636,7 +638,7 @@ func enableOperatorNetworkPolicy(cfg config.Config, clientset kubernetes.Interfa
 	return nil
 }
 
-func addDependencies(_ context.Context, mgr ctrl.Manager, cfg config.Config) error {
+func addInstrumentationUpgrader(_ context.Context, mgr ctrl.Manager, cfg config.Config) error {
 	// adds the upgrade mechanism to be executed once the manager is ready
 	err := mgr.Add(manager.RunnableFunc(func(c context.Context) error {
 		u := instrumentationupgrade.NewInstrumentationUpgrade(
--- a/tests/e2e-no-crds/instrumentation-no-crds/00-install-collector.yaml
+++ b/tests/e2e-no-crds/instrumentation-no-crds/00-install-collector.yaml
@@ -1,0 +1,78 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: collector-config
+data:
+  config.yaml: |
+    receivers:
+      otlp:
+        protocols:
+          grpc:
+            endpoint: "0.0.0.0:4317"
+          http:
+            endpoint: "0.0.0.0:4318"
+    exporters:
+      debug:
+        verbosity: detailed
+    service:
+      pipelines:
+        metrics:
+          receivers: [otlp]
+          exporters: [debug]
+        traces:
+          receivers: [otlp]
+          exporters: [debug]
+        logs:
+          receivers: [otlp]
+          exporters: [debug]
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: collector
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: collector
+  template:
+    metadata:
+      name: collector
+      labels:
+        app: collector
+    spec:
+      automountServiceAccountToken: false
+      containers:
+        - image: otel/opentelemetry-collector-contrib:latest
+          name: collector
+          imagePullPolicy: IfNotPresent
+          volumeMounts:
+            - name: config-volume
+              mountPath: /etc/otelcol-contrib/config.yaml
+              subPath: config.yaml
+          ports:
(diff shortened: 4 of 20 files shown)


internal/config/instrumentation.go
```diff
@@ -0,0 +1,412 @@
...
+	// XRay represents AWS X-Ray.
+	XRay Propagator = "xray"
+	// OTTrace represents OT Trace.
+	OTTrace Propagator = "ottrace"
+	// None represents automatically configured propagator.
+	None Propagator = "none"
+)
+
+// InstrumentationSpec defines the desired state of OpenTelemetry SDK and instrumentation.
+type InstrumentationSpec struct {
```
jaronoff97: can we not reuse the instrumentation spec from the CRD package?

atoulme: you cannot because you get import cycles.

If we choose to have Config define the instrumentation CRD struct, we'd need to move it, and we no longer would generate it from the CRD.

jaronoff97:
I don't think that makes sense to me – if the CRDs are disabled we shouldn't run autoinstrumentation because we have nothing defined?

internal/config/instrumentation.go
swiatekm: This dependency is mostly incidental though. The CRD itself doesn't depend on the config struct, the webhooks do, and they just happen to be in the same package. There's several things we can do to fix this, like moving webhook implementations to a different package, or having them depend on their own config struct rather than the full operator one.

I would also much rather not define the instrumentation spec twice.

atoulme: Is it? I assumed the Struct was derived from the CRD generation step. I can look further and break this down into smaller chunks.

atoulme: Are the types meant to be exported as Go API for some reason or is it ok to not expose them?

atoulme: a very rough cut - moving the types out to an internal package here: https://github.com/open-telemetry/opentelemetry-operator/pull/4796 @swiatekm is this in line with what you asked for here? I feel like I miss something.

jaronoff97: going to review that new pr and then come back here.

atoulme: good to go now, rebased on top of #4846

jaronoff97 (approved):
heck yes! way better!

.chloggen/crds_wip.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: webhook
```
pavolloffay: I think this should be auto-instrumentation

.chloggen/crds_wip.yaml
```diff
@@ -0,0 +1,16 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Allow to run the mutating webhook using static configuration, without the need for CRDs.
+
+# One or more tracking issues related to the change
+issues: [4201]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
pavolloffay: Can you share more details how it is configured, and when this is useful ?

.github/workflows/integration-test.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - name: make java autoinstrumentation image
+        run: |
+          cd autoinstrumentation/java && docker build -t java-autoinstrumentation:dev --build-arg version=$(cat version.txt) .
+      - name: Create kind cluster
+        uses: helm/kind-action@v1.11.0
+        with:
+          cluster_name: kind
+          node_image: kindest/node:v1.27.3
+      - name: Deploy certificate signing request
+        run: kubectl apply -f tests/integration/certs.yaml
```
pavolloffay: I would prefer to run these test via Makefile and run them as we run the other tests.

.github/workflows/integration-test.yaml
```diff
@@ -0,0 +1,49 @@
...
+        run: kubectl apply -f tests/integration/collector.yaml
+      - name: Deploy operator
+        run: kubectl apply -f tests/integration/operator.yaml
+      - name: Sleep 10s
+        run: sleep 10
+      - name: Deploy java app
+        run: kubectl apply -f tests/integration/java.yaml
+      - name: Sleep 20s
+        run: sleep 20
+      - name: Check logs
```
pavolloffay: Maybe I am missing something, but does this e2e test do any assertion?

.github/workflows/integration-test.yaml
atoulme: you cannot run it like the other tests since all other tests first install CRDs then run tests. I can certainly move this to a make target and call make from the test.

.github/workflows/integration-test.yaml
atoulme: `echo ${LOGS} | grep "Starting Servlet engine"` 

https://linux.die.net/man/1/grep says
> the exit status is 0 if selected lines are found and 1 otherwise.

.chloggen/crds_wip.yaml
atoulme:
```suggestion
component: auto-instrumentation
```

.chloggen/crds_wip.yaml
atoulme: added explanation

.github/workflows/integration-test.yaml
atoulme: OK I am midway through integrating with chainsaw now...

atoulme:
Ready for another round @pavolloffay

internal/instrumentation/upgrade/upgrade.go
```diff
@@ -65,13 +65,16 @@ func NewInstrumentationUpgrade(client client.Client, logger logr.Logger, recorde
...
 	list := &v1alpha1.InstrumentationList{}
-	if err := u.Client.List(ctx, list); err != nil {
-		return fmt.Errorf("failed to list: %w", err)
+	if cfg.EnableInstrumentationCRDs {
+		u.Logger.Info("looking for managed Instrumentation instances to upgrade")
+		if err := u.Client.List(ctx, list); err != nil {
+			return fmt.Errorf("failed to list: %w", err)
+		}
+	} else {
+		list.Items = []v1alpha1.Instrumentation{cfg.Instrumentation}
```
swiatekm: Is this right? I would've expected that disabling the CRD would also imply disabling the upgrades.

If we do want this, I think it'd be nicer to save these values in the InstrumentationUpgrade struct, whose constructor already takes the config as a parameter.

internal/instrumentation/upgrade/upgrade_test.go
```diff
@@ -94,7 +96,7 @@ func TestUpgrade(t *testing.T) {
 	}
 	up := NewInstrumentationUpgrade(k8sClient, ctrl.Log.WithName("instrumentation-upgrade"), &record.FakeRecorder{}, cfg)
 
-	err = up.ManagedInstances(context.Background())
+	err = up.ManagedInstances(context.Background(), cfg)
```
swiatekm: Since we're already changing this, let's do t.Context().

tests/e2e-no-crds/instrumentation-no-crds/00-install-operator.yaml
```diff
@@ -0,0 +1,144 @@
+apiVersion: apps/v1
```
swiatekm: Writing tests this way is an antipattern and duplicates a lot of what we already made common in the Makefile. Historically, tests like these, which spawn another operator deployment, have been a source of a lot of flakiness. I'm not going to block the PR over this, but I'd like a follow-up to encapsulate the logic of deploying the operator with a config file into the `prepare-e2e-no-crds` make target.

swiatekm (commented):
Looks mostly good to me, what I primarily don't like is the e2e test.

tests/e2e-no-crds/instrumentation-no-crds/00-install-operator.yaml
atoulme: Just so I understand - you're ok with the test, you just want the operator installed in the prepare step. I guess that doesn't change the substance of the test and can make the test more reliable down the road.

internal/instrumentation/upgrade/upgrade.go
atoulme: no this is not right, I'll revert changes to that file.

tests/e2e-no-crds/instrumentation-no-crds/00-install-operator.yaml
swiatekm: That's basically the idea, yeah. The way our e2e tests are structured is that you install the operator with whatever configuration you want in a preparation step, and then proceed with your test logic. Only exception are the operator upgrade tests.

atoulme: I'll give it a shot, learning how to use kustomize for it.

atoulme: done, refactored. Going to keep with the CI and make sure it passes, but you can take a look.

tests/e2e-no-crds/instrumentation-no-crds/00-install-collector.yaml
```diff
@@ -0,0 +1,78 @@
...
+          exporters: [debug]
+        traces:
+          receivers: [otlp]
+          exporters: [debug]
+        logs:
+          receivers: [otlp]
+          exporters: [debug]
+---
+apiVersion: apps/v1
+kind: Deployment
```
swiatekm, line 30: Is there a reason we're not using the OpenTelemetryCollector CR for this setup?

main.go
```diff
@@ -367,10 +367,12 @@ func main() {
 		}
 	}
 
-	err = addDependencies(ctx, mgr, cfg)
-	if err != nil {
-		setupLog.Error(err, "failed to add/run bootstrap dependencies to the controller manager")
-		os.Exit(1)
+	if cfg.EnableInstrumentationCRDs {
+		err = addDependencies(ctx, mgr, cfg)
```
swiatekm: We should really rename this function, we don't want more stuff unrelated to instrumentation to be added to it in the future. `addInstrumentationUpgrader` maybe?

config/no-crds/kustomization.yaml
```diff
@@ -0,0 +1,61 @@
+# Adds namespace to all resources.
```
swiatekm, line 1: I'm not going to block the PR over this, can be cleaned up afterwards, but I think this stuff should go into config/overlays and be similar to the openshift patches. You're duplicating a bunch of setup here for no benefit.

swiatekm (commented):
LGTM overall, had a few more questions about tests and the kustomize setup.

tests/e2e-no-crds/instrumentation-no-crds/00-install-collector.yaml
atoulme: we are not installing CRDs at all as part of the test. There is no fine grained way to install one CRD but not the other, so it was easier to deploy a collector.

main.go
atoulme: done

config/no-crds/kustomization.yaml
atoulme: Sure, please help me understand how to reuse the logic, I'm new to this framework, and I can clean things up.

pavolloffay: Are we going to produce a manifest file to install the operator without webhook? Or is this just for the e2e test?

pavolloffay (commented):
LGTM, I would resolve @swiatekm comment about aligning the config directory with overlays. 

On the other hand if we are not planning to release a manifest without the webhook, maybe there is a simpler approach without touching the `config` directory? (E.g. use the normal installation and remove the webooks in the test? )

config/no-crds/kustomization.yaml
atoulme: I'd like to build a helm chart to install the webhook without CRDs next.

swiatekm: I've gone through a bunch of alternative approaches, and this one is the most straightforward imo. I'm fine merging this PR as-is after you address the conflicts, and then thinking about how to organize the manifests afterwards.

atoulme:
Not stale - just no bandwidth. Will revisit and get it in shape for merging.

atoulme:
I note https://github.com/open-telemetry/opentelemetry-operator/pull/4981 is similar - no CRDs loaded, the software is deployed standalone. The difference is those tests are in pure go. Is that what we should do here?

swiatekm:
> I note #4981 is similar - no CRDs loaded, the software is deployed standalone. The difference is those tests are in pure go. Is that what we should do here?

If you want to, I'd be ok with that, but it looks like chainsaw works fine for your purposes? Those TA tests wanted to verify some semantic properties about target allocation that are very awkward to do in a bash script.

I'm fine merging this PR as is, for the record.

swiatekm (approved):

atoulme:
@swiatekm let's merge as is and refine over time please

iblancasa (approved):

Issue #4204, opened 2025-07-19
Run operator e2e tests nightly on the latest contrib collector image


swiatekm:
### Component(s)

collector

### Describe the issue you're reporting

We've recently faced two major collector bugs which prevented us from releasing. These bugs were immediately caught by our e2e tests. We should run these tests periodically on the latest contrib image. This way, we can catch issues early and report them upstream, ideally allowing them to be fixed before the collector releases.

In terms of implementation, we need to turn the current [e2e tests](https://github.com/open-telemetry/opentelemetry-operator/blob/main/.github/workflows/e2e.yaml) into a reusable workflow. Then we need to add a scheduled workflow running the tests with the `otel/opentelemetry-collector-contrib-dev` docker image. This workflow should create an issue assigned to all the operator approvers if the tests fail. Ideally, it should also retry execution once to account for flaky tests.

PR #4206, opened 2025-07-20
Run e2e tests nightly using the contrib dev image

swiatekm:
Run our E2E tests nightly using the collector contrib image built from `main`. When the test run fails, the workflow will create an issue to track resolving the failure. This way, we can detect breakage in both contrib itself and our E2E tests much earlier than during releases.

To achieve this, I've made the following changes:

* Moved the E2E tests into a reusable workflow
* Added a new workflow to be called nightly, which creates an issue if the test run fails
* Modified some of the E2E tests to be more relaxed about checking the collector image

- Resolves: https://github.com/open-telemetry/opentelemetry-operator/issues/4204

**Testing:**

E2E tests continue to be run normally on this PR. I've also initially set the contrib nightly tests to run in PR context, you can see an example execution [here](https://github.com/open-telemetry/opentelemetry-operator/actions/workflows/e2e-nightly.yaml). Finally, here's an [example](https://github.com/swiatekm/opentelemetry-operator/issues/103) of the issue that's created on failure in my fork.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -12,3 +12,21 @@ permissions:
 jobs:
   run-e2e-tests:
     uses: ./.github/workflows/e2e-reusable.yaml
+  
+  # This job is here to ensure the check has the right name we have set in Github's required checks.
+  e2e-tests-check:
+    runs-on: ubuntu-latest
+    if: always()
+    needs: [ run-e2e-tests ]
+    steps:
+      - name: Print result
+        run: echo ${{ needs.run-e2e-tests.result }}
+      - name: Interpret result
+        run: |
+          if [[ success == ${{ needs.run-e2e-tests.result }} ]]
+          then
+            echo "All matrix jobs passed!"
+          else
+            echo "One or more matrix jobs failed."
+            false
+          fi


.github/workflows/e2e-nightly.yaml
```diff
@@ -0,0 +1,99 @@
...
+    needs: [run-e2e-tests]
+    if: failure()
+    uses: ./.github/workflows/e2e-reusable.yaml
+    with:
+      collector-image: "otel/opentelemetry-collector-contrib-dev:latest"
+
+  create-issue-on-failure:
+    runs-on: ubuntu-latest
+    needs: [run-e2e-tests, retry-on-failure]
+    if: always() && (needs.run-e2e-tests.result == 'failure' && needs.retry-on-failure.result == 'failure')
```
jaronoff97, line 29: why do we need always here?

jaronoff97 (approved):
single question. Otherwise, LGTM.

.github/workflows/e2e-nightly.yaml
swiatekm: Because by default, if you put a job under `needs` and it fails, you don't run at all. Though maybe just `if: failure() && (needs.retry-on-failure.result == 'failure')` would work?

jaronoff97: ah i see, makes sense. i think this is fine.

Issue #3600, opened 2025-01-10
Support Prometheus Operator ScrapeClass.


nicolastakashi:
### Component(s)

target allocator

### Is your feature request related to a problem? Please describe.

In the Prometheus Operator, we have developed a cool feature called ScrapeClass, which allows us to define common configurations that are automatically injected into every scrape configuration.

This feature is particularly useful for standardizing configurations such as common relabeling rules, certificates, and other settings. It ensures consistency across monitored objects without requiring users to repeatedly define the same settings for each scrape target.

I’d love to hear your thoughts—do you think this could be a valuable addition to the PrometheusCR integration?

### Describe the solution you'd like

Extend TA Config to allow define scrape class structure while using PrometheusCR integration.

jaronoff97:
@swiatekm this sounds like exactly what we want for the TA rather than embedding the whole thing in the TA config

swiatekm:
Sounds good to me. One point worth mentioning is that unlike in prometheus-operator, in a target allocator + otel collector setup, service discovery and scraping happen in different applications. Right now we solve the issue around credentials by encrypting traffic between the target allocator and prometheus receiver, and simply exposing them via target allocator endpoints. Not sure if that makes any difference for scrape classes specifically.

nicolastakashi:
@swiatekm thanks for mention, I don't see any issue from our side with that, I think it's fine.

PR #4216, opened 2025-07-24
feat(target-allocator): Add  support for scrape-classes

ChristianCiach:

Adds support for ScrapeClasses as offered as a feature of the Prometheus Operator. Users can use these to add global configurations to multiple (or even all) PodMonitors and ServiceMonitors. 

I personally need this feature to get rid of the default labels (pod, container, namespace, ...) that the Prometheus-Operator automatically adds to all PodMonitors and ServiceMonitors as described in https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/running-exporters.md#podmonitors I consider these labels problematic and redundant, because they are already present as resource-attributes using proper Otel Semconv names. But I cannot safely drop these labels at the collector, because at this stage I cannot distinguish *actual* metric labels from the labels added by the Prometheus Operator. So I want to use a *default* ScrapeClass to globally drop these problematic labels before the metrics are scraped.

In https://github.com/open-telemetry/opentelemetry-operator/issues/3600#issuecomment-2592235122 @swiatekm raised a concern:

> One point worth mentioning is that unlike in prometheus-operator, in a target allocator + otel collector setup, service discovery and scraping happen in different applications. Right now we solve the issue around credentials by encrypting traffic between the target allocator and prometheus receiver, and simply exposing them via target allocator endpoints. Not sure if that makes any difference for scrape classes specifically.

Looking at the code, I believe this concern does not apply. The configuration of the scrape-classes are simply merged with the configurations found in PodMonitors and ServiceMonitors. The TargetAllocator simply retrieves the merged result from the Prometheus Operator, so the Target Allocator cannot even distinguish whether the configuration originates from a PodMonitor or a ScrapeClass.

**Link to tracking Issue(s):** #3600

- Resolves: #3600

I added a test-case that adds a simple scrape-class to the PrometheusCR configuration and let a PodMonitor reference it. The configuration of the scrape-class is correctly added to the resulting prometheus scrape-config.

TODO


nicolastakashi:
> @nicolastakashi I just see that you offered to work on this in [#3600 (comment)](https://github.com/open-telemetry/opentelemetry-operator/issues/3600#issuecomment-2592868683). This was a while ago though and the changes are pretty small, so I hope you are okay with me taking the plunge on this.

Thank you very much for working on that @ChristianCiach 🙏🏽

nicolastakashi (approved):
LGTM

ChristianCiach:
The `OpenTelemetryCollector` CRD doesn't yet offer to configure these ScrapeClasses via `.spec.targetAllocator.prometheusCR.scrapeClasses`. I use the TargetAllocator [standalone](https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-target-allocator), so I don't really care about the rest of the Operator its CRDs. If possible I would like this PR being merged without touching the rest of the operator. If you like I could try to extend the operator (and the CRDs) in a follow-up PR.

ChristianCiach:
Looks like I didn't properly generated the CRD yamls. Sorry, I know next to nothing about building Operators and CRDs. I will look into it.

apis/v1beta1/targetallocator_types.go
```diff
@@ -4,6 +4,7 @@
 package v1beta1
 
 import (
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
```
frzifus: @swiatekm While its added to the CR, it seems the Operator itself has already a dependency on the prometheus-operator:
https://github.com/open-telemetry/opentelemetry-operator/blob/ff9dd9360c602a86b6c222536360b71e0ad92695/go.mod#L28-L30

ChristianCiach:
I am back from vacation and I wonder if there have been any news regarding this PR. 

As far as I understand it, the only point of contention is whether to expose part of the Prometheus-Operator-API inside the `OpentelemetryCollector` CRD, making the CRD a lot larger.

The more I think about this, the more I think that this is the right thing to do. If the size of the CRD is the main concern, I could probably expose the `scrapeClasses` attribute as type `[]map[string]any` and then convert it back to `[]*monitoringv1.ScrapeClass` internally. The main downside of this would be the lack of validation in editors and on admission.

apis/v1beta1/targetallocator_types.go
swiatekm: Currently, the operator code has a dependency on it, yes. But this would make our CRDs depend on it, which means they would potentially change whenever we update these packages. I'm not sure whether that's something we want to maintain, and what the long-term consequences would be.  For example, if prometheus-operator does a breaking change on this field, does it mean we need to release a new version of our own CRDs if we want to update the dependency?

swiatekm:
@ChristianCiach apologies for the late response. We've had a lot of long vacations and other life events among the maintainers as well recently, so we're not as prompt in responding as we would've liked.

For reference, I ran a quick check on the size increase of the TargetAllocator CRD. We go from ~140KB to ~150KB, with a practical limit of around 250KB (the maximum size of an annotation value in K8s). I *think* that's acceptable, but we'll need to properly evaluate both this change and the dependency it creates for us.

@nicolastakashi do you know what kind of stability guarantees we can expect for the struct we're importing in this PR? Us using it this way would also introduce more friction if prometheus-operator ever wanted to make breaking changes to it, too.

simonpasquier:
:wave: prometheus-operator maintainer here!

Regarding our change policy, we stick to the Kubernetes API conventions as described in https://prometheus-operator.dev/docs/community/contributing/#changes-to-the-apis

> For stable API versions (e.g. v1), we don’t allow to break backward and forward compatibility.

Regarding the CRD size, going above the 250KB is fine as long as users know how to bypass the potential issue wrt annotations: https://prometheus-operator.dev/docs/platform/troubleshooting/#customresourcedefinition--is-invalid-metadataannotations-too-long-issue

apis/v1beta1/targetallocator_types.go
ChristianCiach: This should be answered by https://github.com/open-telemetry/opentelemetry-operator/pull/4216#issuecomment-3302783256

But if I change this CRD attribute to `runtime.RawExtension`, this issue should be avoided altogether.

PR #4218, opened 2025-07-24
fix(target-allocator): Stop pretending to be Prometheus 2.55.1

ChristianCiach:

Stop pretending to be Prometheus 2.55.1 when generating the Scrape-Configs using PrometheusCRs.

I mentioned in https://github.com/open-telemetry/opentelemetry-operator/pull/4000#issuecomment-3107764602 that this hardcoded version number results in some ugly warnings. I don't know if there are any worse consequences than just the warnings.

Anyway, @swiatekm agreed that this hardcoded version should be removed in https://github.com/open-telemetry/opentelemetry-operator/pull/4000#issuecomment-3107912757.

**Link to tracking Issue(s):** https://github.com/open-telemetry/opentelemetry-operator/pull/4000#issuecomment-3107912757.

- Resolves: https://github.com/open-telemetry/opentelemetry-operator/pull/4000#issuecomment-3107764602

No testing has been performed. I don't know why the version has been set to 2.55.1 to begin with, and I don't know what happens when we stop doing this. That's why this PR is a draft.

### DIFF
--- a/cmd/otel-allocator/internal/watcher/promOperator.go
+++ b/cmd/otel-allocator/internal/watcher/promOperator.go
@@ -82,7 +82,6 @@ func NewPrometheusCRWatcher(
 				ProbeSelector:                   cfg.PrometheusCR.ProbeSelector,
 				ProbeNamespaceSelector:          cfg.PrometheusCR.ProbeNamespaceSelector,
 				ServiceDiscoveryRole:            &serviceDiscoveryRole,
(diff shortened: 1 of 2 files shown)


ChristianCiach:
The faked prometheus version was originally added by https://github.com/open-telemetry/opentelemetry-operator/pull/3831, which happens to be authored by @swiatekm. So if he says that this should now be removed, he probably knows what he's talking about :smile: So I will undraft this PR,

Issue #4095, opened 2025-06-12
Support configuring "target reload interval" via CRD


bohub12:
### Component(s)

target allocator, collector

### Describe the issue you're reporting

Background
- https://github.com/open-telemetry/opentelemetry-operator/issues/4090#issuecomment-2959826356

Currently, the OpenTelemetry Operator does not expose a way to configure the ["target reload interval"](https://github.com/elastic/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/README.md#opentelemetry-operator) for the collector via the Custom Resource Definition (CRD). This value is useful for controlling how frequently the collector reloads its target list, especially in dynamic environments.

I’m planning to work on adding a new field to the CRD to allow configuring ["target reload interval"](https://github.com/elastic/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/README.md#opentelemetry-operator). This change will improve flexibility and simplify configuration for users.

I'm thinking of naming the new field CollectorTargetReloadInterval, but I'm open to suggestions if there's a more appropriate or consistent naming convention you'd recommend.

Please let me know if there are any concerns or guidelines I should follow before starting the implementation.

Thanks!

jaronoff97:
@bohub12 Can you share the configuration you have today so I can understand the problem first?

bohub12:
@jaronoff97, This is not about a problem with how the target reload interval currently works.

As mentioned in the “Background” section of this issue, I experienced a problem where some metrics were lost when collector pods scaled out or in. According to [the response I received](https://github.com/open-telemetry/opentelemetry-operator/issues/4090#issuecomment-2959826356), the target allocator waits until a collector pod becomes healthy, while the Prometheus receiver attempts to scrape targets immediately. As a result, the target list may temporarily be returned as empty.

To avoid this problem, I was advised to either adjust the target reload interval or modify the Prometheus receiver. Since the current CRD spec does not allow configuring the target reload interval, I created this issue to propose making it configurable. 

Are you suggesting that it might be sufficient to modify the target reload interval directly in the Prometheus receiver configuration, without adding a new field to the CRD spec?

jaronoff97:
[...]

PR #4222, opened 2025-07-27
Add `CollectorTargetReloadInterval` option

bohub12:

I've added a new `CollectorTargetReloadInterval` option that allows configuring how often the Prometheus receiver reloads targets from the target allocator.

- Resolves: #4095 

I've verified in a local Kubernetes cluster that:
- Embedded TargetAllocator
  - If not specified, `collectorTargetReloadInterval` is 30s (default)
  - If specified, `collectorTargetReloadInterval` uses the custom value
- TargetAllocator CRD
  - If not specified, `collectorTargetReloadInterval` is 30s (default)
  - If specified, `collectorTargetReloadInterval` uses the custom value


internal/manifests/collector/config_replace.go
```diff
@@ -58,6 +58,13 @@ func ReplaceConfig(otelcol v1beta1.OpenTelemetryCollector, targetAllocator *v1al
 		return "", validateCfgPromErr
 	}
 
+	// Set the interval - use CollectorTargetReloadInterval if specified, otherwise default to 30s
+	interval := "30s"
+	if targetAllocator.Spec.CollectorTargetReloadInterval != nil {
+		interval = targetAllocator.Spec.CollectorTargetReloadInterval.Duration.String()
+	}
```
swiatekm: I'd rather not define this default in two places. Since it's already defined in the CRD, you can either assume it's non-nil here, or check and return an error.

apis/v1alpha1/targetallocator_types.go
```diff
@@ -93,4 +93,11 @@ type TargetAllocatorSpec struct {
 	// +kubebuilder:default:="30s"
 	// +kubebuilder:validation:Format:=duration
 	CollectorNotReadyGracePeriod *metav1.Duration `json:"collectorNotReadyGracePeriod,omitempty"`
+	// CollectorTargetReloadInterval defines the interval at which the Prometheus receiver will reload targets from the target allocator.
+	// The default is 30s.
+	//
+	// +optional
+	// +kubebuilder:default:="30s"
+	// +kubebuilder:validation:Format:=duration
+	CollectorTargetReloadInterval *metav1.Duration `json:"collectorTargetReloadInterval,omitempty"`
```
swiatekm: Do we even need this here? This setting only affects collector behavior, so it should be enough to have it in the collector CRD.

internal/manifests/targetallocator/configmap.go
```diff
@@ -127,6 +127,10 @@ func ConfigMap(params Params) (*corev1.ConfigMap, error) {
 		taConfig["collector_not_ready_grace_period"] = taSpec.CollectorNotReadyGracePeriod.Duration
 	}
 
+	if taSpec.CollectorTargetReloadInterval.Size() > 0 {
+		taConfig["collector_target_reload_interval"] = taSpec.CollectorTargetReloadInterval.Duration
+	}
```
swiatekm: There is no target allocator setting like this, so this block has no effect.

swiatekm (commented):

I think you might be overcomplicating this change a bit. You're adding a setting that changes collector behavior, but which happens to be part of a config block the operator generates automatically. It's enough for this setting to be in the Collector CRD, there's no reason to change the TargetAllocator CRD. Does that make sense?

internal/manifests/collector/config_replace.go
bohub12: I also agree that having default values managed in multiple places is not ideal! So, I fix the logic assuming that the default values come from the CRD.

bohub12:
Thanks for the quick review! 👍 

I agree with your suggestion to provide `collectorTargetReloadInterval` only in the Collector CRD. So I added the option only to the Collector CRD as you mentioned. If there's anything in the code that needs to be fixed, feel free to let me know.

swiatekm:
@bohub12 your change looks good to me. Could you rebase on `main` and run `make update` to get rid of the conflicts in generated files?

jaronoff97 (approved):

swiatekm (approved):

PR #4230, opened 2025-07-30
Add network policy for operator

pavolloffay:

This PR adds k8s https://kubernetes.io/docs/concepts/services-networking/network-policies/ for the operator pod. 

The policy allows traffic:
* to operator's /metrics port
* to operator's webhook
* from operator to k8s API server

The network policy is enabled by default, it can be disabled with a flag.

The follow up PR will add network policy for operands created by the operator. For instance:
* to collector receiver pods
* to collector metrics endpoint
* for all egress traffic from the collector

- Updates https://github.com/open-telemetry/opentelemetry-operator/issues/4231 


tests/e2e/operator-restart/assert-operator-pod.yaml
```diff
@@ -14,3 +14,32 @@ status:
     ready: true
     started: true
   phase: Running
+---
```
frzifus: Should we move this into `assert-operator-networkpolicy.yaml`?

internal/controllers/opentelemetrycollector_controller.go
```diff
@@ -219,6 +219,8 @@ func NewReconciler(p Params) *OpenTelemetryCollectorReconciler {
 // +kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get;list;create;update
 // +kubebuilder:rbac:groups=monitoring.coreos.com,resources=servicemonitors;podmonitors,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=networking.k8s.io,resources=ingresses,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=networking.k8s.io,resources=networkpolicies,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=apps,resources=deployments/finalizers,verbs=get;watch;update;patch
```
frzifus, line 223: Why did we add `deployments/finalizers`?

internal/manifests/networkpolicies/operatornetworkpolicies.go
```diff
@@ -0,0 +1,108 @@
...
+
+	np := &networkingv1.NetworkPolicy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "opentelemetry-operator",
+			Namespace: n.operatorNamespace,
+		},
+		Spec: networkingv1.NetworkPolicySpec{
+			PodSelector: metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"app.kubernetes.io/name": "opentelemetry-operator",
```
frzifus: Is this always the name used or can it be configured via olm e.g. on  Openshift or on a helm chart?

pavolloffay: yes it is

internal/controllers/opentelemetrycollector_controller.go
pavolloffay: This is needed because the NP has ownerreference to the operator deployment for proper cleanup.

Otherwise I was getting 

```
error creating NetworkPolicy for the operator: networkpolicies.networking.k8s.io "opentelemetry-operator" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>
```

tests/e2e/operator-restart/assert-operator-pod.yaml
pavolloffay: +1

swiatekm:
At first glance, this feels like it should be a part of packaging rather than operator logic. Why not add it to the kustomize manifests and the Helm Chart instead?

internal/config/config.go
```diff
@@ -81,6 +81,8 @@ type Config struct {
 	AutoInstrumentationJavaImage string `yaml:"auto-instrumentation-java-image"`
 	// OpenshiftCreateDashboard creates an OpenShift dashboard for monitoring the OpenTelemetryCollector instances
 	OpenshiftCreateDashboard bool `yaml:"openshift-create-dashboard"`
+	// CreateNetworkPolicies creates network policies for the operator and operands created by the operator.
```
frzifus:
```suggestion
	// CreateNetworkPolicies creates network policies for the operator created by the operator.
```
When adding the support for the Operands, should you be able to disable it on the CR itself?

internal/manifests/networkpolicies/operatornetworkpolicies.go
```diff
@@ -0,0 +1,108 @@
...
+	clientset         kubernetes.Interface
+	operatorNamespace string
+	scheme            *runtime.Scheme
+}
+
+var _ manager.Runnable = (*networkPolicies)(nil)
+var _ manager.LeaderElectionRunnable = (*networkPolicies)(nil)
+
+func NewOperatorNetworkPolicies(operatorNamespace string, clientset kubernetes.Interface, scheme *runtime.Scheme) manager.Runnable {
+	return networkPolicies{
```
frzifus: Is it ok to make a copy here?
```suggestion
	return &networkPolicies{
```

internal/manifests/networkpolicies/operatornetworkpolicies.go
```diff
@@ -0,0 +1,108 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package networkpolicies
```
frzifus: Should we move this into a different path?
e.g. `internal/manifests/operator/networkpolicies.go` or `internal/manifests/manifestutils/networkpolicies.go`?
I expect, `collector`, `opampbridge`, `targetallocator` will all have their own policies?

internal/manifests/networkpolicies/operatornetworkpolicies.go
```diff
@@ -0,0 +1,108 @@
...
+				{
+					Ports: []networkingv1.NetworkPolicyPort{
+						{
+							Protocol: &tcp,
+							Port:     &apiServerPort,
+						},
+					},
+				},
+			},
+			PolicyTypes: []networkingv1.PolicyType{"Ingress", "Egress"},
```
frzifus:
```suggestion
			PolicyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress, networkingv1.PolicyTypeEgress},
```

internal/config/config.go
pavolloffay: yes, It will be done in a follow up PR

internal/manifests/networkpolicies/operatornetworkpolicies.go
pavolloffay: `internal/manifests/manifestutils` - it does not seem like a good fit there

The `internal` package contains a similar functionality in packages:
*  `internal/operator-metrics` 
* `internal/openshift/dashboards` 

I have moved the functionality temporarily to the `intenal/operatornetworkpolicies`. Once this is merged, let's  reorganize it better.

pavolloffay: > I expect, collector, opampbridge, targetallocator will all have their own policies?

that will fall into the appropriate `/internal/manifests/<component>` package

internal/operatornetworkpolicies/operatornetworkpolicies.go
```diff
@@ -0,0 +1,108 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package operatornetworkpolicies
```
frzifus: Personally I would rename this to just `internal/networking/operator(_policies).go` or `internal/manifests/manifestutils/networking/operator.go`. But up to the other @open-telemetry/operator-approvers

internal/operatornetworkpolicies/operatornetworkpolicies.go
```diff
@@ -0,0 +1,108 @@
...
+		clientset:         clientset,
+		operatorNamespace: operatorNamespace,
+		scheme:            scheme,
+	}
+}
+
+func (n *networkPolicies) Start(ctx context.Context) error {
+	tcp := corev1.ProtocolTCP
+	webhookPort := intstr.FromInt32(9443)
+	metricsPort := intstr.FromInt32(8443)
```
frzifus: Can this lead to issues when users configure different ports?
The helm chart default seems to look like this:
```
    Args:
      --metrics-addr=0.0.0.0:8080
      --enable-leader-election
      --health-probe-addr=:8081
      --webhook-port=9443
      --collector-image=otel/opentelemetry-collector-k8s:0.129.1
```

pavolloffay: I have added dynamic port setting from the supplied operator config

internal/operatornetworkpolicies/operatornetworkpolicies.go
pavolloffay: I would like us to avoid ambiguous naming, there are many files in the codebase named `main.go`, `operator.go` which to me seems more like an anti pattern.

main.go
```diff
@@ -194,6 +197,36 @@ func main() {
...
+				setupLog.Error(errParse, "failed to parse port from metrics address", "metricsAddr", cfg.MetricsAddr)
+				os.Exit(1)
+			}
+			policyOpts = append(policyOpts, operatornetworkpolicy.WithMetricsPort(int32(metricsPort)))
+		}
+		operatorNetworkPoliciesErr := mgr.Add(operatornetworkpolicy.NewOperatorNetworkPolicy(clientset, mgr.GetScheme(), policyOpts...))
+		if operatorNetworkPoliciesErr != nil {
+			setupLog.Error(operatorNetworkPoliciesErr, "failed to create the Operator network policies")
+		}
+	}
```
frzifus: Since this is all port parsing for the network policies, should we move it into `internal/operatornetworkpolicy`?

Ether into the construction or as option (`func WithSettingsFromConfig(cfg config.Config) (Option, error)`)?.

pavolloffay: > func WithSettingsFromConfig(cfg config.Config) (Option, error))

I would prefer to keep the `networkpolicy` package without a dependency on the `config` package. I also prefer to have the options pattern operate on options that are explicitly needed by the creator function.

config/overlays/openshift/manager-patch.yaml
```diff
@@ -11,3 +11,4 @@
     - '--enable-cr-metrics=true'
     - '--create-sm-operator-metrics=true'
     - '--feature-gates=operator.sidecarcontainers.native'
+    - '--feature-gates=operator.networkpolicy'
```
frzifus:
```suggestion
    - '--feature-gates=operator.sidecarcontainers.native,operator.networkpolicy'
```

I am not sure if that will disable native sidecars. I am currently adapting the e2e tests for native sidecars on k8s.  Until then we need to validate it manually. But using one argument should work.

pavolloffay: I have merged them

PR #4243, opened 2025-08-03
fix(labels): ensure app.kubernetes.io/version reflects image tag on workloads

### LINKED ISSUE #4175: Collector pod label `app.kubernetes.io/version` is wrong when using global `--collector-image` option
### Component(s)

collector

## Description

I noticed that the label `app.kubernetes.io/version` of the created collector workloads (DaemonSet, ...) always has the value `latest`, even though I configured the operator with this CLI option:

```
--collector-image=ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-contrib:0.129.1
```

It looks like the operator only uses the `.spec.image` attribut of the CR to generate this label and fallbacks to `latest` if not specified there, even when a global default image has been set.

This is an issue because the Otel Collector uses this label to determine the `service.version` resource attribute.

## Steps to Reproduce

* Deploy the operator with the option shown option.
* Do *not* set `.spec.image` in the OpenTelemetryCollector custom resource.

## Expected Result

The label should be `app.kubernetes.io/version: '0.129.1'`.

## Actual Result

The label is `app.kubernetes.io/version: 'latest'`.

### Kubernetes Version

1.33.1

### Operator version

v0.129.1

### Collector version

v0.129.1

### Environment information

_No response_

### Log output

```shell

```

### Additional context

_No response_

andrefrco:
**Description:** 

This PR makes sure the `app.kubernetes.io/version` label on workloads (Deployment, DaemonSet, and StatefulSet) matches the tag from the image passed via the `--collector-image`, `--target-allocator-image`, or `--operator-opamp-bridge-image` CLI flags in cases where `.spec.image` isn’t set directly in the OpenTelemetryCollector, OpAMPBridge, or TargetAllocator CR.

Before this, if `.spec.image` wasn’t defined, the operator would fall back to setting "latest" as the label even though the actual image (like 0.129.1) came from the CLI flag. This caused the `service.version` resource attribute to be incorrect in the workloads.

- Resolves: #4175

**Testing:**

Added 3 unit tests per workload type:

1. When .spec.image is defined: The label reflects the tag from .spec.image
2. When only the global operator config is defined: The label reflects the tag from the configured image in config.Config
3. When neither .spec.image nor the global config is set (config.Config{}): The label defaults to "latest"

Fixed a DefaultTests that incorrectly expected "latest" when using config.New(). Since config.New() returns a config with a [fallback value (e.g.)](https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/version/main.go#L89-L90) used for tests, the test was updated to reflect that version "0.0.0" instead of "latest".

I've also tested the change locally using minikube and confirmed that the app.kubernetes.io/version label is now correctly set on the generated workloads.

**Documentation:**

This is a bug fix, no doc changes required.

**Design consideration:**

I thought about changing the ``labels.Labels()`` function to accept a ``config.Config`` parameter, which would let us handle the version resolution in a single place. Something like:
```
func Labels(instance metav1.ObjectMeta, name string, image string, component string, config.Config) map[string]string
```
That way, we could reuse this logic across all components (services, serviceMonitors, etc) using the resourceImage and filterLabels together.

But I was a bit worried it might make the labels package more complex than it needs to be, since it would have to know about operator components and global config.

So for now, I went with a simpler approach and kept the logic outside. I'm new here and happy to improve this if there's a better way 🙂


andrefrco:
I'll check the e2e tests failure asap

internal/manifests/collector/daemonset.go
```diff
@@ -16,7 +16,12 @@ import (
 // DaemonSet builds the deployment for the given instance.
 func DaemonSet(params manifests.Params) (*appsv1.DaemonSet, error) {
 	name := naming.Collector(params.OtelCol.Name)
-	labels := manifestutils.Labels(params.OtelCol.ObjectMeta, name, params.OtelCol.Spec.Image, ComponentOpenTelemetryCollector, params.Config.LabelsFilter)
+
+	image := params.OtelCol.Spec.Image
+	if len(image) == 0 {
```
iblancasa:
```suggestion
	if image == "" {
```

.chloggen/fix-labels-version-reflects-image-tag.yaml
```diff
@@ -0,0 +1,18 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: labels
```
iblancasa: I think this should be 
```suggestion
component: [collector, target allocator, opamp]
```

But not sure if multiple components are supported...

andrefrco:
> I'll check the e2e tests failure asap

Fixed the e2e test by removing the hardcoded version assertion from kubeletstats test (to avoid version-specific checks). Also added the same image resolution logic to UpdateCollectorStatus for consistency since it creates labelSelectors.

Issue #2585, opened 2024-01-30
Updating the AutoScale/HPA causing the pods to redeploy


JustinMason:
### Component(s)

operator

## Description

When updating the AutoScale HPA settings the containers are redeployed.

## Steps to Reproduce

Deploy Otel Collector with AutoScale set to min: 3
Validate that 3 replicas are running.
Update AutoScale to Min: 4
Observer that the deployment/statefulSet gets redeployed cause the running replicas to need to all be restarted.

## Expected Result
HPA CRD would be updated without recycling the deployment.

## Actual Result
All Replicas are restarted

### Kubernetes Version

v1.27.7-gke.1121000

### Operator version

v0.92.1

### Collector version

v0.92.0

## Environment
GKE

### Additional context

Ideally an UPDATE to an existing HPA would only update the HPA so that running instances are not restarted.
When a system is under heavy load there may be a need to increase the Max Replicas.
Restarting all the containers causes a huge backlog as the replicas come back online.
If the goal was to mitigate load by increasing replicas this behavior makes it much worse before it can use the new HPA.

changexd:
Hi, Is it possible for you to provide configurations of otelcol with before and after modifications?
As I've tried only editing `autoscaler.minReplicas` in the otelcol resource with the steps you mentioned, it did not redeploy/restart pods

yuriolisa:
@JustinMason, could you please provide the manifests you applied in order to try to reproduce the error?

yuriolisa:
@JustinMason, did you have the opportunity to check it out?

JustinMason:
Sorry for the delay, I have stopped using the autoscale for the persistent collector because it was not helping our use case.

This was the section I had been using:
```yaml
autoscaler:
  minReplicas: 2
  maxReplicas: 2
  targetCPUUtilization: 60
  targetMemoryUtilization: 80
  behavior:
    scaleDown:
      policies:
        - periodSeconds: 120
          type: 'Pods'
          value: 1
      selectPolicy: 'Max'
      stabilizationWindowSeconds: 400
    scaleUp:
      policies:
        - periodSeconds: 15
          type: 'Pods'
          value: 2
      selectPolicy: 'Max'
      stabilizationWindowSeconds: 60
```

PR #4255, opened 2025-08-05
Fix deployments restarts issue

yuriolisa:

Fix a Deployment restart issue caused when the HPA settings changed, the webhook would modify spec.replicas.

- Resolves: #2585 

### DIFF
--- a/.chloggen/fix-hpa-cause-restarts.yaml
+++ b/.chloggen/fix-hpa-cause-restarts.yaml
@@ -2,7 +2,7 @@
 change_type: 'bug_fix'
 
 # The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
-component: 'operator'
+component: 'collector'
 
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
 note: Fix a Deployment restart issue caused when the HPA settings changed, the webhook would modify spec.replicas.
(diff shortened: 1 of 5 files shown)


.chloggen/fix-hpa-cause-restarts.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'bug_fix'
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: 'operator'
```
iblancasa:
```suggestion
component: 'collector'
```

iblancasa (approved):

.chloggen/fix-hpa-cause-restarts.yaml
yuriolisa: Just fixed.

atoulme (approved):

iblancasa (approved):

pavolloffay (approved):

cskinfill:
I think this update may be causing a problem with the HPA scaling.  My understanding of how the HPA works, is that it will calculate a new number of replicas and then update the `.spec.replicas` field of whatever the HPA's `.spec.scaleTargetRef` is.  If the `.spec.autoscaler.minReplicas` is set, I believe this update will cause each time the operator's reconcile loop runs, it will set the child `Deployment`'s `.spec.replicas` to the `OpenTelemetryCollector`'s `.spec.autoscaler.minReplicas` and ignore what the HPA determined the number of replicas should be.

Issue #4242, opened 2025-08-01
Support `spec.observability.metrics.disableMonitoringService`


kurarrr:
### Component(s)

collector

### Is your feature request related to a problem? Please describe.

Hi all, I have a cluster with edge nodes and am using OpenTelemetryCollector with daemonset mode.
It looks the MonitoringService is always created when the collector resource is created.

https://github.com/open-telemetry/opentelemetry-operator/blob/4fcb50f45c1b8e85a5f86fff25b96aa65eece904/internal/manifests/collector/collector.go#L44

In this situation, this ClusterIP service have many endpoints but I'm not using this service.

### Describe the solution you'd like

Can I implement `spec.observability.metrics.disableMonitoringService` for collector?

kurarrr:
@iblancasa Thank you! I've created a PR https://github.com/open-telemetry/opentelemetry-operator/pull/4256 .

Issue #4205, opened 2025-07-19
Option to disable creation of collector's K8s services


sky333999:
### Component(s)

collector

### Is your feature request related to a problem? Please describe.

Currently, the Operator creates several K8s services corresponding to a collector custom resource. In scenarios where the collector is deployed as a daemonset and does not play the role of a server and is used purely as a scraping client (say prometheus scraping as an example), the K8s services are not required.

Creating the K8s service by default, especially the ClusterIp one, adds a lot of overhead by wasting rev space in etcd, creating NAT tables for kube-proxy etc. This is particularly impactful on large clusters where a churn in the daemonset pods can lead more impact.

### Describe the solution you'd like

An option to disable the creation of K8s services, similar to https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.yaml#L516.

yuriolisa:
As you mentioned, for large environments, I believe it does make sense. Please feel free to submit a PR.

PR #4256, opened 2025-08-05
Add spec.observability.metrics.disableMonitoringService

kurarrr:
In v1beta1 collector, add `spec.observability.metrics.disableMonitoringService`, which disables monitoring service that is always created.

- Resolves: https://github.com/open-telemetry/opentelemetry-operator/issues/4242

I added a test case asserting `nil, nil` when `spec.observability.metrics.disableMonitoringService` is true

### DIFF
--- a/.chloggen/add_observability_metrics_disableMonitoringService.yaml
+++ b/.chloggen/add_observability_metrics_disableMonitoringService.yaml
@@ -1,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: collector
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Add spec.observability.metrics.disableMonitoringService"
+
+# One or more tracking issues related to the change
+issues: [4242]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Add `spec.observability.metrics.disableMonitoringService` to disable the monitoring service in the collector.


yuriolisa:
@kurarrr, do you believe this PR would be related to that [issue](#4205
) as well?

iblancasa (commented):
Can you add this as an step to one of our e2e tests?

sky333999:
> > @kurarrr, do you believe this PR would be related to #4205 as well?
> 
> Thank you for letting me know, this looks exactly the same issue as mine! Though it's also an option to always disable a monitoring service for daemonset mode.

My use case is actually a super set of this. 
For this issue, I think it makes sense to add an option to disable the monitoring service under `spec.observability.metrics`, but I dont think it can be extended to the ClusterIP service, headless service and extension service since those dont make sense under `spec.observability.metrics`.

So to keep things consistent, I was just going to expose configuring all the services at the root level of the collector CRD. 

Additionally, today its only the one enable/disable flag that we need, but using an explicit `disableMonitoringService` would mean it cant be easily extended to other configuration options related to the service in the future. Hence I tried to just make each of these separate objects and for now, provide the ability to override the name of the service in [my PR](https://github.com/open-telemetry/opentelemetry-operator/pull/4258).

Let me know what y'all think.

internal/manifests/collector/service.go
```diff
@@ -62,6 +62,10 @@ func HeadlessService(params manifests.Params) (*corev1.Service, error) {
 }
 
 func MonitoringService(params manifests.Params) (*corev1.Service, error) {
+	if params.OtelCol.Spec.Observability.Metrics.DisableMonitoringService {
```
sky333999, line 65: If this is disabled, I dont think there is a point to creating a `ServiceMonitor` either, so Id expect that to also get disabled automatically.

sky333999:
> @sky333999 In my case, services other than monitoring service are not set up because they are working like prometheus crawler and don't have receiving ports, so it was enough for me to close the monitoring service But in your case, does the pod have open ports and export to the upstream? But in anyway, your implementation looks good to me, so I can close this PR.

Right it depends on what the collector is actually configured to do. My collector does expose ports and plays the role of both a client and a server. So I get all 4 services by default.

Issue #4111, opened 2025-06-16
Non Default collector internal telemetry getting overwritten by the prometheus defaults


Kidsan:
### Component(s)

collector

## Description
We have the operator version 0.126.0 installed in kubernetes and have set it to deploy our own collector distrubution image, which works. We now want to enable the internal telemetry of the collectors and push to an otlp endpoint, however any configuration we add to the readers field of the collector spec gets overwritten with the default prometheus spec by the operator webhooks. Looking through the code, we found that if telemetry is set, the defaults shouldn't get applied, but this is not the behaviour we are seeing.
Interestingly, we have found if we manually specify the default config, but change the host or port, that change _does_ get applied.

## Steps to Reproduce
In an effort to simplify demoing the issue, I have created a test which fails: https://github.com/open-telemetry/opentelemetry-operator/compare/main...Kidsan:opentelemetry-operator:main

```yaml
# operator-values.yaml
manager:
  resources:
    limits:
      cpu: 100m
      memory: 256Mi
      ephemeral-storage: 50Mi
    requests:
      cpu: 100m
      memory: 256Mi
      ephemeral-storage: 50Mi

  autoInstrumentationImage:
    java:
      tag: 1.33.6
    nodejs:
      tag: 0.58.0
    python:
      tag: 0.53b1

  securityContext:
    allowPrivilegeEscalation: false
    capabilities:
      drop:
        - ALL
    runAsNonRoot: true
    readOnlyRootFilesystem: true
    seccompProfile:
      type: RuntimeDefault

  podAnnotations:
    "cluster-autoscaler.kubernetes.io/safe-to-evict": "true"

  extraEnvs:
    - name: GOMEMLIMIT
      value: 200MiB

kubeRBACProxy:
  enabled: true

  resources:
    limits:
      cpu: 500m
      memory: 128Mi
    requests:
      cpu: 5m
      memory: 64Mi

  securityContext:
    allowPrivilegeEscalation: false
    capabilities:
      drop:
        - ALL
    runAsNonRoot: true
    readOnlyRootFilesystem: true
    seccompProfile:
      type: RuntimeDefault
    runAsGroup: 1000
    runAsUser: 1000

admissionWebhooks:
  create: true

  certManager:
    enabled: true
    issuerRef: {}

pdb:
  create: false

securityContext:
  fsGroup: 1000
  runAsGroup: 1000
  runAsNonRoot: true
  runAsUser: 1000
```

```yaml
[...]

PR #4257, opened 2025-08-05
Add detailed message to Unmarshall issues

yuriolisa:

Adding a detailed message when the Operator fails the unmarshall process for `spec.service.telemetry` helps avoid silent failures caused by invalid custom telemetry configurations, which can lead to mysterious overwriting with defaults.

- Resolves: #4111


apis/v1beta1/config.go
```diff
@@ -476,17 +489,27 @@ func (s *Service) ApplyDefaults(logger logr.Logger) error {
...
 	host, port, err := s.MetricsEndpoint(logger)
 	if err != nil {
+		logger.Error(err, "failed to determine metrics endpoint for default configuration")
 		return err
 	}
 
 	reader := AddPrometheusMetricsEndpoint(host, port)
 	tel.Metrics.Readers = append(tel.Metrics.Readers, reader)
 
+	logger.Info("applied default Prometheus telemetry configuration",
```
iblancasa: I think this message should not be in the default verbosity level.

apis/v1beta1/config.go
```diff
@@ -573,19 +596,43 @@ func AddPrometheusMetricsEndpoint(host string, port int32) otelConfig.MetricRead
...
 	// Unmarshal JSON into the provided struct
 	if err := json.Unmarshal(jsonData, t); err != nil {
+		if logger != nil {
+			logger.Error(err, "failed to unmarshal telemetry configuration - this may indicate invalid configuration",
+				"json", string(jsonData), "originalConfig", s.Telemetry.Object)
+		}
 		return nil
 	}
+
+	if logger != nil {
```
iblancasa: You are doing this because of the tests, right? Maybe you can just use https://pkg.go.dev/go.uber.org/zap#NewNop instead of providing `nil` as parameter.

apis/v1beta1/config.go
yuriolisa: I don't agree 100% because most of the OTEL Config validations are in the default level, like when you forget to configure a receiver properly. So, I'd leave in the Info one.

apis/v1beta1/config.go
yuriolisa: Sure, let me do it.

iblancasa (approved):

pavolloffay (approved):

apis/v1beta1/config.go
jaronoff97: I think i agree with Israel here, we don't have a ton of default verbosity logs as far as i can tell ([link](https://github.com/search?q=repo:open-telemetry/opentelemetry-operator%20logger.Info\(&type=code), [link](https://github.com/search?q=repo:open-telemetry/opentelemetry-operator+logger.V(1)&type=code))

yuriolisa: I see, but specifically in that case, the default Prometheus telemetry config will be applied in case of an unmarshalling error, so as a user, I'd expect to see it straightforward in the Collector logs, instead of starting a debug to do it.

jaronoff97: We don't do an info log for applying the default endpoint configurations in here. The unmarshalling error is really just something we would receive if the user didn't set anything at all. I think giving this V(2) makes sense

apis/v1beta1/config.go
```diff
@@ -465,9 +465,22 @@ func (s *Service) MetricsEndpoint(logger logr.Logger) (string, int32, error) {
...
+		logger.V(2).Info("original telemetry configuration before parsing", "rawConfig", s.Telemetry.Object)
+
+		if metricsConfig, ok := s.Telemetry.Object["metrics"]; ok {
+			if metricsMap, ok := metricsConfig.(map[string]interface{}); ok {
+				if readers, hasReaders := metricsMap["readers"]; hasReaders && readers != nil {
+					logger.Info("found telemetry readers in original configuration", "readers", readers)
+				}
+			}
+		}
+	}
```
jaronoff97: Why are we doing the manual check here given we do the same thing below?

apis/v1beta1/config_test.go
```diff
@@ -208,7 +208,8 @@ func TestGetTelemetryFromYAML(t *testing.T) {
 			Address: "0.0.0.0:8888",
 		},
 	}
-	assert.Equal(t, telemetry, cfg.Service.GetTelemetry())
+	logger := logr.Discard()
+	assert.Equal(t, telemetry, cfg.Service.GetTelemetry(&logger))
```
jaronoff97: I think it would be useful to bring in the user's test case and Israel's solution [here](https://github.com/open-telemetry/opentelemetry-operator/issues/4111#issuecomment-2979990961) to verify we get the logs we want

apis/v1beta1/config.go
yuriolisa: As we agreed through slack, I will create an event instead.

apis/v1beta1/config.go
yuriolisa: Just removed in favor of the logs below.

apis/v1beta1/config_test.go
yuriolisa: Just implemented a test case covering the case mentioned above.

apis/v1beta1/config.go
```diff
@@ -464,10 +467,11 @@ func (s *Service) MetricsEndpoint(logger logr.Logger) (string, int32, error) {
 }
 
 // ApplyDefaults inserts configuration defaults if it has not been set.
-func (s *Service) ApplyDefaults(logger logr.Logger) error {
-	tel := s.GetTelemetry()
+func (s *Service) ApplyDefaults(logger logr.Logger, recorder record.EventRecorder, obj runtime.Object) error {
+	tel := s.GetTelemetry(&logger)
 
 	if tel == nil {
+		logger.V(1).Info("no telemetry configuration parsed, creating default")
```
jaronoff97: I think this should be v(2)

apis/v1beta1/config.go
```diff
@@ -476,17 +480,28 @@ func (s *Service) ApplyDefaults(logger logr.Logger) error {
 
 	if tel.Metrics.Address != "" || len(tel.Metrics.Readers) != 0 {
 		// The user already set the address or the readers, so we don't need to do anything
+		logger.V(1).Info("telemetry configuration already provided by user, skipping defaults",
+			"metricsAddress", tel.Metrics.Address,
+			"readersCount", len(tel.Metrics.Readers))
 		return nil
 	}
 
+	logger.Info("no telemetry readers configuration found, applying default Prometheus endpoint")
```
jaronoff97: IMO this is a debug log and / or can be removed

apis/v1beta1/config.go
```diff
@@ -262,7 +265,7 @@ func (c *Config) getEnvironmentVariablesForComponentKinds(logger logr.Logger, co
 
 // applyDefaultForComponentKinds applies defaults to the endpoints for the given ComponentKind(s).
 func (c *Config) applyDefaultForComponentKinds(logger logr.Logger, componentKinds ...ComponentKind) error {
-	if err := c.Service.ApplyDefaults(logger); err != nil {
+	if err := c.Service.ApplyDefaults(logger, nil, nil); err != nil {
```
jaronoff97: i believe we need to pass the recorder here

apis/v1beta1/config.go
```diff
@@ -476,17 +480,28 @@ func (s *Service) ApplyDefaults(logger logr.Logger) error {
...
 	}
 
 	reader := AddPrometheusMetricsEndpoint(host, port)
 	tel.Metrics.Readers = append(tel.Metrics.Readers, reader)
 
+	if recorder != nil && obj != nil {
+		recorder.Event(obj, corev1.EventTypeNormal, "Spec.Service.Telemetry.DefaultsApplied",
+			fmt.Sprintf("Applied default Prometheus telemetry configuration (host: %s, port: %d)", host, port))
+	}
+
```
jaronoff97: one option to maybe simplify the passing of the records would be to just return a (set of events, error) and propagate them back up to the object so we don't need to pass the object or recorder down at all. that also would make this more testable too

apis/v1beta1/collector_webhook.go
```diff
@@ -98,7 +100,7 @@ func (c CollectorWebhook) Default(_ context.Context, obj runtime.Object) error {
 		trueVal := true
 		otelcol.Spec.NetworkPolicy.Enabled = &trueVal
 	}
-	return otelcol.Spec.Config.ApplyDefaults(c.logger)
+	return otelcol.Spec.Config.ApplyDefaults(c.logger, c.recorder, otelcol)
```
jaronoff97: why are we passing otelcol down to itself?

jaronoff97: I also think another option here would be to have the ApplyDefaults method return a set of events which would allow us to test this more effectively. LMK what you think about that approach. this would allow us to not need to drill down the recorder and separate runtime.Obj reference.

jaronoff97 (commented):
I think this approach is fine, just want to see if attempted to return a list of events for recording rather than drilling the params all the way down

jaronoff97 (approved):
SO much cleaner, i like this approach a lot more. Thank you :bow: 

Do we want to add some tests in to the webhook_test file to verify we get the events we expect? (not a blocker)

PR #4258, opened 2025-08-06
Ability to configure Collector's Service, Headless Service, Monitoring Service and Extension Service

### LINKED ISSUE #4205: Option to disable creation of collector's K8s services

sky333999:
**Description:**
* Introduces new Collector CRD config options for configuring & managing the ClusterIP Service, Headless Service, Monitoring Service & Extension Service.
* For each of the Services, config to Enable/Disable the service and override the Name of the service.
* To maintain backwards compatibility, I considered the following two options and decided to go with the second.
  * Call the flag `Disabled` instead of `Enabled` such that it initializes to false when not set and the behavior stays the same.
  * Make the flags pointers instead of bools such that we can know whether the flag was explicitly provided i.e. `nil` implies retain old behavior.

- Resolves: https://github.com/open-telemetry/opentelemetry-operator/issues/4205, https://github.com/open-telemetry/opentelemetry-operator/issues/4242

**Testing:**
#### Unit Testing
* Updated existing unit tests and added new ones accordingly
#### Local Testing
* Setup: Created a new image with my local changes, pushed to ECR, deployed the Operator via the [helm chart](https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-operator) using my custom image.
* Scenario 1: Install a Collector CR (used this example) as-is.
  Validation: Daemonset is created, all 4 services are created. No observable change in behavior.
* Scenario 2: Update the Collector CR to disable ClusterIP service.
  Validation: Apply failed complaining the CRD did not recognize the new field. Expected behavior.
* Scenario 3: Updated CRD to include new fields, re-applied the Collector CR to disable the ClusterIP service.
   Validation: Apply was successful; The ClusterIP service is deleted.
* Scenario 4: Updated Collector CR to disable all 4 services.
  Validation: All 4 services are deleted.
* Scenario 5: Updated Collector CR to enable only the ClusterIP service and override the Name.
  Validation: Only the ClusterIP service is recreated and has the overridden name. 
#### E2E Testing
* Added E2E tests to capture default scenario, disabling scenario and name override scenario.

**Documentation:** CRDs are self-documenting, no further changes made.


Makefile
```diff
@@ -657,6 +657,7 @@ TMP_DIR=$$(mktemp -d) ;\
 cd $$TMP_DIR ;\
 go mod init tmp ;\
 echo "Downloading $(2)" ;\
+echo "replace fybrik.io/crdoc => github.com/fybrik/crdoc v0.5.2" >> go.mod ;\
```
sky333999: `fybrik.io` domain seems to have expired. So adding a replace temporarily.

Idk why this is a problem just for me. Is is that the go mod cached on the GH runners for the actions?

internal/controllers/builder_test.go
```diff
@@ -558,6 +558,48 @@ service:
 						Annotations: map[string]string{},
 					},
 				},
+				&networkingv1.Ingress{
```
sky333999, line 606: Just a re-ordering since now services are created after the ingress in `collector.go`.

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -239,6 +239,22 @@ type OpenTelemetryCollectorSpec struct {
 	// +optional
 	ServiceName string `json:"serviceName,omitempty"`
 
+	// Service to override configuration of the generated Collector Service.
+	// +optional
+	Service ServiceSpec `json:"service,omitempty"`
```
jaronoff97: there's no need to edit the v1alpha1 resources, these are deprecated and we shouldnt be touching them anymore.

sky333999: Sounds good. I can revert the changes and remove the changes to the converter as well.

kurarrr:
@sky333999 @jaronoff97 @yuriolisa 
Hi, is there any updates for this PR?

Issue #4117, opened 2025-06-18
Operator still try to create clusterRole and clusterRoleBinding when serivceAccount is provided in OpenTelemetryCollector and has clusterRole attached


fiona-huang-tyro:
### Component(s)

opamp

## Description

The OTEL operator will generate a serviceAccount, clusterRole, and clusterRoleBinding automatically in the app logic. This is useful when the serviceAccount is not provided in the OpenTelemetryCollector custom resource. However, there is a bug in the operator: even when a serviceAccount is provided, the operator will still try to create a clusterRole and clusterRoleBinding for the serviceAccount, even though there is already a clusterRole and clusterRoleBinding attached to the provided serviceAccount.

This will cause unsync issue when the project is gitOps-ed with argo because

1. the clusterRole and clusterRoleBinding generated by the operator will inherited all the labels from the OpenTelemetryCollector which contains a  ArgoCD-tracked label
2. there is no `ownerReference` in the enerated RBAC resources. and also given the clusterRole and clusterRoleBinding are cluster-scoped resources which is not alloed/recommended to have a `ownerReference` attached.

## Steps to Reproduce

The OpenTelemetryCollector i have
```
apiVersion: opentelemetry.io/v1beta1
kind: OpenTelemetryCollector
metadata:
  name: collector-deployment
  namespace: monitoring
  labels:
    argocd.argoproj.io/instance: observability-otel-operator-configs
spec:
  mode: deployment
  podSecurityContext:
    runAsUser: 1000
    runAsGroup: 1000
    seccompProfile:
      type: RuntimeDefault
  observability:
    metrics:
      enableMetrics: true
  securityContext:
    allowPrivilegeEscalation: false
    capabilities:
      drop:
      - ALL
    readOnlyRootFilesystem: true
    runAsNonRoot: true
    seccompProfile:
      type: RuntimeDefault
  serviceAccount: opentelemetry-collector
  config:
     .... some otel configs ...
```

existing clusertRoleBinding 

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: opentelemetry-collector
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: opentelemetry-collector
subjects:
  - kind: ServiceAccount
    name: opentelemetry-collector
    namespace: monitoring
```

existing clusterRole

```
apiVersion: rbac.authorization.k8s.io/v1
[...]

PR #4263, opened 2025-08-07
Skip RBAC creation when SA defined

yuriolisa:

Skip RBAC creation if a customer ServiceAccount is provided.
- Resolves: #4117 

### DIFF
--- a/internal/manifests/collector/rbac_test.go
+++ b/internal/manifests/collector/rbac_test.go
@@ -126,19 +126,15 @@ func TestDesiredClusterRolBinding(t *testing.T) {
 }
 
 func TestSkipRBACWhenCustomServiceAccount(t *testing.T) {
-	// Test that ClusterRole and ClusterRoleBinding are not created when custom ServiceAccount is provided
 	params, err := newParams("", "testdata/rbac_resourcedetectionprocessor_k8s.yaml", nil)
 	assert.NoError(t, err)
 
-	// Set custom ServiceAccount
 	params.OtelCol.Spec.ServiceAccount = "my-custom-service-account"
 
-	// Should return nil for ClusterRole when custom SA is provided
 	cr, err := ClusterRole(params)
 	require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


.chloggen/fix-rbac-duplication.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'bug_fix'
```
iblancasa, line 2: I would say this is a breaking change because if somebody is relying on this to make it work... it will stop working...

internal/manifests/collector/rbac.go
```diff
@@ -17,6 +17,11 @@ import (
 )
 
 func ClusterRole(params manifests.Params) (*rbacv1.ClusterRole, error) {
+	// Skip RBAC creation if custom ServiceAccount is provided
```
iblancasa, line 20: Maybe it makes more sense to check if the service account already has permission for the needed resources... what do you think?

.chloggen/fix-rbac-duplication.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: 'collector'
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Skip RBAC creation if a customer ServiceAccount is provided.
```
pavolloffay, line 8: `if a custom`

.chloggen/fix-rbac-duplication.yaml
pavolloffay: +1

yuriolisa:
I believe I should close this PR because we are already a possible way to do it through a [feature-flag.](https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/config/cli.go#L25)

Issue #4020, opened 2025-05-22
Bridge doesn't restart on all config changes


jaronoff97:
### Component(s)

opamp

## Description
The bridge pod doesn't restart on every config change made and requires a manual rollout

## Steps to Reproduce
change something in the non identifying attributes

## Expected Result
the bridge pod restarts when its config is modified

## Actual Result
nothing changes

### Kubernetes Version

1.25

### Operator version

0.123.0

### Collector version

n/a

## Environment
OS: (e.g., "Ubuntu 20.04")
Compiler(if manually compiled): (e.g., "go 14.2")

### Log output

```shell

```

### Additional context

I think we just need to set a config hash on the deployment

PR #4264, opened 2025-08-07
Fixes placement of pod annotations for opamp bridge

jaronoff97:
**Description:**
We were applying the annotations for the bridge in the wrong place, we needed to set the hash in the pod annotations, not the deployment annotations.

- Resolves: #4020

**Testing:** unit

**Documentation:** auto generated


jaronoff97:
realized that i really should just be using the pod annotations in the opamp bridge crd spec. making the change now.

yuriolisa (approved):

frzifus (approved):

Issue #4214, opened 2025-07-23
Add support for setting the OpenTelemetryCollector hostPID field


olandr:
### Component(s)

collector

### Is your feature request related to a problem? Please describe.

I have a situation where I want to run an openTelemetry-collector that accesses the audit-daemon (i.e. host namespace). It is connected to a feature request here https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/40836.

### Describe the solution you'd like

The issue seems to be that I am not allowed to set `spec.hostPID (boolean)`. Because when instantiating an OpenTelemetryCollector as DaemonSet manually (with `spec.hostPID: true`) I am able to run it as expected.

Proposed solution:
- allow for a user to set the `spec.hostPID`
- modify OpenTelemetryCollector CRD to allow setting `spec.hostPID`

### Describe alternatives you've considered

I have not found another way that would allow me to run a privileged openTelemetry-collector as a DaemonSet.

PR #4280, opened 2025-08-13
feat(collector): adds spec.hostPID for all collector modes

olandr:
**Description:**
At the moment it is not possible to set the `spec.hostPID` field for the collector (regardless of type). This change adds a featuregate `operator.security.hostpid` that allows a user to configure the `spec.hostPID` field for the collector. 

Inspiration for the code follows from the functionality of the `hostNetwork boolean` field
**Link to tracking Issue(s):** 4214

- Resolves: #4214

- adds e2e tests for all types
- adds unit tests for configuring the hostPID depending on featureflag activation.

Autogenerated docs from the type defs.

---------
on-behalf-of: @SAP Simon Olander (simon.olander@sap.com)

### DIFF
--- a/tests/e2e/daemonset-features/01-install.yaml
+++ b/tests/e2e/daemonset-features/01-install.yaml
@@ -1,4 +1,4 @@
-apiVersion: opentelemetry.io/v1alpha1
+apiVersion: opentelemetry.io/v1beta1
 kind: OpenTelemetryCollector
 metadata:
   name: daemonset
@@ -7,7 +7,7 @@ spec:
   terminationGracePeriodSeconds: 600
   hostNetwork: true
   hostPID: true
-  config: |
+  config:
     receivers:
       jaeger:
         protocols:


bundle/community/manifests/opentelemetry.io_targetallocators.yaml
```diff
@@ -1275,6 +1275,8 @@ spec:
                 type: object
               hostNetwork:
                 type: boolean
+              hostPID:
+                type: boolean
```
olandr, line 1339: This was auto generated when I did make bundle, I did not expect to see changes for the targetallocator. Would this then warrant test-cases changes fort the targetallocator component as well?

apis/v1beta1/common.go
```diff
@@ -186,6 +186,9 @@ type OpenTelemetryCommonFields struct {
 	// HostNetwork indicates if the pod should run in the host networking namespace.
 	// +optional
 	HostNetwork bool `json:"hostNetwork,omitempty"`
+	// HostPID indicates if the pod should have access to the host process ID namespace.
+	// +optional
+	HostPID bool `json:"hostPID,omitempty"`
```
atoulme, line 200: because you add this in common.go, I guess this was also added to target-allocator. Please remove and regenerate.

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: docker.io/pavolloffay/opentelemetry-operator
-  newTag: dev-e6d96f00-1755261473
+  newName: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
+  newTag: 0.131.0-36-ga732ae06
```
olandr: Also the auto-gen overwriting stuff from #4287. Should this docker image be used or is the auto-gen correct? cc @pavolloffay?

bundle/community/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -540,7 +540,7 @@ spec:
                   valueFrom:
                     fieldRef:
                       fieldPath: metadata.namespace
-                image: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator:0.131.0
+                image: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator:0.131.0-36-ga732ae06
```
pavolloffay: this should be reverted

config/manager/kustomization.yaml
frzifus:
```suggestion
  newTag: 0.131
```

Oh I guess that should be correct.

frzifus: I am not sure if it should be there at all. But if, it should be the opentelemetry one.

bundle/community/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -540,7 +540,7 @@ spec:
                   valueFrom:
                     fieldRef:
                       fieldPath: metadata.namespace
-                image: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator:0.136.0
+                image: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator:0.136.0-9-g247d92bf
```
pavolloffay: it should be reverted

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -205,6 +205,9 @@ type OpenTelemetryCollectorSpec struct {
 	// HostNetwork indicates if the pod should run in the host networking namespace.
 	// +optional
 	HostNetwork bool `json:"hostNetwork,omitempty"`
+	// HostPID indicates if the pod should have access to the host process ID namespace.
+	// +optional
+	HostPID bool `json:"hostPID,omitempty"`
```
pavolloffay: We should make it clear in the comment that the fields is only enabled with the feature gate.

olandr: After the discussion a few weeks ago we said that a feature flag is probably not needed. Because there are no other settings that use one. And as this change merely allows users to set a particular field (not a change in the logic of the operator) a feature flag was thought to be a bit superfluous.

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
+  newTag: 0.136.0-9-g247d92bf
```
frzifus:
```suggestion
```

apis/v1alpha1/opentelemetrycollector_types.go
swiatekm: We don't need to add this to v1alpha1 at all, as it's deprecated and slated for removal.

tests/e2e/smoke-collector-hostpid/00-assert-daemonset.yaml
```diff
@@ -0,0 +1,40 @@
+apiVersion: apps/v1
```
swiatekm: All of these E2E tests are somewhat overkill just to verify we're passing a field down to the core resource. It'd be sufficient to add it to [this test](https://github.com/open-telemetry/opentelemetry-operator/tree/main/tests/e2e/daemonset-features). The separate checks for deployments and statefulsets are unnecessary in my opinion.

olandr: Alright, sounds reasonable to me.

Issue #4288, opened 2025-08-15
API conversion webhook failure can cause cluster-wide outage


bderrly:
## Description

In a rare scenario, it is possible for an invalid v1alpha1 opentelemetrycollector CR to cause the conversion webhook to fail. The failure results in all Kubernetes watches failing on the cluster.  If this happens it causes disruptions in various components of the cluster ranging from the control-plane to the opentelemetry-operator itself.

The [Kubernetes documentation](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#response) seems to be aware such a situation can occur (see the "Warning" callout). As such, greater care must be taken in the conversion webhook to ensure it **does not fail** even with invalid configuration.

## Steps to Reproduce

1. Create an `otelcol` resource like the following while a version of the operator & CRD version that will allow creating a v1alpha1 resource stored as such in etcd:
   ```yaml
   apiVersion: opentelemetry.io/v1alpha1
   kind: OpenTelemetryCollector
   metadata:
     name: crashtime
   spec:
     mode: deployment
     autoscaler:
       null
     hostNetwork: false
     resources:
       limits:
         cpu: 256m
         memory: 512Mi
     image: "ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator:0.131.0"
     resources:
       limits:
         cpu: 256m
         memory: 512Mi
     config: |
       null
   ```
1. Upgrade the opentelemetry-operator to a relatively new version like 0.129.1 in which the CRD is updated to have the v1beta1 be the stored version.
1. 💥 

## Expected Result
The conversion webhook should not fail in such a way that kubernetes watches/informers and `kubectl get` requests fail. The operator should also not eventually crash.

## Actual Result
The opentelemetry-operator should not fail conversion webhook even with an invalid configuration.

### Kubernetes Version

1.31.10

### Operator version

0.129.1

### Collector version

0.125.0

## Environment
OS: GKE
Compiler: go 1.24

### Log output

```shell
[...]

PR #4289, opened 2025-08-15
Return an empty, valid configuration struct in v1alpha1 conversion webhook

bderrly:
**Description:** Returns a valid, empty configuration struct in the v1alpha1 conversion webhook rather than returning a failure.

**Link to tracking Issue(s):** #4288

- Resolves: #4288 

**Testing:** Added a few tests with invalid YAML configuration strings to ensure that a valid configuration, and no error, is returned from the webhook.

**Documentation:** None

### DIFF
--- a/tests/e2e/conversion-webhook-invalid-config/00-assert-created.yaml
+++ b/tests/e2e/conversion-webhook-invalid-config/00-assert-created.yaml
@@ -1,0 +1,19 @@
+apiVersion: opentelemetry.io/v1alpha1
+kind: OpenTelemetryCollector
+metadata:
+  name: invalid-config-collector
+spec:
+  mode: deployment
+  config: |
+    receivers: {}
+    exporters: {}
+    service:
+      telemetry:
+        metrics:
+          readers:
+            - pull:
+                exporter:
+                  prometheus:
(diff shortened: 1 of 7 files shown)


bderrly:
Thanks for taking a look @frzifus. I added the changelog file as well as an e2e test.

tests/e2e/conversion-webhook-invalid-config/00-assert-created.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: opentelemetry.io/v1alpha1
```
pavolloffay, line 1: does  v1beta1 fail on the admision with this config?

frzifus: Yes, since `Config` changed from string to an object:
```
Error from server: error when creating "otel-test.yaml": admission webhook "mopentelemetrycollectorbeta.kb.io" denied the request: json: cannot unmarshal string into Go struct field OpenTelemetryCollectorSpec.spec.config of type v1beta1.Config
```

swiatekm (approved):
LGTM, thank you for contributing this fix!

PR #4275, opened 2025-08-11
Use native sidecar on k8s 1.29+

frzifus:
**Description:**

The operator will automatically use native sidecars whenever a Kubernetes version higher than 1.29 is discovered
<Describe what has changed.>

**Link to tracking Issue(s):** <Issue number if applicable>

- Resolves: #3356

**Testing:** <Describe what testing was performed and which tests were added.>

**Documentation:** <Describe the documentation added.>

---

cc @IshwarKanse

PR #4310, opened 2025-08-26
avoid sidecar usage on e2e-instrumentation tests

frzifus:

- Resolves: #issue-number

cc  @IshwarKanse

### DIFF
--- a/tests/e2e-instrumentation/instrumentation-apache-httpd/01-assert.yaml
+++ b/tests/e2e-instrumentation/instrumentation-apache-httpd/01-assert.yaml
@@ -95,6 +95,10 @@ metadata:
     app.kubernetes.io/name: deployment-collector
     app.kubernetes.io/part-of: opentelemetry
   name: deployment-collector
+status:
+  availableReplicas: 1
+  readyReplicas: 1
+  replicas: 1
 ---
 apiVersion: v1
 kind: Service
@@ -105,3 +109,15 @@ metadata:
     app.kubernetes.io/name: deployment-collector
     app.kubernetes.io/part-of: opentelemetry
   name: deployment-collector
+spec:
+  ports:
+  - appProtocol: grpc
+    name: otlp-grpc
+    port: 4317
+    protocol: TCP
+    targetPort: 4317
+  - appProtocol: http
+    name: otlp-http
+    port: 4318
+    protocol: TCP
+    targetPort: 4318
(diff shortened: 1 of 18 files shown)


iblancasa:
Why?

iblancasa:
> > Why?
> 
> Otherwise, the change to the native sidecar will cause chaos.
> 
> * [Graduate native sidecar feature gate from alpha to beta #4275](https://github.com/open-telemetry/opentelemetry-operator/pull/4275)

Please, can you explain a bit more? I don't see what is the problem. Does this mean the instrumentation doesn't work with the native sidecar?

frzifus:
> > > Why?
> > 
> > 
> > Otherwise, the change to the native sidecar will cause chaos.
> > 
> > * [Graduate native sidecar feature gate from alpha to beta #4275](https://github.com/open-telemetry/opentelemetry-operator/pull/4275)
> 
> Please, can you explain a bit more? I don't see what is the problem. Does this mean the instrumentation doesn't work with the native sidecar?

When we enable native sidecar support by default, the sidecar gets injected differently. On k8s 1.28 and earlier, the `otc-container` will show up in the `containers` section on the pod. While on k8s 1.29 and above its registered in the `initContainers` section.

There are a few options to overcome this problem.
1. We duplicate all the asserts (e.g. `01-assert.yaml` and `01-assert-native.yaml`)
2. We write some more enhanced evaluation script.
3. We simply change the tests to make use of `mode: deployment` instead of `mode: sidecar`.

I picked option 3, mainly because we test the sidecar and native sidecar functionality already independently.

tests/e2e-instrumentation/instrumentation-apache-httpd/01-assert.yaml
```diff
@@ -83,12 +79,29 @@ status:
...
-    ready: true
-    started: true
   initContainerStatuses:
   - name: otel-agent-source-container-clone
     ready: true
   - name: otel-agent-attach-apache
     ready: true
   phase: Running
+---
+apiVersion: apps/v1
```
IshwarKanse, line 89: For deployment and service we can assert the status and ports as well.
examples:
https://github.com/open-telemetry/opentelemetry-operator/blob/main/tests/e2e-openshift/monitoring/01-assert.yaml#L10
https://github.com/open-telemetry/opentelemetry-operator/blob/main/tests/e2e-openshift/monitoring/01-assert.yaml#L45

frzifus: Thanks @IshwarKanse, I extended the asserts: https://github.com/open-telemetry/opentelemetry-operator/pull/4310/commits/d1faf5b65fff08fd194fb80e0d8be5c7f3e2983b

jaronoff97 (approved):
seems fine to me, this is all with the assumption that we'll have other tests for the native sidecar, yes?

PR #4378, opened 2025-09-23
Set TA and OpAMP status version when upgrading

pavolloffay:

- Resolves: #issue-number

### DIFF
--- a/internal/status/targetallocator/handle.go
+++ b/internal/status/targetallocator/handle.go
@@ -32,7 +32,12 @@ func HandleReconcileStatus(ctx context.Context, log logr.Logger, params targetal
 		return ctrl.Result{}, err
 	}
 	changed := params.TargetAllocator.DeepCopy()
-	changed.Status.Version = version.TargetAllocator()
+
+	// If the user provided a custom image we don't set the version.
+	// There is no guarantee that the custom image version matches the intended version.
+	if changed.Spec.Image == "" {
+		changed.Status.Version = version.TargetAllocator()
+	}
 
 	statusPatch := client.MergeFrom(&params.TargetAllocator)
 	if err := params.Client.Status().Patch(ctx, changed, statusPatch); err != nil {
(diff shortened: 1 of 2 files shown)


internal/status/targetallocator/handle.go
```diff
@@ -26,16 +26,14 @@ const (
...
-	log.V(2).Info("updating opampbridge status")
+	log.V(2).Info("updating target allocator status")
 	if err != nil {
 		params.Recorder.Event(&params.TargetAllocator, eventTypeWarning, reasonError, err.Error())
 		return ctrl.Result{}, err
 	}
 	changed := params.TargetAllocator.DeepCopy()
+	changed.Status.Version = version.TargetAllocator()
 
-	if changed.Status.Version == "" {
```
pavolloffay, line 36: @swiatekm if the custom image is used the operator sets the version from the `versions.txt`

pavolloffay: but after the upgrade the version was not updated

pavolloffay: I think if a user uses a custom image they should make sure it is compatible with the version that the operator expects. The expectation is that an operator version works well with a specific component version.

swiatekm: Sure, but they could easily have their own build of it, with a different versioning scheme, that is nonetheless compatible. So I'd take the container image tag if the image isn't the default one. @jaronoff97 @frzifus WDYT?

pavolloffay: https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/compatibility.md#opentelemetry-operator-vs-opentelemetry-collector

> By default, the OpenTelemetry Operator ensures consistent versioning between itself and the managed OpenTelemetryCollector resources. That is, if the OpenTelemetry Operator is based on version 0.40.0, it will create resources with an underlying OpenTelemetry Collector at version 0.40.0.

> When a custom Spec.Image is used with an OpenTelemetryCollector resource, the OpenTelemetry Operator will not manage this versioning and upgrading. In this scenario, it is best practice that the OpenTelemetry Operator version should match the underlying core version. Given a OpenTelemetryCollector resource with a Spec.Image configured to a custom image based on underlying OpenTelemetry Collector at version 0.40.0, it is recommended that the OpenTelemetry Operator is kept at version 0.40.0.

PR #4380, opened 2025-09-24
Allow setting PriorityClassName on TargetAllocator

apelisse:
**Description:**

PriorityClass can't be updated for the target allocation deploymnet.


swiatekm (changes requested):
PriorityClassName can be set on the [TargetAllocator CR](https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/api/targetallocators.md#targetallocatorspec). We specifically created that CRD so we could avoid embedding every single Pod attribute in the OpenTelemetryCollector CRD twice.

PR #4399, opened 2025-09-30
Validate portt from the collector config in the webhook

pavolloffay:

- Resolves: #issue-number

### DIFF
--- a/apis/v1beta1/collector_webhook.go
+++ b/apis/v1beta1/collector_webhook.go
@@ -225,6 +225,17 @@ func (c CollectorWebhook) Validate(ctx context.Context, r *OpenTelemetryCollecto
 	if err := ValidatePorts(r.Spec.Ports); err != nil {
 		return warnings, err
 	}
+	ports, errPorts := r.Spec.Config.GetAllPorts(c.logger)
+	if errPorts != nil {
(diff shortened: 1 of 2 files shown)


apis/v1beta1/collector_webhook.go
```diff
@@ -225,6 +225,17 @@ func (c CollectorWebhook) Validate(ctx context.Context, r *OpenTelemetryCollecto
 	if err := ValidatePorts(r.Spec.Ports); err != nil {
 		return warnings, err
 	}
+	ports, errPorts := r.Spec.Config.GetAllPorts(c.logger)
+	if errPorts != nil {
+		return warnings, fmt.Errorf("the OpenTelemetry config is incorrect. The port numbers are invalid: %w", errPorts)
+	}
+	for _, p := range ports {
+		nameErrs := validation.IsValidPortName(p.Name)
```
swiatekm: This would only happen due to bugs in the operator, right? Maybe we should indicate it in the error message.

pavolloffay: No, the name is dynamically constructed from the component name e.g. `zipkin/foo` creates a port name  `zipkin-foo`.

swiatekm (approved):

Issue #4400, opened 2025-09-30
Autoscaler doesn't scale deployment when MinReplicas is set


andrefrco:
### Component(s)

collector

## Description

When using HPA with OpenTelemetryCollector, the deployment doesn't scale when HPA decides it needs more replicas. The issue is that GetInitialReplicas() always returns MinReplicas when autoscaler is configured, ignoring the Spec.Replicas field that HPA modifies.

https://github.com/open-telemetry/opentelemetry-operator/blob/258418827fcd7ab5c3918e96f64e635fbc81fae6/internal/manifests/manifestutils/replicas.go#L10-L15

## Steps to Reproduce

1. Create an OpenTelemetryCollector with autoscaler configuration:
   ```yaml
   apiVersion: opentelemetry.io/v1beta1
   kind: OpenTelemetryCollector
   metadata:
     name: backend-traces
     namespace: monitoring
   spec:
     autoscaler:
       minReplicas: 3
       maxReplicas: 6
       targetCPUUtilization: 80
       targetMemoryUtilization: 50
     # ... rest of collector config
   ```

2. Check HPA status after load:
   ```bash
    ❯ k get hpa backend-traces-collector
    NAME                       REFERENCE                               TARGETS            MINPODS   MAXPODS   REPLICAS   AGE
    backend-traces-collector   OpenTelemetryCollector/backend-traces   62%/50%, 42%/80%   3         6        5           386d
   ```

3. Check deployment replicas:
   ```bash
   ❯ k get deployment backend-traces-collector -o jsonpath='{.spec.replicas}'
   3⏎
   ```

7. Check OpenTelemetryCollector spec.replicas:
   ```bash
   ❯ k get otelcol backend-traces -o jsonpath='{.spec.replicas}'
   5⏎
   ```

8. Verify the issue:
   ```bash
   ❯ k get pods -l app.kubernetes.io/name=backend-traces-collector
   NAME                                        READY   STATUS    RESTARTS   AGE
   backend-traces-collector-59f8cbdbcb-grstd   1/1     Running   0          19h
   backend-traces-collector-59f8cbdbcb-k6268   1/1     Running   0          3h14m
   backend-traces-collector-59f8cbdbcb-rqvhg   1/1     Running   0          19h
   ```

## Expected Result

When HPA decides it needs more replicas (e.g. 5), the deployment should scale to 5 pods, respecting HPA's decision.

## Actual Result

[...]

PR #4401, opened 2025-10-01
fix(collector): respect HPA scale; don't cap replicas at minReplicas

shuheiktgw:
The issue was already well described in the original report. However, the operator currently always returns the values defined in spec, and autoscaler.minValues never scales up. I’ve fixed this problem. I also renamed the function to GetDesiredReplicas, since it’s not only called during initialization but also on every reconciliation.

- Resolves: https://github.com/open-telemetry/opentelemetry-operator/issues/4400

### DIFF
--- a/internal/manifests/collector/deployment.go
+++ b/internal/manifests/collector/deployment.go
@@ -35,7 +35,7 @@ func Deployment(params manifests.Params) (*appsv1.Deployment, error) {
 			Annotations: annotations,
 		},
 		Spec: appsv1.DeploymentSpec{
-			Replicas: manifestutils.GetInitialReplicas(params.OtelCol),
+			Replicas: manifestutils.GetDesiredReplicas(params.OtelCol),
 			Selector: &metav1.LabelSelector{
 				MatchLabels: manifestutils.SelectorLabels(params.OtelCol.ObjectMeta, ComponentOpenTelemetryCollector),
 			},
--- a/.chloggen/bugfix-autoscaler-replicas-respect-spec.yaml
+++ b/.chloggen/bugfix-autoscaler-replicas-respect-spec.yaml
@@ -1,0 +1,6 @@
+change_type: bug_fix
+component: collector
+note: "Fix autoscaler not scaling above minReplicas; replicas now respect the scale subresource and never fall below autoscaler.minReplicas."
+issues: [4400]
+subtext: |
+  Also rename helper `GetInitialReplicas` to `GetDesiredReplicas` to reflect reconcile-time behavior.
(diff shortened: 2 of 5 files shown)


internal/manifests/collector/deployment.go
```diff
@@ -35,7 +35,7 @@ func Deployment(params manifests.Params) (*appsv1.Deployment, error) {
 			Annotations: annotations,
 		},
 		Spec: appsv1.DeploymentSpec{
-			Replicas: manifestutils.GetInitialReplicas(params.OtelCol),
+			Replicas: manifestutils.GetDesiredReplicas(params.OtelCol),
```
yuriolisa, line 38: I believe if `GetDesiredReplicas` will produce the same effect as described on #2585

yuriolisa: One idea would be adding a condition about the HPA [here](https://github.com/open-telemetry/opentelemetry-operator/blob/5f536470f25a16a8fab4131700de64590e3959a3/internal/manifests/mutate.go#L329)

shuheiktgw: Thank you so much for your comment! I tried reproducing the behavior described in the original issue, but from what I can see, even when we change minReplicas, the Operator doesn’t seem to create a new ReplicaSet, and the Pods aren’t restarted. Could you let me know if I’m missing something here?

```
kind: OpenTelemetryCollector
metadata:
  name: simplest
spec:
  autoscaler:
    minReplicas: 2
    maxReplicas: 4
  config:
    receivers:
      otlp:
        protocols:
          grpc:
            endpoint: 0.0.0.0:4317
          http:
            endpoint: 0.0.0.0:4318
    processors:
      memory_limiter:
        check_interval: 1s
        limit_percentage: 75
        spike_limit_percentage: 15
      batch:
        send_batch_size: 10000
        timeout: 10s

    exporters:
      debug: {}

    service:
      pipelines:
        traces:
          receivers: [otlp]
          processors: [memory_limiter, batch]
          exporters: [debug]

```

```
NAME                                 READY   STATUS    RESTARTS   AGE
simplest-collector-c9fb74dcd-njqd8   1/1     Running   0          28s
simplest-collector-c9fb74dcd-plvd8   1/1     Running   0          28s
```

Update the minReplicas from 2 to 3.

```
spec:
  autoscaler:
    minReplicas: 3
    maxReplicas: 4
```

Add one Pod without restarting the others.

```
NAME                                 READY   STATUS    RESTARTS   AGE
simplest-collector-c9fb74dcd-6rt44   1/1     Running   0          6s
simplest-collector-c9fb74dcd-njqd8   1/1     Running   0          52s
simplest-collector-c9fb74dcd-plvd8   1/1     Running   0          52s
```

pepperkick:
Can we merge this PR? We are having the same issue with our project and we believe this will fix it.

internal/manifests/collector/deployment.go
swiatekm: I also tested this manually and have noticed no problems. We should add a e2e tests for this scenario in a follow-up, but I'm not going to block this PR with that requirement. I'm going to merge now. @jaronoff97 can you make sure to pick it up for 0.138.0?

PR #4420, opened 2025-10-11
Promote the operator.collector.targetallocatorcr feature flag to Stable

swiatekm:
**Description:**

Promote the `operator.collector.targetallocatorcr` feature flag to Stable. The plan is to remove it in 0.138.0.

- Relates to #4025

### DIFF
--- a/.chloggen/feat_targetallocator-cr-stable.yaml
+++ b/.chloggen/feat_targetallocator-cr-stable.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: target allocator
+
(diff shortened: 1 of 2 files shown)


.chloggen/feat_targetallocator-cr-stable.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: target allocator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Promote the operator.collector.targetallocatorcr feature flag to Stable
```
pavolloffay, line 8: We should be more descriptive here. A regular user does not know what moving the FF to stable actually might mean.

swiatekm: I added a line of subtext. For users who aren't setting this flag, this doesn't change anything. The current behaviour has been the default for more than 5 releases.

Issue #3382, opened 2024-10-23
Deprecate `VolumeSizeLimit` in the Instrumentation CRD


swiatekm:
### Component(s)

auto-instrumentation

### Describe the issue you're reporting

With https://github.com/open-telemetry/opentelemetry-operator/pull/3285, it's possible to customize the instrumentation volume more deeply in general, including changing the size. Changing the volume size limit can now be accomplished by using a custom volume, so the individual attribute isn't necessary anymore.

pavolloffay:
discussed at sig meeting. No objections.

apt-x4869:
Hi @pavolloffay ,

From what I understand, the usual procedure to deprecate fields in a CRD is to define a new version (e.g., v1alpha2) and set `served: false` on the previous version.

Please let me know if I’m mistaken. I’m new to this project, so I may not be aware of other constraints or considerations.

If it’s okay, could you please assign this issue to me? I would love to work on it and learn more about the project.

Thanks!

PR #4431, opened 2025-10-16
[chore]: Deprecated VolumeSizeLimit in the Instrumentation CRD

ShreyashSri:
**Description:**
Deprecated the `volumeSizeLimit` field for all language specific instrumentations:
- The `volumeSizeLimit` field in `instrumentation_types.go` is now marked as deprecated.
- Validating webhook has been implemented to generate warnings whenever a user attempts to apply an `Instrumentation` resource that still uses the deprecated `volumeSizeLimit` field.
- API documentation has been updated to show the field's deprecated status.

- Resolves: #3382

**Testing:**
  ```
  go test ./apis/v1alpha1 -run TestInstrumentation -count=1
  ```
- Result: all tests passed:
  - `ok github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1 0.019s`
  
**Documentation:**
- Updated `docs/api/instrumentations.md` to mark the `volumeSizeLimit` field as deprecated and specify `spec.<lang>.volume.size` as its replacement.

### DIFF
--- a/bundle/community/manifests/opentelemetry-operator.clusterserviceversion.yaml
+++ b/bundle/community/manifests/opentelemetry-operator.clusterserviceversion.yaml
@@ -99,7 +99,7 @@ metadata:
     categories: Logging & Tracing,Monitoring
     certified: "false"
     containerImage: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
-    createdAt: "2025-10-17T09:24:13Z"
+    createdAt: "2025-10-23T18:06:48Z"
     description: Provides the OpenTelemetry components, including the Collector
     operators.operatorframework.io/builder: operator-sdk-v1.29.0
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
--- a/bundle/community/manifests/opentelemetry.io_instrumentations.yaml
+++ b/bundle/community/manifests/opentelemetry.io_instrumentations.yaml
@@ -329,7 +329,7 @@ spec:
                     required:
                     - spec
                     type: object
-                  volumeSizeLimit:
+                  volumeLimitSize:
                     anyOf:
                     - type: integer
                     - type: string
@@ -556,7 +556,7 @@ spec:
                     required:
                     - spec
                     type: object
-                  volumeSizeLimit:
+                  volumeLimitSize:
                     anyOf:
                     - type: integer
                     - type: string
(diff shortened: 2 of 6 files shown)


apis/v1alpha1/instrumentation_types.go
```diff
@@ -163,7 +163,8 @@ type Java struct {
 
 	// VolumeSizeLimit defines size limit for volume used for auto-instrumentation.
 	// The default size is 200Mi.
-	VolumeSizeLimit *resource.Quantity `json:"volumeLimitSize,omitempty"`
+	// Deprecated: use spec.<lang>.volume.size instead. This field will be removed in a future release.
```
pavolloffay: > This field will be removed in a future release.

We don't remove fields, but we can make it noop. 

> This field will be inactive in a future release.

ShreyashSri:
@pavolloffay made the changes

.chloggen/fix_deprecate_VolumeSizeLimit.yaml
```diff
@@ -0,0 +1,18 @@
...
+note: Deprecate VolumeSizeLimit in the Instrumentation CRD
+
+# One or more tracking issues related to the change
+issues: [3382]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  - The volumeSizeLimit field in instrumentation_types.go is now marked as deprecated.
```
pavolloffay: It would be a good practice to mention what is the alternative.

I would avoid mentioning the file name, it's not important for a user.

ShreyashSri: Ok, I'll make the changes

apis/v1alpha1/instrumentation_types.go
```diff
@@ -163,7 +163,8 @@ type Java struct {
 
 	// VolumeSizeLimit defines size limit for volume used for auto-instrumentation.
 	// The default size is 200Mi.
-	VolumeSizeLimit *resource.Quantity `json:"volumeLimitSize,omitempty"`
+	// Deprecated: use spec.<lang>.volume.size instead. This field will be inactive in a future release.
+	VolumeSizeLimit *resource.Quantity `json:"volumeSizeLimit,omitempty"`
```
swiatekm: Changing the name of the field in the CRD would be a breaking change. We should keep the CRD field the same and update the name in the struct and documentation instead.

PR #4447, opened 2025-10-21
Set V(4) logging when pdb is created by default

mveitas:
Set the logging level to 4 when a PDB is not set in the spec and created by default to reduce log volume

**Testing:**

**Documentation:**


.chloggen/pdb-log-level.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
pavolloffay: This is rather a bug_fix than an enhancement.

iblancasa (approved):

.chloggen/pdb-log-level.yaml
mveitas: Updated the change_type

frzifus (approved):

PR #4461, opened 2025-10-29
Migrate simple validation from webhook to annotations

shuheiktgw:
While reading the source code, I noticed that some of the validations implemented in the webhook could be simplified using annotations. Using annotations has a few advantages:

1. We can enforce validation without relying on the webhook.
2. We can apply validation wherever the struct is used. In the current case, Spec.StatusProbe was not validated because ValidateProbe was never called on it.

I migrated some of the simple validations to annotations and added tests. Thanks for the review!

- Resolves: #issue-number

### DIFF
--- a/.chloggen/migrate_validation_webhook_to_annotations.yaml
+++ b/.chloggen/migrate_validation_webhook_to_annotations.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: collector
+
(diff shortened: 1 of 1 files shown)


swiatekm (commented):
Thank you for taking this up! The changes look good to me. I think that we should add a changelog entry for this PR, given that we're modifying the CRDs. Could you add one?

shuheiktgw:
@swiatekm Thank you for the quick review! I've added a changelog 🙂

apis/v1beta1/collector_webhook_test.go
```diff
@@ -1189,138 +1185,6 @@ func TestOTELColValidatingWebhook(t *testing.T) {
 			},
 			expectedErr: "does not support the attribute 'affinity'",
 		},
```
iblancasa, line 1187: How are we testing now that the validation works?

swiatekm: There's tests for it below using envtest.

PR #4469, opened 2025-11-02
Remove duplicated Sidecar mode validations

shuheiktgw:
This PR is part of the migration from webhook-based validation to CEL. As a first step, I removed the duplicate validations from the webhook. Thank you for reviewing!

- https://github.com/open-telemetry/opentelemetry-operator/issues/3319


apis/v1beta1/collector_webhook.go
```diff
@@ -295,11 +272,6 @@ func (c CollectorWebhook) Validate(ctx context.Context, r *OpenTelemetryCollecto
 		)
 	}
 
-	if r.Spec.Ingress.Type == IngressTypeIngress && r.Spec.Mode == ModeSidecar {
-		return warnings, fmt.Errorf("the OpenTelemetry Spec Ingress configuiration is incorrect. Ingress can only be used in combination with the modes: %s, %s, %s",
-			ModeDeployment, ModeDaemonSet, ModeStatefulSet,
-		)
-	}
```
shuheiktgw, line 302: This was simply duplicated with the validation just above.

Makefile
```diff
@@ -87,16 +87,17 @@ ENABLE_WEBHOOKS ?= false
 # Additional flags for go test command
 GOTEST_EXTRA_OPTS ?=
 
-# If we are running in CI, run go test in verbose mode
+# If we are running in CI, run go test in verbose mode without caching
 ifeq (,$(CI))
 GOTEST_OPTS=-race $(if $(GOTEST_EXTRA_OPTS),$(GOTEST_EXTRA_OPTS))
 else
-GOTEST_OPTS=-race -v $(if $(GOTEST_EXTRA_OPTS),$(GOTEST_EXTRA_OPTS))
+GOTEST_OPTS=-race -v -count=1 $(if $(GOTEST_EXTRA_OPTS),$(GOTEST_EXTRA_OPTS))
```
shuheiktgw, line 94: The `-count=1` flag is required to ensure that `test-min-k8s` doesn’t use cached results from `test`.

apis/v1beta1/collector_webhook_test.go
```diff
@@ -1145,62 +1133,6 @@ func TestOTELColValidatingWebhook(t *testing.T) {
 			},
 			expectedErr: fmt.Sprintf("Ingress can only be used in combination with the modes: %s, %s, %s", v1beta1.ModeDeployment, v1beta1.ModeDaemonSet, v1beta1.ModeStatefulSet),
 		},
-		{
-			name: "invalid mode with priorityClassName",
```
pavolloffay, line 1149: Are we removing these tests because we are not longer able to test the CEL expressions?

Makefile
swiatekm: Won't this cause us to never cache test results when using any of the unit test targets in the Makefile? That's too extreme. According to the [documentation](https://pkg.go.dev/cmd/go#hdr-Test_packages), just reading the env variable in the course of the test is enough to do this:

```
Tests that open files within the package's module or that consult environment variables only match future runs in which the files and environment variables are unchanged. A cached test result is treated as executing in no time at all, so a successful package test result will be cached and reused regardless of -timeout setting. 
```

So a much simpler and more elegant solution would be to read `KUBEBUILDER_ASSETS` in TestMain and write out a log message about its contents.

shuheiktgw: > Won't this cause us to never cache test results when using any of the unit test targets in the Makefile? That's too extreme.

This is only for the CI environment, so I'm not too sure if it’s a big issue or not. The suggested approach above has one drawback: we’d need to reference KUBEBUILDER_ASSETS in every package where we want to run tests against both Kubernetes versions. If you think that’s acceptable, I’m happy to look into the approach described above 🙂

apis/v1beta1/collector_webhook_test.go
shuheiktgw: They’re now being tested in TestValidationViaCRDAnnotations using envtest 🙂

Makefile
shuheiktgw: I did a bit of research and testing, and it seems we can’t use TestMain. But if we simply reference the environment variable inside one of the tests in the package, it seems to work fine!

- https://github.com/golang/go/issues/24589
- https://github.com/golang/go/issues/24122

swiatekm: My bad, I missed the condition in the Makefile. In an ideal world, it should also be easy to run the unit tests on both versions locally, so the environment variable approach seems better to me. But I'm fine not blocking your PR with that request, we can merge it as-is.

swiatekm (approved):

Issue #3820, opened 2025-03-19
Create a first managed deployment of the collector


atoulme:
### Describe the issue you're reporting

Build enough of the CR that we have a collector deployment similar to what is allowed with the helm chart.

PR #4475, opened 2025-11-03
initial commit for clusterobservability CRD

jinja2:

Draft for Design Discussion - This PR introduces an initial implementation of the ClusterObservability controller as defined in the [RFC](https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/rfcs/managed.md) in repo.

The new ClusterObservability CR is meant to allow users to deploy a complete observability stack with a single Custom Resource:

```yaml
kind: ClusterObservability
metadata:
  name: cluster-observability
  namespace: opentelemetry-operator-system
spec:
  exporter:
    endpoint: "https://otel-backend.example.com:4318"
    headers:
      "Authorization": "Bearer token"
```

This initial implemantation creates the following resources:
- Agent Collector (DaemonSet): Node-level metrics, logs, and OTLP receiver
- Cluster Collector (Deployment): Cluster-level k8s metrics and events  
- Instrumentation CR: Auto-instrumentation configuration

The details of the current proposed design can we found in [docs/cluster-observability.md](https://github.com/open-telemetry/opentelemetry-operator/pull/4475/files#diff-db741f1924b37891bdcb3050a600df11c849f537e7253996d4586a3daca9fe5e) in draft PR. There is still more work to do, for e.g. adding tests, making the controller more extendable for introducing more distro specific implementation, more telemetry configs, etc. But I'd like to get some feedback on the fundamentals like design, CR config options, code structure.

- Resolves: #3820

Tests are missing for now, will be added once we discuss the design/implementation details.

I have added the new controller's design doc [here](https://github.com/open-telemetry/opentelemetry-operator/pull/4475/files#diff-db741f1924b37891bdcb3050a600df11c849f537e7253996d4586a3daca9fe5e)


config/samples/clusterobservability_v1alpha1_clusterobservability.yaml
```diff
@@ -0,0 +1,55 @@
...
+kind: ClusterObservability
+metadata:
+  name: cluster-observability
+  namespace: opentelemetry-operator-system
+spec:
+  # Signals to collect - must include at least one
+  signals:
+    - traces
+    - metrics
+    - logs
```
frzifus: Would it make sense to provide here some specific usecases instead of signals or something like "prepared" configurations?

The doc outlines the `ClusterObservability` CR  also creates e.g. an Auto-Instr. CR. Maybe we want system logs but ignore k8s logs or the other way around.

What comes to my mind would be something like this:
```suggestion
  collections:
    - all
    # - metrics_system
    # - metrics_ingestion
    # - logs_system
    # - logs_k8s
    # - traces_ingestion
    # - instr_all
    # - instr_java
```

That  way you could deploy a collector to collect metrics and logs from a host. While not accepting reported logs and not creating e.g. auto-instr. CRD.

internal/status/clusterobservability/handle.go
```diff
@@ -0,0 +1,489 @@
...
+	return componentStatus{
+		ready: true,
+		message: fmt.Sprintf("Cluster collector Deployment ready: %d/%d replicas ready",
+			deployment.Status.ReadyReplicas, deployment.Status.Replicas),
+	}
+}
+
+// checkInstrumentationStatus checks the status of the single Instrumentation CR.
+func checkInstrumentationStatus(ctx context.Context, cli client.Client, co *v1alpha1.ClusterObservability) componentStatus {
+	instrumentationName := "default-instrumentation"
```
frzifus: Can this lead to reconcile loops when two CRs are created on the same namespace? Maybe we could prefix this using the ClusterObservability CR Name.

frzifus (commented):
Thanks @jinja2 for picking this up! Thats rly great stuff! Just left some comments.

config/samples/clusterobservability_v1alpha1_clusterobservability.yaml
atoulme: I have been pushing for less toggles, not more. Here is a bit of reasoning around it:
I don't think we'll hear on feedback and specific use cases (such as logs_system vs logs_k8s) until we put this in the hands of users
If we make those choices now, there's a good chance we get this wrong. It looks somewhat easy to list use cases right now, but I'd like them discussed in the RFC with descriptions, justifications, stability level, etc.
I would decouple signals and use cases. Signals are technical plumbing toggles. Use cases can work across multiple signals.
Finally, I am very lazy and I want us to support less combinations of possible problems.

I want us to ship something that is too short, too clumsy, and gets us user feedback.

apis/v1alpha1/clusterobservability_types.go
```diff
@@ -0,0 +1,295 @@
...
+
+const (
+	ObservabilitySignalLogs     ObservabilitySignal = "logs"
+	ObservabilitySignalTraces   ObservabilitySignal = "traces"
+	ObservabilitySignalMetrics  ObservabilitySignal = "metrics"
+	ObservabilitySignalProfiles ObservabilitySignal = "profiles"
+)
+
+// OTLPHTTPExporter defines OTLP HTTP exporter configuration.
+// This structure mirrors the official OpenTelemetry Collector otlphttpexporter configuration.
```
swiatekm, line 11: There's a danger associated with doing this, where a breaking change in the exporter will force a breaking change on us. In addition, we'd need to keep up with additions made by the exporter. At that point, we may as well import the upstream config struct directly. 

I wonder if it isn't safer to make this into an opaque type and just pass it down.

config/crd/kustomization.yaml
```diff
@@ -6,6 +6,7 @@ resources:
 - bases/opentelemetry.io_instrumentations.yaml
 - bases/opentelemetry.io_opampbridges.yaml
 - bases/opentelemetry.io_targetallocators.yaml
+- bases/opentelemetry.io_clusterobservabilities.yaml
```
swiatekm: Adding the CRD here will cause it to be shipped in our release manifest, which I'd prefer to avoid for now. For testing, we can include it manually.

docs/api/clusterobservabilities.md
```diff
@@ -0,0 +1,576 @@
+# API Reference
```
swiatekm: We might want to avoid creating public documentation for now as well.

internal/controllers/clusterobservability_controller.go
```diff
@@ -0,0 +1,650 @@
...
+
+// cleanupManagedResources deletes all resources managed by ClusterObservability.
+func (r *ClusterObservabilityReconciler) cleanupManagedResources(ctx context.Context, log logr.Logger, instance *v1alpha1.ClusterObservability) error {
+	// Clean up OpenTelemetryCollector CRs (both agent and cluster collectors)
+	if err := r.cleanupCollectors(ctx, log, instance); err != nil {
+		return fmt.Errorf("failed to cleanup collectors: %w", err)
+	}
+
+	// Clean up the single Instrumentation CR in operator namespace
+	if err := r.cleanupInstrumentations(ctx, log, instance); err != nil {
```
swiatekm: This is always in the same namespace as the ClusterObservability resource, isn't it? Can't we just let the normal Kubernetes garbage collection mechanism take care of it?

swiatekm (commented):
Some preliminary comments about the API and the logistics of introducing this CRD to the operator.

config/samples/clusterobservability_v1alpha1_clusterobservability.yaml
frzifus: I totally see the point of keeping things simple. Would it then maybe make sense to skip the `signals` entry entirely and add something in case we get a request?

Just the naming of the signals seems to be a bit intransparent what it actually does. That means Ive to lookup the docs. At the same time I think it should be easy to block telemetry reported by internal services without needing to configure e.g. a network policy.

atoulme: I agree if we can live without toggles for now we should.

apis/v1alpha1/clusterobservability_types.go
```diff
@@ -0,0 +1,295 @@
...
+
+// ClusterObservabilitySpec defines the desired state of ClusterObservability.
+// This follows a simplified design using a single OTLP HTTP exporter for all signals.
+type ClusterObservabilitySpec struct {
+	// Signals defines which observability signals to collect and export.
+	// Must contain at least one signal type from: logs, traces, metrics, profiles
+	// +required
+	// +kubebuilder:validation:MinItems=1
+	// +listType=set
+	Signals []ObservabilitySignal `json:"signals"`
```
pavolloffay: I like the simplicity of disabling/enabling by signal type. 

However I would prefer if the CR design was future proof and allowed us to add more detailed "nobs" in the future if we have to (once we collect feedback from the end users). 

Perhaps we can change the ObservabilitySignal from string to a struct.

config/samples/clusterobservability_v1alpha1_clusterobservability.yaml
frzifus: Nice, then we simply get rid of it for now? :smiley:

frzifus:
@jinja2 Ive been unable to create a PR to your branch. I just created a draft addressing the comments on top. Feel free to merge/cherry-pick changes:
- https://github.com/open-telemetry/opentelemetry-operator/pull/4504

I placed a summary into the PR description.

cc @atoulme

internal/controllers/clusterobservability_controller.go
jinja2: cherry-picked from @frzifus updates

config/samples/clusterobservability_v1alpha1_clusterobservability.yaml
jinja2: I see a few different opinions ([here](https://github.com/open-telemetry/opentelemetry-operator/pull/4475#discussion_r2500276532) is another comment on this) on how many knobs we should expose. I think we should keep the toggle for telemetry types at least. We could make this optional, with all signals enabled by default, so users don’t need to set it explicitly unless they want to customize it. Let me know if that sounds reasonable.

frzifus: I personally would still question whats the benefit of having the option to disable specific signal types. Especally when a signal itself remains quite intransparent. Disabling `metrics` will disable collecting hostmetrics, metrics ingestion and whatever is behind it. Personally, I would advocate for leaving the option out completely for now. Adding knobs in the future should be easy. That way we also can move forward and have this discussion in the future.

Would be good to get some views form the others ( @open-telemetry/operator-approvers ) on this.

apis/v1alpha1/clusterobservability_types.go
frzifus: We also could trim it down to a more minimal set just to unblock this PR.

config/samples/clusterobservability_v1alpha1_clusterobservability.yaml
jinja2: My original intent here was to give user a simple knob for controlled/gradual adoption (try one telemetry area at a time, reduce initial blast radius), but I understand your concerns. I’ll drop the option for now and keep the default behavior simple.

apis/v1alpha1/clusterobservability_types.go
jinja2: Hey folks, apologies for the long delay but am back working on this.

I can make the exporter an opaque anyconfig type. I can add checks similar to [`validateExporter` in Instrumentation](https://github.com/open-telemetry/opentelemetry-operator/blob/v0.144.0/apis/v1alpha1/instrumentation_webhook.go#L288). Does that work?

config/samples/clusterobservability_v1alpha1_clusterobservability.yaml
jinja2: updated to remove signals config

apis/v1alpha1/clusterobservability_types.go
frzifus: That sounds great to me :)

frzifus: I am also fine with merging this large PR right now and change it  on a followup PR. wdyt @swiatekm ?

apis/v1alpha1/clusterobservability_types.go
```diff
@@ -0,0 +1,276 @@
...
+	// +optional
+	Insecure bool `json:"insecure,omitempty"`
+
+	// ServerName for TLS handshake. If empty, uses the hostname from endpoint.
+	// +optional
+	ServerName string `json:"server_name,omitempty"`
+}
+
+// SendingQueueConfig defines configuration for the sending queue.
+type SendingQueueConfig struct {
```
jaronoff97, line 105: I think this is slightly out of date, there are a few more dials [here](https://github.com/open-telemetry/opentelemetry-collector/blob/0c5a4ffc7d2a506f7d6463c3619233d01cdd9c98/exporter/exporterhelper/README.md?plain=1#L21-L31). I understand we don't want to expose everything, do we have guidelines for what config will be present / omitted?

bundle/community/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -429,18 +429,10 @@ spec:
...
-          - opentelemetrycollectors
-          verbs:
-          - get
-          - list
-          - patch
-          - update
-          - watch
-        - apiGroups:
-          - opentelemetry.io
-          resources:
```
jaronoff97, line 442: this is a combination of the below right?

docs/cluster-observability.md
jaronoff97, line 1: the diagram in here makes it seem like the openshift integration is a requirement, can we make it more clear when it's used / optional?

jaronoff97 (approved):
a few thoughts/questions but overall looks good

apis/v1alpha1/clusterobservability_types.go
atoulme: No, I think we need to build on your feedback and https://github.com/open-telemetry/opentelemetry-operator/pull/4475/changes/BASE..39be612211e98d9ccb126872b9e4541edd975242#r2499485317 to make a new issue and follow up separately.

apis/v1alpha1/clusterobservability_types.go
swiatekm: Yep, that works for me too.

docs/cluster-observability.md
frzifus: I guess the openshift part is optional, right? Aside from that we should add a part to determine the platform at some point anyway.

jinja2: Yeah, I updated the documentation to clarify the auto-detection feature in [this PR.](https://github.com/open-telemetry/opentelemetry-operator/pull/4668/changes)

Issue #4361, opened 2025-09-15
Support HTTPRoute creation by v1beta1 otel CR


frzifus:
### Component(s)

collector

### Is your feature request related to a problem? Please describe.

Gateway API users may dont want to deal with ingress / route objects.

### Describe the solution you'd like

Extend the CollectorCR by a httpRoute type.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

Rohlik:
This would be really helpful as the Gateway API is more popular now than it was a year ago.

PR #4510, opened 2025-11-19
support gatewayApi HTTPRoute creation by v1beta1 for otel CR

Horiodino:
### **Description**

Adds support for generating Gateway API `HTTPRoute` resources from the `OpenTelemetryCollector` v1beta1 spec. Includes the new `httpRoute` configuration.

### **Link to tracking Issue(s)**

* Resolves: **#4361**

### **Testing**

* added unit tests for `httpRoute` 
* ran controller locally with `httpRoute` enabled

### **Documentation**

* added API docs and changelog.

  ```yaml
  kind: OpenTelemetryCollector
  spec:
    mode: deployment
    httpRoute:
      enabled: true
      gateway: my-gateway        # required
      gatewayNamespace: default   (defaults to collector namespace)
      hostName: "otel.example.com" 
  ```

  When enabled, the operator automatically generates an `HTTPRoute` per HTTP port exposed by the collector’s configuration.

### DIFF
--- a/bundle/community/manifests/opentelemetry-operator.clusterserviceversion.yaml
+++ b/bundle/community/manifests/opentelemetry-operator.clusterserviceversion.yaml
@@ -409,6 +409,18 @@ spec:
           - get
           - list
         - apiGroups:
+          - gateway.networking.k8s.io
+          resources:
+          - httproutes
+          verbs:
+          - create
+          - delete
+          - get
+          - list
+          - patch
+          - update
+          - watch
+        - apiGroups:
           - monitoring.coreos.com
           resources:
           - podmonitors
--- a/bundle/community/manifests/opentelemetry.io_opentelemetrycollectors.yaml
+++ b/bundle/community/manifests/opentelemetry.io_opentelemetrycollectors.yaml
@@ -6324,6 +6324,22 @@ spec:
                 type: boolean
               hostUsers:
                 type: boolean
+              httpRoute:
+                properties:
+                  enabled:
+                    type: boolean
+                  gateway:
+                    type: string
+                  gatewayNamespace:
+                    type: string
+                  hostnames:
+                    items:
+                      type: string
+                    type: array
+                required:
+                - enabled
+                - gateway
+                type: object
               image:
                 type: string
               imagePullPolicy:
--- a/bundle/openshift/manifests/opentelemetry-operator.clusterserviceversion.yaml
+++ b/bundle/openshift/manifests/opentelemetry-operator.clusterserviceversion.yaml
@@ -409,6 +409,18 @@ spec:
           - get
           - list
         - apiGroups:
+          - gateway.networking.k8s.io
+          resources:
+          - httproutes
+          verbs:
+          - create
+          - delete
+          - get
+          - list
+          - patch
+          - update
+          - watch
+        - apiGroups:
           - monitoring.coreos.com
           resources:
           - podmonitors
--- a/bundle/openshift/manifests/opentelemetry.io_opentelemetrycollectors.yaml
+++ b/bundle/openshift/manifests/opentelemetry.io_opentelemetrycollectors.yaml
@@ -6323,6 +6323,22 @@ spec:
                 type: boolean
               hostUsers:
                 type: boolean
+              httpRoute:
+                properties:
+                  enabled:
+                    type: boolean
+                  gateway:
+                    type: string
+                  gatewayNamespace:
+                    type: string
+                  hostnames:
+                    items:
+                      type: string
+                    type: array
+                required:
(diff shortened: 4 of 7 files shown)


Horiodino:
cc @frzifus

jaronoff97:
@frzifus mind reviewing this one?

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,15 @@
...
+- name: controller
+  newName: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
+  newTag: 0.138.0-22-g1a1d9992
+patches:
+- patch: '[{"op":"add","path":"/spec/template/spec/containers/0/args/-","value":"--target-allocator-image=ghcr.io/open-telemetry/opentelemetry-operator/target-allocator:v0.138.0-22-g1a1d9992"}]'
+  target:
+    kind: Deployment
+- patch: '[{"op":"add","path":"/spec/template/spec/containers/0/args/-","value":"--operator-opamp-bridge-image=ghcr.io/open-telemetry/opentelemetry-operator/operator-opamp-bridge:v0.138.0-22-g1a1d9992"}]'
+  target:
+    kind: Deployment
```
frzifus:
```suggestion
```

internal/controllers/opentelemetrycollector_controller.go
```diff
@@ -379,6 +382,8 @@ func (r *OpenTelemetryCollectorReconciler) GetOwnedResourceTypes() []client.Obje
 		ownedResources = append(ownedResources, &routev1.Route{})
 	}
 
+	// TODO: Add Gateway API availability check similar to OpenShift Routes
+
```
frzifus, line 373: We should add this before merging this.

internal/manifests/collector/httpRoute.go
```diff
@@ -0,0 +1,150 @@
...
+func HTTPRoutes(params manifests.Params) ([]*gatewayv1.HTTPRoute, error) {
+	if !params.OtelCol.Spec.HttpRoute.Enabled {
+		return nil, nil
+	}
+
+	if params.OtelCol.Spec.Mode == v1beta1.ModeSidecar {
+		params.Log.V(3).Info("HTTPRoute settings are not supported in sidecar mode")
+		return nil, nil
+	}
+
```
frzifus, line 27: Maybe we can reuse the logic from the ingress entry (which also constructs routes) wdyt?

On the other hand having a httproute entry would allow us to simply deprecate `ingress` on a future `v1` version.

internal/manifests/collector/httpRoute.go
```diff
@@ -0,0 +1,150 @@
...
+			continue
+		}
+
+		name := naming.HTTPRoute(params.OtelCol.Name, port.Name)
+		labels := manifestutils.Labels(params.OtelCol.ObjectMeta, name, params.OtelCol.Spec.Image, ComponentOpenTelemetryCollector, params.Config.LabelsFilter)
+
+		// auto-generate ParentRefs from gateway name and namespace
+		gatewayNamespace := httpRouteConfig.GatewayNamespace
+		if gatewayNamespace == "" {
+			// default to collector namespace if gateway namespace is not specified or it should be default?
```
frzifus: The collector namespace should be fine. We should inform the user at some point that we used the collector namespace as fallback.

bundle/community/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -531,6 +543,8 @@ spec:
                 - --zap-log-level=info
                 - --zap-time-encoding=rfc3339nano
                 - --enable-nginx-instrumentation=true
+                - --target-allocator-image=ghcr.io/open-telemetry/opentelemetry-operator/target-allocator:v0.138.0-22-g1a1d9992
+                - --operator-opamp-bridge-image=ghcr.io/open-telemetry/opentelemetry-operator/operator-opamp-bridge:v0.138.0-22-g1a1d9992
```
frzifus:
```suggestion
```

internal/manifests/collector/httpRoute.go
Horiodino: > On the other hand, having an HTTPRoute entry would allow us to simply deprecate ingress in a future v1 version.

I don’t mind that suggestion, but I’d prefer to continue with the current changes since they keep things simpler.  and  migration from Ingress to HTTPRoute, allows both things to evolve independently.

Horiodino:
I’ve resolved the merge conflicts. If everything looks good on your end, please go ahead and merge the PR delaying it further may lead to more conflicts.

PR #4517, opened 2025-11-20
[Chore] Use full image names in tests

IshwarKanse:
**Description**

Replace short container image names with fully qualified image names (e.g., rabbitmq → docker.io/library/rabbitmq:latest)
 across all test manifests and Dockerfiles.

**Why**

Image pulls are failing on OpenShift due to short name enforcement. OpenShift enforces fully qualified image names as a
security measure to prevent ambiguous registry resolution and potential image spoofing attacks.

Without explicit registry paths, pods fail to start with errors related to short name resolution being blocked by the
container runtime policy.

**Changes**

- Updated all test YAML manifests with explicit docker.io registry paths
- Added explicit tags (:latest, :3, etc.) where previously omitted
- Updated Dockerfiles to use fully qualified base image names

### DIFF
--- a/tests/e2e-instrumentation/instrumentation-apache-multicontainer/01-install-app.yaml
+++ b/tests/e2e-instrumentation/instrumentation-apache-multicontainer/01-install-app.yaml
@@ -46,7 +46,7 @@ spec:
             capabilities:
               drop: ["ALL"]
         - name: myrabbit
-          image: rabbitmq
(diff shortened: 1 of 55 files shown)


swiatekm (approved):
We should really switch as many of these as possible to GHCR or ECR to avoid rate limits. Or, ideally, put them in the E2E test image archive.

Issue #4441, opened 2025-10-19
failed to verify certificate: x509: certificate signed by unknown authority


ImDevinC:
## Description
I have installed both cert-manager and opentelemetry-operator. Other services work fine with cert-manager. However opentelemetry-operator keeps showing this error in the log and I'm not getting metrics due to it.

## Steps to Reproduce
1. Apply config
2. Look at logs

## Expected Result
No errors about certificate, and metrics show up

## Actual Result
Log is filled with errors about invalid certificate, no metrics showing up

### Kubernetes Version

v1.31.4

### Operator version

0.136.0

### Collector version

0.136.0

## Environment
OS: Arch 6.11.8-arch1-2

### Log output

```shell
2025-10-19T21:52:46.839Z	error	targetallocator/manager.go:126	Failed to retrieve job list	{"resource": {"service.instance.id": "c6954f62-10ff-49df-9515-1965610f81d4", "service.name": "otelcol-contrib", "service.version": "0.136.0"}, "otelcol.component.id": "prometheus", "otelcol.component.kind": "receiver", "otelcol.signal": "metrics", "error": "Get \"https://otelcol-targetallocator:443/scrape_configs\": tls: failed to verify certificate: x509: certificate signed by unknown authority (possibly because of \"crypto/rsa: verification error\" while trying to verify candidate authority certificate \"otelcol-ca-cert\")"}
github.com/open-telemetry/opentelemetry-collector-contrib/receiver/prometheusreceiver/targetallocator.(*Manager).sync
	github.com/open-telemetry/opentelemetry-collector-contrib/receiver/prometheusreceiver@v0.136.0/targetallocator/manager.go:126
github.com/open-telemetry/opentelemetry-collector-contrib/receiver/prometheusreceiver/targetallocator.(*Manager).Start.func1
	github.com/open-telemetry/opentelemetry-collector-contrib/receiver/prometheusreceiver@v0.136.0/targetallocator/manager.go:81
github.com/cenkalti/backoff/v5.Retry[...]
	github.com/cenkalti/backoff/v5@v5.0.3/retry.go:87
github.com/open-telemetry/opentelemetry-collector-contrib/receiver/prometheusreceiver/targetallocator.(*Manager).Start
	github.com/open-telemetry/opentelemetry-collector-contrib/receiver/prometheusreceiver@v0.136.0/targetallocator/manager.go:91
[...]

PR #4526, opened 2025-11-24
fix(target-allocator): Extend target allocator CA certificate duration to prevent renewal race

schahal:
**Description:**

See https://github.com/open-telemetry/opentelemetry-operator/issues/4441 for details

The root cause (from the 3 separate accounts in the thread) seems to be that all three TargetAllocator certificates (CA, server, client) were created at the same time with the same 90-day duration.

When they all came up for renewal simultaneously (around day 60), there was a race condition where, **for example:*

- One of either the Server or Client cert gets renewed and signed by the original CA
- The CA gets renewed second (new CA cert + key)
- The other (e.g., Client) cert gets renewed and signed by the new CA
- This causes "certificate signed by unknown authority" errors

By giving the CA a 1-year duration while client/server certs keep the default 90-day duration, we ensure:

- The CA remains stable (e.g., longer period between renewals) while client/server certs renew
- Client and server certs are always signed by the same CA (unless there's a very tiny chance that they renew at same time as CA's 11-month mark)

- Resolves: https://github.com/open-telemetry/opentelemetry-operator/issues/4441

**Testing:**

- Updated test to validate duration

```
go test ./internal/manifests/targetallocator/ -run Certificate -v
=== RUN   TestCACertificate
=== RUN   TestCACertificate/Default_CA_Certificate
--- PASS: TestCACertificate (0.00s)
    --- PASS: TestCACertificate/Default_CA_Certificate (0.00s)
=== RUN   TestServingCertificate
=== RUN   TestServingCertificate/Default_Serving_Certificate
--- PASS: TestServingCertificate (0.00s)
    --- PASS: TestServingCertificate/Default_Serving_Certificate (0.00s)
=== RUN   TestClientCertificate
=== RUN   TestClientCertificate/Default_Client_Certificate
--- PASS: TestClientCertificate (0.00s)
    --- PASS: TestClientCertificate/Default_Client_Certificate (0.00s)
PASS
ok  	github.com/open-telemetry/opentelemetry-operator/internal/manifests/targetallocator	(cached)
```

Added a changelog gen entry

### DIFF
--- a/.chloggen/fix-ca-cert-renewal-race.yaml
+++ b/.chloggen/fix-ca-cert-renewal-race.yaml
@@ -5,7 +5,7 @@ change_type: bug_fix
 component: target allocator
 
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Fix CA certificate renewal race condition by extending CA certificate duration to 1 year
+note: Fix CA certificate race condition with client cert renewals by extending its duration and and renewal attempt.
 
 # One or more tracking issues related to the change
 issues: [4441]
@@ -14,6 +14,6 @@ issues: [4441]
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
 subtext: |
-  The CA certificate now has a 1-year duration (instead of the default 90 days) to prevent race conditions
+  The CA certificate now has a 2-year duration (instead of the default 90 days) to prevent race conditions
   where client and server certificates could be signed by different CA versions during simultaneous renewal.
   This ensures the CA remains stable while dependent certificates renew regularly.


internal/manifests/targetallocator/certificate.go
```diff
@@ -28,6 +29,10 @@ func CACertificate(params Params) *cmv1.Certificate {
 		Spec: cmv1.CertificateSpec{
 			IsCA:       true,
 			CommonName: naming.CACertificate(params.TargetAllocator.Name),
+			// Set CA certificate to 1 year (much longer than the default 90-day duration of client/server certs)
```
pavolloffay: Is the client/server certs renewal configurable? 

Would it make sense to make the CA renewal shorter? / What is the lowest safe value possible for the CA renewal?

schahal: > client/server certs renewal configurable?

No, when I'd looked at it, it was using cert-manager's defaults (which I had to look up, hence the comments for reference because i _**thought**_ it was 90 days, but had to look it up).

I think that's fine, for simplicity

>  make the CA renewal shorter? / What is the lowest safe value possible for the CA renewal?

In general, I've seen CAs pretty long-lived. 1-year seems reasonable (I think even like 10 years would be ok, but guessing original author(s) were erring on shorter lifecycle for CA - I don't have that history, so I picked 1 year.

I think 1 year, combined with @swiatekm suggestion adding a `renewBefore` would actually eliminate the race.

Will push a change.

schahal: https://github.com/open-telemetry/opentelemetry-operator/pull/4526/commits/1186ee270f27f0c0b7f6c22caa9983b0755e7475

internal/controllers/builder_test.go
```diff
@@ -3260,9 +3261,11 @@ prometheus_cr:
 						Subject: &cmv1.X509Subject{
 							OrganizationalUnits: []string{"opentelemetry-operator"},
 						},
-						CommonName: "test-ca-cert",
-						IsCA:       true,
-						SecretName: "test-ca-cert",
+						CommonName:  "test-ca-cert",
+						Duration:    &metav1.Duration{Duration: 8760 * time.Hour},
+						RenewBefore: &metav1.Duration{Duration: 2400 * time.Hour},
```
swiatekm: Can we make these constants and add a comprehensive comment explaining why the values are what they are? This is a subtle issue and I want it to be very clear to future readers why we set these to what they are.

We should probably explicitly set the duration on the client and server certificates, even if we're keeping it at the default. It's not inconceivable for cert-manager to change these defaults in the future, and I'd like to be as explicit as possible about the effect we're trying to achieve here.

What I'd prefer is somethling like:

```go
const (
  clientCertDuration = time.Hour * 24 * 90
  caCertRenewBefore = clientCertDuration * 2
  caCertDuration = clientCertDuration * 8
)
```

schahal: Added in https://github.com/open-telemetry/opentelemetry-operator/pull/4526/commits/3e33c644404616d338171e6fa8da83e3eb88207d 

(one tweak is I added a "+ 1day" offset on CACertRenewBefore to handle prevent cases where all 3 are simultaneously renewed at same time - explained in comment).

schahal:
Updated!

> I'd accept if you compile the operator with these values set very low and verify that it works in a test cluster.

Looks like cert-manager has a minimum duration of 1 hour enforced by its admission webhook:

```
"admission webhook \"webhook.cert-manager.io\" denied the request: spec.duration: Invalid value: 24m0s: certificate duration must be greater than 1h0m0s"
```

So made it 1h, updated the image and side-loaded into my KIND cluster... not waiting for renewal because of the time, but validated certs created with expected renewal dates:

```
$ kubectl get certificate -n cert-test -o json | jq -r '.items[] | "\(.metadata.name):\n  Duration: \(.spec.duration)\n  RenewBefore: \(.spec.renewBefore // "not set")\n  NotBefore: \(.status.notBefore)\n  NotAfter: \(.status.notAfter)\n  RenewalTime: \(.status.renewalTime)\n"'
cert-test-collector-ca-cert:
  Duration: 12h0m0s
  RenewBefore: 4h0m0s
  NotBefore: 2025-11-30T16:46:04Z
  NotAfter: 2025-12-01T04:46:04Z
  RenewalTime: 2025-12-01T00:46:04Z

cert-test-collector-ta-client-cert:
  Duration: 1h30m0s
  RenewBefore: not set
  NotBefore: 2025-11-30T16:46:07Z
  NotAfter: 2025-11-30T18:16:07Z
  RenewalTime: 2025-11-30T17:46:07Z

cert-test-collector-ta-server-cert:
  Duration: 1h30m0s
  RenewBefore: not set
  NotBefore: 2025-11-30T16:46:07Z
  NotAfter: 2025-11-30T18:16:07Z
  RenewalTime: 2025-11-30T17:46:07Z

File: internal/manifests/targetallocator/certificate.go (hunk 1 above)
Comment (schahal) [reply internal/manifests/targetallocator/certificate.go]: A small update, @pavolloffay: thru progression of this PR, we did end up setting the client/server cert duration as explicit constants (rather than relying on cert-manager defaults).

File: .chloggen/fix-ca-cert-renewal-race.yaml [hunk 3]
```diff
@@ -0,0 +1,19 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: target allocator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix CA certificate renewal race condition by extending CA certificate duration to 1 year
```
atoulme: this changelog message seems out of step with the changes. The CA certificate duration is set to 720 days, roughly 2 years. Am I missing something?

swiatekm: Good catch, we ended up increasing it, but haven't updated the changelog message.

schahal: Ah yes, this had changed thru the progression of the PR

Updated to be more accurate in https://github.com/open-telemetry/opentelemetry-operator/pull/4526/commits/48393dad76882a1c1bcf4400b52a29772838af02

schahal: Of course, now after the changelog change, some CI tests seem to randomly fail /s 🤦

swiatekm: I restarted them. Github has been flaky these past few days, and the tests were having trouble installing tools.

PR #4538, opened 2025-12-03
[Chore] Update and fix tests

IshwarKanse:
The PR adds minor test fixes.

### DIFF
--- a/tests/e2e-openshift/export-to-cluster-logging-lokistack/otel-collector.yaml
+++ b/tests/e2e-openshift/export-to-cluster-logging-lokistack/otel-collector.yaml
@@ -42,7 +42,7 @@ metadata:
   name: otel
   namespace: openshift-logging
 spec:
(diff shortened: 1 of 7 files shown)


tests/e2e-openshift/export-to-cluster-logging-lokistack/otel-collector.yaml
```diff
@@ -42,7 +42,7 @@ metadata:
   name: otel
   namespace: openshift-logging
 spec:
-  image: ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-contrib:0.116.1
+  image: ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-contrib:0.140.0
```
swiatekm, line 45: Any reason we can't just leave this empty?

IshwarKanse: k8sattributes processor, bearertokenauth extension are not available in the default image which are required for the test.

PR #4550, opened 2025-12-05
[chore]: prepare v0.141.0 release

TylerHelmuth:


RELEASE.md
```diff
@@ -4,22 +4,22 @@ Steps to release a new version of the OpenTelemetry Operator:
 
 1. Create a `Prepare release x.y.z` pull request with the following content:
    1. Set the version you're releasing as an environment variable for convenience: `export VERSION=0.n+1.0`
-   1. Update `versions.txt`
+   2. Update `versions.txt`
```
pavolloffay: This can be reverted.

frzifus (approved):

RELEASE.md
TylerHelmuth: I recognize markdown can handle rendering all the `1`s correctly based on the tabbing, but my IDE complains about it. Does it have to be reverted?

swiatekm: My understanding is that this is a the recommended way of numbering long lists, as it avoids needing to renumber after adding an element in the middle.

TylerHelmuth: Ok reverted

Issue #3134, opened 2024-07-15
Question about unknown/unknown Tag in ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator Docker Image


wzy531a:
### Describe the issue you're reporting

Hi,

I noticed that the Docker image [ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator](https://github.com/open-telemetry/opentelemetry-operator/pkgs/container/opentelemetry-operator%2Fopentelemetry-operator) contains a tag labeled unknown/unknown. Can someone clarify what this tag represents and what its purpose is?

Thank you!

jaronoff97:
@wzy531a can you link that image? I looked through GHCR and couldn't find it.

wzy531a:
> @wzy531a can you link that image? I looked through GHCR and couldn't find it.

Yes, when I tried to pull the unknown image, Docker indicated that the operating system is not supported. For now, I tried the following:
`[root@k8s-master ~]# docker pull ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator:main@sha256:bdf756140620a2e71239185c1fc136291d2ea78c65c36ed861cc5c87f65129fb

ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator@sha256:bdf756140620a2e71239185c1fc136291d2ea78c65c36ed861cc5c87f65129fb: Pulling from open-telemetry/opentelemetry-operator/opentelemetry-operator
e00d8f0bad61: Pulling fs layer
operating system is not supported`

wzy531a:
> what OS are you using?

I am using Linux-x64. The 'unknown' tag is provided for systems where the specific OS is not identified. If the system is not Linux, you can try using the 'unknown' tag image.

jaronoff97:
Im unclear on what the issue you're experiencing is now. Can you share more information, including logs and OS.

wzy531a:
> Im unclear on what the issue you're experiencing is now. Can you share more information, including logs and OS.

I want to know what the meaning of unknown OS/Arch is?
For example, Linux/amd64 corresponds to a Linux system running on a 64 bit AMD architecture, while Linux/arm64 corresponds to a Linux system running on a 64 bit ARM architecture. So, what is unknown

pavolloffay:
[...]

PR #4551, opened 2025-12-07
ci: disable provenance and sbom to fix the `unknown/unknown` platform on ghcr

gautamkrishnar:
**Description:** 
 Added `provenance: false` and `sbom: false` parameters to all Docker/build-push-action steps across 12 workflow
  files. This will fix the issue of ghcr showing `unknown/unknown` in the GHCR UI.

**Caveats:**
This change disables two security/transparency features:
1. **Provenance Attestations**: Provides cryptographically signed build transparency information (build platform,
   source repository, build parameters)
2. **SBOM** (Software Bill of Materials): Provides dependency information for security scanning and compliance

Ref: https://docs.docker.com/build/ci/github-actions/attestations/

I also noticed that you guys are pushing to Docker.io as well. For Docker.io, this fix is not really necessary.  Let me know if you guys want to separate the pushing to Docker.io separately enabling both provenance and SBOM.

- Resolves: #3134

**Testing:** Fixes done as per the discussion here: https://github.com/orgs/community/discussions/45969#discussioncomment-13352049
**Documentation:** n/a

### DIFF
--- a/.github/workflows/publish-autoinstrumentation-apache-httpd.yaml
+++ b/.github/workflows/publish-autoinstrumentation-apache-httpd.yaml
@@ -72,5 +72,7 @@ jobs:
           build-args: version=${{ env.VERSION }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          provenance: false
+          sbom: false
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache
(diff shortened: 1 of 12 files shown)


.github/workflows/publish-autoinstrumentation-apache-httpd.yaml
```diff
@@ -72,5 +72,7 @@ jobs:
           build-args: version=${{ env.VERSION }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          provenance: false
+          sbom: false
```
jaronoff97, line 76: I would rather us figure out how to make provenance and sboms work than disable this feature. We should be improving our security posture when possible.

Issue #4367, opened 2025-09-17
Finalizer in the open telemetry collector CO prevents successful namespace deletion when managed operator is uninstalled prior to CO uninstall/deletion


thiruvikraman-subburaman:
### Component(s)

collector

## Description
When opentelemetry operator and CO(operand) installed separately , whenever OpenTelemetry Collector custom resource( CO) gets created/modifies, the opentelemetry collector controller adds the finalizer `opentelemetrycollector.opentelemetry.io/finalizer`  to the Custom Resource and removes it when CR gets deleted. But, it then brings the strict ordering on uninstall/cleanup that operand(CR) to be uninstalled/deleted before Operator and which leads to preventing the successful namespace deletion as the finalizer block still exists. 

## Steps to Reproduce
1) Install OpenTelemetry Operator 
2) Install OpenTelemetry collector (CR)
3) uninstall OpenTelemetry Operator 
4) namespace deletion or uninstall OpenTelemetry collector

## Expected Result
Namespace deletion should succeed, and it should not depend on the order of uninstallation of operator and managed CO

## Actual Result
namespace deletion gets stuck due to the finalizer object references not removed because of ordering issues on uninstallation

### Kubernetes Version

1.27.0

### Operator version

0.135.0

### Collector version

0.135.0

## Environment
OS: (e.g., "Ubuntu 20.04")
Compiler(if manually compiled): (e.g., "go 14.2")

separate helm chart of operator and collector 

![Image](https://github.com/user-attachments/assets/67d7911d-29c5-4301-ac11-6becb3ec06b5)

### Log output

```shell
Attached the sample screenshot
```

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

PR #4574, opened 2025-12-17
Add finalizer only for cluster scope objects

smanojj:
**Description:**

- Add finalizers to CR only when cluster roles and bindings for SA are created by Operator
- Caused by #2938

- Resolves: #4367

### DIFF
--- a/internal/controllers/opentelemetrycollector_controller.go
+++ b/internal/controllers/opentelemetrycollector_controller.go
@@ -285,12 +285,10 @@ func (r *OpenTelemetryCollectorReconciler) Reconcile(ctx context.Context, req ct
 	}
 
 	// Add finalizer for this CR
-	if !controllerutil.ContainsFinalizer(&instance, collectorFinalizer) {
-		if controllerutil.AddFinalizer(&instance, collectorFinalizer) {
-			err = r.Update(ctx, &instance)
-			if err != nil {
-				return ctrl.Result{}, err
-			}
+	if maybeAddFinalizer(params, &instance) {
+		err = r.Update(ctx, &instance)
+		if err != nil {
+			return ctrl.Result{}, err
 		}
 	}
 
@@ -394,4 +392,11 @@ func (r *OpenTelemetryCollectorReconciler) finalizeCollector(ctx context.Context
 		return deleteObjects(ctx, r.Client, r.log, objects)
 	}
 	return nil
+}
+
+func maybeAddFinalizer(params manifests.Params, instance *v1beta1.OpenTelemetryCollector) bool {
+	if params.Config.CreateRBACPermissions == rbac.Available && !controllerutil.ContainsFinalizer(instance, collectorFinalizer) {
+		return controllerutil.AddFinalizer(instance, collectorFinalizer)
+	}
+	return false
 }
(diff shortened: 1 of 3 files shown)


internal/controllers/opentelemetrycollector_controller.go
```diff
@@ -249,7 +249,7 @@ func (r *OpenTelemetryCollectorReconciler) Reconcile(ctx context.Context, req ct
 
 	// We have a deletion, short circuit and let the deletion happen
 	if deletionTimestamp := instance.GetDeletionTimestamp(); deletionTimestamp != nil {
-		if controllerutil.ContainsFinalizer(&instance, collectorFinalizer) {
+		if params.Config.CreateRBACPermissions == rbac.Available && controllerutil.ContainsFinalizer(&instance, collectorFinalizer) {
```
swiatekm: We shouldn't have a condition here. If the CR has a finalizer, we should honor it.

internal/controllers/opentelemetrycollector_controller.go
```diff
@@ -285,7 +285,7 @@ func (r *OpenTelemetryCollectorReconciler) Reconcile(ctx context.Context, req ct
 	}
 
 	// Add finalizer for this CR
-	if !controllerutil.ContainsFinalizer(&instance, collectorFinalizer) {
+	if params.Config.CreateRBACPermissions == rbac.Available && !controllerutil.ContainsFinalizer(&instance, collectorFinalizer) {
 		if controllerutil.AddFinalizer(&instance, collectorFinalizer) {
```
swiatekm: I'd prefer both this check and `AddFinalizer` to be rolled into a single `maybeAddFinalizer` function.

It'd also be nice to have a unit test for this behavior.

smanojj: Done

swiatekm:
@smanojj the changes look good to me. Please add a changelog entry by running `make chlog-new` and filling out the template it creates. You'll also need to sign the CLA before we can merge your PR.

internal/controllers/opentelemetrycollector_controller.go
```diff
@@ -395,3 +393,10 @@ func (r *OpenTelemetryCollectorReconciler) finalizeCollector(ctx context.Context
 	}
 	return nil
 }
+
+func maybeAddFinalizer(params manifests.Params, instance *v1beta1.OpenTelemetryCollector) bool {
+	if params.Config.CreateRBACPermissions == rbac.Available && !controllerutil.ContainsFinalizer(instance, collectorFinalizer) {
```
pavolloffay, line 398: Alternatively we could dynamically check the managed objects e.g. 

```
func IsNamespaced(obj client.Object, scheme *runtime.Scheme, restMapper meta.RESTMapper) (bool, error) {
	gvk, err := apiutil.GVKForObject(obj, scheme)
	if err != nil {
		return false, err
	}
	mapping, err := restMapper.RESTMapping(gvk.GroupKind(), gvk.Version)
	if err != nil {
		return false, err
	}
	return mapping.Scope.Name() == meta.RESTScopeNameNamespace, nil
}
```

PR #4597, opened 2026-01-09
fix(target-allocator): add TLS certificate hot-reload for mTLS

CharlieTLe:
**Description:** Fix TLS certificate hot-reload for mTLS connections in the Target Allocator.

The Target Allocator now automatically reloads TLS certificates when they are renewed
by cert-manager. Previously, certificate renewals required a pod restart because
certificates were only loaded once at startup.

The fix introduces a `CertificateReloader` struct that:
- Uses fsnotify to watch the certificate directory for changes
- Dynamically reloads certificates when files are modified
- Provides a `GetCertificate` callback for seamless TLS handshakes with updated certs

This enables seamless certificate rotation without downtime.

- Resolves: #4368

**Testing:** Added e2e chainsaw test that verifies certificate rotation by:
1. Capturing the initial certificate date before renewal
2. Running `cmctl renew` to trigger certificate renewal
3. Verifying the new certificate date differs from the initial date
4. Confirming the Target Allocator continues serving with the renewed certificate

**Documentation:** Added chloggen changelog entry describing the fix.


cmd/otel-allocator/internal/config/tls_reloader.go
```diff
@@ -0,0 +1,135 @@
...
+			return ctx.Err()
+
+		case event, ok := <-watcher.Events:
+			if !ok {
+				return nil
+			}
+
+			// In Kubernetes, secret updates create a new symlink target.
+			// We look for Create or Write events on any file in the directory.
+			if event.Op&(fsnotify.Create|fsnotify.Write) != 0 {
```
pavolloffay: This implementation seems fine. In the contrib repo we had a race condition in one watcher impl https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/44104/changes#diff-6e4e17a95db08bf6e5ca6307cea3be19b08d3f3efe62bd6808ec4ffec7871938L128

pavolloffay (commented):
Great work. Just some minor comments.

cmd/otel-allocator/internal/config/tls_reloader.go
CharlieTLe: Thanks for the review and link to another implementation, @pavolloffay. This implementation looks at the cert directory instead of the actual files, so we should be good.

CharlieTLe:
Hi @jaronoff97, could you help review? Thanks!

cmd/otel-allocator/internal/config/tls_reloader.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	defer watcher.Close()
+
+	// Watch the directory containing the certificates.
+	// In Kubernetes, secrets are mounted as symlinks that get updated atomically,
+	// so we need to watch the directory for changes.
+	certDir := filepath.Dir(r.certPath)
+	if err := watcher.Add(certDir); err != nil {
+		return err
+	}
```
swiatekm: Technically, the directories for the three files could be different, so I think we should add all of them here.

cmd/otel-allocator/internal/config/tls_reloader.go
swiatekm: Don't we also need to check if the event involves the file we're interested in? We could have other files in this directory. Technically, this would just cause spurious reloads, so not a big deal, but if we can avoid them easily, we should do so.

cmd/otel-allocator/internal/config/tls_reloader.go
```diff
@@ -0,0 +1,135 @@
...
+	keyPath   string
+	caPath    string
+	cert      *tls.Certificate
+	clientCAs *x509.CertPool
+	mu        sync.RWMutex
+	logger    logr.Logger
+}
+
+// NewCertificateReloader creates a new CertificateReloader and loads the initial certificates.
+func NewCertificateReloader(certPath, keyPath, caPath string, logger logr.Logger) (*CertificateReloader, error) {
```
swiatekm: We should have some basic unit tests for this object.

jaronoff97 (commented):
It looks like @swiatekm got to all the comments i was going to say and more haha, otherwise this looks great, excited for this functionality to be in here!

cmd/otel-allocator/internal/config/tls_reloader.go
CharlieTLe: Added some, thanks!

cmd/otel-allocator/internal/config/tls_reloader.go
CharlieTLe: Added a debouncer since we're dealing with symlinks.

cmd/otel-allocator/internal/config/tls_reloader.go
```diff
@@ -68,6 +80,11 @@ func (r *CertificateReloader) Reload() error {
 		"keyPath", r.keyPath,
 		"caPath", r.caPath)
 
+	// Call test callback if set (only used in tests)
+	if r.testReloadCallback != nil {
+		r.testReloadCallback()
+	}
+
```
swiatekm: Instead of adding fields to this struct just for testing, you can check the logs. Or verify that the cert values have changed.

cmd/otel-allocator/internal/config/tls_reloader.go
```diff
@@ -86,6 +103,32 @@ func (r *CertificateReloader) GetClientCAs() *x509.CertPool {
...
 
+// scheduleReload schedules a certificate reload after the debounce delay.
+// If a reload is already scheduled, it resets the timer.
+func (r *CertificateReloader) scheduleReload() {
+	r.timerMu.Lock()
+	defer r.timerMu.Unlock()
+
+	if r.reloadTimer != nil {
+		// Stop existing timer and drain channel if it already fired
+		if !r.reloadTimer.Stop() {
```
swiatekm: Doesn't this cause us to cancel the current timer every time we call `scheduleReload`? In principle, we could keep cancelling forever and never reload if the file is updated more frequently than the debounce delay.

cmd/otel-allocator/internal/config/tls_reloader_test.go
```diff
@@ -0,0 +1,637 @@
...
+	defer cancel()
+
+	// Start watcher in background
+	watcherDone := make(chan error, 1)
+	go func() {
+		watcherDone <- reloader.Watch(ctx)
+	}()
+
+	// Wait for watcher to start
+	time.Sleep(50 * time.Millisecond)
```
swiatekm: This is likely to lead to flaky tests. Please find a way to wait for this deterministically.

cmd/otel-allocator/internal/config/tls_reloader.go
CharlieTLe: Yeah, I've added a maximum time it will wait so it won't keep cancelling forever. Thanks for calling this out.

cmd/otel-allocator/internal/config/tls_reloader_test.go
CharlieTLe: Fixed.

cmd/otel-allocator/internal/config/tls_reloader.go
```diff
@@ -0,0 +1,221 @@
...
+	return r.clientCAs
+}
+
+// scheduleReload schedules a certificate reload after the debounce delay.
+// If a reload is already scheduled, it resets the timer.
+//
+// To prevent timer starvation from continuous events, this implements a
+// maximum debounce wait time. Even if events keep arriving, a reload is
+// guaranteed to happen within maxDebounceWait from the first event.
+func (r *CertificateReloader) scheduleReload() {
```
swiatekm: This implementation just seems unnecessarily complicated to me. Our debounce time is short and we don't really care if we reload immediately after the files change on disk. How about:

1. Add a `needsReload` boolean attribute.
2. `scheduleReload` just sets `needsReload` to true.
3. We run a goroutine which uses a ticker and reloads if `needsReload` is set to true.

Would there be a problem with that?

CharlieTLe: Yeah, I did some looking around and saw the existence of https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/certwatcher which seems to fit our needs. It didn't include support for reloading the CA cert as well, but I've added  a callback to also reload the CA cert whenever the tls server cert gets reloaded.

cmd/otel-allocator/internal/config/config.go
```diff
@@ -388,27 +389,57 @@ func ValidateConfig(config *Config) error {
...
+	caReloader, err := NewCAReloader(c.CAFilePath, logger)
 	if err != nil {
-		return nil, err
+		return nil, nil, fmt.Errorf("failed to create CA reloader: %w", err)
 	}
 
-	caCertPool := x509.NewCertPool()
-	caCertPool.AppendCertsFromPEM(caCert)
+	// Register callback to reload CA when server cert changes
+	// Since Kubernetes updates secrets atomically, the CA will be updated at the same time
```
swiatekm, line 406: Is this a guarantee Kubernetes explicitly gives, or does it happen to be its current behavior?

cmd/otel-allocator/internal/config/ca_reloader_test.go
```diff
@@ -0,0 +1,111 @@
...
+	go func() {
+		defer wg.Done()
+		for i := 0; i < 10; i++ {
+			newCAPEM, _ := generateTestCertificate(t)
+			require.NoError(t, os.WriteFile(caPath, newCAPEM, 0600))
+			err := reloader.Reload()
+			assert.NoError(t, err)
+			time.Sleep(10 * time.Millisecond)
+		}
+	}()
```
swiatekm: If we're going to test this, we should try to ensure the accesses and reloads actually happen concurrently. As you have these written, you're at the mercy of the Go runtime scheduler and when it decides to start your goroutines, which can also vary wildly depending on the host machine's resource constraints.

I'd also be fine removing this test, because I don't think it does much other than verifying that you've put locks in the appropriate spots, even if it does work.

cmd/otel-allocator/internal/config/ca_reloader.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	caCertPool := x509.NewCertPool()
+	if !caCertPool.AppendCertsFromPEM(caCert) {
+		return fmt.Errorf("failed to parse CA certificate at %s", r.caPath)
+	}
+
+	r.mu.Lock()
+	r.clientCAs = caCertPool
+	r.mu.Unlock()
```
swiatekm: There's a potential race condition here, where if Reload is called concurrently as the ca file changes, we can end up with the older value if we're unlucky. I'd just make the mutex cover the whole function. This isn't performance sensitive, we don't mind blocking while we read the file.

Alternatively, you can check timestamps on the file, but I don't think that's necessary.

cmd/otel-allocator/internal/config/config.go
CharlieTLe: Yeah, the CA cert is part of the Certificate resource that cert-manager manages. It gets updated at the same time as the cert.

swiatekm: Are you sure about that? My understanding is that the `Certificate` resource manages a `Secret`, and that's what gets mounted into the container. If the `Secret` changes, kubelet will eventually remount the files, and that's what we observe. But I'm not confident the remounts are guaranteed to happen atomically.

cmd/otel-allocator/internal/config/ca_reloader.go
CharlieTLe: Added a lock to the entire function. Thanks for the callout.

cmd/otel-allocator/internal/config/ca_reloader_test.go
CharlieTLe: Test removed.

cmd/otel-allocator/internal/config/config.go
CharlieTLe: Yep, Certificate manages a Secret which has the tls.crt, tls.key, and ca.crt. All of these are mounted in the pod in a timestamped directory by the kubelet. The files that the TA knows about are symlinks(1) to a symlink(2) to the timestamped directory. When an update occurs, the kubelet will create a new timestamped directory with the tls.crt, tls.key, and ca.crt files then update the symlink(2) to point to the new timestamped directory. This is how an atomic update will happen for all three files at once.

The implementation for this happens in [AtomicWriter](https://pkg.go.dev/k8s.io/kubernetes/pkg/volume/util#AtomicWriter). A nice write up of how it works can be found [here](https://ahmet.im/blog/kubernetes-inotify/#atomic-file-writes-on-kubernetes).

swiatekm: Understood, thanks for the detailed answer! I'm good with this.

Issue #3308, opened 2024-09-27
Allow SDK injection into initContainers


Joibel:
### Component(s)

auto-instrumentation

### Is your feature request related to a problem? Please describe.

I am adding tracing to argo-workflows to allow collection of a trace of the actual workflow.

When a workflow executes a workload in a pod it will create the pod with an initContainer which does various bits of setup of the workflow, including downloading of any artifacts (files) which the pod may use during execution.

I would like to be able to capture this as part of the overall workflow trace.

### Describe the solution you'd like

When a `container-names` annotation is present on a pod I would like the SDK environment variables to be present in the initContainers whose names are present in the annotation.

### Describe alternatives you've considered

I can manually inject the constant SDK variables, but this is also the case for anyone using the operator annotations for SDK injection. I would like the convenience of the operator to help me here too, especially for the non-constant values.

### Additional context

I have a POC which does this. The code change is relatively small.

I am unsure whether any of the specific injectors should also honor this on initContainers - so would like guidance on that.

pavolloffay:
This makes sense. I think we should not limit this only to SDK but offer support for all languages.

pavolloffay:
A PR is welcome

Joibel:
> This makes sense. I think we should not limit this only to SDK but offer support for all languages.

Can you clarify what you mean here please? I only have limited understanding of the auto-instrumentation mechanisms. I'm pretty sure we can't auto-instrument a go init container as that involves a sidecar which I don't believe you can do with an initContainer.

The other auto-instrumentation all(?) rely on initContainers with shared volumes, so I guess they could be made to work and auto-instrumentation if we ensure the instrumentation initContainer runs before them, is that what you were hoping for them?

Yamakaky:
Additional usecase: Kubernetes recently added native support for sidecars, so having auto instrumentation for init containers should be more frequent.

https://kubernetes.io/blog/2023/08/25/native-sidecar-containers/

Joibel:
I do intend to return to this, but my focus has been elsewhere for a while. Equally, if someone else wants to take it on, please do.

cdemi:
This would be really useful for us. We have a database migration app that runs as initContainer and it supports OpenTelemtry. Would be really cool if we could inject auto-instrumentation environment variables

PR #4598, opened 2026-01-09
refactor(instrumentation): use common language lists

Joibel:
**Description:** 

This is a pure refactor, creating a single list of all instrumentations and iterate over it.

I need to change `areInstrumentedContainersCorrect` for #3308 and to avoid having to do it 8 times, did this refactor.

- part of #3308, does not close it

**Testing:**

Built locally, and ran unit tests

**Documentation:**

None required, pure refactor


.chloggen/refactor-instrumentation.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
```
iblancasa: I think we don't need a changelog for this

Joibel: Would you rather I titled it with chore? (I can't add labels)
https://github.com/open-telemetry/opentelemetry-operator/blob/main/CONTRIBUTING.md#adding-a-changelog-entry

Joibel: Done

swiatekm (approved):
LGTM, thanks for cleaning this up.

Issue #4603, opened 2026-01-12
Operator ServiceMonitor not created on OpenShift - missing args and invalid scheme value


IshwarKanse:
### Component(s)

Operator

## Description
  The OpenTelemetry Operator fails to create a ServiceMonitor for its own metrics on OpenShift clusters. This prevents operator controller metrics (e.g., `controller_runtime_reconcile_time_seconds_count{controller="opentelemetrycollector"}) from being scraped by Prometheus.`

  Two issues contribute to this failure:

  1. Missing operator args: PR #4576 (Remove kube-rbac-proxy) inadvertently removed critical operator flags from the OpenShift deployment. The manager_auth_proxy_tls_patch.yaml strategic merge patch overwrites the args array set by manager-patch.yaml, removing flags like --create-sm-operator-metrics=true, --enable-leader-election, --enable-cr-metrics=true, etc.
  2. Invalid scheme value: PR #3858 (Bump prometheus group) updated the prometheus-operator dependency where SchemeHTTPS constant changed to uppercase "HTTPS". The ServiceMonitor CRD validation only accepts lowercase values, causing creation to fail.

## Steps to Reproduce
  1. Deploy the OpenTelemetry Operator on an OpenShift cluster via OLM built off the latest upstream main branch.
`  oc label namespace opentelemetry-operator openshift.io/cluster-monitoring=true`
  3. Check if ServiceMonitor is created:
  `oc get servicemonitor -n opentelemetry-operator`
  4. Check operator logs for errors:
`  oc logs -n opentelemetry-operator -l app.kubernetes.io/name=opentelemetry-operator | grep -i servicemonitor`

## Expected Result
  - A ServiceMonitor named opentelemetry-operator-metrics-monitor should be created in the operator namespace
  - Operator controller metrics should be available in Thanos/Prometheus

## Actual Result
  - No ServiceMonitor is created for the operator
  - Operator logs show:
`  error creating Service Monitor for operator metrics: ServiceMonitor.monitoring.coreos.com "opentelemetry-operator-metrics-monitor" is invalid: spec.endpoints[0].scheme: Unsupported value: "HTTPS": supported values: "http", "https"`
  - The --create-sm-operator-metrics=true flag is missing from the operator deployment args (only TLS cert args are present)

### Kubernetes Version

v1.33.6

### Operator version

opentelemetry-operator.v0.142.0

### Collector version

0.142.0

PR #4576, opened 2025-12-17
Remove kube-rbac-proxy and use builtin controller-runtime approach

pavolloffay:
**Description:** <Describe what has changed.>

  The operator now uses controller-runtime built-in auth for the metrics server. It is disabled by default.
  The client accessing the metrics endpoint must have the following RBAC rule `nonResourceURLs: "/metrics", verbs: get`.
  Which didn't change from the previous approach of using kube-rbac-proxy.
  
  This changes adds following flags to the operator:
  `metrics-secure` - enables authentication and authorization for the metrics server. If no TLS certificates are provided, self signed certificates will be generated.
  `metrics-tls-cert-file` - TLS certificate file for the metrics server
  `metrics-tls-key-file` - TLS private key file for the metrics server

**Link to tracking Issue(s):** <Issue number if applicable>

- Resolves: #3369

**Testing:** <Describe what testing was performed and which tests were added.>

```yaml
# This file contains a deployment that queries the operator metrics endpoint every 5 seconds
# along with the required RBAC to access the metrics endpoint.
#
# Usage:
#   kubectl apply -f hack/metrics-test-client.yaml
#   kubectl logs -f -n opentelemetry-operator-system deployment/metrics-test-client
#
# Cleanup:
#   kubectl delete -f hack/metrics-test-client.yaml

apiVersion: v1
kind: ServiceAccount
metadata:
  name: metrics-test-client
  namespace: opentelemetry-operator-system

---
# ClusterRole to access the metrics endpoint
# This grants permission to GET /metrics on the operator
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: metrics-test-client
rules:
# Required for accessing the /metrics endpoint with authentication
- nonResourceURLs:
  - /metrics
  verbs:
  - get

---
# Bind the ClusterRole to the ServiceAccount
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: metrics-test-client
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: metrics-test-client
subjects:
- kind: ServiceAccount
  name: metrics-test-client
  namespace: opentelemetry-operator-system

---
# Deployment that queries the metrics endpoint every 5 seconds
apiVersion: apps/v1
kind: Deployment
metadata:
  name: metrics-test-client
  namespace: opentelemetry-operator-system
  labels:
    app: metrics-test-client
spec:
  replicas: 1
  selector:
[...]

PR #4604, opened 2026-01-12
Fix operator ServiceMonitor not being created for controller metrics

IshwarKanse:
  This PR fixes a bug where the OpenTelemetry Operator fails to create a ServiceMonitor for its own metrics on OpenShift clusters.

  Two issues were identified and fixed:

  1. Missing operator args in OpenShift deployment: PR #4576 (Remove kube-rbac-proxy) inadvertently removed critical operator flags from the OpenShift deployment. The manager_auth_proxy_tls_patch.yaml is a strategic merge patch that replaces the entire args array, overwriting the args set by manager-patch.yaml. This removed flags like --create-sm-operator-metrics=true, --enable-leader-election, --enable-cr-metrics=true, etc.

Fix: Move TLS cert args (--metrics-tls-cert-file, --metrics-tls-key-file) to manager-patch.yaml and remove the args section from manager_auth_proxy_tls_patch.yaml.

  2. Invalid scheme value rejected by CRD validation: PR #3858 (Bump prometheus group) updated the prometheus-operator dependency where the SchemeHTTPS constant is uppercase "HTTPS". The ServiceMonitor CRD validation only accepts lowercase values, causing creation to fail with:
  spec.endpoints[0].scheme: Unsupported value: "HTTPS": supported values: "http", "https"

Fix: Change ptr.To(monitoringv1.SchemeHTTPS) to ptr.To(monitoringv1.Scheme("https")) in internal/operator-metrics/metrics.go.

- Resolves: #4603
- https://issues.redhat.com/browse/TRACING-5919 

  Tested on OpenShift cluster by running the e2e-openshift monitoring test:
  chainsaw test --skip-delete tests/e2e-openshift/monitoring

  The test verifies that:
  - Operator ServiceMonitor is created in the operator namespace
  - Operator controller metrics (controller_runtime_reconcile_time_seconds_count{controller="opentelemetrycollector"}) are available in Thanos/Prometheus

No documentation changes required.

### DIFF
--- a/tests/e2e/operator-metrics/assert-servicemonitor.yaml
+++ b/tests/e2e/operator-metrics/assert-servicemonitor.yaml
@@ -1,0 +1,18 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: opentelemetry-operator-metrics-monitor
+  namespace: ($otelnamespace)
+  labels:
+    app.kubernetes.io/name: opentelemetry-operator
+    app.kubernetes.io/part-of: opentelemetry-operator
+    control-plane: controller-manager
+spec:
+  endpoints:
+  - interval: 30s
+    path: /metrics
+    scheme: https
+    scrapeTimeout: 10s
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: opentelemetry-operator
--- a/tests/e2e/operator-metrics/chainsaw-test.yaml
+++ b/tests/e2e/operator-metrics/chainsaw-test.yaml
@@ -4,8 +4,9 @@ metadata:
   name: operator-metrics
 spec:
   steps:
-    - name: Assert operator metrics.
+    - name: Enable ServiceMonitor and assert operator metrics
       try:
+        # Get the operator namespace
         - command:
             entrypoint: kubectl
             args:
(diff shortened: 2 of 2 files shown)


internal/operator-metrics/metrics.go
```diff
@@ -166,7 +166,7 @@ func (om operatorMetrics) createOperatorMetricsServiceMonitor(ctx context.Contex
 					BearerTokenFile: bearerTokenFile,
 					Interval:        "30s",
 					Path:            "/metrics",
-					Scheme:          ptr.To(monitoringv1.SchemeHTTPS),
+					Scheme:          ptr.To(monitoringv1.Scheme("https")),
```
swiatekm, line 169: This is kind of nasty. The validation in the CRD version we depend on actually accepts the uppercase version as well, but older versions do not. Technically, we're misusing the type here, but I don't know if there's another way. I'm surprised this doesn't work as-is considering monitoringv1.Scheme serializes to the lowercase version.

Relatedly, I'm surprised we don't have integration tests for this. Or do we, and we always use newer prometheus-operator CRDs?

IshwarKanse: @swiatekm   The issue is that CRD validation happens before serialization. When we pass monitoringv1.SchemeHTTPS (which is the string "HTTPS"), the Kubernetes API server validates the value against the CRD's enum constraint before the object is created.

  The prometheus-operator library defines:
  const SchemeHTTPS URIScheme = "HTTPS"

  With validation annotation:
  // +kubebuilder:validation:Enum=http;https;HTTP;HTTPS

  Newer CRD versions accept both cases, but older versions only accept lowercase http and https. Since we need to support all versions, using the lowercase string directly seems to be the pragmatic solution.
  
  For the test, I have updated the operator-metrics test to enable the creation of controller manager metrics service monitor and assert it.

swiatekm: We're defining the ServiceMonitor as a Go struct, but it gets serialized to JSON before it's submitted to the API Server. And the Scheme type has a `String()` method that returns it in lowercase. So my assumption was that it would get serialized to JSON as lowercase too, but this is clearly not the case, and I'm not sure why.

It's good that you added that test, but even if we had it, it wouldn't have caught this problem, because for e2e tests we install prometheus-operator CRDs straight from their main branch. We should probably use the ones from the version we import, but that still wouldn't have caught the problem, as it only appears on older versions. I don't know that we can do anything about it, it's just nasty.

IshwarKanse: Agreed, this is difficult to catch in upstream, but we were able to catch this issue using the OpenShift-specific test case at tests/e2e-openshift/monitoring, which runs against actual OpenShift clusters with their shipped CRD versions. Although there is some gap between when changes are merged upstream and when our OpenShift CI jobs pick up those changes and run the tests, it still provides a safety net for these platform-specific edge cases.

Issue #4091, opened 2025-06-10
Service Monitor/Pod Monitor in TargetAllocator doesnt pick up updates to secrets


rashmichandrashekar:
### Component(s)

TargetAllocator

## Description
When the service/pod monitors use basic auth and read the keys from secrets, it is able to get them using the new https support. However, when the secret gets updated, the pod/service monitors do not pick them up.
Another related issue is that is the pod/service monitor is created before the secret object is created, the pod/svc monitors don't pick them up and makes it dependent on the order of resource creation.

## Steps to Reproduce
Create pod/service monitor using basic auth reading the username and password from secret. Update the username/password in the secret. But the pod and service monitors don't pick them up.

Create secret being referenced in the service monitor after the service monitor is created. The secret doesn't get picked up and the pod and service monitors do not work as expected.

## Expected Result
When the secret is updated the new values should be updated in the [objStore](https://github.com/prometheus-operator/prometheus-operator/blob/c4ebc762d0d2263541c67ebfe1ba7f2b419ed547/pkg/assets/store.go#L68) which the pod/service monitors [use](https://github.com/open-telemetry/opentelemetry-operator/blob/df01cc85b2f802c2cdf55f556a01331285d07d7b/cmd/otel-allocator/internal/watcher/promOperator.go#L98).
The order of secret creation shouldn't affect pod/svc monitor evaluation, should get picked up whenever the secret is created.

## Actual Result
Updated secret values don't get picked up from pod/svc monitors.
If secret is created after pod/svc monitor creation, it gets skipped.

### Kubernetes Version

1.29.15

### Operator version

v0.127.0

### Collector version

v0.127.0

## Environment
OS: cbl-mariner - 2.0
Compiler(if manually compiled): 1.23.8

### Log output
Example service monitor for the second part of the issue.

```yaml
  endpoints:
  - basicAuth:
      password:
        key: password
        name: my-secret-with-basic-auth-creds
      username:
        key: username
        name: my-secret-with-basic-auth-creds

```

```shell
[...]

PR #4608, opened 2026-01-13
Bug fix - Service Monitor/Pod Monitor in TargetAllocator doesnt pick up updates to secrets

rashmichandrashekar:
**Description:** 
This PR fixes the issue where - When the service/pod monitors use basic auth and read the keys from secrets, it is able to get them using the new https support. However, when the secret gets updated, the pod/service monitors do not pick them up.
Another related issue is that is the pod/service monitor is created before the secret object is created, the pod/svc monitors don't pick them up and makes it dependent on the order of resource creation.

Also, updating the groupname to read the value from **promMonitoring.GroupName** instead of hardcoding it to **"monitoring.coreos.com"** since prometheus-operator allows for this [customization ](https://github.com/prometheus-operator/prometheus-operator/blob/7bc5a41a76b062293ed14044e70e1a11c75b508e/pkg/apis/monitoring/register.go#L17C1-L25C2)and we are dependent on it.

**Link to tracking Issue(s):** [Issue 4091](https://github.com/open-telemetry/opentelemetry-operator/issues/4091)

- Resolves: 4091

**Testing:** Unit test is added and have done e2e testing to evaluate functionality

### DIFF
--- a/apis/v1beta1/targetallocator_rbac.go
+++ b/apis/v1beta1/targetallocator_rbac.go
@@ -13,7 +13,6 @@ import (
 )
 
 // targetAllocatorCRPolicyRules are the policy rules required for the CR functionality.
-// These rules are expected to be granted via a ClusterRole.
 var targetAllocatorCRPolicyRules = []*rbacv1.PolicyRule{
 	{
 		APIGroups: []string{"monitoring.coreos.com"},
--- a/internal/webhook/collector_webhook_test.go
+++ b/internal/webhook/collector_webhook_test.go
@@ -726,7 +726,6 @@ func TestOTELColValidatingWebhook(t *testing.T) {
 				"missing the following rules for system:serviceaccount:test-ns:adm-warning-targetallocator - nodes: [get,list,watch]",
 				"missing the following rules for system:serviceaccount:test-ns:adm-warning-targetallocator - pods: [get,list,watch]",
 				"missing the following rules for system:serviceaccount:test-ns:adm-warning-targetallocator - configmaps: [get]",
-				"missing the following rules for system:serviceaccount:test-ns:adm-warning-targetallocator - secrets: [get,list,watch]",
 				"missing the following rules for system:serviceaccount:test-ns:adm-warning-targetallocator - discovery.k8s.io/endpointslices: [get,list,watch]",
 				"missing the following rules for system:serviceaccount:test-ns:adm-warning-targetallocator - nonResourceURL: /metrics: [get]",
 				"missing the following rules for system:serviceaccount:test-ns:adm-warning-targetallocator - nonResourceURL: /api: [get]",
--- a/internal/webhook/targetallocator_webhook_test.go
+++ b/internal/webhook/targetallocator_webhook_test.go
@@ -236,7 +236,6 @@ func TestTargetAllocatorValidatingWebhook(t *testing.T) {
 				"missing the following rules for system:serviceaccount:test-ns:test-ta-targetallocator - nodes: [get,list,watch]",
 				"missing the following rules for system:serviceaccount:test-ns:test-ta-targetallocator - pods: [get,list,watch]",
 				"missing the following rules for system:serviceaccount:test-ns:test-ta-targetallocator - configmaps: [get]",
-				"missing the following rules for system:serviceaccount:test-ns:test-ta-targetallocator - secrets: [get,list,watch]",
 				"missing the following rules for system:serviceaccount:test-ns:test-ta-targetallocator - discovery.k8s.io/endpointslices: [get,list,watch]",
 				"missing the following rules for system:serviceaccount:test-ns:test-ta-targetallocator - nonResourceURL: /metrics: [get]",
 				"missing the following rules for system:serviceaccount:test-ns:test-ta-targetallocator - nonResourceURL: /api: [get]",


rashmichandrashekar:
@swiatekm - Please take a look at this pr, this contains a fix as discussed here - https://github.com/open-telemetry/opentelemetry-operator/issues/4091#issuecomment-2966045165
thanks!

cmd/otel-allocator/internal/watcher/promOperator.go
```diff
@@ -375,29 +389,131 @@ func (w *PrometheusCRWatcher) Watch(upstreamEvents chan Event, upstreamErrors ch
...
-				default:
-				}
-			},
-		})
+		// Use a custom event handler for secrets since secret update requires asset store to be updated so that CRs can pick up updated secrets.
+		if name == string(v1.ResourceSecrets) {
+			w.logger.Info("Using custom event handler for secrets informer", "informer", name)
+			// only send an event notification if there isn't one already
+			resource.AddEventHandler(cache.ResourceEventHandlerFuncs{
+				// these functions only write to the notification channel if it's empty to avoid blocking
```
swiatekm, line 404: This custom handler does a lot. Does prometheus-operator handle this in a similar way?

cmd/otel-allocator/internal/watcher/promOperator.go
```diff
@@ -375,29 +389,131 @@ func (w *PrometheusCRWatcher) Watch(upstreamEvents chan Event, upstreamErrors ch
...
+
+					if err != nil {
+						w.logger.Error("unexpected store error when getting updated secret - ", "secret", secretName, "error", err)
+						return
+					}
+					w.logger.Info("Updating secret in store", "newObjName", newMeta.GetObjectMeta().GetName(), "newobjnamespace", newMeta.GetObjectMeta().GetNamespace())
+					if err := w.store.UpdateObject(newSecret); err != nil {
+						w.logger.Error("unexpected store error when updating secret  - ", "secret", newMeta.GetObjectMeta().GetName(), "error", err)
+					} else {
+						w.logger.Info(
```
swiatekm: These logs should be `Debug` imo.

cmd/otel-allocator/internal/watcher/promOperator_test.go
```diff
@@ -1217,6 +1217,113 @@ func TestNamespaceLabelUpdate(t *testing.T) {
...
+
+	// Update the secret data.
+	sec, err := w.k8sClient.CoreV1().Secrets(namespace).Get(context.Background(), "basic-auth", metav1.GetOptions{})
+	require.NoError(t, err)
+	sec.Data["username"] = []byte("newadmin")
+	sec.Data["password"] = []byte("newpassword")
+	_, err = w.k8sClient.CoreV1().Secrets(namespace).Update(context.Background(), sec, metav1.UpdateOptions{})
+	require.NoError(t, err)
+
+	// Update the asset store to simulate the secrets informer behavior.
```
swiatekm: Aren't we running that informer in this test? Why do we need to simulate the updates?

cmd/otel-allocator/internal/watcher/promOperator.go
rashmichandrashekar: yes, this logic is extracted from the prometheus-operator. It maintains an object store and there is a [secret informer event handler](https://github.com/prometheus-operator/prometheus-operator/blob/b76c97db18442ae3bc57f6e09acd999c166bc829/pkg/prometheus/server/operator.go#L566) which is missing in the promOperator. When an event is received it uses the [eventhandler](https://github.com/prometheus-operator/prometheus-operator/blob/b76c97db18442ae3bc57f6e09acd999c166bc829/pkg/operator/event_handler.go) and eventually uses the [enqueueForPrometheusNamespace ](https://github.com/prometheus-operator/prometheus-operator/blob/b76c97db18442ae3bc57f6e09acd999c166bc829/pkg/prometheus/server/operator.go#L640)which processes it. In this repo we need a way to keep the store up to date so that when pod and service monitors reference secrets the store is up to date.

cmd/otel-allocator/internal/watcher/promOperator_test.go
rashmichandrashekar: Updated test to use informer instead of updating the asset store

cmd/otel-allocator/internal/watcher/promOperator.go
rashmichandrashekar: Changed to Debug

cmd/otel-allocator/internal/watcher/promOperator.go
```diff
@@ -375,29 +389,131 @@ func (w *PrometheusCRWatcher) Watch(upstreamEvents chan Event, upstreamErrors ch
...
+			resource.AddEventHandler(cache.ResourceEventHandlerFuncs{
+				// these functions only write to the notification channel if it's empty to avoid blocking
+				// if scrape config updates are being rate-limited
+				AddFunc: func(obj interface{}) {
+					select {
+					case notifyEvents <- struct{}{}:
+					default:
+					}
+				},
+				UpdateFunc: func(oldObj, newObj interface{}) {
```
swiatekm: The update and delete handler are quite complex, it'd be better to move them to separate methods. These would handle the update and return a boolean indicating whether anything changed and we need to reload the config.

rashmichandrashekar:
> The logic of this PR makes sense to me, I'd like it to be refactored a bit more. I left a comment explaining what I'd like to see.
> 
> Independently, is it possible for us to add a e2e test showing that this mechanism works correctly?

Thanks @swiatekm. I have refactored to address the comments and added an e2e test as well, pls take a look

cmd/otel-allocator/internal/watcher/promOperator_test.go
```diff
@@ -1228,6 +1229,166 @@ func TestNamespaceLabelUpdate(t *testing.T) {
...
+			Name:            "basic-auth",
+			Namespace:       namespace,
+			ResourceVersion: "2",
+		},
+	}
+	err = mdClient.Tracker().Update(secretGVR, updatedSecretMeta, namespace)
+	require.NoError(t, err)
+
+	// Give time for the tracker update to propagate to the informer
+	time.Sleep(100 * time.Millisecond)
```
swiatekm: Do we need this sleep? Doesn't the assertion below cover it?

Alternatively, if we want to keep the sleep, I'd prefer using `synctest`.

rashmichandrashekar: Updated the tests to remove sleep. PTAL, thanks!

swiatekm:
Can we leave the RBAC changes out of this for now? I'd prefer not to increase the complexity of this PR, and I'm not convinced we should warn users about needing secret access even if they're not using auth on their Prometheus endpoints. It's fine for that to go into a follow-up PR.

Issue #1234, opened 2022-11-09
Does the "configmaps/status" resource serve any purpose?


Allex1:
According to https://github.com/kubernetes-sigs/kubebuilder/issues/1473 and https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#configmap-v1-core the "configmaps/status" resource does not exist 😕 but it's still present in our roles in https://github.com/open-telemetry/opentelemetry-operator/blob/main/config/rbac/leader_election_role.yaml#L22 and https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/templates/role.yaml#L25
I've tested the operator chart in my env without that role and could not find any apparent issues.
I'd be happy to remove it from both repos if I'm not missing something.

jaronoff97:
This is a good question, and i'm not sure i know the answer @open-telemetry/operator-approvers any ideas?

swiatekm-sumo:
I'm guessing we just inherited it from kubebuilder? If it doesn't exist, we should delete it.

apt-x4869:
Hi @swiatekm,

I'm new to contributing to open-source projects and am eager to get involved. Could you please assign this issue to me? 
I will test if removing it won't cause any issue.
I look forward to contributing and learning.

Thank you!

swiatekm:
Resolved in https://github.com/open-telemetry/opentelemetry-operator/pull/4609, closing.

PR #4609, opened 2026-01-13
[Chore] Removed ``configmaps/status`` resource

tushar1977:
**Description:** 
Removed the depreciated  ``configmaps/status`` resource.

#1234

### DIFF
--- a/internal/instrumentation/apachehttpd.go
+++ b/internal/instrumentation/apachehttpd.go
@@ -48,12 +48,8 @@ const (
 	6) Inject mounting of volumes / files into appropriate directories in application container
 */
 
-func injectApacheHttpdagent(_ logr.Logger, apacheSpec v1alpha1.ApacheHttpd, pod corev1.Pod, useLabelsForResourceAttributes bool, index int, otlpEndpoint string, resourceMap map[string]string, instSpec v1alpha1.InstrumentationSpec) corev1.Pod {
-
+func injectApacheHttpdagent(_ logr.Logger, apacheSpec v1alpha1.ApacheHttpd, pod corev1.Pod, useLabelsForResourceAttributes bool, container *corev1.Container, otlpEndpoint string, resourceMap map[string]string, instSpec v1alpha1.InstrumentationSpec) corev1.Pod {
 	volume := instrVolume(apacheSpec.VolumeClaimTemplate, apacheAgentVolume, apacheSpec.VolumeSizeLimit)
-
-	// caller checks if there is at least one container
-	container := &pod.Spec.Containers[index]
 
 	// inject env vars
 	container.Env = appendIfNotSet(container.Env, apacheSpec.Env...)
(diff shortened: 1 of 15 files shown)


config/rbac/leader_election_role.yaml
```diff
@@ -16,14 +16,6 @@ rules:
   - update
   - patch
   - delete
-- apiGroups:
```
iblancasa, line 19: With this change, you will need to regenerate the manifests.

swiatekm: You can do this by running `make update`.

tushar1977: sure, i'll do it

tushar1977: I also have a suggestion regarding Makefiles.

In opentelemetry-go-compile-instrumentation, running make displays the available Make targets. Could we implement the same behavior in opentelemetry-operator as well?

I understand that the OpenTelemetry Go compiler bootstraps Go projects and enables tracing, metrics, and other instrumentation. Adding a similar make help output in the operator repo would  improve the new developer experience by making common workflows easier to discover.

swiatekm: Sure thing, I'd absolutely accept a PR adding that. I'd be concerned about it becoming outdated unless we have a check for it, but it'd be an improvement regardless.

For reference, a lot of the targets relevant to new contributors are documented in [CONTRIBUTING.md](https://github.com/open-telemetry/opentelemetry-operator/blob/main/CONTRIBUTING.md).

tushar1977: > Sure thing, I'd absolutely accept a PR adding that. I'd be concerned about it becoming outdated unless we have a check for it, but it'd be an improvement regardless.

so should i create a new issue and raise PR there? 

> For reference, a lot of the targets relevant to new contributors are documented in [CONTRIBUTING.md](https://github.com/open-telemetry/opentelemetry-operator/blob/main/CONTRIBUTING.md).

thanks, i'll check it

bundle/openshift/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -548,7 +548,7 @@ spec:
                   valueFrom:
                     fieldRef:
                       fieldPath: metadata.namespace
-                image: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator:0.142.0
+                image: controller
```
pavolloffay: This should not change. It should point to a real image.

config/rbac/leader_election_role.yaml
tushar1977: image

@swiatekm here is a overall layout of the all commands. Here i am using ``awk`` to gather the comments and commands from makefile using a specific format
```bash
##@ PR   <----- SECTION
# Verify generated code, manifests, bundles, and API docs are up to date   <---- Description
.PHONY: ensure-update-is-noop    <---- Command
ensure-update-is-noop: VERSION=$(OPERATOR_VERSION)

```
Just need to add description in few commands

swiatekm: Looks good to me. I don't think we need an issue, but please create a new PR for this change.

tushar1977: sure

PR #4610, opened 2026-01-13
refactor(instrumentation): pass container pointer instead of index

Joibel:

This is a refactor, changing instrumentation to pass container pointers instead of indexes.

For https://github.com/open-telemetry/opentelemetry-operator/issues/3308 we'll need to get the containers from the initContainer and container list for most of the injection functions.

I could do this by passing a tuple of index and a boolean fiagging which container list to use, but this feels cleaner. It's quite invasive, so I've split it into a separate PR from actually doing initContainers.

I've added an `ensureContainer()` function which implements the repeated code:
```go
if len(insts.Java.Containers) == 0 {
    insts.Java.Containers = []string{pod.Spec.Containers[0].Name}
}
```

and added `containersToInstrument()` to collect the list of containers in each case. This function will learn about initContainers in a future patch. Note: containers and initContainers [must have unique names (across both lists)](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#detailed-behavior) which allows us to just search both lists when finding containers.

Note: I'm in place modifying the passed in pod or container in a number of places rather than always returning a copied pod. If this is unacceptable would the alternative of DeepCopying the pod at the top of inject() be a suitable alternative. Because the pod is just passed by value, a lot of the underlying slices and maps will be sharing contents in the old form of the code anyway, so this didn't feel like it was a bad change to have made.

- Part of #3308, but does not close it.

Existing tests continue to pass - it is a refactor only.
The tests themselves look more messy, but the code looks cleaner to me.

None, refactor only

### DIFF
--- a/internal/instrumentation/helper_test.go
+++ b/internal/instrumentation/helper_test.go
@@ -315,3 +315,121 @@ func TestInstrWithoutContainers(t *testing.T) {
 		})
 	}
 }
+
+func TestEnsureContainer(t *testing.T) {
+	tests := []struct {
+		name               string
+		inst               instrumentationWithContainers
+		pod                corev1.Pod
+		expectedContainers []string
+	}{
+		{
+			name: "empty containers list",
+			inst: instrumentationWithContainers{Containers: []string{}},
+			pod: corev1.Pod{
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{{Name: "app"}},
+				},
+			},
+			expectedContainers: []string{"app"},
+		},
+		{
+			name: "already has containers",
+			inst: instrumentationWithContainers{Containers: []string{"my-container"}},
+			pod: corev1.Pod{
+				Spec: corev1.PodSpec{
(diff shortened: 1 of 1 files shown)


internal/instrumentation/sdk.go
```diff
@@ -168,60 +149,65 @@ func (i *sdkInjector) inject(ctx context.Context, insts languageInstrumentations
...
+
+	for _, containerName := range inst.Containers {
+		for i := range pod.Spec.Containers {
+			if pod.Spec.Containers[i].Name == containerName {
+				containers = append(containers, &pod.Spec.Containers[i])
+			}
+		}
+	}
+	return containers
+}
```
swiatekm, line 209: Could you add docstrings and some basic unit tests for these functions?

Joibel: Yep, sounds good to me, I'll add some

Joibel: Done

Issue #4623, opened 2026-01-16
PodDisruptionBudget's MatchLabels copied over from openTelemetryCollector


yelshall:
### Component(s)

collector

## Description
The `matchLabels` of the PDB of the collector is just copied over from the collector custom resource. This breaks the PDB temporarily on every rollout as we have deploy labels that our tooling adds to the custom resource which changes with every deploy tag.

[Code can be found here](https://github.com/open-telemetry/opentelemetry-operator/blob/3f990f3ba2fd65745f520fdb8a1ec9a7a6ccc162/internal/manifests/collector/poddisruptionbudget.go#L34)

As for the `targetAllocator`, the `matchLabels` is limited to a selection of selector labels as can be seen [here](https://github.com/open-telemetry/opentelemetry-operator/blob/3f990f3ba2fd65745f520fdb8a1ec9a7a6ccc162/internal/manifests/manifestutils/labels.go#L79).

Ideally, we'd have similar behavior to targetAllocator's PDB or add more customization of the selector labels and pod labels.

## Steps to Reproduce
This is the default behavior when creating the collector's PDB.

Just deploy a collector CR and the PDB will show the behavior in the description.

## Expected Result
N/A

## Actual Result
N/A

### Kubernetes Version

1.33

### Operator version

0.141.0

### Collector version

N/A

### Environment information

N/A

### Log output

```shell
N/A
```

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

PR #4626, opened 2026-01-17
fix: use selector labels for PodDisruptionBudget

Arpit529Srivastava:
**Description:** 
fix the collector PDB to use stable selector labels so it doesn’t become invalid during rollouts.

- Resolves: #4623 

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,1 @@
+CONTRIBUTING.md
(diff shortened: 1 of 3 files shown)


Arpit529Srivastava:
/cc @swiatekm. PTAL
thanks.

Arpit529Srivastava:
> @Arpit529Srivastava could you add a unit test for this behavior? Looks like we don't have one right now.

@swiatekm made the requested changes.
thanks

Issue #3729, opened 2025-02-20
Failed pod creating nginx instrumentation


Faig92:
### Component(s)

auto-instrumentation

### What happened?

While trying use nginx autoinstrumentation this error occurs:

Error creating: Pod app is invalid: spec.initContainers[1].resizePolicy: Invalid value: []core.ContainerResizePolicy{core.ContainerResizePolicy{ResourceName:"cpu", RestartPolicy:"NotRequired"}, core.ContainerResizePolicy{ResourceName:"memory", RestartPolicy:"NotRequired"}}: must not be set for init containers
```
apiVersion: apps/v1
kind: Deployment
spec:
  replicas: 1
  selector:
    matchLabels:
      name: app
  template:
    metadata:
      name: app
      labels:
        123: xxx
      annotations:
        vault.hashicorp.com/agent-cache-enable: 'false'
        vault.hashicorp.com/agent-image: myimage
        vault.hashicorp.com/agent-init-first: 'true'
        vault.hashicorp.com/agent-inject: 'true'
        vault.hashicorp.com/agent-pre-populate-only: 'true'
        instrumentation.opentelemetry.io/inject-nginx: 'instrumentation'
        instrumentation.opentelemetry.io/container-names: 'app'
   spec:
      containers:
        - name: app
          image: myimage
          resources:
            limits:
              cpu: xxx
              ephemeral-storage: xxx
              memory: xxx
            requests:
              cpu: xxx
              ephemeral-storage: xxx
              memory: xxx  
          livenessProbe:
          readinessProbe:
          imagePullPolicy: Always
```

### Kubernetes Version

1.27

### Operator version

0.117.0

### Collector version

0.119.0

## Environment
OS: (e.g., "Ubuntu 20.04")
Compiler(if manually compiled): (e.g., "go 14.2")

### Log output

```shell

```

Faig92:
I've tried replace vault.hashicorp.com/agent-init-first: 'true' annotation but still recieving error while pod creating

swiatekm:
This is a legitimate bug. We [clone](https://github.com/open-telemetry/opentelemetry-operator/blob/a30b66a75ccb9e81d95cf3ff93c5df5f453e28f4/internal/instrumentation/nginx.go#L88) the nginx container to create our instrumentation init container, which isn't right and can lead to problems like this. We should instead create the container spec from scratch, like we do for other instrumentations.

PR #4627, opened 2026-01-18
Fix invalid fields copied to NGINX init container

ozzywalsh:
Fixes a bug in the NGINX and Apache instrumentations where the init containers were created by doing a `DeepCopy` on the main container. This caused init-container-incompatible fields (eg. `resizePolicy`) to be unintentionally copied, leading to invalid pod specs.

This change creates the instrumentation init containers from scratch, explicitly setting only the required fields. This aligns the implementation with other instrumentations (like nodejs) and prevents incompatible fields from being propagated.

Link to tracking Issue(s):
Resolves: #3729

Testing:
Updated unit test cases in both nginx_test.go and apachehttpd_test.go to ensure init-container-incompatible fields are not copied to the init containers.

Documentation:
No documentation changes were required.

### DIFF
--- a/.chloggen/fix_nginx-init-container-from-scratch.yaml
+++ b/.chloggen/fix_nginx-init-container-from-scratch.yaml
@@ -5,12 +5,13 @@ change_type: bug_fix
 component: auto-instrumentation
 
 # A brief description of the change.
-note: "Fix NGINX instrumentation init container creation to avoid copying init-container-incompatible fields."
+note: "Fix NGINX and Apache instrumentation init container creation to avoid copying init-container-incompatible fields."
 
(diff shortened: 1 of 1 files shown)


swiatekm (approved):
Thank you for the fix! I'd love to get the same problem fixed in the apache instrumentation - if you're willing to take that up, I'm ok both adding it to this PR and submitting a separate one.

ozzywalsh:
> Thank you for the fix! I'd love to get the same problem fixed in the apache instrumentation - if you're willing to take that up, I'm ok both adding it to this PR and submitting a separate one.

Ok, I added the same fix & test for the apache instrumentation to this PR also.

PR #4635, opened 2026-01-19
feat: init container instrumentation

Joibel:
**Description:** 

Add support for instrumenting init containers for Python, Java, Node.js, .NET and SDK-only.

It would be presumably possible to do this for go once official long-running sidecar support allowed the go sidecar to run alongside the init container, but I considered that out of scope.

Apache and Nginx are unlikely to run in init containers, so supporting them felt strange, so this is not possible here to avoid making this more complicated.

If no annotation is found the behaviour of instrumenting the first regular container is found.
The injected init containers are still injected last unless someone wants instrumented language init containers, when they are injected before the first instrumented init container.

Disclaimer: This PR was assisted by Claude AI, in particular for tests.

- Resolves: #3308

**Testing:** 

Added unit tests, and a e2e single test per language (+SDK only) to prove at least some telemetry is emitted in all cases.

**Documentation:**

The README.md is updated to cover the support of init containers, including an example.
The support didn't require any change in the annotations - [init containers and regular containers contain no overlapping names](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#detailed-behavior).

Please guide me if I should also update another repository with more documentation once this is merged.

### DIFF
--- a/internal/instrumentation/javaagent_test.go
+++ b/internal/instrumentation/javaagent_test.go
@@ -5,6 +5,7 @@ package instrumentation
 
 import (
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -410,8 +411,7 @@ func TestInjectJavaagent(t *testing.T) {
 					continue
 				}
 				// Skip extension init containers
-				if len(pod.Spec.InitContainers[i].Name) > len(initContainerName+"-extension-") &&
-					pod.Spec.InitContainers[i].Name[:len(initContainerName+"-extension-")] == initContainerName+"-extension-" {
+				if strings.HasPrefix(pod.Spec.InitContainers[i].Name, initContainerName+"-extension-") {
 					continue
 				}
 				injector.injectDefaultJavaEnvVars(&pod.Spec.InitContainers[i], test.Java)
--- a/internal/instrumentation/sdk.go
+++ b/internal/instrumentation/sdk.go
@@ -87,17 +87,13 @@ func (i *sdkInjector) injectJava(ctx context.Context, inst instrumentationWithCo
 		for _, container := range containers {
 			if err := injectJavaagentToContainer(otelinst.Spec.Java, container); err != nil {
 				i.logger.Info("Skipping javaagent injection", "reason", err.Error(), "container", container.Name)
-				return pod
+			} else {
+				i.injectCommonEnvVar(otelinst, container)
+				i.injectDefaultJavaEnvVars(container, otelinst.Spec.Java)
+				pod = i.injectCommonSDKConfig(ctx, otelinst, ns, pod, container, container)
+				pod = i.setInitContainerSecurityContext(pod, container.SecurityContext, javaInitContainerName)
 			}
 		}
-
-		for _, container := range containers {
-			i.injectCommonEnvVar(otelinst, container)
-			i.injectDefaultJavaEnvVars(container, otelinst.Spec.Java)
-			pod = i.injectCommonSDKConfig(ctx, otelinst, ns, pod, container, container)
-			pod = i.setInitContainerSecurityContext(pod, container.SecurityContext, javaInitContainerName)
-		}
-
 		pod = injectJavaagentToPod(otelinst.Spec.Java, pod, containers[0].Name, otelinst.Spec)
 	}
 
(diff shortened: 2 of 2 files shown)


tests/e2e-instrumentation/instrumentation-dotnet-initcontainer/00-install-collector.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: opentelemetry.io/v1beta1
```
swiatekm, line 1: As a general comment about the e2e tests you've added, I don't think we specifically need to check that we receive data from these. There isn't much difference between an init container and a normal container in terms of application instrumentation, so this is somewhat overkill. These tests tend to be a bit flaky as well, so I'd rather not add more of them without good reason.

I think it'd be fine if you just checked that the instrumentation was injected correctly and that the Pod started successfully.

internal/instrumentation/javaagent_test.go
```diff
@@ -330,15 +384,40 @@ func TestInjectJavaagent(t *testing.T) {
...
 			for i := range pod.Spec.Containers {
 				injector.injectDefaultJavaEnvVars(&pod.Spec.Containers[i], test.Java)
 			}
+			for i := range pod.Spec.InitContainers {
+				// Skip the instrumentation init container we added
+				if pod.Spec.InitContainers[i].Name == javaInitContainerName {
+					continue
+				}
+				// Skip extension init containers
+				if len(pod.Spec.InitContainers[i].Name) > len(initContainerName+"-extension-") &&
```
iblancasa: Maybe:
```go
if strings.HasPrefix(pod.Spec.InitContainers[i].Name, initContainerName+"-extension-") {
      continue
  }
```

tests/e2e-instrumentation/instrumentation-dotnet-initcontainer/00-install-collector.yaml
Joibel: I did have concerns that if the init container ordering was incorrect it would pass that kind of test - also if I had implemented a golang version of this this kind of CI test would pass but in reality the initContainer instrumentation wouldn't be functional, which is why I had implemented this as I did.

I've removed the telemetry generation and checks now

swiatekm: Hm, that's a good point. @open-telemetry/operator-maintainers WDYT? Should we include the extended tests? We can merge this PR as is and decide on this independently. It is true that a change might break this by accident by rearranging the init containers, and we wouldn't have a good way of knowing.

internal/instrumentation/helper_test.go
```diff
@@ -433,3 +469,213 @@ func TestContainersToInstrument(t *testing.T) {
...
+			name:      "empty pod",
+			pod:       corev1.Pod{},
+			target:    "any",
+			expectNil: true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			result := findContainerByName(test.target, &test.pod)
```
iblancasa, line 579: If the target does not exist... do we provide any feedback to the user?

internal/instrumentation/sdk.go
```diff
@@ -49,140 +50,226 @@ func (i *sdkInjector) inject(ctx context.Context, insts languageInstrumentations
...
+	if len(containers) > 0 {
+		for _, container := range containers {
+			if err := injectJavaagentToContainer(otelinst.Spec.Java, container); err != nil {
 				i.logger.Info("Skipping javaagent injection", "reason", err.Error(), "container", container.Name)
-			} else {
-				i.injectCommonEnvVar(otelinst, container)
-				i.injectDefaultJavaEnvVars(container, otelinst.Spec.Java)
-				pod = i.injectCommonSDKConfig(ctx, otelinst, ns, pod, container, container)
-				pod = i.setInitContainerSecurityContext(pod, container.SecurityContext, javaInitContainerName)
+				return pod
```
iblancasa: If there are 3 containers, validation fails for the first one and containers 2 and 3 can be injected, this will skip them.

internal/instrumentation/sdk.go
```diff
@@ -49,140 +50,226 @@ func (i *sdkInjector) inject(ctx context.Context, insts languageInstrumentations
...
 			}
 		}
+
+		for _, container := range containers {
```
iblancasa: Why do we need 2 loops?

internal/instrumentation/sdk.go
Joibel: Yes, sorry. This is a reversion from the current code - this went through quite a few iterations to get it to work right, and it ended up more dissimilar to the previous code than necessary.

I should have caught this and the double loop in my own code review.

internal/instrumentation/sdk.go
Joibel: We don't. Removed - same reason as before.

internal/instrumentation/javaagent_test.go
Joibel: Done, yeah, that was overcomplex.

internal/instrumentation/helper_test.go
Joibel: findContainerByName returns nil when we can't find anything - I'm not sure I understand what else you'd like?

At a top level, the behaviour of `containersToInstrument` and feedback from it hasn't changed in this commit - we don't tell the user if they've typed rubbish in.

iblancasa: Mm that's true. Something we should fix :/ it's fine

iblancasa (approved):

tests/e2e-instrumentation/instrumentation-dotnet-initcontainer/00-install-collector.yaml
jaronoff97: yeah i think init container instrumentation is honestly going to just be _very_ tricky. like super tricky. I'm not sure if there's a great way for us to test it.

jaronoff97 (approved):
the actual logic here isn't too different, but reviewing this type of change further reveals my frustrations with the pre-existing logic, thank you very much for working through those.

jaronoff97:
@Joibel I appreciate the offer! We've been discussing in our SIG meetings what the future of this code path looks like and already have a few folks working through implementations. We've been waiting for the injector SIG to get to a release phase as well as for declarative config to be stable (both have recently happened) so hopefully the implementation work will begin shortly. Thank you!

PR #4640, opened 2026-01-20
[target-allocator] Add feature flag for labeled targets_remaining metric with OTel conventions

### LINKED ISSUE #4637: Add namespace/job_name labels to opentelemetry_allocator_targets_remaining metric
### Component(s)

target allocator

### Problem

  The `opentelemetry_allocator_targets_remaining` metric currently reports a scalar count of unallocated targets, making it impossible to distinguish
  between:

  1. **Intentionally filtered targets** (e.g., namespaces excluded via `allowNamespaces`)
  2. **Orphaned targets due to bugs** (e.g., stale collector UIDs, version skew during rollouts)

  This makes alerting extremely difficult - we can't set meaningful thresholds without either:
  - Creating false positives (alerting on intentional filters)
  - Missing real issues (high threshold that ignores single critical target losses)

  ### Use Case

  In our production environment:
  - We filter `istio-system` namespace targets (~1600 targets) to control costs
  - This baseline makes `targets_remaining` unusable for detecting orphaned targets
  - A recent incident (target assigned to deleted collector pod) went undetected for 14 hours

### Proposed Solution

  Add labels to `opentelemetry_allocator_targets_remaining` to break down unallocated targets:

  ```go
  targetsRemaining = promauto.NewGaugeVec(
      prometheus.GaugeOpts{
          Name: "opentelemetry_allocator_targets_remaining",
          Help: "Number of targets that could not be allocated to any collector",
      },
      []string{"job_name", "namespace"},  // Add these labels
  )
```
  This would allow alerts like:
```
  # Alert on orphaned targets, excluding known filtered namespaces
  opentelemetry_allocator_targets_remaining{namespace!="istio-system"} > 10
```
  Alternative

  If per-target labeling is too expensive (high cardinality), consider:
  - Add just reason label: {reason="namespace_filtered"} vs {reason="no_collectors"}
  - Or expose separate metrics: ..._targets_filtered vs ..._targets_orphaned

_No response_

### Additional context

_No response_

### Tip

### PR DESCRIPTION
**Pull request by hkf57**
## Description

Adds `k8s.namespace.name` and `job.name` labels to the `opentelemetry_allocator_targets_remaining` metric (when enabled via feature gate) to enable distinguishing between intentionally filtered targets and orphaned targets due to bugs.

**Breaking Change Mitigation:** This change is behind feature gates (alpha) to provide a migration path for existing users.

## Motivation and Context

Currently, `opentelemetry_allocator_targets_remaining` reports a scalar count, making it impossible to create meaningful alerts. In production environments with namespace filtering (e.g., excluding `istio-system` to control costs), the baseline can be ~1600 targets. This masks genuine issues like orphaned targets from deleted collectors or version skew scenarios.

### Feature Gates (Alpha)

**Operator gate:** `operator.targetallocator.labeledmetrics`
- Registered in `pkg/featuregate/featuregate.go`
- When enabled, the operator sets `--feature-gates=target-allocator.labeled-metrics` on the TA container args

**Target Allocator gate:** `target-allocator.labeled-metrics`
- Registered in `cmd/otel-allocator/internal/featuregate/featuregate.go`
- TA parses `--feature-gates` via CLI flags
- Controls labelled metric emission at runtime

This two-gate approach follows the reviewer's guidance: the operator gate propagates to the TA via `--feature-gates`, avoiding any config struct changes that would be a breaking change to remove when the gate graduates.

### Label Names (OTel Semantic Conventions)
- `job.name`
- `k8s.namespace.name`

### Implementation
Modified `SetTargets()` in `allocator.go` to:
1. Check feature gate state
2. When enabled: Group remaining targets by job and namespace, emit labelled metrics
3. When disabled: Emit single unlabelled gauge (original behaviour)
4. Extract namespace from `__meta_kubernetes_namespace` label
5. Use "unknown" for targets without namespace metadata

## Usage

Enable via the operator's feature gates:
```bash
--feature-gates=operator.targetallocator.labeledmetrics
```

The operator propagates this to the target allocator container automatically.

## After this change (when feature enabled)

```promql
opentelemetry_allocator_targets_remaining{k8s_namespace_name!="istio-system"} > 10
opentelemetry_allocator_targets_remaining{job_name="kube-state-metrics"} > 0
```

## Testing

- [x] Code compiles successfully
- [x] Mock interfaces updated
- [x] Feature gates registered (operator + TA)
- [x] Container args propagation test
- [x] User-specified feature gates not overridden
- [ ] Unit tests for labelled/unlabelled behaviour
- [ ] E2E tests with feature flag enabled/disabled

Fixes #4637

### DIFF
--- a/cmd/otel-allocator/internal/allocation/allocator.go
+++ b/cmd/otel-allocator/internal/allocation/allocator.go
@@ -94,6 +94,7 @@ type allocator struct {
 	log logr.Logger
 
 	filter                Filter
+	useLabeledMetrics     bool
 	targetsPerCollector   metric.Int64Gauge
 	collectorsAllocatable metric.Int64Gauge
 	timeToAssign          metric.Float64Histogram
@@ -111,6 +112,11 @@ func (a *allocator) SetFallbackStrategy(strategy Strategy) {
 	a.strategy.SetFallbackStrategy(strategy)
 }
 
+// SetLabeledMetrics sets whether to use labeled metrics.
+func (a *allocator) SetLabeledMetrics(enabled bool) {
+	a.useLabeledMetrics = enabled
+}
+
 // SetTargets accepts a list of targets that will be used to make
 // load balancing decisions. This method should be called when there are
 // new targets discovered or existing targets are shutdown.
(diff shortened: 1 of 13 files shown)


swiatekm:
I'd prefer if we made this a feature flag in the target allocator as well.

- Add a featuregates.go in cmd/otel-allocator similar to the one in pkg/featuregates.go
- Add the target allocator feature gates to command line flags in main.go
- Add your new feature gate and use it for any checks
- The operator feature gate stays, and causes the operator to set the feature gate on the target allocator

Not only will this remove the need to set a new attribute on the structs in target allocator, you also won't need to add a new config option. Adding that option and then removing it later will be an unnecessary breaking change.

cmd/otel-allocator/internal/allocation/allocator.go
```diff
@@ -124,7 +130,35 @@ func (a *allocator) SetTargets(targets []*target.Item) {
 		targets = a.filter.Apply(targets)
 	}
 
-	a.targetsRemaining.Record(context.Background(), int64(len(targets)))
+	// Record remaining targets, with labels if feature flag is enabled
+	if a.useLabeledMetrics {
```
swiatekm, line 134: I think you can just use the feature gate directly here and avoid passing this option down from `main.go`.

PR #4644, opened 2026-01-22
feat: implement pod mutation observability metrics

### LINKED ISSUE #3702: Improve Operator Pod Mutation Observability
### Component(s)

auto-instrumentation

I'm currently working on rolling out the OpenTelemetry Operator across all of the kubernetes (OpenShift) clusters in our environment.  The capability of auto-instrumenting our application workloads will become crucial in our ability to support our systems.  If something happens to the operator that results in pods NOT getting auto-instrumented, we'd potentially be "flying blind".

I'd like the ability to have finer insights into the counts of auto-instrumentation attempts and failures to build the proper alerting (SLOs).

Instrument the pod mutator to create/increment metrics that indicate that a pod contains the instrumentation annotation and is subject to receive auto-instrumentation.  Some initial ideas on the types of scenarios/metrics to expose:

- pod contained instrumentation/sidecar annotation (may or may not be valid config) -> increment some counter saying "the podmutator will attempt to process"
- pod contained invalid "inject" type -> pod mutation didn't happen, increment a counter to reflect this scenario
- pod contained invalid instrumentation or sidecar reference in the annotation value -> pod mutation didn't happen, increment a counter to reflect this scenario
- pod contained valid instrumentation or sidecar annotation/reference, but an unexpected error occurred -> pod mutation failed, increment a counter to reflect

I know some of these scenarios may be available in container or kubernetes logs, but for managing a fleet of operator across multiple clusters is much easier to do with aggregate metrics to feed to our alerting infrastructure.

I'm currently leveraging the metrics provided by the kubernetes api server admission controller to see the counts of webhook invocations sent to the mpod.kb.io and it does provide some insights, but not all pod creations will be eligible for OTel instrumentation (i.e. they may or may not have the instrumentation.opentelemetry.io annotations.

### Additional context

_No response_

Arpit529Srivastava:

adds observability to the pod mutation webhook by introducing a new prometheus counter `opentelemetry_operator_pod_mutations_total`. this allows operators to track success, failure and skip rates for both sidecar injection and auto-instrumentation across the cluster. the metric includes detailed labels such as status, reason, type, language, and namespace, making it easier to distinguish between cases like configuration errors, disabled feature gates, or missing resources.

- Resolves: #3702

new unit test added at `internal/webhook/podmutation/metrics_test.go` to verify that the counter increments correctly with the expected label set.

integration checks confirm that both the sidecar and instrumentation mutators invoke the metrics recorder at all critical decision points.

existing unit tests were updated to ensure there are no regressions in the mutator logic.


internal/webhook/podmutation/metrics.go
```diff
@@ -0,0 +1,71 @@
...
+		return nil, err
+	}
+
+	return &PodMutationMetrics{
+		mutationsTotal: mutationsTotal,
+	}, nil
+}
+
+func (m *PodMutationMetrics) RecordSidecarMutation(ctx context.Context, status, reason, namespace string) {
+	if m == nil {
```
iblancasa: I think this can happen for tests, right? How about creating an interface and have a `noopMetrics` type for the tests?

internal/webhook/podmutation/metrics.go
```diff
@@ -0,0 +1,71 @@
...
+		return
+	}
+	attrs := []attribute.KeyValue{
+		attribute.String("mutation_type", "sidecar"),
+		attribute.String("status", status),
+		attribute.String("namespace", namespace),
```
iblancasa: I'm not sure about this... because of cardinality

internal/instrumentation/podmutator.go
```diff
@@ -231,92 +234,107 @@ func (pm *instPodMutator) Mutate(ctx context.Context, ns corev1.Namespace, pod c
 
 	if inst, err = pm.getInstrumentationInstance(ctx, ns, pod, annotationInjectJava); err != nil {
 		// we still allow the pod to be created, but we log a message to the operator's logs
+		pm.metrics.RecordInstrumentationMutation(ctx, "error", "lookup_failed", "java", ns.Name)
```
iblancasa: I think we should define some constants:
```go
  const (
      StatusSuccess  = "success"
      StatusSkipped  = "skipped"
      StatusRejected = "rejected"
      StatusError    = "error"

      ReasonAlreadyInstrumented = "already_instrumented"
      ReasonFeatureDisabled     = "feature_disabled"
      ReasonLookupFailed        = "lookup_failed"
  )
```

internal/webhook/podmutation/metrics.go
iblancasa: Can we use semantic conventions package for those attributes that are already there? Like `k8s.namespace.name`

.chloggen/issue-3702-pod-mutation-metrics.yaml
```diff
@@ -0,0 +1,16 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Add Prometheus metrics to the pod mutation webhook to track success, failures, and skips for both sidecar injection and auto-instrumentation."
+
+# One or more tracking issues related to the change
+issues: [3702]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
iblancasa: Can you explain here what metrics this PR is adding and other stuff?

internal/instrumentation/podmutator.go
```diff
@@ -195,7 +196,7 @@ func (langInsts *languageInstrumentations) setLanguageSpecificContainers(ns meta
 
 var _ podmutation.PodMutator = (*instPodMutator)(nil)
 
-func NewMutator(logger logr.Logger, client client.Client, recorder record.EventRecorder, cfg config.Config) podmutation.PodMutator {
+func NewMutator(logger logr.Logger, client client.Client, recorder record.EventRecorder, cfg config.Config, metrics *podmutation.PodMutationMetrics) podmutation.PodMutator {
```
iblancasa: Instead of repeating here for each language.. maybe we can do something more table-driven

main.go
```diff
@@ -358,20 +358,24 @@ func main() {
 	}
 
 	if cfg.EnableWebhooks {
-		var crdMetrics *otelv1beta1.Metrics
+		meterProvider, metricsErr := otelv1beta1.BootstrapMetrics()
+		if metricsErr != nil {
+			setupLog.Error(metricsErr, "Error bootstrapping metrics")
```
iblancasa, line 363: Since we continue, it can happen `meterProvider` is `nil`. And that cause a panic later if methods using `meterProvider`. Can you add checks to avoid that?

iblancasa (commented):
Maybe we should add a new option to enable/disable recording these metrics.
I also miss some E2E tests to ensure the feature works properly.

internal/webhook/podmutation/metrics.go
Arpit529Srivastava: i recommend/think keeping it since it’s critical for debugging and has bounded cardinality. with the addition of the `--enable-webhook-metrics` flag, this is now opt-in and minimizes risk. happy to discuss other options.

main.go
Arpit529Srivastava: fixed by adding a check for `meterProvider != nil` before initializing metrics and using the `noop metrics` pattern to prevent panics in other code paths.

swiatekm:
> what’s the recommended way to verify secured metrics in this e2e environment? should the test disable secure metrics (`--metrics-secure=false`), or is there a supported way (token / auth mechanism) to authenticate against the metrics endpoint?

I think the simplest way is to fetch the metrics in a Job and do your checks inside it. The job has to have the right RBAC, but that's not too difficult to add inside the cluster.  To scrape metrics e.g. via Prometheus the client needs a ClusterRole with the following rule: * nonResourceURLs: "/metrics", verbs: get .

Issue #4366, opened 2025-09-16
Support enabling user namespaces for collector instances


t3mi:
### Component(s)

collector

### Is your feature request related to a problem? Please describe.

With Kubernetes 1.33, [user namespaces](https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/) are supported to enhance pod security by isolating pod processes in a separate user namespace, reducing the risk of privilege escalation. It would be awesome if operator could support setting that for collector instances.

### Describe the solution you'd like

Expand collector CRD to include toggle for `hostUsers` property

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

alliasgher:
This appears to already be done — #4659 added `hostUsers` to `OpenTelemetryCommonFields` ([apis/v1beta1/common.go L244-246](https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1beta1/common.go#L244-L246)) and wired it into the collector's deployment, daemonset, and statefulset PodSpecs (plus tests) back in March. The same was done for the TargetAllocator. Should this issue be closed?

PR #4659, opened 2026-01-28
Support enabling user namespaces for collector instances

am1r021:
### Description

This change adds the `hostUsers` field to `OpenTelemetryCommonFields` to enable toggling support for isolating pod processes under a separate user namespace.

### Fixes #
#4366 

### Testing

Added unit tests to verify `hostUsers` propagation for all workload builders:
* Collector Deployment
* Collector DaemonSet
* Collector StatefulSet
* Target Allocator Deployment

Each test validates three cases:
* unset (nil) in CR => pod.spec.hostUsers remains unset
* explicit true => pod.spec.hostUsers: true
* explicit false => pod.spec.hostUsers: false

### Documentation

Updated API docs and generated manifests to reflect the new `hostUsers` field on `OpenTelemetryCommonFields`.

PR Assisted by OpenAI Codex

### DIFF
--- a/.chloggen/hostusers-podspec.yaml
+++ b/.chloggen/hostusers-podspec.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: collector
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add the hostUsers field to OpenTelemetryCommonFields to enable toggling support for isolating pod processes under a separate user namespace
+
+# One or more tracking issues related to the change
+issues: [4366]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:


apis/v1beta1/common.go
```diff
@@ -241,6 +241,9 @@ type OpenTelemetryCommonFields struct {
 	// This is only applicable to Service resources.
 	// +optional
 	TrafficDistribution *string `json:"trafficDistribution,omitempty"`
+	// HostUsers isolates pod processes in a separate user namespace, reducing the risk of privilege escalation.
+	// +optional
+	HostUsers bool `json:"hostUsers,omitempty"`
```
iblancasa: I think you need to pipeline this to the pod spec.

apis/v1beta1/common.go
```diff
@@ -241,6 +241,9 @@ type OpenTelemetryCommonFields struct {
 	// This is only applicable to Service resources.
 	// +optional
 	TrafficDistribution *string `json:"trafficDistribution,omitempty"`
+	// HostUsers isolates pod processes in a separate user namespace, reducing the risk of privilege escalation.
```
iblancasa, line 244: I think you need to run `make generate` to autogenerate some go code

apis/v1beta1/common.go
am1r021: I switched the type to be a pointer to keep it the same as the `corev1.PodSpec.HostUsers` spec, as seen [here](https://pkg.go.dev/k8s.io/api/core/v1#PodSpec). Then I plugged it into the PodSpec builders.

apis/v1beta1/common.go
am1r021: Since it was a `bool` before, `make generate` was silently succeeding since it didn't need special deep copy handling. After changing its type to a `*bool`, `make update` started generating the deep copy code.

PR #4669, opened 2026-02-02
Add support for TLS profile

pavolloffay:

  Added operator flag `--tls-cluster-profile` which obtains the TLS min version and cipher suites from the OpenShift `APIServer` `cluster` custom resource (CR).
  It overrides the `--tls-min-version` and `--tls-cipher-suites` flags if set.
  The flags is disabled by default on Kubernetes and enabled on OpenShift.
  
  Added operator flag `--tls-configure-operands` which configures operands TLS settings (min version, cipher suites)
  based on the supplied operator TLS flags (`--tls-cipher-suites` and `--tls-min-version`) or from the OpenShift `APIServer` CR
  if `--tls-cluster-profile` is enabled.
  The flag is disabled by default on Kubernetes and enabled on OpenShift.
  
  The `--tls-min-version` defaults to `TLSv1.2` which matches the collector's default.
  The `--tls-cipher-suites` is empty by default which matches the collector's default.
  Therefore enabling `--tls-configure-operands` with the default TLS flags should not change the collector's behavior.

**Follow up work**

* The operator is restarted on TLS change.
* The collector config is updated in the defaulting webhook.
* Therefore the update of the TLS profile in the APIServer is not applied once the CR is created.

##### TLS between collector and TA

* TLS is enabled with `--feature-gates=+operator.targetallocator.mtls`.
* TA server config does not expose TLS min version and ciphers - https://github.com/open-telemetry/opentelemetry-operator/blob/605a8d7479c0a3f1734afe4cf0112ffade0bffd1/cmd/otel-allocator/internal/config/config.go#L97-L103

- Resolves: #issue-number

Test manifest:

```yaml
kind: OpenTelemetryCollector
metadata:
  name: otel
spec:
#  image: registry.redhat.io/rhosdt/opentelemetry-collector-rhel8@sha256:aaf9bfbe6fcc0459ee1860d08dc832b666bf67af04e9d62300ce31c9177d74d3
  volumeMounts:
    - name: serving-certs
      mountPath: /etc/serving-certs
      readOnly: true
  volumes:
    - name: serving-certs
      secret:
        secretName: otel-collector-headless-tls
  config:
    receivers:
      otlp:
        protocols:
          grpc:
            endpoint: 0.0.0.0:4317
            tls:
              cert_file: /etc/serving-certs/tls.crt
              key_file: /etc/serving-certs/tls.key
          http:
            endpoint: 0.0.0.0:4318
            tls:
              cert_file: /etc/serving-certs/tls.crt
              key_file: /etc/serving-certs/tls.key
      zipkin:
        tls:
          cert_file: /etc/serving-certs/tls.crt
          key_file: /etc/serving-certs/tls.key
    processors:
    exporters:
      debug: {}
    service:
      pipelines:
        traces:
          receivers: [otlp, zipkin]
          exporters: [debug]
        metrics:
          receivers: [otlp]
          exporters: [debug]
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: openshift-service-ca
  annotations:
    service.beta.openshift.io/inject-cabundle: "true"
data: {}
---
kind: OpenTelemetryCollector
metadata:
  name: metrics-sender
spec:
  image: registry.redhat.io/rhosdt/opentelemetry-collector-rhel8@sha256:aaf9bfbe6fcc0459ee1860d08dc832b666bf67af04e9d62300ce31c9177d74d3
  volumeMounts:
    - name: openshift-service-ca
      mountPath: /etc/pki/ca-trust/source/service-ca
      readOnly: true
  volumes:
    - name: openshift-service-ca
      configMap:
        name: openshift-service-ca
  config:
    receivers:
      hostmetrics:
        collection_interval: 30s
        scrapers:
          cpu: {}
          memory: {}
          disk: {}
    processors:
      batch: {}
    exporters:
      otlp:
        endpoint: otel-collector-headless.ploffay.svc.cluster.local:4317
        tls:
          # The exporter defaults to insecure false
          # If TLS block is missing.
          insecure: true
          ca_file: /etc/pki/ca-trust/source/service-ca/service-ca.crt
    service:
      pipelines:
        metrics:
          receivers: [hostmetrics]
          processors: [batch]
          exporters: [otlp]
---
kind: OpenTelemetryCollector
metadata:
  name: otelko
spec:
  image: registry.redhat.io/rhosdt/opentelemetry-collector-rhel8@sha256:aaf9bfbe6fcc0459ee1860d08dc832b666bf67af04e9d62300ce31c9177d74d3
  config:
    receivers:
      otlp:
        protocols:
          grpc: {}
          http: {}
      zipkin: {}
    processors:
    exporters:
      debug: {}
    service:
      pipelines:
        traces:
          receivers: [otlp, zipkin]
          exporters: [debug]
        metrics:
          receivers: [otlp]
          exporters: [debug]
```

### DIFF
--- a/internal/components/multi_endpoint.go
+++ b/internal/components/multi_endpoint.go
@@ -79,23 +79,23 @@ func (m *MultiPortReceiver) GetDefaultConfig(logger logr.Logger, config any, opt
 	}
 	defaultedConfig := map[string]any{}
 	for protocol, ec := range multiProtoEndpointCfg.Protocols {
-		if defaultSvc, ok := m.portMappings[protocol]; ok {
-			port := defaultSvc.Port
-			if ec != nil {
-				port = ec.GetPortNumOrDefault(logger, port)
-			}
-			addr := m.defaultRecAddr
-			if defaultAddr, ok := m.addrMappings[protocol]; ok {
-				addr = defaultAddr
-			}
-			conf, err := AddressDefaulter(logger, defaultCfg, addr, port, ec)
-			if err != nil {
-				return nil, err
-			}
-			defaultedConfig[protocol] = conf
-		} else {
+		defaultSvc, ok := m.portMappings[protocol]
+		if !ok {
 			return nil, fmt.Errorf("unknown protocol set: %s", protocol)
 		}
+		port := defaultSvc.Port
+		if ec != nil {
+			port = ec.GetPortNumOrDefault(logger, port)
+		}
+		addr := m.defaultRecAddr
+		if defaultAddr, ok := m.addrMappings[protocol]; ok {
+			addr = defaultAddr
+		}
+		conf, err := AddressDefaulter(logger, defaultCfg, addr, port, ec)
+		if err != nil {
+			return nil, err
+		}
+		defaultedConfig[protocol] = conf
 	}
 	return map[string]any{
 		"protocols": defaultedConfig,
--- a/internal/components/tls_profile.go
+++ b/internal/components/tls_profile.go
@@ -39,12 +39,12 @@ func TLSVersionToCollectorFormat(version uint16) string {
 		return "1.0"
 	case tls.VersionTLS11:
 		return "1.1"
-	case tls.VersionTLS12:
-		return "1.2"
 	case tls.VersionTLS13:
 		return "1.3"
+	case tls.VersionTLS12:
+		fallthrough
 	default:
-		return "1.2" // Default to TLS 1.2
+		return "1.2"
 	}
 }
 
--- a/internal/controllers/opentelemetrycollector_controller.go
+++ b/internal/controllers/opentelemetrycollector_controller.go
@@ -247,7 +247,7 @@ func (r *OpenTelemetryCollectorReconciler) Reconcile(ctx context.Context, req ct
 
 	// We have a deletion, short circuit and let the deletion happen
 	// Remove finalizer if RBAC permission not available
-	deletionTimestamp, err := removeFinalizer(r, ctx, params, &instance)
+	deletionTimestamp, err := removeFinalizer(ctx, r, params, &instance)
 	if err != nil || deletionTimestamp != nil {
 		return ctrl.Result{}, err
 	}
@@ -384,7 +384,7 @@ func maybeAddFinalizer(params manifests.Params, instance *v1beta1.OpenTelemetryC
 	return false
 }
 
-func removeFinalizer(r *OpenTelemetryCollectorReconciler, ctx context.Context, params manifests.Params, instance *v1beta1.OpenTelemetryCollector) (*metav1.Time, error) {
+func removeFinalizer(ctx context.Context, r *OpenTelemetryCollectorReconciler, params manifests.Params, instance *v1beta1.OpenTelemetryCollector) (*metav1.Time, error) {
 	deletionTimestamp := instance.GetDeletionTimestamp()
 	if deletionTimestamp != nil || params.Config.CreateRBACPermissions != rbac.Available {
 		if controllerutil.ContainsFinalizer(instance, collectorFinalizer) {
--- a/internal/controllers/opentelemetrycollector_reconciler_test.go
+++ b/internal/controllers/opentelemetrycollector_reconciler_test.go
@@ -247,7 +247,7 @@ func TestRemoveFinalizer(t *testing.T) {
 				},
 			}
 
-			deletionTS, err := removeFinalizer(reconciler, context.Background(), params, instance)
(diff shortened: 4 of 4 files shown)


config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: docker.io/pavolloffay/opentelemetry-operator
+  newTag: dev-d852dcab-1770051364
```
jaronoff97: remove from commit

apis/v1beta1/config.go
```diff
@@ -298,7 +308,7 @@ func (c *Config) applyDefaultForComponentKinds(logger logr.Logger, componentKind
 		for componentName := range enabledComponents[componentKind] {
 			parser := retriever(componentName)
 			componentConf := cfg.Object[componentName]
-			newCfg, err := parser.GetDefaultConfig(logger, componentConf)
+			newCfg, err := parser.GetDefaultConfig(logger, componentConf, tlsSettings)
```
jaronoff97: I understand why we need to drill it this way, but i'm wondering if there's a better strategy here using the options pattern instead of this. I could imagine other parsers will want to take advantage of this type of arg passing and I think it may be cleaner for the design for this to take variadic param of options that are loaded in the generic parsers. Using variadic args would mean that none of the other tests would need to change and we could also send down more config later too.

pavolloffay:
@IshwarKanse  re

> Bug: The TLS observer in main.go does not handle permission errors gracefully, operator crashes if it cannot access the APIServer CR at startup.

Can the `APIServer` `cluster` be missing in the cluster? Is it a valid state of the cluster?

IshwarKanse:
@pavolloffay 

> Bug: The TLS observer in main.go does not handle permission errors gracefully, operator crashes if it cannot access the APIServer CR at startup.
> 
> Can the APIServer cluster be missing in the cluster? Is it a valid state of the cluster?

No, I was simulating an inaccessible API server for testing. 

For 

> ⏺ Bug Found: TLS Observer poll loop is not started

When the cluster TLS profile changes (e.g., Intermediate → Modern), existing collectors don't pick up the new profile because AddressDefaulter skips non-empty fields. The only way to trigger re-injection is to either delete/recreate the collector or patch the CR to null out min_version/cipher_suites. Is this the intended behavior, or should the operator handle this automatically?

Apart from this query, I have finished testing the PR and have also added a e2e test for this in https://github.com/open-telemetry/opentelemetry-operator/pull/4680

apis/v1beta1/config.go
```diff
@@ -354,8 +369,34 @@ func (c *Config) GetAllRbacRules(logger logr.Logger) ([]rbacv1.PolicyRule, error
...
+type DefaultsOption func(*DefaultsConfig)
+
+// WithTLSProfile sets the TLS profile to use when applying defaults.
+// When set, TLS settings (min_version, cipher_suites) are injected into
+// components that have a tls: block with cert_file configured.
+func WithTLSProfile(tls components.TLSProfile) DefaultsOption {
+	return func(cfg *DefaultsConfig) {
+		cfg.TLSProfile = tls
+	}
+}
```
jaronoff97: why do we also need options here? can main.go not just create the internal components option and we pass that option all the way down?

pavolloffay: I am not sure I understand the question. 

For components the TLS settings are sourced during reconciliation. When config/configmap is being created the `.Config.ApplyDefaults` is being called e.g. 

```
// Apply TLS defaults at reconciliation time if configured.
	// This ensures collectors get updated TLS settings when the operator restarts
	// after a cluster TLS profile change, without requiring CR updates.
	if params.Config.Internal.OperandTLSProfile != nil {
		_, err := otelCol.Spec.Config.ApplyDefaults(params.Log, v1beta1.WithTLSProfile(params.Config.Internal.OperandTLSProfile))
		if err != nil {
			params.Log.Error(err, "failed to apply TLS defaults to collector config")
			return nil, err
		}
	}
```

The `ApplyDefaults` uses the options pattern as requested here https://github.com/open-telemetry/opentelemetry-operator/pull/4669#discussion_r2764713415

jaronoff97: i understand that, i'm wondering why we need two identical structs? we have one in config here and another in the internal package... actually i think i understand why we need both... nvm.

jaronoff97: nvm i think there's a way to do this without both. going to fork this PR and try it out.

pavolloffay: thanks @jaronoff97

PR #4689, opened 2026-02-12
adopt otel exporter names

frzifus:

https://github.com/open-telemetry/opentelemetry-collector/commit/f33c609d976f2682f7d9c19cc92a05ffbe17427e

- Resolves: #issue-number

### DIFF
--- a/.chloggen/otlp-exporter-migration.yaml
+++ b/.chloggen/otlp-exporter-migration.yaml
@@ -1,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. filelogreceiver)
+component: operator
+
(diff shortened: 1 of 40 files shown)


.chloggen/otlp-exporter-migration.yaml
```diff
@@ -0,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. filelogreceiver)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add automatic migration and deprecation warnings for OTLP exporter name changes (otlp→otlp_grpc, otlphttp→otlp_http)
```
iblancasa, line 10: I like the deprecation warnings but not 100% sure about the renaming. Do we really want to do it? Because users will need to change their manifests anyway and the collector will continue working (at least for now).

Or is the idea to keep the logic even after the collector drops the support?

Issue #3128, opened 2024-07-15
An algorithm for new target allocation, based on the average job algorithm.


zh-w:
### Component(s)

target allocator

### Is your feature request related to a problem? Please describe.

In the scenario where I use target-allocator, there are usually different jobs, and the metric datapoints of targets for each job vary significantly. For example, there are five types of collection jobs: A, B, C, D, and E. Suppose each job has the same collection interval and each job has 10 targets. The number of datapoints pulled by each target of job A is 1000 (e.g., KSM), for job B it is 100, for jobs C and D it is 50, and for job E it is 10.

At the same time, assume I have 5 collector instances deployed in StatefulSets. When using consistent-hashing or least-weighted algorithms, the targets for each job are not evenly distributed across each collector instance. In the assumed collection scenario, it is possible that collector-0 is assigned 3 targets of job A, while collector-4 is assigned 0 targets of job A. This can result in a significant disparity in the number of datapoints collected by each collector, leading to an imbalance in load.

In my actual use case, this situation occurs quite frequently. Below is a diagram showing the load distribution of collectors in a large cluster I deployed (using the consistent-hashing algorithm), illustrating the extreme imbalance in resource utilization across each collector.
![image](https://github.com/user-attachments/assets/36af843e-c443-4e17-8df3-638b97f78763)

### Describe the solution you'd like

I have implemented a load-balancing algorithm based on jobs. The algorithm is designed as follows:

zh-w:
The algorithm is designed as follows:                                                                                                                                                     if service discovery targets change {
    for each deleted target:
        remove the corresponding target from the original collector without triggering rebalancing
    for each new target:
        for the job's targets, find the collector with the smallest number of assigned targets for that job. If multiple collectors meet the criteria, choose the collector with the smallest total number of targets.
} else if the list of collector instances changes {
    for each new collector:
        add the collector to the list and reassign all targets
    for each removed collector:
        remove the collector from the list and reassign all targets
}

jaronoff97:
@zh-w what do the metrics from the target allocator look like for the targets per collector metric? It's possible that the distribution is even, but the memory usage is not in the scenario where some of your scrape targets emit more metrics than others.

swiatekm:
I wouldn't really mind if we just made the `least-weighted` strategy behave this way. @zh-w would that work? Out of the set of collectors with the least amount of targets, we'd pick the one with the least amount of targets from the assigned target's job. Right now we just pick randomly.

[...]

PR #4704, opened 2026-02-16
Weight-class configuration for LeastWeighted target allocation strategy

suprjinx:
**Description:**

This PR implements weight-based target allocation for the `least-weighted` strategy, so that heavier targets are spread across collectors rather than piling up on one. In the absence of weight information, the strategy falls back to existing `NumTargets` balancing.

### How to assign weights to targets

Add the `opentelemetry.io/target-allocation-weight` annotation to the target pod with an integer value from 1 to 100:

```yaml
metadata:
  annotations:
    opentelemetry.io/target-allocation-weight: "10"
```

For pods you don't control (e.g. kube-state-metrics, nginx-ingress), you can patch the deployment's pod template:

```bash
kubectl patch deployment nginx-ingress -p '{"spec":{"template":{"metadata":{"annotations":{"opentelemetry.io/target-allocation-weight":"10"}}}}}'
```

Kubernetes SD automatically surfaces this as the meta label `__meta_kubernetes_pod_annotation_opentelemetry_io_target_allocation_weight`, which the allocator reads directly.

### Weight resolution

1. Pod annotation value (integer 1-100) — if present and valid
2. Default: **1**

Values outside the 1-100 range, non-numeric, or missing annotations all fall back to the default weight of 1.

### How it works

The `least-weighted` strategy tracks `WeightedLoad` (sum of weights) per collector instead of just target count. When assigning a new target, it picks the collector with the lowest `WeightedLoad`. This ensures heavy targets get spread across collectors.

- Relates: #3128

**Testing:**
Unit tests covering: weighted load balancing, unlabeled targets defaulting to weight 1, invalid annotations (non-numeric, zero, negative, >100), and balanced distribution of heavy targets.

**Documentation:**
Needed

PR Assisted by Claude Code


cmd/otel-allocator/internal/prehook/relabel.go
```diff
@@ -51,7 +51,7 @@ func (tf *relabelConfigTargetFilter) Apply(targets []*target.Item) []*target.Ite
 			targets[writeIndex] = target.NewItem(
 				tItem.JobName,
 				tItem.TargetURL,
-				tItem.Labels,
+				builder.Labels(),
```
suprjinx: this is a significant change -- the filtered labels were not persisted to the item previously

jaronoff97: I understand the goal of this PR, but this change could indeed be decently dangerous. I have to spend some time thinking through the consequences of this. cc @swiatekm

swiatekm: There's no way we can do this, it's a major breaking change. My original thought was that we were going to use a Pod annotation, because we get those straight from K8s service discovery, and don't need relabeling.

suprjinx: > There's no way we can do this, it's a major breaking change. My original thought was that we were going to use a Pod annotation, because we get those straight from K8s service discovery, and don't need relabeling.

Thanks @swiatekm @jaronoff97 -- I will remove the relabeling. I can use pod annotation, but i think it will also require some new config option for TA (in order to supply a weight config for Pods we don't control)

suprjinx: Ready for another look, with TA CR changes for WeightOverrides

Issue #3772, opened 2025-03-06
Explore a way to set readiness of the operator deployment checking for the webhook server


iblancasa:
> Just blinding searching since I'm not familiar with the operator but I think here: https://github.com/open-telemetry/opentelemetry-operator/blob/00f2d89038f96a30a9b5758163d9a661d2ea6731/main.go#L504-L507 

 _Originally posted by @nrfox in [#3194](https://github.com/open-telemetry/opentelemetry-operator/issues/3194#issuecomment-2701479249)_

It would be a good idea to explore adding here something like a wrapper of the `Ping` function that checks the status of the webhook server or something.

PR #4778, opened 2026-03-02
Add webhook readiness check to operator /readyz endpoint

jinja2:

The operator's `/readyz` endpoint currently uses `healthz.Ping` which returns 200 before the webhook server is listening. This means k8s marks the pod as `Ready` and starts routing webhook traffic to it while the TLS listener is still spinning up. The result is intermittent connection refused errors when CRs are created right after deployment.
This is easy to reproduce with Helm: `helm install --wait` returns as soon as the readiness probe passes, and a post-install hook that creates an Instrumentation CR will sometimes fail because the webhook isn't serving yet.

With this PR, when webhooks are enabled, we register a webhook ready check using controller-runtime's `StartedChecker()` ([reference](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.22.5/pkg/webhook/server.go#L273)). This checker does a real TLS dial to the webhook port, so the pod won't report ready until the listener is genuinely accepting connections.

Linking to an already existing issue which mentions adding a similar feature.

- Resolves: #3772

Replaced the ` go run hack/check-operator-ready.go 300 ` in e2e-test with a simple `kubectl wait` for the operator deployment to be available. With the webhook check now in place, we just need to wait for pod to be available (ready + minReadySeconds if set) instead of trying to create a CR.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -126,8 +126,6 @@ linters:
           disabled: true # FIXME
         - name: struct-tag
           disabled: true # FIXME
-        - name: superfluous-else
-          disabled: true # FIXME
         - name: unchecked-type-assertion
           disabled: true
         - name: unexported-return
@@ -138,8 +136,6 @@ linters:
           disabled: true # FIXME
         - name: unsecure-url-scheme
           disabled: true
-        - name: unused-parameter
-          disabled: true # FIXME
         - name: use-waitgroup-go
           disabled: true
         - name: var-naming
--- a/apis/v1alpha1/instrumentation_webhook.go
+++ b/apis/v1alpha1/instrumentation_webhook.go
@@ -46,7 +46,7 @@ type InstrumentationWebhook struct {
 	scheme *runtime.Scheme
 }
 
-func (w InstrumentationWebhook) Default(ctx context.Context, obj runtime.Object) error {
+func (w InstrumentationWebhook) Default(_ context.Context, obj runtime.Object) error {
 	instrumentation, ok := obj.(*Instrumentation)
 	if !ok {
 		return fmt.Errorf("expected an Instrumentation, received %T", obj)
@@ -54,7 +54,7 @@ func (w InstrumentationWebhook) Default(ctx context.Context, obj runtime.Object)
 	return w.defaulter(instrumentation)
 }
 
-func (w InstrumentationWebhook) ValidateCreate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+func (w InstrumentationWebhook) ValidateCreate(_ context.Context, obj runtime.Object) (admission.Warnings, error) {
 	inst, ok := obj.(*Instrumentation)
 	if !ok {
 		return nil, fmt.Errorf("expected an Instrumentation, received %T", obj)
@@ -62,7 +62,7 @@ func (w InstrumentationWebhook) ValidateCreate(ctx context.Context, obj runtime.
 	return w.validate(inst)
 }
 
-func (w InstrumentationWebhook) ValidateUpdate(ctx context.Context, oldObj, newObj runtime.Object) (admission.Warnings, error) {
+func (w InstrumentationWebhook) ValidateUpdate(_ context.Context, _, newObj runtime.Object) (admission.Warnings, error) {
 	inst, ok := newObj.(*Instrumentation)
 	if !ok {
 		return nil, fmt.Errorf("expected an Instrumentation, received %T", newObj)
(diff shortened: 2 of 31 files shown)


jinja2:
The failing tests are related to TA and look unrelated to my changes - 

```
=== NAME  chainsaw/targetallocator-kubernetessd
    sink.go:56: | 18:46:54 | targetallocator-kubernetessd | step-01                     | ASSERT    | ERROR | batch/v1/Job @ chainsaw-whole-aphid/check-metrics
        === ERROR
        -----------------------------------------------
        batch/v1/Job/chainsaw-whole-aphid/check-metrics
        -----------------------------------------------
        * status.succeeded: Required value: field not found in the input object
        
        --- expected
        +++ actual
        @@ -3,6 +3,5 @@
         metadata:
           name: check-metrics
           namespace: chainsaw-whole-aphid
        -status:
        -  succeeded: 1
        +status: {}
    sink.go:56: | 18:46:54 | targetallocator-kubernetessd | step-01                     | ASSERT    | RUN   | batch/v1/Job @ chainsaw-whole-aphid/check-ta-jobs
=== NAME  chainsaw/targetallocator-namespace
    sink.go:56: | 18:46:58 | targetallocator-namespace | @chainsaw                                                                            | CLEANUP   | END   |
=== NAME  chainsaw/targetallocator-prometheuscr
    sink.go:56: | 18:52:40 | targetallocator-prometheuscr | step-01                     | ASSERT    | ERROR | batch/v1/Job @ chainsaw-becoming-gorilla/check-ta-jobs
        === ERROR
        ----------------------------------------------------
        batch/v1/Job/chainsaw-becoming-gorilla/check-ta-jobs
        ----------------------------------------------------
        * status.succeeded: Required value: field not found in the input object
        
        --- expected
        +++ actual
        @@ -3,6 +3,5 @@
         metadata:
           name: check-ta-jobs
           namespace: chainsaw-becoming-gorilla
        -status:
        -  succeeded: 1
        +status: {}
    sink.go:56: | 18:52:40 | targetallocator-prometheuscr | step-01                     | ASSERT    | RUN   | batch/v1/Job @ chainsaw-becoming-gorilla/check-ta-scrape-configs
=== NAME  chainsaw/targetallocator-kubernetessd
    sink.go:56: | 18:52:54 | targetallocator-kubernetessd | step-01                     | ASSERT    | ERROR | batch/v1/Job @ chainsaw-whole-aphid/check-ta-jobs
        === ERROR
        -----------------------------------------------
        batch/v1/Job/chainsaw-whole-aphid/check-ta-jobs
        -----------------------------------------------
        * status.succeeded: Required value: field not found in the input object
        
        --- expected
        +++ actual
        @@ -3,6 +3,5 @@
         metadata:
           name: check-ta-jobs
           namespace: chainsaw-whole-aphid
        -status:
        -  succeeded: 1
        +status: {}
    sink.go:56: | 18:52:54 | targetallocator-kubernetessd | step-01                     | ASSERT    | RUN   | batch/v1/Job @ chainsaw-whole-aphid/check-ta-scrape-configs
=== NAME  chainsaw/targetallocator-prometheuscr
    sink.go:56: | 18:58:40 | targetallocator-prometheuscr | step-01                     | ASSERT    | ERROR | batch/v1/Job @ chainsaw-becoming-gorilla/check-ta-scrape-configs
        === ERROR
        --------------------------------------------------------------
        batch/v1/Job/chainsaw-becoming-gorilla/check-ta-scrape-configs
        --------------------------------------------------------------
        * status.succeeded: Required value: field not found in the input object
        
        --- expected
        +++ actual
        @@ -3,6 +3,5 @@
         metadata:
           name: check-ta-scrape-configs
           namespace: chainsaw-becoming-gorilla
        -status:
        -  succeeded: 1
        +status: {}
 ```

File: tests/e2e/webhook-readiness/chainsaw-test.yaml [hunk 1]
```diff
@@ -0,0 +1,82 @@
+apiVersion: chainsaw.kyverno.io/v1alpha1
```
swiatekm: Given that we verify operator readiness by submitting CRDs before even starting e2e tests, I don't think this test achieves anything. Instead, you can try replacing uses of https://github.com/open-telemetry/opentelemetry-operator/blob/8a56b9b3ae32b17c630707ca4e68567f77596637/Makefile#L324 with a simpler Pod readiness check using kubectl. If your change is effective, this should suffice.

swiatekm (commented):
Thank you for fixing this! I recall wondering why we needed to work around the readiness check being so unreliable. Turns out we were just holding it wrong.

tests/e2e/webhook-readiness/chainsaw-test.yaml
jinja2: Makes sense. Updated this and removed the test

Makefile
```diff
@@ -321,7 +321,7 @@ add-rbac-permissions-to-operator: manifests kustomize
 .PHONY: deploy
 deploy: set-image-controller
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
-	go run hack/check-operator-ready.go 300
```
iblancasa, line 324: Maybe we can now remove this script?

swiatekm: This can happen in a follow-up, once we're confident this change doesn't lead to any regressions.

pavolloffay: If it's not used, I would remove it right away.

pavolloffay (approved):

Makefile
swiatekm: It is used for the upgrade tests. Maybe we can get rid of it there as well, but it's less obvious, and I'd rather not hold this PR up with that discussion. WDYT?

PR #4818, opened 2026-03-08
Add collectorSelector field to standalone TargetAllocator CR

### LINKED ISSUE #5016: OpenTelemetry Operator TargetAllocator CR does not support collectorSelector
### Component(s)

target allocator

## Description

The current implementation of the OpenTelemetry Operator and CRD does not support collectorSelectors for the TargetAllocator CR.

 

This is mandatory as without any workload in the Cluster with a certain label will be assigned as collector depending on the RBAC permission of the serviceAccount.

The Env OTELCOL_NAMESPACE can limit the selected collectors but still in the example I end up with none OTC pods being assigned and therefor missing Scrape Targets.

## Steps to Reproduce

- create a namespace

```
oc create reproducer
```

- create a TargetAllocator

```
cat <<'EOF' | oc -n reproducer create -f-
kind: TargetAllocator
metadata:
  name: reproducer
spec:
  allocationStrategy: consistent-hashing
  collectorNotReadyGracePeriod: 30s
  filterStrategy: relabel-config
  ipFamilyPolicy: SingleStack
  managementState: managed
  observability:
    metrics:
      enableMetrics: false
  prometheusCR:
    enabled: false
    scrapeInterval: 30s
  replicas: 1
  scrapeConfigs:
    - fallback_scrape_protocol: PrometheusText1.0.0
      job_name: reproducer
      scheme: http
      static_configs:
        - targets:
            - host01:9090
            - host02:9090
            - host03:9090
            - host04:9090
            - host05:9090
            - host06:9090
            - host07:9090
            - host08:9090
            - host09:9090
            - host10:9090
            - host11:9090
            - host12:9090
            - host13:9090
            - host14:9090
            - host15:9090
            - host16:9090
            - host17:9090
            - host18:9090
            - host19:9090
            - host20:9090
EOF
```

- use port-forwarding to see which collectors have been assigned 

```
oc -n reproducer port-forward service/reproducer-targetallocator 8080:80 & 
cat <<'EOF' | python 
import re, requests
pattern = r'.*href="/debug/collector.+?>([^<]+)</a>'
rsp = requests.get("http://localhost:8080").text
for match in re.finditer(pattern, rsp):
    print(match.groups()[0])

EOF
```

- will output the targetAllocator itself and or multiple other pods in that namespace

- add a simple httpd pod to verify targetAllocator selects it for target dispatching

```
cat <<'EOF' | oc -n reproducer create -f-
apiVersion: v1
kind: Pod
metadata:
  name: example
  labels:
    app: httpd
  namespace: targetallocatorreproducer
spec:
  securityContext:
    runAsNonRoot: true
    seccompProfile:
      type: RuntimeDefault
  containers:
    - name: httpd
      image: 'image-registry.openshift-image-registry.svc:5000/openshift/httpd:latest'
      ports:
        - containerPort: 8080
      securityContext:
        allowPrivilegeEscalation: false
        capabilities:
          drop:
            - ALL
EOF
```

- re-check the targetAllocator collector list

```
oc -n reproducer port-forward service/reproducer-targetallocator 8080:80 & 
cat <<'EOF' | python 
import re, requests
pattern = r'.*href="/debug/collector.+?>([^<]+)</a>'
rsp = requests.get("http://localhost:8080").text
for match in re.finditer(pattern, rsp):
    print(match.groups()[0])

EOF
```

-  this for example outputs now (after adding two httpd pods)

```
example
example2
reproducer-targetallocator-75596bd594-mp4np
```

- the UI of the targetAllocator also shows assigned Job and TargetCounts

```
Collector    Job Count    Target Count
example    1    6
example2    1    10
reproducer-targetallocator-75596bd594-mp4np    1    4
```

### Kubernetes Version

1.32

### Operator version

1.48.0

### Collector version

1.48.0

## Environment
OS: (e.g., "Ubuntu 20.04")

### Log output

```shell

```

### Additional context

OpenShift, orignally:
https://redhat.atlassian.net/browse/TRACING-5977

### Tip

frzifus:

- Resolves: https://github.com/open-telemetry/opentelemetry-operator/issues/5016

### DIFF
--- a/.chloggen/ta_col_selector.yaml
+++ b/.chloggen/ta_col_selector.yaml
@@ -1,0 +1,19 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: target allocator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add collectorSelector field to standalone TargetAllocator CR to prevent all pods in the namespace from being treated as collectors.
+
+# One or more tracking issues related to the change
+issues: []
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  When using the standalone TargetAllocator CR without a linked OpenTelemetryCollector, the collector_selector
+  was set to nil, causing all pods in the namespace to be treated as collectors. The new collectorSelector field
+  allows users to specify which pods should receive target assignments.
--- a/.chloggen/ta_col_selector.yaml
+++ b/.chloggen/ta_col_selector.yaml
@@ -8,7 +8,7 @@ component: target allocator
 note: Add collectorSelector field to standalone TargetAllocator CR to prevent all pods in the namespace from being treated as collectors.
 
 # One or more tracking issues related to the change
-issues: []
+issues: [4818]
 
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
--- a/.chloggen/ta_col_selector.yaml
+++ b/.chloggen/ta_col_selector.yaml
@@ -8,7 +8,7 @@ component: target allocator
 note: Add collectorSelector field to standalone TargetAllocator CR to prevent all pods in the namespace from being treated as collectors.
 
 # One or more tracking issues related to the change
-issues: [4818]
+issues: [5016]
 
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
--- a/apis/v1alpha1/targetallocator_types.go
+++ b/apis/v1alpha1/targetallocator_types.go
@@ -86,6 +86,12 @@ type TargetAllocatorSpec struct {
 	// NetworkPolicy defines the network policy to be applied to the Target Allocator pods.
 	// +optional
 	NetworkPolicy v1beta1.NetworkPolicy `json:"networkPolicy,omitempty"`
+	// CollectorSelector defines the label selector for collector pods that the TargetAllocator should
+	// assign targets to. This is required when using the standalone TargetAllocator CR without an
+	// associated OpenTelemetryCollector CR. If not set and no collector is linked, all pods in the
+	// namespace will be treated as collectors.
+	// +optional
+	CollectorSelector *metav1.LabelSelector `json:"collectorSelector,omitempty"`
 	// CollectorNotReadyGracePeriod defines the grace period after which a TargetAllocator stops considering a collector is target assignable.
 	// The default is 30s, which means that if a collector becomes not Ready, the target allocator will wait for 30 seconds before reassigning its targets. The assumption is that the state is temporary, and an expensive target reallocation should be avoided if possible.
 	//
--- a/apis/v1alpha1/zz_generated.deepcopy.go
+++ b/apis/v1alpha1/zz_generated.deepcopy.go
@@ -1673,6 +1673,11 @@ func (in *TargetAllocatorSpec) DeepCopyInto(out *TargetAllocatorSpec) {
 	in.PrometheusCR.DeepCopyInto(&out.PrometheusCR)
 	in.Observability.DeepCopyInto(&out.Observability)
 	in.NetworkPolicy.DeepCopyInto(&out.NetworkPolicy)
+	if in.CollectorSelector != nil {
+		in, out := &in.CollectorSelector, &out.CollectorSelector
+		*out = new(metav1.LabelSelector)
+		(*in).DeepCopyInto(*out)
+	}
 	if in.CollectorNotReadyGracePeriod != nil {
 		in, out := &in.CollectorNotReadyGracePeriod, &out.CollectorNotReadyGracePeriod
 		*out = new(metav1.Duration)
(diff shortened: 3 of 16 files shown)


swiatekm:
The change looks correct to me, but why are we adding it?

Independently, I'd love to see an e2e test for it.

frzifus:
> The change looks correct to me, but why are we adding it?

Without `collectorSelector` the TA treats every pod in the namespace as a collector and hands it scrape targets - whether it's an actual OTel collector or a random pod. :smile: 


> Independently, I'd love to see an e2e test for it.

Sure, I  will add some!

swiatekm:
> Without collectorSelector the TA treats every pod in the namespace as a collector and hands it scrape targets - whether it's an actual OTel collector or a random pod. 😄

Unless you use the label, then it works correctly. I was curious if you have a use case where that's not sufficient and you actually need a selector.

michaelalang:
@swiatekm sorry if I misunderstood your comment but what we see when deploying a targetAllocator with following configuration:
```
kind: TargetAllocator
metadata:
  name: ta-test
  namespace: ta-test
spec:
  allocationStrategy: consistent-hashing
  collectorNotReadyGracePeriod: 30s
  filterStrategy: relabel-config
  ipFamilyPolicy: SingleStack
  managementState: managed
  prometheusCR:
    allowNamespaces:
      - ta-test
    enabled: true
    scrapeInterval: 30s
    serviceMonitorSelector:
      matchExpressions:
        - key: ta-test-scrape
          operator: In
          values:
            - 'true'
  replicas: 1
---
```
the targetAllocator return's any pod in the namespace (or returned by KubeAPI) as collector including itself.

```
oc port-forward service/ta-test-targetallocator 8080:80 &
curl localhost:8080

            <a href="/debug/collector?collector_id=httpd-859dc75bf-8dblg">httpd-859dc75bf-8dblg</a>
            <a href="/debug/collector?collector_id=ta-test-targetallocator-7cb6b4dc7-gqjvk">ta-test-targetallocator-
...
```

the targetAllocator is missing the `collectorSelector` to be configurable and avoid listing all pods if receives from the Kubernetes API 

```
$ oc explain TargetAllocator.spec --recursive | grep collectorSelector | wc -l
0
```

modifying the Operand to `managementState: unmanaged` and adjusting the configMap from
```
allocation_strategy: consistent-hashing
collector_not_ready_grace_period: 30s
collector_selector: null
filter_strategy: relabel-config
prometheus_cr:
  allow_namespaces:
  - ta-test
  enabled: true
  pod_monitor_selector: null
  probe_selector: null
  scrape_config_selector: null
  scrape_interval: 30s
  service_monitor_selector:
    matchlabels: {}
    matchexpressions:
    - key: ta-test-scrape
      operator: In
      values:
      - "true"
```
to a more proper `collector_selector` like

```
allocation_strategy: consistent-hashing
collector_not_ready_grace_period: 30s
collector_selector:
  matchLabels:
    # app.kubernetes.io/name: ta-collector # < if being very specific
filter_strategy: relabel-config
prometheus_cr:
  allow_namespaces:
  - ta-test
  enabled: true
  pod_monitor_selector: null
  probe_selector: null
  scrape_config_selector: null
  scrape_interval: 30s
  service_monitor_selector:
    matchlabels: {}
    matchexpressions:
    - key: ta-test-scrape
      operator: In
      values:
      - "true"
 ```
 
 will result in only the particular collectors matching being shown

swiatekm:
@michaelalang returning all Pods from the namespace is incorrect behavior, it should return none. But f you add the right label to an OpenTelemetryCollector CR in the same namespace, it'll use that.

To be clear, I don't mind making this configurable via a explicit selector. But this isn't a bug fix, it's a feature. And the implementation has to decide what to do if both the label and the selector are set. I suspect that's the reason the e2e test is failing for this PR.

I'm asking these questions because I suspect there's a misunderstanding on how this currently works.

michaelalang:
> @michaelalang returning all Pods from the namespace is incorrect behavior, it should return none. But f you add the right label to an OpenTelemetryCollector CR in the same namespace, it'll use that.
> 
> To be clear, I don't mind making this configurable via a explicit selector. But this isn't a bug fix, it's a feature. And the implementation has to decide what to do if both the label and the selector are set. I suspect that's the reason the e2e test is failing for this PR.
> 
> I'm asking these questions because I suspect there's a misunderstanding on how this currently works.

I totally agree with your Point of view that it should return None but, it's not which is why we consider it being a Bug not a feature.

swiatekm:
> > @michaelalang returning all Pods from the namespace is incorrect behavior, it should return none. But f you add the right label to an OpenTelemetryCollector CR in the same namespace, it'll use that.
> > 
> > To be clear, I don't mind making this configurable via a explicit selector. But this isn't a bug fix, it's a feature. And the implementation has to decide what to do if both the label and the selector are set. I suspect that's the reason the e2e test is failing for this PR.
> > 
> > I'm asking these questions because I suspect there's a misunderstanding on how this currently works.
> 
> I totally agree with your Point of view that it should return None but, it's not which is why we consider it being a Bug not a feature.
> 
> 

If you'd like to submit a fix to that bug, I'll happily accept it. But that's not the only thing this PR does. It adds a feature allowing the selector to be set via the CRD. That's also not controversial, but I'd prefer these changed to be separate.

michaelalang:
> > > @michaelalang returning all Pods from the namespace is incorrect behavior, it should return none. But f you add the right label to an OpenTelemetryCollector CR in the same namespace, it'll use that.
> > > To be clear, I don't mind making this configurable via a explicit selector. But this isn't a bug fix, it's a feature. And the implementation has to decide what to do if both the label and the selector are set. I suspect that's the reason the e2e test is failing for this PR.
> > > I'm asking these questions because I suspect there's a misunderstanding on how this currently works.
> > 
> > 
> > I totally agree with your Point of view that it should return None but, it's not which is why we consider it being a Bug not a feature.
> 
> If you'd like to submit a fix to that bug, I'll happily accept it. But that's not the only thing this PR does. It adds a feature allowing the selector to be set via the CRD. That's also not controversial, but I'd prefer these changed to be separate.

Ack will talk with @frzifus to coordinate both as you state the default behavior should be fixed too 
thanks for clarification and quick responses <3

PR #4846, opened 2026-03-10
[apis/v1beta1] move webhooks to internal/webhook

atoulme:
**Description:**
Similar to #4837


iblancasa (approved):
You need to regenerate the docs, I think

Issue #4848, opened 2026-03-11
Auto-instrumentation init container security context no longer inherited from app container


JulesTriomphe:
### Component(s)

auto-instrumentation

## Description

Due to the changes In #4635, the security context of auto-instrumentation init containers is no longer set in pods instrumented using Operator version v0.146.0.
As a result, auto-instrumentation fails in namespaces with restricted Pod Security Standards.

The bug is in `setInitContainerSecurityContext` being called before the instrumentation init container is added to the pod. Here's the exact broken sequence in each `injectXxx` function (shown for Python, same pattern for Java, NodeJS, DotNet):

```go
// CURRENT (BROKEN) — sdk.go, injectPython()
if len(containers) > 0 {
    for _, container := range containers {
        if err := injectPythonSDKToContainer(...); err != nil {
            ...
        } else {
            i.injectCommonEnvVar(otelinst, container)
            i.injectDefaultPythonEnvVars(container)
            pod = i.injectCommonSDKConfig(...)
            pod = i.setInitContainerSecurityContext(pod, container.SecurityContext, pythonInitContainerName)
            //   ↑ NO-OP: "opentelemetry-auto-instrumentation-python" is NOT in pod.Spec.InitContainers yet
        }
    }
    pod = injectPythonSDKToPod(...)  // ← init container added HERE, without SecurityContext
}
```

`setInitContainerSecurityContext` iterates over `pod.Spec.InitContainers` looking for `pythonInitContainerName`, finds nothing (it doesn't exist yet), and returns the pod unchanged. Then `injectPythonSDKToPod` appends the init container with no SecurityContext, which violates the *restricted* Pod Security Standard.

This was working in all pre-v0.146.0 versions because the init container was appended inside the same loop body, before `setInitContainerSecurityContext` was called.

### Minimal Fix

Move `setInitContainerSecurityContext` to after `injectXxxToPod` in all four affected functions in `internal/instrumentation/sdk.go`:

```go
// FIXED — injectPython()
if len(containers) > 0 {
    for _, container := range containers {
        if err := injectPythonSDKToContainer(...); err != nil {
            ...
        } else {
            i.injectCommonEnvVar(otelinst, container)
            i.injectDefaultPythonEnvVars(container)
            pod = i.injectCommonSDKConfig(...)
            // ← remove setInitContainerSecurityContext from here
        }
    }
[...]

PR #4849, opened 2026-03-11
fix: Fix instrumentation init container security context

JulesTriomphe:

Due to the changes In #4635, the security context of auto-instrumentation init containers is no longer set in pods instrumented using Operator version v0.146.0.
As a result, auto-instrumentation fails in namespaces with restricted Pod Security Standards.

The bug is in `setInitContainerSecurityContext` being called before the instrumentation init container is added to the pod. Here's the exact broken sequence in each `injectXxx` function (shown for Python, same pattern for Java, NodeJS, DotNet):

```go
// CURRENT (BROKEN) — sdk.go, injectPython()
if len(containers) > 0 {
    for _, container := range containers {
        if err := injectPythonSDKToContainer(...); err != nil {
            ...
        } else {
            i.injectCommonEnvVar(otelinst, container)
            i.injectDefaultPythonEnvVars(container)
            pod = i.injectCommonSDKConfig(...)
            pod = i.setInitContainerSecurityContext(pod, container.SecurityContext, pythonInitContainerName)
            //   ↑ NO-OP: "opentelemetry-auto-instrumentation-python" is NOT in pod.Spec.InitContainers yet
        }
    }
    pod = injectPythonSDKToPod(...)  // ← init container added HERE, without SecurityContext
}
```

`setInitContainerSecurityContext` iterates over `pod.Spec.InitContainers` looking for `pythonInitContainerName`, finds nothing (it doesn't exist yet), and returns the pod unchanged. Then `injectPythonSDKToPod` appends the init container with no SecurityContext, which violates the *restricted* Pod Security Standard.

This was working in all pre-v0.146.0 versions because the init container was appended inside the same loop body, before `setInitContainerSecurityContext` was called.

- Resolves: #4848

**Testing:** Added functionality tests.

### DIFF
--- a/internal/instrumentation/podmutator_test.go
+++ b/internal/instrumentation/podmutator_test.go
@@ -4,7 +4,6 @@
 package instrumentation
 
 import (
-	"context"
 	"errors"
 	"fmt"
 	"testing"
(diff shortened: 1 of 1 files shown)


iblancasa (commented):
Please, fix the CI issues.

internal/instrumentation/podmutator_test.go
```diff
@@ -5227,3 +5227,126 @@ func TestInitContainerInstrumentation(t *testing.T) {
...
+
+func TestInitContainerInstrumentationCopiesSecurityContext(t *testing.T) {
+	inst := &v1alpha1.Instrumentation{
+		ObjectMeta: metav1.ObjectMeta{Name: "python-inst-sc", Namespace: "default"},
+		Spec: v1alpha1.InstrumentationSpec{
+			Python:   v1alpha1.Python{Image: "otel/python:1"},
+			Exporter: v1alpha1.Exporter{Endpoint: "http://collector:4318"},
+		},
+	}
+	err := k8sClient.Create(context.Background(), inst)
```
swiatekm:
```suggestion
	err := k8sClient.Create(t.Context(), inst)
```

Same thing where we use the context elsewhere in this test.

swiatekm (commented):
LGTM, minor nitpick about the use of context in tests. @jaronoff97 @pavolloffay could either of you have a look as well?

internal/instrumentation/podmutator_test.go
JulesTriomphe: Shall I do the same for all uses of the context in this file? It's used as `context.Background()` also in other tests.

swiatekm: If you can do it in a separate commit, that'd be a nice improvement.

JulesTriomphe: Sure --> https://github.com/open-telemetry/opentelemetry-operator/pull/4849/commits/86433c928a2995163c424ce9bbc1ac846036df49

PR #4871, opened 2026-03-18
Add TLS profile support for extensions and  receivers                                                                                                                                                                                   

ozzywalsh:
- Add TLS profile injection to the health_check extension                                                                              
 - Add TLS profile injection to jaeger_query extension                                                                                                                                            
 - Add TLS profile support to all scraper receivers                                                                                                                                                                                      
 - Add PrometheusScraper to apply TLS profile settings when scraping Prometheus endpoints (prometheys uses a  different tls config format)

### DIFF
--- a/internal/components/receivers/scraper.go
+++ b/internal/components/receivers/scraper.go
@@ -64,14 +64,7 @@ func scraperDefaulter(defaultCfg *components.DefaultConfig, config *ScraperConfi
 		config = &ScraperConfig{}
 	}
 
-	if defaultCfg != nil && defaultCfg.TLSProfile != nil && config.TLS != nil {
-		if config.TLS.MinVersion == "" && defaultCfg.TLSProfile.MinTLSVersionOTEL() != "" {
-			config.TLS.MinVersion = defaultCfg.TLSProfile.MinTLSVersionOTEL()
-		}
-		if config.TLS.Ciphers == nil && len(defaultCfg.TLSProfile.CipherSuiteNames()) > 0 {
-			config.TLS.Ciphers = defaultCfg.TLSProfile.CipherSuiteNames()
-		}
-	}
+	config.TLS.ApplyTLSProfileDefaults(defaultCfg.TLSProfile)
 
 	res := make(map[string]any)
 	err := mapstructure.Decode(config, &res)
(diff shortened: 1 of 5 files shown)


.chloggen/TRACING-5985.yaml
pavolloffay, line 1: Rename the file and remove the link to internal jira from the PR description

.chloggen/tls-profile-extensions-scrapers.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+# One or more tracking issues related to the change
+issues: [4871]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  When a TLS profile is configured on the cluster, the operator now injects min_version and cipher_suites
+  into health_check and jaeger_query extension configs, as well as scraper receiver configs (including Prometheus),
```
pavolloffay: `scraper receiver configs` this seems like an internal terminology of the codebase. I would avoid it here.

ozzywalsh: Agreed. I've removed that terminology.

internal/components/receivers/scraper.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	if defaultCfg != nil && defaultCfg.TLSProfile != nil && config.TLS != nil {
+		if config.TLS.MinVersion == "" && defaultCfg.TLSProfile.MinTLSVersionOTEL() != "" {
+			config.TLS.MinVersion = defaultCfg.TLSProfile.MinTLSVersionOTEL()
+		}
+		if config.TLS.Ciphers == nil && len(defaultCfg.TLSProfile.CipherSuiteNames()) > 0 {
+			config.TLS.Ciphers = defaultCfg.TLSProfile.CipherSuiteNames()
+		}
+	}
```
swiatekm: Seems like this logic is repeated at least three times in this PR, and it also looks really ugly. Can we move it to an appropriately named helper function?

ozzywalsh: Good idea; I  have moved it into a shared helper.

PR #4892, opened 2026-03-23
Add TLS profile injection for exporters

ozzywalsh:
## Summary
- Introduce `ExporterParser` that overrides `GetDefaultConfig` to apply TLS profile defaults (`min_version`, `cipher_suites`) to exporter `tls:` blocks
- Change `ParserFor` default fallback to use `NewExporterParser` so all unregistered exporters get TLS profile injection
- Prometheus exporter already covered via `NewSinglePortParserBuilder` / `AddressDefaulter`
- Follows the same pattern as `ScraperParser` for receivers in #4871

### DIFF
--- a/internal/components/exporters/exporter.go
+++ b/internal/components/exporters/exporter.go
@@ -61,14 +61,7 @@ func exporterDefaulter(defaultCfg *components.DefaultConfig, config *ExporterCon
 		config = &ExporterConfig{}
 	}
 
-	if defaultCfg != nil && defaultCfg.TLSProfile != nil && config.TLS != nil {
-		if config.TLS.MinVersion == "" && defaultCfg.TLSProfile.MinTLSVersionOTEL() != "" {
-			config.TLS.MinVersion = defaultCfg.TLSProfile.MinTLSVersionOTEL()
-		}
-		if config.TLS.Ciphers == nil && len(defaultCfg.TLSProfile.CipherSuiteNames()) > 0 {
-			config.TLS.Ciphers = defaultCfg.TLSProfile.CipherSuiteNames()
-		}
-	}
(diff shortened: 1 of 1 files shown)


internal/components/exporters/exporter_test.go
```diff
@@ -0,0 +1,156 @@
...
+		expectMinVersion string
+		expectCiphers    []string
+	}{
+		{
+			name: "applies min_version and cipher_suites to tls block",
+			config: map[string]any{
+				"endpoint": "tempo.example.com:4317",
+				"tls":      map[string]any{},
+			},
+			expectMinVersion: "1.2",
```
pavolloffay: It's going to be more readable if we have just `expectedConfig` field

pavolloffay: The assertion will be simpler

ozzywalsh: Ok good idea; I updated it with a simpler test case & assertion structure.

internal/components/exporters/exporter.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/go-logr/logr"
+	"github.com/mitchellh/mapstructure"
+
+	"github.com/open-telemetry/opentelemetry-operator/internal/components"
+)
+
+var _ components.Parser = &ExporterParser{}
+
+// ExporterConfig represents the minimal config for push-based exporters.
+// Only the tls: block is relevant for defaulting; other fields are preserved
```
pavolloffay: I would remove this line. or move the to the TLS

ozzywalsh:
Updated to use tls helper function from merged PR https://github.com/open-telemetry/opentelemetry-operator/pull/4871

Issue #4918, opened 2026-04-01
Increase tests coverage of target allocator


odubajDT:
### Component(s)

target allocator

### Describe the issue you're reporting

# Improve Unit Test Coverage for Target Allocator

## Summary

The Target Allocator has significant unit test coverage gaps across several internal packages. Existing tests focus primarily on happy-path scenarios and core functionality, while error paths, edge cases, constructors, and utility functions remain untested.

## Current State (Before Changes)

| Package | Test File | Existing Tests | Focus |
|---------|-----------|---------------|-------|
| `internal/diff` | `diff_test.go` | 1 (`TestDiffMaps` with 3 sub-cases) | Basic map diffing only |
| `internal/allocation` | `allocator_test.go` | 7 | Core allocation lifecycle |
| `internal/server` | `server_test.go` | 12 | Happy-path handler responses |
| `internal/prehook` | `relabel_test.go` | 5 | Core relabel filtering |
| `internal/target` | `target_test.go` | ~1 | Minimal hash test |

### `internal/diff` — `diff_test.go`

- **Missing edge cases for `Maps()`**: both-empty maps, empty-current/non-empty-new, non-empty-current/empty-new, identical maps, complete entry swap
- **No test for `NewChanges()` constructor** or its `Additions()`/`Removals()` accessors
- **No nil-input test** for `NewChanges`
- **No test with non-string key types** — `Maps` is generic but only tested with `string` keys

### `internal/allocation` — `allocator_test.go`

- **No test for `GetTargetsForCollectorAndJob` with non-existent collector/job** — should return empty slice, not panic
- **No test for `SetTargets` with empty slice** — clearing all targets
- **No test for `SetCollectors` with empty map** — removing all collectors; targets should become unassigned
- **No idempotency test** — calling `SetTargets` with the same targets shouldn't change state
- **No error path for `New()` with invalid strategy name** — should return error
- **No test for `GetRegisteredAllocatorNames()`** utility function
- **No constructor test for `Collector` struct** — `NewCollector`, `Hash()`, `String()` methods
- **No test for `WithFilter` option** — filter callback integration with allocation
- **No test for repeated `SetCollectors`** — idempotency when collector set doesn't change
- **No multi-job allocation test** — targets from multiple jobs allocated simultaneously

### `internal/server` — `server_test.go`

- **No test for `/jobs/{job}/targets` without `collector_id`** — should return by-collector map response
[...]

PR #4919, opened 2026-04-01
[chore][target-allocator] increase unit tests coverage

odubajDT:
Resolves https://github.com/open-telemetry/opentelemetry-operator/issues/4918


cmd/otel-allocator/internal/allocation/allocator_test.go
```diff
@@ -186,3 +186,188 @@ func TestAllocationCollision(t *testing.T) {
...
+		// since MakeNTargetsForJob doesn't add node labels)
+		assignedCount := 0
+		for _, item := range allocator.TargetItems() {
+			if item.CollectorName != "" {
+				assignedCount++
+			}
+		}
+		// For least-weighted and consistent-hashing, all should be assigned
+		// For per-node, none will be assigned due to missing node labels
+		assert.Len(t, allocator.TargetItems(), 6)
```
swiatekm: This part doesn't make any sense. `assignedCount` isn't used, and the assertion at the end is the same as on line 359. What is the intent here?

cmd/otel-allocator/internal/diff/diff_test.go
```diff
@@ -95,3 +201,52 @@ func TestDiffMaps(t *testing.T) {
...
+func TestNewChanges(t *testing.T) {
+	additions := map[string]Hasher[string]{"a": HasherString("1")}
+	removals := map[string]Hasher[string]{"b": HasherString("2")}
+	c := NewChanges(additions, removals)
+	if !reflect.DeepEqual(c.Additions(), additions) {
+		t.Errorf("Additions() = %v, want %v", c.Additions(), additions)
+	}
+	if !reflect.DeepEqual(c.Removals(), removals) {
+		t.Errorf("Removals() = %v, want %v", c.Removals(), removals)
+	}
```
swiatekm: We should use testify assertions here. I think `assert.Equal` is the equivalent here. Same for the other tests.

cmd/otel-allocator/internal/server/server_test.go
```diff
@@ -1140,3 +1140,224 @@ func TestServer_ScrapeConfigResponse(t *testing.T) {
 func newLink(jobName string) linkJSON {
 	return linkJSON{Link: fmt.Sprintf("/jobs/%s/targets", url.QueryEscape(jobName))}
 }
+
+func TestServer_TargetsHandlerNoCollectorID(t *testing.T) {
+	leastWeighted, _ := allocation.New("least-weighted", logger)
+	listenAddr := ":8080"
```
swiatekm: We don't actually bind to the address in these tests, so this port being here is confusing. An empty string works just as well.

cmd/otel-allocator/internal/server/server_test.go
```diff
@@ -1140,3 +1140,224 @@ func TestServer_ScrapeConfigResponse(t *testing.T) {
...
+	request := httptest.NewRequestWithContext(t.Context(), http.MethodGet, "/jobs/nonexistent/targets?collector_id=test-collector", http.NoBody)
+	w := httptest.NewRecorder()
+
+	s.server.Handler.ServeHTTP(w, request)
+	result := w.Result()
+
+	assert.Equal(t, http.StatusOK, result.StatusCode)
+	bodyBytes, err := io.ReadAll(result.Body)
+	require.NoError(t, err)
+	assert.Equal(t, "[]\n", string(bodyBytes))
```
swiatekm: It'd be better to unmarshal this and assert the list is empty.

swiatekm (changes requested):
This all looks very good, thank you for contributing these tests. I have some questions and minor requests, but nothing very substantial.

cmd/otel-allocator/internal/allocation/allocator_test.go
odubajDT: ah, small mistake from my side, thanks for catching it! Intent is to test that all the jobs are tracked and verifies that the target assignment is all or nothing

PR #4924, opened 2026-04-03
[chore] add standalone TA e2e test

odubajDT:
Part-of: #4918

### Motivation

The Target Allocator (TA) can run as a standalone component without the OpenTelemetry Operator — users deploy it directly with raw Kubernetes manifests when they need Prometheus target sharding across collectors without adopting the full operator lifecycle. Today, no e2e tests validate the TA in this standalone mode. All existing TA tests go through the operator-managed flow, leaving a gap in coverage for:

- Target distribution correctness across collectors
- Target redistribution on collector scale-up/down
- Standalone deployment viability

### What this PR adds

A new Chainsaw e2e test suite (`tests/e2e-ta-standalone/`) that deploys the TA as a raw Deployment with real collector pods — no operator involved. The test validates the core TA functionality end-to-end.

### Test flow

1. **Deploy** — RBAC, TA config, TA deployment, 2 collector replicas
2. **Verify initial allocation** — All 6 targets distributed across `collector-0` and `collector-1` via `GET /jobs/test-targets/targets`
3. **Scale up to 3** — `kubectl scale statefulset/collector --replicas=3`
4. **Verify redistribution** — `collector-2` receives targets, total remains 6
5. **Scale down to 2** — Remove `collector-2`
6. **Verify reassignment** — `collector-2`'s targets move back to remaining collectors


odubajDT:
> > I have two reservations about this test:
> > ```
> > 1. We have tests involving the operator, using the operator-generated manifests. Your test verifies that this also works using an arbitrary manifest which isn't published anywhere. For this to be valuable, we should have an official kustomize manifest for TA which we could then verify in this test.
> > ```
> 
> Makes total sense, would you be against introducing the manifests in a separate PR? We have an internal use-case for the TA, where we effectively do not need to deploy it via operator at this stage, therefore is important for us that this path is tested. Instead of doing downstream testing, we prefer to contribute the tests upstream, so the whole comunity can potentially benefit from them.

related issue https://github.com/open-telemetry/opentelemetry-operator/issues/4945
> 
> > ```
> > 2. The target allocation test logic looks really nasty. Could we rewrite it in Go? Ideally as a Go test that runs against the live cluster. I don't want to maintain any more shell code than strictly necessary.
> > ```
> 
> Makes sense, will try to rewrite them in a nicer way

odubajDT:
closing in favor of https://github.com/open-telemetry/opentelemetry-operator/pull/4981

PR #4942, opened 2026-04-08
Refactor nginx and httpd instrumentation injection scripts

swiatekm:
**Description:**

The Nginx and HTTPD instrumentations require some more elaborate shell scripting to inject the instrumentation. Until now, these scripts lived as plain strings, with variables injected through printf. This PR moves the scripts to actual .sh files under `./internal/instrumentation/scripts`, which are then embedded into the Go code as strings via `go:embed`.

We also add more validation to the CRD fields for passing config paths into these instrumentations.

- Resolves: https://github.com/open-telemetry/opentelemetry-operator/issues/4925

**Testing:**

Updated existing tests.


bundle/community/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -99,7 +99,7 @@ metadata:
     categories: Logging & Tracing,Monitoring,Observability
     certified: "false"
     containerImage: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
-    createdAt: "2026-04-27T14:53:29Z"
+    createdAt: "2026-05-02T14:35:06Z"
```
atoulme, line 102: Is this change needed?

iblancasa: The file is modified when rendering the manifests. It is ok to have this change here.

.chloggen/fix_nginx-instrumentation-arguments.yaml
```diff
@@ -0,0 +1,19 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: auto-instrumentation
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Stop interpolating user-controlled CRD fields into init container shell scripts when injecting Apache HTTPD and Nginx instrumentation
```
iblancasa: I think this sends a different message to what you did. It sounds like "now these values are ignored"

iblancasa (approved):
Just a nit.

.chloggen/fix_nginx-instrumentation-arguments.yaml
swiatekm: Yeah, maybe I should only mention the CRD changes. The rest is effectively a refactor that shouldn't change anything functionally unless you're putting shell commands in these fields. Let me rewrite.

iblancasa (approved):

jaronoff97 (approved):
lgtm, when the merge conflicts are resolved we should be ✅

PR #4949, opened 2026-04-10
Expose prometheus receiver api_server port

ozzywalsh:
## Summary
- Rewrite prometheus receiver parser to extract port from `api_server.server_config.endpoint` using `GenericParser`, so the port is exposed on the collector Service and NetworkPolicy.
- The previous `ScraperParser`-based implementation never exposed a port for the API server.

## Test plan
- [x] Unit tests covering: disabled api_server, missing endpoint, empty endpoint, valid endpoints, coexistence with scrape_configs
- [x] Manual verification with collector CR using `api_server.enabled: true` + `operand.networkpolicy` feature gate

### DIFF
--- a/.chloggen/fix-prometheus-receiver-api-server-port.yaml
+++ b/.chloggen/fix-prometheus-receiver-api-server-port.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: collector
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Expose prometheus receiver api_server port on collector Service and NetworkPolicy
+
+# One or more tracking issues related to the change
+issues: [4949]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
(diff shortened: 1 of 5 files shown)


swiatekm:
It looks like you removed some existing unit tests, why is that?

ozzywalsh:
> It looks like you removed some existing unit tests, why is that?

Good question; I should have mentioned this in PR description.

@swiatekm The removed tests cover TLS profile injection for scrape_configs, which is being removed in #4939 . This PR overlaps with that change (touches same receiver). 

I'd suggest we merge #4939 first.

pavolloffay:
The PR can be now rebased

pavolloffay (approved):

ozzywalsh:
> The PR can be now rebased

@pavolloffay Rebased. Should be ready to merge.

.chloggen/fix-prometheus-receiver-api-server-port.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
```
iblancasa, line 2: Shouldn't his be `enhancement`? For me it is "something was not exposed before but now it is"

iblancasa (approved):

.chloggen/fix-prometheus-receiver-api-server-port.yaml
ozzywalsh: The api_server config was already being passed through to the collector, meaning the feature was fully functional from the collector's side. The bug is that the operator wasn't doing its job; exposing the port on the Service and NetworkPolicy.

With that context; does it make sense as a bug_fix (to me it does)?

I could update the chglog to include this; at the moment it just says what was changed; not what was going wrong before.

swiatekm (approved):

PR #4951, opened 2026-04-11
Allow instrumentation versions to be marked as unupgradable

### LINKED ISSUE #4646: Allow blocking Instrumentation upgrades for versions with major breaking changes
### Component(s)

auto-instrumentation

Major breaking changes in language SDKs and instrumentation libraries are not unheard of. They may even be a consequences of breaking changes in semantic conventions, as seen in https://github.com/open-telemetry/opentelemetry-operator/issues/2542. This presents a particularly difficult problem for the operator due to the way a new default Instrumentation image is rolled out after an operator upgrade. An OpenTelemetryCollector will typically just crash, Kubernetes normal bad rollout countermeasures will kick in, and the damage will be contained.

An Instrumentation, on the other hand, is injected via a Mutating Pod Webhook. This means that there's no immediate effect of an upgrade. It's only new Pods that are instrumented with the new version. The effect of breakage might not be immediately visible, either, as in https://github.com/open-telemetry/opentelemetry-operator/issues/2542, where the data has a different shape, but everything works correctly otherwise. As a result, users may not immediately notice the breakage, and will likely have trouble tracing the root cause back to the operator upgrade. 

This makes major breaking changes in language instrumentations difficult for the operator to manage. It's why the operator uses old versions of the Java and .NET instrumentations by default right now.

My proposed solution is to allow instrumentation versions to be marked as unupgradable in the code. The user would need to handle the upgrade manually on their own. The operator would do its best to inform the user that this is an issue they should deal with, by:

* Emitting an Event describing the problem.
* Adding a description of the problem to the Instrumentation status.
* Returning a warning whenever the Instrumentation resource in question is updated.
* Emitting a warning log line.

This way, the operator could always offer the latest versions for new Instrumentation resources, without breaking older ones with upgrades. Ideally, each such version would contain an informative message instructing the user how to resolve the breakage.

The only alternative I can think of is to let the upgrade happen and put a lot of warnings all over the changelog and operator repo, including a pinned issue, etc. But I don't think that's sufficient, or acceptable.

### Additional context

See https://github.com/open-telemetry/opentelemetry-operator/issues/2542, it's the issue that prompted this proposal.

### Tip

swiatekm:
**Description:**

An instrumentation version can now be marked as unupgradable. The logic is that when we're upgrading from a `source` version to a `target` version, we block the upgrade if `source < blocked <= target`.

An upgrade being blocked has the following consequences:
- A warning is emitted by the admission webhook if the Instrumentation version would be subject to a blocked upgrade.
- If the upgrader runs into a blocked upgrade, it emits an Event and includes the information in the status. 
- The operator also logs blocked upgrades at Info level.

- Resolves: https://github.com/open-telemetry/opentelemetry-operator/issues/4646
- Relates: https://github.com/open-telemetry/opentelemetry-operator/issues/2542

**Testing:**
Unit tests. Adding e2e tests for this feature is a pain, as they require an entry in the blocked versions map. I have e2e tests for it in a separate branch, where I also bump default Java instrumentation to the latest version.


internal/version/unupgradable.go
```diff
@@ -0,0 +1,91 @@
...
+	"strings"
+
+	"github.com/Masterminds/semver/v3"
+
+	"github.com/open-telemetry/opentelemetry-operator/pkg/constants"
+)
+
+// unupgradableInstrumentationVersions contains instrumentation versions that cannot be automatically upgraded from.
+// Outer key is the language, inner key is the version tag, value is the warning message.
+var unupgradableInstrumentationVersions = map[constants.InstrumentationLanguage]map[string]string{}
```
jaronoff97, line 16: how will we fill this in later?

swiatekm: Something like:

```json
{
  "java": {
    "2.0.0": "This version contains breaking changes to the http semantic conventions. See issue X for more information."
  }
}

```

pavolloffay (commented):
Shall we modify the RELEASE.md to update the blocked version. We should also define what a blocked version is.

swiatekm:
> Shall we modify the RELEASE.md to update the blocked version. We should also define what a blocked version is.

I don't think we ever need to update them. Blocked versions exist to let users know they need to upgrade manually and deal with breakage. They don't disappear over time.

pavolloffay:
What I mean is, how do we figure out that a new auto-instrumentation should be categorized and "blocked".

swiatekm:
> What I mean is, how do we figure out that a new auto-instrumentation should be categorized and "blocked".

I don't think this will ever have a uniform process. Ideally, the instrumentation library maintainers will let us know, or it's part of a broader change like https://github.com/open-telemetry/opentelemetry-operator/issues/2542 and it'll be publicized via various channels before the libraries start making the change.

jaronoff97 (approved):

pavolloffay (approved):

yuriolisa (approved):

PR #4970, opened 2026-04-16
fix(opamp-bridge): guard nil processors in validateComponents

byehn:

The opamp-bridge panics with a nil pointer dereference in validateComponents when a remote config omits the `processors` field. `v1beta1.Config.Processors` is *AnyConfig with omitempty tag, so nil is a valid case, but dereferencing it unconditionally.

- Added testdata/no-processors-collector.yaml (minimal config without processors)
- Added a no processors case to `TestClient_Apply`
- Updated `TestClient_Apply` to pass a non-empty componentsAllowed map


swiatekm (approved):

PR #4971, opened 2026-04-17
target allocator: add configurable mTLS option for TA crds

### LINKED ISSUE #4915: Add a config option for enabling target allocator mTLS
### Component(s)

target allocator

Target Allocator has a [feature](https://github.com/open-telemetry/opentelemetry-operator/tree/main/cmd/otel-allocator#service--pod-monitor-endpoint-credentials) where it can talk to the collector over mTLS, securing the connection and allowing metric endpoint credentials to be transferred over it. This feature is currently hidden behind an alpha feature flag `operator.targetallocator.mtls` and depends on cert-manager for provisioning the actual certificates.

It's a feature flag because the intent was to eventually enable it for everyone and be secure by default. However, the feature turned out to be quite complex and has resulted in several subtle bugs that were only uncovered after sustained production use. After discussing this with @jaronoff97 and @TylerHelmuth, we prefer to make it configurable via an attribute on the `OpenTelemetryCollector` and `TargetAllocator` CRDs, and disabled by default. Our reasons are as follows:

- The complexity of the mTLS setup is significant and can produce subtle bugs.
- Authenticated metrics endpoints are relatively niche, so we'd be foisting this complexity on the majority of users without much benefit to them.
- In terms of the default security of scrape targets, we're currently in line with prometheus-operator. The Prometheus web UI isn't secure by default either.

We don't want to completely close the door on enabling the option by default in some farther future, but it should definitely remain configurable.

@open-telemetry/operator-approvers @CharlieTLe please comment if you have a contrary opinion.

### Tip

Horiodino:
**Description:**  
Added  `mtls.enabled` option to Target Allocator (in both `OpenTelemetryCollector` and `TargetAllocator` ). This replaces the `operator.targetallocator.mtls` feature gate.

- Resolves: #4915

**Testing:**  
- Updated unit tests.  
- Updated e2e tests.

**Documentation:**  
- Updated API docs and added changelog.


swiatekm:
You don't need to add this to v1alpha1, we're going to be removing it shortly.

Horiodino:
cc @swiatekm

apis/v1beta1/opentelemetrycollector_types.go
```diff
@@ -237,6 +237,16 @@ type TargetAllocatorEmbedded struct {
...
+
+	// Mtls defines the mTLS configuration for the target allocator. If enabled, the target allocator will communicate with the collector over mTLS.
+	// +optional
+	Mtls *TargetAllocatorMTLS `json:"mtls,omitempty"`
+}
+
+type TargetAllocatorMTLS struct {
+	// Enabled indicates whether to enable mTLS between the target allocator and the collector.
+	// +optional
+	Enabled bool `json:"enabled,omitempty"`
```
swiatekm, line 249: Considering we're going to be adding a way to manually configure this, how about we add a `UseCertManager` attribute here and set it to true by default? Then it's more explicit how this works by default, and we can later validate that this option and the manual one are mutually exclusive.

.github/workflows/e2e-reusable.yaml
```diff
@@ -96,7 +96,7 @@ jobs:
         - group: e2e-metadata-filters
           setup: "add-operator-arg OPERATOR_ARG='--annotations-filter=.*filter.out --annotations-filter=config.*.gke.io.* --labels-filter=.*filter.out' prepare-e2e"
         - group: e2e-ta-collector-mtls
-          setup: "add-operator-arg OPERATOR_ARG='--feature-gates=operator.targetallocator.mtls' add-certmanager-permissions prepare-e2e"
+          setup: "add-certmanager-permissions prepare-e2e"
```
swiatekm, line 99: Honestly, we can just always add these permissions during setup and move these tests to e2e-targetallocator.

swiatekm (commented):
You also need to update the documentation in cmd/otel-allocator to reflect this change.

internal/controllers/opentelemetrycollector_controller.go
```diff
@@ -156,8 +157,18 @@ func getCollectorConfigMapsToKeep(configVersionsToKeep int, configMaps []*corev1
...
 func (r *OpenTelemetryCollectorReconciler) GetParams(ctx context.Context, instance v1beta1.OpenTelemetryCollector) (manifests.Params, error) {
+	// Recheck cert-manager RBAC on each reconciliation to handle late-added permissions.
+	// Only attempt to upgrade from NotAvailable → Available
+	if r.reviewer != nil && r.config.CertManagerAvailability != internalCertmanager.Available {
+		if warnings, err := internalCertmanager.CheckCertManagerPermissions(ctx, r.reviewer); err == nil && warnings == nil {
+			r.log.V(2).Info("cert-manager permissions detected during reconciliation, upgrading availability")
+			r.config.CertManagerAvailability = internalCertmanager.Available
+		}
+	}
+	cfg := r.config
```
swiatekm: Is this really necessary? We're adding another API request in the hot path for very user. If we want to do something like this, it should be a new feature that reloads all the config values based on a watch.

swiatekm (commented):
I think we need to think about this more carefully. This feature should also work for the TargetAllocator CRD, which I don't believe is the case right now. The way this is typically implemented is that there's an attribute on TargetAllocator.spec, and setting it on the Collector forwards it there in https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/manifests/collector/targetallocator.go.

In this case, we also need to do some stuff in the Collector itself, and the right check for this is by looking at params.TargetAllocator, the same way this is done for config manipulation. The reverse of what this PR does, where it checks the collector CR while generating the TargetAllocator manifests.

Does that make sense, @Horiodino ?

ozzywalsh:
@Horiodino Are you working on this? It's something we need; I could take over the further changes required.

Horiodino:
Hey @ozzywalsh, thank you for your concern. I'm currently working on this issue , it's almost done,was busy with end sem exams, will give update shortly.

swiatekm:
This PR was effectively superseded by #5136, so I'm closing it. Feel free to reopen if you'd like to continue any of your work here @Horiodino.

Issue #3188, opened 2024-08-01
Allow overriding the default collector's image entrypoint


rogercoll:
### Component(s)

collector

### Is your feature request related to a problem? Please describe.

The current API provides a configuration option to specify a custom collector's image, but it assumes that the entrypoint of the container will be the actual collector binary and it cannot be overwritten. That is the case of the OpenTelemetry collector images (core, contrib and k8s), but for distributions or extended images the entrypoint of the container might be different. For example, to launch the binary collector within the `elastic-agent`, an `otel` subcommand mode must be provided: `./elastic-agent otel --config otel.yml`.

This not an issue with the OpenTelemetry Helm as the container's entrypoint can be configured with the `command` value: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.yaml#L180

We use the upstream Helm charts with the following values:

```yaml
image:
  repository: docker.elastic.co/beats/elastic-agent
  tag: 8.15.0-SNAPSHOT
command:
  name: "/usr/share/elastic-agent/elastic-agent"
  extraArgs: ["otel"]
  ```
  
The `command` config option is not availble in the operator, thus we cannot launch the Elastic distribution with the current container's entrypoint. (see also https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/)

### Describe the solution you'd like

Ability to override the container's command:

```
apiVersion: opentelemetry.io/v1alpha1
kind: OpenTelemetryCollector
metadata:
  name: daemonset
spec:
  mode: daemonset
  hostNetwork: true
  image: docker.elastic.co/beats/elastic-agent:8.15.0-SNAPSHOT
  command:
    - /usr/share/elastic-agent/elastic-agent
    - otel
  ```

blewis12:
The PR implementing this just got merged, so it should be available in the next minor version of the operator I imagine. @rogercoll I don't have permission to mark this as closed/completed but perhaps you do?

PR #4973, opened 2026-04-17
feat: Introduce Command Arg to Collector Operator

blewis12:
**Description:**
Addresses [this feature request](https://github.com/open-telemetry/opentelemetry-operator/issues/3188) - and implements the ability to configure the collector entry point from the operator. Previously, it was only possible to set the `args` that were passed through and the command would fall back to the default (ie whatever is defined as the entry point in the docker image). This PR allows a user to set arguments like:

```
spec:
  command:
    - "/path/to/agent"
    - "otel"
```

Which would set the entry point/command as `/path/to/agent otel` - currently the helm chart offers the [same capability ](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/c738eda8d0bfa36513acdc9601165a308d9f506b/charts/opentelemetry-collector/values.yaml#L265-L267)

**Testing:**
`internal/webhook/collector_webhook_test.go` => tests the validating webhook’s create path 
`apis/v1alpha1/convert_test.go` => tests that converting from `v1beta1` to `v1alpha1` and vice versa works as expected
`internal/manifests/collector/container_test.go`=> tests manifest generation

**Documentation:** 
New spec.command / CollectorCommand fields in the Go types (v1beta1 and v1alpha1, including kubebuilder markers for CRD validation) - opentelemetrycollectors.md includes generated docs from that (if there's somewhere else to include docs please let me know!)


swiatekm:
Why are you making this a struct instead of just keeping `Command` as a slice, the same as a Pod spec?

CC @open-telemetry/operator-approvers I vaguely recall there was a reason we didn't allow this to be changed, am I misremembering?

blewis12:
thanks for taking a look @swiatekm - I used a struct here because I was emulating the structure that exists in the [helm chart](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/c738eda8d0bfa36513acdc9601165a308d9f506b/charts/opentelemetry-collector/values.yaml#L265-L267)  so thought it made sense for it to look similar on the operator level - do you think it makes more sense to have it be a slice instead?

i'm also curious if there were reasons this wasn't implemented before, since the related issue has been open for a while and has some upvotes

blewis12:
ah, though I modeled this after the upstream helm chart with the struct - it probably doesn't make sense here as the operators already defines [args](https://github.com/open-telemetry/opentelemetry-operator/blob/893cc7a25d6481d230c7ac643b5fa8c1474fb579/apis/v1beta1/common.go#L108) (which the helm chart doesn't) so there's overlap with `command.extraArgs` there. A slice might be conceptually better and also aligns with the pod spec

internal/manifests/collector/container.go
```diff
@@ -130,6 +131,13 @@ func Container(cfg config.Config, logger logr.Logger, otelcol v1beta1.OpenTeleme
 	}
 }
 
+func containerCommandArgv(cmd []string) []string {
+	if len(cmd) == 0 {
+		return nil
+	}
```
swiatekm: `slices.Clone` returns nil when passed nil, so it already does what you want. We don't need this function at all.

internal/webhook/collector_webhook.go
```diff
@@ -166,6 +166,14 @@ func (c CollectorWebhook) ValidateDelete(ctx context.Context, otelcol *v1beta1.O
 func (c CollectorWebhook) Validate(ctx context.Context, r *v1beta1.OpenTelemetryCollector) (admission.Warnings, error) {
 	warnings := admission.Warnings{}
 
+	if len(r.Spec.Command) > 0 {
+		for i, a := range r.Spec.Command {
+			if strings.TrimSpace(a) == "" {
+				return warnings, fmt.Errorf("spec.command[%d] must be non-empty", i)
+			}
+		}
+	}
```
swiatekm: Is this a validation that Pods do as well for this field? I'd like it more if we could express it using a kubebuilder or CEL validation instead.

apis/v1alpha1/opentelemetrycollector_types.go
```diff
@@ -161,6 +161,13 @@ type OpenTelemetryCollectorSpec struct {
 	// Image indicates the container image to use for the OpenTelemetry Collector.
 	// +optional
 	Image string `json:"image,omitempty"`
+	// Command overrides the container entrypoint (Pod.spec.containers[].command), for example
+	// ["/path/to/binary", "subcommand"]. When omitted, the image ENTRYPOINT is used.
+	// +optional
+	// +listType=atomic
+	// +kubebuilder:validation:MaxItems=50
+	// +kubebuilder:validation:MinItems=1
+	Command []string `json:"command,omitempty"`
```
swiatekm: We don't need to touch v1alpha1 or the conversion. We're going to deprecate and remove v1alpha1 Collector, so there's no point in touching it.

internal/webhook/collector_webhook.go
pavolloffay: +1 on CEL

apis/v1alpha1/opentelemetrycollector_types.go
blewis12: ah ok 👍 i'll remove that then

internal/manifests/collector/container.go
blewis12: good point 👍 https://github.com/open-telemetry/opentelemetry-operator/pull/4973/changes/4d63db12e9d974161bb75aaee66a77149a0e34dd

apis/v1alpha1/opentelemetrycollector_types.go
blewis12: https://github.com/open-telemetry/opentelemetry-operator/pull/4973/changes/f0f8ad590205793fc949244f92800bc46db4601d

internal/webhook/collector_webhook.go
blewis12: Hm right, looking at the pod logic I can see there's no enforcements on non-empty command argv (looking [here](https://github.com/kubernetes/kubernetes/blob/ba220d79acaa62d78f6377e70bc5166faf1e2633/pkg/apis/core/validation/validation.go#L3878))

blewis12: Removing any sort of validation on empty argv would probably just result in a runtime error. The helm chart tries to protect against this seemingly by [removing empty values](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-helm-charts%20%20%7B%7B-%20if%20.Values.command.name%20%7D%7D&type=code) (ie falling back to the default cmd if something like "" is passed through).

I think it would be a cleaner experience to validate on empty argv's (via CEL also makes more sense), but open to the argument of just falling back to pod validation if ya'll think that makes more sense?

blewis12:
hm, the e2e tests that failed here seem unrelated to this PR and pass locally 🤔 could be flaky?

swiatekm (approved):

internal/webhook/collector_webhook.go
swiatekm: I'm ok with the current state of requiring a nonempty array.

apis/v1beta1/opentelemetrycollector_types.go
```diff
@@ -143,6 +143,11 @@ type OpenTelemetryCollectorSpec struct {
 	// This is only applicable to Deployment mode.
 	// +optional
 	DeploymentUpdateStrategy appsv1.DeploymentStrategy `json:"deploymentUpdateStrategy,omitempty"`
+	// Command overrides the container entrypoint (Pod.spec.containers[].command). When omitted, the image ENTRYPOINT is used.
+	// +optional
+	// +listType=atomic
+	// +kubebuilder:validation:MinItems=1
+	Command []string `json:"command,omitempty"`
```
frzifus, line 150: We are now getting close to the point where the operator can be used to deploy almost anything. 💦

frzifus (commented):
@blewis12 could you please rebase the PR? It has some conflicts due to the 152 release.

apis/v1beta1/opentelemetrycollector_types.go
blewis12: Just so I understand, do you think this could be a negative/slippery slope? Imo the operator has a surface area that is flexible but still adheres to what you'd expect of a collector distro, the entry point of which isn't clearly defined. But I definitely want to make sure this PR doesn't push the operator spec into fuzzy or confusing territory

blewis12: (btw let me know if you think this is something worth bringing up in the SIG, happy to put it on the agenda for the next one!)

blewis12: heya @frzifus - just following up on this one

jaronoff97 (approved):

apis/v1beta1/opentelemetrycollector_types.go
swiatekm: We discussed this during a SIG meeting and decided we're ok with it.

swiatekm:
@blewis12 could you fix the failing tests? Afterwards, we can merge this.

apis/v1beta1/opentelemetrycollector_types.go
blewis12: oh nice! ok - will fix tests now to finalize the PR

blewis12:
> @blewis12 could you fix the failing tests? Afterwards, we can merge this.

done 👍

PR #4986, opened 2026-04-22
Standalone mode for OpAMP k8s bridge

### LINKED ISSUE #4913: Standalone OpAMP-to-Kubernetes binary for updating Collector configs without the operator
## Background

The `operator-opamp-bridge` is currently bundled within this repository and tightly coupled to the `OpenTelemetryCollector` CRD. It acts as the Kubernetes-native equivalent of the OpAMP Supervisor: it connects to an OpAMP server, receives remote configurations, and applies them by creating/updating `OpenTelemetryCollector` custom resources via `controller-runtime`.

This model requires the OpenTelemetry Operator to be installed in the cluster. However, there is a significant set of users who:

- Run collectors deployed as plain Deployments/DaemonSets with Kubernetes `ConfigMap`-based configuration (no operator CRD installed),
- Want remote configuration management via OpAMP without adopting the full operator stack,
- Use GitOps tooling (e.g., ArgoCD, Flux) where the collector workload is managed externally, but the collector config (a `ConfigMap`) can be updated independently.

## Problem statement

Today, there is no supported way to use OpAMP remote configuration to update a collector's `ConfigMap` directly in Kubernetes. Users must choose between:
1. Installing the full OpenTelemetry Operator and using the `OpenTelemetryCollector` CRD, or
2. Managing collector configuration entirely out-of-band (no OpAMP).

## Proposed solution

Extract the common OpAMP agent logic from `operator-opamp-bridge` into a shared library and ship a second standalone binary — a **ConfigMap-targeting OpAMP bridge** — that:

1. Connects to an OpAMP server (same protocol, same `opamp-go` client),
2. Receives remote configuration,
3. Applies the configuration by patching Kubernetes `ConfigMap` objects (rather than `OpenTelemetryCollector` CRs),
4. Optionally triggers a rolling restart of the collector workload (Deployment/DaemonSet) so the new config is picked up.

### Shared library / common layer

- The OpAMP client lifecycle (`Start`, `Shutdown`, heartbeating, `onMessage` dispatch, `RemoteConfigStatus` reporting) should be extracted so both binaries share the same code path.
- The `ConfigApplier` interface (already defined in `internal/operator/client.go`) is a natural seam — a `ConfigMap`-backed implementation would satisfy it without touching the agent or OpAMP protocol logic.
- Health reporting should remain extensible so each backend (CRD vs. ConfigMap) can describe the health of its managed resources appropriately.

### Standalone binary (`cmd/configmap-opamp-bridge` or similar)

- Must run without the OpenTelemetry Operator CRDs being present in the cluster.
- Must be able to identify target `ConfigMap`(s) by label selectors or explicit name/namespace, matching keys received from the OpAMP server (following the same `name/namespace` key convention as the existing bridge).
- Must require minimal RBAC: `get`/`update` on `ConfigMaps` and optionally `patch` on `Deployments`/`DaemonSets` for restart signaling.
- Should support the same configuration surface as the existing bridge (endpoint, TLS, headers, capabilities).
- Should not depend on any `operator-framework` or `controller-runtime` manager/webhook machinery — a plain Kubernetes client is sufficient.

### Out of scope

- Full parity with all `OpenTelemetryCollector`-specific features (component allow-listing validation, proxy/pass-through mode, per-pod health rollup via the OpAMP proxy).
- Replacing the existing bridge — both binaries should coexist; the operator-managed path remains unchanged.

## Why here and not in `opamp-go`?

The Kubernetes-specific adapter belongs in this repository. The `opamp-go` library is intentionally Kubernetes-agnostic. The refactoring is purely internal to the operator repository.

owais:
**Description:**
This PR implements a basic standlone mode for the OpAMP bridge that works directly on collector configmap(s) instead of operator CRDs. It does not depend on the operator. The ConfigApplier interface almost already satisfies everything we need with some minor tweaks. The change is mostly a new implementation of the ConfigApplier interface that works directly with configmaps. 

It watches configmaps with a specific label and reports them to the OpAMP server. Users can then remotely manage that config from the server.

- Resolves: #4913 

**How to review**:

- `internal/agent` & `internal/operator` packages moves things around a bit to allow the new standalone mode to implement the existing Config Applier interface. No new features or changes to existing features in these packages.
- `internal/standalone` contains all of the new feature code. It implements a standalone client and an implementation of the ConfigApplier interface. 
- `config/standalone-bridge` contains the k8s manifests used to deploy the bridge in standlone mode.

**Testing:**
- Tests added for the new implementation & updated existing tests where needed
- Manually tested in a local kind cluster


owais:
The failing VULN does not affect us. It only affect plugin installs when running docker as a daemon which we do not. We only use it indirectly as an API. 

more details: https://github.com/moby/moby/security/advisories/GHSA-pxq6-2prw-chj9

Patching is not as simple as docker has deprecated the github.com/docker/docker package and replaced it with multiple smaller packages in github.com/moby/moby/*; and we don't use the docker package directly. The call path is:

```
go: github.com/open-telemetry/opentelemetry-operator/apis/v1beta1 imports
	go.opentelemetry.io/contrib/otelconf/v0.3.0 imports
	go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploggrpc imports
	google.golang.org/grpc imports
	google.golang.org/grpc/balancer/roundrobin imports
	google.golang.org/grpc/balancer/endpointsharding tested by
	google.golang.org/grpc/balancer/endpointsharding.test imports
	google.golang.org/grpc/internal/testutils/roundrobin imports
	
go: github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/internal/watcher imports
	github.com/prometheus-operator/prometheus-operator/pkg/operator imports
	github.com/prometheus/prometheus/model/rulefmt imports
	github.com/prometheus/prometheus/promql tested by
	github.com/prometheus/prometheus/promql.test imports
	github.com/prometheus/prometheus/tsdb imports
	github.com/prometheus/prometheus/tsdb/index imports
	github.com/bboreham/go-loser: github.com/docker/docker@v29.3.1+incompatible: reading github.com/docker/docker/go.mod at revision v29.3.1: unknown revision v29.3.1
	
go: github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/internal/config imports
	github.com/prometheus/prometheus/discovery/install imports
	github.com/prometheus/prometheus/discovery/scaleway imports
	github.com/scaleway/scaleway-sdk-go/api/instance/v1 tested by
	github.com/scaleway/scaleway-sdk-go/api/instance/v1.test imports
	github.com/scaleway/scaleway-sdk-go/internal/testhelpers/httprecorder imports
	github.com/scaleway/scaleway-sdk-go/vcr imports
	gopkg.in/dnaeon/go-vcr.v4/pkg/cassette: github.com/docker/docker@v29.3.1+incompatible: reading github.com/docker/docker/go.mod at revision v29.3.1: unknown revision v29.3.1
	
	
go: github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/internal/config imports
	github.com/prometheus/prometheus/discovery/install imports
	github.com/prometheus/prometheus/discovery/gce imports
	google.golang.org/api/compute/v1 imports
	google.golang.org/api/internal/gensupport imports
	github.com/googleapis/gax-go/v2/callctx tested by
	github.com/googleapis/gax-go/v2/callctx.test imports
	google.golang.org/genproto/googleapis/type/color: github.com/docker/docker@v29.3.1+incompatible: reading github.com/docker/docker/go.mod at revision v29.3.1: unknown revision v29.3.1
```

So each one of those packages will need to migrate to moby to pick up the security patch.  I suggest we ignore this vuln as it is essentially a false positive for us and upgrading to a fixed moby version means waiting on close to a dozen OSS packages to do it.

migration details: https://github.com/moby/moby/releases/tag/docker-v29.0.0

atoulme (approved):

jaronoff97:
this is top on my review list, thank you very much for the contribution here and apologies for the delay, i was on PTO.

cmd/operator-opamp-bridge/internal/agent/agent.go
```diff
@@ -118,21 +116,20 @@ func (agent *Agent) generateCollectorPoolHealth() (map[string]*protobufs.Compone
...
 	for _, col := range cols {
 		key := newKubeResourceKey(col.GetNamespace(), col.GetName())
-		podMap, err := agent.generateCollectorHealth(agent.getCollectorSelector(col), col.GetNamespace())
+		podMap, err := agent.generateCollectorHealth(col.GetSelectorLabels(), col.GetNamespace())
 		if err != nil {
 			return nil, err
 		}
 		isPoolHealthy := true
 		for podName, pod := range podMap {
-			// we need to remove the prefix here as we don't have the namespace from the hostname.
```
jaronoff97, line 129: let's leave in the comment

cmd/operator-opamp-bridge/internal/agent/kube_resource_key.go
```diff
@@ -12,21 +12,33 @@ import (
...
 	s := strings.Split(key, "/")
-	// We expect map keys to be of the form name/namespace
-	if len(s) != 2 {
+	// We expect map keys to be of the form namespace/name or kind/namespace/name.
+	switch len(s) {
+	case 2:
+		return newKubeResourceKey(s[0], s[1]), nil
+	case 3:
+		return newKubeResourceKeyWithKind(s[1], s[2], s[0]), nil
+	default:
```
jaronoff97: can we document this in the readme as well for how this should work?

cmd/operator-opamp-bridge/internal/agent/kube_resource_key.go
```diff
@@ -12,21 +12,33 @@ import (
...
 func kubeResourceFromKey(key string) (kubeResourceKey, error) {
 	s := strings.Split(key, "/")
-	// We expect map keys to be of the form name/namespace
-	if len(s) != 2 {
+	// We expect map keys to be of the form namespace/name or kind/namespace/name.
+	switch len(s) {
+	case 2:
+		return newKubeResourceKey(s[0], s[1]), nil
+	case 3:
+		return newKubeResourceKeyWithKind(s[1], s[2], s[0]), nil
```
jaronoff97: we should validate what `kind` can be as i would imagine we should only accept cases where kind is either configmap or otelcol

cmd/operator-opamp-bridge/internal/agent/kube_resource_key.go
```diff
@@ -12,21 +12,33 @@ import (
...
+}
+
 func kubeResourceFromKey(key string) (kubeResourceKey, error) {
 	s := strings.Split(key, "/")
-	// We expect map keys to be of the form name/namespace
-	if len(s) != 2 {
+	// We expect map keys to be of the form namespace/name or kind/namespace/name.
+	switch len(s) {
+	case 2:
+		return newKubeResourceKey(s[0], s[1]), nil
```
jaronoff97: rather than setting empty kind here, let's set it to `otelcol` (a constant)

cmd/operator-opamp-bridge/internal/config/config.go
```diff
@@ -91,6 +91,10 @@ type Config struct {
 	HeartbeatInterval time.Duration       `yaml:"heartbeatInterval,omitempty"`
 	Name              string              `yaml:"name,omitempty"`
 	AgentDescription  AgentDescription    `yaml:"description,omitempty"`
+
+	// Mode selects the operating mode: "operator" (default) uses OpenTelemetryCollector CRDs,
+	// "standalone" discovers Deployments/DaemonSets by label and manages their ConfigMaps.
+	Mode string `yaml:"mode,omitempty"`
```
jaronoff97, line 104: given we're making standalone mode work this way, should we require kind to only ever equal `configmap` when we receive a message?

cmd/operator-opamp-bridge/internal/config/config.go
```diff
@@ -225,15 +229,25 @@ func (c *Config) RemoteConfigEnabled() bool {
...
+			return nil, err
+		}
 	}
 	return client.New(c.ClusterConfig, client.Options{
 		Scheme: scheme.Scheme,
 	})
 }
 
+func (c *Config) IsStandaloneMode() bool {
+	return c.Mode == "standalone"
```
jaronoff97: let's make standalone a constant variable

jaronoff97: also make `operator` a constant that we actually default it to

cmd/operator-opamp-bridge/internal/operator/client.go
```diff
@@ -207,7 +204,19 @@ func (c Client) Delete(name, namespace string) error {
 	return c.k8sClient.Delete(ctx, &result)
 }
 
-func (c Client) ListInstances() ([]v1beta1.OpenTelemetryCollector, error) {
+func (c Client) ListInstances() ([]CollectorInstance, error) {
+	rawInstances, err := c.listRawInstances()
```
jaronoff97: `raw` instance isn't the right descriptor here IMO, it should be something that specifies this is an otelcol crd in someway.

cmd/operator-opamp-bridge/internal/operator/client.go
```diff
@@ -207,7 +204,19 @@ func (c Client) Delete(name, namespace string) error {
...
+	if err != nil {
+		return nil, err
+	}
+	result := make([]CollectorInstance, len(rawInstances))
+	for i := range rawInstances {
+		result[i] = CRDInstance{Col: rawInstances[i]}
```
jaronoff97: can we use a constructor instead of direct instantiation this way?

cmd/operator-opamp-bridge/internal/operator/crd_instance.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func (c CRDInstance) GetNamespace() string {
+	return c.Col.GetNamespace()
+}
+
+func (c CRDInstance) GetConfigMapKey() string {
+	return fmt.Sprintf("%s/%s", c.GetNamespace(), c.GetName())
+}
+
```
jaronoff97: should we move kube_resource_key to be a shared type so we can just return that here?

cmd/operator-opamp-bridge/internal/standalone/client.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+	if _, err = informer.AddEventHandler(handler); err != nil {
+		return fmt.Errorf("failed to add ConfigMap event handler: %w", err)
+	}
+
+	go func() {
+		if err := ca.Start(ctx); err != nil {
+			c.log.Error(err, "ConfigMap cache stopped with error")
+		}
+	}()
```
jaronoff97, line 108: does this method handle graceful shutdown well? Or will we be stuck in a lock?

cmd/operator-opamp-bridge/internal/standalone/client.go
```diff
@@ -0,0 +1,287 @@
...
+	if apierrors.IsNotFound(err) {
+		cm := &v1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      name,
+				Namespace: namespace,
+				Labels:    map[string]string{managedByLabel: managedByValue},
+			},
+			Data: received.Config,
+		}
+		if createErr := c.k8sClient.Create(ctx, cm); createErr != nil {
```
jaronoff97: should we do a permissions check prior to this or during startup to ensure that we can indeed do this?

cmd/operator-opamp-bridge/internal/standalone/client.go
```diff
@@ -0,0 +1,287 @@
...
+	if updateErr := c.k8sClient.Update(ctx, existing); updateErr != nil {
+		return fmt.Errorf("failed to update ConfigMap %s/%s: %w", namespace, name, updateErr)
+	}
+	c.log.Info("Updated ConfigMap", "name", name, "namespace", namespace)
+
+	return c.triggerRollout(ctx, namespace, workloads)
+}
+
+// Delete removes the ConfigMap identified by name/namespace.
+func (c *Client) Delete(name, namespace string) error {
```
jaronoff97: unlike the operator mode, delete here could be dangerous or fail in very unexpected ways. How do we ensure that if we delete a configmap it's not attached to an existing collector? I don't know if deletes are particularly safe to do here.

cmd/operator-opamp-bridge/internal/standalone/client.go
```diff
@@ -0,0 +1,287 @@
...
+	for _, ref := range workloads {
+		parts := strings.SplitN(ref, "/", 2)
+		if len(parts) != 2 {
+			c.log.Error(fmt.Errorf("invalid workload reference %q", ref), "skipping rollout entry")
+			continue
+		}
+		kind, name := parts[0], parts[1]
+
+		switch strings.ToLower(kind) {
+		case "deployment":
```
jaronoff97: do we not allow updating statefulsets?

cmd/operator-opamp-bridge/internal/standalone/client.go
```diff
@@ -0,0 +1,287 @@
...
+}
+
+// GetCollectorPods is not used in standalone mode (pod health is not reported).
+func (*Client) GetCollectorPods(_ map[string]string, _ string) (*v1.PodList, error) {
+	return &v1.PodList{}, nil
+}
+
+// triggerRollout patches the pod template of each workload in the list with a
+// restart annotation, causing a rolling restart.
+func (c *Client) triggerRollout(ctx context.Context, namespace string, workloads []string) error {
```
jaronoff97: like above, i do think we should validate that we have permissions to do this in main

cmd/operator-opamp-bridge/internal/standalone/schema.go
```diff
@@ -0,0 +1,46 @@
...
+	if c.Name != name {
+		return fmt.Errorf("standalone config name %q does not match target name %q", c.Name, name)
+	}
+	if c.Namespace != namespace {
+		return fmt.Errorf("standalone config namespace %q does not match target namespace %q", c.Namespace, namespace)
+	}
+	if len(c.Config) == 0 {
+		return errors.New("standalone config data is required")
+	}
+	for key := range c.Config {
```
jaronoff97: should we do more validation here to ensure it's of the correct form for a collector?

cmd/operator-opamp-bridge/manifests/standalone/configmap.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: otel-opamp-bridge-standalone-config
+  namespace: opentelemetry-opamp-bridge
+data:
+  config.yaml: |
```
jaronoff97, line 7: we have pretty extensive testing for the opamp bridge in e2e and i would like to see some e2e tests for this use case.

jaronoff97 (changes requested):
some requested changes, but overall i would like to see some better test coverage as well as some more thought around the code boundaries. definitely in favor of this, but there's a few open questions I'd like to hear more about.

cmd/operator-opamp-bridge/internal/agent/kube_resource_key.go
owais: I wanted to do that but didn't to preserve backward compat for existing users but would love to if that is not an issue. would actually simplify things.

cmd/operator-opamp-bridge/internal/agent/kube_resource_key.go
owais: I think there are other cases for health check on pods, etc that don't have a kind. Proxy mode might also accept identifiers without it but I'm not sure. We do checks on config sent from the server and don't accept anything other than the supported types. So I think we should be ok but happy to add it.

cmd/operator-opamp-bridge/internal/config/config.go
owais: This is already in place. We reject configs that don't match the supported type.

cmd/operator-opamp-bridge/internal/standalone/client.go
owais: We could but does it make a considerable difference? This error would log and tell the user about lack of permissions anyway. Maybe throwing errors eagerly on startup is better UX but we'd have to collect all permissions the program might ever need and check for all of them. We probably need a thin sub-system to maintain that over time. I'm not opposed to it but probably should do it as a separate work item IMO.

cmd/operator-opamp-bridge/internal/agent/kube_resource_key.go
jaronoff97: it would still preserve backwards compat, we would accept both cases split on the `/`, just internally we would represent this with the same triplet.

cmd/operator-opamp-bridge/internal/standalone/client.go
owais: We aren't. The responsibility is on the user. We make sure we only deleted if the configmap is managed by the remote server and don't touch any other resource but happy to remove delete support altogether. We can reconsider adding back in a future version if we get feedback from users.

cmd/operator-opamp-bridge/internal/standalone/client.go
owais: added support for sts

cmd/operator-opamp-bridge/internal/agent/kube_resource_key.go
owais: I implemented it but it made things more complicated for no strong benefits. I think resource key can stay agnostic and bridge mode can validate what it passes in.

owais:
Addressed some things. Will add more e2e tests.

cmd/operator-opamp-bridge/internal/standalone/client.go
owais: It stops when context is cancelled which reacts to OS interrupt

cmd/operator-opamp-bridge/internal/standalone/client.go
owais: What is the goal here? Is it to fail fast and report this early or to avoid making these calls if we don't have the permissions? Rollout is an optional feature and users might not even enable it. Even if we do check for permissions, we'd have to do it in all namespaces as any configmap in any namespace could be marked to be managed by the user.

cmd/operator-opamp-bridge/internal/standalone/client.go
owais: removed delete path. we don't support it now. we can consider adding ti with guardrails if we see a strong use case for it

owais:
@jaronoff97 please take another look when you can

jaronoff97:
@owais this is top of my list of reviews and i hope to get to this tomorrow morning, apologies for the delay D:

cmd/operator-opamp-bridge/internal/operator/client.go
```diff
@@ -29,20 +30,17 @@ const (
 )
 
 type ConfigApplier interface {
-	// Apply receives a name and namespace to apply an OpenTelemetryCollector CRD that is contained in the configmap.
-	Apply(name, namespace string, configmap *protobufs.AgentConfigFile) error
+	// Apply receives an OpAMP config key and applies the corresponding configuration.
+	Apply(key string, configmap *protobufs.AgentConfigFile) error
```
jaronoff97, line 36: Why are we not passing the key and namespace here anymore? shouldn't we centralize on the kube resource key or at least keep this here so we don't need to split on the key everywhere?

cmd/operator-opamp-bridge/internal/standalone/client.go
```diff
@@ -0,0 +1,384 @@
...
+	existing.Data[entry.Key] = string(configFile.Body)
+	if updateErr := c.k8sClient.Update(context.Background(), existing); updateErr != nil {
+		return fmt.Errorf("failed to update ConfigMap %s/%s: %w", namespace, entry.Name, updateErr)
+	}
+	c.log.Info("Updated ConfigMap key", "name", entry.Name, "namespace", namespace, "key", entry.Key)
+
+	if err := c.triggerRollout(context.Background(), namespace, workloadType, workloadName); err != nil {
+		return fmt.Errorf("failed to trigger rollout for %s/%s: %w", namespace, workloadName, err)
+	}
+	return nil
```
jaronoff97, line 304: i do worry about the security / permissioning here, do we do some type of check on startup to ensure that this will work and we are doing so correctly?

cmd/operator-opamp-bridge/README.md
```diff
@@ -20,6 +20,64 @@ There are two main ways to install the OpAMP Bridge:
...
+capabilities:
+  AcceptsRemoteConfig: true
+  ReportsEffectiveConfig: true
+  ReportsRemoteConfig: true
+standalone:
+  agents:
+    - workloadName: my-collector
+      namespace: default
+      type: otel-collector
+      workloadType: deployment
```
jaronoff97: IMO we should be using a label selector here, something more deliberate and allows the user to specify in a more familiar syntax.

jaronoff97 (commented):
looking a lot better, just two main comments about structure / configuration, otherwise great.

cmd/operator-opamp-bridge/internal/standalone/client.go
owais: If we don't have the right permissions, this op will just fail and log to console about missing permissions so users can go and fix them. I'm not sure if pro-active permission checks add a ton of value but happy to do it.  

I guess it might be helpful in environments where remote ops are very rare and users might not immediately be able to see errors? 

I'll add it in the next push.

cmd/operator-opamp-bridge/README.md
owais: That is what I started with but I realized it was not going to be a good fit for what we're doing. We already have quite a bit of config options which will be too heavy for labels+annotations. We'll have more config options going forward (e.g, specifying custom attributes or config merge behavior etc). Ownership is another issue. If we have multiple bridges deployed, they'll either need to coordinate with each other or we'll need additional config heuristics to determine who owns what.

cmd/operator-opamp-bridge/internal/operator/client.go
owais: OpAMP server doesn't care about k8s semantics. It just needs a configmap key per connected agent. For standalone mode, we don't need to expose these semantics to the server. The server sees an agent like `otel-collector` and then a config file with name `collector-config` (determined by bridge configuration). So the interface doesn't assume this anymore and let's the implementation choose the key. Operator uses ns/name as that is what makes most sense to operator mode. Standalone uses user configured name for the config file.

cmd/operator-opamp-bridge/internal/standalone/client.go
owais: image


we now check for permissions on startup and error out. I added a permission checker implementation that can be re-used. I think we'll need to improve on this sub-system and figure out a way to ensure perm checks remain up to date. We can do that under a separate ticket.

cmd/operator-opamp-bridge/internal/operator/client.go
jaronoff97: yes but this is the operator client which _does_ care about k8s semantics. A caller should do the work to ensure the keys are well formed.
```
func (c Client) Apply(key string, configmap *protobufs.AgentConfigFile) error {
	resource, err := kubeResourceFromKey(key)
	if err != nil {
		return err
	}
	name, namespace := resource.name, resource.namespace
```

is a leaky abstraction. we should either take in the kube resource type here or take in the name and namespace.

cmd/operator-opamp-bridge/README.md
jaronoff97: can you explain more about this? I'm still not convinced this shouldn't be a label selector. You could have the same multiple bridges deployed issue with the existing approach? As you mentioned, we could solve this by stamping each collector instance with a label on it to determine who owns the modifications.

This is one of the _many_ reasons I was initially opposed to this proposal, "need to coordinate with each other" is precisely why we have the operator and why I am still hesitant here. IMO if someone is in this mode, we **must** disallow HA bridges at least for now.

jaronoff97: If you feel that we should stick with this, however, we should follow Argo Rollouts' example here ([reference](https://github.com/argoproj/argo-rollouts/blob/017676bf7317840547f659147967ce2b97154d92/examples/workload-ref/workload-ref.yaml#L29-L32))

cmd/operator-opamp-bridge/internal/operator/client.go
owais: The only use for key here is to name the configmap file in opamp server. It has nothing else to do with k8s. Operator mode overloads the opamp configmap file key with additional responsibility of mapping the remote config file with a local CRD to avoid having an in memory state. There is no reason for the opamp config file to be named `<namespace>/<crd_name>` or `<namespace>/<configmap>`. They are already scoped under the registered agent. 

Example UI from an opamp server:

* Agent details page has info about the agent and its status
image

* Agent config page lists all configs for the agent. it makes sense to have a logical config file name here instead of `<namespace>/<k8s_configmap_name>`. opamp server shouldn't have to deal with that. 
image

The reason operator mode does this is to have a mapping by convention which is fine but standalone does not need that. Even if we added it to standalone, we'd need to add `<namespace>/<k8s_configmap>/<config_file>` as each k8s configmap can contain multiple files. Slapping an entire configmap in there results in worse UX and inconsistent config to that reported by individual collectors themselves.

cmd/operator-opamp-bridge/README.md
owais: Thanks. That looks much better.

owais: > As you mentioned, we could solve this by stamping each collector instance with a label on it to determine who owns the modifications.

yeah but that adds yet one more config field. We could use labels/annotations. It'd just bloat them. We'd have too many annotations/labels and they'd grow in the future as we add more features. I feel argo like config makes more sense and is simpler for us and end users both.

cmd/operator-opamp-bridge/main.go
```diff
@@ -22,18 +23,33 @@ func main() {
...
+		<-signalCtx.Done()
+		standaloneManager.Shutdown()
+		return
+	}
 
 	opampClient := cfg.CreateClient()
+	applier := operator.NewClient(cfg.Name, l.WithName("operator-client"), kubeClient, cfg.GetComponentsAllowed())
 	opampProxy := proxy.NewOpAMPProxy(l.WithName("server"), cfg.ListenAddr)
-	opampAgent := agent.NewAgent(l.WithName("agent"), operatorClient, cfg, opampClient, opampProxy)
+	opampAgent := agent.NewAgent(l.WithName("agent"), applier, cfg, opampClient, opampProxy)
```
jaronoff97: @owais i just realized why i was having a problem with the key conversation we were having above, and i think this is the core of it. The standalone manager should work the same as the `operator.NewClient` no? We may need to make other parts of the code more generic, but ultimately these other components should continue working correctly.

owais: I'm not sure if I completely understand but yes, standalone client has the same overall flow as the operator client. They both take some k8s objects and expose them as an opamp agent to the server with health and config. Depending on the nature of the agent (k8s CRD vs otel config), they may label things like opamp config files differently (determined by key) but on k8s side they follow the exact same flow and are driven by the same logic. Each of them understands how to map opamp config file name to the local k8s resource they are managing.

owais: Both the clients work exactly the same way at a higher level. 

- They monitor k8s resources and register them as agents (CRD, workload).
- They export k8s objects as opamp config files (CR, configmap).
- When opamp remote updates the config files, they update the local k8s counterpart.

The difference is what the opamp config file on the server is named. Operator names it `<namespace>/<cr_name>`. Standalone names it the name given to it in config which is usually the same as the name the config data has in the otel k8s configmap. This is what the `key` controls.

jaronoff97: yes, so why don't we make the standalone just a different kind of applier and make the rest of the mechanisms more generic? i.e. the opamp client accepts the config based on its configured mode and propagates that and we swap out the applier which knows how to handle the 'standalone' mode.

owais: That's pretty much what we're doing. standalone and operator have their own config applier and they're used by a common opamp client. Naming a config file in opamp as `<k8s_namespace>/<name>` isn't ideal for the server but if you feel strongly about it, I think we can use that naming `namespace/name` in both modes.

jaronoff97: I don't believe that's what you're doing here, were that the case we would be using the same NewAgent method and simply swapping the applier rather than shortcircuiting on the mode. We should be able to just bring the applier interface up a level and then have an operator applier and a standalone applier, i don't love that we are duplicating the manager logic and forking the path in this way.

owais: I've updated to have a common manager run both appliers. We now instantiate either config applier and supporting objects conditionally and then pass them into the exact same manager. Now we don't have any duplication. LMK what you think.

cmd/operator-opamp-bridge/internal/config/config.go
```diff
@@ -101,13 +112,38 @@ type AgentDescription struct {
...
 
+type StandaloneConfig struct {
+	Agents []StandaloneAgentConfig `yaml:"agents,omitempty"`
+}
+
+type StandaloneAgentConfig struct {
+	Namespace   string                           `yaml:"namespace"`
+	Type        string                           `yaml:"type"`
+	WorkloadRef StandaloneWorkloadRef            `yaml:"workloadRef"`
+	Config      map[string]StandaloneConfigEntry `yaml:"config"`
```
jaronoff97, line 123: is this like setting a default config for the collector? If this is managed by an opamp server why do we need this?

jaronoff97: oh i see what this does now. Given Kind for the standalone entry can only be configmap right now, should we omit it or make an enum to prevent with validation problems?

cmd/operator-opamp-bridge/internal/config/config.go
```diff
@@ -300,17 +463,80 @@ func LoadFromCLI(target *Config, flagSet *pflag.FlagSet) error {
...
+	if err := validateNoDuplicateStandaloneConfigKeys(envExpandedYaml); err != nil {
+		return err
+	}
+	if err := yaml.Unmarshal(envExpandedYaml, cfg); err != nil {
 		return fmt.Errorf("error unmarshaling YAML: %w", err)
 	}
 	return nil
 }
+
+func validateNoDuplicateStandaloneConfigKeys(data []byte) error {
```
jaronoff97: what does this method do exactly? This just validates someone doesn't have two keys for the same entry? I'm not even sure this is possible / I thought the YAML library would give you an error? https://www.besthub.dev/articles/why-go-s-yaml-v3-fails-to-convert-yaml-with-duplicate-keys-to-json-and-how-to-fix-it-15ca6eaa178c

cmd/operator-opamp-bridge/internal/manager/manager.go
```diff
@@ -0,0 +1,126 @@
...
+	healthServer            *healthcheck.Server
+	opampProxy              *proxy.OpAMPProxy
+	kubernetesClient        KubernetesClient
+	permissionReviewClient  PermissionReviewClient
+	listRequiredPermissions func() ([]Permission, error)
+	cancelKubernetesClient  context.CancelFunc
+	shutdownOnce            sync.Once
+}
+
+func New(opts ManagerOpts) *Manager {
```
jaronoff97: this is not the convention for the option pattern, either use direct arguments here or refactor to use the proper options pattern please.

cmd/operator-opamp-bridge/internal/operator/client.go
```diff
@@ -65,15 +63,20 @@ func NewClient(name string, log logr.Logger, c client.Client, componentsAllowed
 	}
 }
 
-func (c Client) Apply(name, namespace string, configmap *protobufs.AgentConfigFile) error {
+func (c Client) Apply(key string, configmap *protobufs.AgentConfigFile) error {
+	resource, err := kubeResourceFromKey(key)
+	if err != nil {
+		return err
+	}
+	name, namespace := resource.name, resource.namespace
```
jaronoff97, line 75: It's fine this is changing here, but we should document on the interface the expectation for this kube resource key here.

cmd/operator-opamp-bridge/internal/standalone/manager_opts.go
```diff
@@ -0,0 +1,51 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	opampagent "github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/internal/agent"
+	"github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/internal/config"
+	"github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/internal/healthcheck"
+	bridgemanager "github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/internal/manager"
+	"github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/internal/proxy"
+)
+
+func NewManagerOpts(log logr.Logger, cfg *config.Config, c client.Client, restCfg *rest.Config) bridgemanager.ManagerOpts {
```
jaronoff97: Same as the other manager_opts file.

cmd/operator-opamp-bridge/internal/operatorbridge/manager_opts.go
```diff
@@ -0,0 +1,50 @@
...
+
+	opampagent "github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/internal/agent"
+	bridgeconfig "github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/internal/config"
+	"github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/internal/healthcheck"
+	bridgemanager "github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/internal/manager"
+	"github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/internal/operator"
+	"github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/internal/proxy"
+)
+
+func NewManagerOpts(log logr.Logger, cfg *bridgeconfig.Config, c client.Client, _ *rest.Config) bridgemanager.ManagerOpts {
```
jaronoff97: We should move to properly use the options pattern, it's also unclear why this needs to be its own file as opposed to just a function in main?

cmd/operator-opamp-bridge/internal/standalone/validate.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1beta1"
+)
+
+func validateCollectorConfigEntry(body string) error {
+	if strings.TrimSpace(body) == "" {
+		return errors.New("config value is empty")
+	}
+
+	var cfg v1beta1.Config
```
jaronoff97, line 25: given we're using the unmarshal here, i think there's a validate function in the operator we should be able to use here?

cmd/operator-opamp-bridge/main.go
```diff
@@ -22,34 +22,28 @@ func main() {
...
-		l.Error(err, "Cannot start OpAMP client")
-		os.Exit(1)
+	newManagerOpts := operatorbridge.NewManagerOpts
+	if cfg.IsStandaloneMode() {
+		newManagerOpts = standalone.NewManagerOpts
 	}
-	if err := opampProxy.Start(); err != nil {
-		l.Error(err, "failed to start OpAMP Server")
+	opts := newManagerOpts(l, cfg, kubeClient, cfg.GetRestConfig())
+	manager := bridgemanager.New(opts)
```
jaronoff97: this is bad practice for dependency injection. Required dependencies should be arguments and optionals should be provided with the options pattern.

cmd/operator-opamp-bridge/internal/config/config.go
```diff
@@ -91,6 +97,11 @@ type Config struct {
 	HeartbeatInterval time.Duration       `yaml:"heartbeatInterval,omitempty"`
 	Name              string              `yaml:"name,omitempty"`
 	AgentDescription  AgentDescription    `yaml:"description,omitempty"`
+	Standalone        StandaloneConfig    `yaml:"standalone,omitempty"`
```
jaronoff97: we also need to update the opampbridge CRD to pass these.

jaronoff97 (commented):
Looking much better, a few comments and questions but the shape of this is much better.

cmd/operator-opamp-bridge/internal/config/config.go
owais: We could but it'd be a breaking change if we ever add support for anything other than configmaps. The whole field would need to be changed from enum to struct so this is more future proof even if it might feel a bit redundant right now. We do have validation checks that help the user if they try to use anything other than a configmap.

cmd/operator-opamp-bridge/internal/config/config.go
owais: Thanks. This is using yaml v2 which silently lets the later one win. Will migrate to v3 and remove this validator.

owais:
Thanks @jaronoff97 made the requested changes. PTAL and let me know if anything else is needed.

cmd/operator-opamp-bridge/internal/config/config.go
jaronoff97: yes, i don't love the ways we do those validation checks today. Were we to add support for things other than configmaps it would be a new enum value and a new field that would be available for that value.

owais: Done.

cmd/operator-opamp-bridge/internal/standalone/manager_opts.go
```diff
@@ -0,0 +1,51 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	opampagent "github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/internal/agent"
+	"github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/internal/config"
+	"github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/internal/healthcheck"
+	bridgemanager "github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/internal/manager"
+	"github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/internal/proxy"
+)
+
+func NewManagerOptions(log logr.Logger, cfg *config.Config, c client.Client, restCfg *rest.Config) []bridgemanager.Option {
```
jaronoff97: It's good that we're using the options pattern now, but this is still an anti-pattern, all of this should be done in the main.go not in a separate function like this.

cmd/operator-opamp-bridge/internal/config/config.go
```diff
@@ -176,18 +225,124 @@ func (c *Config) GetDescription() *protobufs.AgentDescription {
...
-			keyValuePair("host.name", hostname),
-		),
+		NonIdentifyingAttributes: c.AgentDescription.nonIdentifyingAttributes(map[string]string{
+			"os.family": runtime.GOOS,
+			"host.name": hostname,
+		}),
+	}
+}
+
+func (c *Config) ForStandaloneAgent(agent StandaloneAgentConfig) *Config {
```
jaronoff97: Doing this type of shadowing and needing to do custom cloning like this is very odd. We should instead make a different constructor that makes a new instance of a config rather than a shadow.

cmd/operator-opamp-bridge/internal/manager/manager.go
```diff
@@ -0,0 +1,155 @@
...
+	return func(m *Manager) {
+		m.listRequiredPermissions = listRequiredPermissions
+	}
+}
+
+func New(options ...Option) *Manager {
+	manager := &Manager{}
+	for _, option := range options {
+		option(manager)
+	}
```
jaronoff97, line 90: if there are required fields, we should check that in the constructor and eagerly error rather than waiting until start.

cmd/operator-opamp-bridge/internal/standalone/validate.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	if len(cfg.Service.Pipelines) == 0 {
+		return errors.New("collector config must define at least one service pipeline")
+	}
+
+	for pipelineName, pipeline := range cfg.Service.Pipelines {
+		if pipeline == nil {
+			return fmt.Errorf("pipeline %s is empty", pipelineName)
+		}
+	}
```
jaronoff97, line 76: now that we use the webhook's validate method here, we can remove this, no?

jaronoff97 (commented):
looking much much better only a couple more conventional go things.

cmd/operator-opamp-bridge/internal/standalone/manager_opts.go
owais: The options created are specific to each module so each module holds the knowledge of how to initialize it. I didnt' want to pollute main with it but I'll move it there.

owais:
Thanks @jaronoff97. Appreciate the quick review. PTAL when you can.

cmd/operator-opamp-bridge/internal/config/config.go
```diff
@@ -257,6 +272,26 @@ func cloneStringMap(in map[string]string) map[string]string {
...
+		out[key] = append([]string(nil), values...)
+	}
+	return out
+}
+
+func cloneCapabilityMap(in map[Capability]bool) map[Capability]bool {
+	if in == nil {
+		return nil
+	}
+	out := make(map[Capability]bool, len(in))
```
jaronoff97: I think we don't need this method and can instead just use maps.Copy here because all we need here is a shallow copy?

jaronoff97 (approved):
only a single minor comment, but otherwise this is good. thank you very much for enduring the long back and forth here – i'm happy with where we've ended up here. I'm going to ask for one more maintainer to review before we merge.

cmd/operator-opamp-bridge/internal/config/config.go
owais: It wouldn't copy slices within the structs but we probably don't need to be that defensive. These values shouldn't change ever after initialization anyway. Will simplify.

jaronoff97: this is only for the capability -> bool map, not the slice copy

owais:
@jaronoff97  Thanks for all the time you invested and for the thorough review. I have a few improvements/simplifications lined up that I think you'll like.

tests/e2e-opampbridge/standalone/00-assert.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: apps/v1
```
swiatekm, line 1: If we're adding these tests, we should run them as part of the CI as well. Alternatively, extract them out and add in a follow-up.

cmd/operator-opamp-bridge/internal/agent/agent.go
```diff
@@ -95,124 +93,74 @@ func (agent *Agent) getHealth() *protobufs.ComponentHealth {
...
 }
 
-// generateCollectorPoolHealth allows the bridge to report the status of the collector pools it owns.
-// TODO: implement enhanced health messaging using the collector's new healthcheck extension:
-// https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/26661
-func (agent *Agent) generateCollectorPoolHealth() (map[string]*protobufs.ComponentHealth, error) {
-	cols, err := agent.applier.ListInstances()
-	if err != nil {
-		return nil, err
+func (agent *Agent) componentHealthFromInternal(health operator.Health, statusTime uint64, root bool) (*protobufs.ComponentHealth, error) {
```
swiatekm, line 113: For this and other functions added by this PR, it would help to have short docstrings explaining what they do and what the parameters mean. It makes review more difficult to need to infer all of this.

swiatekm (commented):
The logic and content of this change makes sense to me. I'd like to try to give it a more narrow scope, so it's not +4k lines mixing different concerns.

- Could we move the CRD modifications out into a follow-up? It's not even clear to me if those are valuable, considering the whole point is to decouple from the operator.
- We're not running the E2E tests in CI. Either they should be run, or they should be removed from this PR and added in a follow-up.
- I'd rather move the manifests in `config/` out as well. These are something like a public API in the sense that users might just `kubectl -k` them, and so should require more careful review.
- On a different note, I think this PR could use more docstrings for the functions it adds. It would make it easier to review, if nothing else.

Also @dmitryax, you're the author of the issue, could you have a look and see if this does what you'd envisioned?

tests/e2e-opampbridge/standalone/00-assert.yaml
owais: We do run them in CI. You can find them under `e2e-opampbridge`

cmd/operator-opamp-bridge/internal/agent/agent.go
owais: Added doc strings to new functions that were not self-explanatory.

owais:
> I'd rather move the manifests in config/ out as well. These are something like a public API in the sense that users might just kubectl -k them, and so should require more careful review.

@swiatekm They're very useful during development so I've moved them to a new `dev/` dir for now. I can circle back with another PR after this is merged to add public version of these to config like we have for other things.

owais:
> Could we move the CRD modifications out into a follow-up? It's not even clear to me if those are valuable, considering the whole point is to decouple from the operator.

I've removed them for now. You're right they're not very useful right now.  I think we might need them in the future. I'll add them back in another PR if needed/requested.

owais:
@swiatekm PTAL when you can. Thanks.

tests/e2e-opampbridge/standalone/00-assert.yaml
swiatekm: Ah, my bad, I was looking for where we were installing it with the new make target. I think this is what the test actually should do, maybe with some post-install modifications.

swiatekm:
> > I'd rather move the manifests in config/ out as well. These are something like a public API in the sense that users might just kubectl -k them, and so should require more careful review.
> 
> @swiatekm They're very useful during development so I've moved them to a new `dev/` dir for now. I can circle back with another PR after this is merged to add public version of these to config like we have for other things.

How about moving them to cmd/operator-opamp-bridge/manifests/standalone and adding a small README.md saying that there are dev manifests not intended for production usage?

owais:
@swiatekm moved to cmd and added readme

cmd/operator-opamp-bridge/internal/healthcheck/server_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestServerStartsHealthEndpoint(t *testing.T) {
+	server := NewServer(logr.Discard(), "127.0.0.1:0")
+
+	require.NoError(t, server.Start(t.Context()))
+	t.Cleanup(func() {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second)
```
swiatekm:
```suggestion
		ctx, cancel := context.WithTimeout(t.Context(), time.Second)
```

cmd/operator-opamp-bridge/internal/manager/manager.go
```diff
@@ -0,0 +1,172 @@
...
+
+func (m *Manager) Shutdown() {
+	m.shutdownOnce.Do(func() {
+		if m.cancelKubernetesClient != nil {
+			m.cancelKubernetesClient()
+		}
+		for _, runtime := range m.runtimes {
+			runtime.OpAMPAgent.Shutdown()
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
swiatekm: `Shutdown` should take a context as an argument and pass it in here instead of using `context.Background()`.

swiatekm (commented):
I don't like the copious use of context.Background, but this is something the codebase unfortunately already does, so I'm not going to ask for it to be fixed. I highlighted two spots where it should be fixed in this PR, but we should do a follow-up where we thread context through all the methods that need it @jaronoff97.

owais:
@swiatekm updated the ctx usage. thanks.

swiatekm:
@owais I'll be happy to approve after you fix the conflict.

owais:
@swiatekm resolved.

dmitryax (approved):
LGTM. Thanks @owais!

dmitryax:
The PR description needs to be updated. Looks like we don't support `app.kubernetes.io/managed-by: opamp-bridge-standalone` in this PR (and probably don't need to)

swiatekm (approved):
LGTM

swiatekm:
@jaronoff97 you did the majority of the review for this PR and are the opamp bridge expert, so I'm leaving it to you to merge.

PR #4997, opened 2026-04-24
feat(auto-instrumentation): add securityContext to language specs

### LINKED ISSUE #4894: Instrumentation CRD v1alpha1 missing securityContext on language specs — breaks restricted PSA
## What happened?

The `v1alpha1` Instrumentation CRD schema does not include `securityContext` on the language-specific specs (e.g. `spec.java.securityContext`). This means there is no way to explicitly set the security context on auto-instrumentation init containers via the Instrumentation CR.

## Relationship with #4848

This issue is related to but separate from #4848:

- **#4848** (regression in v0.146.0+): The operator stopped *inheriting* the security context from the app container to the init container. This is a bug with a clear fix (reorder the function calls in `sdk.go`).
- **This issue** (schema gap): Even when #4848 is fixed and inheritance works, there is no way to *explicitly* configure the init container security context via the CR. Users must rely on the implicit inheritance behavior, which is undocumented and fragile.

With the operator pinned to pre-v0.146.0, the implicit inheritance works as a workaround. But the CRD should support explicit configuration for:
1. Deployments where the app container security context differs from what the init container needs
2. GitOps workflows where the desired state should be fully declarative in the CR
3. Validation tooling that checks CRD manifests against the schema

## Steps to reproduce

1. Create a namespace with `pod-security.kubernetes.io/enforce: restricted`
2. Apply an Instrumentation CR with `spec.java.securityContext` set:
   ```yaml
   spec:
     java:
       securityContext:
         runAsNonRoot: true
         allowPrivilegeEscalation: false
         capabilities:
           drop: ["ALL"]
         seccompProfile:
           type: RuntimeDefault
   ```
3. `kubectl apply` fails: `strict decoding error: unknown field "spec.java.securityContext"`
4. `kubectl apply --server-side` also fails: `field not declared in schema`

## Expected behavior

The `securityContext` field should be included in the Instrumentation CRD schema for all language specs, so that the init container security context can be set explicitly.

## Environment

- Operator Helm chart version: `0.108.0` (latest)
- Operator image: `0.145.0` (pinned to avoid #4848)
- Kubernetes: 1.30
- Pod Security Admission: `restricted` (enforce)

## Workaround

Pin the operator to pre-v0.146.0. The init container security context is inherited from the app container automatically. This breaks in v0.146.0+ due to #4848.

Sanil2108:
## Summary

Fixes #4894 — adds an optional \`securityContext\` field to every language-specific section of the Instrumentation CRD (Java, NodeJS, Python, DotNet, Go, ApacheHttpd, Nginx). Setting it explicitly configures the security context of the auto-instrumentation init containers for that language (or the eBPF sidecar, for Go).

This unblocks restricted Pod Security Admission environments where the app container's security context isn't a valid init-container security context, and it makes the CR fully declarative for GitOps workflows that can't rely on the implicit inheritance from the app container.

## Behavior

- **Field unset (default):** existing behavior preserved.
  - Init containers (Java, NodeJS, Python, DotNet, Apache, Nginx) continue to inherit the security context of the first application container being instrumented.
  - The Go sidecar continues to use its hardcoded defaults (\`Privileged: true\`, \`RunAsUser: 0\`) required for eBPF uprobe attachment.
- **Field set:** the provided \`*corev1.SecurityContext\` is applied directly to the language's init container(s) / Go sidecar, overriding the inherited / default value.

A small helper \`resolveInitContainerSecurityContext\` in \`internal/instrumentation/sdk.go\` centralizes the precedence logic (spec wins when non-nil, else inherit).

## Relationship to #4848 / #4849

This is the schema-gap piece described in the issue — separate from the #4848 regression that #4849 addresses. This PR does not change how the operator inherits the security context; it just adds an explicit way to override it. The two land well together.

## What changed

- \`apis/v1alpha1/instrumentation_types.go\` — \`SecurityContext *corev1.SecurityContext\` added to \`Java\`, \`NodeJS\`, \`Python\`, \`DotNet\`, \`Go\`, \`ApacheHttpd\`, \`Nginx\`.
- \`apis/v1alpha1/zz_generated.deepcopy.go\` — regenerated.
- \`config/crd/bases/opentelemetry.io_instrumentations.yaml\` — regenerated.
- \`internal/instrumentation/sdk.go\` — new helper + wired into Java, NodeJS, Python, DotNet, Apache init-container security context setting.
- \`internal/instrumentation/nginx.go\` — clone + main init containers respect the new spec.
- \`internal/instrumentation/golang.go\` — Go sidecar honors \`spec.go.securityContext\` when set.
- \`internal/instrumentation/sdk_securitycontext_test.go\` — unit tests for the precedence helper.
- \`.chloggen/4894-instrumentation-securitycontext.yaml\` — changelog entry.

## Testing

- \`make generate manifests\` — clean, no drift.
- \`go build ./...\` — green.
- \`KUBEBUILDER_ASSETS=... go test ./internal/instrumentation/ -count=1\` — green.
- New \`TestResolveInitContainerSecurityContext\` covers the four precedence cases (spec wins, nil falls back, both nil, container nil).

Fixes #4894


swiatekm:
The change looks good to me with one caveat: We're duplicating the struct eight times in the actual CRD, leading to 500 additional lines in the rendered yaml. Could we not get away with defining this once, at the top level, and using it for all instrumentation containers? Maybe with the exception of Go, since that needs different permissions. The rest of them fundamentally do the same thing, though.

Sanil2108:
Thanks for the review, @swiatekm — good call on the duplication. Pushed a follow-up that consolidates the field to spec-level:

- New: `spec.initContainerSecurityContext` applies to all six init-container languages (Java, NodeJS, Python, DotNet, Apache HTTPD, Nginx).
- Kept: `spec.go.securityContext` stays on the `Go` struct — its sidecar has different requirements (eBPF needs `Privileged: true`, `RunAsUser: 0`) and isn't an init container, so sharing the name would be misleading.
- CRD yaml delta: dropped ~504 lines from `opentelemetry.io_instrumentations.yaml`.

Chloggen + deepcopy + CRD manifests regenerated, unit tests still green. Let me know if you'd rather see `SecurityContext` at the top level instead (without the `InitContainer` prefix) — happy to adjust the field name.

swiatekm:
@Sanil2108 you need to run `make update` to ensure all the generated files are correct.

Sanil2108:
Done — pushed `make update` output. The bundle manifests and `docs/api/instrumentations.md` now reflect the new `spec.initContainerSecurityContext` / `spec.go.securityContext` fields.

swiatekm (approved):
Looks good to me, thank you for the contribution!

pavolloffay (approved):

PR #5000, opened 2026-04-25
Ability to customize namespaces for watching namespaces for pod/service monitors

### LINKED ISSUE #4999: Ability to customize namespaces for watching namespaces for pod/service monitors
### Component(s)

target allocator

Currently the promoperator watches for secrets only in the [namespace as the collector](https://github.com/open-telemetry/opentelemetry-operator/blob/71218d5080c5873650d5ebc8981f6ff17ed0d9f5/cmd/otel-allocator/internal/watcher/promOperator.go#L70). We need an ability to customize the namespaces to watch secrets.

Provide an ability in the [targetallocator's configuration](https://github.com/open-telemetry/opentelemetry-operator/blob/71218d5080c5873650d5ebc8981f6ff17ed0d9f5/cmd/otel-allocator/internal/config/config.go#L77) to be able to customize the namespaces to watch the secrets for,

_No response_

### Additional context

_No response_

### Tip

rashmichandrashekar:
**Description:** 
Currently the promoperator watches for secrets only in the [namespace as the collector](https://github.com/open-telemetry/opentelemetry-operator/blob/71218d5080c5873650d5ebc8981f6ff17ed0d9f5/cmd/otel-allocator/internal/watcher/promOperator.go#L70). We need an ability to customize the namespaces to watch secrets. This PR adds that ability

**Link to tracking Issue(s):**  #4999 

- Resolves: #4999 
**Testing:** unit tests and e2e tests

**Documentation:** added relevant documentation


swiatekm:
From what I can see, prometheus-operator just uses the generic allowList for this. Can we not do that as well?

rashmichandrashekar:
> From what I can see, prometheus-operator just uses the generic allowList for this. Can we not do that as well?

Thanks @swiatekm . The problem with that approach is that, the basic auth config watch by the secret informer will be for all the namespaces that the pod/service monitors are being watched for. By default if nothing is configured in the allow/deny list, it defaults to watch all namespaces, this is problem for secrets because we don't want to provide permissions for the targetallocator to watch all namespaces by default instead it should be a conscious opt in, because of security concerns. I also see issues created in prometheus-operator repository for the default secret watch permissions which makes it insecure - 

https://github.com/prometheus-operator/prometheus-operator/issues/6302
https://github.com/prometheus-operator/prometheus-operator/issues/8355#issuecomment-4067442170
https://github.com/prometheus-operator/prometheus-operator/issues/3544

apis/v1beta1/targetallocator_types.go
```diff
@@ -18,6 +18,10 @@ type TargetAllocatorPrometheusCR struct {
 	// DenyNamespaces Namespaces to scope the interaction of the Target Allocator and the apiserver (deny list). This is mutually exclusive with AllowNamespaces.
 	// +optional
 	DenyNamespaces []string `json:"denyNamespaces,omitempty"`
+	// SecretsAccessNamespaces Namespaces to scope the watching of secrets for the Target Allocator.
+	// If not configured, no namespaces are watched for secrets.
+	// +optional
+	SecretsAccessNamespaces []string `json:"secretsAccessNamespaces,omitempty"`
```
swiatekm: I think I'd prefer just `SecretNamespaces`. Also, shouldn't the default value be the namespace where the target allocator is running? Otherwise this is a breaking change.

swiatekm (commented):
Ok, I can see how adding a new attribute is justified here. I still have a comment about the specific implementation.

apis/v1beta1/targetallocator_types.go
rashmichandrashekar: Thanks @swiatekm. Fixed.
I think providing access to the target allocator's namespace by default is also a security risk. For instance if it is running in kube-system namespace, there may be secrets and having get, list and watch on all of them might pose a security risk. 
Another thing to note is that, this was only added in the last month so there might not be a lot of users who have adapted it already. But I agree this is a breaking change and needs to be documented.

rashmichandrashekar: @swiatekm - gentle ping on this pr. Thanks!

swiatekm: @rashmichandrashekar I don't think it's true that this was added last month. Your own [PR](https://github.com/open-telemetry/opentelemetry-operator/pull/4608) fixing the updates is from January, and before that the auth support was still functional for immutable secrets with the feature gate turned on. So the breaking change will probably affect everyone using that gate. Let's not change this behavior in this PR, we can discuss it in a separate issue.

rashmichandrashekar: Thanks @swiatekm. The PR was created in January but it was only released in [March](https://github.com/open-telemetry/opentelemetry-operator/releases/tag/v0.147.0). Either way I agree its a breaking change. If needed I update this PR to keep existing behavior and add ability to update namespace and create a separate issue to track the namespace update as a security concern.

swiatekm:
@rashmichandrashekar I think the cla failure is due to copilot being a coauthor on https://github.com/open-telemetry/opentelemetry-operator/pull/5000/changes/84f47d801a524dcb32227a01120427093bdddbe9.

rashmichandrashekar:
> @rashmichandrashekar I think the cla failure is due to copilot being a coauthor on [84f47d8](https://github.com/open-telemetry/opentelemetry-operator/commit/84f47d801a524dcb32227a01120427093bdddbe9).

Thanks @swiatekm. Fixed, pls take a look.

swiatekm (approved):
LGTM

Issue #5010, opened 2026-04-29
Separate webhooks to subcommand / allow webhook HA


pavolloffay:
### Component(s)

auto-instrumentation

### Is your feature request related to a problem? Please describe.

I would like to ensure high availability for the instrumentation webhook. Right now the webhook is part of the operator controller deployment with a single replica and `failurePolicy: Ignore` (if the webhook is not reachable the start continues).

HA is required to ensure all annotated pods in the cluster are instrumented. The HA should handle operator upgrades and cluster disruptions (node drain, failures)


The instrumentation uses a pod mutation webhook deployed as part of the operator controller. 
The Kubernetes controllers are usually deployed as singletons (1 replica). If multiple instances are deployed, the leader election is used to select the master. 

The webhooks can be deployed and scaled separately. The API server calls the webhook via a service that loadbalances requests to all instances. 


### Describe the solution you'd like

I would like to be able to configure/opt-in for the instrumentation webhook HA.
 
I should be able to:
*  scale webhook to 2 and more replicas
 
The operator should automatically:
* define PodDisruptionBudget for the webhook/operator deployment
* define PodAntiAffinity for the webhook/operator deployment
* define PriorityClasses for the webhook/operator deployment



### Describe alternatives you've considered

_No response_

### Additional context

## How other projects handle HA for webhooks

### Cert manager

```yaml
 apiVersion: operator.openshift.io/v1alpha1                                                                                                                                                                                                                                                             
  kind: CertManager
  metadata:
    name: cluster                                                                                                                                                                                                                                                                                        
  spec:                                                                                                                                                                                                                                                                                                  
    webhookConfig:                                                                                                                                                                                                                                                                                       
      overrideReplicas: 3                                                                                                                                                                                                                                                                               
```

Cert manager operator creates webhook object and deployment https://github.com/openshift/cert-manager-operator/blob/master/bundle/manifests/cert-manager-operator.clusterserviceversion.yaml#L461 

### Istio 

```yaml
apiVersion: sailoperator.io/v1
  kind: Istio                                                                                                                                                                                                                                                                                            
  metadata:                                                                                                                                                                                                                                                                                              
    name: default
  spec:                                                                                                                                                                                                                                                                                                  
    namespace: istio-system                                                                                                                                                                                                                                                                              
    values:                
      pilot:                                                                                                                                                                                                                                                                                             
        autoscaleEnabled: false
        replicaCount: 2
```

### OpenShift COO operator
[...]

PR #5012, opened 2026-04-29
Make operator HA on OpenShift

pavolloffay:

- Updates #5010

### DIFF
--- a/bundle/openshift/manifests/opentelemetry-operator.clusterserviceversion.yaml
+++ b/bundle/openshift/manifests/opentelemetry-operator.clusterserviceversion.yaml
@@ -99,7 +99,7 @@ metadata:
     categories: Logging & Tracing,Monitoring,Observability
     certified: "false"
     containerImage: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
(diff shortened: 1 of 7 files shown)


iblancasa (approved):

swiatekm (approved):

pavolloffay:
Do not merge this yet!

ps-mir (approved):

bundle/openshift/manifests/opentelemetry-operator.clusterserviceversion.yaml
```diff
@@ -540,18 +540,32 @@ spec:
           control-plane: controller-manager
         name: opentelemetry-operator-controller-manager
         spec:
-          replicas: 1
+          replicas: 2
```
frzifus, line 543: How would that look like on a Single Node Openshift setting the anti affinity?

ps-mir: I believe it should still schedule 2 replicas due to the soft affinity rule. 

I'm curious if a single-node setup will block the node drain due to PDB, but if we are talking about draining a single-node setup, is this even a valid concern?

PR #5022, opened 2026-05-01
chore(rbac): split kubebuilder RBAC markers and remove unused delete permissions

### LINKED ISSUE #3156: Restrict Cluster Role access authorizations
### Component(s)

_No response_

**Context**
Our current setup uses the OpenTelemetry Operator to make the application traceable. The operator is deployed through a [Helm Chart](https://open-telemetry.github.io/opentelemetry-helm-charts). However, our Trivy scanner identifies that the operator has broad permissions via the Kubernetes ClusterRole.

**Revise**
Based on my understanding, the OpenTelemetry Operator's current permissions allow it to delete various Kubernetes resources like pods, services, and service accounts. This level of access seems unnecessary for the operator's intended functionality.

The RBAC are generated via go maker comments. The most relevant ones are on the `OpenTelemetryCollectorReconciler` struct in the Reconcile function.

**Suggestion**
- For more granular access, define for each required Kubernetes kind access via `+kubebuilder:rbac` and do not group them in a single go comment marker.
- Remove delete access

**Hint**
As these are cluster roles, this applies to all namespaces.

**Version**
Helm Chart v0.58.2

Sanil2108:
## Summary

Closes #3156.

The operator's ClusterRole granted broad `create;update;patch;delete` permissions on `pods`, but the operator never directly manages Pod objects — pods are created and owned by `Deployment`, `DaemonSet`, and `StatefulSet` controllers. This PR splits the grouped RBAC markers so each resource has only the verbs it actually needs.

**Changes:**

- Split `pods;configmaps;services;serviceaccounts` (which had full write permissions) into two separate annotations:
  - `pods`: `get;list;watch` only (no create/update/patch/delete)
  - `configmaps;serviceaccounts;services`: unchanged full write permissions (these are directly managed and pruned by the operator)
- Reduced `targetallocators/finalizers` from `get;list;watch;create;update;patch;delete` down to `update` — finalizer subresources only need `update` to add/remove the finalizer field
- Regenerated `config/rbac/role.yaml` via `make manifests`

**Rationale:**

The `reconcileDesiredObjects` and `deleteObjects` functions in `internal/controllers/common.go` call `client.Delete` to prune owned objects, so `delete` is legitimately needed for `configmaps`, `services`, `serviceaccounts`, and other directly-managed resources. However, `pods` never appear in `GetOwnedResourceTypes()` and are never passed to `CreateOrUpdate` or `Delete` — the pod mutation webhook operates on admission request objects, not via a direct API client call with write permissions.

## Test plan

- [ ] `make manifests` produces the updated `config/rbac/role.yaml` with reduced pod permissions
- [ ] Existing unit tests pass (`go test ./internal/controllers/...`)
- [ ] Manual verification: operator reconciles collectors in all modes (Deployment, DaemonSet, StatefulSet) without permission errors

### DIFF
--- a/.chloggen/split-rbac-markers.yaml
+++ b/.chloggen/split-rbac-markers.yaml
@@ -1,0 +1,19 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Split kubebuilder RBAC markers and restrict pod permissions to get/list/watch only
+
+# One or more tracking issues related to the change
+issues: [3156]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The operator does not create, update, patch, or delete Pod objects directly;
+  pods are managed by Deployment, DaemonSet, and StatefulSet controllers.
+  Also reduced targetallocators/finalizers to only the `update` verb.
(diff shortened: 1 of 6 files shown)


swiatekm:
Looks like a reasonable change to me. You'll need to run `make update` and commit updates to the manifests to make the CI happy.

swiatekm (approved):

swiatekm:
@open-telemetry/operator-maintainers I'd love a second review before we merge, in case these bits are here for some obscure reason.

atoulme (approved):
checks out for me, approved.

iblancasa (approved):

PR #2703, opened 2024-03-04
Reconcile v1beta1 collector CRD

pavolloffay:
**Description:** <Describe what has changed.>

Do not merge this. the PR will be split into smaller PRs. 

**Link to tracking Issue:** <Issue number if applicable>
Resolves: #2620 

This can be merged after https://github.com/open-telemetry/opentelemetry-operator/issues/2708 

Follow ups:

- https://github.com/open-telemetry/opentelemetry-operator/issues/2736
- https://github.com/open-telemetry/opentelemetry-operator/issues/2727
- https://github.com/open-telemetry/opentelemetry-operator/issues/2737
- https://github.com/open-telemetry/opentelemetry-operator/issues/2746

**Testing:** <Describe what testing was performed and which tests were added.>

**Documentation:** <Describe the documentation added.>

PR #5023, opened 2026-05-01
feat(api): restore CRD field descriptions for operator-owned types

### LINKED ISSUE #2737: Add CRD descriptions
### Component(s)

collector, auto-instrumentation, target allocator, opamp

The https://github.com/open-telemetry/opentelemetry-operator/pull/2703/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R43  removed CRD descriptions otherwise `kubectl apply -f crd-file.yaml` would fail due to large size. 

## Proposal
Bring back CRD descriptions at least for structs that come from the OTEL operator.

Sanil2108:
## Summary

Restores CRD field descriptions that were removed in #2703 to keep CRD sizes manageable.

- Sets `maxDescLen=128` (previously `maxDescLen=0`) in the `controller-gen` CRD generation options, restoring truncated but meaningful descriptions for all operator-defined fields
- Updates the `make install` target to use `kubectl apply --server-side` to support the larger CRD payloads

## Background

PR #2703 set `maxDescLen=0` because `kubectl apply -f crd-file.yaml` was failing due to CRD size. This removed all field descriptions from the generated CRDs, making `kubectl explain` and API documentation less useful.

With `maxDescLen=128`, the largest CRD (OpenTelemetryCollectors) grows from ~400KB to ~900KB, which is within etcd's object size limits. Using `--server-side` apply avoids the `last-applied-configuration` annotation that doubles the payload for client-side apply.

## CRD sizes with this change

| CRD | Before (maxDescLen=0) | After (maxDescLen=128) |
|-----|----------------------|------------------------|
| opentelemetrycollectors | 405KB | 900KB |
| targetallocators | 153KB | 346KB |
| instrumentations | 92KB | 185KB |
| opampbridges | 73KB | 166KB |
| clusterobservabilities | 5KB | 11KB |

## Example

Before:
```yaml
configVersions:
  default: 3
  minimum: 1
  type: integer
```

After:
```yaml
configVersions:
  default: 3
  description: ConfigVersions defines the number versions to keep for
    the collector config. Each config version is stored in a separate ConfigMap.
  minimum: 1
  type: integer
```

Resolves: #2737

## Test plan

- [x] `make manifests` regenerates CRDs with descriptions present
- [x] Verified `kubectl explain` shows field descriptions for operator-owned types
- [x] CRD sizes verified to be within etcd object size limits

### DIFF
--- a/.chloggen/restore-crd-field-descriptions.yaml
+++ b/.chloggen/restore-crd-field-descriptions.yaml
@@ -1,0 +1,19 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Restore CRD field descriptions for operator-owned types by setting maxDescLen=128.
+
+# One or more tracking issues related to the change
+issues: [2737]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The `make install` target now uses `kubectl apply --server-side` to support the larger
+  CRD payloads that include field descriptions. This allows `kubectl explain` and similar
+  tools to show useful documentation for operator-defined fields.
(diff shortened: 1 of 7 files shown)


iblancasa (changes requested):
I'm not against enabling this again but only if we have some kind of test that probes the operator is not broken because of this.

swiatekm:
I'm not sure we want to do this. Or at least not until we drop support for v1alpha1 Collectors.

PR #5024, opened 2026-05-01
feat(collector): skip service provisioning for sidecar mode

Sanil2108:
**Description:** In sidecar mode, the operator does not control the Pod lifecycle — the collector container is injected into the user's workload Pod by a mutating webhook. Provisioning Services, Ingress, NetworkPolicies, and HorizontalPodAutoscalers for sidecar-mode collectors is therefore unnecessary and creates confusing, non-functional resources in the namespace.

This change moves the following manifest factories behind a `mode != sidecar` guard, so they are only built when the operator actually manages the Pod (Deployment, StatefulSet, DaemonSet modes):

- `Service` (base)
- `HeadlessService`
- `MonitoringService`
- `ExtensionService`
- `HorizontalPodAutoscaler`
- `Ingress`
- `NetworkPolicy`

`PodMonitor` continues to be provisioned in sidecar mode when `spec.observability.metrics.enableMetrics` is set, because PodMonitors select Pods by label and work regardless of who owns the Pod. `ConfigMap`, `ServiceAccount`, and `TargetAllocator` are also still provisioned unconditionally.

- Resolves: #4934

**Testing:** Updated the existing sidecar-mode unit test to expect 2 objects (ConfigMap + ServiceAccount) instead of 3, and added a new test case verifying that a PodMonitor is still built in sidecar mode when metrics are enabled.

**Documentation:** No documentation changes required; the behavior change follows naturally from the sidecar semantics already described in the docs.

### DIFF
--- a/.chloggen/skip-services-sidecar-mode.yaml
+++ b/.chloggen/skip-services-sidecar-mode.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: collector
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Skip provisioning Services, Ingress, NetworkPolicy, and HPA for sidecar mode collectors since the operator does not control the Pod lifecycle in that mode. PodMonitors are still provisioned when metrics are enabled.
+
+# One or more tracking issues related to the change
+issues: [4934]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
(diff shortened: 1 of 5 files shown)


Sanil2108:
Fixed the resource ordering issue — TargetAllocator is now appended after the services block to preserve the original slice order.

The Govulncheck failures (GO-2026-4887, GO-2026-4883) are pre-existing in the repo and also fail on main — not introduced by this PR.

swiatekm (commented):
The changes look good to me. Can you also add some assertions to an existing e2e test that these resources are not created?

Sanil2108:
Added e2e assertions in the latest commit — both `smoke-sidecar` and `smoke-sidecar-native` tests now include an `error:` step that verifies the base Service, headless Service, and monitoring Service are not created after the collector is installed in sidecar mode. @swiatekm could you take another look when you get a chance?

swiatekm (approved):

swiatekm:
I'll merge after the next release, thank you for the contribution!

iblancasa (approved):

PR #5071, opened 2026-05-12
[chore] Pin collector image in filelog receiver test

swiatekm:
The current `main` fails due to a change we need to handle in 0.152.0. Pinning to 0.150.1 until I figure out how to make it align with the operator's default.

Also temporarily skip the 3.9 Python test to unblock CI, see https://github.com/open-telemetry/opentelemetry-operator/issues/5072.

### DIFF
--- a/tests/e2e-instrumentation/instrumentation-python/chainsaw-test.yaml
+++ b/tests/e2e-instrumentation/instrumentation-python/chainsaw-test.yaml
@@ -48,6 +48,7 @@ kind: Test
 metadata:
   name: instrumentation-python-oldest
 spec:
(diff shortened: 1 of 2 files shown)


iblancasa:
> The current `main` fails due to a change we need to handle in 0.152.0. Pinning to 0.150.1 until I figure out how to make it align with the operator's default.

What is that change?

swiatekm:
> > The current `main` fails due to a change we need to handle in 0.152.0. Pinning to 0.150.1 until I figure out how to make it align with the operator's default.
> 
> What is that change?

https://github.com/open-telemetry/opentelemetry-operator/issues/5026#issuecomment-4391753635

iblancasa (approved):

Issue #5060, opened 2026-05-08
Stability and Production Readiness of Instrumentation CRD (v1alpha1)


Ranganatha-D:
## 🧩 Background
We are currently evaluating the `Instrumentation` Custom Resource Definition (CRD) for managing auto-instrumentation across our clusters. We noticed the current API version is `v1alpha1`.

## ❓ Questions
I would like to get confirmation from the maintainers regarding the following:

1.  **Production Readiness:** Is the `v1alpha1` version recommended for production environments?
2.  **API Stability:** Are there significant breaking changes planned for the `Instrumentation` spec in the near future?
3.  **Promotion Roadmap:** Is there a timeline or a tracking issue for promoting this CRD to `v1beta1` or `v1`?

## 🛠️ Environment Details
*   **OpenTelemetry Operator version:** [latest]
*   **Kubernetes version:** [ v1.33]
*   **Install Method:** [Helm chart]

## Additional Context
We are planning a large-scale rollout and want to ensure we are aligned with the project's long-term stability goals. Any guidance would be greatly appreciated!

iblancasa:
There is some work in progress for a new version
- https://github.com/open-telemetry/opentelemetry-operator/issues/4093
- https://github.com/open-telemetry/opentelemetry-operator/pull/4583

dayananda30:
@iblancasa, Does that mean it is not production-ready yet?

swiatekm:
It is production ready, and will be maintained past the v1beta1 release. In principle, being at alpha allows us to do breaking changes to the CRD, but that hasn't actually happened in a long time - we're cognizant that it has a lot of users. The changes we're planning in v1beta1 will allow us to improve the webhook performance and eventually standardize on the Otel declarative configuration, but they won't break any of the existing functionality. @pavolloffay can you also chime in?

pavolloffay:
I have just submitted RFC for v1bta1 https://github.com/open-telemetry/opentelemetry-operator/pull/5079. As @swiatekm the v1alpha1 will continue to function and the new API will allow us introduce new features and clean the API.

PR #5079, opened 2026-05-13
Add RFC: Instrumentation v1beta1

pavolloffay:

This PR adds RFC for Instrumenation v1beta1

Related to #5060

Explanation fo managing multiple CRD versions: https://gist.github.com/pavolloffay/e8bf0ad2c42cd94e82f2354a3559894f 

### DIFF
--- a/docs/rfcs/instrumentation-v1beta1.md
+++ b/docs/rfcs/instrumentation-v1beta1.md
@@ -1,0 +1,406 @@
+# Instrumentation v1beta1
+
+This document outlines the next version of the Instrumentation CRD - `v1beta1`.
+
+## Motivation
+
+The current `v1alpha1` Instrumentation CRD has been widely adopted in production environments despite its `v1alpha1` version. Promoting to `v1beta1` signals API stability and provides an opportunity to address accumulated design issues. The `v1beta1` version should be a breaking change from `v1alpha1` that:
+
+1. Aligns with OpenTelemetry's [declarative configuration](https://github.com/open-telemetry/opentelemetry-configuration) initiative
+2. Fixes structural inconsistencies in the current API
+
+## Objectives
+
+1. Support strongly-typed [declarative configuration](https://github.com/open-telemetry/opentelemetry-configuration) in `spec.declarativeConfig` alongside existing env-var-based configuration in `spec.envConfig` ([#4093](https://github.com/open-telemetry/opentelemetry-operator/issues/4093))
+2. Add explicit OTLP exporter protocol field to avoid ambiguity between HTTP and gRPC endpoints ([#3658](https://github.com/open-telemetry/opentelemetry-operator/issues/3658))
+3. Normalize per-language resource fields — remove deprecated [`volumeLimitSize`](https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1alpha1/instrumentation_types.go#L178) and unify inconsistent JSON tags ([`json:"resources"`](https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1alpha1/instrumentation_types.go#L188) vs [`json:"resourceRequirements"`](https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1alpha1/instrumentation_types.go#L228))
+4. Consolidate `Resource` and `Defaults` into a single top-level `spec.resource` field for operator-level resource attribute configuration
+
+## Non-Goals (for initial v1beta1)
+
+* Label selectors for targeting workloads ([#2744](https://github.com/open-telemetry/opentelemetry-operator/issues/2744), [#821](https://github.com/open-telemetry/opentelemetry-operator/issues/821)) - additive feature, can be added later without breaking changes
+* Webhook architecture separation ([#5010](https://github.com/open-telemetry/opentelemetry-operator/issues/5010), [#4115](https://github.com/open-telemetry/opentelemetry-operator/issues/4115)) - operational concern, not CRD spec
+* Windows node support ([#642](https://github.com/open-telemetry/opentelemetry-operator/issues/642)) - can be added without breaking changes
+* New language support - can be added incrementally
+
+## Proposed Changes
+
+### 1. SDK Declarative Configuration
+
+**Issues:** [#4093](https://github.com/open-telemetry/opentelemetry-operator/issues/4093), [#4607](https://github.com/open-telemetry/opentelemetry-operator/issues/4607)
+
+OpenTelemetry is standardizing on [file-based declarative configuration](https://opentelemetry.io/docs/specs/otel/configuration/) as the preferred way to configure SDKs. The v1beta1 CRD supports two mutually exclusive configuration approaches:
+
+- **`spec.declarativeConfig`** — strongly-typed Go structs matching the [OTel SDK configuration schema](https://github.com/open-telemetry/opentelemetry-configuration). The operator serializes this to a YAML file, mounts it into the workload, and sets `OTEL_CONFIG_FILE`.
+- **`spec.envConfig`** — the existing env-var-based configuration (`exporter`, `sampler`, `propagators`, `resource`), moved under a dedicated field. This preserves the current v1alpha1 behavior.
+
+Setting both `declarativeConfig` and `envConfig` is invalid and rejected by the webhook.
+
+#### Declarative config example
+
+```yaml
+apiVersion: opentelemetry.io/v1beta1
+kind: Instrumentation
+metadata:
+  name: declarative-example
+spec:
+  declarativeConfig:
+    file_format: "0.4"
+    resource:
+      attributes:
+        - name: service.namespace
+          value: production
+    tracer_provider:
+      sampler:
+        parent_based:
+          root:
+            trace_id_ratio_based:
+              ratio: 0.25
+      processors:
+        - batch: {}
+      exporters:
+        - otlp:
+            endpoint: http://collector:4318
+            protocol: http/protobuf
+```
+
+#### Environment variable substitution in declarative config
+
+The declarative config supports [environment variable substitution](https://opentelemetry.io/docs/specs/otel/configuration/file-configuration/#environment-variable-substitution) using the `${VAR}` syntax. This is useful for injecting secrets like API tokens without hardcoding them in the CR. Environment variables can be set via `spec.env` or per-language `env` fields.
+
+This also addresses a limitation in v1alpha1 where Kubernetes `$(VAR)` substitution fails due to env var ordering ([#3022](https://github.com/open-telemetry/opentelemetry-operator/issues/3022)). Since `${VAR}` substitution in declarative config happens at SDK runtime rather than Kubernetes pod creation time, it works regardless of the order in which env vars are defined.
+
+```yaml
+apiVersion: opentelemetry.io/v1beta1
+kind: Instrumentation
+metadata:
+  name: declarative-with-secret
+spec:
+  env:
+    - name: OTEL_EXPORTER_API_KEY
+      valueFrom:
+        secretKeyRef:
+          name: otel-secrets
+          key: api-key
+  declarativeConfig:
+    file_format: "0.4"
+    tracer_provider:
+      processors:
+        - batch: {}
+      exporters:
+        - otlp:
+            endpoint: https://otlp.example.com:4318
+            protocol: http/protobuf
+            headers:
+              - name: x-api-key
+                value: ${OTEL_EXPORTER_API_KEY}
+```
+
+#### Env-var config example (current behavior)
+
+```yaml
+apiVersion: opentelemetry.io/v1beta1
+kind: Instrumentation
+metadata:
+  name: env-example
+spec:
+  envConfig:
+    exporter:
+      endpoint: http://collector:4318
+      protocol: http/protobuf
+    sampler:
+      type: parentbased_traceidratio
+      argument: "0.25"
+    propagators:
+      - tracecontext
+      - baggage
+  resource:
+    attributes:
+      service.namespace: production
+```
+
+### 2. Explicit Exporter Protocol
+
+**Issues:** [#3658](https://github.com/open-telemetry/opentelemetry-operator/issues/3658)
+
+The v1alpha1 `spec.exporter` struct has a single `endpoint` field with no indication of whether it expects HTTP or gRPC. This is a common source of confusion because different SDK auto-instrumentation images default to different protocols:
+
+| Language | Default OTLP Protocol | Default Port | Operator Override |
+|----------|----------------------|--------------|-------------------|
+| Java | `http/protobuf` | 4318 | No (SDK default since Java agent 2.x) |
+| NodeJS | `http/protobuf` | 4318 | No (SDK default) |
+| Python | `grpc` | 4317 | Yes — operator forces `http/protobuf` (port 4318) |
+| DotNet | `http/protobuf` | 4318 | No (auto-instrumentation default; differs from .NET SDK which defaults to `grpc`) |
+| Go | `http/protobuf` | 4318 | No (auto-instrumentation default; Go SDK itself defaults to `grpc`) |
+| Apache HTTPD | `grpc` | 4317 | No (otel-webserver-module only supports gRPC; [proposal to add HTTP](https://github.com/open-telemetry/opentelemetry-cpp-contrib/issues/614)) |
+| Nginx | `grpc` | 4317 | No (otel-webserver-module only supports gRPC; [proposal to add HTTP](https://github.com/open-telemetry/opentelemetry-cpp-contrib/issues/614)) |
+
+The v1beta1 adds an explicit `protocol` field to the `Exporter` struct. When set, the operator injects `OTEL_EXPORTER_OTLP_PROTOCOL` alongside the endpoint. Valid values are `grpc`, `http/protobuf`, and `http/json`.
+
+```go
+type Exporter struct {
+    Endpoint string `json:"endpoint,omitempty"`
+    Protocol string `json:"protocol,omitempty"`
+    TLS      *TLS   `json:"tls,omitempty"`
+}
+```
+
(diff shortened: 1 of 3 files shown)


docs/rfcs/instrumentation-v1beta1.md
```diff
@@ -0,0 +1,406 @@
...
+2. Add explicit OTLP exporter protocol field to avoid ambiguity between HTTP and gRPC endpoints ([#3658](https://github.com/open-telemetry/opentelemetry-operator/issues/3658))
+3. Normalize per-language resource fields — remove deprecated [`volumeLimitSize`](https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1alpha1/instrumentation_types.go#L178) and unify inconsistent JSON tags ([`json:"resources"`](https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1alpha1/instrumentation_types.go#L188) vs [`json:"resourceRequirements"`](https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1alpha1/instrumentation_types.go#L228))
+4. Consolidate `Resource` and `Defaults` into a single top-level `spec.resource` field for operator-level resource attribute configuration
+
+## Non-Goals (for initial v1beta1)
+
+* Label selectors for targeting workloads ([#2744](https://github.com/open-telemetry/opentelemetry-operator/issues/2744), [#821](https://github.com/open-telemetry/opentelemetry-operator/issues/821)) - additive feature, can be added later without breaking changes
```
swiatekm: I think this is a breaking change in a practical sense. The whole point of it is that we want to scope the mutating webhook by label. We can only do that if we are confident users are using the label to indicate instrumented resources, rather than the annotation. It's a breaking change to our public API, though not to the CRD.

docs/rfcs/instrumentation-v1beta1.md
```diff
@@ -0,0 +1,406 @@
...
+### 1. SDK Declarative Configuration
+
+**Issues:** [#4093](https://github.com/open-telemetry/opentelemetry-operator/issues/4093), [#4607](https://github.com/open-telemetry/opentelemetry-operator/issues/4607)
+
+OpenTelemetry is standardizing on [file-based declarative configuration](https://opentelemetry.io/docs/specs/otel/configuration/) as the preferred way to configure SDKs. The v1beta1 CRD supports two mutually exclusive configuration approaches:
+
+- **`spec.declarativeConfig`** — strongly-typed Go structs matching the [OTel SDK configuration schema](https://github.com/open-telemetry/opentelemetry-configuration). The operator serializes this to a YAML file, mounts it into the workload, and sets `OTEL_CONFIG_FILE`.
+- **`spec.envConfig`** — the existing env-var-based configuration (`exporter`, `sampler`, `propagators`, `resource`), moved under a dedicated field. This preserves the current v1alpha1 behavior.
+
+Setting both `declarativeConfig` and `envConfig` is invalid and rejected by the webhook.
```
swiatekm, line 39: How do we handle differences in SDK support for the declarative configuration? I think we need to keep a list of supported languages, probably overridable through config for easier testing. How will we update this list?

docs/rfcs/instrumentation-v1beta1.md
```diff
@@ -0,0 +1,406 @@
...
+  resource:
+    attributes:
+      service.namespace: production
+```
+
+### 2. Explicit Exporter Protocol
+
+**Issues:** [#3658](https://github.com/open-telemetry/opentelemetry-operator/issues/3658)
+
+The v1alpha1 `spec.exporter` struct has a single `endpoint` field with no indication of whether it expects HTTP or gRPC. This is a common source of confusion because different SDK auto-instrumentation images default to different protocols:
```
swiatekm, line 142: This is just for `envConfig`, right?

docs/rfcs/instrumentation-v1beta1.md
```diff
@@ -0,0 +1,406 @@
...
+| Change | v1alpha1 | v1beta1 | Migration |
+|--------|----------|---------|-----------|
+| SDK configuration | `exporter`, `sampler`, `propagators`, `resource` at top level | `exporter`, `sampler`, `propagators` moved under `spec.envConfig`, or use new `spec.declarativeConfig` | Wrap existing fields under `envConfig`, or migrate to declarative config |
+| Resource attributes | `spec.resource.resourceAttributes`, `spec.resource.addK8sUIDAttributes`, `spec.defaults.useLabelsForResourceAttributes` | Consolidated into `spec.resource.attributes`, `spec.resource.k8sMetadata`, and `spec.resource.serviceMetadata` | Rename `resourceAttributes` to `attributes`, use `k8sMetadata` for k8s.* attributes, use `serviceMetadata` for service.* derivation |
+| Per-language resources JSON tag | Mixed (`resources` / `resourceRequirements`) | `resources` (all) | Rename in YAML for NodeJS, Python, DotNet, Go, ApacheHttpd, Nginx |
+| Volume size limit | `volumeLimitSize` (deprecated) | Removed | Use `volumeClaimTemplate` |
+
+## Migration Strategy
+
+1. **Conversion webhook**: Implement a conversion webhook that translates between v1alpha1 and v1beta1, handling field renames and removals automatically.
```
swiatekm, line 489: Let's be sure to make this optional. It was a massive pain to rely on it for backwards compatibility during the Collector CR migration. The way I imagine this is that supporting v1alpha1 will involve internally converting to v1beta1, and the conversion webhook will simply use the same method.

swiatekm (commented):
Looks good to me overall. I have a major question involving the annotation -> label change, and some minor ones.

docs/rfcs/instrumentation-v1beta1.md
pavolloffay: The SDK support is documented here https://github.com/open-telemetry/opentelemetry-configuration/blob/main/language-support-status.md 

The language support is even more complicated because it's split per config feature. Given the large number of features I am not sure what validation we can perform in the operator. 

I have added a section with supported languages. Only .net does not support it at all.

docs/rfcs/instrumentation-v1beta1.md
pavolloffay: yes. The declarative config encodes protocol by the exporter type:

```
 The protocol is determined by the exporter type:
  - otlp_http ? HTTP transport (with encoding: protobuf or json)
  - otlp_grpc ? gRPC transport 

  tracer_provider:
    exporters:
      - otlp_http:
          endpoint: http://collector:4318
          encoding: protobuf

```

docs/rfcs/instrumentation-v1beta1.md
pavolloffay: Do you mean to make the conversion webhook optional?

swiatekm: Yes. If users want it, it should be available, but we should also support v1alpha directly.

docs/rfcs/instrumentation-v1beta1.md
swiatekm: So are we planning to just let users figure out what is supported and what isn't? I wouldn't mind returning an admission error if someone tries to use declarative config with .net, for example. I'm not sure we're able to do much else without overloading ourselves with compatibility checks.

docs/rfcs/instrumentation-v1beta1.md
swiatekm: Correction, what I wanted to do is actually impossible. It is possible to do all the breaking changes in v1alpha1, but I don't think that's an option for us given how widely it's used. So we're left with the webhook, the only question is whether we do the same thing we did for v1beta1 Collector, or if we implement some kind of CRD self-patching logic.

docs/rfcs/instrumentation-v1beta1.md
pavolloffay: The operator will perform some checks during the injection. This is what I have in the doc:

> The operator will **skip injection** for languages that do not support declarative configuration when `spec.declarativeConfig` is set. For example, using `declarativeConfig` with `.NET` auto-instrumentation will result in the pod mutation webhook skipping injection and emitting a warning event.

docs/rfcs/instrumentation-v1beta1.md
pavolloffay: Added the label as a breaking change and objective for v1beta1

docs/rfcs/instrumentation-v1beta1.md
swiatekm: I'm personally ok with a conversion webhook, possibly live-patched at runtime by the operator.

docs/rfcs/instrumentation-v1beta1.md
```diff
@@ -216,6 +232,79 @@ The operator handles this differently depending on the active config mode:
...
+| Present | Not set | Label controls injection (v1beta1 approach) |
+| Present | Present | Label takes precedence |
+| Not set | Not set | No injection |
+
+This allows gradual migration: users can switch to labels at their own pace while existing annotation-based configurations continue to work.
+
+#### Deprecation path
+
+1. **v1beta1**: Both labels and annotations supported. Annotations deprecated but functional.
+2. **v1alpha1 removal**: Annotations no longer documented, labels become the primary mechanism. Annotations support is eventually dropped.
```
swiatekm: This isn't fully clear to me. Are we planning to support annotations until we go to v1? Or do they go away with v1alpha1?

docs/rfcs/instrumentation-v1beta1.md
```diff
@@ -216,6 +232,79 @@ The operator handles this differently depending on the active config mode:
...
+    objectSelector:
+      matchExpressions:
+        - key: instrumentation.opentelemetry.io/inject-java
+          operator: Exists
+        - key: instrumentation.opentelemetry.io/inject-python
+          operator: Exists
+        # ... other languages
+```
+
+This reduces webhook invocations to only pods that have opted in via labels, improving cluster performance. Pods using legacy annotations will still pass through the webhook (no `objectSelector` match), maintaining backwards compatibility.
```
swiatekm: This only works if we're not supporting annotations, right? So it'll have to be an option in either the kustomize manifests or the Helm Chart.

swiatekm (commented):
LGTM overall, some questions about the annotation -> label migration.

docs/rfcs/instrumentation-v1beta1.md
pavolloffay: yes it can be only enforced once we don't support annotations. I would give it as opt-in before we fully drop support for the annotation.

docs/rfcs/instrumentation-v1beta1.md
pavolloffay: I have clarified it

> 2. **v1alpha1 removal**: Labels become the primary mechanism. Annotations support is dropped.

docs/rfcs/instrumentation-v1beta1.md
pavolloffay: Added some clarification ot the readme.

swiatekm (approved):

docs/rfcs/instrumentation-v1beta1.md
```diff
@@ -0,0 +1,504 @@
...
+See also: [CRD Version Graduation Strategies](crd-version-graduation-strategies.md) for background on CRD versioning approaches.
+
+## Motivation
+
+The current `v1alpha1` Instrumentation CRD has been widely adopted in production environments despite its `v1alpha1` version. Promoting to `v1beta1` signals API stability and provides an opportunity to address accumulated design issues. The `v1beta1` version should be a breaking change from `v1alpha1` that:
+
+1. Aligns with OpenTelemetry's [declarative configuration](https://github.com/open-telemetry/opentelemetry-configuration) initiative
+2. Fixes structural inconsistencies in the current API
+
+## Objectives
```
jaronoff97, line 14: One thing i think we should also do is let users target specific otelcol CRDs rather than needing to hand write a svc / url. i.e. someone could target `namespace/otelcol` or something like that and we would validate that the collector exists and then write the url for them.

docs/rfcs/instrumentation-v1beta1.md
```diff
@@ -0,0 +1,504 @@
...
+
+```yaml
+apiVersion: opentelemetry.io/v1beta1
+kind: Instrumentation
+metadata:
+  name: declarative-example
+spec:
+  declarativeConfig:
+    file_format: "0.4"
+    resource:
```
jaronoff97, line 65: I know why it's like this, but i don't love that we mix camel case and snake case.

docs/rfcs/instrumentation-v1beta1.md
```diff
@@ -0,0 +1,504 @@
...
+        - otlp:
+            endpoint: http://collector:4318
+            protocol: http/protobuf
+```
+
+#### Environment variable substitution in declarative config
+
+The declarative config supports [environment variable substitution](https://opentelemetry.io/docs/specs/otel/configuration/file-configuration/#environment-variable-substitution) using the `${VAR}` syntax. This is useful for injecting secrets like API tokens without hardcoding them in the CR. Environment variables can be set via `spec.env` or per-language `env` fields.
+
+This also addresses a limitation in v1alpha1 where Kubernetes `$(VAR)` substitution fails due to env var ordering ([#3022](https://github.com/open-telemetry/opentelemetry-operator/issues/3022)). Since `${VAR}` substitution in declarative config happens at SDK runtime rather than Kubernetes pod creation time, it works regardless of the order in which env vars are defined.
```
jaronoff97, line 87: One thing that I think would be useful here is to add a "presets" section for setting the resource hierarchy like we do today, this would make it easier for users to opt out and then all we do internally is set the declarative config and let that figure out the layering later on.

docs/rfcs/instrumentation-v1beta1.md
```diff
@@ -0,0 +1,504 @@
...
+
+#### Env-var config example (current behavior)
+
+```yaml
+apiVersion: opentelemetry.io/v1beta1
+kind: Instrumentation
+metadata:
+  name: env-example
+spec:
+  envConfig:
```
jaronoff97, line 123: we should note that this will be removed/deprecated when we go to v1

jaronoff97 (approved):
a few small thoughts, otherwise looks great. thank you Pavol! :bow:

docs/rfcs/instrumentation-v1beta1.md
pavolloffay: I don't think that is the goal. We might remove it if the SDKs stop supporting the env config

docs/rfcs/instrumentation-v1beta1.md
mmanciop, line 14: Out of experience with multiple OTel adopters, treating the trace-context propagation format (and, specifically, opt-in to AWS XRay) as a first-class citizen in the CRD, pays off

docs/rfcs/instrumentation-v1beta1.md
```diff
@@ -0,0 +1,504 @@
...
+
+#### Language support
+
+Not all language SDKs support declarative configuration yet. See [language-support-status.md](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/language-support-status.md) for the current status.
+
+- **Java** — supported
+- **Python** — supported ([#4856](https://github.com/open-telemetry/opentelemetry-python/issues/4856))
+- **Node.js** — in progress
+- **Go** — in progress
+- **.NET** — not supported ([#6380](https://github.com/open-telemetry/opentelemetry-dotnet/issues/6380))
```
mmanciop: Int seems in-progress now: https://github.com/open-telemetry/opentelemetry-dotnet/pull/7413

docs/rfcs/instrumentation-v1beta1.md
```diff
@@ -0,0 +1,504 @@
...
+
+The operator will **skip injection** for languages that do not support declarative configuration when `spec.declarativeConfig` is set. For example, using `declarativeConfig` with `.NET` auto-instrumentation will result in the pod mutation webhook skipping injection and emitting a warning event.
+
+The operator will have a flag to control declarative configuration support  (e.g., `--instrumentation-declarative-config=java,python,dotnet,cpp,nodejs`). This allows enabling support as SDKs mature without requiring operator upgrades.
+
+#### Declarative config example
```
mmanciop, line 55: In the system packages SIG, we are having issues with the fact that all implementations of the declarative SDK but Java [seem not to allow per-language overrides](https://github.com/open-telemetry/opentelemetry-packaging/pull/10/changes#r3429403577), e.g. to disable instrumentations. It seems to me you may face the same issue.

pavolloffay: Are you referring to this part of the config https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md#experimentalinstrumentation- ?

docs/rfcs/instrumentation-v1beta1.md
pavolloffay: added

docs/rfcs/instrumentation-v1beta1.md
pavolloffay: We already support in as a first-class citizen in the CRD. The goal is to support as much as possible from the stable config symbols.

docs/rfcs/instrumentation-v1beta1.md
mmanciop: yes

docs/rfcs/instrumentation-v1beta1.md
pavolloffay: The resource attributes in v1beta1 will be handled by:

  - spec.resource.k8sMetadata.enabled — toggle for K8s resource attributes
  - spec.resource.k8sMetadata.includeUIDs — toggle for UID attributes
  - spec.resource.serviceMetadata.enabled — toggle for service identity attributes

They are mutually exclusive. The service name resolution uses the spec https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/#service-attributes

Issue #3821, opened 2025-03-19
Create an integration test checking the output of the managed CR application


atoulme:
### Describe the issue you're reporting

Using chainsaw, create a situation where we run a managed CR deployment, and check the output is what is expected, especially when it comes to metrics and logs expressed.

Issue #3818, opened 2025-03-19
[Managed] Create a best practice managed experience with the operator


atoulme:
### Describe the issue you're reporting

This feature request tracks the creation of a new custom resource which automates the provisioning of collectors, target allocators, auto-instrumentation webhook managers and so on.

This custom resource will set up all the resources to create a cohesive, out of the box experience of OpenTelemetry, with the same feature set as https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-collector.

This is a bit of a change compared to the existing CR in the sense that this is an opinionated CR which will define behaviors. The user configuration exposed on this CR will be minimal, typically just setting the data destination.

iblancasa:
I think it would make sense to create one RFC https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/rfcs/template.md

jaronoff97:
@atoulme do you think should supersede, close, and/or reference some of the prior work (cc @frzifus) #3297 #3279 #1906

atoulme:
This prior work seems intent on fixing specific issues, I see this CR as a user experience where the operator in its totality is simply managed, and therefore doesn't talk about sampling, or exposes the collector configuration.

PR #5080, opened 2026-05-13
[clusterobservability] fix config for bundled collectors

jinja2:
### Collector image
- Generate the bundled collectors with the **K8s Collector distribution** ([`opentelemetry-collector-k8s`](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-k8s)) at the same version as the operator's default collector, exposed as a new `Config.ClusterObservabilityCollectorImage` and overridable via the `--clusterobservability-collector-image` flag. This keeps the version tied to the existing `opentelemetry-collector` entry in [`versions.txt`](https://github.com/open-telemetry/opentelemetry-operator/blob/main/versions.txt).

### Agent collector (DaemonSet)
- Inject `K8S_NODE_NAME` via the downward API so the bundled kubeletstats receiver's `https://${env:K8S_NODE_NAME}:10250` endpoint resolves.
- Mount the whole host root at `/hostfs` (read-only, `HostToContainer`) instead of granular per-directory hostPath mounts, and set the hostmetric receiver's [`root_path`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/hostmetricsreceiver#collecting-host-metrics-from-inside-a-container-linux-only) to `/hostfs`, so the receiver can resolve symlinks and stat every host mount point.
- Exclude virtual / container-runtime filesystems from the `hostmetrics` `filesystem` scraper (`exclude_mount_points` + `exclude_fs_types`) to avoid `/sysroot` and `/boot` scrape errors on RHCOS.
- On OpenShift, run the agent as root (`runAsUser: 0`, with the [`spc_t` SELinux type](https://docs.openshift.com/container-platform/latest/authentication/managing-security-context-constraints.html) supplied by the generated SCC) so the filelog receiver can read the root-owned container logs under `/var/log/pods`. Capabilities stay dropped and privilege escalation disabled.

### Cluster collector
- `checkClusterCollectorStatus` now reads the StatefulSet workload, so the CR no longer stays in `Pending` when the cluster collector is fully healthy (the cluster collector was moved to a StatefulSet in #5139).

### Auto-instrumentation
- Point the generated `Instrumentation` at the agent's OTLP HTTP port (`:4318`); most auto-instrumentations default to `http/protobuf`.

### Refactor
- Move distro detection in the manifests builder to the shared `config.DistroProvider`, and extract agent security contexts / volumes / mounts into distro-gated helper functions.

### Link to tracking Issue(s)
- Create an integration test checking the output of the managed CR application #3821
- [Managed] Create a best practice managed experience with the operator #3818

### Testing
- New `make e2e-clusterobservability` chainsaw target. `make prepare-e2e-clusterobservability` runs `add-rbac-permissions-to-operator` (opt-in kustomize patches) and bakes in `--feature-gates=+operator.clusterobservability` for the e2e job.
- The Chainsaw suite asserts telemetry flow end-to-end: signals from `kubeletstats`, `hostmetrics`, `k8s_cluster`, `filelog`, and `k8s_events` reach the sinks. A Node.js app with auto-instrumentation also produces spans that arrive at the cluster collector (`otelcol_receiver_accepted_spans`).
- The suite checks component restarts and logs, along with refused or dropped telemetry.
- Reusable e2e workflow gets the new group, `experimental: true` with matching `continue-on-error`, so failures don't block other operator e2e jobs.
- Unit tests cover the generated collector configuration and StatefulSet-based status handling.
- Manually validated with a freshly built operator on ROSA (OpenShift): the full Chainsaw suite passed, including container-log collection, all bundled receiver signals, auto-instrumented traces, OpenShift SCC and SELinux settings, kubelet CA verification, component health, and `ClusterObservability` status.
- Manually validated on Talos


internal/manifests/clusterobservability/clusterobservability.go
```diff
@@ -43,22 +43,25 @@ const (
...
-// getCollectorImage returns a sensible default collector image when build-time version is not set.
+// getCollectorImage rewrites the configured collector image to the contrib distribution.
 func getCollectorImage(configuredImage string) string {
 	// If the configured image has a 0.0.0 tag (fallback during development builds)
 	// replace it with latest
 	if configuredImage == "ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector:0.0.0" {
 		return "ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-contrib:latest"
 	}
-	return configuredImage
+	return strings.Replace(configuredImage, "/opentelemetry-collector:", "/opentelemetry-collector-contrib:", 1)
```
pavolloffay: I wonder if we could use the k8s distribution instead?

frzifus: Currently the default images are provided by setting via ldflags and reading the version from the `version.txt` file. 
Right now the default is the core collector in version `0.152.1`. During one of the SIG calls we discussed to  maybe use the k8s distro as default. In that case, I think we dont need to change anything.

If we continue to use core, we may want to add a separate version entry for CollectorCRs generated by the  `ClusterObservability` reconciler.

frzifus (commented):
Awesome, I will test it on talos and openshift during the week!

internal/manifests/clusterobservability/clusterobservability.go
jinja2: Thanks! Switched to the k8s distribution. Instead of string-rewriting the image, the reconciler now reads a dedicated `Config.ClusterObservabilityCollectorImage`, which defaults to `opentelemetry-collector-k8s:<version>` using the existing `opentelemetry-collector` entry in versions.txt, with `--clusterobservability-collector-image` as an override.

I derived the version rather than adding a separate versions.txt entry for now, given the experimental maturity here. That said, I can see the case for an explicit `opentelemetry-collector-k8s` entry - it'd be self-documenting and let maintainers pin/patch the k8s distro independently if the release process ever diverges. Do you have a preference?

jinja2:
Hi @frzifus , I tested ClusterObservability end to end on ROSA/OpenShift 4.22.6 and Talos 1.13.7 with K8s 1.36. Both runs passed, covering all bundled receivers, container logs, events, and auto-instrumented traces. The OpenShift run also verified the generated SCC, SELinux/UID settings, and kubelet CA verification.

Could you please take another look at this? I’d like to follow up with more ClusterObservability features once these fixes and tests land. I’d also appreciate your thoughts on the opentelemetry-collector-k8s image choice. It keeps ClusterObservability functional without changing the default image for existing Collector CRs, but I’m happy to add a separate versions.txt entry if preferred.

frzifus (approved):

PR #5087, opened 2026-05-17
docs: document Node.js module lookup path

### LINKED ISSUE #3653: Add documentation for instrumentation-winston lookup path
### Component(s)

_No response_

I'm using the `opentelemetry-operator` in k8s to auto-instrument and found that `instrumentation-winston` was running but it couldn't load `@opentelemetry/winston-transport` even though it was added to the app's package.json.

The issue is the operator mounts the auto-instrumentation packages at `/otel-auto-instrumentation-nodejs/autoinstrumentation.js` so the [Node module lookup order](https://nodejs.org/api/modules.html#loading-from-node_modules-folders) means it only looks upwards for `node_modules` until `/`.

If you've built your Dockerfile following [a typical tutorial](https://www.digitalocean.com/community/tutorials/how-to-build-a-node-js-application-with-docker#step-3-writing-the-dockerfile) then your `package.json` is installing into `/home/node/app/node_modules` which won't be on the search path during the `require` call.  This can be fixed by adding the env-var:
```
NODE_PATH=/home/node/app/node_modules
```
which can also be done in the `Instrumentation` resource that's injecting into containers.  Suggest adding this somewhere in documentation.

popsiclelmlm:
Fixes #3653

## Summary
- explain when Node.js auto-instrumentation may need `NODE_PATH` to find application dependencies
- add an `Instrumentation.spec.nodejs.env` example for setting `NODE_PATH`

## Verification
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,6 +292,18 @@ NodeJS:
 instrumentation.opentelemetry.io/inject-nodejs: "true"
 ```
 
+Node.js resolves modules from the injected auto-instrumentation script path. If an instrumentation
+package needs to load an application dependency from the application's `node_modules` directory, set
+`NODE_PATH` with the application dependency path in the `Instrumentation` resource:
+
+```yaml
+spec:
+  nodejs:
+    env:
+      - name: NODE_PATH
+        value: /home/node/app/node_modules
+```
+
 Python:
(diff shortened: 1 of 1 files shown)


swiatekm (approved):

swiatekm (approved):

iblancasa (approved):

PR #5088, opened 2026-05-17
feat(target-allocator): add allowInsecureAuthSecrets option

### LINKED ISSUE #3746: ScrapeConfig targetAllocator bad password using basicAuth
### Component(s)

targetAllocator

### What happened?

Hi ;) 

It seems that the Target Allocator is picking the wrong password when using ScrapeConfig from Prometheus Operator (no issue with username in the same secret) 

## Description

The following ScrapeConfig is working fine using Prometheus Operator : 
```
apiVersion: monitoring.coreos.com/v1alpha1
kind: ScrapeConfig
metadata:
  name: demo-scrapeconfig
spec:
  staticConfigs:
    - labels:
        job: demo
      targets:
        - demo-remote-service:9999
  metricsPath: /metrics
  scheme: HTTPS
  tlsConfig:
    insecureSkipVerify: true
  basicAuth:
    username:
      name: basic-auth
      key: user
    password:
      name: basic-auth
      key: password
```

But when using targetAllocator, I get 401 errors in OtelColector logs:

```
2025-02-25T18:41:37.945Z	debug	scrape/scrape.go:1347	Scrape failed	{"kind": "receiver", "name": "prometheus", "data_type": "metrics", "scrape_pool": "scrapeConfig/demo/demo-scrapeconfig", "target": "https://demo-remote-service:9999/metrics", "error": "server returned HTTP status 401 Unauthorized"}
```

Using `curl 'localhost:8080/scrape_configs' | jq` to get the scrape config, I can see that the password is not the one created in the k8s secret

```
  "scrapeConfig/demo/demo-scrapeconfig": {
    "basic_auth": {
      "password": "<secret>",
      "username": "demo-user"
    },
```

Using a remote webhook receiver to troubleshoot the final query, I can confirm that the secret value is not good 

Image

## Steps to Reproduce

1. Create secret `k create secret generic basic-auth-demo --from-literal='user=username-example' --from-literal='password=password-example'`

2. Create a ScrapeConfig
```
apiVersion: monitoring.coreos.com/v1alpha1
kind: ScrapeConfig
metadata:
  name: demo-scrape-config
  labels:
    app: kube-prometheus-stack-prometheus
    release: prometheus-operator
    prometheus: system-monitoring-prometheus
spec:
  staticConfigs:
    - labels:
        job: demo-scrape-config
      targets:
        - webhook.site
  metricsPath: /3ad6310f-79ed-4816-b74c-5616ff68d2a1 #Change here with your endpoint
  scheme: HTTPS
  tlsConfig:
    insecureSkipVerify: true 
  basicAuth:
    username:
      name: basic-auth-demo
      key: user
    password:
      name: basic-auth-demo
      key: password
```

3. Use portforward / curl on TA to get config 
```
curl 'localhost:8080/scrape_configs' | jq
... 
  "scrapeConfig/test-sna/demo-scrape-config": {
    "basic_auth": {
      "password": "<secret>",
      "username": "username-example"
    },
```
4. Check the webhook site 

Image

## Expected Result

The right password:
```
    "basic_auth": {
      "password": "password-example",
      "username": "username-example"
    },
```

## Actual Result
```
    "basic_auth": {
      "password": "<secret>",
      "username": "username-example"
    },
```

### Kubernetes Version

v1.30.9

### Operator version

otel/opentelemetry-collector-k8s:0.117.0

### Collector version

otel/opentelemetry-collector-k8s:0.117.0

## Environment
OS: (e.g., "Ubuntu 20.04")

### Log output

```shell
2025-02-25T18:41:37.945Z	debug	scrape/scrape.go:1347	Scrape failed	{"kind": "receiver", "name": "prometheus", "data_type": "metrics", "scrape_pool": "scrapeConfig/demo/demo-scrapeconfig", "target": "https://demo-remote-service:9999/metrics", "error": "server returned HTTP status 401 Unauthorized"}
```

### Additional context

My OTEL Deployment: 

```
kind: OpenTelemetryCollector
metadata:
  name: collector-with-ta
spec:
  mode: statefulset
  targetAllocator:
    enabled: true
    prometheusCR:
      enabled: true
      serviceMonitorSelector: {}
      scrapeConfigSelector: {}
  config:
    receivers:
      prometheus:
        config:
          scrape_configs:
          - job_name: 'otel-collector'
            scrape_interval: 10s
            static_configs:
            - targets: [ '0.0.0.0:8888' ]
            metric_relabel_configs:
            - action: labeldrop
              regex: (id|name)
            - action: labelmap
              regex: label_(.+)
              replacement: $$1

    exporters:
      debug: {}

    service:
      pipelines:
        metrics:
          receivers: [prometheus]
          exporters: [debug]
      telemetry:
        logs:
          level: "debug"
        metrics:
          address: 0.0.0.0:8888
```

tillwulfram:
**Description:** 
Currently the TA can only serve discovered auth data, eg. Kubernetes Secrets from a ServiceMonitor with basicauth, over mtls. This great feature would overcomplicate setups for teams who have existing service mesh which already handles the mtls auth between the services or high maintenance for teams who can't use the operator / crds and also for dev setups. 
Its configurable bool via config file, CLI flag (`--allow-insecure-auth-secrets`), or environment variable (`ALLOW_INSECURE_AUTH_SECRETS`) - allowing the most flexibility so the operator crds can cover this setting and also the standalone helm chart from the targetAllocator.
-> When enabled, a "warn" log will be written at startup that secret are exposed via plaintext. 
Defaults to `false` to maintain backwards compatibility.

A next feature could be that also https can be configured so no "heavy" mtls is needed, but at least the transport is secured.

- Resolves: #3746

**Testing:**
Expanded the tests for the new feature in the config, flags and the server files.
All tests, precommits and lints were successful.

**Documentation:** 
Added the new config parameter to the targetAllocator and a short description how it can be used, for which environments it should be used and which problem it can solve.

### DIFF
--- a/cmd/otel-allocator/README.md
+++ b/cmd/otel-allocator/README.md
@@ -426,23 +426,23 @@ Prerequisites:
 
 #### Alternative: allow insecure auth secrets
 
-If the connection between the target allocator and collectors is already secured by a service mesh (e.g. Istio, Linkerd) or equivalent transport-level security, you can skip mTLS setup and serve auth secret values over plain HTTP instead.
+If transport security is already handled by a service mesh or equivalent, you can skip the mTLS setup and serve auth secret values over plain HTTP.
 
-Set `allow_insecure_auth_secrets: true` in the target allocator config file, pass `--allow-insecure-auth-secrets` as a CLI flag, or set the `ALLOW_INSECURE_AUTH_SECRETS=true` environment variable.
-
-When using the `OpenTelemetryCollector` CR with an embedded target allocator, set it via the `env` field:
+**With the Operator (CRD):**
 
 ```yaml
 targetAllocator:
   enabled: true
-  env:
-    - name: ALLOW_INSECURE_AUTH_SECRETS
-      value: "true"
+  allowInsecureAuthSecrets: true
 ```
 
-> **Warning:** Only enable this when transport-level security is guaranteed by other means. Without mTLS or a service mesh, auth secrets will be transmitted in plaintext.
+This works on both the `OpenTelemetryCollector` CR (embedded target allocator) and the standalone `TargetAllocator` CR.
 
+**Standalone Target Allocator (without Operator):**
 
+Set `allow_insecure_auth_secrets: true` in the target allocator config file, or set the `ALLOW_INSECURE_AUTH_SECRETS=true` environment variable.
+
+> **Warning:** Only enable this when transport-level security is guaranteed by other means.
 
 # Design
 
--- a/cmd/otel-allocator/README.md
+++ b/cmd/otel-allocator/README.md
@@ -27,6 +27,7 @@ The Target Allocator uses a configuration file (by default under `/conf/targetal
 | `filter_strategy`                  | Filter strategy to apply to metrics                                           | `relabel-config`                              |                      |
 | `prometheus_cr`                    | Whether to watch Prometheus Custom Resources                                  |                                               |                      |
 | `https`                            | Whether to expose the target allocator endpoint over https                    |                                               |                      |
+| `allow_insecure_auth_secrets`      | Serve auth secret values over plain HTTP without mTLS                         | `false`                                       | `ALLOW_INSECURE_AUTH_SECRETS` |
 | `collector_not_ready_grace_period` | Wait time before assigning jobs to a new collector.                           | 30s                                           |                      |
 
 Additional configuration options are present under [./internal/config/config.go](./internal/config/config.go).
@@ -422,6 +423,24 @@ Prerequisites:
         ```
 
 - Enable the `operator.targetallocator.mtls` feature gate in the operator's deployment. 
+
+#### Alternative: allow insecure auth secrets
+
+If the connection between the target allocator and collectors is already secured by a service mesh (e.g. Istio, Linkerd) or equivalent transport-level security, you can skip mTLS setup and serve auth secret values over plain HTTP instead.
+
+Set `allow_insecure_auth_secrets: true` in the target allocator config file, pass `--allow-insecure-auth-secrets` as a CLI flag, or set the `ALLOW_INSECURE_AUTH_SECRETS=true` environment variable.
+
+When using the `OpenTelemetryCollector` CR with an embedded target allocator, set it via the `env` field:
+
+```yaml
+targetAllocator:
+  enabled: true
+  env:
+    - name: ALLOW_INSECURE_AUTH_SECRETS
+      value: "true"
+```
+
+> **Warning:** Only enable this when transport-level security is guaranteed by other means. Without mTLS or a service mesh, auth secrets will be transmitted in plaintext.
 
 
 
--- a/cmd/otel-allocator/internal/config/config_test.go
+++ b/cmd/otel-allocator/internal/config/config_test.go
@@ -794,6 +794,31 @@ func TestLoadFromEnv(t *testing.T) {
 	assert.Equal(t, namespace, cfg.CollectorNamespace)
 }
 
+func TestLoadFromEnvAllowInsecureAuthSecrets(t *testing.T) {
+	t.Run("not set defaults to false", func(t *testing.T) {
+		cfg := &Config{}
+		err := LoadFromEnv(cfg)
+		require.NoError(t, err)
+		assert.False(t, cfg.AllowInsecureAuthSecrets)
+	})
+
+	t.Run("set to true", func(t *testing.T) {
+		t.Setenv("ALLOW_INSECURE_AUTH_SECRETS", "true")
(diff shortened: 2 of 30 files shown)


cmd/otel-allocator/internal/config/config_test.go
```diff
@@ -1086,3 +1111,91 @@ kube_config_file_path: "` + kubeConfigPath + `"
 		assert.True(t, config.PrometheusCR.Enabled, "CLI should override config file for prometheus CR enabled")
 	})
 }
+
+func TestAllowInsecureAuthSecrets(t *testing.T) {
```
swiatekm: Do we really need these tests specifically for this one config value? Consider adding it to an existing test instead.

cmd/otel-allocator/internal/server/server.go
```diff
@@ -121,6 +129,11 @@ func NewServer(log logr.Logger, allocator allocation.Allocator, listenAddr strin
 		opt(s)
 	}
 
+	if s.allowInsecureAuthSecrets {
+		slog.New(logr.ToSlogHandler(s.logger)).Warn("allowInsecureAuthSecrets is enabled - auth secret values will be served over plain HTTP. " +
```
swiatekm: Why are you using slog here, as opposed to using the existing logger directly?

cmd/otel-allocator/README.md
```diff
@@ -423,6 +424,24 @@ Prerequisites:
...
+```yaml
+targetAllocator:
+  enabled: true
+  env:
+    - name: ALLOW_INSECURE_AUTH_SECRETS
+      value: "true"
+```
+
+> **Warning:** Only enable this when transport-level security is guaranteed by other means. Without mTLS or a service mesh, auth secrets will be transmitted in plaintext.
+
```
swiatekm, line 442: Instead of suggesting this, I'd prefer to add a `allowInsecureAuthSecrets` on the TargetAllocator CRD.

swiatekm (commented):
I think we should make this a field on the TargetAllocator CRD rather than force users to go through env variables.

We should also add an e2e test for this functionality, similar to the existing mtls tests.

tillwulfram:
thanks for the feedback @swiatekm! :)
I'll implement the improvements for the tests.
Regarding the other topics I would like to discuss to understand / clarify it. I only implemented `slog` because I wasn't aware of—or hadn't noticed—any native warning level in `logr`, and since `slog` is also used in other files (config, etc.), I based my implementation on that. If the message is supposed to be just an info level and should include "WARNING" or something similar in the message text, I can change that as well.
From what I gathered from our discussion in the previous issue, the CRD adaptation would require too much effort or would be an implementation in a separate pr for such a "niche" topic, and with the three approaches (Env, Args, ConfigMap), we would have covered both the Operator CRDs and also the standalone Helm TA. But if wished I can look into the implementation about the crd field

cmd/otel-allocator/internal/server/server.go
swiatekm: Let's just log this as Info and leave a comment that it should be a warning after we migrate away from logr.

swiatekm:
> From what I gathered from our discussion in the previous issue, the CRD adaptation would require too much effort or would be an implementation in a separate pr for such a "niche" topic, and with the three approaches (Env, Args, ConfigMap), we would have covered both the Operator CRDs and also the standalone Helm TA. But if wished I can look into the implementation about the crd field.

To be clear, I don't mind merging this PR as-is, but we should then have a follow-up that adds the field to the CRDs. We're adding the feature, I don't want to make it annoying for users to adopt. What I'd like to happen in this PR, is for you to remove the documentation section. Whether you want to do the CRD part here, or in a follow-up, is up to you.

cmd/otel-allocator/internal/config/config_test.go
tillwulfram: added to existing test

cmd/otel-allocator/internal/server/server.go
tillwulfram: used logr and add comment

cmd/otel-allocator/README.md
tillwulfram: done

tillwulfram:
Thanks for your input @swiatekm
I addressed your reviews and also included the crd fields for both the embedded ta inside the collector crd and the standalone ta crd. The standalone TA without Operator can just use the parameter via configmap or env.
I thought it would be easier to include the crds in this pr, so that the end-to-end tests, etc., could directly test the preferred CRD fields and you wouldn't have to adjust the code again in the follow-up. 
If something is missing pls let me know. :)

swiatekm (approved):
Looks good to me, thank you for the contribution!

frzifus (approved):

PR #5091, opened 2026-05-18
[Chore] (tests): make OpenShift e2e tests compatible with OLM-managed installs

IshwarKanse:
## Summary

Three e2e tests assumed direct Deployment patching and hardcoded the `opentelemetry-operator-system` namespace, causing failures on OpenShift clusters where OLM manages the operator under the `opentelemetry-operator` namespace.

- **env-config**: OLM can require multiple reconciliation cycles to propagate `Subscription spec.config.env` to the running Deployment (AtLatestKnown state). The test raced ahead after OLM's first (pre-change) rollout cycle, before `LABELS_FILTER` was active. Fix: poll the Deployment spec until `LABELS_FILTER` appears before calling `rollout status`.

- **instrumentation-blocked-upgrade**: Direct Deployment arg patches are reconciled back by OLM within ~30s, so `--auto-instrumentation-dotnet-image` / `--auto-instrumentation-java-image` overrides were immediately reverted, preventing the upgrade-blocking behavior from being tested. Fix: detect OLM installs via CSV presence — on OLM clusters patch the CSV and poll the Deployment spec for propagation; on kind (no CSV) patch the Deployment directly. Also replace the hardcoded `opentelemetry-operator-system` namespace and `podLogs` catch block with dynamic discovery via pod label selectors.

- **tls-profile**: Step-09's `Running+Ready` pod check found the old (about-to-terminate) pod after the step-08 APIServer revert and exited immediately. The subsequent `OtelCollector` apply then hit the webhook during pod shutdown → `connection refused` → `no endpoints available`. Fix: capture old pod UIDs before the delete, detect when a new UID appears (restart begun), wait for `rollout status`, then poll the webhook Service endpoint for ready addresses before proceeding.

## Test plan

- [x] `env-config` verified passing on OpenShift OLM cluster
- [x] `instrumentation-blocked-upgrade` verified passing on OpenShift OLM cluster; upstream kind CI compatibility preserved via CSV-detection branch
- [x] `tls-profile` verified passing on OpenShift OLM cluster (152s, all 11 steps)

### DIFF
--- a/tests/e2e-instrumentation/instrumentation-blocked-upgrade/chainsaw-test.yaml
+++ b/tests/e2e-instrumentation/instrumentation-blocked-upgrade/chainsaw-test.yaml
@@ -24,20 +24,69 @@ spec:
         content: |
           #!/bin/bash
           set -eo pipefail
-          NS=opentelemetry-operator-system
           DEPLOY=opentelemetry-operator-controller-manager
           OLD_DOTNET_IMG="ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-dotnet:1.2.0"
           OLD_JAVA_IMG="ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-java:1.33.6"
 
-          # Add the dotnet and java image args to override the compiled-in defaults
-          ARGS=$(kubectl get deploy -n $NS $DEPLOY -o json | jq -c '.spec.template.spec.containers[0].args')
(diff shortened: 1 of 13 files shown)


iblancasa (approved):

swiatekm (approved):
I don't mind this change, but we could also just use the same install namespace for plain e2e tests as we do for OLM. It's not really important which namespace it is, and it'd save us some annoying bash code. WDYT @IshwarKanse ?

IshwarKanse:
@swiatekm Agreed, I have updated the tests to use opentelemetry-operator-system.

PR #5101, opened 2026-05-19
fix(target-allocator): scrapeClass tlsConfig fields silently dropped

ozzywalsh:
## Summary

The TA's config parser can't deserialize prometheus-operator types that use `json:",inline"` embedded structs (like `TLSConfig`). Fields like `caFile` and `insecureSkipVerify` are present in the ConfigMap but silently dropped when the TA loads them.

The root cause: the TA config pipeline uses `mapstructure` with `TagName:"yaml"` and `yaml.v2` for unmarshaling. But prometheus-operator types embed sub-structs with `json:",inline"` — a tag that `yaml.v2` doesn't understand. So any field that lives behind an inline embedding gets lost.

Fixed by adding a mapstructure decode hook (`MapToMonitoringV1`) that routes all `monitoring/v1` types through `sigs.k8s.io/yaml`, which converts YAML to JSON first and then uses `json.Unmarshal` — correctly handling `json:",inline"`.

This is the same pattern as the existing `MapToLabelSelector` hook (which handles `metav1.LabelSelector` having `json:` tags instead of `yaml:` tags), but generic for all monitoring/v1 types rather than manually renaming individual fields.

## Test plan
- [x] Unit test: camelCase tlsConfig fields load correctly from YAML config
- [x] E2e: new step-04 in `targetallocator-prometheuscr` verifies scrapeClass TLS config is applied to ServiceMonitor scrape configs
- [x] E2e: updated step-02 assertion for `service_monitor_selector` (no longer null after type changes)
- [x] Manual: confirmed on CRC — stock TA shows no `tls_config` in `/scrape_configs`, fixed TA shows `ca_file` and `insecure_skip_verify` correctly


ozzywalsh:
Failing e2e appears to be flaky tests unrelated to change (everything was green before hitting publish for review).

.chloggen/fix_ta-scrapeclass-tls-deserialization.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: target allocator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix deserialization of prometheus-operator types (e.g. scrapeClasses tlsConfig) in target allocator config by handling json:",inline" embedded structs.
```
pavolloffay: types or a single type? 

I would keep the note simple and provide more details in the subtext with a config that was fixed.

ozzywalsh: Agreed; I have simplified the note and added and example to the subtext.

swiatekm (approved):
LGTM, as awful as the Prometheus config loading is becoming.

PR #5104, opened 2026-05-19
feat: add support for arb sm toggle

jaronoff97:
**Description:**
Add support to disable service and podmonitor endpoints that read arbitrary files. The current behavior is to keep this disabled to continue previous expectations. Users may want to enable this feature and manually mount secrets themselves in the future.

- Resolves: #issue-number

**Testing:** unit

**Documentation:** crd

### DIFF
--- a/cmd/otel-allocator/internal/watcher/promOperator.go
+++ b/cmd/otel-allocator/internal/watcher/promOperator.go
@@ -687,6 +687,9 @@ func (w *PrometheusCRWatcher) LoadConfig(ctx context.Context) (*promconfig.Confi
 // references (caFile, certFile, keyFile). This is the equivalent guard from
 // ArbitraryFSAccessThroughSMs.Deny in the Prometheus Operator.
 func (w *PrometheusCRWatcher) validateAndFilterScrapeConfigs(promCfg *promconfig.Config) error {
+	if !w.denyFSAccessThroughSMs {
+		return nil
+	}
 	for _, sc := range promCfg.ScrapeConfigs {
 		// Check the HTTPClientConfig.Authorization.CredentialsFile
 		auth := sc.HTTPClientConfig.Authorization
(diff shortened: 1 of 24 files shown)


cmd/otel-allocator/internal/watcher/promOperator.go
```diff
@@ -665,6 +678,50 @@ func (w *PrometheusCRWatcher) LoadConfig(ctx context.Context) (*promconfig.Confi
 	return promCfg, nil
 }
 
+// validateAndFilterScrapeConfigs validates scrape configs and returns an error if
+// any config references arbitrary files on the file system. This prevents
+// tenants from stealing the Collector's service account token via ServiceMonitor
+// bearerTokenFile (via authorization.credentials_file) or tlsConfig file
+// references (caFile, certFile, keyFile). This is the equivalent guard from
+// ArbitraryFSAccessThroughSMs.Deny in the Prometheus Operator.
+func (w *PrometheusCRWatcher) validateAndFilterScrapeConfigs(promCfg *promconfig.Config) error {
```
swiatekm: I think the right behavior should be to drop only the scrape configs with the denied fields.

swiatekm (commented):
We should add an e2e test for this too.

swiatekm (approved):

cmd/otel-allocator/internal/watcher/promOperator.go
```diff
@@ -649,6 +651,13 @@ func (w *PrometheusCRWatcher) LoadConfig(ctx context.Context) (*promconfig.Confi
 			return nil, unmarshalErr
 		}
 
+		// If denyFSAccessThroughSMs is enabled, drop scrape configs that reference
+		// arbitrary files on the file system. This prevents tenants from stealing
+		// the Collector's service account token.
+		if w.denyFSAccessThroughSMs {
+			w.filterScrapeConfigs(promCfg)
+		}
```
TylerHelmuth, line 659: Do we need to check the boolean here and in filterScrapeConfigs?

TylerHelmuth (approved):

swiatekm (approved):

PR #5105, opened 2026-05-20
Restrict automatic RBAC creation from users without permissions

swiatekm:
**Description:**

Ensure the auto-RBAC mechanism doesn't assign permissions the submitting user doesn't have. This happens at admission and is very similar to the mechanism that emits warnings during reconciliation. If the submitting user doesn't have permissions needed by the collector configuration that aren't covered by the existing ServiceAccount, then the submission is rejected.

**Testing:**

Added unit tests and e2e tests.

### DIFF
--- a/.chloggen/rbac-privilege-escalation-check.yaml
+++ b/.chloggen/rbac-privilege-escalation-check.yaml
@@ -1,0 +1,22 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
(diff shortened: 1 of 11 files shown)


jaronoff97 (approved):

ps-mir (approved):
I'm still new to the code, so did some digging.

The change is for pull-based receivers, where role+binding gets added to SA depending on k8s api access needs. The logic now checks for the new permissions required by SA and compares them with the user permissions.

The delta of required permissions will be more for the create case compared to the update case. Both scenarios are covered in Tests.

Looks ok.

frzifus (approved):

PR #5111, opened 2026-05-22
[chore] Read proxy env vars at startup and store them in config

swiatekm:

Instead of calling `proxy.ReadProxyVarsFromEnv()` inside each container manifest builder at reconcile time, read the proxy variables once during operator startup in `ApplyEnvVars` and store them in `config.Config.ProxyEnvVars`.

The manifest builders (collector, targetallocator, opampbridge) now read from `cfg.ProxyEnvVars` instead of calling `os.LookupEnv` out-of-band. This eliminates an antipattern where reconcile-time code had hidden dependencies on the process environment, and makes the values easy to control in tests by setting `cfg.ProxyEnvVars` directly rather than via `t.Setenv`.

A particular environment this fixes is a coding agent sandbox, where all external connections go through a proxy. Tests would fail there because of environment variables. This change isolates them.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 
 	"github.com/goccy/go-yaml"
+	corev1 "k8s.io/api/core/v1"
 
 	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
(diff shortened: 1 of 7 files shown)


frzifus (approved):
cc @open-telemetry/operator-approvers

PR #5115, opened 2026-05-24
[chore] Remove OpenShift-specific kubelet CA volume from ClusterObservability agent

frzifus:
## Summary
- Replace the OpenShift-specific `kubelet-ca.crt` hostPath volume with the `kube-root-ca.crt` ConfigMap for kubelet TLS verification
- Remove the OpenShift-specific `ca_file` override — the base config now handles all platforms
- Set `ca_file` in the base agent config to `/etc/kube-root-ca/ca.crt`

## Why
The previous implementation mounted `/etc/kubernetes/kubelet-ca.crt` via hostPath, which is OpenShift-specific. On Talos Linux this path doesn't exist (causing `FailedMount`), and on vanilla k8s the CA lives at `/etc/kubernetes/pki/ca.crt` with root-only permissions (causing `permission denied`).

The `kube-root-ca.crt` ConfigMap is automatically created by Kubernetes in every namespace and contains the cluster CA. Mounting it as a volume works universally across all distributions without platform-specific paths or file permission issues.

## Test plan
- [x] Unit tests pass (`make test`)
- [x] Verify kubeletstats scrapes metrics successfully on Talos
- [x] Verify kubeletstats scrapes metrics successfully on OpenShift
- [x] Verify `kube-root-ca.crt` ConfigMap exists in the target namespace

### DIFF
--- a/internal/manifests/clusterobservability/clusterobservability.go
+++ b/internal/manifests/clusterobservability/clusterobservability.go
@@ -204,12 +204,6 @@ func buildAgentCollector(params manifests.Params) (*v1beta1.OpenTelemetryCollect
 						MountPath: hostPathDockerContain,
 						ReadOnly:  true,
 					},
-					// OpenShift kubelet CA certificate mount (direct file)
-					{
-						Name:      "kubelet-serving-ca",
-						MountPath: "/etc/kubelet-serving-ca/ca-bundle.crt",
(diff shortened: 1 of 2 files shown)


frzifus:
skipping changelog, since its not part of the operator bundle right now.

swiatekm (approved):

PR #5119, opened 2026-05-24
[chore] Use /etc/os-release instead of /usr/lib/os-release in ClusterObservability agent

frzifus:
## Summary
- Switch os-release hostPath mount from `/usr/lib/os-release` to `/etc/os-release` in the ClusterObservability agent collector

## Why
`/usr/lib/os-release` does not exist on Talos Linux (and potentially other minimal/immutable distributions), causing `CreateContainerError` because the container runtime fails to mkdir on the read-only container filesystem. `/etc/os-release` is guaranteed to exist on all Linux distributions per the [os-release specification](https://www.freedesktop.org/software/systemd/man/os-release.html).

### DIFF
--- a/internal/manifests/clusterobservability/clusterobservability.go
+++ b/internal/manifests/clusterobservability/clusterobservability.go
@@ -175,8 +175,8 @@ func buildAgentCollector(params manifests.Params) (*v1beta1.OpenTelemetryCollect
 						ReadOnly:  true,
 					},
 					{
-						Name:      "host-usr-lib-osrelease",
-						MountPath: "/hostfs/usr/lib/os-release",
+						Name:      "host-etc-osrelease",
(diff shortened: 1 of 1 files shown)


swiatekm:
Can you add a comment explaining why these specific paths are used? Ideally move them to constants as well.

swiatekm (approved):

frzifus:
```bash
$ oc version
Client Version: 4.17.1
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: 4.19.4
Kubernetes Version: v1.32.6
$ kubectl debug node/ip-10-0-1-182.ec2.internal -it --image=busybox -- ls -l /host/etc/os-release
lrwxrwxrwx    1 root     root            21 May 24 13:43 /host/etc/os-release -> ../usr/lib/os-release
```

PR #5136, opened 2026-05-25
target allocator: allow configuring mtls between TargetAllocator & Collector

### LINKED ISSUE #4915: Add a config option for enabling target allocator mTLS

ozzywalsh:
## Description

Adds a configurable `Mtls` field to `TargetAllocatorSpec` (v1alpha1) for enabling mutual TLS between the Target Allocator and collectors. Users set `mtls.enabled: true` on the TA CR instead of relying on the `operator.targetallocator.mtls` feature gate, which is removed by this PR.

The `Mtls` field reuses the existing `v1beta1.TargetAllocatorMTLS` type and works for both standalone TA CRDs and those created by a Collector CR. When a Collector creates a TA, it forwards its mTLS config as a spec field on the TA CR.

Replaces #4971 — builds on @Horiodino's original work and addresses [review feedback](https://github.com/open-telemetry/opentelemetry-operator/pull/4971#pullrequestreview-4198044674) from @swiatekm requesting the config live on the TA spec.

### Key changes

- Add `Mtls` field to `v1alpha1.TargetAllocatorSpec`
- Remove `operator.targetallocator.mtls` feature gate — mTLS is now purely CRD-driven
- Decouple mTLS enablement from cert-manager availability: volumes, ports, and TLS config are always applied when `mtls.enabled: true`
- Cert-manager resources (Issuer, Certificate CRs) are only created when `useCertManager` is true (default) **and** cert-manager is detected at startup
- Add checks in the mutating webhook for both the Collector & the TargetAllocator; to throw an error if the user tries to create a TargetAllocator with `mtls.useCertManager: false` when cert manager is not available to the operator.

## Link to tracking Issue(s)

- Resolves: #4915


swiatekm (approved):
LGTM, thank you and @Horiodino for this contribution!

PR #5151, opened 2026-05-27
[chore] Improve docker caching in CI

swiatekm:
Clean up docker image caching in our workflows. The docker action we use has built-in support for using the GHA cache as a sort of impromptu registry, which simplifies the setup and deduplicates the cache. Right now, the different jobs stomp on each others caches and it doesn't really do much of use.


frzifus (approved):

frzifus:
I had in mind we did that in the past already

PR #5170, opened 2026-06-01
fix(opamp-bridge): exclude terminating collectors from EffectiveConfig

byehn:

`applyRemoteConfig` calls `applier.Delete` on collectors that are no longer in the config. Delete is asynchronous in k8s, sets deletionTimestamp and leaves the actual cleanup to finalizer. `UpdateEffectiveConfig` call that fires right after still sees those terminating collectors, so the bridge ends up reporting collectors that are on their way out as part of the effective config.

This patch filters out any OpenTelemetryCollector with a non-nil deletionTimestamp.

- Added `TestAgent_getEffectiveConfig/skip_terminating`

### DIFF
--- a/cmd/operator-opamp-bridge/internal/agent/agent.go
+++ b/cmd/operator-opamp-bridge/internal/agent/agent.go
@@ -348,6 +348,10 @@ func (agent *Agent) getEffectiveConfig(context.Context) (*protobufs.EffectiveCon
 	}
 	instanceMap := map[string]*protobufs.AgentConfigFile{}
 	for _, instance := range instances {
+		// Skip collectors pending deletion
+		if deletionTimestamp := instance.GetDeletionTimestamp(); deletionTimestamp != nil {
+			continue
+		}
 		col := instance
 		marshaled, err := yaml.Marshal(&col)
 		if err != nil {
--- a/cmd/operator-opamp-bridge/internal/agent/agent.go
+++ b/cmd/operator-opamp-bridge/internal/agent/agent.go
@@ -349,7 +349,7 @@ func (agent *Agent) getEffectiveConfig(context.Context) (*protobufs.EffectiveCon
 	instanceMap := map[string]*protobufs.AgentConfigFile{}
 	for _, instance := range instances {
 		// Skip collectors pending deletion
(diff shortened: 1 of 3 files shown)


cmd/operator-opamp-bridge/internal/agent/agent.go
```diff
@@ -348,6 +348,10 @@ func (agent *Agent) getEffectiveConfig(context.Context) (*protobufs.EffectiveCon
 	}
 	instanceMap := map[string]*protobufs.AgentConfigFile{}
 	for _, instance := range instances {
+		// Skip collectors pending deletion
+		if deletionTimestamp := instance.GetDeletionTimestamp(); deletionTimestamp != nil {
+			continue
```
swiatekm: Why not just

```suggestion
		if instance.GetDeletionTimestamp() != nil {
			continue
```

?

cmd/operator-opamp-bridge/internal/agent/agent_test.go
```diff
@@ -580,6 +580,55 @@ func TestAgent_getHealth(t *testing.T) {
...
+		}
+
+		applier := getFakeApplier(t, conf, collectorList)
+		mp := newMockProxy(nil, nil, nil)
+		agent := NewAgent(l, applier, conf, mockClient, mp)
+		err := agent.Start()
+		defer agent.Shutdown()
+		require.NoError(t, err, "should be able to start agent")
+
+		effectiveConfig, err := agent.getEffectiveConfig(context.Background())
```
swiatekm:
```suggestion
		effectiveConfig, err := agent.getEffectiveConfig(t.Context())
```

swiatekm (commented):
Some nitpicks, but looks good to me otherwise. @jaronoff97 do you think we need an e2e test for this behavior?

cmd/operator-opamp-bridge/internal/agent/agent.go
byehn: Saw it used that way in the operator's reconciler so I followed the same pattern, but if there's no real reason, agreed it's better to change it.

swiatekm (approved):

frzifus (approved):

PR #5173, opened 2026-06-02
chore: update Go to 1.26

frzifus:
## Summary
- Update `go` directive in all `go.mod` files (root, `apis/`, `tests/test-e2e-apps/bridge-server/`) from 1.25.x to 1.26.3
- Update Dockerfiles in `tests/test-e2e-apps/` to use `golang:1.26-alpine`
- Run `go mod tidy` on all modules

## Test plan
- [x] `go build ./...` compiles cleanly
- [x] CI passes


apis/go.mod
```diff
@@ -1,29 +1,25 @@
 module github.com/open-telemetry/opentelemetry-operator/apis
 
-go 1.25.0
+go 1.26.3
```
pavolloffay: Could we set it to 1.26.0 ? From the compatibility standpoint it's good to keep the patch version as low as possible.

pavolloffay: https://github.com/kubernetes/kubernetes/blob/master/go.mod#L9

frzifus:
@swiatekm

pavolloffay (approved):

Issue #4539, opened 2025-12-03
Go auto-instrumentation sidecar prevents Kubernetes Jobs/CronJobs from completing


brachi-gytpol:
### Component(s)

auto-instrumentation

## Description
When using Go auto-instrumentation on Kubernetes Jobs or CronJobs, the injected sidecar container does not terminate when the main application container exits. This causes the Job to remain in Running state indefinitely.

```
apiVersion: batch/v1
kind: CronJob
metadata:
  name: test-job
spec:
  schedule: "*/5 * * * *"
  jobTemplate:
    spec:
      template:
        metadata:
          annotations:
            instrumentation.opentelemetry.io/inject-go: "grafana/apm-instrumentation"
            instrumentation.opentelemetry.io/otel-go-auto-target-exe: /app
        spec:
          containers:
            - name: app
              image: my-go-app:latest
              command: ["/app"]
```
## Expected Behavior

When the main container exits (code 0 or 1), the sidecar should terminate and the Job should complete.

## Actual Behavior

- Main container: `Terminated` (exit code 0 or 1)
- Sidecar container: `Running` (indefinitely)
- Pod status: `Running`
- Job status: Never completes

### Kubernetes Version

v1.33.5-eks-3cfe0ce

### Operator version

opentelemetry-operator-0.99.1

### Collector version

-

### Log output

```shell

```

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

yuriolisa:
@brachi-gytpol, do you wanna to submit a fix for this?

iblancasa:
I'm not sure if we should fix this. But deprecate go instrumentation. The repository seems to be unmaintained.

yuriolisa:
> I'm not sure if we should fix this. But deprecate go instrumentation. The repository seems to be unmaintained.

Let's discuss it at SIG.

Horiodino:
i'll identify the cause and discuss it before making any changes.

iblancasa:
> i'll identify the cause and discuss it before making any changes.

We need to discuss about this during the SIG.

iypetrov:
@iblancasa I didn't attend the SIG meeting, what was the conclusion?

[...]

PR #5180, opened 2026-06-03
fix(tests): pin Go e2e test app to 1.26.1 for eBPF offset compatibility

frzifus:
## Summary
- Pins the Go e2e test app base image to `golang:1.26.1-alpine` instead of the floating `1.26-alpine` tag
- The Go auto-instrumentation (`v0.24.0`) only has cached eBPF offsets up to Go 1.26.1. The recent Go 1.26 upgrade (#5173) resolved to Go 1.26.3 via the floating tag, causing `instrumentation-go` e2e tests to fail with no telemetry data produced
- Once a new `opentelemetry-go-instrumentation` release adds offsets for Go 1.26.3+, this pin can be removed

See: https://github.com/open-telemetry/opentelemetry-go-instrumentation/releases/

## Test plan
- [x] `e2e-instrumentation` CI job passes (specifically `instrumentation-go` test)

### DIFF
--- a/tests/test-e2e-apps/golang/Dockerfile
+++ b/tests/test-e2e-apps/golang/Dockerfile
@@ -1,5 +1,5 @@
 # Stage 1: Build the Go application
-FROM docker.io/library/golang:1.26-alpine@sha256:91eda9776261207ea25fd06b5b7fed8d397dd2c0a283e77f2ab6e91bfa71079d AS builder
+FROM docker.io/library/golang:1.26.1-alpine@sha256:2389ebfa5b7f43eeafbd6be0c3700cc46690ef842ad962f6c5bd6be49ed82039 AS builder
 
 # Set the working directory inside the container for the build stage.
 WORKDIR /app


iblancasa (approved):

swiatekm:
Doesn't look like this is fixing the failures. Maybe we should revert the Go test app change until we figure this out?

iblancasa:
> Doesn't look like this is fixing the failures. Maybe we should revert the Go test app change until we figure this out?

I think we should deprecate Go instrumentation https://github.com/open-telemetry/opentelemetry-operator/issues/4539

swiatekm:
> > Doesn't look like this is fixing the failures. Maybe we should revert the Go test app change until we figure this out?
> 
> I think we should deprecate Go instrumentation #4539

That, too, but first we need to unblock our CI in the simplest possible change.

frzifus:
> > > Doesn't look like this is fixing the failures. Maybe we should revert the Go test app change until we figure this out?
> > 
> > 
> > I think we should deprecate Go instrumentation #4539
> 
> That, too, but first we need to unblock our CI in the simplest possible change.

@iblancasa  @swiatekm  In any case  I think  we have to force merge  something. :sweat:

swiatekm:
> > > > Doesn't look like this is fixing the failures. Maybe we should revert the Go test app change until we figure this out?
> > > 
> > > 
> > > I think we should deprecate Go instrumentation #4539
> > 
> > 
> > That, too, but first we need to unblock our CI in the simplest possible change.
> 
> @iblancasa @swiatekm In any case I think we have to force merge something. 😓

I don't think we do? The Go 1.26.4 update PR passes govulncheck. We just need to disable the Go e2e test there. Let me have a stab at it.

swiatekm:
@frzifus can you build this locally and sure it passes e2e tests?

PR #5188, opened 2026-06-05
Add a way to customize the command/entrypoint of the main container

agagniere:
## Motivation

When using the collector Helm chart, we can specify ([link](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.yaml#L327-L329)):
```yaml
# OpenTelemetry Collector executable
command:
  name: ""
  extraArgs: []
```
1. `command.name` allows overriding the container's entrypoint. Note that a `/` is prepended to the value provided
2. `command.extraArgs` allows adding arguments after the `--config=/conf/relay.yaml` one.

With the operator, only the args are customizable (only `--key=value` args, which is fine for collectors)

## Description

To align what is customizable with the Helm chart vs with the operator, I propose adding a `command` field to the CRD. It will be equivalent to the chart's `command.name`, and allow overriding the container's entrypoint.

## Testing

Added a unit test confirming the generated container has its command set to the provided value.

### DIFF
--- a/.chloggen/customize-command.yaml
+++ b/.chloggen/customize-command.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: collector
+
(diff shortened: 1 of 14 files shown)


agagniere:
Done in https://github.com/open-telemetry/opentelemetry-operator/pull/4973

PR #5205, opened 2026-06-11
Fix failing TA tests on IBM Z

pavolloffay:

- Resolves: #issue-number

The `curlimages/curl` is not available on IBM Z, causing our internal tests pipelines to fail. I am switching curl images to `alpine` and wget which is well supported for all platforms. 

### DIFF
--- a/tests/e2e-prometheuscr/create-sm-prometheus-exporters/08-install.yaml
+++ b/tests/e2e-prometheuscr/create-sm-prometheus-exporters/08-install.yaml
@@ -120,7 +120,7 @@ spec:
       - args:
         - /bin/sh
         - -c
-        - curl -s http://simplest-targetallocator/jobs | grep "targetallocator"
-        image: mirror.gcr.io/curlimages/curl:latest
+        - wget -q -O - http://simplest-targetallocator/jobs | grep "targetallocator"
+        image: mirror.gcr.io/alpine:latest
         name: check-metrics
       restartPolicy: OnFailure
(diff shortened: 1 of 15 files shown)


andreasgerstmayr (approved):

swiatekm:
Can we not get an image containing curl? wget is less nice.

IshwarKanse:
@swiatekm @pavolloffay Should we add a workflow to build our own image with curl like https://github.com/grafana/tempo-operator/pkgs/container/tempo-operator%2Ftest-utils or can we use the Tempo one ?

frzifus:
> @swiatekm @pavolloffay Should we add a workflow to build our own image with curl like https://github.com/grafana/tempo-operator/pkgs/container/tempo-operator%2Ftest-utils or can we use the Tempo one ?

I think using the tempo one should  be fine

IshwarKanse:
> > @swiatekm @pavolloffay Should we add a workflow to build our own image with curl like https://github.com/grafana/tempo-operator/pkgs/container/tempo-operator%2Ftest-utils or can we use the Tempo one ?
> 
> I think using the tempo one should be fine

We didn't add another workflow to build image cause we are building too many images already in this repo.

pavolloffay:
I am not sure what other think but I would prefer to build an image for OTEL tests here to keep the repo self-contained.  @swiatekm  WDYT?

pavolloffay:
> Can we not get an image containing curl? wget is less nice.

I didn't find any multiarch upstream image containing curl. We would have to maintain it ourself

swiatekm:
> I am not sure what other think but I would prefer to build an image for OTEL tests here to keep the repo self-contained. @swiatekm WDYT?

Sure, that's better. But we should set it up from the start to get automatically updated by renovate and actually tested in PRs that change it.

PR #5211, opened 2026-06-13
RFC: Documentation rework

swiatekm:
We've talked about reorganizing and improving the operator documentation since at least KubeCon Amsterdam 2023. This RFC proposes a concrete plan to do so, including a PoC.

### DIFF
--- a/docs/rfcs/documentation.md
+++ b/docs/rfcs/documentation.md
@@ -1,0 +1,117 @@
+# Documentation restructure
+
+**Status:** *Draft*
+
+**Author:** Mikołaj Świątek (@swiatekm)
+
+**Date:** 2026-06-13
+
+
+## Objective
+
+Reorganize the operator's user documentation into a structured `docs/` tree, fill known content gaps, and add small autogeneration for reference material that's currently hand-maintained.
+
+The current state has four problems:
+
+- The root [`README.md`](../../README.md) is ~970 lines, with most user-facing documentation living several headings deep inside it.
+- `docs/` is a small set of standalone files with no coherent organization ([`compatibility.md`](../compatibility.md), [`cluster-observability.md`](../cluster-observability.md), [`crd-changelog.md`](../crd-changelog.md), [`api/`](../api/), `rfcs/`).
+- Target Allocator user documentation lives at [`cmd/otel-allocator/README.md`](../../cmd/otel-allocator/README.md), next to its code, where users don't look.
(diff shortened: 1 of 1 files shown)


docs/rfcs/documentation.md
```diff
@@ -0,0 +1,117 @@
...
+- `operator-flags.md` - from the `pflag.FlagSet` constructed in [`main.go`](../../main.go).
+- `target-allocator-flags.md` - from [`cmd/otel-allocator/main.go`](../../cmd/otel-allocator/main.go).
+- `feature-gates.md` - from [`pkg/featuregate/featuregate.go`](../../pkg/featuregate/featuregate.go) via `featuregate.GlobalRegistry().VisitAll(...)`.
+
+CI gains a `docs-check` target mirroring the existing api-docs verify step in the [`Makefile`](../../Makefile).
+
+Phase 3 has no dependency on Phase 2 and can land in parallel if a contributor picks it up.
+
+
+## Open questions
```
jaronoff97, line 102: I would also like us to discuss the strategy for our docs vs the opentelemetry.io docs, should we just point everyone to our repo docs?

jaronoff97 (approved):

iblancasa (commented):
I think it would be good if we add a section about operator self telemetry

swiatekm:
> I think it would be good if we add a section about operator self telemetry

Yeah, that should probably exist. Do you think it should be its own top-level section? Target Allocator has its own telemetry, for example.

docs/rfcs/documentation.md
swiatekm: I'm honestly not sure, which is why I'm dodging the question in this RFC. :sweat_smile: Once we have all the content we need to have, organized properly, I think we can ask one of the docs maintainers for advice on how to proceed. I don't mind duplicating content as long as there's a reconciliation process in play.

iblancasa:
> > I think it would be good if we add a section about operator self telemetry
> 
> Yeah, that should probably exist. Do you think it should be its own top-level section? Target Allocator has its own telemetry, for example.

Mmm I think very likely. Or at least as part of reference maybe?

swiatekm:
> > > I think it would be good if we add a section about operator self telemetry
> > 
> > 
> > Yeah, that should probably exist. Do you think it should be its own top-level section? Target Allocator has its own telemetry, for example.
> 
> Mmm I think very likely. Or at least as part of reference maybe?

Added it there. Eventually, we can even autogenerate it, if we declare it in a Weaver manifest.

frzifus (approved):

pavolloffay (approved):

Issue #4571, opened 2025-12-15
Add support for managing OTel eBPF Instrumentation (aka OBI)


cyrille-leclerc:
### Is your feature request related to a problem? Please describe.

Make [OpenTelemetry eBPF Instrumentation](https://opentelemetry.io/docs/zero-code/obi/) (aka OBI) manageable with the OpenTelemetry Operator similarly to the way the OpenTelemetry Collector and OpenTelemetry language instrumentation SDKs are manageable using the OTel Operator.

### Describe the solution you'd like

I would like to define OBI's configuration through a CRD that looks similar to the OTelCollector and OTel Instrumentation CRDs and then get the OTel Operator manage OBI accordingly so that all the OTel instrumentation of my K8s clusters are managed through one single mchanism that ensures proper correlation between all the telemetry types and layers.

### Describe alternatives you've considered

Today I manage OBI with a distinct deployment model. This causes downsides such as
* Risk of gaps, overlaps, or conflicts between the OTel instrumentation
* Challenge to correlate OTel telemetry types and layers like 
   * OBI metrics and traces
   * OTel SDKs metrics, traces, and logs
   * K8s pod logs and metrics

### Additional context

This Enhancement request was discussed with @grcevski

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

swiatekm:
Nice! The next thing I'd like to see before agreeing to implement this is a simple PoC, a set of K8s manifests showing a sample application instrumented with this image. I want to understand what it is that the operator needs to do as part of the injection process.

The next step after that would be an RFC document, as this is a major enough change to warrant it.

iblancasa:
> All deployment forms supported by the OTel Collector are available through the OTel Operator: deployment, daemonset, statefulset or sidecar ([here](https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/api/opentelemetrycollectors.md)). I would assume that the OTel eBPF Instrumentation would follow the sam. flexibility of offering through the OTel Operator all the deployment modes it supports. Does this help?

[...]

PR #5219, opened 2026-06-15
Add OBI eBPF RFC

ozzywalsh:
## Description

RFC proposing a new cluster-scoped CRD (`ClusterOBIAgent`) to integrate the [OpenTelemetry eBPF Instrumentation (OBI)](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation) agent into the operator as a first-class managed workload.

Covers CRD design, security model (privileged vs capability-based), namespace/RBAC management, and OpenShift SCC handling.

**Rendered:** [preview](https://github.com/ozzywalsh/opentelemetry-operator/blob/obi-rfc/docs/rfcs/obi-agent.md)

## Link to tracking Issue(s)

N/A

## Testing

Documentation only — no code changes.

## Documentation

`docs/rfcs/obi-agent.md`


docs/rfcs/obi-agent.md
```diff
@@ -0,0 +1,293 @@
...
+    open_port: '8080'
+    discovery:
+      instrument:
+        - k8s_namespace: tenant-alpha
+exporters:
+  otlp:
+    endpoint: <your-otlp-endpoint>
+```
+
+Unfortunately this is not yet available in an existing collector distribution, such as `otelcol-contrib`. Users would need to build their own collector image using `ocb`. Therefore this is unviable currently for initial OBI support in the operator. Eventually the operator could support both a dedicated `DaemonSet` (via the proposal in this RFC) and deployment via a collector `DaemonSet`. Since both options require configuring a `DaemonSet` with necessary privileges, volume mounts etc; the work in this RFC can be reused to later support the collector receiver model.
```
iblancasa: Maybe https://github.com/open-telemetry/opentelemetry-collector-releases/issues/1490 should be linked here.

ozzywalsh: Yes; actually the collector receiver approach could be very convenient; the existing `Collector` CR is already capable of supporting it; it just requires a lot of extra config for the linux capabilities, volume mounts etc.

Perhaps the operator could do this automatically. 

It would be nice to get input from the OBI team also; related issue https://github.com/open-telemetry/opentelemetry-operator/issues/4571

For example (includes SCC for openshift; not needed for vanilla k8s):
```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: obi-system
  labels:
    pod-security.kubernetes.io/enforce: privileged
    pod-security.kubernetes.io/warn: privileged
    pod-security.kubernetes.io/audit: privileged
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: obi-collector-collector
  namespace: obi-system
---
apiVersion: security.openshift.io/v1
kind: SecurityContextConstraints
metadata:
  name: obi-collector
allowPrivilegedContainer: false
allowHostPID: true
allowHostNetwork: false
allowHostPorts: false
allowHostDirVolumePlugin: true
allowPrivilegeEscalation: true
readOnlyRootFilesystem: true
runAsUser:
  type: RunAsAny
seLinuxContext:
  type: RunAsAny
fsGroup:
  type: RunAsAny
supplementalGroups:
  type: RunAsAny
allowedCapabilities:
  - BPF
  - SYS_PTRACE
  - NET_RAW
  - CHECKPOINT_RESTORE
  - DAC_READ_SEARCH
  - PERFMON
  - SYS_ADMIN
volumes:
  - emptyDir
  - hostPath
  - configMap
  - projected
users:
  - system:serviceaccount:obi-system:obi-collector-collector
groups: []
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: obi-collector
rules:
  - apiGroups: [""]
    resources: ["pods", "nodes", "services"]
    verbs: ["get", "list", "watch"]
  - apiGroups: [""]
    resources: ["replicationcontrollers"]
    verbs: ["get", "list", "watch"]
  - apiGroups: ["apps"]
    resources: ["deployments", "replicasets", "statefulsets", "daemonsets"]
    verbs: ["get", "list", "watch"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: obi-collector
subjects:
  - kind: ServiceAccount
    name: obi-collector-collector
    namespace: obi-system
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: obi-collector
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: obi-collector-scc
rules:
  - apiGroups: ["security.openshift.io"]
    resources: ["securitycontextconstraints"]
    resourceNames: ["obi-collector"]
    verbs: ["use"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: obi-collector-scc
subjects:
  - kind: ServiceAccount
    name: obi-collector-collector
    namespace: obi-system
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: obi-collector-scc
---
kind: OpenTelemetryCollector
metadata:
  name: obi-collector
  namespace: obi-system
spec:
  mode: daemonset
  managementState: managed
  image: quay.io/ozwalsh/otelcol-obi:v0.9.0
  serviceAccount: obi-collector-collector
  config:
    receivers:
      obi:
        discovery:
          instrument:
            - k8s_namespace: tenant-alpha
              k8s_pod_annotations:
                obi.instrument: "true"
            - k8s_namespace: tenant-beta
              k8s_pod_annotations:
                obi.instrument: "true"
        attributes:
          kubernetes:
            enable: "true"
    processors:
      transform:
        trace_statements:
          - context: resource
            statements:
              - set(attributes["service.name"], Concat([attributes["k8s.namespace.name"], attributes["service.name"]], "/"))
    exporters:
      debug:
        verbosity: detailed
      otlp/jaeger:
        endpoint: jaeger.observability.svc.cluster.local:4317
        tls:
          insecure: true
    service:
      pipelines:
        traces:
          receivers: [obi]
          processors: [transform]
          exporters: [debug, otlp/jaeger]
  hostPID: true
  securityContext:
    runAsUser: 0
    readOnlyRootFilesystem: true
    capabilities:
      add:
        - BPF
        - SYS_PTRACE
        - NET_RAW
        - CHECKPOINT_RESTORE
        - DAC_READ_SEARCH
        - PERFMON
        - SYS_ADMIN
  volumeMounts:
    - name: var-run-obi
      mountPath: /var/run/obi
    - name: cgroup
      mountPath: /sys/fs/cgroup
  volumes:
    - name: var-run-obi
      emptyDir: {}
    - name: cgroup
      hostPath:
        path: /sys/fs/cgroup
```

docs/rfcs/obi-agent.md
```diff
@@ -0,0 +1,294 @@
...
+
+**Date:** 2026-06-15
+
+## Objective
+
+Integrate the [OpenTelemetry eBPF Instrumentation (OBI)](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation) agent into the operator as a first-class managed workload, enabling automatic zero-code instrumentation of applications via eBPF without requiring application changes or restarts.
+
+## Summary
+
+This RFC introduces a new cluster-scoped CRD, `ClusterOBIAgent`, that the operator reconciles into a privileged DaemonSet running the OBI eBPF agent. The operator manages the full lifecycle of the agent: namespace creation, RBAC, security context, and configuration.
```
frzifus: Can we handle the OBI deployment without introducing a new CRD?

Does OBI replace the need of auto-instr. injection?

Could we provide a FG (`--enable-obi-instr="otel-obi"`) to enable to enable the agent deployment into a specific namespace?

If the instrumentation CR could be reused, he operator can deploy obi agents and configure it to send data to a collector service that can be configured on the fly, based on existing instrumentation CRs.
The obi agent config will be maintained by the operator.  If someone needs tweaks, the operator will respect a config-overwrite.

TLDR;
if `--enable-obi-instr="<namespace>"` != `""`
- disable webhook for sdk injection
- deploy collector 
- Check if obi config overwrite exists (respect it)
- deploy obi daemonset, send data to collector
- configure collector based on existing instrumentation CRs

else:
- keep existing auto-instr. sdk injetion behavior

iblancasa: If I'm not mistaken, during KubeCon we talked about this and the decision was use both instrumentation mechanisms and not replace one by the other. Because they are complementary.

ozzywalsh: From my understanding they are complementary; this [blog post about OBI's release](https://opentelemetry.io/blog/2025/obi-announcing-first-release/#should-i-just-use-opentelemetry-ebpf-instrumentation-for-everything) talks about that.

It's also able to detect that an application is instrumented using another sdk; and not duplicate the signals.

swiatekm: I don't like the Instrumentation CRD doing wildly different things depending on configuration. Right now it just informs a mutating pod webhook, but the OBI instrumentation spawns a cluster-wide DaemonSet. I'd rather not mix these together.

I'm aware this is something of a controversial topic, so it's also worth talking through in a SIG meeting.

And we definitely don't intend OBI to replace the per-language autoinstrumentation.

ozzywalsh: But I like the approach of not requiring a new `ClusterOBIAgent` CR at all necessarily; and utilizing the existing `Instrumentation` CR.

docs/rfcs/obi-agent.md
swiatekm: OBI having an implementation using the existing Collector CR would be very nice indeed. I'm definitely in favor if we can make it happen.

docs/rfcs/obi-agent.md
```diff
@@ -0,0 +1,294 @@
...
+
+    // Capabilities allows adding to or overriding the operator-computed
+    // capability set. Only applies when privileged: false.
+    // If Override is set it replaces the operator-computed capabilities
+    // entirely. If only Add is set, those capabilities are appended.
+    Capabilities *OBICapabilitySpec `json:"capabilities,omitempty"`
+
+    // +kubebuilder:pruning:PreserveUnknownFields
+    // Config represents the raw OBI agent YAML configuration.
+    Config *apiextensionsv1.JSON `json:"config,omitempty"`
```
swiatekm: Might be simpler to use string here, I recall this type being something of a pain to manage when we tried it for the collector config.

docs/rfcs/obi-agent.md
```diff
@@ -0,0 +1,294 @@
...
+Matches the [OBI Helm chart default](https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-ebpf-instrumentation). Maximum compatibility with kind, Docker-in-Docker, minikube, and managed Kubernetes.
+
+```yaml
+securityContext:
+  privileged: true
+```
+
+### `privileged: false` — operator-computed capabilities
+
+The operator injects the minimum capabilities for the configured mode. The base set for application observability:
```
swiatekm: How complex is this to compute for a given environment? Can we import a library that will do it for us?

docs/rfcs/obi-agent.md
```diff
@@ -0,0 +1,294 @@
...
+    add: [BPF, PERFMON, DAC_READ_SEARCH, CHECKPOINT_RESTORE, SYS_PTRACE, NET_RAW]
+```
+
+### Compatibility note
+
+`SYS_PTRACE` interacts with the `RuntimeDefault` seccomp profile applied by default on some distributions (GKE Autopilot, hardened EKS). A seccomp profile and a capability grant are orthogonal — both must permit the syscall. Users on such clusters may need `privileged: true` or a custom seccomp profile. This is the primary reason `privileged: true` remains the default.
+
+## Namespace Management
+
+The operator creates and owns a dedicated `obi-system` namespace for all OBI resources. This is not user-configurable. The operator labels it for Pod Security Admission on all platforms:
```
swiatekm: I don't see why this wouldn't be configurable, to be honest. Maybe we should rethink if this needs to be a singleton. I can easily see users wanting to run differently configured agents on different node groups, for example.

docs/rfcs/obi-agent.md
ozzywalsh: I think the operator could set the capabilities using the the same logic as this function here in OBI https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/obi/os.go#L174

docs/rfcs/obi-agent.md
ozzywalsh: @swiatekm Is your concern mainly about the Instrumentation CRD triggering the creation of a DaemonSet?

I would also find that undesirable.

What I have in mind (and I think that's what @frzifus means) is:

The `--enable-obi-agent="<namespace>"` flag is passed to the operator at startup; this triggers the OBI DaemonSet (or Collector with OBI receiver) to be created in that namespace. The Instrumentation CR never creates or manages the DaemonSet.

Then Instrumentation CRs with `spec.obi: {}` are reconciled by the operator into OBI's discovery config. The existing SDK injection webhook is completely unaffected — `spec.python` still triggers pod mutation as today. `spec.obi` is only consumed by a separate reconciler that writes discovery entries into the OBI ConfigMap.

For example, two tenants each create an Instrumentation in their namespace:

  ```yaml
  kind: Instrumentation
  metadata:
    name: default
    namespace: tenant-alpha
  spec:
    exporter:
      endpoint: http://otel-collector:4318
    python:
      image: ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-python:latest
    obi: {}
```
```yaml
  kind: Instrumentation
  metadata:
    name: default
    namespace: tenant-beta
  spec:
    exporter:
      endpoint: http://otel-collector:4318
    obi: {}
```

The operator reconciles all Instrumentation CRs with spec.obi and builds the OBI ConfigMap. The namespace is injected automatically from the CR's metadata, and each entry requires the
  instrumentation.opentelemetry.io/inject-obi: "true" pod annotation — so tenants must explicitly opt in workloads:

```yaml
  # OBI ConfigMap (managed by operator)
  discovery:
    instrument:
      - k8s_namespace: tenant-alpha
        k8s_pod_annotations:
          instrumentation.opentelemetry.io/inject-obi: "true"
      - k8s_namespace: tenant-beta
        k8s_pod_annotations:
          instrumentation.opentelemetry.io/inject-obi: "true"
```
tenant-alpha gets both Python SDK injection (via the webhook) and OBI eBPF tracing. tenant-beta gets OBI only. OBI detects SDK-instrumented applications and avoids duplicating signals.

ozzywalsh:
I have updated the RFC and tried to take in account the feedback; it now focuses on adding support for the obi collector receiver component to the operator (this has also been added to the collector contrib distro since the RFC was opened)

[RFC](https://github.com/ozzywalsh/opentelemetry-operator/blob/obi-rfc/docs/rfcs/obi-agent.md)

The recently merged PR to add obi receiver to collector contrib distro: https://github.com/open-telemetry/opentelemetry-collector-releases/pull/1386

docs/rfcs/obi-agent.md
```diff
@@ -0,0 +1,240 @@
...
+
+### `Instrumentation` CR with `spec.obi` for tenant delegation
+
+An `Instrumentation` CR with `spec.obi: {}` could allow namespace owners to opt their workloads into OBI instrumentation, with the operator aggregating these into OBI's discovery config. However, OBI already has built-in Kubernetes-aware discovery that filters by namespace, pod labels, and pod annotations natively. Pod labels may be sufficient for tenant opt-in without any CR involvement. This is deferred pending real-world feedback on whether OBI's native selectors are adequate.
+
+## Custom collector image requirement
+
+Previously it was only possible to use the obi receiver by building a custom collector image.
+
+The obi receiver has (as of 2026-07-02) been [added](https://github.com/open-telemetry/opentelemetry-collector-releases/pull/1386) to the
```
pavolloffay, line 55: A better is probably add a collector version since it is available or will be

docs/rfcs/obi-agent.md
```diff
@@ -0,0 +1,240 @@
...
+          insecure: true
+    service:
+      pipelines:
+        traces:
+          receivers: [obi]
+          processors: [routing]
+          exporters: [otlp/tenant-alpha, otlp/tenant-beta]
+```
+
+The operator auto-configures hostPID, privileged security context, and required volumes. The user only specifies the discovery selectors and export destinations.
```
pavolloffay, line 113: it would help to provide more details , maybe as a fragment of the deployment spec what is added by the operator.

atoulme:
Does it need a CRD, if we can just enable it as part of a collector deployed as a daemonset?

swiatekm:
> Does it need a CRD, if we can just enable it as part of a collector deployed as a daemonset?

We talked about this during a recent SIG meeting and the conclusion was that we should start with documentation showing how to use OBI end to end using the operator. We still might want to have a thin CRD to abtract away some toil and provide a more explicit security boundary, as OBI requires very privileged containers to work.

PR #5224, opened 2026-06-17
fix(collector): default fsGroup from namespace range on OpenShift

ozzywalsh:
## Summary

- On OpenShift, permissive SCCs (e.g. `anyuid`) do not inject `fsGroup` from the namespace range, unlike the `restricted` SCC. When a user grants privileges to the collector ServiceAccount (e.g. for RBAC), the effective SCC changes and PVC volumes become inaccessible: `mkdir /queues/work/: permission denied`
- The controller now defaults `podSecurityContext.fsGroup` from the namespace's `supplemental-groups` annotation (falling back to `uid-range`) when running on OpenShift and no explicit `fsGroup` is configured
- An explicitly set `fsGroup` is never overwritten

## Approach

Reads the namespace's `openshift.io/sa.scc.supplemental-groups` annotation (falling back to `openshift.io/sa.scc.uid-range`) and parses the first ID from the allocated range using `library-go/pkg/security/uid.ParseBlock`. 

Supersedes #5106, which hardcoded `fsGroup: 65532` in the webhook — that value would be rejected by the `restricted` SCC if outside the namespace's allocated range.

## Test plan

- [x] Unit tests
- [x] Manual verification on OpenShift cluster with anyuid SCC + PVC storage


ozzywalsh:
@pavolloffay @frzifus

frzifus (approved):

internal/controllers/opentelemetrycollector_controller.go
```diff
@@ -175,6 +181,47 @@ func (r *OpenTelemetryCollectorReconciler) GetParams(ctx context.Context, instan
...
+//
+// See: https://github.com/migtools/crane/blob/440432b/cmd/transfer-pvc/transfer-pvc.go#L554
+func (r *OpenTelemetryCollectorReconciler) defaultFSGroupOnOpenShift(ctx context.Context, instance *v1beta1.OpenTelemetryCollector) {
+	if instance.Spec.PodSecurityContext != nil && instance.Spec.PodSecurityContext.FSGroup != nil {
+		return
+	}
+
+	ns := &corev1.Namespace{}
+	if err := r.Get(ctx, types.NamespacedName{Name: instance.Namespace}, ns); err != nil {
+		r.log.V(4).Info("unable to fetch namespace for fsGroup defaulting", "namespace", instance.Namespace, "error", err)
```
pavolloffay: Maybe it's better to use the info logging for this and below as well?

internal/controllers/opentelemetrycollector_controller.go
```diff
@@ -175,6 +181,47 @@ func (r *OpenTelemetryCollectorReconciler) GetParams(ctx context.Context, instan
...
+	}
+
+	block, err := uid.ParseBlock(rangeAnnotation)
+	if err != nil {
+		r.log.V(4).Info("unable to parse group range annotation", "annotation", rangeAnnotation, "error", err)
+		return
+	}
+
+	fsGroup := int64(block.Start)
+	if instance.Spec.PodSecurityContext == nil {
```
pavolloffay, line 219: I think this is redundant. The expectation is that the context is nill

tests/e2e-openshift/fsgroup-anyuid-pvc/01-assert.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: fsgroup-test-collector
+spec:
+  template:
+    spec:
+      (securityContext.fsGroup != null): true
```
pavolloffay: This assertion checks that fsGroup is not null, but doesn't verify the actual value. The value check happens later in `verify-fsgroup.sh`.

If fsGroup gets set to a wrong value (e.g., hardcoded instead of from namespace range), this assertion passes and the failure only shows up in the shell script step — which is harder to debug.

Consider checking the expected value here directly. Chainsaw supports bindings to capture the namespace UID range first, then assert against it:

```yaml
(securityContext.fsGroup == $expected_fsgroup): true
```

Not a blocker — the shell script catches it — but inline value assertions fail faster and are easier to diagnose.

pavolloffay (commented):
Minor suggestion on the e2e test assertion.

internal/controllers/opentelemetrycollector_controller.go
ozzywalsh: The nil check isn't redundant; the guard at the top lets through two cases:

1. PodSecurityContext == nil -> need to initialize the struct before setting FSGroup
 2. PodSecurityContext != nil but FSGroup == nil  -> user set other fields (e.g. RunAsUser), skip initialization to preserve them, only set FSGroup

Removing the conditional and always initializing would wipe user-specified fields in case 2.

tests/e2e-openshift/fsgroup-anyuid-pvc/01-assert.yaml
ozzywalsh: Ok; I have changed it to check inline instead.

internal/controllers/opentelemetrycollector_controller.go
ozzywalsh: fixed!

pavolloffay (approved):

PR #5226, opened 2026-06-18
[chore] Fix inverted ApiAvailability.String() values

byehn:
The string slice in gatewayapi.ApiAvailability.String() was ordered inversely.

### DIFF
--- a/internal/autodetect/gatewayapi/api.go
+++ b/internal/autodetect/gatewayapi/api.go
@@ -14,5 +14,5 @@ const (
 )
 
 func (p ApiAvailability) String() string {
-	return [...]string{"Available", "NotAvailable"}[p]
+	return [...]string{"NotAvailable", "Available"}[p]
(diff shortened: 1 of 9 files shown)


iblancasa (approved):

internal/autodetect/gatewayapi/api.go
```diff
@@ -14,5 +14,5 @@ const (
 )
 
 func (p ApiAvailability) String() string {
-	return [...]string{"Available", "NotAvailable"}[p]
```
iblancasa, line 17: Should we add some kind of test? I don't know how we didn't catch this before.

swiatekm:
I'll second a test.

byehn:
Agreed, thanks. I've added tests for all String().

swiatekm (approved):
👍 Thanks for fixing this!

PR #5229, opened 2026-06-19
fix(collector): support declarative telemetry resource format

### LINKED ISSUE #5207: Mutating webhook overwrites service.telemetry when using declarative resource.attributes schema
### Component(s)

collector

## Description
[OpenTelemetry Collector v0.151.0](https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.151.0) added support for declarative `service.telemetry.resource.attributes` configuration.

However, the OpenTelemetry Operator mutating webhook [still parses `service.telemetry.resource` as a legacy inline map (`map[string]*string`)](https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/otelconfig/config.go#L66). Because of this, an `OpenTelemetryCollector` CR using the new declarative resource schema can be mutated unexpectedly.
  -  When [GetTelemetry](https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/otelconfig/config.go#L526) fails to unmarshal the new schema, ServiceApplyDefaults treats it as an empty config and overwrites s.Telemetry.Object, dropping the user input before injecting the default Prometheus reader.

The operator webhook should support both service.telemetry.resource formats:

Legacy inline map:
```
  resource:
    service.name: my-collector
```
Declarative resource schema:
```
  resource:
    attributes:
      - name: service.name
        value: my-collector
```

## Steps to Reproduce
Given this Collector-valid configuration:
```yaml
  service:
    telemetry:
      resource:
        attributes:
          - name: deployment.environment.name
            value: production
      metrics:
        level: detailed
        readers:
          - periodic:
              interval: 60000
              exporter:
                otlp:
                  protocol: grpc
                  endpoint: otel-collector:4317
                  insecure: true
```

## Expected Result
The webhook leaves service.telemetry untouched (the user explicitly configured metrics.readers), and the rendered ConfigMap preserves both resource.attributes and the OTLP periodic reader.

## Actual Result
The mutating webhook injects the default Prometheus metrics reader:

```
  service:
    telemetry:
      metrics:
        readers:
          - pull:
              exporter:
                prometheus:
                  host: 0.0.0.0
                  port: 8888
                  without_scope_info: false
                  without_type_suffix: false
                  without_units: false
```
User-provided resource, metrics.level, and configured readers are gone.

As a workaround, we can use the legacy inline map format. But, Collector v0.153.0 then emits a warning:
>   Using legacy service.telemetry.resource inline map format; prefer service.telemetry.resource.attributes
>   (array of maps with `name` and `value` keys)

### Kubernetes Version

1.26.4

### Operator version

0.152.0

### Collector version

0.153.0

## Environment
OS: (e.g., "Ubuntu 20.04")

### Log output

```shell

```

### Additional context

_No response_

### Tip

ozzywalsh:
## Summary

Fixes #5207.

- The `Telemetry.Resource` field was typed as `map[string]*string`, which rejected the declarative `resource.attributes` array format introduced in Collector v0.151.0
- `GetTelemetry` failed to unmarshal, returned nil, and `ServiceApplyDefaults` overwrote the entire telemetry block with defaults — dropping user-provided resource attributes, metrics level, and custom readers
- Changed `Resource` to `json.RawMessage` so both legacy and declarative formats pass through without interpretation

## Test plan

- Added 7 unit tests covering legacy and declarative resource formats through `GetTelemetry`, `ServiceApplyDefaults`, and `TelemetryToAnyConfig`

### DIFF
--- a/internal/otelconfig/config_test.go
+++ b/internal/otelconfig/config_test.go
@@ -1595,166 +1595,175 @@ func TestAddPrometheusMetricsEndpointPreservesShapeWhenGateDisabled(t *testing.T
 	require.False(t, *reader.Pull.Exporter.Prometheus.WithoutScopeInfo)
 }
 
-func TestGetTelemetryWithLegacyResource(t *testing.T) {
-	svc := v1beta1.Service{
-		Telemetry: &v1beta1.AnyConfig{
-			Object: map[string]any{
-				"metrics":  map[string]any{"level": "detailed"},
-				"resource": map[string]any{"service.name": "my-collector"},
-			},
-		},
-	}
-
-	tel := GetTelemetry(&svc, logr.Discard())
-	require.NotNil(t, tel)
-	assert.Equal(t, "detailed", tel.Metrics.Level)
-	assert.JSONEq(t, `{"service.name":"my-collector"}`, string(tel.Resource))
-}
-
-func TestGetTelemetryWithDeclarativeResource(t *testing.T) {
-	svc := v1beta1.Service{
-		Telemetry: &v1beta1.AnyConfig{
-			Object: map[string]any{
-				"metrics": map[string]any{"level": "detailed"},
-				"resource": map[string]any{
-					"attributes": []any{
-						map[string]any{"name": "service.name", "value": "my-collector"},
-						map[string]any{"name": "deployment.environment.name", "value": "production"},
+func TestGetTelemetryResourceFormats(t *testing.T) {
+	tests := []struct {
+		name             string
+		service          v1beta1.Service
+		expectedResource string // JSON; empty means nil
+	}{
+		{
+			name: "no resource field",
+			service: v1beta1.Service{
+				Telemetry: &v1beta1.AnyConfig{
+					Object: map[string]any{
+						"metrics": map[string]any{"level": "basic"},
 					},
 				},
 			},
 		},
-	}
-
-	tel := GetTelemetry(&svc, logr.Discard())
-	require.NotNil(t, tel, "GetTelemetry must not return nil for declarative resource format")
-	assert.Equal(t, "detailed", tel.Metrics.Level)
-}
-
-func TestServiceApplyDefaultsPreservesLegacyResource(t *testing.T) {
-	cfg := &v1beta1.Config{
-		Service: v1beta1.Service{
-			Telemetry: &v1beta1.AnyConfig{
-				Object: map[string]any{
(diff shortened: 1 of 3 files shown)


internal/otelconfig/config_test.go
```diff
@@ -1594,3 +1594,167 @@ func TestAddPrometheusMetricsEndpointPreservesShapeWhenGateDisabled(t *testing.T
...
+			Object: map[string]any{
+				"metrics":  map[string]any{"level": "detailed"},
+				"resource": map[string]any{"service.name": "my-collector"},
+			},
+		},
+	}
+
+	tel := GetTelemetry(&svc, logr.Discard())
+	require.NotNil(t, tel)
+	assert.Equal(t, "detailed", tel.Metrics.Level)
```
pavolloffay: if possible it's better to assert on the entire telemetry object, rather than pieces of it.

internal/otelconfig/config_test.go
```diff
@@ -1594,3 +1594,167 @@ func TestAddPrometheusMetricsEndpointPreservesShapeWhenGateDisabled(t *testing.T
 	require.NotNil(t, reader.Pull.Exporter.Prometheus.WithoutScopeInfo)
 	require.False(t, *reader.Pull.Exporter.Prometheus.WithoutScopeInfo)
 }
+
+func TestGetTelemetryWithLegacyResource(t *testing.T) {
```
pavolloffay: could you please dry this test with table driven approach? In in single test we could execute the apply defaults and get telemetry.

elliehmy:
@pavolloffay @iblancasa @jaronoff97 — I'm the reporter of #5207. Gentle ping on this one, since it looks like it may have gotten buried. The Jun 24 review comments look addressed and checks are green. Could one of you take another look when you get a chance?

bourbonkk:
@ozzywalsh heads up that this has picked up merge conflicts since the last activity, so it can't go in as-is even once a reviewer gets to it.

Context on why we're watching it: we hit #5207 downstream. Collector 0.160.0 warns on the legacy inline map format across all of our collectors, but we had to stay on it — with the declarative `resource.attributes` format our Argo CD deployments go perpetually out-of-sync, because the webhook drops our `metrics.readers` OTLP config and substitutes the default Prometheus reader.

If you don't have time to pick this back up, I'm happy to rebase and carry it forward in a follow-up PR with your commits preserved. Either way is fine — just let us know.

PR #5241, opened 2026-06-22
[chore] Update opamp-bridge CRD and docs

owais:
**Description:** 
A few things were added to the opamp-bridge configuration but not reflected in the API types, CRD and docs. Some of the things such as standalone mode is probably not needed in operator mode but it is better to have the CRD in sync for completeness otherwise some tooling, e2e tests, etc appear to have a hard time.


owais:
@swiatekm added CRD changes independently. I think it is better to have them in sync than not even if we don't expect users to use them.

apis/v1alpha1/opampbridge_types.go
```diff
@@ -29,6 +29,31 @@ type OpAMPBridgeSpec struct {
 	// Description allows the customization of the non identifying attributes for the OpAMP Bridge.
 	// +optional
 	Description *AgentDescription `json:"description,omitempty"`
+	// HeartbeatInterval is the interval to use for sending a heartbeat. Setting it to 0 disables the heartbeat.
```
iblancasa, line 32: Maybe this needs a changelog now?

iblancasa (commented):
Can you explain where did you sync from and how? Maybe it is something we need to document/fix to avoid non-synced stuff.

owais:
> Can you explain where did you sync from and how? Maybe it is something we need to document/fix to avoid non-synced stuff.

I added the missing types to `apis/v1alpha1/opampbridge_types.go` and ran make update.

jaronoff97:
@owais originally I was thinking that someone may want to create a bridge via CRD in standalone mode hence why I thought it was valuable to have the CRD change for the bridge. Now I'm wondering if that's something we actually should support, is that an unreasonable expectation or do you think it makes sense? If we think it makes sense to have that CRD change, we should pipe the settings through the operator's reconciliation loop rather than having them disconnected like it is in the current state. What do you think?

owais:
I think choosing to explicitly not support standalone mode when configured via CRD is not unreasonable. We can take that stance until we run into a use case where someone actually needs it. 

I've been thinking of getting rid of the "mode" thing and support both types of workloads (workloads + CRs) at the same time and we might need it then but we can always bring it in if I find a good enough solution to support that.

I'm fine with leaving standalone mode config out of the CRD for now.

PR #5189, opened 2026-06-06
add observedGeneration and conditions to Collector and Instrumentation status

Horiodino:
**Description:**
This PR implements support for `status.observedGeneration` and `status.conditions` for both `OpenTelemetryCollector` and `Instrumentation` custom resources.

The change enables controllers to report the latest reconciled generation and expose reconciliation status through standard Kubernetes conditions. A `Ready` condition is now maintained during reconciliation, and `observedGeneration` is updated whenever the controller processes a new resource generation.

**Link to tracking Issue(s):**

* Resolves: #4312

**Testing:**

* Added reconciliation tests for `OpenTelemetryCollector` status updates.

**Documentation:**

* Updated API documentation.

PR #5243, opened 2026-06-22
Revert Instrumentation status fields; fix upgrade-path status handling

ozzywalsh:
## Summary

PR #5189 added `status.observedGeneration` and `status.conditions` to the Instrumentation CRD, setting them inside the upgrade path (`InstrumentationUpgrade.ManagedInstances`). Since that code only runs once on operator startup, status was never updated in response to spec changes — the e2e test had to restart the operator and sleep to trigger it.

After discussion on #5189, this PR reverts the Instrumentation status fields and cleans up the upgrade path instead:

- Removes `observedGeneration` and `conditions` from `InstrumentationStatus` (and regenerated CRDs, bundles, docs)
- Removes status-setting code from the upgrade path — upgrade now only handles `upgradeBlockedVersions`
- Switches the blocked-versions status write from `Status().Update()` to a `MergeFrom` patch so it only touches the fields it owns
- Restores `continue` on spec update failure so the status patch is skipped when the preceding update fails (avoids operating on stale resourceVersion)
- Removes the operator restart + sleep hacks from the e2e test
- Narrows the #5189 changelog entry to OpenTelemetryCollector only


swiatekm:
Can't we update this in the admission webhook instead? In some sense, these fields are trivial for an Instrumentation CR, given that it's not actually reconciled.

ozzywalsh:
> Can't we update this in the admission webhook instead? In some sense, these fields are trivial for an Instrumentation CR, given that it's not actually reconciled.

@swiatekm As I understand; it's not possible to set the CR status from the admission webhook https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#status-subresource

swiatekm:
> > Can't we update this in the admission webhook instead? In some sense, these fields are trivial for an Instrumentation CR, given that it's not actually reconciled.
> 
> @swiatekm As I understand; it's not possible to set the CR status from the admission webhook https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#status-subresource

That makes sense given how the status API works, but is annoying. As is, setting observed generations for Instrumentation doesn't really do anything. I'm a bit concerned that the field existing will introduce confusion with the mutating webhook. It wouldn't be an unreasonable belief that if the observed generation is equal to the actual generation, then the webhook should apply that version of the Instrumentation resource, but that's not at all guaranteed, especially in horizontally scaled webhook servers. @open-telemetry/operator-approvers WDYT?

ozzywalsh:
I see what you mean. I was coming at it from a bug fix angle; since what the previous PR added (not yet released); which only sets `status.observedGeneration` on operator startup is obviously broken. 

Certainly the condition should not be set; since that's misleading; instead of:
```yaml
status:
  conditions:
  - lastTransitionTime: "2026-06-23T10:20:23Z"
    message: Successfully reconciled
    observedGeneration: 1
    reason: Reconciled
    status: "True"
    type: Ready
  observedGeneration: 1
```
we should just have:
```yaml
status:
  observedGeneration: 1
```

But yes, even then `status.observedGeneration` on the Instrumentation CR seems meaningless.

How about we:
* Do no set `status.observedGeneration` on the Instrumentation at all; remove it from the `internal/instrumentation/upgrade/upgrade.go` and also do no add a new controller for the Instrumentation.
* Remove `InstrumentationStatus.conditions` also; since it will be unused (was added by the previous PR)

swiatekm:
Yeah, I'm kind of leaning in that direction. It doesn't do anything useful for us, and is potentially confusing. If the webhook wants to stamp the resource version or generation on the instrumented Pod, it can do so without touching status.

ozzywalsh:
I have updated the PR to stop setting `status.observervedGeneration` & `status.conditions` on the Instrumenation CR altogether. 

Looks like flaky test failing.

swiatekm (approved):

ozzywalsh:
Seperate PR that I believe should address the flaky test https://github.com/open-telemetry/opentelemetry-operator/pull/5247

jaronoff97 (approved):

Issue #5242, opened 2026-06-22
Improve target relabeling and hashing test coverage for target allocator


swiatekm:
The target allocator needs to handle target labels in a way that is compatible with both prometheus and prometheus-operator. Without relabel filtering, this only involves calculating target identity from labels, but even this can be subject to drift from upstream Prometheus semantics. This has been a source of many subtle bugs over the years, and even refactoring to improve the situation is hampered by fear of inadvertently introducing more such bugs.

We should improve test coverage to the level where major changes to the target allocator logic are possible without fear of breakage. Unfortunately, there isn't an official conformance test suite we can use here - the relabelling logic is an implementation detail of prometheus. But their code is public, so we should be able to at least cover what they do.

My rough plan for this is the following:

1. Run the target allocation pipeline against a suite of conformance tests generated from the results of [`promtool check service-discovery`](https://prometheus.io/docs/prometheus/latest/command-line/promtool/#promtool-check-service-discovery)

1. Run the same suite as an integration test combining the target allocator and prometheus receiver in-process. This can actually verify the shape of the data.

1. Add an e2e test in a kind cluster which uses the same configurations to scrape the same data using both Prometheus itself and the target allocator + otel collector setup exporting to Prometheus, and verify that the target labels are the same.

Issues blocked on this:

- https://github.com/open-telemetry/opentelemetry-operator/issues/4074

PR #5248, opened 2026-06-24
Add target allocator conformance integration tests

swiatekm:
**Description:**

Add in process integration tests for target allocator. The tests start a test http server exposing prometheus metrics in exposition format, a target allocator pinned to a single collector, and an in-process prometheus receiver configured to use the target allocator. The prometheus receiver writes to a metrics sink from the otel collector's component test framework, and the tests assert on the shape of the data therein.

We use the existing scrape configurations for conformance tests, adding a `prometheus.yaml` file containing the raw exposition data and expected metrics metadata. The only adjustment done to the scrape configs is target rewriting, so the test http server is scraped.

This all lives in a separate go module. The reason is that we don't want to deal with importing prometheus receiver into the operator go.mod and all the resultant headaches. The integration module can do its own thing without impacting the rest of the project. I'm going to make renovate play well with both this new module and the apis module in a follow up.

- Relates: https://github.com/open-telemetry/opentelemetry-operator/issues/5242

**Documentation:**

The new module has a doc.go summarizing what it contains.


cmd/otel-allocator/integrationtest/harness.go
```diff
@@ -0,0 +1,198 @@
...
+	"go.opentelemetry.io/collector/component/componenttest"
+	"go.opentelemetry.io/collector/confmap"
+	"go.opentelemetry.io/collector/consumer/consumertest"
+	"go.opentelemetry.io/collector/receiver/receivertest"
+	"go.uber.org/zap/zaptest"
+
+	prometheusreceiver "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/prometheusreceiver"
+
+	"github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/internal/allocation"
+	taconfig "github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/internal/config"
```
jaronoff97, line 33: the import sorting i think is incorrect here?

jaronoff97 (approved):
tests look fine, one linting question

pavolloffay (approved):

cmd/otel-allocator/integrationtest/harness.go
swiatekm: It might be. I'll fix this in a follow-up for both this module and the apis module.

PR #5251, opened 2026-06-25
[chore] Add e2e test for collector internal metrics

IshwarKanse:
## Summary
- Adds an upstream e2e test (`tests/e2e/collector-internal-metrics/`) that validates the collector's internal metrics endpoint on port 8888
- Deploys a collector with traces, metrics, and logs pipelines, sends telemetry via telemetrygen, then asserts 18 key metric names exist with expected values
- Catches regressions from semantic convention migrations that rename collector internal metrics (e.g. `otelcol_exporter_sent_spans`, `otelcol_process_uptime`)
- Works on both upstream kind clusters and OpenShift — uses `kubectl get --raw` API proxy, no Prometheus/ServiceMonitor required.


tests/e2e/collector-internal-metrics/chainsaw-test.yaml
```diff
@@ -0,0 +1,86 @@
...
+              (length($metrics[?as_string(metric."__name__") == 'otelcol_process_uptime']) > `0`): true
+              (length($metrics[?as_string(metric."__name__") == 'otelcol_process_cpu_seconds']) > `0`): true
+              (length($metrics[?as_string(metric."__name__") == 'otelcol_process_memory_rss']) > `0`): true
+              (length($metrics[?as_string(metric."__name__") == 'otelcol_process_runtime_heap_alloc_bytes']) > `0`): true
+              (length($metrics[?as_string(metric."__name__") == 'otelcol_process_runtime_total_alloc_bytes']) > `0`): true
+              (length($metrics[?as_string(metric."__name__") == 'otelcol_process_runtime_total_sys_memory_bytes']) > `0`): true
+              # Exporter metrics (value > 1 proves data flowed)
+              ($metrics[?as_string(metric."__name__") == 'otelcol_exporter_sent_spans'].value | [0] > `1`): true
+              ($metrics[?as_string(metric."__name__") == 'otelcol_exporter_sent_metric_points'].value | [0] > `1`): true
+              ($metrics[?as_string(metric."__name__") == 'otelcol_exporter_sent_log_records'].value | [0] > `1`): true
```
pavolloffay, line 70: Some tests in the repo use `_total` suffix https://github.com/open-telemetry/opentelemetry-operator/blob/6574bd44cd84de03b8b3ea8f89cd3f7887d63f48/tests/e2e/filelog-receiver/chainsaw-test.yaml#L76 

Why there is this inconsistency? I know that you recently fixed some tests, but I wonder why the filelog receiver test pasases.

swiatekm: Filelog receiver test is pinned to a specific collector version, and this behavior does depend on that version - otel core recently fixed a bug that affects this. The solution is for the filelog test to read the version from the `OTELCOL_VERSION` env variable, I think and use `latest` if it's unset.

IshwarKanse: yeah, we can use renovate to make sure the pinned collector version is the same as the operator provided one. I'll check and create a new PR to fix it since its not related to this one.

IshwarKanse: Added PR https://github.com/open-telemetry/opentelemetry-operator/pull/5313 to remove the hardcoded collector versions in tests.

pavolloffay (approved):

PR #5100, opened 2026-05-19
Make instrumentation webhook HA

pavolloffay:
**Description:** <Describe what has changed.>

**Link to tracking Issue(s):** <Issue number if applicable>

- Resolves: #5010 

## Summary

 Split the pod mutation webhook (`/mutate-v1-pod`) into a standalone deployment on OpenShift to enable high availability (HA). The webhook handles auto-instrumentation and sidecar injection.

  - Add a separate `opentelemetry-operator-pod-webhook` deployment to the OpenShift CSV with 2 replicas by default
  - Add a `PodWebhookReconciler` controller that reconciles the pod-webhook deployment replica count based on the `POD_WEBHOOK_REPLICAS` env var (scales up and down, capped at CSV default)
  - Split `main.go` into subcommands: `operator` (full operator) and `auto-instrumentation` (webhook-only mode for the standalone deployment)
  - Add `POD_WEBHOOK_REPLICAS` to the operator config (`internal/config/`) with a default of 2
  - Add kustomize overlay resources for the pod-webhook deployment (`config/overlays/openshift/`)
  - Add e2e tests verifying:
    - `LABELS_FILTER` env var propagation via OLM Subscription
    - Scaling pod-webhook down via `POD_WEBHOOK_REPLICAS=1`
    - Scaling pod-webhook back to default (2) when env var is removed

## Design 

  See `docs/rfcs/pod-mutation-high-availability.md` included in this PR.

## Notes

```
 # Build and push the operator image
 IMG=ghcr.io/pavolloffay/opentelemetry-operator:latest make bundle container container-push

 # Build and push the OpenShift bundle
BUNDLE_VARIANT=openshift IMG=ghcr.io/pavolloffay/opentelemetry-operator:latest BUNDLE_IMG=ghcr.io/pavolloffay/opentelemetry-operator-bundle:latest make bundle-build bundle-push

 # Deploy via OLM
 ./bin/operator-sdk run bundle ghcr.io/pavolloffay/opentelemetry-operator-bundle:latest

Cleanup
kubectl delete catalogsources.operators.coreos.com --all
kubectl delete clusterserviceversion.operators.coreos.com --all
kubectl delete subscriptions.operators.coreos.com --all

Upgrade
OPERATOR_VERSION=0.150.1 IMG=ghcr.io/pavolloffay/opentelemetry-operator:latest BUNDLE_IMG=ghcr.io/pavolloffay/opentelemetry-operator-bundle:latest make bundle bundle-build bundle-push
[...]

PR #5253, opened 2026-06-26
Webhook command only

pavolloffay:
This is a split PR from https://github.com/open-telemetry/opentelemetry-operator/pull/5100 

This is part of https://github.com/open-telemetry/opentelemetry-operator/issues/5010 

Add a new `pod-webhook` subcommand that runs only the pod mutation webhook without the controllers. This enables High Availability (HA) deployment where the webhook can be scaled independently of the operator.

Key changes:
- Refactor main.go to use cobra subcommands (operator and pod-webhook)
- Add RFC documentation for pod mutation HA design

The pod-webhook mode runs without leader election since multiple webhook instances can serve requests concurrently.

- Resolves: #issue-number


docs/rfcs/pod-mutation-high-availability.md
```diff
@@ -0,0 +1,302 @@
+# Pod Mutation High Availability
```
swiatekm: Can we just make this webhook high availability? This would be more generic if it could be all the webhooks, which we can make configurable if we want to.

docs/rfcs/pod-mutation-high-availability.md
```diff
@@ -0,0 +1,302 @@
...
+│                    opentelemetry-operator-system                │
+│                                                                 │
+│  ┌──────────────────────────┐    ┌──────────────────────────┐  │
+│  │ opentelemetry-operator-  │    │ opentelemetry-operator-  │  │
+│  │ controller-manager       │    │ pod-webhook              │  │
+│  │ (1 replica, leader       │    │ (2 replicas, no leader   │  │
+│  │  election)               │    │  election)               │  │
+│  │                          │    │                          │  │
+│  │ - CR controllers         │    │ - /mutate-v1-pod         │  │
+│  │ - CR webhooks            │    │   (instrumentation +     │  │
```
swiatekm: Is there a problem moving admission webhooks into the other deployment?

docs/rfcs/pod-mutation-high-availability.md
```diff
@@ -0,0 +1,302 @@
...
+```
+
+On vanilla Kubernetes (community bundle), everything remains in a single deployment (no change).
+
+### How It Works
+
+#### OpenShift with OLM
+
+1. **Kustomize overlay** adds a second deployment definition via `config/overlays/openshift/pod-webhook-deployment.yaml`. This becomes part of the CSV during `make bundle`.
+2. **Bundle generation** patches the `mpod.kb.io` webhookdefinition to point to the pod-webhook deployment instead of the controller-manager.
```
swiatekm: I think it'd be easier to patch the Service instead.

main.go
```diff
@@ -104,30 +105,62 @@ func main() {
...
-		ec.MessageKey = cfg.Zap.MessageKey
-		ec.LevelKey = cfg.Zap.LevelKey
-		ec.TimeKey = cfg.Zap.TimeKey
-		if cfg.Zap.LevelFormat == "lowercase" {
-			ec.EncodeLevel = zapcore.LowercaseLevelEncoder
-		} else {
-			ec.EncodeLevel = zapcore.CapitalLevelEncoder
-		}
-	})
+func newAutoInstrumentationCmd() *cobra.Command {
```
swiatekm: Can we make this a generic webhook server command, maybe with options controlling which webhooks we're serving?

main.go
```diff
@@ -675,3 +624,258 @@ func parseFipsFlag(fipsFlag string) (receivers, exporters, processors, extension
 	}
 	return receivers, exporters, processors, extensions
 }
+
+func applyConfigAndSetupLogger(cfg *config.Config, configFile string, opts zap.Options) {
```
swiatekm: I like moving these to separate functions a lot! Should we also move them to a new package to keep `main.go` lean? This would also help setup in envtest e2e tests, which now need to reproduce parts of it independently. How about moving the command definitions to `./cmd/manager` and `./cmd/webhook` and the shared bits to `internal/cmd/common`?

swiatekm (commented):
I don't like needing to do a separate controller just to work around OLM limitations, but I can live with it. But I'd like the webhook server to be more generic and support all the webhooks we have prospectively.

docs/rfcs/pod-mutation-high-availability.md
pavolloffay: Is there a reason why we would do that? The whole point for moving the pod mutation to a separate deployment is for the HA sidecar/auto-instrumentation use-case. 

I would prefer keeping the the defaulting/validating as close to the controller as possible.

docs/rfcs/pod-mutation-high-availability.md
pavolloffay: The same question is asked below, let's keep the conversation there.

docs/rfcs/pod-mutation-high-availability.md
swiatekm: In my view any argument in favor of moving webhooks to a separate deployment applies equally to all of them. The admission webhook not being available until the operator proper starts is an issue in some cases too.

docs/rfcs/pod-mutation-high-availability.md
pavolloffay: Could you please calrify? The `admissionReviewVersions` in the CSV  points to the `deploymentName`

https://github.com/open-telemetry/opentelemetry-operator/blob/4e806894a853f13b3db3601a039ace1d84d8b57e/bundle/openshift/manifests/opentelemetry-operator.clusterserviceversion.yaml#L831 


```
 operator-sdk generate regenerates the CSV from scratch every time you run make bundle. It always sets all webhooks' deploymentName to opentelemetry-operator-controller-manager (the default/only deployment it knows about).

  So the flow is:

  1. make bundle runs operator-sdk generate → CSV gets generated with deploymentName: opentelemetry-operator-controller-manager for every webhook, including mpod.kb.io
  2. The sed patch runs immediately after → changes just the mpod.kb.io entry back to opentelemetry-operator-pod-webhook

  Without the patch, every make bundle would revert line 831 to the default, and CI (make ensure-update-is-noop) would fail because the committed CSV wouldn't match what make bundle produces.

  In short: operator-sdk doesn't support per-webhook deployment routing natively, so we patch it as a post-generation step.
```

docs/rfcs/pod-mutation-high-availability.md
pavolloffay: You're right. I looked at how other operators handle this and they all move **all** webhook types (mutating + validating + defaulting) into a single separate webhook deployment:

- **cert-manager** — dedicated [`cert-manager-webhook`](https://github.com/cert-manager/cert-manager/tree/master/cmd/webhook) deployment serves mutating, validating, and CRD conversion webhooks. Controller has no webhooks.
- **Kyverno** — separate [`admission-controller`](https://github.com/kyverno/kyverno/tree/main/cmd/kyverno) deployment handles all admission webhooks, separate from background/reports/cleanup controllers.
- **Gatekeeper (OPA)** — [`controller-manager`](https://github.com/open-policy-agent/gatekeeper/tree/master/cmd/manager) (webhook) is separate from `audit`. Validating + mutating webhooks run together, with granular `--operation` flags to split further.
- **Knative Serving** — dedicated [`webhook`](https://github.com/knative/serving/tree/main/cmd/webhook) deployment separate from `controller`, serving all validation and defaulting webhooks.

None of them keep CRD validation in the controller while splitting out only pod mutation.

I'll rename the subcommand to `webhook` (instead of `pod-webhook`) and make it serve all webhooks. In the OLM/OpenShift follow-up PR we can then configure which webhooks to route to the separate deployment.

main.go
pavolloffay: Noted as TODO in https://github.com/open-telemetry/opentelemetry-operator/issues/5010

docs/rfcs/pod-mutation-high-availability.md
pavolloffay: noted as TODO in https://github.com/open-telemetry/opentelemetry-operator/issues/5010

main.go
```diff
@@ -104,30 +105,61 @@ func main() {
...
+	opts := zap.Options{}
+	var zapFlagSet flag.FlagSet
+	opts.BindFlags(&zapFlagSet)
+	cliFlags.AddGoFlagSet(&zapFlagSet)
 
+	var configFile string
+	cliFlags.StringVar(&configFile, "config-file", "", "Path to config file")
+
+	cmd := &cobra.Command{
+		Use:   "webhook",
```
swiatekm: `webhook-server`, maybe?

main.go
```diff
@@ -104,30 +105,61 @@ func main() {
...
-		ec.MessageKey = cfg.Zap.MessageKey
-		ec.LevelKey = cfg.Zap.LevelKey
-		ec.TimeKey = cfg.Zap.TimeKey
-		if cfg.Zap.LevelFormat == "lowercase" {
-			ec.EncodeLevel = zapcore.LowercaseLevelEncoder
-		} else {
-			ec.EncodeLevel = zapcore.CapitalLevelEncoder
-		}
-	})
+func newAutoInstrumentationCmd() *cobra.Command {
```
swiatekm:
```suggestion
func newWebhookServerCmd() *cobra.Command {
```

main.go
```diff
@@ -675,3 +623,258 @@ func parseFipsFlag(fipsFlag string) (receivers, exporters, processors, extension
...
+			setupLog.Error(err, "unable to set up webhook ready check")
+			os.Exit(1)
+		}
+	}
+}
+
+// runAutoInstrumentationWebhook runs only the auto-instrumentation webhook without the controllers.
+// This enables High Availability (HA) deployment where the webhook can be scaled independently.
+func runAutoInstrumentationWebhook(cfg config.Config, configFile string, opts zap.Options) {
+	applyConfigAndSetupLogger(&cfg, configFile, opts)
```
swiatekm: nitpick: A lot of the setup that happens here feels similar to the operator setup, but we can deduplicate it in a follow up.

main.go
```diff
@@ -675,3 +623,258 @@ func parseFipsFlag(fipsFlag string) (receivers, exporters, processors, extension
...
+	logger := ctrl.Log
+
+	configLog := ctrl.Log.WithName("config")
+
+	v := version.Get()
+
+	logger.Info("Starting the OpenTelemetry Auto-Instrumentation Webhook",
```
swiatekm:
```suggestion
	logger.Info("Starting the OpenTelemetry Webhook Server",
```

main.go
```diff
@@ -675,3 +623,258 @@ func parseFipsFlag(fipsFlag string) (receivers, exporters, processors, extension
...
+
+	validateFilterPatterns(cfg)
+
+	// Register the Instrumentation webhook for validation
+	if err = wh.SetupInstrumentationWebhook(mgr, cfg); err != nil {
+		setupLog.Error(err, "unable to create webhook", "webhook", "Instrumentation")
+		os.Exit(1)
+	}
+
+	// Register the pod mutation webhook with only the instrumentation mutator
```
swiatekm: We register the sidecar webhook below as well, so either this comment should be modified, or the sidecar should be removed.

docs/rfcs/pod-mutation-high-availability.md
swiatekm: For OLM that makes sense. For the generic bundle, the Service should be enough, right?

main.go
pavolloffay: removed

swiatekm (approved):

PR #5276, opened 2026-06-30
Fix operator NetworkPolicy crash on OpenShift: separate ipBlock and selector peers

### LINKED ISSUE #5275: Operator CrashLoopBackOff on OpenShift: invalid NetworkPolicy mixing ipBlock with podSelector
## Description

The operator crashes on startup when the `operator.networkpolicy` feature gate is enabled (default on OpenShift). The operator pod enters CrashLoopBackOff with:

```
NetworkPolicy.networking.k8s.io "opentelemetry-operator" is invalid:
  spec.egress[0].to[0]: Forbidden: may not specify both ipBlock and another peer
```

## Root Cause

The Kubernetes NetworkPolicy API forbids combining `ipBlock` with `podSelector` or `namespaceSelector` within the same egress rule's `to` list. On OpenShift, the operator NetworkPolicy code appends PodSelector/NamespaceSelector peers into the same egress rule that already contains IPBlock peers for the API server IPs.

This was a latent bug introduced across PRs #4490 and #4426, but was activated by PR #5253 which changed the startup order so that OpenShift autodetect now runs before NetworkPolicy creation.

## Impact

- **Severity:** High — the operator cannot start at all on OpenShift clusters.
- **Affected environments:** Any OpenShift cluster where `operator.networkpolicy` is enabled (default in OpenShift overlay).

## Fix

Place PodSelector/NamespaceSelector peers in a separate egress rule from IPBlock peers.

## Environment

- OpenShift 4.12+
- OpenTelemetry Operator v0.153.0 (main branch)
- Feature gate `operator.networkpolicy` enabled

IshwarKanse:
## Summary

- Fix CrashLoopBackOff on OpenShift caused by mixing `ipBlock` with `podSelector`/`namespaceSelector` in the same NetworkPolicy egress rule peer list. Place them in separate egress rules instead.
- Add unit tests for the `operatornetworkpolicy` package (previously had zero tests).
- Add OpenShift e2e test validating both operator and operand NetworkPolicies.
- Update kind e2e assertion to also verify egress rules.

## Root Cause

The Kubernetes API forbids combining `ipBlock` with `podSelector` or `namespaceSelector` within the same `NetworkPolicyPeer` list ([docs](https://kubernetes.io/docs/concepts/services-networking/network-policies/#behavior-of-to-and-from-selectors)). On OpenShift, after #5253 changed the startup order so autodetect runs before NetworkPolicy creation, both IPBlock and selector peers were added to the same egress rule, triggering:

```
spec.egress[0].to[0]: Forbidden: may not specify both ipBlock and another peer
```

## Test plan

- [x] Unit tests pass: `go test ./internal/operatornetworkpolicy/... -v` (8 tests)
- [x] Full test suite passes: `make test` (2218 tests)
- [x] Lint passes: `make lint`
- [x] Changelog validated: `make chlog-validate`
- [x] OpenShift e2e test passes on OCP 4.22: `chainsaw test --test-dir ./tests/e2e-openshift/networkpolicy`

Fixes #5275

### DIFF
--- a/internal/operatornetworkpolicy/operatornetworkpolicy_test.go
+++ b/internal/operatornetworkpolicy/operatornetworkpolicy_test.go
@@ -1,0 +1,341 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package operatornetworkpolicy
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/kubernetes/fake"
+	kubeTesting "k8s.io/client-go/testing"
+)
+
+func newTestScheme() *runtime.Scheme {
+	s := runtime.NewScheme()
+	_ = appsv1.AddToScheme(s)
+	_ = networkingv1.AddToScheme(s)
+	_ = corev1.AddToScheme(s)
+	return s
+}
+
+func operatorDeployment(namespace string) *appsv1.Deployment {
+	return &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      operatorName,
+			Namespace: namespace,
+			UID:       "test-uid",
+		},
+		Spec: appsv1.DeploymentSpec{
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{"app.kubernetes.io/name": "opentelemetry-operator"},
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{"app.kubernetes.io/name": "opentelemetry-operator"},
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{{Name: "manager", Image: "test"}},
+				},
+			},
+		},
+	}
+}
+
+// startAndCapture runs Start() and captures the created NetworkPolicy
+// by using a reactor that cancels the context after creation.
+func startAndCapture(t *testing.T, clientset *fake.Clientset, scheme *runtime.Scheme, opts ...Option) *networkingv1.NetworkPolicy {
+	t.Helper()
+
+	var captured *networkingv1.NetworkPolicy
+	ctx, cancel := context.WithCancel(context.Background())
+
+	clientset.PrependReactor("create", "networkpolicies", func(action kubeTesting.Action) (bool, runtime.Object, error) {
+		createAction := action.(kubeTesting.CreateAction)
+		np := createAction.GetObject().(*networkingv1.NetworkPolicy)
+		captured = np.DeepCopy()
+		cancel()
+		return false, np, nil
+	})
+
+	n := NewOperatorNetworkPolicy(clientset, scheme, opts...)
+	err := n.(*networkPolicy).Start(ctx)
+	require.NoError(t, err)
+	require.NotNil(t, captured, "NetworkPolicy was not created")
+	return captured
+}
+
+func ownerRef() []metav1.OwnerReference {
+	trueVal := true
+	return []metav1.OwnerReference{
+		{
+			APIVersion:         "apps/v1",
+			Kind:               "Deployment",
+			Name:               operatorName,
+			UID:                types.UID("test-uid"),
+			Controller:         &trueVal,
+			BlockOwnerDeletion: &trueVal,
+		},
+	}
+}
+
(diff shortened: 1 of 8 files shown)


IshwarKanse:
cc @pavolloffay

tests/e2e-openshift/networkpolicy/00-assert-operator-network-policy.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: networking.k8s.io/v1
```
pavolloffay, line 1: could we move the e2e test to the normal test suite? The network policy is behind a feature gate and cane be enabled on k8s as well.

internal/operatornetworkpolicy/operatornetworkpolicy_test.go
```diff
@@ -0,0 +1,252 @@
...
+				},
+			},
+		},
+	}
+}
+
+// startAndCapture runs Start() in a goroutine and captures the created NetworkPolicy
+// by using a reactor that cancels the context after creation.
+func startAndCapture(t *testing.T, clientset *fake.Clientset, scheme *runtime.Scheme, opts ...Option) *networkingv1.NetworkPolicy {
+	t.Helper()
```
pavolloffay, line 57: what is this call doing?

pavolloffay:
Nice fix and great test coverage! One suggestion on the test style:

Consider asserting on the entire `NetworkPolicy` object rather than individual fields. For example:

```go
expected := &networkingv1.NetworkPolicy{
        Name:      "opentelemetry-operator",
        Namespace: namespace,
        // ...
    },
    Spec: networkingv1.NetworkPolicySpec{
        Egress: []networkingv1.NetworkPolicyEgressRule{
            {Ports: ..., To: []networkingv1.NetworkPolicyPeer{{IPBlock: &networkingv1.IPBlock{CIDR: "10.0.0.1/32"}}}},
            {Ports: ..., To: []networkingv1.NetworkPolicyPeer{{PodSelector: podSelector, NamespaceSelector: nsSelector}}},
        },
        Ingress: []networkingv1.NetworkPolicyIngressRule{{}},
        PolicyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress, networkingv1.PolicyTypeEgress},
        PodSelector: metav1.LabelSelector{MatchLabels: map[string]string{"app.kubernetes.io/name": "opentelemetry-operator"}},
    },
}
assert.Equal(t, expected, np)
```

This would improve readability — you see the full expected policy at a glance (similar to the e2e YAML assertions) and you can't accidentally miss asserting a field.

internal/operatornetworkpolicy/operatornetworkpolicy_test.go
IshwarKanse: t.Helper() method marks this function as a test helper. When a test assertion fails inside startAndCapture (e.g. the require.NoError or require.NotNil), Go reports the failure at the line in the calling test function rather than at the line inside this helper. It makes test failure output easier to read — you immediately see which test case failed without having to trace back from the helper.

tests/e2e-openshift/networkpolicy/00-assert-operator-network-policy.yaml
IshwarKanse: This bug is OpenShift-specific: the pod/namespace selectors for the API server egress are only added when OpenShiftRoutesAvailability == RoutesAvailable (cmd/operator/operator.go:372). On kind that condition is false, so only a single egress rule with ipBlock peers is created — the kind test can't trigger the invalid mixed-peer scenario this PR fixes. The OpenShift e2e test validates that the 2 separate egress rules are created correctly, which is the core regression check for this fix.

IshwarKanse:
> Nice fix and great test coverage! One suggestion on the test style:
> 
> Consider asserting on the entire `NetworkPolicy` object rather than individual fields. For example:
> 
> ```go
> expected := &networkingv1.NetworkPolicy{
>     ObjectMeta: metav1.ObjectMeta{
>         Name:      "opentelemetry-operator",
>         Namespace: namespace,
>         // ...
>     },
>     Spec: networkingv1.NetworkPolicySpec{
>         Egress: []networkingv1.NetworkPolicyEgressRule{
>             {Ports: ..., To: []networkingv1.NetworkPolicyPeer{{IPBlock: &networkingv1.IPBlock{CIDR: "10.0.0.1/32"}}}},
>             {Ports: ..., To: []networkingv1.NetworkPolicyPeer{{PodSelector: podSelector, NamespaceSelector: nsSelector}}},
>         },
>         Ingress: []networkingv1.NetworkPolicyIngressRule{{}},
>         PolicyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress, networkingv1.PolicyTypeEgress},
>         PodSelector: metav1.LabelSelector{MatchLabels: map[string]string{"app.kubernetes.io/name": "opentelemetry-operator"}},
>     },
> }
> assert.Equal(t, expected, np)
> ```
> 
> This would improve readability — you see the full expected policy at a glance (similar to the e2e YAML assertions) and you can't accidentally miss asserting a field.

Added the suggested changes.

pavolloffay (approved):

internal/operatornetworkpolicy/operatornetworkpolicy_test.go
pavolloffay: I see, the OCP NP slightly differs from the vanilla k8s one.

PR #5288, opened 2026-07-03
Fix, enable network policy on webhook ports

pavolloffay:

- Resolves: #issue-number

### DIFF
--- a/.chloggen/fix-networkpolicy-webhook-port.yaml
+++ b/.chloggen/fix-networkpolicy-webhook-port.yaml
@@ -4,7 +4,7 @@ change_type: bug_fix
 component: operator
 
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Always include webhook port in the operator NetworkPolicy regardless of ENABLE_WEBHOOKS setting, fixing CR creation failures when webhooks run in a separate deployment.
+note: Fix NetworkPolicy and operand network policy defaulting when webhooks run in a separate deployment.
 
 # One or more tracking issues related to the change
 issues: [5288]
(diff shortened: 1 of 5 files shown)


ozzywalsh (approved):

IshwarKanse:
I tested this PR on OpenShift with the operator deployed via OLM using the OpenShift bundle overlay, and found a related issue that this PR alone does not fix.

**The problem:** The OpenShift overlay defines a separate webhook deployment (`config/overlays/openshift/webhook-deployment.yaml`), but that deployment is missing the `FEATURE_GATES` env var. Without it, the webhook's `featuregate.EnableOperandNetworkPolicy.IsEnabled()` check in `collector_webhook.go` returns `false`, so `spec.networkPolicy.enabled` is never defaulted to `true` on new `OpenTelemetryCollector` CRs. As a result, even though the controller-manager has the feature gate enabled and creates the operator NetworkPolicy, operand NetworkPolicies are never created because the webhook never sets the field.

**Root cause:** In `config/overlays/openshift/webhook-deployment.yaml`, the webhook container's `env` list includes `ENABLE_NGINX_AUTO_INSTRUMENTATION` and `ENABLE_GO_AUTO_INSTRUMENTATION`, but not `FEATURE_GATES`. Meanwhile, the controller-manager patch in `config/overlays/openshift/manager-patch.yaml` correctly sets `FEATURE_GATES: "operator.networkpolicy,operand.networkpolicy"`.

**Fix:** Add the `FEATURE_GATES` env var to `config/overlays/openshift/webhook-deployment.yaml`:

```yaml
env:
- name: ENABLE_NGINX_AUTO_INSTRUMENTATION
  value: "true"
- name: ENABLE_GO_AUTO_INSTRUMENTATION
  value: "true"
- name: FEATURE_GATES
  value: "operand.networkpolicy"
```

**Verification:** After applying this fix and rebuilding the bundle, both the `networkpolicy` and `operator-restart` e2e tests pass — the webhook correctly defaults `spec.networkPolicy.enabled = true`, and the controller-manager creates the operand NetworkPolicies as expected.

Could this fix be included in this PR, or should it be a separate one?

pavolloffay:
@IshwarKanse I have added the fix here

frzifus (approved):

PR #5293, opened 2026-07-06
feat(target allocator): add OTLP self-telemetry API to CRDs

odubajDT:
CRD + operator wiring for the TargetAllocator OTLP self-telemetry feature:
- spec.targetAllocator.telemetry.metrics.otlp (OpenTelemetryCollector CR) and spec.telemetry.metrics.otlp (TargetAllocator CR), typed fields for endpoint, protocol (grpc/http), temporality, headers, TLS/insecure, export interval/timeout
- operator renders the telemetry block into the TargetAllocator ConfigMap
- generated CRDs, bundles, API docs, deepcopy

Refs open-telemetry#5047

Based on https://github.com/open-telemetry/opentelemetry-operator/pull/5068


swiatekm:
I think this configuration should at the very least be a subset of the declarative config, same as what we implemented for the target allocator itself. @pavolloffay as the person who investigated generating CRDs based on the declarative config spec, what do you think the easiest way to accomplish this is?

odubajDT:
> I think this configuration should at the very least be a subset of the declarative config, same as what we implemented for the target allocator itself. @pavolloffay as the person who investigated generating CRDs based on the declarative config spec, what do you think the easiest way to accomplish this is?

Hi, totally agree and already working on it. Honestly, I forgot to change the PR to draft, as I didn't manage to finish and push the changes yesterday. Sorry for that, will try to put it together today.

The version here still contained the old state of CRDs (before the TA binary OTLP selfmon changes were merged).

odubajDT:
@swiatekm should be ready :)

internal/manifests/targetallocator/configmap.go
```diff
@@ -152,6 +152,42 @@ func ConfigMap(params Params) (*corev1.ConfigMap, error) {
 		taConfig["collector_not_ready_grace_period"] = taSpec.CollectorNotReadyGracePeriod.Duration
 	}
 
+	if len(taSpec.Telemetry.Metrics.Readers) > 0 {
+		readers := make([]any, 0, len(taSpec.Telemetry.Metrics.Readers))
```
swiatekm: Have you tried serializing the whole struct using mapstructure? Or serialize to json, then deserialize to a map? I'd rather avoid this manual conversion if possible.

swiatekm (commented):
Can you add an e2e test that checks that the config renders correctly and that the target allocator exports the data to an otel collector?

internal/manifests/targetallocator/configmap.go
odubajDT: I followed the pattern that is already present, would it makes sense to do a major refactoring here afterwards?

swiatekm: We can keep it as is here, I was just curious if you tried it and walked back because of problems.

swiatekm:
@odubajDT all of this looks good to me. One thing missing is a test similar to what we had for plain target allocator config, where we load the CRD section serialized to yaml or json into the declarative config and verify that we're compatible. After you add that, I'll happily approve.

odubajDT:
> @odubajDT all of this looks good to me. One thing missing is a test similar to what we had for plain target allocator config, where we load the CRD section serialized to yaml or json into the declarative config and verify that we're compatible. After you add that, I'll happily approve.

Thank you! Added

swiatekm (approved):
LGTM, thank you for taking this on! 

Since this is a major addition to a CRD, I'd like to get at least one more approval before we merge. @open-telemetry/operator-approvers

iblancasa (approved):

swiatekm:
@odubajDT merging main will fix the govulncheck failures - we've just had the Go 1.26.6 release publish some fresh CVEs.

odubajDT:
Hi @swiatekm, are we good to merge? Thank you

swiatekm:
@odubajDT can you have a look at the e2e test failures?

For merging, you're making a large addition to our CRDs, so I wanted to give at least one fellow maintainer a chance to review. But if there's no objection by the end of the week, I'll merge this on Monday. Is that ok with you?

odubajDT:
> @odubajDT can you have a look at the e2e test failures?
> 
> For merging, you're making a large addition to our CRDs, so I wanted to give at least one fellow maintainer a chance to review. But if there's no objection by the end of the week, I'll merge this on Monday. Is that ok with you?

Sure, thank you!

apis/v1beta1/opentelemetrycollector_types.go
```diff
@@ -420,6 +427,90 @@ type MetricsConfigSpec struct {
 	DisablePrometheusAnnotations bool `json:"disablePrometheusAnnotations,omitempty"`
 }
 
+// TargetAllocatorTelemetry defines the self-telemetry configuration for the TargetAllocator.
+type TargetAllocatorTelemetry struct {
```
jaronoff97: In the future we may want to name this something more generic as I would imagine we would re-use this for other resources like the bridge

internal/manifests/targetallocator/configmap.go
```diff
@@ -170,6 +176,107 @@ func ConfigMap(params Params) (*corev1.ConfigMap, error) {
 	}, nil
 }
 
+// taTelemetryConfig and its nested types are private intermediate structs used
+// solely for YAML serialization. Their yaml tags match the TA binary's config
+// schema exactly, so no manual map construction is needed.
+
+type taTelemetryConfig struct {
+	Metrics *taMetricsConfig `yaml:"metrics,omitempty"`
+}
```
jaronoff97, line 185: can we not use the actual otel configuration go struct at this point and serialize that?

swiatekm: We can't, because target allocator itself does not support all of those options yet. That can be done in a follow-up. There were some complexities in retrofitting that struct to the target allocator while preserving its current behavior, so we left it alone for now.

iblancasa:
I think it would be a good idea to create a follow up ticket to add resource detection to the generated telemetry data, as we are adding to the collector or opamp supervisor

jaronoff97 (approved):

pavolloffay (approved):

swiatekm:
@odubajDT could you take care of the failing e2e tests? Once CI is green, we can merge.

odubajDT:
> @odubajDT could you take care of the failing e2e tests? Once CI is green, we can merge.

Already working on it, having some issue resolving the problems. If you would have a hint where the issue lies, it would be very helpful

PR #5294, opened 2026-07-06
feat(target allocator): export self-telemetry via OTLP

odubajDT:
TA binary side of the OTLP self-telemetry feature (no CRD changes):
- OTLP metric export from the TA config file (telemetry.metrics.otlp)
- Prometheus bridge so SD/Go/process metrics reach OTLP alongside /metrics
  (SDK metrics on a dedicated registry to avoid double counting; /metrics = union)
- service.name resource, graceful meter-provider shutdown
- ${env:VAR} expansion in header/endpoint values; gRPC accepts URL or host:port
- config validation + unit and end-to-end export tests

Based on https://github.com/open-telemetry/opentelemetry-operator/pull/5068

### DIFF
--- a/cmd/otel-allocator/internal/config/config.go
+++ b/cmd/otel-allocator/internal/config/config.go
@@ -537,6 +537,9 @@ func validateTelemetry(t TelemetryConfig) error {
 		if exp.OTLPGrpc == nil && exp.OTLPHttp == nil {
 			return fmt.Errorf("telemetry.metrics.readers[%d].periodic: must configure otlp_grpc or otlp_http exporter", i)
 		}
+		if exp.OTLPGrpc != nil && exp.OTLPHttp != nil {
+			return fmt.Errorf("telemetry.metrics.readers[%d].periodic: otlp_grpc and otlp_http are mutually exclusive, configure only one", i)
+		}
 		if exp.OTLPGrpc != nil {
 			if exp.OTLPGrpc.Endpoint == "" {
 				return fmt.Errorf("telemetry.metrics.readers[%d].periodic.otlp_grpc: endpoint must be set", i)
--- a/cmd/otel-allocator/internal/config/config.go
+++ b/cmd/otel-allocator/internal/config/config.go
@@ -114,39 +114,83 @@ type HTTPSServerConfig struct {
 	TLSKeyFilePath  string `yaml:"tls_key_file_path,omitempty"`
 }
 
-// TelemetryConfig mirrors the collector's service.telemetry structure for the Target Allocator.
+// TelemetryConfig holds the self-telemetry settings for the Target Allocator.
+// The Metrics field follows the OTel declarative configuration spec (MeterProvider
+// section), which means a telemetry YAML fragment can be validated with
+// otelconf.ParseYAML for compatibility checking.
 type TelemetryConfig struct {
-	Metrics MetricsConfig `yaml:"metrics,omitempty"`
+	Metrics *MetricsConfig `yaml:"metrics,omitempty"`
 }
 
-// MetricsConfig holds metric-export settings for the Target Allocator's own telemetry.
+// MetricsConfig mirrors otelconf's MeterProvider schema for the readers list.
 type MetricsConfig struct {
-	// OTLP configures an optional OTLP metric exporter. When set, metrics are
-	// exported via OTLP in addition to the existing Prometheus /metrics endpoint.
-	OTLP *OTLPExporterConfig `yaml:"otlp,omitempty"`
+	// Readers configures one or more metric readers, following the OTel
+	// declarative configuration spec.
+	Readers []MetricReader `yaml:"readers,omitempty"`
 }
 
-// OTLPExporterConfig holds connection settings for the OTLP metric exporter.
-type OTLPExporterConfig struct {
-	// Endpoint is the OTLP receiver base URL or address.
-	// For HTTP, provide the base URL without the signal path
-	// (e.g. "https://example.com/api/v2/otlp"); /v1/metrics is appended automatically.
-	// For gRPC, provide host:port (e.g. "example.com:4317").
-	Endpoint string `yaml:"endpoint"`
-	// Protocol selects the transport: "grpc" (default) or "http".
-	Protocol string `yaml:"protocol,omitempty"`
+// MetricReader mirrors otelconf's MetricReader type.
+type MetricReader struct {
+	Periodic *PeriodicMetricReader `yaml:"periodic,omitempty"`
+}
+
+// PeriodicMetricReader mirrors otelconf's PeriodicMetricReader type.
+// Interval and Timeout are in milliseconds, matching the otelconf spec.
+type PeriodicMetricReader struct {
+	// Interval is the delay between consecutive exports in milliseconds (default 60000).
+	Interval int `yaml:"interval,omitempty"`
+	// Timeout is the maximum allowed export duration in milliseconds (default 30000).
+	Timeout int `yaml:"timeout,omitempty"`
+	// Exporter configures the push exporter for this reader.
+	Exporter MetricExporter `yaml:"exporter"`
+}
+
+// MetricExporter mirrors otelconf's PushMetricExporter type.
+type MetricExporter struct {
+	// OTLPGrpc configures an OTLP/gRPC metric exporter.
+	OTLPGrpc *OTLPGrpcExporterConfig `yaml:"otlp_grpc,omitempty"`
+	// OTLPHttp configures an OTLP/HTTP metric exporter.
+	OTLPHttp *OTLPHttpExporterConfig `yaml:"otlp_http,omitempty"`
+}
+
+// OTLPGrpcExporterConfig mirrors otelconf's OTLPGrpcMetricExporter type.
+type OTLPGrpcExporterConfig struct {
+	// Endpoint is the gRPC receiver address. Accepts host:port or a full URL with scheme.
+	Endpoint string `yaml:"endpoint,omitempty"`
 	// Headers are additional key/value pairs sent with every export request.
-	Headers map[string]string `yaml:"headers,omitempty"`
+	// Values support ${env:VAR} substitution.
+	Headers []NameValuePair `yaml:"headers,omitempty"`
+	// TemporalityPreference sets aggregation temporality: "cumulative" (default),
+	// "delta", or "low_memory".
+	TemporalityPreference string `yaml:"temporality_preference,omitempty"`
+	// Tls configures TLS for the gRPC connection.
+	Tls *GrpcTlsConfig `yaml:"tls,omitempty"`
+}
+
+// OTLPHttpExporterConfig mirrors otelconf's OTLPHttpMetricExporter type.
+type OTLPHttpExporterConfig struct {
+	// Endpoint is the OTLP/HTTP receiver base URL (e.g. "https://example.com:4318").
+	// /v1/metrics is appended automatically unless already present.
+	// Values support ${env:VAR} substitution.
+	Endpoint string `yaml:"endpoint,omitempty"`
+	// Headers are additional key/value pairs sent with every export request.
+	// Values support ${env:VAR} substitution.
+	Headers []NameValuePair `yaml:"headers,omitempty"`
+	// TemporalityPreference sets aggregation temporality: "cumulative" (default),
+	// "delta", or "low_memory".
+	TemporalityPreference string `yaml:"temporality_preference,omitempty"`
+}
+
+// NameValuePair mirrors otelconf's NameStringValuePair type for HTTP/gRPC headers.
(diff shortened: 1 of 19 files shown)


cmd/otel-allocator/main.go
```diff
@@ -77,12 +84,42 @@ func main() {
 		os.Exit(1)
 	}
 
-	metricExporter, promErr := otelprom.New()
+	// The OTel SDK metrics are exported to Prometheus through a dedicated registry
```
swiatekm: Let's move all the telemetry setup code to `telemetry.go`. We already have separate tests for it, and the main file is growing large and messy.

swiatekm (commented):
Conceptually, the changes look good and make sense to me. The main issue now is how do make them configurable for the user.

The end result I'd ultimately like to achieve is that the otel declarative config is the source of truth for all telemetry of the otel allocator, and the existing prometheus endpoint more of a legacy option for backwards compatibility.

There's three questions I'm trying to answer right now:

1. Can we use the declarative config and customize the resulting sdk with the prometheus bridge and exporter?
2. If so, is it better to embed the declarative config or keep a separate config file.
3. If the above is not feasible, too complex, or we don't want to do it in this PR, then can we make our config compatible with the declarative spec?

odubajDT:
Thanks! Moved all the telemetry setup out of `main.go` into `telemetry.go`

On the three questions, having dug into `otelconf`:

**1. Can we use the declarative config and customize the resulting SDK with the Prometheus bridge + exporter?**

The OTLP exporter side is fully covered by otelconf - `NewSDK(WithOpenTelemetryConfiguration(...))` builds the meter provider with the periodic OTLP reader straight from the config.

The Prometheus bridge is the catch. otelconf builds the readers internally and doesn't expose a way to attach a `sdkmetric.Producer` to them. `WithMeterProviderOptions` lets you append SDK options, but adding a separate reader carrying the bridge producer would double-collect the SDK metrics, since every reader on a meter provider collects all of its metrics. So we can't cleanly graft the bridge onto an otelconf-built SDK now.

**2. Embed vs. separate config file?**

For the plain TA config I don't think a separate file is needed; embedding the declarative config in the existing config file is consistent with how the collector does it, and keeps everything in one place.

**3. If (1) isn't clean, can we make our config compatible with the declarative spec?**

Yes, and I think that might be the sweet spot for this PR - adopt otelconf's config types as the schema for the `telemetry` block and parse through otelconf (so its validation + env substitution run), but keep building the meter provider ourselves so the Prometheus bridge and the `/metrics` endpoint keep working. That makes the config forward-compatible with the declarative spec now, and leaves a clean migration to full `NewSDK` later.

The genuinely clean end state for declarative config is to stop registering the metrics on the Prometheus default registry and register them on the OTel SDK instead - then the bridge disappears and `NewSDK` covers everything, with the Prometheus endpoint reduced to the legacy `pull` reader. That's a larger change, so I'd suggest it as a follow-up rather than blocking this PR.

So my proposal - keep this PR as the working feature, and land it as-is with our current config and do the otelconf-schema swap next

Let me know what you think!

swiatekm:
@odubajDT sorry for the wait. I'm ok with keeping the mechanics of this PR as they are, and making the config compatible with the declarative config. It'd be nice to have a smoke test where we load a test yaml file into declarative config to verify that this works.

odubajDT:
@swiatekm thanks! should be adapted and ready for another round of review.

note: the CI failures seem unrelated to these changes, as they are currently failing on `main` as well

cmd/otel-allocator/telemetry.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+// telemetryResource builds the OpenTelemetry resource attached to exported metrics.
+// It sets a default service.name and honors the standard OTEL_SERVICE_NAME /
+// OTEL_RESOURCE_ATTRIBUTES environment variables, so additional attributes (e.g. k8s.*)
+// can be injected without code changes.
+func telemetryResource(ctx context.Context) (*resource.Resource, error) {
+	return resource.New(ctx,
+		resource.WithTelemetrySDK(),
+		resource.WithAttributes(semconv.ServiceName("target-allocator")),
```
callegarimattia: `telemetryResource()` only sets `service.name`, not `service.instance.id`. With more than one TA replica (or once logs/traces get resources too), there's no way to tell instances apart in a backend. Worth defaulting it to the pod hostname? 
```suggestion
		resource.WithAttributes(
		  semconv.ServiceName("target-allocator")
		  semconv.ServiceInstanceID(hostname),
		),
```

cmd/otel-allocator/internal/config/config.go
```diff
@@ -439,6 +519,45 @@ func ValidateConfig(config *Config) error {
...
+			}
+		}
+		if exp.OTLPHttp != nil {
+			if exp.OTLPHttp.Endpoint == "" {
+				return fmt.Errorf("telemetry.metrics.readers[%d].periodic.otlp_http: endpoint must be set", i)
+			}
+			switch exp.OTLPHttp.TemporalityPreference {
+			case "", "cumulative", "delta", "low_memory":
+			default:
+				return fmt.Errorf("telemetry.metrics.readers[%d].periodic.otlp_http: temporality_preference must be 'cumulative', 'delta', or 'low_memory', got %q", i, exp.OTLPHttp.TemporalityPreference)
```
callegarimattia, line 560: in the case in which both are set, the grpc is used without any log message or validation error. should we clarify it somehow?

cmd/otel-allocator/telemetry.go
```diff
@@ -0,0 +1,311 @@
...
+			}
+			deltaDPs := make([]metricdata.DataPoint[float64], 0, len(sum.DataPoints))
+			for _, dp := range sum.DataPoints {
+				key := deltaSumKey{
+					scope:  scope,
+					metric: m.Name,
+					attrs:  fmt.Sprintf("%v", dp.Attributes),
+				}
+				prev, seen := d.prev[key]
+				d.prev[key] = prevSumPoint{value: dp.Value, t: dp.Time}
```
callegarimattia: the deltaproducer prev map is unbounded and never cleaned up. isn't that going to break for long running TA?

odubajDT:
Thanks @callegarimattia for the review! Adapted

callegarimattia (approved):

cmd/otel-allocator/telemetry.go
```diff
@@ -0,0 +1,330 @@
...
+// forwarded as-is. For the Prometheus bridge this means counters arrive as
+// cumulative sums. This wrapper converts them to delta so that backends
+// configured to expect delta temporality receive the correct data.
+//
+// The prev map grows to one entry per unique {scope, metric, attribute-set}
+// combination that has ever been observed. For a stable Prometheus scrape target
+// (fixed label cardinality, no ephemeral label values such as pod IPs) this set
+// is bounded in practice. High-cardinality label churn would cause unbounded
+// growth; in that case the Prometheus scrape target itself would already be a
+// problem, so this is an acceptable tradeoff for the current use case.
```
swiatekm: I'd be much more comfortable with this PR if we could avoid this in general. It's very hard for me to determine if we're missing any edge cases in this implementation. Is there really no way to do this using existing processors?

odubajDT: `deltaProducer` is necessary because the `PeriodicReader` intentionally does not apply the `TemporalitySelector` to external `Producer` output. Prometheus counters have fixed cumulative temporality by definition and the SDK does not rewrite them. The key metrics that need bridging are registered on the default Prometheus registry by the SD library itself (`discovery.CreateAndRegisterSDMetrics(prometheus.DefaultRegisterer)`). That code is not ours and there is no OTel SDK API to inject there. A Collector-side `cumulativetodelta` processor would only help if there were a Collector in the path, which this feature is specifically designed to avoid - to enable direct export to backends accepting only delta metrics.

The clean long-term fix is to wrap the SD library's registerer with one that bridges into the SDK, or upstream a change to accept an OTel meter. Once SD metrics are SDK-native the bridge and `deltaProducer` both disappear and `TemporalitySelector` works end-to-end. That is a larger change better suited as a follow-up.

For reference, this is how the Collector handles it: It registers all internal metrics natively on the OTel SDK from the start, so `TemporalitySelector` applies directly without any bridge. The Collector can do this because it does not use the Prometheus SD library, but has its own scraper pipeline. The TA's entire purpose is the Prometheus SD engine, so it cannot take the same approach today.

cmd/otel-allocator/telemetry.go
```diff
@@ -0,0 +1,330 @@
...
+			sum, ok := m.Data.(metricdata.Sum[float64])
+			if !ok || !sum.IsMonotonic || sum.Temporality != metricdata.CumulativeTemporality {
+				continue
+			}
+			deltaDPs := make([]metricdata.DataPoint[float64], 0, len(sum.DataPoints))
+			for _, dp := range sum.DataPoints {
+				key := deltaSumKey{
+					scope:  scope,
+					metric: m.Name,
+					attrs:  fmt.Sprintf("%v", dp.Attributes),
```
swiatekm: Are you sure this is correct? I'd feel more confident if we had a helper that looks at `dp.Attributes.All()` and writes their `AsString()` to a hasher.

cmd/otel-allocator/telemetry.go
```diff
@@ -0,0 +1,330 @@
...
+				}
+				prev, seen := d.prev[key]
+				d.prev[key] = prevSumPoint{value: dp.Value, t: dp.Time}
+				if !seen {
+					// First observation: emit with delta = current value so the
+					// backend sees the counter from the start rather than skipping it.
+					startTime := dp.StartTime
+					if startTime.IsZero() {
+						startTime = dp.Time
+					}
```
swiatekm: Is this surgery defensive or something you've actually seen be needed in real production data?

cmd/otel-allocator/telemetry_test.go
```diff
@@ -0,0 +1,177 @@
...
+		// An empty body is a valid ExportMetricsServiceResponse.
+		w.WriteHeader(http.StatusOK)
+		select {
+		case received <- struct{}{}:
+		default:
+		}
+	}))
+	defer srv.Close()
+
+	ctx := context.Background()
```
swiatekm:
```suggestion
	ctx := t.Context()
```
?

cmd/otel-allocator/internal/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,177 @@
...
+	headers := expandHeaders([]config.NameValuePair{{Name: "Authorization", Value: &val}})
+	assert.Equal(t, "Api-Token secret-value", headers["Authorization"])
+	assert.Nil(t, expandHeaders(nil))
+}
+
+// TestOTLPExportEndToEnd starts a mock OTLP/HTTP receiver and verifies that a recorded
+// metric is actually exported, that the endpoint's signal path is used, and that a header
+// value referencing an environment variable is expanded before being sent.
+func TestOTLPExportEndToEnd(t *testing.T) {
+	t.Setenv("TA_TEST_TOKEN", "secret-value")
```
swiatekm, line 112: The deltaProducer is the most substantial addition in this PR and it's far from obvious, so I'd like to see more unit tests for it, covering all the obvious cases.

cmd/otel-allocator/telemetry.go
swiatekm: I'm willing to accept it if we add more tests for it. More unit tests at the very minimum. In an ideal world, there should be a e2e test for the standalone TA (we have these already) verifying that some metric from the SD has correct values when exporter via otlp. You might even be able to write that in-process without involving a K8s cluster.

cmd/otel-allocator/telemetry.go
odubajDT: defensive, teoretically it can happen, but then we have an issue somewhere else. Removing it, thanks.

cmd/otel-allocator/telemetry.go
odubajDT: thanks! added more unit test including an e2e test covering the delta conversion with real data increments

swiatekm:
@odubajDT these changes look good to me. One final request I have is that we've added a fair amount of new code, and it'd be nice to move all of it into a new `telemetry` package. Just moving the files and updating imports is good enough. Fix the dependency conflict too, and we're good to go.

odubajDT:
> @odubajDT these changes look good to me. One final request I have is that we've added a fair amount of new code, and it'd be nice to move all of it into a new `telemetry` package. Just moving the files and updating imports is good enough. Fix the dependency conflict too, and we're good to go.

Thanks! All fixed

swiatekm (approved):
Thank you for the contribution, and for your patience with the review process!

odubajDT:
> Thank you for the contribution, and for your patience with the review process!

Thank you for your input, appreciate it!

PR #5305, opened 2026-07-10
[chore] docs: add manager configuration reference

### LINKED ISSUE #3565: Document all flags associated with manager

amarkdotdev:
## Summary

- Adds `docs/reference/configuration.md` documenting manager settings across YAML config file, environment variables, and CLI flags
- Links the new page from `docs/reference/README.md`
- Fixes `--tls-min-version` CLI parsing (flag name mismatch in `ApplyCLI`)
- Adds a unit test for the TLS min version flag

## Context

#3565 asks for documented, uniform configuration surfaces for the operator manager. The harmonization refactor added CLI/env parity for most settings, but no reference doc existed on main. This PR adds that reference and fixes a small CLI bug discovered while documenting TLS settings.

## Test plan

- [x] `go test ./internal/config/ -run TestTLSMinVersionFlag -count=1`
- [ ] Review tables against `internal/config/{config,cli,env}.go`

Fixes #3565


internal/config/cli.go
```diff
@@ -147,8 +147,8 @@ func ApplyCLI(cfg *Config) error {
 				cfg.TLS.UseClusterProfile, _ = f.GetBool("tls-cluster-profile")
 			case "tls-configure-operands":
 				cfg.TLS.ConfigureOperands, _ = f.GetBool("tls-configure-operands")
-			case "min-tls-version":
```
iblancasa, line 153: This is a breaking change. I would prefer to avoid it.

docs/reference/configuration.md
```diff
@@ -0,0 +1,163 @@
...
+
+- `open-shift-routes-availability`
+- `gateway-apis-availability`
+- `prometheus-cr-availability`
+- `cert-manager-availability`
+- `target-allocator-availability`
+- `collector-availability`
+- `opampbridge-availability`
+
+HTTP(S) proxy settings (`HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`) are captured from the operator pod environment and propagated to managed workloads; they are not part of the YAML schema.
```
swiatekm: This is not configuration, it's operator implementation details. It doesn't need to be here.

docs/reference/configuration.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Helm chart
+
+When deploying with Helm, set `manager.config` in chart values to inject a YAML snippet mounted via `--config-file`. Precedence matches the standalone operator: chart defaults, then file, then env, then any extra manager args.
```
swiatekm: This should be in the helm chart documentation instead.

docs/reference/configuration.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Source of truth
+
+Configuration structs and defaults: [`internal/config/config.go`](https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/config/config.go).
+
+Future auto-generated flag documentation is tracked in [docs/rfcs/documentation.md](https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/rfcs/documentation.md) (Phase 3).
```
swiatekm: I'd much rather do this from the start, it's going to be annoying to retrofit it afterwards. The format you have right now looks very structured, can't we generate it?

.chloggen/manager-configuration-reference.yaml
```diff
@@ -0,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
```
swiatekm, line 1: I don't think we need a changelog entry for documentation improvements.

internal/config/cli.go
```diff
@@ -147,8 +147,8 @@ func ApplyCLI(cfg *Config) error {
...
-				cfg.TLS.MinVersion, _ = f.GetString("min-tls-version")
+			case "tls-min-version":
+				cfg.TLS.MinVersion, _ = f.GetString("tls-min-version")
 			case "tls-cipher-suites":
```
swiatekm, line 155: Seconding Israel's request not to change this. Let's just document what it is right now.

docs/reference/README.md
```diff
@@ -1,5 +1,6 @@
 # Reference
 
 - [API reference](../api/README.md)
+- [Manager configuration](configuration.md)
```
swiatekm:
```suggestion
- [Operator configuration](configuration.md)
```

amarkdotdev:
@swiatekm @iblancasa Pushed an update: dropped the changelog entry, reverted the cli.go change, removed the runtime-only and helm sections from the reference page, and fixed the README link. This should be doc-only now.

amarkdotdev:
@swiatekm @iblancasa Thanks for the generate-docs pushback — I looked at whether we can do that in this PR.

Today the only docs generation path in-tree is `make api-docs` (CRD → markdown via `crdoc`). That doesn't cover manager configuration: YAML keys, CLI flags, and env vars are registered separately in `internal/config/{config,cli,env}.go`, with several intentional name mismatches (e.g. `--create-sm-operator-metrics` vs `create-service-monitor-operator-metrics`, image flag/env naming). There's no single metadata source we can feed a generator without a larger refactor.

So this page stays hand-written for now, scoped as a docs-only `[chore]` reference. A realistic later path would be:

1. Introduce a shared field registry (or struct tags) for yaml / flag / env / default / description / section
2. Drive `CreateCLIParser` + `ApplyEnvVars` from that registry
3. Add something like `make config-docs` that emits `docs/reference/configuration.md` (or a generated fragment) and wire it into `ensure-update-is-noop`

Happy to open a follow-up issue/PR for that if you want it tracked. Also left the existing `--tls-min-version` / `ApplyCLI` mismatch alone per your earlier feedback (document current behavior, no breaking rename in this PR).

pavolloffay:
Could we find an approach to keep this up-to-date? I would prefer to generate parts of this docs from the source code or at least validate.

swiatekm:
> Could we find an approach to keep this up-to-date? I would prefer to generate parts of this docs from the source code or at least validate.

I think this PR is an improvement over the status quo, which is that this isn't documented at all. There might be some code changes needed before we can generate parts of it automatically, and I think those can land independently. So I'd be inclined to merge this rather than block on the lack of automation. Do you have objections against the content?

amarkdotdev:
@pavolloffay agreed on generation/validation as a follow-up. flags are not in the crdoc path today, so this keeps a hand-maintained reference for now.

swiatekm (approved):

amarkdotdev:
rebased onto main, branch was carrying stale merge artifacts

PR #5306, opened 2026-07-10
feat(operator-opamp-bridge): Add support for the capability AcceptsRestartCommand

syepes:
[operator-opamp-bridge](https://github.com/open-telemetry/opentelemetry-operator/tree/main/cmd/operator-opamp-bridge)

Adds support for the OpAMP capability AcceptsRestartCommand is now fully implemented in both operator and standalone modes, the OnCommand is wired and Restart is on the ConfigApplier interface, workload rollout patch is implemented for all collector modes.

This feature now works as the [supervisor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/cmd/opampsupervisor) and allows remote restarting of the collector


cmd/operator-opamp-bridge/internal/agent/agent.go
```diff
@@ -452,6 +453,23 @@ func (agent *Agent) Shutdown() {
...
+	case protobufs.CommandType_CommandType_Restart:
+		agent.logger.Info("Received restart command, triggering collector restart")
+		if err := agent.applier.Restart(ctx); err != nil {
+			agent.logger.Error(err, "Failed to restart collector on command")
+			return err
+		}
+		return nil
+	default:
+		agent.logger.Info("Received unsupported command type, ignoring", "type", command.GetType().String())
+		return nil
```
swiatekm: Is there a way for us to return an error here instead, notifying the server that we don't support the command? Dropping these silently is not great.

cmd/operator-opamp-bridge/internal/operator/client.go
```diff
@@ -218,6 +226,86 @@ func (c Client) Delete(key string) error {
...
+		} else {
+			joinedErr = fmt.Errorf("%w; %w", joinedErr, e)
+		}
+	}
+	return joinedErr
+}
+
+// triggerRollout patches the pod template restart annotation on the collector's managed workload,
+// causing Kubernetes to perform a rolling restart identical to `kubectl rollout restart`.
+// Sidecar mode collectors have no standalone workload, so they are skipped.
```
swiatekm: Why aren't we plainly doing a rollout restart via the API as opposed to patching the template? Is there any additional benefit to the kubectl method? The standalone client already has a `triggeRollout` method that does this, so we could generalize it.

cmd/operator-opamp-bridge/internal/agent/agent.go
syepes: Yes, that would be useful

cmd/operator-opamp-bridge/internal/operator/client.go
syepes: Effectively, this can be modularize so that it can be used by other components

syepes:
@atoulme what do you know about the last chances? do you think there is something else that we need to adjust so that this suits the standards or missing functionalities?

swiatekm:
Looks like the integration tests are failing for some definite reason, though I'm not sure at a glance what it is.

syepes:
@swiatekm 
> Looks like the integration tests are failing for some definite reason, though I'm not sure at a glance what it is.

Found the issue, it was related to the extra k8s RBAC needed for the restarts. If this gets merged I will also update the RBAC in the [helm chart](https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-operator)

swiatekm:
The changes look good to me. There's two things I'd like you to add:
- A changelog entry. Run `make chlog-new` and fill in the template.
- A E2E test actually using the capability. If you don't know how to get started with this, feel free to ask questions here or on the CNCF Slack, we'll get you on the right track.

syepes:
@swiatekm just added the dedicated tests and chlog

tests/e2e-opampbridge/standalone/chainsaw-test.yaml
```diff
@@ -119,9 +121,8 @@ spec:
               )
 
               printf '%s' "$remote_config" | kubectl run -n "$NAMESPACE" send-valid-remote-config \
-                --image=mirror.gcr.io/alpine/curl:latest \
-                --restart=Never \
-                --rm -i \
+                --image=mirror.gcr.io/curlimages/curl:8.20.0 \
```
swiatekm: Why are you changing this image? We've specifically set it to what it is for a good reason, see https://github.com/open-telemetry/opentelemetry-operator/pull/5301. These tests need to run on some slightly obscure architectures. This comment applies to other changes in this PR where you use the same image, too.

syepes: Resolved with [6c3a510](https://github.com/open-telemetry/opentelemetry-operator/pull/5306/commits/6c3a51042df7cb45c8390d1eadf68fb12d69711c)

tests/e2e-opampbridge/opampbridge/chainsaw-test.yaml
```diff
@@ -40,7 +50,6 @@ spec:
               (length(values($result))): 1
               (values($result)[0].status.sequence_num >= `2`): true
               (values($result)[0].status.health.healthy): true
-              (length(values($result)[0].effective_config)): 0
```
swiatekm, line 61: Why was this removed?

tests/e2e-opampbridge/opampbridge/chainsaw-test.yaml
```diff
@@ -65,58 +74,70 @@ spec:
...
-              # it would be great if I could wildcard for the initial instance id key and then
-              # construct a compound key of `$NAMESPACE/simplest` for the collector I want to assert
               (length(values($result))): 1
-              (values($result)[0].status.sequence_num >= `2`): true
               (values($result)[0].status.health.healthy): true
-              (length(values($result)[0].status.health.component_health_map)): 1
-              (values(values($result)[0].status.health.component_health_map)[0].healthy): true
-              (length(values($result)[0].effective_config)): 1
-              (length(values($result)[0].status.effective_config.config_map.config_map)): 1
-              (length(values(values($result)[0].status.effective_config.config_map.config_map)[0].body) > `0`): true
```
swiatekm, line 120: Why should any of these assertions change? We haven't changed anything about the setup.

tests/e2e-opampbridge/opampbridge/chainsaw-test.yaml
```diff
@@ -137,16 +158,163 @@ spec:
...
+              set -eu
+              kubectl run -n "$NAMESPACE" send-restart-cmd \
+                --image=mirror.gcr.io/alpine/curl:latest \
+                --restart=Never --rm -i \
+                --command -- curl -fsS -X POST \
+                "http://e2e-test-app-bridge-server:4321/agents/send-command" \
+                -H "Content-Type: application/json" \
+                -d '{"command_type":"restart"}'
+        # Poll until the restartedAt annotation appears on the collector Deployment.
+        - script:
```
swiatekm, line 211: You don't need a script action for this, it's a resource attribute. You can instead put the expected Deployment metadata here and chainsaw will wait for it correctly.

tests/e2e-opampbridge/opampbridge/chainsaw-test.yaml
```diff
@@ -137,16 +158,163 @@ spec:
...
+    # Scenario 3: Bridge remains healthy and connected after processing the
+    #             restart command (bridge itself does not restart; it only
+    #             triggers a rollout on the managed collector).
+    # ---------------------------------------------------------------------------
+    - catch:
+        - podLogs:
+            selector: app=e2e-test-app-bridge-server
+        - podLogs:
+            selector: app.kubernetes.io/component=opentelemetry-opamp-bridge
+      name: "scenario 3: bridge remains healthy after restart command"
```
swiatekm, line 244: If we're checking if the bridge itself is healthy, can't we do it through the Custom Resource status? Or even just the bridge Deployment?

tests/e2e-opampbridge/opampbridge/chainsaw-test.yaml
```diff
@@ -137,16 +158,163 @@ spec:
...
+    # ---------------------------------------------------------------------------
+    - catch:
+        - podLogs:
+            selector: app=e2e-test-app-bridge-server
+        - podLogs:
+            selector: app.kubernetes.io/component=opentelemetry-opamp-bridge
+      name: "scenario 2: unknown command is rejected, collector not restarted"
+      try:
+        # Record the current restartedAt timestamp before sending the unknown command.
+        - script:
```
swiatekm: I don't think we need to check this. It's an awkward test, because we're asserting something will not happen, so we have to wait some bounded amount of time. Instead we should simply check if the bridge returns an error when given an unsupported command. If this is annoying to do here, I'll accept not having a test for it.

tests/e2e-opampbridge/standalone/chainsaw-test.yaml
```diff
@@ -169,6 +170,7 @@ spec:
       name: verify valid remote config status
       try:
         - script:
+            timeout: 75s
```
swiatekm: Why are these timeouts necessary?

swiatekm (commented):
I think the code proper is good, but I had some comments and questions about the e2e tests.

tests/e2e-opampbridge/opampbridge/chainsaw-test.yaml
syepes: Effectively this is still needed

tests/e2e-opampbridge/opampbridge/chainsaw-test.yaml
syepes: Restore the original asserts

tests/e2e-opampbridge/opampbridge/chainsaw-test.yaml
syepes: Simplified the test by looking at the Deployment replicas

tests/e2e-opampbridge/opampbridge/chainsaw-test.yaml
syepes: Remove unneeded test

tests/e2e-opampbridge/opampbridge/chainsaw-test.yaml
syepes: Resolved

tests/e2e-opampbridge/standalone/chainsaw-test.yaml
swiatekm: @syepes Please revert the other ones too, unless you have a good reason to set them.

syepes: Done

swiatekm (approved):
LGTM, thank you for your patience with the review process! @jaronoff97 can you have a quick look as well?

syepes:
@jaronoff97 Any news on this one?

jaronoff97 (approved):
apologies for the delay, didn't see this in my inbox. this looks great. give me a ping when the merge conflicts are done so i can merge

syepes:
@jaronoff97 We should be good now, thanks!

PR #4983, opened 2026-04-22
fix(processors): register k8s_attributes alongside k8sattributes for RBAC

SAY-5:
## Summary

Fixes #4922.

open-telemetry/opentelemetry-collector-contrib#45901 renamed the processor to the snake-case `k8s_attributes` while keeping the original `k8sattributes` spelling accepted by the collector. The operator only registered the camel spelling, so a collector configured with `k8s_attributes` produced no RBAC and the collector logged `pods is forbidden` / `replicasets is forbidden` at runtime against the default `node-collector` service account.

## Fix

Register a second `Parser` for `k8s_attributes` under the same `K8sAttributeConfig` type and `GenerateK8SAttrRbacRules` generator. `k8sattributes` keeps its existing registration so both spellings continue to work — the collector accepts either, and now so does the operator's RBAC pass.

## Tests

- New testdata fixture `rbac_k8s_attributes_snake_case.yaml` (same content as `rbac_k8sattributes_service_name.yaml`, but using the snake-case processor name).
- New `TestDesiredClusterRoles` table entry asserting the same RBAC rules are generated for the snake-case spelling.

`go test ./internal/components/processors/... ./internal/manifests/collector/...` green; `go vet` clean.

PR #5317, opened 2026-07-13
Add new snake_case spellings of remaining receivers and processors

agagniere:
**Description:** 

Add new snake_case spellings for:
- `kubelet_stats`
- `k8s_objects`
- `resource_detection`
- `fluent_forward`
- `tcp_log`
- `udp_log`
- `ssh_check`
- `cloud_foundry`
- `http_check`
- `flink_metrics`

Part of https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45339

Similar to #4983

The Operator looks for components in the config in the CR, using their legacy name without an underscore.
We must add the new name to allow both names to be used.

### DIFF
--- a/internal/components/receivers/helpers.go
+++ b/internal/components/receivers/helpers.go
@@ -120,17 +120,12 @@ var componentParsers = []components.Parser{
 		WithProtocol(corev1.ProtocolTCP).
 		WithTargetPort(3100).
 		MustBuild(),
-	components.NewBuilder[kubeletStatsConfig]().WithName("kubeletstats").
-		WithRbacGen(generateKubeletStatsRbacRules).
-		WithEnvVarGen(generateKubeletStatsEnvVars).
-		MustBuild(),
-	// kubelet_stats is the snake-case alias introduced in
-	// open-telemetry/opentelemetry-collector-contrib#47957. Both names
-	// resolve to the same receiver in the collector, so the operator must
-	// generate the same RBAC and env vars for either spelling.
+	// kubelet_stats, formerly kubeletstats
+	// (open-telemetry/opentelemetry-collector-contrib#47957).
 	components.NewBuilder[kubeletStatsConfig]().WithName("kubelet_stats").
 		WithRbacGen(generateKubeletStatsRbacRules).
 		WithEnvVarGen(generateKubeletStatsEnvVars).
+		WithAlias("kubeletstats").
 		MustBuild(),
 	components.NewBuilder[k8seventsConfig]().WithName("k8s_events").
 		WithRbacGen(generatek8seventsRbacRules).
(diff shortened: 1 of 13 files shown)


swiatekm:
I'm not fond of the builder repetition. Could we create a shared builder and then just `.WithName(...).MustBuild()` twice? We can move the initialization into a function to avoid defining more global variables.

internal/components/receivers/helpers.go
```diff
@@ -124,6 +124,14 @@ var componentParsers = []components.Parser{
 		WithRbacGen(generateKubeletStatsRbacRules).
 		WithEnvVarGen(generateKubeletStatsEnvVars).
 		MustBuild(),
+	// kubelet_stats is the snake-case alias introduced in
+	// open-telemetry/opentelemetry-collector-contrib#47957. Both names
+	// resolve to the same receiver in the collector, so the operator must
+	// generate the same RBAC and env vars for either spelling.
+	components.NewBuilder[kubeletStatsConfig]().WithName("kubelet_stats").
```
jaronoff97, line 135: very minor suggestion, but rahter than duplicating we could also just add another builder method called "WithAlias" so we don't need to duplicate

jaronoff97: given this is the standard for collector components we could also try to automate this by attempting both paths programmatically

agagniere: Agreed, I just replicated what was done in the previous PR, but now that it is used at 4 places, it's time to factorize the logic

swiatekm (approved):
LGTM, thank you for fixing this!

internal/components/receivers/helpers.go
```diff
@@ -120,18 +120,24 @@ var componentParsers = []components.Parser{
 		WithProtocol(corev1.ProtocolTCP).
 		WithTargetPort(3100).
 		MustBuild(),
-	components.NewBuilder[kubeletStatsConfig]().WithName("kubeletstats").
+	// kubelet_stats, formerly kubeletstats
+	// (open-telemetry/opentelemetry-collector-contrib#47957).
+	components.NewBuilder[kubeletStatsConfig]().WithName("kubelet_stats").
 		WithRbacGen(generateKubeletStatsRbacRules).
 		WithEnvVarGen(generateKubeletStatsEnvVars).
+		WithAlias("kubeletstats").
```
jaronoff97, line 138: should we automate what the alias could be by just removing the `_` and changing the other names?

swiatekm: I'd rather be explicit. I'm not sure we can assume this - to an extent, it's an arbitrary decision by the collector SIG. Hopefully this will be the only time we're going to have component renames, now that the naming convention is set.

agagniere: @jaronoff97 
I understand how your suggestion would handle future renames.
But I believe that introducing a different logic than the collector would confuse both users and maintainers

jaronoff97 (approved):

PR #5322, opened 2026-07-14
target allocator: add consistent-hashing-endpoint allocation strategy

### LINKED ISSUE #5321: target allocator: add a consistent-hashing-endpoint strategy keyed on the full scrape endpoint
**Is your feature request related to a problem? Please describe.**

The `consistent-hashing` target allocator strategy keys its hash ring on `__address__` alone (`item.TargetURL`). Any scrape endpoints that share a host:port but are distinguished only by metrics path or query params - for example, multiple exporters running behind the same proxy/port, or a single host scraped under several distinct paths - always hash to the same value and therefore always land on the same collector, regardless of how many collector replicas exist.

This defeats the point of choosing `consistent-hashing` for those workloads: instead of spreading scrape load across the target allocator's replicas, every endpoint sharing a `host:port` piles onto a single collector.

**Describe the solution you'd like**

Add a new allocation strategy, `consistent-hashing-endpoint`, that keys the hash ring on the full scrape endpoint - address, scheme, metrics path, and query params - rather than on the address alone. Endpoints sharing a host:port but differing by path or params would then spread across collectors as expected.

To avoid churn from the more sensitive key, the hash intentionally excludes labels that don't affect the actual scrape URL (instance metadata, service-discovery annotations, etc.), so mutable label changes don't reshuffle assignments - only changes to the effective scrape endpoint do.

This is proposed as a **new, separate strategy** rather than a change to the existing `consistent-hashing` behavior, so current users relying on address-only hashing see no behavior change; the new strategy is opt-in via `allocationStrategy: consistent-hashing-endpoint`.

**Describe alternatives you've considered**

- Changing `consistent-hashing`'s existing key in place - rejected, since it would silently reassign targets for existing users on upgrade.
- Hashing on the full target identity (all relabeled labels + job name) - tried this first, but it turned out to be too churn-sensitive: any label change (including labels unrelated to the scrape URL) reassigns the target to a different collector. Scoping the key to just the scrape-URL-relevant labels keeps the spreading benefit without that downside.

I have an implementation ready (new strategy file + unit tests + CRD enum/doc updates) and would like to open a PR referencing this issue.

yonatan-ess:
## Summary

Fixes #5321.

`consistent-hashing` keys its ring on `__address__` alone, so scrape endpoints that share a host:port but are distinguished only by metrics path or query params (e.g. multiple exporters behind the same proxy, or one host scraped under several paths) always land on the same collector.

This adds a new, opt-in `consistent-hashing-endpoint` strategy that hashes the full scrape endpoint (address, scheme, metrics path, and query params) instead, spreading those endpoints across collectors. The key intentionally excludes labels that don't affect the scrape URL (instance metadata, service-discovery annotations, etc.), so mutable label churn doesn't reshuffle assignments.

It's added as a **new strategy** rather than a change to `consistent-hashing`'s existing behavior, so current users relying on address-only hashing see no change on upgrade.

- New `cmd/otel-allocator/internal/allocation/consistent_hashing_endpoint.go` + unit tests
- Registered in the strategy map; added as a valid `allocationStrategy` enum value in both `v1alpha1` and `v1beta1` CRDs
- Treated the same as `consistent-hashing` for HA-related defaulting (PodDisruptionBudget)
- `make update` regenerated CRDs/bundles/docs

## Test plan

- [x] `go test ./cmd/otel-allocator/...` passes, including the new strategy-specific tests and the existing cross-strategy test suite (`RunForAllStrategies`)
- [x] `make precommit` (fmt, vet, lint, test, ensure-update-is-noop) passes clean


yonatan-ess:
This PR was developed with Claude Code assistance (implementation, tests, and CRD/doc regeneration), reviewed and submitted by me.

vadyanuch13-ai (approved):

swiatekm (changes requested):
I'd prefer to make the existing strategy configurable rather than add a new one. Let's discuss in the issue how this should work.

yonatan-ess:
> I'd prefer to make the existing strategy configurable rather than add a new one. Let's discuss in the issue how this should work.

Makes sense, will follow up in the issue and rework this PR once we agree on the config shape.

PR #5334, opened 2026-07-20
Allow nginx and apache instrumentations to send data to IP

### LINKED ISSUE #5333: Allow Apache and Nginx Autoinstrumentations to send data without using a service
### Component(s)

auto-instrumentation

Apache and NGINX instrumentations only work when you are using a service to send the telemetry data. In our case, we send the data to our Daemonset via Node IP and not using a Service.

The init containers should receive the Instrumentation spec to read the `OTEL_NODE_IP` environment variable to send the data there.

_No response_

### Additional context

_No response_

### Tip

iblancasa:

- Resolves: #5333

**Testing:**
- Added new unit tests


swiatekm (changes requested):
I'm not sure we want to do this when a solution using K8s primitives is available. See my comment on the issue.

iblancasa:
> I'm not sure we want to do this when a solution using K8s primitives is available. See my comment on the issue.

I disagree. That solution forces to use a `Service`. I don't want to use a `Service`. I want to send the data to an OTLP endpoint that is in the same node but I don't have a `Service` for that.

My specific use case is that I'm deploying the operator without the CRDs to use the instrumentation mechanism. And I'm deploying a Collector on my own. As a `DaemonSet`. And I don't want to create a `Service` just for that.

swiatekm:
> > I'm not sure we want to do this when a solution using K8s primitives is available. See my comment on the issue.
> 
> I disagree. That solution forces to use a `Service`. I don't want to use a `Service`. I want to send the data to an OTLP endpoint that is in the same node but I don't have a `Service` for that.
> 
> My specific use case is that I'm deploying the operator without the CRDs to use the instrumentation mechanism. And I'm deploying a Collector on my own. As a `DaemonSet`. And I don't want to create a `Service` just for that.

My view is that these nginx/httpd scripts are among the least maintainable parts of the whole codebase and I would really like to avoid adding any more logic to them than is strictly necessary. So the bar for acceptance is quite high. I'd prefer to look at other approaches first. Or, at the very least, have a more generic mechanism. What you want to do - using an environment variable in the definition of an exporter endpoint - is valid, and we should probably support that for all Instrumentations if we aren't already. I'd accept a PR wanting to support interpolation in general for Apache/HTTPD. But a hack that just hardcodes two specific ones in the script, I do not want.

iblancasa:
> I'd prefer to look at other approaches first.

I'm totally open to that. I implemented this in the best way I could find.

> Or, at the very least, have a more generic mechanism.

I will try to do something more generic for the different environment variables.

internal/instrumentation/helper.go
```diff
@@ -171,3 +171,9 @@ func setContainersFromAnnotation(inst *instrumentationWithContainers, annotation
 	inst.Containers = append(inst.Containers, languageContainers...)
 	return nil
 }
+
+func initContainerUserEnv(required, languageEnv, commonEnv []corev1.EnvVar) []corev1.EnvVar {
+	env := append([]corev1.EnvVar{}, languageEnv...)
+	env = appendIfNotSet(env, commonEnv...)
+	return append(env, required...)
+}
```
swiatekm, line 180: I think we have a helper function for other instrumentations which merges these in a more clever way. You need to be a bit careful because if the user sets one the variables the operator sets, we want to use theirs rather than ours. I get the feeling the apache/httpd instrumentation didn't get those bug fixes.

swiatekm (commented):
That looks much better, thanks for making it happen! The logic seems sound, I just want to make sure we're not reproducing annoying bugs we've fixed for other instrumentations.

swiatekm (approved):

pavolloffay (approved):

Issue #5255, opened 2026-06-26
Require instrumentation version to be specified by users in Instrumentation CR


jack-berg:
### Component(s)

auto-instrumentation

### Is your feature request related to a problem? Please describe.

Follow to this [slack conversation](https://cloud-native.slack.com/archives/C033BJ8BASU/p1782416961364339).

Right now the version of language instrumentations is optional. This couples operator version scheme to the collective version scheme of all the language instrumentations it supports. 

We saw this when the opentelemetry java agent went to 2.x and there was a long delay in upgrading the version of the java instrumentation image to 2.x. Doing so meant updating the default, which meant exposing users to those breaking changes who rely on the default.

The OpenTelemetry project has a goal of instrumentations to be stable by default. That means marking things as 1.x rather then the defacto stable 0.x we've been going with. Maintainers will only feel comfortable marking things 1.x if there is an allowance to have a 2.x. And so we need to expect regular major version bumps from instrumentation packages, which will include breaking changes.

The plan is to release opentelemetry java agent 3.x next month. And 4.x will come probably sometime 12-24 months after that. Other languages will likely follow suit since opentelemetry java agent's release pattern of regular major version updates with support for `LATEST_MAJOR_VERSION - 1` is being touted as a pattern for other languages to follow.

The operator doesn't want to be in the business of cutting a new major version for each new major version of any instrumentation. And it doesn't want to be in the business of holding users back from using the latest versions of instrumentations while it figures out when its appropriate to update the new default to the next major version.

### Describe the solution you'd like

The only solution I can think of is to force users to specify the version of instrumentations in their instrumentation CRs.

Proposal:

- Update the Instrumentation CRD to require users specify the version of language instrumentations
- Setup automation to update and publish new instrumentation images for each and every release of respective language instrumentation. No curation decisions. The images are just a repackaging of the instrumentations, with lockstep versioning schemes.
- Any snippets in the README or elsewhere in the repo to help users get started should have the latest language instrumentation versions included, such that users can simply copy and paste the CR. 
- Setup the language instrumentation version update process to automatically update the versions referenced in these snippets.

With this, operator can publish updates with a versioning scheme independent of language instrumentation releases. Users can confidently update their operator version without any impact to the schema of the telemetry produced by instrumentation. The UX impact of these required fields in minimal because nobody writes CRs from scratch - everyone copy / pastes from the examples in github, which now just have a few extra lines.

ps-mir:
[...]

PR #5345, opened 2026-07-21
Add instrumentation v1beta1 go spec

pavolloffay:

First part of https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/rfcs/instrumentation-v1beta1.md

* This PR adds only the Go struct without enabling the CRD or adding it to the bundle. 
* The fields for the declarative config will be added in a separate PR.

Updates https://github.com/open-telemetry/opentelemetry-operator/issues/5346 


swiatekm (approved):

apis/v1beta1/instrumentation_types.go
```diff
@@ -0,0 +1,287 @@
...
+	// Enabled controls whether service identity attributes are automatically derived.
+	// When false, no service.* attributes are added by the operator. Defaults to true.
+	// +optional
+	Enabled *bool `json:"enabled,omitempty"`
+}
+
+// CommonLanguageSpec contains fields shared by all language-specific configurations.
+type CommonLanguageSpec struct {
+	// Image is a container image with auto-instrumentation.
+	// +optional
```
jaronoff97, line 178: IMO we should make this required and non-optional

apis/v1beta1/instrumentation_types.go
```diff
@@ -0,0 +1,287 @@
...
+
+	// VolumeClaimTemplate defines an ephemeral volume used for auto-instrumentation.
+	// If omitted, an emptyDir is used with a default size limit.
+	// +optional
+	VolumeClaimTemplate corev1.PersistentVolumeClaimTemplate `json:"volumeClaimTemplate,omitempty"`
+
+	// Env defines language-specific env vars.
+	// Precedence: original container env > language-specific env > common env > SDK config.
+	// +optional
+	Env []corev1.EnvVar `json:"env,omitempty"`
```
jaronoff97: Similarly, I thought we would get rid of these entirely? There's going to be even more ways to set environment variables now, and i think we should simplify these here.

apis/v1beta1/instrumentation_types.go
```diff
@@ -0,0 +1,287 @@
...
+
+// Nginx defines Nginx SDK and instrumentation configuration.
+type Nginx struct {
+	CommonLanguageSpec `json:",inline"`
+
+	// Attrs defines Nginx agent specific attributes. The precedence order is:
+	// `agent default attributes` > `instrument spec attributes` .
+	// Attributes are documented at https://github.com/open-telemetry/opentelemetry-cpp-contrib/tree/main/instrumentation/otel-webserver-module
+	// +optional
+	Attrs []corev1.EnvVar `json:"attrs,omitempty"`
```
jaronoff97: same here, shouldn't these just be environment variables?

jaronoff97 (commented):
a few thoughts from our recent discussion

apis/v1beta1/instrumentation_types.go
pavolloffay: Let's tackle it separately as part of the https://github.com/open-telemetry/opentelemetry-operator/issues/5255. @TylerHelmuth is also interested in this. 

I have added the issue to v1beta1 milestone

apis/v1beta1/instrumentation_types.go
pavolloffay: >  There's going to be even more ways to set environment variables now

Do you mean the declarative config? 

I think we should move the `Env []corev1.EnvVar `json:"env,omitempty"` under `spec.envConfig` because the env vars are not applicable in the declarative config:

```go

// EnvConfig defines the env-var-based SDK configuration.
type EnvConfig struct {
	// Exporter defines exporter configuration.
	// +optional
	Exporter Exporter `json:"exporter,omitempty"`

	// Propagators defines inter-process context propagation configuration.
	// +optional
	Propagators []Propagator `json:"propagators,omitempty"`

	// Sampler defines sampling configuration.
	// +optional
	Sampler Sampler `json:"sampler,omitempty"`

       Env []corev1.EnvVar `json:"env,omitempty"
}
```

jaronoff97 (approved):

PR #5347, opened 2026-07-21
Use otel injector

swiatekm:
**Description:**

Add a feature flag to switch to the otel injector for injecting instrumentation. The injector is included in instrumentation images unconditionally.

- Resolves: #issue-number


tests/e2e-instrumentation-injector/instrumentation-dotnet/01-assert.yaml
```diff
@@ -0,0 +1,102 @@
...
+          fieldPath: status.podIP
+    - name: ASPNETCORE_URLS
+      value: http://+:8080
+    - name: OTEL_METRIC_EXPORT_INTERVAL
+      value: "30000"
+    - name: OTEL_TRACES_SAMPLER
+      value: always_on
+    - name: LD_PRELOAD
+      value: /otel-auto-instrumentation-dotnet/libotelinject.so
+    - name: DOTNET_AUTO_INSTRUMENTATION_AGENT_PATH_PREFIX
```
mmanciop, line 29: The injector supports setting this via files too, and I think it’d make the change less cluttered

mmanciop (commented):
I have not tried the PR, just read the diff. Looks legit, but I think the biggest gap is documentation, especially in explaining why users should opt in to the Injector (in this PR is mostly resilience against modifications to `JAVA_TOOL_OPTIONS` and similar env vars in the entrypoint of the container).

swiatekm:
> I have not tried the PR, just read the diff. Looks legit, but I think the biggest gap is documentation, especially in explaining why users should opt in to the Injector (in this PR is mostly resilience against modifications to `JAVA_TOOL_OPTIONS` and similar env vars in the entrypoint of the container).

It's intended as a PoC, so it's deliberately not included. In all honesty though, this particular change is an implementation detail in my view. Users don't have to know, or care, and in an ideal world they won't even notice it happened, aside from some old bugs being resolved.

mmanciop:
> > I have not tried the PR, just read the diff. Looks legit, but I think the biggest gap is documentation, especially in explaining why users should opt in to the Injector (in this PR is mostly resilience against modifications to `JAVA_TOOL_OPTIONS` and similar env vars in the entrypoint of the container).
> 
> It's intended as a PoC, so it's deliberately not included. In all honesty though, this particular change is an implementation detail in my view. Users don't have to know, or care, and in an ideal world they won't even notice it happened, aside from some old bugs being resolved.

I agree, it feels like the kind of feature gate that gets flipped on by default a few months down the line.

tests/e2e-instrumentation-injector/instrumentation-dotnet/01-assert.yaml
swiatekm: But then I'd need to set an env variable for the config file location (since it has to be on a shared volume), so it's the same number of variables, but the image can more easily control it. I like that because it would make it easier for users to build their own images.

pavolloffay:
By how much does it increase the image sizes?

swiatekm:
> By how much does it increase the image sizes?

~ 3MB uncompressed. To me, the main argument in favor of keeping the injector in a separate image is that it will make building custom images easier. If the injector comes separately, the only condition a custom image would need fulfill would be putting an injector-compatible artifact in the right path. If we bake the injector into our images, users building custom images would need to do it too.

PR #5348, opened 2026-07-21
[chore] Add a new testing/e2e package 

swiatekm:
**Description:**

The purpose of this package is to hold helpers for writing e2e tests against Kubernetes clusters in Go. The intent is to migrate tests currently using painful amounts of bash in chainsaw to this framework, where semantic assertions about data are much easier to make.

Other than adding the package, we migrate the existing standalone target allocator e2e tests to it.


internal/testing/e2e/names_test.go
yuriolisa, line 1: Am I wrong, or do these tests never actually run? The e2e build tag is needed here because the package itself is tagged, but nothing executes them, once the `make e2e-ta-standalone` only runs `./tests/e2e-ta-standalone/...`, and make test builds without the tag so the package doesn't even exist there.

internal/testing/e2e/repo.go
yuriolisa, line 1: small nit: using config/target-allocator/clusterrole.yaml ties a generic RepoRoot() to a TA-specific file. Would go.mod survive any config layout reshuffle?

yuriolisa (approved):

internal/testing/e2e/repo.go
swiatekm: It could reach a false positive, but I think that's the only risk? If that file moves, it's good that the tests fail. I suppose I could shell out to git, think that's a better option?

internal/testing/e2e/names_test.go
swiatekm: Good catch, I removed the build tag, it's not necessary for the package itself, just the tests proper.

internal/testing/e2e/repo.go
yuriolisa: True, for that reason I've commented as a nit. LGTM.

Issue #2181, opened 2023-09-29
Use opentelemetry.io* namespace for extracing resource and attribute labels from Kubernetes manifests


austinlparker:
Currently, the Operator infers certain OpenTelemetry resource values (like service.name) by looking at various labels on the manifest (https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/instrumentation/sdk.go#L352-L371) such as Deployment Name, Stateful Set Name, etc. 

I would like for OpenTelemetry to reserve the `*.opentelemetry.io/*` namespace (as in `resource.opentelemetry.io/service.version` or `attribute.opentelemetry.io/my.cool.attributte`), and have the operator (and various resource detectors) parse this namespace in order to set the associated values.

jaronoff97:
Yeah i think we could use these and give them priority over what's there currently. Would also make it easier to inject resource attributes generally. Would we allow these to use environment variables? I'm not sure how that would work

Horiodino:
hi @austinlparker @jaronoff97   i would love to work on this issue , i have used the API and manifests so many times.
so it wont be trouble for me .
can i work on it ?

jaronoff97:
@Horiodino that would be great, thank you!

Horiodino:
> @Horiodino that would be great, thank you!

Thank you @jaronoff97  I will start working on it right now🙂.

Horiodino:
@jaronoff97 so i need to reserve a namespace so all the resources with mentioned labels falls under it .
can you  provide the file to make changes .

pavolloffay:
> Currently, the Operator infers certain OpenTelemetry resource values (like service.name) by looking at various labels on the manifest

The operator is not looking at k8s object labels. It queries API server and gets deployment name, RS name through owner references.  On top of that users can define additional attributes in the CR https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1alpha1/instrumentation_types.go#L87


> I would like for OpenTelemetry to reserve the *.opentelemetry.io/* namespace (as in resource.opentelemetry.io/service.version or attribute.opentelemetry.io/my.cool.attributte),

@austinlparker  @jaronoff97 is there a spec documenting these values?  I am not sure I fully understand the proposal in this ticket. Is the intention to change there resouce attribute names used by the operator, or make the operator to support these new `opentelemetry.io` attbiutes by propagating them from labels/annotations?

Horiodino:
[...]

PR #5350, opened 2026-07-22
Added arm64 support for .NET instrumentation

bhogayatakb:
The autoinstrumentation-dotnet image is already built and published for both linux/amd64 and linux/arm64 (see
  .github/workflows/publish-autoinstrumentation-dotnet.yaml), but the Dockerfile doesn't use buildx's TARGETARCH build arg, so it always downloads the x64 .NET
  auto-instrumentation artifacts regardless of the platform being built. In practice this means the arm64 variant of the image has always silently shipped an
  x64 native profiler binary (OpenTelemetry.AutoInstrumentation.Native.so), which fails to load on arm64 nodes.

  This follows the direction proposed by @swiatekm on #3270:
  1. Make the image genuinely multi-platform by using ARG TARGETARCH to select the correct .NET RID artifact (x64 or arm64) for the platform actually being
  built.
  2. Expose the native binaries under architecture-agnostic linux / linux-musl directories, since the correct architecture is now resolved when the multi-arch
  manifest is pulled — not by the webhook.
  3. Keep linux-x64 / linux-musl-x64 as symlinks to linux / linux-musl for backward compatibility, so CORECLR_PROFILER_PATH values already in use (and the
  instrumentation.opentelemetry.io/otel-dotnet-auto-runtime annotation values) don't change.

  As a result, the operator's webhook doesn't need to know a pod's target CPU architecture at admission time (which it generally can't know reliably anyway) —
  Kubernetes/containerd already resolve that automatically via the image's multi-arch manifest when the pod is scheduled.

  Changes:
  - autoinstrumentation/dotnet/Dockerfile: download only the TARGETARCH-appropriate glibc/musl artifacts, extract to linux/linux-musl, symlink
  linux-x64/linux-musl-x64 for compatibility.
  - internal/instrumentation/dotnet.go: point CORECLR_PROFILER_PATH at the new arch-agnostic paths. No change to the supported annotation values (linux-x64,
  linux-musl-x64 remain the only two, and both now work correctly on amd64 and arm64 nodes).
  - docs/auto-instrumentation/languages/dotnet.md: note that the image is multi-arch and the runtime annotation only distinguishes glibc vs musl, not CPU
  architecture.

  Link to tracking Issue(s):

  - Resolves: #3270 (and the underlying request in open-telemetry/opentelemetry-dotnet-instrumentation#2314, which was blocked on
  open-telemetry/opentelemetry-dotnet-instrumentation#1865 and #2181 — both now resolved upstream)


docs/auto-instrumentation/languages/dotnet.md
```diff
@@ -10,4 +10,6 @@ instrumentation.opentelemetry.io/otel-dotnet-auto-runtime: "linux-x64" # for Lin
 instrumentation.opentelemetry.io/otel-dotnet-auto-runtime: "linux-musl-x64"  # for Linux musl based images
 ```
 
+> **Note:** The auto-instrumentation image is published for both `linux/amd64` and `linux/arm64`. The runtime annotation above only distinguishes glibc vs musl based application images; the CPU architecture of the actual native profiler binary used is resolved automatically when the image is pulled onto the node, so no separate arm64 annotation values are needed.
+
```
swiatekm: We don't need this comment here, it just explains how multi-architecture docker images work. If there's some broader confusion about this among users, it should be in the generic instrumentation section, rather than under dotnet.

internal/instrumentation/dotnet.go
```diff
@@ -22,18 +22,23 @@ const (
...
-	dotNetAdditionalDepsPath            = "/otel-auto-instrumentation-dotnet/AdditionalDeps"
-	dotNetOTelAutoHomePath              = "/otel-auto-instrumentation-dotnet"
-	dotNetSharedStorePath               = "/otel-auto-instrumentation-dotnet/store"
-	dotNetStartupHookPath               = "/otel-auto-instrumentation-dotnet/net/OpenTelemetry.AutoInstrumentation.StartupHook.dll"
-	dotnetVolumeName                    = volumeName + "-dotnet"
-	dotnetInitContainerName             = initContainerName + "-dotnet"
-	dotnetInstrMountPath                = "/otel-auto-instrumentation-dotnet"
+	// The image's native profiler binaries live under architecture-agnostic linux/linux-musl
+	// directories; the actual amd64 or arm64 binary is selected when the multi-arch image is
+	// pulled, not by this webhook. See autoinstrumentation/dotnet/Dockerfile.
```
swiatekm: This is how all the webhooks work and how image selection works in general incontainer runtimes. We don't need to document it here.

autoinstrumentation/dotnet/Dockerfile
```diff
@@ -5,27 +5,41 @@
...
+    esac && \
+    wget -q -O glibc.zip "https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/releases/download/v${version}/opentelemetry-dotnet-instrumentation-linux-glibc-${dotnetArch}.zip" && \
+    wget -q -O musl.zip "https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/releases/download/v${version}/opentelemetry-dotnet-instrumentation-linux-musl-${dotnetArch}.zip" && \
+    unzip glibc.zip && \
+    unzip musl.zip "linux-musl-${dotnetArch}/*" -d . && \
+    mv "linux-${dotnetArch}" linux && \
+    mv "linux-musl-${dotnetArch}" linux-musl && \
+    ln -s linux "linux-${dotnetArch}" && \
+    ln -s linux-musl "linux-musl-${dotnetArch}" && \
+    rm glibc.zip musl.zip && \
```
swiatekm: I'd prefer not to have this whole thing in a single bash command, it can be hard to debug when something fails.

internal/instrumentation/dotnet.go
```diff
@@ -22,18 +22,23 @@ const (
...
+	dotNetSharedStorePath          = "/otel-auto-instrumentation-dotnet/store"
+	dotNetStartupHookPath          = "/otel-auto-instrumentation-dotnet/net/OpenTelemetry.AutoInstrumentation.StartupHook.dll"
+	dotnetVolumeName               = volumeName + "-dotnet"
+	dotnetInitContainerName        = initContainerName + "-dotnet"
+	dotnetInstrMountPath           = "/otel-auto-instrumentation-dotnet"
 )
 
 // Supported .NET runtime identifiers (https://learn.microsoft.com/en-us/dotnet/core/rid-catalog), can be set by instrumentation.opentelemetry.io/inject-dotnet.
+// Both values work on amd64 and arm64 nodes: the image is multi-arch and the operator does not
+// need to know the target pod's CPU architecture.
```
swiatekm: Same here.

swiatekm (commented):
Thanks for taking this up! The content looks mostly correct to me, but some of the comments aren't necessary in my view.

docs/auto-instrumentation/languages/dotnet.md
bhogayatakb: Dropped the comment from the PR, as suggested.

internal/instrumentation/dotnet.go
bhogayatakb: Removed the comment from the PR, as suggested.

internal/instrumentation/dotnet.go
bhogayatakb: Removed the comment from the PR, as suggested.

autoinstrumentation/dotnet/Dockerfile
bhogayatakb: Updated Dockerfile so that each command is under seperate RUN statement, so we can debug easily.

autoinstrumentation/dotnet/Dockerfile
```diff
@@ -5,28 +5,43 @@
...
 #  - For auto-instrumentation by container injection, the Linux command cp is
 #    used and must be available in the image.
+#  - This image is built for linux/amd64 and linux/arm64. The .NET Runtime Identifier (RID) specific
+#    artifact for the target platform is downloaded and exposed under the architecture-agnostic
+#    `linux`/`linux-musl` directories, with `linux-x64`/`linux-musl-x64` kept as symlinks for
+#    backward compatibility with the CORECLR_PROFILER_PATH values used before multi-arch support
+#    was added.
 
-FROM busybox:1.38.0@sha256:fd8d9aa63ba2f0982b5304e1ee8d3b90a210bc1ffb5314d980eb6962f1a9715d AS downloader
+FROM --platform=$BUILDPLATFORM busybox:1.38.0@sha256:fd8d9aa63ba2f0982b5304e1ee8d3b90a210bc1ffb5314d980eb6962f1a9715d AS downloader
```
swiatekm, line 26: What exactly happens when we drop this addition? I don't see why it should be necessary.

bhogayatakb: This was not part of original commit, however the build check was failing.
Some of the wget commands were not fetching the right binaries.

However build might be failing due to one of the two reasons.
1. The Dockerfile changes were earlier into a common RUN statement
2. --platform=$BUILDPLATFORM

Let me drop this change, and if build passes then there is no necessity.

pavolloffay (approved):

iblancasa (approved):

autoinstrumentation/dotnet/Dockerfile
bhogayatakb: The build crashed again, after I reverted this change

bhogayatakb: @swiatekm 
I think we will have to stick to the `--platform=$BUILDPLATFORM`
because the workflow is running via QEMU
So, arm64 workflows are also running on `busybox:amd64`, which is not working well for `wget`

Earlier, this issue was not present, as Dockerfile used to have `ADD`
However now `wget` is required as we want to take targetarch dynamically based on the `case`

Let me know what do you think ?
Should I add this back ? `--platform=$BUILDPLATFORM`

swiatekm: It looks like a problem originating in QEMU, interestingly. You can add it back and I'll approve the PR, but add a comment explaining why it needs to exist.

bhogayatakb: Added --platform setting with explanation, as suggested

swiatekm (approved):

PR #5356, opened 2026-07-22
Enable mTLS for TA irrespective of cert-manager

### LINKED ISSUE #3982: Enable mtls irrespective of certManager resource
### Component(s)

_No response_

## Description

Trying to enable mtls between TA and collector without the support of certManager
We have our own cert management service to issue certs and mount these secrets.

## Steps to Reproduce
Enable mtls flag in operator
Override tls config in prometheus receiver under target_allocator:
by pointing the cert locations

## Expected Result
Without the dependency on certManager , operator shall enable mtls and 
use the certs provided via the config

## Actual Result
As certManager resource is not present, operator is overriding the config
with no-tls config (http://<ta>:80)

### Kubernetes Version

1.23.0

### Operator version

0.123.0

### Collector version

0.125.0

## Environment
OS: (e.g., "Ubuntu 20.04")

### Log output

```shell

```

### Additional context

args:
   enable-https-server: "true"
   https-ca-file: /run/secrets/ta-ca/ca.crt
   https-tls-cert-file: /run/secrets/ta/certs/cert.crt
   https-tls-key-file: /run/secrets/ta/certs/cert.key
  volumes:
   - name: ta-certs
     secret:
        secretName: ta-cert-secret
   - name: ta-ca-certs
     secret:
        secretName: ta-ca-secret
  volumeMounts:
  - name: ta-certs
    mountPath: /run/secrets/ta/certs
    readOnly: true
  - name: ta-ca-certs
    mountPath: /run/secrets/ta-ca/certs
    readOnly: true

config in prometheus receiver:
prometheus:
    target_allocator:
       endpoint: http://localhost:8080
       timeout: 5s
       tls:
          ca_file: /run/secrets/ca/ca.crt
         cert_file: /run/secrets/client-cert/tls.crt
          key_file: /run/secrets/client-cert/tls.key
          interval: 30s
          collector_id: collector-1        

and the respective certs are mounted from secrets without certManager resource

iypetrov:

OpenTelemetry Operator can only enable mTLS between the Target Allocator and the Collector when cert-manager is available - it provisions the certificates via cert-manager `Issuer`/`Certificate` CRDs. There is no way to point the Operator at externally-managed certificates.

- Resolves: https://github.com/open-telemetry/opentelemetry-operator/issues/3982

**Testing:**
Added unit-test coverage for the new user-provided-cert paths (helpers, volume/mount wiring, and webhook validation). All new tests pass alongside the existing suite.

**Documentation:**
A new `mtls.tls` block lets you provide the mTLS certificates directly when `useCertManager` is set to `false`. Example configuration:
- Standalone TargetAllocator CR:
```yaml
kind: TargetAllocator
metadata:
  name: ta
spec:
  mtls:
    # Enable mTLS between the Target Allocator and the Collector.
    enabled: true
    # false = provide certs yourself (via tls: below); true/omitted = use cert-manager (default).
    useCertManager: false
    # Only consulted when useCertManager is false.
    tls:
      # CA used to verify the peer. Required in this mode. Exactly one of secret/configMap.
      certificateAuthorityCertificate:
        configMap:
          name: my-ca                 # required
          key: ca.crt                 # optional, default: ca.crt
        # ...or from a Secret instead:
        # secret:
        #   name: my-ca
        #   key: ca.crt               # optional, default: ca.crt
      # Target Allocator's HTTPS server cert + key (may be in different Secrets).
      serverCertificate:
        certificateSecret:
          name: my-server             # required
          key: tls.crt                # optional, default: tls.crt
        keySecret:
          name: my-server             # required
          key: tls.key                # optional, default: tls.key
      # Collector's client cert + key (used when talking to the TA).
      clientCertificate:
        certificateSecret:
          name: my-client             # required
          key: tls.crt                # optional, default: tls.crt
        keySecret:
          name: my-client             # required
          key: tls.key                # optional, default: tls.key
```
- OpenTelemetryCollector with embedded TargetAllocator:
```yaml
kind: OpenTelemetryCollector
metadata:
  name: collector-with-ta
spec:
  targetAllocator:
    enabled: true
    mtls:
      enabled: true
      useCertManager: false
      tls:
        certificateAuthorityCertificate:
          configMap:
            name: my-ca               # CA from a ConfigMap (or use secret: instead)
        serverCertificate:
          certificateSecret:
            name: my-server
          keySecret:
            name: my-server
        clientCertificate:
          certificateSecret:
            name: my-client
          keySecret:
            name: my-client
  config:
    receivers:
      prometheus:
        config:
          scrape_configs:
            - job_name: otel-collector
              static_configs:
                - targets: ["0.0.0.0:8888"]
    exporters:
      debug: {}
    service:
      pipelines:
        metrics:
          receivers: [prometheus]
          exporters: [debug]
```


iypetrov:
@swiatekm can you rerun the failing tests, seems that there was a problem with e2e cluster setup
```bash
❯ gh run view --job=89149486257 --log | grep ERROR
run-e2e-tests / e2e-tests (1.25, e2e-ta-collector-mtls)	UNKNOWN STEP	2026-07-23T07:58:24.5489671Z ERROR: failed to create cluster: failed to pull image "kindest/node:v1.25.16@sha256:6110314339b3b44d10da7d27881849a87e092124afab5956f2e10ecdb463b025": command "docker pull kindest/node:v1.25.16@sha256:6110314339b3b44d10da7d27881849a87e092124afab5956f2e10ecdb463b025" failed with error: exit status 1
```

apis/v1beta1/opentelemetrycollector_types.go
```diff
@@ -268,6 +268,50 @@ type TargetAllocatorMTLS struct {
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	SecretName string `json:"secretName"`
+	// DataKeyCertificate is the key within the Secret's data that holds the certificate.
+	// Defaults to tls.crt.
+	// +optional
+	// +kubebuilder:default:=tls.crt
+	DataKeyCertificate string `json:"dataKeyCertificate,omitempty"`
+	// DataKeyKey is the key within the Secret's data that holds the private key.
+	// Defaults to tls.key. It is not required for CA certificate references.
```
swiatekm: Can we add a separate struct for CA certificate, if the key isn't required? It's a bit more fuss for us, but I think clarity for users is worth it.

internal/manifests/manifestutils/mtls.go
```diff
@@ -22,3 +30,185 @@ func IsTAMTLSCertManagerEnabled(ta *v1alpha1.TargetAllocator, cfg config.Config)
...
+		return ref.DataKeyKey
+	}
+	return constants.TACollectorTLSKeyFileName
+}
+
+// ValidateTAMTLS validates the target allocator mTLS configuration. When mTLS relies on cert-manager
+// it requires cert-manager to be available. When cert-manager is disabled it requires the user to
+// provide the server and client certificate Secrets; the CA certificate may either be referenced
+// separately or bundled in the leaf Secrets under the ca.crt key.
+func ValidateTAMTLS(ta *v1alpha1.TargetAllocator, certManagerAvailable bool) error {
```
swiatekm: This is only run in webhooks, right? Might be easier to only take the mtls section as an argument. On which note, do all the functions this calls need the full TargetAllocator struct, or just the mtls section?

swiatekm (commented):
The changes look good to me, had some minor bookkeeping comments. One thing we're missing that I'd like to include are e2e tests. The equivalent of what we have for cert-manager mode should do - might even be able to reuse a lot of the plumbing.

apis/v1beta1/opentelemetrycollector_types.go
```diff
@@ -268,6 +268,50 @@ type TargetAllocatorMTLS struct {
...
+	// +optional
+	TLS *TargetAllocatorTLS `json:"tls,omitempty"`
+}
+
+// TargetAllocatorTLS references user-provided Secrets holding the certificates used for mTLS
+// between the target allocator and the collector. Each reference is independent, so different Secrets
+// may back the CA, certificate and key. The referenced keys are projected into the pods via subPath
+// volume mounts, which means certificate rotation requires the pods to be restarted.
+type TargetAllocatorTLS struct {
+	// CertificateAuthorityCertificate references a Secret containing the CA certificate used to
```
pavolloffay: It's also common to distribute CA bundle via a config map

apis/v1beta1/opentelemetrycollector_types.go
iypetrov: Done — added a dedicated `CAReference` struct (no `dataKeyKey`, since the CA reference only needs the certificate). The leaf server/client certs continue to use `CertificateReference`. Regenerated the CRDs and API docs so the `certificateAuthorityCertificate` field no longer exposes `dataKeyKey`.

internal/manifests/manifestutils/mtls.go
iypetrov: Done — `ValidateTAMTLS` (and `IsTAMTLSEnabled` / `IsTAMTLSCertManagerEnabled`) now take just the `*v1beta1.TargetAllocatorMTLS` section instead of the whole `TargetAllocator`. The volume-building helpers (`TAServerCertificateVolumes`/`TAClientCertificateVolumes`) still take the full struct because they need the resource name for the Secret/volume naming. Updated all callers accordingly.

apis/v1beta1/opentelemetrycollector_types.go
```diff
@@ -268,6 +268,50 @@ type TargetAllocatorMTLS struct {
...
+	// +optional
+	ServerCertificate *CertificateReference `json:"serverCertificate,omitempty"`
+	// ClientCertificate references a Secret containing the client certificate and key used by the
+	// collector when talking to the target allocator's HTTPS server.
+	// +optional
+	ClientCertificate *CertificateReference `json:"clientCertificate,omitempty"`
+}
+
+// CertificateReference points to a certificate (and optionally its private key) stored in a Secret.
+type CertificateReference struct {
```
pavolloffay: I wonder if we should make this more flexible and allow consuming the cert and key from different secrets. 

In another operator we used this structure to consume parts from secrets/configmaps 

```go
// SecretKeySelector encodes a reference to a single key in a Secret in the same namespace.
type SecretKeySelector struct {
	// Key contains the name of the key inside the referenced Secret.
	//
	// +kubebuilder:validation:Required
	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Key Name",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
	Key string `json:"key"`

	// SecretName contains the name of the Secret containing the referenced value.
	//
	// +kubebuilder:validation:Required
	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Secret Name",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
	Name string `json:"name"`
}

// ConfigMapKeySelector encodes a reference to a single key in a ConfigMap in the same namespace.
type ConfigMapKeySelector struct {
	// Key contains the name of the key inside the referenced Secret.
	//
	// +kubebuilder:validation:Required
	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Key Name",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
	Key string `json:"key"`

	// SecretName contains the name of the Secret containing the referenced value.
	//
	// +kubebuilder:validation:Required
	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Secret Name",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
	Name string `json:"name"`
}

```

iypetrov: I like this idea. IMO it is better to keep our implementation as flexible as possible.

apis/v1beta1/opentelemetrycollector_types.go
iypetrov: Okay, I wasn't aware of that. Will extend the implementation.

swiatekm:
Looks good to me, can you update the documentation here: https://github.com/open-telemetry/opentelemetry-operator/tree/main/docs/target-allocator#service--pod-monitor-endpoint-credentials?

apis/v1beta1/opentelemetrycollector_types.go
```diff
@@ -268,6 +268,86 @@ type TargetAllocatorMTLS struct {
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name"`
+	// Key within the ConfigMap's data. Defaults to ca.crt when omitted.
+	// +optional
+	Key string `json:"key,omitempty"`
+}
+
+// CAReference points to a CA certificate stored in either a Secret or a ConfigMap. CA certificates
+// are public, so distributing them via a ConfigMap is common. Exactly one of secret or configMap
```
pavolloffay, line 325: are we doing any check for the "exactly one must be provided" ?

apis/v1beta1/opentelemetrycollector_types.go
```diff
@@ -268,6 +268,86 @@ type TargetAllocatorMTLS struct {
...
+	Secret *SecretKeySelector `json:"secret,omitempty"`
+	// ConfigMap sources the CA certificate from a ConfigMap.
+	// +optional
+	ConfigMap *ConfigMapKeySelector `json:"configMap,omitempty"`
+}
+
+// CertificateReference points to a certificate and its private key. The certificate and the key may
+// be stored in different Secrets.
+type CertificateReference struct {
+	// CertificateSecret selects the certificate. Its key defaults to tls.crt.
```
pavolloffay, line 339: Did you consider using kubebuilder markers for defaults? 

cc) @swiatekm

pavolloffay (commented):
LGTM, just some small comments

apis/v1beta1/opentelemetrycollector_types.go
iypetrov: I do, at `internal/manifests/manifestutils/mtls.go` I have a helper function:
```go
func validateCAReference(ref *v1beta1.CAReference) error {
	if (ref.Secret != nil) == (ref.ConfigMap != nil) { // neither or both
		return errors.New("tls.certificateAuthorityCertificate must set exactly one of secret or configMap")
	}
	return nil
}
```

apis/v1beta1/opentelemetrycollector_types.go
iypetrov: I did, but because we already define some constants [here](https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/constants/env.go#L37-L40) I decided to reuse them for consistency.

swiatekm: I think we can use both. Or, if it's genuinely a contextless constant, only kubebuilder is superior, as the CRD is self-documenting.

iypetrov: I added it.

swiatekm (approved):
LGTM, thank you for the contribution, and your patience with the review process!

iypetrov:
> LGTM, thank you for the contribution, and your patience with the review process!

Thanks for your reviews and feedback @swiatekm @pavolloffay!

swiatekm:
The link check is failing for some infrastructural reason, but it's not a merge blocker, FYI.

iypetrov:
@pavolloffay when you have time can you give it a look?

pavolloffay (approved):

PR #5364, opened 2026-07-27
fix(target-allocator): use TargetAllocator CR's own name for mTLS endpoint hostname

### LINKED ISSUE #4297: Enable HTTPs on a collector using TargetAllocator CRD when mTLS is enabled
### Component(s)

_No response_

## Description

When a collector used `.spec.targetAllocator` to create a target allocator, it used mTLS between a collector and a target allocator.
But when a collector used `opentelemetry.io/target-allocator` label to use a target allocator, a generated config map didn't use mTLS and plaintext http.

## Steps to Reproduce

1. Enable a feature gate operator.targetallocator.mtls on an operator
1. Create a TargetAllocator manifest
1. Create a OpenTelemetry collector with a label `opentelemetry.io/target-allocator` to use the target allocator in the previous step

## Expected Result

The collector uses https for requests to a target allocator

## Actual Result

The collector uses http for requests to a target allocator.

### Kubernetes Version

v1.32.5

### Operator version

v0.131.0

### Collector version

v0.132.0

## Environment
OS: (e.g., "Ubuntu 20.04")

### Log output

```shell
This is an example of a config map generated

> kubectl get cm test-observability-<REDACTED> -o yaml
...
    receivers:
        prometheus:
            api_server:
                enabled: true
                server_config:
                    endpoint: 0.0.0.0:9090
            config: {}
            target_allocator:
                collector_id: ${POD_NAME}
                endpoint: http://ta-targetallocator:80
                interval: 30s
    service:
...
```

### Additional context

These are manifests I got this bug

```
kind: TargetAllocator
metadata:
  name: ta
spec:
  serviceAccount: test-<redacted>-ta
  prometheusCR:
    enabled: true
    allowNamespaces:
      - test-<redacted>
    scrapeInterval: 10s
    scrapeConfigSelector:
      matchLabels:
        test-targetallocator: "true"
---
kind: OpenTelemetryCollector
metadata:
  name: test-<redacted>
  labels:
    opentelemetry.io/target-allocator: ta
...
  config:
    receivers:
      prometheus:
        api_server:
          enabled: true
          server_config:
            endpoint: "0.0.0.0:9090"
        config:
          scrape_configs:
            - job_name: "otel-collector"
              scrape_interval: 10s
              static_configs:
                - targets: ["0.0.0.0:8888"]
```

### Tip

pujitha24:
Motivation:
Issue #4297 reports that a collector associated with a standalone
TargetAllocator CR via the `opentelemetry.io/target-allocator` label kept
talking to it over plaintext HTTP even when the TargetAllocator had mTLS
enabled. Investigation found that the originally-reported scheme-selection
bug (the mTLS check being based on `otelCol.Spec.TargetAllocator.Enabled`
rather than the resolved TargetAllocator object) was already fixed on main
as a side effect of an unrelated PR (#5136) that introduced
`manifestutils.IsTAMTLSEnabled(ta *v1alpha1.TargetAllocator)`.

However, a second, still-live bug remained in the same code path: when
building the mTLS-enabled https endpoint, `internal/manifests/collector/
configmap.go` used `naming.TAService(otelCol.Name)` -- the collector's own
name -- instead of the associated TargetAllocator CR's name. The
plaintext endpoint built a few lines earlier (in config_replace.go) and
the real TargetAllocator Service (targetallocator/service.go) are both
named after the TargetAllocator CR's own name, so this mismatch caused the
https endpoint to point at the wrong hostname (and thus the wrong
Service/certificate) whenever the collector's name differs from the
TargetAllocator's name -- which is the normal case for the label-based
association described in the issue. This is a real, user-visible bug: the
collector would fail to reach the target allocator over https and fall
back to erroring out its scrape config, not merely a masked/log-only
issue.

Approach:
Change the one line in `internal/manifests/collector/configmap.go` to use
`naming.TAService(params.TargetAllocator.Name)` instead of
`naming.TAService(otelCol.Name)`. `params.TargetAllocator` is guaranteed
non-nil at that call site because it is guarded by
`manifestutils.IsTAMTLSEnabled(params.TargetAllocator)`, which itself
nil-checks.

Updated an existing unit test that had (accidentally) been relying on the
buggy behavior via an unnamed TargetAllocator object, and added two new
regression tests that reproduce the issue's exact scenario -- a
label-based, differently-named standalone TargetAllocator -- one at the
`collector.ConfigMap` level and one exercising the full controller
`GetParams` (including the label-based lookup) end to end. Both fail
against the pre-fix code and pass after the fix.

Validation:
- go build ./...
- go test ./internal/manifests/collector/... ./internal/controllers/... ./internal/manifests/manifestutils/... (all pass)
- ./bin/golangci-lint run ./internal/controllers/... ./internal/manifests/collector/... (0 issues)
- make chlog-validate (pass)

Fixes #4297

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>


internal/controllers/opentelemetrycollector_controller_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
+	"github.com/open-telemetry/opentelemetry-operator/apis/v1beta1"
+	"github.com/open-telemetry/opentelemetry-operator/internal/config"
+	"github.com/open-telemetry/opentelemetry-operator/internal/manifests/collector"
+	"github.com/open-telemetry/opentelemetry-operator/internal/manifests/manifestutils"
+	"github.com/open-telemetry/opentelemetry-operator/pkg/constants"
+)
+
+// TestGetParams_TargetAllocatorMTLSViaLabel guards against a regression of
```
swiatekm: I don't think we need this test, it's basically a restatement of the configmap one. Generally speaking, we should run all target allocator e2e tests in both modes, but that's far outside the scope of your PR. Just delete this file.

.chloggen/fix-ta-mtls-endpoint-hostname.yaml
```diff
@@ -0,0 +1,12 @@
+change_type: bug_fix
+
+component: target allocator
+
+note: Fix the collector's mTLS-enabled target allocator endpoint using the collector's own name instead of the associated TargetAllocator CR's name
```
swiatekm:
```suggestion
note: Fix collector mtls with a TargetAllocator CR
```

We don't need to go into details here.

swiatekm (commented):
Looks good to me overall, had some relatively minor comments.

pujitha24:
Thanks for the review! I addressed both points:

- Deleted `internal/controllers/opentelemetrycollector_controller_test.go` as suggested — the same scenario is already covered by the regression test in `collector.ConfigMap`'s test suite, so it was redundant.
- Reworded the `.chloggen` note to `Fix collector mtls with a TargetAllocator CR`, matching your suggested edit (this also resolves the Copilot bot's comment that the original wording had the fix backwards).

Ran `go build ./...`, the affected unit tests, `golangci-lint`, and `chloggen validate` — all pass.

swiatekm (approved):

swiatekm:

PR #5387, opened 2026-07-28
Add support to set session affinity for Collector Services

### LINKED ISSUE #4455: Support specifying `sessionAffinity` and `sessionAffinityConfig` for ClusterIP services created for Collectors by Operator
### Component(s)

collector

In some cases, we would like to create a Collector service with `sessionAffinity` set to `ClientIP`. But there is no way to specify that parameter in the OpenTelemetry Collector resource. Because the service is managed by the operator and may not be created immediately after the Collector resource is created, there is no way to patch the service during deployment (using Kustomize or similar tools). Also, service might be reconciled by the Operator, and reconciliation might drop the modified attribute.

Add the ability to specify `sessionAffinity` and `sessionAffinityConfig` in the OpentelemetryCollector resource and support these parameters during the provision of the resource by the operator.

A workaround would be to have a mutating admission webhook to add such parameters to the resource.

An alternative solution is to allow disabling service creation to let the user to provide service, but this will be very error-prone, because slight changes to the Collector definition might cause configuration drift between resources. 

### Additional context

_No response_

### Tip

yuriolisa:
**Description:** Add support to set `sessionAffinity` and `sessionAffinityConfig` on the Services the operator creates for the Collector. Follows the same pattern as
 `trafficDistribution`. v1beta1 only — v1alpha1 round-trip drops the field, same as `ipFamilies`.

   - Resolves: #4455

   **Testing:** Unit tests covering set/unset on the generated Service.

   **Documentation:** API docs regenerated via `make api-docs`.

### DIFF
--- a/.chloggen/session-affinity.yaml
+++ b/.chloggen/session-affinity.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: collector
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add support for setting `sessionAffinity` and `sessionAffinityConfig` on the Services created for the Collector
+
+# One or more tracking issues related to the change
+issues: [4455]
+
+# (Optional) One or more lines of additional information to render under the primary note.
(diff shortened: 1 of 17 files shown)


swiatekm:
Can you add this to the smoke tests in https://github.com/open-telemetry/opentelemetry-operator/tree/main/tests/e2e/smoke-collector too? Just to make sure it works in a real cluster. I'd also accept doing this in envtest, but I think chainsaw is simpler.

swiatekm (approved):

frzifus (approved):

PR #5388, opened 2026-07-28
fix(target-allocator): refresh stale meta labels when target address is unchanged

### LINKED ISSUE #4839: The target-allocator does not update __meta_kubernetes_pod_name for daemonsets running over the host network
### Component(s)

target allocator

## Description
When restarting pods, daemonsets running over the host network should update their __meta_kubernetes_pod_name . In fact, the __meta_kubernetes_pod_name remains old. Update __meta_kubernetes_pod_name only when restarting target-allocator

## Steps to Reproduce
  * Deploy the daemonset with hostNetwork: true
  * Update the daemonset. The pod name must change
  * The target Allocator in endpoint: GET jobs/{job_name}/targets to return the __meta_kubernetes_pod_name label with the old pod name

## Expected Result
__meta_kubernetes_pod_name  must have a new pod name

## Actual Result
__meta_kubernetes_pod_name  has an old name. Only restarting the target Allocator helps

### Kubernetes Version

v1.29.7

### Operator version

v0.145.0

### Collector version

v0.146.1

### Log output

```shell

```

### Additional context

_No response_

### Tip

pujitha24:
Fixes #4839

Motivation:
For a hostNetwork DaemonSet pod, __address__ is the node's IP, which
stays the same across pod restarts even though the pod itself (and
its __meta_kubernetes_pod_name) changes. Target-allocator's target
hash (target.HashFromBuilder) deliberately excludes __meta_*-prefixed
labels, since Prometheus discards meta labels after relabeling and two
targets differing only in meta labels are the same scrape target. The
allocator's target map (allocator.targetItems) is keyed by that same
hash, so diff.Maps(a.targetItems, targetMap) can never observe a meta
label change for a hash it already has an entry for: the "found but
changed" branch compares Hash() to Hash(), which are equal by
construction since the map key IS the hash. The result is that a
rediscovered target's stale *target.Item (with the old pod name) is
kept forever, and the /jobs/{job}/targets API and downstream
collectors keep seeing the pod name from before the restart, until
target-allocator itself restarts.

This is a data-correctness issue only: it does not cause a crash,
outage, or scrape failure. Labels are still correct without
Target Allocator (plain Prometheus service discovery), so this is
specific to sharding.

Approach:
Add allocator.refreshExistingTargetLabels, called at the end of
SetTargets under the existing lock. For every hash present in both the
new discovery result and the current target map, if the labels differ,
swap the stored *target.Item for the freshly discovered one, copying
over the existing item's CollectorName so the collector assignment
(which is unaffected, since hash and job name are unchanged) is
preserved.

Validation:
go test ./cmd/otel-allocator/internal/allocation/... -race
(all pass, including added regression test
TestSetTargetsUpdatesMetaLabelsForUnchangedHash, verified to fail
without the fix and pass with it, across all three allocation
strategies)

go build ./... && go vet ./... && go test ./...
(run from cmd/otel-allocator; all pass)

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/cmd/otel-allocator/internal/allocation/allocator.go
+++ b/cmd/otel-allocator/internal/allocation/allocator.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"github.com/prometheus/prometheus/model/labels"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
@@ -126,6 +127,24 @@ func (a *allocator) SetTargets(targets []*target.Item) {
 	// If there are any additions or removals
 	if len(targetsDiff.Additions()) != 0 || len(targetsDiff.Removals()) != 0 {
 		a.handleTargets(targetsDiff)
+	}
+	a.refreshExistingTargetLabels(targetMap)
+}
+
+// refreshExistingTargetLabels updates the stored labels for targets that already exist (same hash)
+// but were rediscovered with different meta labels, e.g. __meta_kubernetes_pod_name after a pod
+// restart of a hostNetwork DaemonSet, where __address__ (and therefore the target's hash, which
+// excludes meta labels, see target.HashFromBuilder) stays the same across restarts. Such changes
+// never surface as additions/removals in targetsDiff, so they must be applied separately. The
+// existing collector assignment is preserved since the target's identity (hash) is unchanged.
+func (a *allocator) refreshExistingTargetLabels(targetMap map[target.ItemHash]*target.Item) {
+	for hash, newItem := range targetMap {
+		existing, ok := a.targetItems[hash]
+		if !ok || labels.Equal(existing.Labels, newItem.Labels) {
+			continue
+		}
+		newItem.CollectorName = existing.CollectorName
+		a.targetItems[hash] = newItem
 	}
 }
 
--- a/cmd/otel-allocator/internal/allocation/allocator.go
+++ b/cmd/otel-allocator/internal/allocation/allocator.go
@@ -13,7 +13,6 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
-	"github.com/prometheus/prometheus/model/labels"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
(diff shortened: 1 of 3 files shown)


swiatekm (changes requested):
I don't think this is the right place to fix this problem. It may be that we simply need to include the meta labels in the hash. A target with different labels is logically a different target, even if the labels are just meta.

It would help a lot if we could add this case to the target allocator conformance test and explicitly show the failure.

cmd/otel-allocator/internal/allocation/allocator.go
```diff
@@ -127,6 +128,24 @@ func (a *allocator) SetTargets(targets []*target.Item) {
...
+// refreshExistingTargetLabels updates the stored labels for targets that already exist (same hash)
+// but were rediscovered with different meta labels, e.g. __meta_kubernetes_pod_name after a pod
+// restart of a hostNetwork DaemonSet, where __address__ (and therefore the target's hash, which
+// excludes meta labels, see target.HashFromBuilder) stays the same across restarts. Such changes
+// never surface as additions/removals in targetsDiff, so they must be applied separately. The
+// existing collector assignment is preserved since the target's identity (hash) is unchanged.
+func (a *allocator) refreshExistingTargetLabels(targetMap map[target.ItemHash]*target.Item) {
+	for hash, newItem := range targetMap {
+		existing, ok := a.targetItems[hash]
+		if !ok || labels.Equal(existing.Labels, newItem.Labels) {
```
swiatekm: I don't think the label comparison is necessary here. Labels can be quite large, and we already have the new target allocated. Always using the new one is cheaper than testing.

cmd/otel-allocator/internal/allocation/allocator.go
```diff
@@ -127,6 +128,24 @@ func (a *allocator) SetTargets(targets []*target.Item) {
...
 	}
+	a.refreshExistingTargetLabels(targetMap)
+}
+
+// refreshExistingTargetLabels updates the stored labels for targets that already exist (same hash)
+// but were rediscovered with different meta labels, e.g. __meta_kubernetes_pod_name after a pod
+// restart of a hostNetwork DaemonSet, where __address__ (and therefore the target's hash, which
+// excludes meta labels, see target.HashFromBuilder) stays the same across restarts. Such changes
+// never surface as additions/removals in targetsDiff, so they must be applied separately. The
+// existing collector assignment is preserved since the target's identity (hash) is unchanged.
```
swiatekm:
```suggestion
// refreshExistingTargetLabels replaces the stored Item of every target that is present
// both before and after this update, so that what we serve reflects the latest
// discovery rather than the first sighting.
//
// A target's identity is the hash of its post-relabel labels, which excludes meta
// labels (see target.HashFromBuilder), mirroring Prometheus: meta labels are
// discarded when the final label set is computed, so two targets differing only in
// them are the same scrape target. That means a target can keep its identity while
// the meta labels discovery reports for it change — a DaemonSet pod with
// hostNetwork replaced on the same host address, for instance, keeps its address
// and its relabeled labels while its pod name and UID change. The diff below sees
// no change for such a target, but Item.Labels is not internal bookkeeping: we
// serve it verbatim on the HTTP SD endpoint, and the collector turns those meta
// labels into resource attributes. Keeping the old Item would attribute the new
// pod's metrics to the departed one, forever, since the hash never moves again.
//
// Prometheus does the same thing for the same reason, overwriting a retained
// target's discovered labels on every sync:
// https://github.com/prometheus/prometheus/blob/v3.12.0/scrape/scrape.go#L491
//
// The collector assignment lives on the Item, so it has to carry over. Dropping it
// would make the least-weighted strategy treat every surviving target as newly
// assigned and reshuffle the whole set on each update, and would leave the
// per-collector target counts unbalanced when the target is eventually removed.
//
```

I'd like the reasoning for this logic existing to be very clear, as this has historically been a source of many subtle bugs.

swiatekm (commented):
Having thought through the problem, this is probably the best way to solve it after all. The cleaner solutions all break something else. This way, our logic is identical to Prometheus', which is what we actually want.

I'm going to add an end-to-end conformance test for this case after this PR is merged, as it's a lot of annoying wiring that I don't want to burden this PR with. Thanks for taking the time to troubleshoot and articulate the problem in a way that I could finally reproduce.

pujitha24:
Thanks for the review! Addressed both inline comments:

- Dropped the `labels.Equal` comparison in `refreshExistingTargetLabels` — we now always overwrite the stored `Item` with the newly-discovered one (still carrying over the existing `CollectorName`), since the new item is already on hand and comparing labels first was pure overhead.
- Replaced the doc comment with your suggested wording explaining the identity/hash rationale, the hostNetwork DaemonSet scenario, why keeping the stale `Item` would misattribute metrics, and why the collector assignment must carry over.

Also removed the now-unused `prometheus/model/labels` import. Ran `go build`/`go vet`/`go test` for the package, including the existing `TestSetTargetsUpdatesMetaLabelsForUnchangedHash` regression test — all pass. I didn't touch anything related to the broader "is this the right fix" discussion since it looks like that's already been resolved in your favor, and I'll leave the conformance-test follow-up to you as you mentioned.

pujitha24:
Thanks for the review! I've addressed both inline comments in bd13d33b:

- Dropped the `labels.Equal` comparison in `refreshExistingTargetLabels` and always overwrite the stored `Item` with the newly discovered one (only carrying over `CollectorName`), since we already have the new target and the comparison was pure overhead.
- Replaced the doc comment with the fuller explanation you suggested, covering target identity/hashing, why meta labels can change without the hash moving, the Prometheus parity/link, and why the collector assignment has to carry over.

`go test ./cmd/otel-allocator/internal/allocation/...` still passes, including `TestSetTargetsUpdatesMetaLabelsForUnchangedHash`, which covers this exact case. I'll leave the e2e conformance test to you as you mentioned.

swiatekm (approved):

PR #5390, opened 2026-07-30
fix(collector): preserve string values written in exponential notation when generating collector config

### LINKED ISSUE #4314: Operator misinterprets strings as numbers when converting configuration
### Component(s)

collector

## Description
The opentelemetery operator appears to fail to preserve types of certain fields when writing them out to the configmap of a collector

## Steps to Reproduce
Deploy a collector with the following opentelemetrycollector definition:
```
kind: OpenTelemetryCollector
metadata:
  name: otelcol2
  namespace: opentelemetry
spec:
  config:
    exporters:
      otlphttp/prometheus:
        endpoint: http://127.0.0.1/api/v1/otlp
    processors:
      metricstransform/cluster-code:
        transforms:
        - action: update
          include: ^.*$
          match_type: regexp
          operations:
          - action: add_label
            new_label: cluster_id
            new_value: "0e12"
    receivers:
      otlp:
        protocols:
          grpc:
            endpoint: 0.0.0.0:4317
          http:
            endpoint: 0.0.0.0:4318
    service:
      pipelines:
        metrics:
          exporters:
          - otlphttp/prometheus
          processors:
          - metricstransform/cluster-code
          receivers:
          - otlp
          - prometheus
  mode: statefulset
  replicas: 1
```

## Expected Result
The "cluster_id" property would be added to all metrics with value "0e12" without being interpreted as an integer (in our use case its a unique identifier for the cluster that is a randomly generated set of hex digits defined at cluster creation)

## Actual Result
Created collector errors with the message `'transforms[0].operations[0].new_value' expected type 'string', got unconvertible type 'float64'`, which indicates that it is being misinterpreted as a string. We have checked, and in the yaml for the collector definition it is correctly quoted on the cluster, but in the configmap created it has lost the quoting

### Kubernetes Version

1.32.6

### Operator version

0.129.1

### Collector version

0.131.0

## Environment
GKE, nodes running on the latest Container-Optimized OS version for that k8s version

Also tested on minikube v1.36.0 with k8s version 1.32.4

### Log output

```shell
Error: failed to get config: cannot unmarshal the configuration: decoding failed due to the following error(s):

'processors' error reading configuration for "metricstransform/cluster-code": decoding failed due to the following error(s):

'transforms[0[].operations[0].new_value' expected type 'string', got unconvertible type 'float64'
2025/08/27 12:40:09 collector server run finished with error: failed to get config: cannot unmarshal the configuration: decoding failed due to the following error(s):

'processors' error reading configuration for "metricstransform/cluster-code": decoding failed due to the following error(s):

'transforms[0[].operations[0].new_value' expected type 'string', got unconvertible type 'float64'
stream closed EOF for opentelemetry/otelcol2-collector-0 (otc-container)
```

### Additional context

_No response_

### Tip

pujitha24:
Strings that happen to look like exponential numbers can currently change type when the Operator generates the Collector YAML.

For example, a value such as `"0e12"` is meant to stay a string, but it can be emitted without quotes. The Collector then parses it as a number and fails when the component expects a string.

I initially looked at removing `AutoInt()`, but that is too broad and changes legitimate numeric serialization. This fix keeps the existing behavior and only forces quoting for the exponent-style strings that would otherwise be ambiguous.

I added a regression test that round-trips this case and verifies the value is still a string. The relevant test suites, build, lint and changelog validation pass.

AI assistance: I used AI tools while investigating this and preparing the change; I have reviewed and tested it myself.

Fixes #4314

### DIFF
--- a/.chloggen/fix-yaml-string-quoting-4314.yaml
+++ b/.chloggen/fix-yaml-string-quoting-4314.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: collector
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Fix collector config string values written in exponential notation without a decimal point (e.g. `\"0e12\"`) being emitted unquoted in the generated ConfigMap, causing the collector to misread them as numbers and fail to start."
+
+# One or more tracking issues related to the change
+issues: [4314]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/apis/v1beta1/config.go
+++ b/apis/v1beta1/config.go
@@ -7,8 +7,11 @@ import (
 	"bytes"
 	"encoding/json"
 	"maps"
+	"regexp"
+	"strconv"
 
 	go_yaml "github.com/goccy/go-yaml"
+	"github.com/goccy/go-yaml/token"
 )
 
 type ComponentKind int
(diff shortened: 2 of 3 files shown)


pujitha24:
The failing `e2e-ta-collector-mtls` / `e2e-targetallocator-cr` checks were caused by my own change, not a flake: registering `go_yaml.CustomMarshaler[string]` to intercept all strings also broke `IndentSequence(true)`'s handling of nested sequences (e.g. Prometheus `static_configs`/`targets` ended up over-indented), which is exactly what those e2e tests caught.

I reworked the fix to scope the quoting to only the ambiguous exponent-notation values, via a dedicated `quotedString` wrapper type applied on a copy of the config tree, instead of overriding marshaling for every string. I reproduced both the original bug and the indentation regression locally, confirmed the new approach fixes both, and reran the affected unit test suites (`internal/otelconfig`, `internal/config`, `internal/manifests`, `internal/webhook`, `pkg/sidecar`) plus `make lint` — all green. I can't run the e2e suite locally, so I'm relying on CI to confirm the target-allocator tests pass now.

pujitha24:
This has been green and rebased for a couple of weeks now — happy to make any changes if something would help move review along.

swiatekm:
> This has been green and rebased for a couple of weeks now — happy to make any changes if something would help move review along.

Hey, sorry for the late response. Having read the PR description, we're essentially working around a bug in `goccy/go-yaml`, right? The main thing I'm concerned about is that what we do here won't solve other similar edge cases, or the regex we use turns out to not be consistent with the spec either.  How did you come up with it?

I'd like to have a more structural solution to this whole class of problems, but I'm not sure yet what that solution looks like.

pujitha24:
Yeah — this is working around a real gap in goccy/go-yaml's quoting check, not the YAML core schema in the abstract. I traced it into their `token.toNumber` (v1.19.2): it only calls a string numeric if it parses as int/octal/hex/binary, or has exactly one dot for the float branch, so `"0e12"` fails the octal parse (`e` isn't a valid octal digit) and has zero dots, and `IsNeedQuoted` never flags it. On the decode side, `yaml.v3`'s resolver falls back to Go's `strconv.ParseFloat` when the int parse fails, and `ParseFloat` does accept exponent notation without a dot, so it comes back as `0.0` — the actual mismatch is two Go libraries disagreeing on what looks numeric. I got the regex by reading goccy's source for exactly what it misses, not from a spec audit, so the skepticism is fair — I only confirmed it closes this one gap, not the whole class.

A more structural version might be to quote any string where `strconv.ParseFloat`/`ParseInt` would succeed, since that's the actual criterion Go-based consumers like the collector use, instead of approximating spec patterns — but I haven't built that and it's a bigger change than this PR. Happy to take a stab at it here if you'd rather have that than the narrow fix, or leave this as-is while you think through the structural approach.

swiatekm:
@pujitha24 This problem was resolved in a structural (and much more complex) way in https://github.com/open-telemetry/opentelemetry-operator/pull/5615, so I think we can close this PR. Thank you for your contribution nonetheless - the solution wasn't wrong, we just weren't comfortable only fixing this one case.

Issue #5346, opened 2026-07-21
Instrumentation v1beta1


pavolloffay:
### Is your feature request related to a problem? Please describe.

This is an umbrealla issue for Instrumentation v1beta1 work.

RFC:  https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/rfcs/instrumentation-v1beta1.md

 

Implementation plan:

- [x] Go struct with cleaned v1alpha1 spec https://github.com/open-telemetry/opentelemetry-operator/pull/5345 
- [ ] Go struct for declarative config  
- [ ] conversion code https://github.com/open-telemetry/opentelemetry-operator/pull/5392 
- [ ] Bring back env config to root level, it is needed for env substitution in the declarative config https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/rfcs/instrumentation-v1beta1.md#environment-variable-substitution-in-declarative-config 
- [ ] add v1beta CRD into bundle and set it as a storage version

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

Samyra312007:
Respected Maintainers, If this issue is still opens i would like to give it a try.

PR #5392, opened 2026-07-30
Add internal conversion functions for instrumentation v1beta1

pavolloffay:

- Updates https://github.com/open-telemetry/opentelemetry-operator/issues/5346 


atoulme (approved):
LGTM

frzifus (approved):

iblancasa (approved):

swiatekm:
What happens if there's overlap between env variables for different SDKs? I thought the ability to set env variables for them individually would be preserved, even if we don't encourage that as a config mechanism.

pavolloffay:
The per-language env vars are preserved for lossless round-trip through an annotation `instrumentation.opentelemetry.io/v1alpha1-fields`). During `v1alpha1` to `v1beta1` conversion, per-language env vars are merged into envConfig.env so they're visible/accessible, and the original per-language values are serialized into the annotation. On the way back (v1beta1→v1alpha1), the annotation is read and the per-language env vars are restored to their original fields, and then subtracted from Spec.Env to recover the original common-only list.

I will look at the env vars more closely in a follow up PR (add as TODO in https://github.com/open-telemetry/opentelemetry-operator/issues/5346) We will need to have env vars in the root of the spec even for the decparative config to set e.g. TOKEN for the API endpoint https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/rfcs/instrumentation-v1beta1.md#environment-variable-substitution-in-declarative-config

swiatekm:
> The per-language env vars are preserved for lossless round-trip through an annotation `instrumentation.opentelemetry.io/v1alpha1-fields`). During `v1alpha1` to `v1beta1` conversion, per-language env vars are merged into envConfig.env so they're visible/accessible, and the original per-language values are serialized into the annotation. On the way back (v1beta1→v1alpha1), the annotation is read and the per-language env vars are restored to their original fields, and then subtracted from Spec.Env to recover the original common-only list.
> 
> I will look at the env vars more closely in a follow up PR (add as TODO in #5346) We will need to have env vars in the root of the spec even for the decparative config to set e.g. TOKEN for the API endpoint https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/rfcs/instrumentation-v1beta1.md#environment-variable-substitution-in-declarative-config

It's a lossless roundtrip, but not a lossless conversion, because we throw all the per-language env vars into a single bag. So our v1beta1 Instrumentation will have different behavior than the v1alpha1 we converted it from, in a non-obvious way. Is that correct?

pavolloffay:
I don't think so, the v1beta1 (which would be created from v1alpha1) would have all the infromation at the injection time, which env vars to inject to guarantee the same behavior.

swiatekm:
> I don't think so, the v1beta1 (which would be created from v1alpha1) would have all the infromation at the injection time, which env vars to inject to guarantee the same behavior.

Let's say two language sections declare `OTEL_SERVICE_NAME` with different values. My understanding of your code is that whichever got merged later would win, making v1beta1 behave differently (and incorrectly) compared to v1alpha1.

swiatekm (approved):

PR #5393, opened 2026-07-30
Created preservedLabels and Annotations

### LINKED ISSUE #2795: spec.podAnnotations entries can not be deleted in reconcile
### Component(s)

collector

## Description

Hi, I tested opentelemetry operator with testing tool [acto](https://github.com/xlab-uiuc/acto), which automatically generates otelcol instance definition yaml files and performs end-to-end testing. Configurations of acto to test opentelemetry operator are [here](https://github.com/xlab-uiuc/acto/pull/358).

I found that `spec.podAnnotations` entries can not be deleted when the user applies new definitions to reconcile.

## Steps to Reproduce

1. create a kind cluster
2. install cert-manager and opentelemetry-operator with `kubectl apply -f`
3. create and change the definition of opentelemetry collector one by one as follows (mutation 1 and 2 are the most important):

```yaml
# mutation 0
kind: OpenTelemetryCollector
metadata:
  name: test-cluster
spec:
  config: "receivers:\n  otlp:\n    protocols:\n      grpc:\n      http:\nprocessors:\n\
    \  memory_limiter:\n    check_interval: 1s\n    limit_percentage: 75\n    spike_limit_percentage:\
    \ 15\n  batch:\n    send_batch_size: 10000\n    timeout: 10s\n\nexporters:\n \
    \ debug:\n\nservice:\n  pipelines:\n    traces:\n      receivers: [otlp]\n   \
    \   processors: []\n      exporters: [debug]"
```

```yaml
# mutation 1
kind: OpenTelemetryCollector
metadata:
  name: test-cluster
spec:
  config: "receivers:\n  otlp:\n    protocols:\n      grpc:\n      http:\nprocessors:\n\
    \  memory_limiter:\n    check_interval: 1s\n    limit_percentage: 75\n    spike_limit_percentage:\
    \ 15\n  batch:\n    send_batch_size: 10000\n    timeout: 10s\n\nexporters:\n \
    \ debug:\n\nservice:\n  pipelines:\n    traces:\n      receivers: [otlp]\n   \
    \   processors: []\n      exporters: [debug]"
  podAnnotations:
    actokey: actokey
```

```yaml
# mutation 2
kind: OpenTelemetryCollector
metadata:
  name: test-cluster
spec:
  config: "receivers:\n  otlp:\n    protocols:\n      grpc:\n      http:\nprocessors:\n\
    \  memory_limiter:\n    check_interval: 1s\n    limit_percentage: 75\n    spike_limit_percentage:\
    \ 15\n  batch:\n    send_batch_size: 10000\n    timeout: 10s\n\nexporters:\n \
    \ debug:\n\nservice:\n  pipelines:\n    traces:\n      receivers: [otlp]\n   \
    \   processors: []\n      exporters: [debug]"
  podAnnotations: {}
```

```yaml
# mutation 3
kind: OpenTelemetryCollector
metadata:
  name: test-cluster
spec:
  config: "receivers:\n  otlp:\n    protocols:\n      grpc:\n      http:\nprocessors:\n\
    \  memory_limiter:\n    check_interval: 1s\n    limit_percentage: 75\n    spike_limit_percentage:\
    \ 15\n  batch:\n    send_batch_size: 10000\n    timeout: 10s\n\nexporters:\n \
    \ debug:\n\nservice:\n  pipelines:\n    traces:\n      receivers: [otlp]\n   \
    \   processors: []\n      exporters: [debug]"
  podAnnotations:
    actokey2: actokey2
```

## Expected Result

In the state `mutation 2` and `mutation 3`, the `spec.podAnnotations` entry `actokey: actokey` in user specification is been deleted, and the annotations of pod and deployment should be changed as what `spec.podAnnotations` defines.

## Actual Result

When I describe the pod and deployment `test-cluster-collector` after applying `mutation 2` and `mutation 3`, the `actokey: actokey` still exists. In `mutation 3` I can see both `actokey: actokey` and `actokey2: actokey2` in the annotations.

### Kubernetes Version

1.28.0

### Operator version

0.95.0

### Collector version

0.95.0

## Environment
OS: Ubuntu 22.04

### Log output

_No response_

### Additional context

When logging the variables related to the annotations in the operator (like desiredObjects in [this line](https://github.com/open-telemetry/opentelemetry-operator/blob/2ddfe3193a2dad57836ebbca5fb92481c35dbeff/controllers/common.go#L81)(https://github.com/open-telemetry/opentelemetry-operator/blob/2ddfe3193a2dad57836ebbca5fb92481c35dbeff/controllers/common.go#L81)), we can find that the data are changed as expected. Thus maybe the problem is related to the k8s controller-runtime API in [this line](https://github.com/open-telemetry/opentelemetry-operator/blob/2ddfe3193a2dad57836ebbca5fb92481c35dbeff/controllers/common.go#L101)(https://github.com/open-telemetry/opentelemetry-operator/blob/2ddfe3193a2dad57836ebbca5fb92481c35dbeff/controllers/common.go#L101).

yuriolisa:

Fixing a bug of spec.podAnnotations entries could never be deleted: once an annotation was set on the CR and reconciled into the managed workloads, removing it from the CR had no effect, because mutatePodTemplate merged the desired annotations/labels into the existing pod template (merge-only semantics) and there was no way to distinguish operator-managed entries from externally added ones.

- Resolves: #2795 

### DIFF
--- a/.chloggen/fix-pod-template-metadata-deletion.yaml
+++ b/.chloggen/fix-pod-template-metadata-deletion.yaml
@@ -1,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: collector
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Make pod template annotations and labels authoritative from the CR, so entries removed from `spec.podAnnotations` are removed from the managed workloads. Keys managed by external tooling (e.g. ArgoCD, `kubectl rollout restart`, policy engines) can be exempted via the new `--preserved-annotations` and `--preserved-labels` operator flags. Annotations and labels on the managed objects themselves (not their pod templates) are still preserved.
+
+# One or more tracking issues related to the change
+issues: [2795]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  This supersedes the `operator.opentelemetry.io/prometheus-annotations-added` ownership marker introduced in #5069, which is no longer necessary and has been removed. `prometheus.io/*` (or any other) annotations applied out of band directly on the managed workloads' pod templates are now removed during reconciliation unless they match `--preserved-annotations` (e.g. `--preserved-annotations=prometheus.io/*`).
--- a/internal/config/cli.go
+++ b/internal/config/cli.go
@@ -51,6 +51,8 @@ func CreateCLIParser(cfg Config) *pflag.FlagSet {
 	f.String("auto-instrumentation-nginx-image", cfg.AutoInstrumentationNginxImage, "The default OpenTelemetry Nginx instrumentation image. This image is used when no image is specified in the CustomResource.")
 	f.StringArray("labels-filter", cfg.LabelsFilter, "Labels to filter away from propagating onto deploys. It should be a string array containing patterns, which are literal strings optionally containing a * wildcard character. Example: --labels-filter=.*filter.out will filter out labels that looks like: label.filter.out: true")
 	f.StringArray("annotations-filter", cfg.AnnotationsFilter, "Annotations to filter away from propagating onto deploys. It should be a string array containing patterns, which are literal strings optionally containing a * wildcard character. Example: --annotations-filter=.*filter.out will filter out annotations that looks like: annotation.filter.out: true")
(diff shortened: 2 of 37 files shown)


swiatekm:
This should be marked as a breaking change at the very least.

swiatekm:
I'm marking this for discussion during the next SIG meeting, as I think we need to carefully consider just how much this breaks users and what we can do to make it less painful. @yuriolisa do you know how other operators solve this problem?

jaronoff97:
I do not think we can make this change safely as it stands today. the problem is that if we make this the default behavior we potentially remove labels that other operators add to deployments or other manifests in our reconcile loop. This would cause us to remove their labels/annotations only for them to be added once again by their operator. if its a label change this could cause an infinite loop of reconciliation to take place and potentially take down a cluster. If we really wanted to remove labels, it would be better to determine which ones may be safe to remove.

swiatekm:
> I do not think we can make this change safely as it stands today. the problem is that if we make this the default behavior we potentially remove labels that other operators add to deployments or other manifests in our reconcile loop. This would cause us to remove their labels/annotations only for them to be added once again by their operator. if its a label change this could cause an infinite loop of reconciliation to take place and potentially take down a cluster. If we really wanted to remove labels, it would be better to determine which ones may be safe to remove.

We currently add a special annotation tracking if it was us who added the prometheus scrape annotations. We could generalize this and have each reconciliation save all the labels and annotations it added in a special annotation. Then we know exactly what we're allowed to remove.

PR #5398, opened 2026-07-31
fix(collector): resolve version label from default collector image

### LINKED ISSUE #4175: Collector pod label `app.kubernetes.io/version` is wrong when using global `--collector-image` option

evillgenius75:
**Description:**
When `.spec.image` is omitted from an `OpenTelemetryCollector` custom resource, the operator falls back to the configured default collector image (`--collector-image` CLI flag / operator default config).

However, manifest generators previously passed the unnormalized `params.OtelCol.Spec.Image` directly to `manifestutils.Labels(...)`. Because `.spec.image` was empty, `manifestutils.Labels` fell back to `app.kubernetes.io/version: latest` instead of extracting the image version from the configured default collector image. Downstream components deriving `service.version` from this label were incorrectly assigned `latest`.

This PR fixes the issue by:
1. Normalizing `.spec.image` upfront in `OpenTelemetryCollectorReconciler.GetParams` and `collector.Build` so that all downstream manifest generators operate on the effective collector image. I originally started to add the fallback logic into every manifest builder but saw that not sustainable with changes and future updates to manifests. By normalizing `.spec.image` upfront it makes sure that all manifests builder (current and future) and container generators get the effective collector image without modifying the CR.
2. Introducing `collector.Labels` to resolve the effective image from CR spec or config fallback.
3. Adding comprehensive unit tests for version label resolution.

- Resolves: #4175

**Testing:**
- Added unit tests in `internal/manifests/collector/labels_test.go` covering:
  - Explicit `.spec.image` tags and SHA256 hashes
  - Fallback to `cfg.CollectorImage` when `.spec.image` is omitted
  - Fallback to `latest` when both are empty
  - Regex label filtering
- Ran full test suite across the repository (`go test ./...`), which passed with 0 errors.

**Documentation:**
- Added entry to `CHANGELOG.md` under `<!-- next version -->` bug fixes.

**Documentation & Testing assisted by AI**

### DIFF
--- a/.chloggen/fix-collector-version-label-4175.yaml
+++ b/.chloggen/fix-collector-version-label-4175.yaml
@@ -1,0 +1,11 @@
+change_type: bug_fix
+component: collector
+issues: [4175]
+
+note: "Fix `app.kubernetes.io/version` label fallback to resolve version from default collector image (`--collector-image`) when `.spec.image` is omitted."
+
+subtext: |
+  When `.spec.image` is omitted from an OpenTelemetryCollector custom resource,
+  the operator now normalizes `.spec.image` upfront during reconciliation to use
+  the configured default collector image, ensuring all manifest generators stamp
+  the correct image version label instead of defaulting to `latest`.
--- a/internal/controllers/opentelemetrycollector_controller.go
+++ b/internal/controllers/opentelemetrycollector_controller.go
@@ -162,6 +162,10 @@ func (r *OpenTelemetryCollectorReconciler) GetParams(ctx context.Context, instan
 		r.defaultFSGroupOnOpenShift(ctx, &instance)
 	}
 
+	if instance.Spec.Image == "" {
+		instance.Spec.Image = r.config.CollectorImage
+	}
+
 	p := manifests.Params{
 		Config:   r.config,
 		Client:   r.Client,
--- a/internal/manifests/collector/collector.go
+++ b/internal/manifests/collector/collector.go
@@ -20,6 +20,9 @@ const (
 
 // Build creates the manifest for the collector resource.
 func Build(params manifests.Params) ([]client.Object, error) {
+	if params.OtelCol.Spec.Image == "" {
+		params.OtelCol.Spec.Image = params.Config.CollectorImage
+	}
 	var resourceManifests []client.Object
 	var manifestFactories []manifests.K8sManifestFactory[manifests.Params]
 	switch params.OtelCol.Spec.Mode {
--- a/internal/manifests/collector/configmap.go
+++ b/internal/manifests/collector/configmap.go
@@ -39,7 +39,7 @@ func ConfigMap(params manifests.Params) (*corev1.ConfigMap, error) {
 	}
 	name := naming.ConfigMap(otelCol.Name, hash)
 	collectorName := naming.Collector(otelCol.Name)
-	labels := manifestutils.Labels(otelCol.ObjectMeta, collectorName, otelCol.Spec.Image, ComponentOpenTelemetryCollector, []string{})
+	labels := Labels(*otelCol, collectorName, params.Config, []string{})
 
 	annotations, err := manifestutils.Annotations(*otelCol, params.Config.AnnotationsFilter)
 	if err != nil {
--- a/internal/manifests/collector/daemonset.go
+++ b/internal/manifests/collector/daemonset.go
@@ -16,7 +16,7 @@ import (
 // DaemonSet builds the deployment for the given instance.
 func DaemonSet(params manifests.Params) (*appsv1.DaemonSet, error) {
 	name := naming.Collector(params.OtelCol.Name)
-	labels := manifestutils.Labels(params.OtelCol.ObjectMeta, name, params.OtelCol.Spec.Image, ComponentOpenTelemetryCollector, params.Config.LabelsFilter)
+	labels := Labels(params.OtelCol, name, params.Config, params.Config.LabelsFilter)
 
 	annotations, err := manifestutils.Annotations(params.OtelCol, params.Config.AnnotationsFilter)
 	if err != nil {
--- a/internal/manifests/collector/daemonset_test.go
+++ b/internal/manifests/collector/daemonset_test.go
@@ -63,7 +63,7 @@ func TestDaemonSetNewDefault(t *testing.T) {
 		"app.kubernetes.io/managed-by": "opentelemetry-operator",
 		"app.kubernetes.io/name":       "my-instance-collector",
 		"app.kubernetes.io/part-of":    "opentelemetry",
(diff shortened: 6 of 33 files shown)


evillgenius75:
### 💡 Context & Design Considerations
My first Contribution to OpenTelemetry! Yay!
Now down to business! 

When investigating this issue, I looked into a few architectural approaches to resolve label and metadata resolution:
1. **Option A (Per-builder fallback)**: Adding individual fallback logic into every manifest builder. This was technical the easiest and fastest but would have a larger blast radius in the future if we add new manifest builders, etc.
2. **Option B (Upfront In-Memory Normalization — implemented in this PR)**: Normalizing `.spec.image` upfront during `GetParams` and `collector.Build`. This ensures all manifest builders and container generators automatically receive the effective collector image without modifying the persisted custom resource in etcd.
3. **Option C (Centralized Metadata Decorator Pipeline — long-term architectural refactor)**: Moving label and annotation stamping out of individual manifest builders entirely and into a centralized post-processing decorator loop inside `collector.Build` (e.g. `ApplyCommonMetadata(obj, params)`).
**Why Option B for this PR:**
Option B delivers a clean, localized fix for #4175 with low blast radius and low risk of regression across other controllers.
**Future Architectural Discussion:**
Longer term, adopting a centralized metadata decorator pipeline (Option C) could help eliminate boilerplate and reduce toil whenever new upstream Kubernetes resource types are introduced to the collector manifest builder. Happy to open a follow-up discussion/issue on this if maintainers feel there is broader interest!

swiatekm:
@open-telemetry/operator-approvers can anyone remind me why we don't just set the image in the webhook defaulter, like we do for instrumentation? That sounds like a much more reasonable solution to this problem.

evillgenius75:
### 🔄 Update & CI E2E Assertion Fixes

1. **Rebased onto latest `main`**: Brought the branch up to date with upstream `main`.
2. **Updated E2E Test Assertions**: 
   - The Chainsaw E2E tests (`receiver-kubeletstats`, `create-sm-prometheus-exporters`, `extension`, etc.) previously asserted `app.kubernetes.io/version: latest` because they were authored under the old behavior where the operator always emitted `latest` when `.spec.image` was omitted.
   - Now that the operator correctly derives the version from the configured default image (e.g. `0.156.0`), those static `latest` assertions have been updated so that the full E2E test matrix verifies the correct operand metadata.
3. **Preserved `latest` Fallback**: If an image is untagged or explicitly specified as `:latest`, the operator still safely defaults to `latest`.

evillgenius75:
@swiatekm Thanks for bringing this up! 
cc @open-telemetry/operator-approvers 

I am not one of the approvers obviously, however, I looked into adopting the webhook defaulter approach (similar to `Instrumentation`), and here is what I deduced:

### Why `Instrumentation` uses Webhook Defaulting:
`Instrumentation` injects init-containers into external application pods rather than managing its own reconciler workload. It writes `spec.<lang>.image` into etcd alongside `operator.opentelemetry.io/default-auto-instrumentation-*` annotations, and relies on an explicit upgrade controller (`internal/instrumentation/upgrade`) to patch CRs in etcd across operator upgrades.

### Why `OpenTelemetryCollector` (I also think TargetAllocator / OpAMPBridge follow the same logic here) avoids Webhook Defaulting:
1. **Automatic Operator Upgrades**: Leaving `.spec.image` unset in etcd allows the collector to automatically upgrade its operand when the operator image is updated, without needing CR patch migrations in etcd. If the webhook stamped `spec.image` directly into the CR, I think the controller would treat it as user-pinned and never upgrade it upon operator updates.
2. **GitOps / ArgoCD Drift**: Mutating `.spec.image` in the admission webhook causes drift warnings against Git manifests where `image` is omitted.
3. **Reconciler Resilience**: If webhooks are disabled or bypassed, the reconciler and manifest builders still need to handle empty `.spec.image` safely.

### Conclusion:
Resolving `.spec.image` in-memory during reconciliation (as done in this PR) keeps `.spec.image` unpinned in etcd for easier upgrades and GitOps compatibility while still propagating the operand version tag to `app.kubernetes.io/version`.

However, if the maintainers would prefer `OpenTelemetryCollector` to align with `Instrumentation` by writing `spec.image` + defaulting annotations in the webhook and handling upgrades via `pkg/collector/upgrade`, I am happy to implement that pattern as well!

iblancasa:
@evillgenius75 thank you for your PR. I’d really appreciate a more personal back-and-forth in your own words, even if brief. It helps us better understand each other’s reasoning and makes the review smoother.

evillgenius75:
Sure @iblancasa 

As I described I looked through the Webhook defaulter as used  for Instrumentation, and from what I was able to gather it writes in the `spec.<lang>.image` into the etcd and uses a specific Upgrade controller to patch all of the CR on upgrade. 

I did not find a similar upgrade controller mechanism for OpenTelemteryCollector. I thought perhaps this may be an anomaly and looked at other CRD controllers and only instrumentation had this webhook process. 

I was perhaps being a bit too clever to try and remove the explicit calls in each manifest builder, as I figured that would be a bit too brittle when changes to new manifest builders need to happen. In my first iteration I did not even look at the way the Instrumentation CRD worked but happy to dive deeper.

swiatekm:
For the record, we do explicitly set defaults for Instrumentation CRs, and they also get automatic upgrades. I'm not sure what the reason for the discrepancy is. I've tagged this PR so we can discuss it during our next SIG meeting on 13.08.2026. Sorry for the wait @evillgenius75, but I want to make sure we're fixing the right thing here.

pavolloffay (commented):
We have discussed this on the SIG meeting. It makes sense to move the image defaulting to the webhook, which should result in a simpler patch.

evillgenius75:
Happy to shift and work on that to solve this bug. Thank you for the guidance.

PR #5410, opened 2026-08-03
Add conformance tests for K8s meta label changes

swiatekm:
**Description:**

Add in-process conformance tests for handling K8s meta label changes to the target allocator. The tests use a fake Prometheus Service Discovery, but otherwise exercise the whole target allocation pipeline.

The DaemonSet test is a regression test for https://github.com/open-telemetry/opentelemetry-operator/issues/4839 and fails without the fix for that issue.

https://github.com/open-telemetry/opentelemetry-operator/issues/4839

### DIFF
--- a/cmd/otel-allocator/integrationtest/conformance_test.go
+++ b/cmd/otel-allocator/integrationtest/conformance_test.go
@@ -144,11 +144,21 @@ func configureFastScrape(scrapeConfigs []*promconfig.ScrapeConfig) {
 	}
 }
 
-// collectMetrics flattens the sink's OTLP batches into a map from metric name to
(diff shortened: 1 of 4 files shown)


iblancasa (approved):
I don't know if for this kind of things it would make sense to explore weaver.

swiatekm:
> I don't know if for this kind of things it would make sense to explore weaver.

Not for these. Weaver is for checking if emitted telemetry follows a spec. It's a lot of ceremony if you just want to know if a given set of labels is correct on an arbitrary metric.

It also has absolutely no opinions on scrape targets, which these tests are very concerned with.

PR #5415, opened 2026-08-04
clusterobservability: make exporter opaque type

jinja2:

`ClusterObservability.spec.exporter` currently maintains a typed copy of the collector's OTLP HTTP exporter configuration. This can fall out of sync as exporter settings change upstream.

This updates `spec.exporter` to use `AnyConfig` and passes it unchanged to the generated agent and cluster Collectors. Existing configurations keep the same YAML shape, while newer OTLP HTTP exporter settings can be used without updating the CRD. Exporter configuration is preserved without translating individual fields. Supported settings are validated by the Collector when it loads the generated configuration.

This follows up on https://github.com/open-telemetry/opentelemetry-operator/pull/4475#discussion_r2499485317.

- Added unit tests for JSON compatibility, deep-copy behavior, and exact exporter pass-through.
- Updated the Chainsaw test to verify an opaque exporter setting in both generated collectors.
- Tested end to end on ROSA/OpenShift, and Talos.

Updated the ClusterObservability documentation and sample to clarify that the exporter configuration is passed through unchanged.


apis/v1alpha1/clusterobservability_types.go
```diff
@@ -5,153 +5,18 @@ package v1alpha1
 
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-)
-
-// OTLPHTTPExporter defines OTLP HTTP exporter configuration.
-// This structure mirrors the official OpenTelemetry Collector otlphttpexporter configuration.
-type OTLPHTTPExporter struct {
```
pavolloffay, line 12: Did you test if this is not breaking users? 

As far as I know, we should not be making breaking changes int the CRDs

jinja2: Yes. I ran an in-place upgrade from the pre-change CRD/operator to this PR. The existing resource kept the same UID and exporter config, reconciled successfully, and both Collectors stayed Ready.

The CRD change is schema-relaxing and keeps the k8s wire format unchanged. Direct Go users of the v1alpha1 type will need to update; I considered that acceptable for v1alpha1, but we can revisit it if Go source compatibility must also be maintained.

swiatekm: Isn't ClusterObservability explicitly experimental? Should be fine to make breaking changes to it if they're helpful.

jinja2: Yeah, it is behind an alpha feature gate and isn’t included by default in any installation.

jinja2: Hey folks, just checking in on this. Given that this is experimental and existing CR YAML remains compatible, I’d prefer to keep it as-is. Does that sound reasonable? Thanks!

swiatekm (changes requested):
This doesn't have anything to do with your PR, but the e2e tests for this CR are currently disabled for being flaky. https://github.com/open-telemetry/opentelemetry-operator/issues/5582 needs to be resolved before we can merge any further changes to it.

PR #5420, opened 2026-08-06
[Chore] Fix OBI smoke tests for OpenShift and multi-node clusters

IshwarKanse:
## Summary

- Use `nginx-unprivileged:stable` on port 8080 instead of `nginx:stable` on port 80 in OBI target workloads. Standard nginx binds port 80, which fails on OpenShift where the restricted SCC runs containers with a random non-root UID.
- Inline telemetry steps (`wait-for-telemetry`, `validate-collector-spans`) to check **all** DaemonSet collector pods instead of only the first. OBI eBPF traces only appear in the collector pod on the same node as the target workload, so picking `items[0]` is flaky on multi-node clusters.
- For the unprivileged OBI test, skip telemetry steps on OpenShift with kernel < 6.4 (OCP 4.x / RHEL 9) where the base-6 capabilities are insufficient for the sk_msg/sockhash eBPF path. Spec reconciliation (DaemonSet, RBAC, CR status) is still validated on all platforms.

## Test plan

- [x] `smoke-collector-obi` passes on OpenShift 4.x (3-node cluster, kernel 5.14)
- [x] `smoke-collector-obi-unprivileged` passes on OpenShift 4.x — spec reconciliation validated, telemetry steps correctly skipped with kernel < 6.4
- [x] Both tests pass when run in parallel
- [x] Verify tests still pass on kind (single-node, kernel >= 6.4 — telemetry steps should run fully)

### DIFF
--- a/tests/e2e/smoke-collector/smoke-collector-obi-unprivileged/01-install-app.yaml
+++ b/tests/e2e/smoke-collector/smoke-collector-obi-unprivileged/01-install-app.yaml
@@ -15,13 +15,13 @@ spec:
     spec:
       containers:
         - name: nginx
-          image: mirror.gcr.io/library/nginx:stable
+          image: mirror.gcr.io/nginxinc/nginx-unprivileged:stable
           ports:
-            - containerPort: 80
+            - containerPort: 8080
           readinessProbe:
             httpGet:
               path: /
-              port: 80
+              port: 8080
             initialDelaySeconds: 3
             periodSeconds: 5
 ---
@@ -34,4 +34,4 @@ spec:
     app: obi-target
   ports:
     - port: 80
-      targetPort: 80
+      targetPort: 8080
--- a/tests/e2e/smoke-collector/smoke-collector-obi-unprivileged/chainsaw-test.yaml
+++ b/tests/e2e/smoke-collector/smoke-collector-obi-unprivileged/chainsaw-test.yaml
@@ -7,12 +7,18 @@ metadata:
 # fine-grained Linux capabilities instead of a privileged container.
 # Validates that the operator reconciles the unprivileged CR into the
 # correct DaemonSet spec and that the OBI receiver produces traces.
+#
+# On OpenShift 4.x (kernel < 6.4), the base-6 capabilities are
+# insufficient for eBPF trace generation — the sk_msg/sockhash path is
+# disabled due to a kernel locking bug. The spec reconciliation is still
+# validated but the telemetry steps are skipped. See PR #5219 capability
+# findings for details.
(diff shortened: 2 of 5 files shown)


swiatekm (commented):
Does chainsaw lot let us check the support once and then skip the rest of the test? Having to do it in each step sucks.

IshwarKanse:
Unfortunately, Chainsaw v0.2.15 doesn't support this — I investigated several approaches:

1. **Step outputs don't propagate across steps.** A `check-kernel` step that sets an `ebpfSupported` output can't be referenced by later steps — Chainsaw scopes outputs to the step they're defined in. I verified this with a minimal repro that fails with `variable not defined`.

2. **No conditional `skip` at the step level.** Chainsaw v0.2 only has `spec.skip: true` (a static boolean that skips the entire test), not a per-step runtime condition.

3. **`use: template:` is mutually exclusive with `try:`.** So we can't combine a guard script with a step template in the same step.

To reduce the duplication, the actual kernel check logic lives in a single `check-ebpf-support.sh` helper script — each step just calls `./check-ebpf-support.sh || exit 0` (one line).

Chainsaw v1 (currently in development) may add richer step-level conditionals, which would let us clean this up further.

iblancasa (approved):

IshwarKanse:
### Upstream CI Verification (kind cluster, Kubernetes 1.35)

Both OBI smoke tests passed with all telemetry steps fully executing:

**`smoke-collector-obi` (privileged)** — PASS (34.41s)
- All 6 steps ran: `setup-scc` → `install-collector` → `install-target-workload` → `send-traffic` → `wait-for-telemetry` → `validate-collector-spans`
- `wait-for-telemetry`: Found `k8s.namespace.name` in collector logs after 5s (1 retry)
- `validate-collector-spans`: `otelcol_exporter_sent_spans = 3` on pod `obi-collector-collector-7lggg` — first attempt success

**`smoke-collector-obi-unprivileged`** — PASS (45.06s)
- All 6 steps ran — `check-ebpf-support.sh` correctly returned 0 on kind (no OpenShift SCC API), so **no steps were skipped**
- `send-traffic`: Successfully hit `nginx-unprivileged` on port 8080
- `wait-for-telemetry`: Found `k8s.namespace.name` after 15s (3 retries)
- `validate-collector-spans`: `otelcol_exporter_sent_spans = 3` on pod `obi-collector-collector-7qd45` — first attempt success

The `check-ebpf-support.sh` guard only activates on OpenShift with kernel < 6.4 — on kind, all telemetry verification runs fully.

swiatekm (approved):

PR #4482, opened 2025-11-06
for daemonset copy only spec.config not entire deepcopy

Nitesh-vaidyanath:
**Description:** <Describe what has changed.>

**Link to tracking Issue(s):** <Issue number if applicable>

- Resolves: #4481

**Testing:** <Describe what testing was performed and which tests were added.>

**Documentation:** <Describe the documentation added.>

PR #5422, opened 2026-08-07
fix(opamp): preserve operator-owned fields when applying remote configs

### LINKED ISSUE #4481: OpAMP Bridge: collector.DeepCopy() causes full spec updates → DaemonSet mode fails (should patch spec.config only)
### Component(s)

opamp

## Description
When using the Operator’s OpAMP Bridge to update an OpenTelemetryCollector that runs in spec.mode: daemonset, the bridge fails with a webhook denial even if the remote config only intends to change spec.config. Manual kubectl patch of only spec.config succeeds, but the bridge’s update fails.

I suspect the bridge sends a full-object update (based on collector.DeepCopy() in Apply()), which unintentionally “touches” fields that the webhook forbids for DaemonSet pools, causing the rejection.

## Steps to Reproduce

`go
// cmd/operator-opamp-bridge/internal/operator/client.go
updatedCollector := collector.DeepCopy()
...
return c.update(ctx, instance, updatedCollector)`

### Kubernetes Version

1.31

### Operator version

0.138.0

### Collector version

0.138.0

## Environment
	•	OpenTelemetry Operator: v0.138.0
	•	OpAMP Bridge image: ghcr.io/open-telemetry/opentelemetry-operator/operator-opamp-bridge:0.138.0
	•	CR namespace: opentelemetry-operator-system
	•	Collector image: otel/opentelemetry-collector-contrib:latest
	•	CR labels: opentelemetry.io/opamp-managed: "true"
	•	CR mode: daemonset

### Log output

```shell
{"level":"error","ts":"2025-11-06T05:55:19Z","logger":"agent","msg":"failed to apply remote config","error":"admission webhook \"vopentelemetrycollectorcreateupdatebeta.kb.io\" denied the request: the OpenTelemetry Collector mode is set to daemonset, which does not support modification","stacktrace":"github.com/open-telemetry/opentelemetry-operator/cmd/operator-opamp-bridge/internal/agent.(*Agent).onMessage
	.../agent.go:467
github.com/open-telemetry/opamp-go/client/types.CallbacksStruct.OnMessage
	.../callbacks.go:161
github.com/open-telemetry/opamp-go/client/internal.(*receivedProcessor).ProcessReceivedMessage
	.../receivedprocessor.go:153
github.com/open-telemetry/opamp-go/client/internal.(*wsReceiver).ReceiverLoop
	.../wsreceiver.go:92"}
```

### Additional context

From testing, the operator webhook blocks DaemonSet updates that touch:
	•	spec.mode
	•	spec.image
	•	spec.targetAllocator
	•	spec.volumes
	•	spec.volumeMounts
	•	spec.env
	•	spec.resources
	•	spec.upgradeStrategy
(and sometimes other non-config spec mutations)

Manual config-only patches work, so the failure likely comes from the bridge’s full update including zero-values or unchanged fields that still register as a spec mutation.

### Tip

theycallmeloki:
fix(opamp): preserve operator-owned fields when applying remote configs

Fixes #4481

## Problem

The OpAMP bridge built the object it submitted from the received remote configuration alone (`yaml.Unmarshal` into a fresh `OpenTelemetryCollector`, then a full `Update` with only `ObjectMeta`/`TypeMeta` taken from the cluster instance). A remote configuration typically carries only `spec.config` (and occasionally `spec.replicas`), so the full update reset the rest of the spec that is owned by the operator — `spec.mode`, image, ingress, target allocation strategy, … — and tripped mode-specific webhook validation.

For **DaemonSet pools** this surfaced as a hard failure: a remote configuration that carries the target allocator settings but not `spec.mode` produced `mode: ""` + `targetAllocator.enabled: true`, which the validating webhook rejects (`validateTargetAllocatorConfig`, collector_webhook.go:321). A manual `kubectl patch` of only `spec.config` succeeds. (There was an earlier attempt at this, #4482, which stalled on a changelog entry and was closed as inactive.)

## Fix

Build the desired object from the **current instance** (fetched on every `Apply`) with only the fields the remote configuration manages overlaid — `spec.config`, and `spec.replicas` when present — and apply it as a merge patch via controller-runtime's `client.MergeFrom`:

```go
desired := o.DeepCopy()
desired.Spec.Config = updated.Spec.Config
if updated.Spec.Replicas != nil {
    desired.Spec.Replicas = updated.Spec.Replicas
}
return c.k8sClient.Patch(ctx, desired, client.MergeFrom(o))
```

- The effective submitted CR is identical to the one in the cluster except for the remote-managed fields: operator-owned fields (`spec.mode`, image, ingress, targetAllocator, …) are preserved by construction, so mode-specific webhook validation passes.
- The validated collector object is what gets applied — no re-parsing of the raw remote body.
- No manual JSON marshalling; controller-runtime computes the merge patch (the same `MergeFrom` pattern used by the operator's status reconciliation).
- A remote configuration carrying neither config nor replicas is rejected as managing nothing.
- The `create` path is unchanged.

## Tests

- `TestClient_ApplyUpdatePreservesOperatorOwnedFields` — regression test for the #4481 shape: a DaemonSet-mode collector with a per-node target allocator updated with a config-only remote configuration keeps `spec.mode=daemonset`, `targetAllocator.enabled`, and `targetAllocator.allocationStrategy=per-node`, while `spec.config` is updated.
- `TestClient_ApplyUpdateRemoteConfigWithoutSpec` — a remote configuration carrying neither config nor replicas is rejected cleanly.
- All existing `cmd/operator-opamp-bridge` tests pass (including the replicas-update contract).


swiatekm:
The change looks fine to me, but I'd like to understand why it's necessary. In principle, the collector CR we submit for an update should be identical to the one in the cluster. So patching just the config feels like working around a bug a layer above. I'd rather fix that bug instead. Can you reproduce this with a specific set of opamp messages?

theycallmeloki:
> The change looks fine to me, but I'd like to understand why it's necessary. In principle, the collector CR we submit for an update should be identical to the one in the cluster. So patching just the config feels like working around a bug a layer above. I'd rather fix that bug instead. Can you reproduce this with a specific set of opamp messages?

Ah, I have missed that consideration, I assumed the bridge built the object it submitted, so the `Apply` op unmarshalled the `AgentConfigFile` into a new `OpenTelemetryCollector` and deep-copies it (`updatedCollector`) so the spec fields carry only the remote config mentions `spec.config` and `spec.replicas` 

The old `update` op only copied `ObjectMeta` and `TypeMeta` and updated only the minimal spec, so downstream fields like `spec.mode`, `image`, `ingress`, `targetAllocator` was reset to its zero value on every remote config 

A minimal OpAmp reproduction: 

A collector: 

```yaml
   kind: OpenTelemetryCollector                                                                                                                                                                                                                                                                                                                               
   metadata:                                                                                                                                                                                                                                                                                                                                                  
     name: simplest                                                                                                                                                                                                                                                                                                                                           
     namespace: default                                                                                                                                                                                                                                                                                                                                       
     labels:                                                                                                                                                                                                                                                                                                                                                  
       opentelemetry.io/opamp-managed: "true"                                                                                                                                                                                                                                                                                                                 
   spec:                                                                                                                                                                                                                                                                                                                                                      
     mode: daemonset                                                                                                                                                                                                                                                                                                                                          
     targetAllocator:                                                                                                                                                                                                                                                                                                                                         
       enabled: true                                                                                                                                                                                                                                                                                                                                          
       allocationStrategy: per-node                                                                                                                                                                                                                                                                                                                           
     config: { ... }                
```

What the control plan manages `spec.config` and `spec.TargetAllocator` barring `spec.mode` 
                                                                                                                                                                                                                                                                                                                                           
 ```yaml                                                                                                                                                                                                                                                                                                                                                      
   kind: OpenTelemetryCollector                                                                                                                                                                                                                                                                                                                               
   metadata:                                                                                                                                                                                                                                                                                                                                                  
     name: simplest                                                                                                                                                                                                                                                                                                                                           
     namespace: default                                                                                                                                                                                                                                                                                                                                       
   spec:                                                                                                                                                                                                                                                                                                                                                      
     targetAllocator:                                                                                                                                                                                                                                                                                                                                         
       enabled: true                                                                                                                                                                                                                                                                                                                                          
       allocationStrategy: per-node                                                                                                                                                                                                                                                                                                                           
     config: { ... }   # updated                                                                                                                                                                                                                                                                                                                              
 ```                               

^ In this case, bridge's `updatedCollector` ends up with `mode: ""` + `targetAllocator.enabled: true` leading to a full update, the validating webhook rejects this with `the OpenTelemetry Collector mode is set to "", which does not support the target allocation deployment` 

^ same case with A manual kubectl patch of just `spec.config` succeeds.     

In the latter's case, config-only remote the update passes, but silentry resets spec.mode and the allocator leading to the same shape as #4481                                             

Let me try to rework the fix to submit a CR that's identical to the cluster by building the desired object from the current instance and overlay the remote fields, then apply a full update. This if it passes the DaemonSet case should address the right layer.

theycallmeloki:
Done, reworked it as per your suggestion. The bridge now builds the update from the current instance (fetched from every Apply) overlays the spec keys the remote carries, and submits a full Update. The CR is identical to the one in the cluster except for remote-managed fields, and the update carries the instance's `resourceVersion` so concurrent changes surface as conflicts instead of being silently clobbered. 

Main thing was remote doesn't carry `spec.Mode` at all, so overlay happens at map level, because a struct-level overlay can't distinguish `mode` absent from `mode: ""` 

The daemonset regression test passes, and the existing can update config and replicas contract remains in effect.

cmd/operator-opamp-bridge/internal/operator/client.go
```diff
@@ -116,7 +117,7 @@ func (c Client) Apply(key string, configmap *protobufs.AgentConfigFile) error {
 	if instance == nil {
 		return c.create(ctx, name, namespace, updatedCollector)
 	}
-	return c.update(ctx, instance, updatedCollector)
+	return c.update(ctx, instance, configmap.Body)
```
jaronoff97: Why are we submitting the configmap body here after we go through all the work to do the proper validation here?

cmd/operator-opamp-bridge/internal/operator/client.go
```diff
@@ -197,12 +198,55 @@ func (c Client) create(ctx context.Context, name, namespace string, collector *v
...
-	n.TypeMeta = o.TypeMeta
+func (c Client) update(ctx context.Context, o *v1beta1.OpenTelemetryCollector, remoteConfig []byte) error {
+	// Build the desired object from the current instance so the CR we submit
+	// is identical to the one in the cluster except for the spec keys the
+	// remote configuration actually carries. Operator-owned fields (mode,
+	// image, ingress, targetAllocator, ...) are preserved by construction,
+	// which keeps mode-specific webhook validation satisfied (e.g. DaemonSet
+	// pools). The overlay happens at the map level because a struct-level
+	// overlay cannot distinguish "field absent" from "field zero" (e.g.
+	// spec.mode: "").
```
jaronoff97: This type of json marshalling and unmarshalling is very fraught and potentially dangerous. I would much rather we use known methods like Patch if the goal is to just overlay the changes.

jaronoff97 (changes requested):

cmd/operator-opamp-bridge/internal/operator/client.go
theycallmeloki: Very fair point, I've reworked `update()` to use the validated `updateCollector` instead of the raw `configmap.Body` without parsing after validation, and the object that passed `validateComponents`/`validateLabels` is applied

cmd/operator-opamp-bridge/internal/operator/client.go
theycallmeloki: Agree completely, I have reworked it to not use the round-trip and instead use the typed path: the full object is built from the current instance and only the managed fields are overlaid (`spec.Config` and `spec.replicas` when present) and it's applied with `client.Patch(ctx, desired, client.MergeFrom(o))` 

The same `MergeFrom` that operator's status reconciliation uses, thanks for pointing me to this. 

Building from the current instance also addresses @swiatekm 's earlier point as the effective submitted CR is now identical to the one in cluster except for the remote fields

jaronoff97:
@theycallmeloki this is better, but we should have some more testing to verify this will solve the underlying problem here.

theycallmeloki:
Added a test that exercises the validating webhook against the #4481 shape                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                              
 TestClient_ApplyUpdateSurvivesWebhookValidation:                                                                                                                                                                                                                                                                                                             
 - DaemonSet-mode collector with a per-node target allocator, updated with a remote configuration that carries the target allocator settings but not `spec.mode`                                                                                                                                                                                                
 - Asserts the updated collector passes `CollectorWebhook.Validate` (the actual validation the API server runs)                                                                                                                                                                                                                                                 
 - Counterfactual: what the old full-object update produced: spec.mode reset to "" with the allocator still enabled: is rejected with "does not support the target allocation deployment"                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                              
So the test proves both halves: the new update path survives the webhook, and the old path demonstrably would not have. 

Had to also mention: the test's collector config now includes a `receivers.prometheus` section because that's a prerequisite the webhook itself enforces for target-allocator-enabled collectors, not sure if right, can't drop it or it won't pass through the validating webhook (I think).

swiatekm:
> DaemonSet-mode collector with a per-node target allocator, updated with a remote configuration that carries the target allocator settings but not spec.mode

This is the part that's confusing me. I don't think opamp-bridge is supposed to do this. You're supposed to send the full custom resource down each time. Do we promise otherwise in any of our documentation?

theycallmeloki:
> > DaemonSet-mode collector with a per-node target allocator, updated with a remote configuration that carries the target allocator settings but not spec.mode
> 
> This is the part that's confusing me. I don't think opamp-bridge is supposed to do this. You're supposed to send the full custom resource down each time. Do we promise otherwise in any of our documentation?

I tried to poke through documentation, I wasn't able to find anything particularly substantial: the OpAmp bridge docs are relatively minimal, and only concrete examples for standalone mode shows config-only content (receivers/exporters/pipelines), not a full CR.
                                                                                                                                                                                                                                                                                 
On #4481 case: the reporter's server sent config-centric documents: the report says the remote config "only intends to change spec.config". The old full Update then touched other spec fields, and the DaemonSet webhook rejected it with "the OpenTelemetry Collector mode is set to daemonset, which does not support modification" (the reporter's  
 log). The reporter's own analysis lists the fields the webhook protects in daemonset mode: `mode, image, targetAllocator, volumes, env, resources, upgradeStrategy`                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                              
That last part is why the merge behavior is needed even for full-CR senders: the DaemonSet webhook rejects any update that touches those protected fields, so the CR we submit must differ from the one in the cluster only in the fields we actually intend to change: whatever the sender sends. A full-CR sender is fine as long as the submitted        
 document matches the cluster CR except for the managed fields; the merge guarantees that by construction.

swiatekm:
> > > DaemonSet-mode collector with a per-node target allocator, updated with a remote configuration that carries the target allocator settings but not spec.mode
> > 
> > 
> > This is the part that's confusing me. I don't think opamp-bridge is supposed to do this. You're supposed to send the full custom resource down each time. Do we promise otherwise in any of our documentation?
> 
> I tried to poke through documentation, I wasn't able to find anything particularly substantial: the OpAmp bridge docs are relatively minimal, and only concrete examples for standalone mode shows config-only content (receivers/exporters/pipelines), not a full CR.
> 
> On #4481 case: the reporter's server sent config-centric documents: the report says the remote config "only intends to change spec.config". The old full Update then touched other spec fields, and the DaemonSet webhook rejected it with "the OpenTelemetry Collector mode is set to daemonset, which does not support modification" (the reporter's log). The reporter's own analysis lists the fields the webhook protects in daemonset mode: `mode, image, targetAllocator, volumes, env, resources, upgradeStrategy`
> 
> That last part is why the merge behavior is needed even for full-CR senders: the DaemonSet webhook rejects any update that touches those protected fields, so the CR we submit must differ from the one in the cluster only in the fields we actually intend to change: whatever the sender sends. A full-CR sender is fine as long as the submitted document matches the cluster CR except for the managed fields; the merge guarantees that by construction.

The claims in the issue are false. All of those fields, aside from `mode` itself, can be changed on a DaemonSet Collector. 

I'm also pretty sure your change is breaking. Right now, with full updates, it's possible to switch from a set field to the default. For example, I can have an OpenTelemetryCollector with a custom image, clear the field, and this will be applied. After your change, this won't be the case.

In any case, the original issue isn't asking for a bug fix, but for a change in behavior. We should document the current behavior, but we're not doing breaking changes unless there's a very good reason. I'm not approving this PR until the project direction is clear, and that discussion belongs in the issue.

PR #1798, opened 2023-06-05
add prometheus metric exporter support for nodejs autoinstrumentation

jakobjoachim:
The Autoinstrumentation Image only supports the OTLPMetricExporter metrics exporter until now. 

This reads the environment variable OTEL_METRICS_EXPORTER and uses the configured one or throws and error if anything else is set.

PR #5425, opened 2026-08-09
fix(auto-instrumentation): default OTEL_METRICS_EXPORTER to otlp for Node.js

### LINKED ISSUE #3768: Latest operator autoinstrumention-nodejs image loses Metrics Reader
### Component(s)

auto-instrumentation

## Description
The latest image opentelemetry-operator/autoinstrumentation-nodejs@sha256:8f46a6efd93e8272122bf000d42bcc53fbe52e5c430ca28073c9f1840e346a02 due to https://github.com/open-telemetry/opentelemetry-operator/pull/3570/files#diff-6d911017f3b5ee42e61e3cea12d02cbcdef6d815d959937231dff56ee938d416 loads a very minimal `NodeSDK`:
```javascript
const sdk = new opentelemetry.NodeSDK({
    instrumentations: (0, utils_1.getNodeAutoInstrumentations)(),
    resourceDetectors: (0, utils_1.getResourceDetectorsFromEnv)(),
});
```
whereas older images like opentelemetry-operator/autoinstrumentation-nodejs@sha256:e25b28350ce8fdfb37c58afa87096484fb4e986b57207942d9bfcbf6a912bc0f loaded a more featured SDK:
```javascript
const sdk = new NodeSDK({
    autoDetectResources: true,
    instrumentations: [getNodeAutoInstrumentations()],
    traceExporter: getTraceExporter(),
    metricReader: getMetricReader(),
```

This specifically breaks users who were exporting metrics.  The [SDK docs](https://www.npmjs.com/package/@opentelemetry/sdk-node#configuration)
show most options have a default except:
> metricReader
> Add a [MetricReader](https://github.com/open-telemetry/opentelemetry-js/blob/opentelemetry-sdk-metrics/src/export/MetricReader.ts) that will be passed to the MeterProvider. If metricReader is not configured, the metrics SDK will not be initialized and registered.

Please advise on fix for projects that get auto-instrumentation via the operator and then decided to start using other SDK features.

## Steps to Reproduce
Use operator image released in last two weeks.

## Expected Result
Existing metrics continue to export.

## Actual Result
Metrics are not exported.

### Kubernetes Version

1.31.5

### Operator version

0.111.0

### Collector version

?

### Environment information

_No response_

### Log output

```shell

```

### Additional context

_No response_

pujitha24:
**Description:** The Node.js SDK only initializes its metrics pipeline when a metric reader is configured. Since the operator's Node.js auto-instrumentation webhook wasn't setting one (or the corresponding env var), metrics silently stopped being exported for anyone using Node.js auto-instrumentation and relying on the previous default behavior, even though traces and logs continued to work. This change defaults `OTEL_METRICS_EXPORTER` to `otlp` in `getDefaultNodeJSEnvVars` (`internal/instrumentation/nodejs.go`) when the user hasn't already set that env var on the container, mirroring the default already applied for Python auto-instrumentation in `getDefaultPythonEnvVars` (`internal/instrumentation/python.go`).

- Resolves: #3768

**Testing:** Updated unit test expectations in `nodejs_test.go`, `sdk_test.go`, and `podmutator_test.go` to assert `OTEL_METRICS_EXPORTER=otlp` is injected by default and left untouched when already set by the user. Confirmed the tests are meaningful by temporarily reverting only `nodejs.go` (keeping the updated tests) and verifying `TestInjectNodeJS`, `TestInjectNodeJSSDK`, and `TestMutatePod` fail without the fix and pass with it restored. Ran and passed `go build ./...`, `go test ./internal/instrumentation/...`, `gofmt -l internal/instrumentation/nodejs.go internal/instrumentation/nodejs_test.go internal/instrumentation/podmutator_test.go internal/instrumentation/sdk_test.go`, `go vet ./internal/instrumentation/...`, and `make chlog-validate`.

**Documentation:** No documentation changes were required; this restores previous default behavior for Node.js metrics export.

---
AI assistance: this change was drafted with Claude Code.

Fixes #3768

### DIFF
--- a/.chloggen/fix-nodejs-metrics-exporter-default.yaml
+++ b/.chloggen/fix-nodejs-metrics-exporter-default.yaml
@@ -1,0 +1,20 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: auto-instrumentation
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Default `OTEL_METRICS_EXPORTER` to `otlp` for Node.js auto-instrumentation so metrics are exported without extra configuration.
+
+# One or more tracking issues related to the change
+issues: [3768]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The Node.js SDK only initializes its metrics pipeline when a metric reader is configured, and the
+  webhook was not setting one. As a result, metrics silently stopped being exported for anyone relying
+  on the previous default behavior. The webhook now sets `OTEL_METRICS_EXPORTER=otlp` by default,
+  matching the existing behavior for Python auto-instrumentation, unless the user already set it.
--- a/internal/instrumentation/nodejs.go
+++ b/internal/instrumentation/nodejs.go
@@ -74,26 +74,32 @@ func injectNodeJSSDK(nodeJSSpec v1alpha1.NodeJS, pod *corev1.Pod, containers []*
 }
 
 func getDefaultNodeJSEnvVars(container *corev1.Container) []corev1.EnvVar {
+	var envVars []corev1.EnvVar
+
 	idx := getIndexOfEnv(container.Env, envNodeOptions)
 	if idx == -1 {
-		return []corev1.EnvVar{
-			{
-				Name:  envNodeOptions,
-				Value: nodeRequireArgument,
-			},
-		}
-	} else if idx > -1 {
-		// Don't modify NODE_OPTIONS if it uses ValueFrom
-		if container.Env[idx].ValueFrom != nil {
-			return []corev1.EnvVar{}
-		}
-		// NODE_OPTIONS is set, append the required argument
-		return []corev1.EnvVar{
-			{
-				Name:  envNodeOptions,
-				Value: container.Env[idx].Value + nodeRequireArgument,
-			},
-		}
+		envVars = append(envVars, corev1.EnvVar{
+			Name:  envNodeOptions,
+			Value: nodeRequireArgument,
+		})
+	} else if container.Env[idx].ValueFrom == nil {
+		// NODE_OPTIONS is set, append the required argument.
+		// Don't modify NODE_OPTIONS if it uses ValueFrom.
+		envVars = append(envVars, corev1.EnvVar{
+			Name:  envNodeOptions,
+			Value: container.Env[idx].Value + nodeRequireArgument,
+		})
 	}
-	return []corev1.EnvVar{}
+
+	// Set OTEL_METRICS_EXPORTER to otlp if not set by the user. The Node.js SDK only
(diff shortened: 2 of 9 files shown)


swiatekm:
What I find confusing about this is that we have tests checking if we get data out of an instrumented application, and they do verify metrics too. I'll have to look into why that is, if the default config doesn't send anything.

pujitha24:
Thanks for looking into this, @swiatekm. I dug into the existing e2e coverage to check that: `tests/e2e-instrumentation/instrumentation-nodejs/chainsaw-test.yaml` (via its shared step templates) does verify Node.js metrics end-to-end — it asserts `otelcol_receiver_accepted_metric_points`/`otelcol_exporter_sent_metric_points` > 1 on the collector's Prometheus endpoint, and greps the collector's debug-exporter logs for a real Node HTTP metric (`http.server.duration`). But that test's `Instrumentation` CR explicitly sets `OTEL_METRICS_EXPORTER: otlp` in `spec.env` (`00-install-instrumentation.yaml`, added in #1798, unchanged since), which is applied via `injectCommonEnvVar` before `getDefaultNodeJSEnvVars` runs — so that test only exercises the "user explicitly set it" path and would pass identically with or without this PR.

I did find a few fixtures that exercise the actual unset-default path this PR changes — e.g. `tests/e2e-multi-instrumentation/instrumentation-multi-multicontainer`, `instrumentation-single-instr-first-container`, and the Node.js app in `tests/e2e-clusterobservability/smoke-clusterobservability` — but they only assert the injected pod env var or that spans reach the sink, not metric delivery through the collector, so they wouldn't have caught the original bug either. I couldn't find any e2e test that both leaves `OTEL_METRICS_EXPORTER` unset for Node.js *and* verifies metrics actually reach the collector. Happy to add one if that would help confirm the behavior — let me know what you find when you dig further, or if you'd like me to add that coverage.

swiatekm:
> Thanks for looking into this, @swiatekm. I dug into the existing e2e coverage to check that: `tests/e2e-instrumentation/instrumentation-nodejs/chainsaw-test.yaml` (via its shared step templates) does verify Node.js metrics end-to-end — it asserts `otelcol_receiver_accepted_metric_points`/`otelcol_exporter_sent_metric_points` > 1 on the collector's Prometheus endpoint, and greps the collector's debug-exporter logs for a real Node HTTP metric (`http.server.duration`). But that test's `Instrumentation` CR explicitly sets `OTEL_METRICS_EXPORTER: otlp` in `spec.env` (`00-install-instrumentation.yaml`, added in #1798, unchanged since), which is applied via `injectCommonEnvVar` before `getDefaultNodeJSEnvVars` runs — so that test only exercises the "user explicitly set it" path and would pass identically with or without this PR.
> 
> I did find a few fixtures that exercise the actual unset-default path this PR changes — e.g. `tests/e2e-multi-instrumentation/instrumentation-multi-multicontainer`, `instrumentation-single-instr-first-container`, and the Node.js app in `tests/e2e-clusterobservability/smoke-clusterobservability` — but they only assert the injected pod env var or that spans reach the sink, not metric delivery through the collector, so they wouldn't have caught the original bug either. I couldn't find any e2e test that both leaves `OTEL_METRICS_EXPORTER` unset for Node.js _and_ verifies metrics actually reach the collector. Happy to add one if that would help confirm the behavior — let me know what you find when you dig further, or if you'd like me to add that coverage.

Thanks for looking into this! Could you unset those variables in the e2e test setup as part of this PR, then?

swiatekm (approved):

iblancasa (approved):

PR #5447, opened 2026-08-12
Add init container options for auto injections

dineshg13:
**Description:**  Add an optional initContainer configuration to each supported language

**Link to tracking Issue(s):** [<Issue number if applicable>](https://github.com/open-telemetry/opentelemetry-operator/issues/5485)

 

**Testing:** Tested locally via Kind cluster. 
 PS: Human (I) wrote this description. Most of the code is built using codex


swiatekm:
What do you need this for? I don't see any issue, nor does your change add documentation showing how these fields could be used.

dineshg13:
@swiatekm We (Datadog) have our own SDKs. They are OTel API compatible. We would like to use those SDKs with OTel operator.

PR #5452, opened 2026-08-12
fix(opamp): rebuild applied collector keys from cluster state on bridge restart

### LINKED ISSUE #5445: OpAMP Bridge forgets which collector CRs it manages after a restart
### Component(s)

opamp

## Description

opamp-bridge keeps the set of remote config keys it has applied in an in-memory map(`agent.appliedKeys`). On restart that map starts out empty and nothing rebuilds it from the cluster.

Once the bridge has restarted, it no longer knows about any collector it applied before. If a later remote config drops one of those entries, the delete never happens.

The map is created empty in `NewAgent`.
```go
// agent/agent.go
agent := &Agent{
	appliedKeys: map[string]bool{},
	...
}
```

Deletion is computed as the difference between that map and the keys in the incoming config, so right after a restart there is nothing to iterate over.
```go
for key := range agent.appliedKeys {
	if _, ok := config.Config.GetConfigMap()[key]; !ok {
		if err := agent.applier.Delete(key); err != nil {
			errs = append(errs, err)
		}
	}
}
```

## Steps to Reproduce

1. Send the bridge a remote config containing a single entry `ns/collector`. One collector CR is created.
2. Restart the bridge pod.
3. Send a remote config whose config map is empty.

## Expected Result

`ns/collector` is deleted.

## Actual Result

`ns/collector` is still there.

### Kubernetes Version

1.26.4

### Operator version

v0.152.0

### Collector version

v0.152.0

### Environment information

_No response_

### Log output

```shell

```

### Additional context

_No response_

### Tip

pujitha24:
**Description:** The OpAMP Bridge tracks which collector remote-config keys it has applied in an in-memory map (`Agent.appliedKeys`). This map is only ever created empty and is never rebuilt from cluster state, so after a bridge restart the bridge forgets about every collector it previously applied. Deletion is computed as the set difference between `appliedKeys` and the keys in an incoming remote config, so right after a restart there is nothing to diff against — a later remote config that drops a previously-applied collector never deletes it.

This change rebuilds `appliedKeys` from the cluster on `Agent.Start()` by listing the collectors the bridge currently manages and seeding their keys. Reporting-only collectors (labeled `opentelemetry.io/opamp-reporting: true` but not `opentelemetry.io/opamp-managed`) are explicitly excluded from this rebuild, since the bridge never applies config to them and must not delete them.

**Link to tracking Issue(s):** 5445

- Resolves: #5445

**Testing:** Added `TestAgent_Start_RebuildsAppliedKeysAcrossRestart` in `cmd/operator-opamp-bridge/internal/agent/agent_test.go`, which seeds a fake Kubernetes client with a previously-applied managed collector and an unrelated reporting-only collector, starts a fresh `Agent` (simulating a post-restart process), and asserts that a subsequent remote config which omits the managed collector deletes it while leaving the reporting-only collector untouched. I confirmed this test fails without the fix (`appliedKeys` stays empty after `Start()`) and passes with it, by temporarily stashing only the `agent.go` change and re-running the test.

Ran from `cmd/operator-opamp-bridge/` (its own Go module):
- `go build ./...` — passes
- `go vet ./...` — passes
- `go test ./...` — all packages pass
- `golangci-lint run ./...` — 0 issues
- `gofmt -l` on changed files — clean

I did not run the e2e OpAMP bridge suite (`make e2e-opampbridge`) against a live cluster; the fix and its regression test are scoped to the in-process `Agent`/`ConfigApplier` logic, which the added unit test exercises directly against a fake Kubernetes client covering the exact reported scenario.

**Documentation:** No user-facing documentation changes; this is an internal bug fix to bridge restart behavior. Added a changelog entry at `.chloggen/opamp-bridge-rebuild-applied-keys.yaml`.

Fixes #5445

### DIFF
--- a/.chloggen/opamp-bridge-rebuild-applied-keys.yaml
+++ b/.chloggen/opamp-bridge-rebuild-applied-keys.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: opamp
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Rebuild the OpAMP Bridge's applied-keys tracking from cluster state on startup, so a remote config that drops a collector after a bridge restart correctly deletes it.
+
+# One or more tracking issues related to the change
+issues: [5445]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/cmd/operator-opamp-bridge/internal/agent/agent.go
+++ b/cmd/operator-opamp-bridge/internal/agent/agent.go
@@ -198,6 +198,11 @@ func (agent *Agent) Start() error {
 		return err
 	}
 	agent.startTime = startTime
+
+	if err = agent.rebuildAppliedKeys(); err != nil {
+		return fmt.Errorf("failed to rebuild applied keys from cluster state: %w", err)
+	}
+
 	settings := types.StartSettings{
 		OpAMPServerURL: agent.config.Endpoint,
 		Header:         agent.config.Headers.ToHTTPHeader(),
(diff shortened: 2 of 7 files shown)


pujitha24:
Just checking in on this one — it's still green and rebased on main, happy to make any changes if something would help move review along.

jaronoff97 (approved):

PR #5459, opened 2026-08-13
fix(collector): don't reintroduce deprecated telemetry address when readers are set

### LINKED ISSUE #5416: Collector Pod Crash During Upgrade
### Component(s)

collector

Hi,
We are upgrading our set in EKS 1.33 cluster:
OTel Operator from 0.117.0 → 0.151.0
ADOT Collector from v0.43.2 → v0.47.0
Chart version used : opentelemetry-operator-0.113.1/opentelemetry-operator-0.113.1.tgz
We install otel collector as part of cluster build using Terraform.
Issue:
When we trying to upgrade Operator and collector, collector POD keep crashing with below error:
````Error: failed to create meter provider: binding address 0.0.0.0:8888 for Prometheus exporter: listen tcp 0.0.0.0:8888: bind: address already in use````
after troubleshooting issue, i found that Operator is adding 2 readers for otel internal metrics, one by default and other one , which we pass in collector config, snippet of CM
````apiVersion: v1
data:
  collector.yaml: |
    receivers:
      otlp:
        protocols:
          grpc:
            endpoint: 0.0.0.0:4317
          http:
            endpoint: 0.0.0.0:4318
    exporters:
      debug:
        verbosity: detailed
    processors:
      batch: null
      memory_limiter:
        check_interval: 1s
        limit_mib: 256
    service:
      telemetry:
        metrics:
          readers:
            - pull:
                exporter:
                  prometheus:
                    host: 0.0.0.0
                    port: 8888
                    without_type_suffix: true
                    without_units: true
            - pull:
                exporter:
                  prometheus:
                    host: 0.0.0.0
                    port: 8888````
But when we delay Collector installation by 2-3 minutes after Operator installation, we don't face this issue.
I want to confirm, is it issue or I am doing something wrong.
Or its always best pratice to keep 2-3  minutes delays between operator & collector install.


### Tip

### PR DESCRIPTION
**Pull request by pujitha24**
**Description:** Fixes the `0.111.0` automatic-upgrade step so it no longer re-adds the deprecated `service.telemetry.metrics.address` field on a collector that already has `service.telemetry.metrics.readers` configured. Previously, `applyDefaults` in `pkg/collector/upgrade/v0_111_0.go` unconditionally backfilled `address` via `mergo.Merge` whenever it wasn't already set, without checking whether an equivalent `readers` entry already existed. The later `0.122.0` upgrade step then migrates any leftover `address` into a *new* Prometheus reader, assuming none exists yet for that endpoint — leaving two readers bound to the same host:port. The collector then fails to start with `failed to create meter provider: binding address 0.0.0.0:8888 for Prometheus exporter: ... bind: address already in use`, exactly as reported in the issue. The fix adds a guard in `applyDefaults` that skips the `address` backfill when `Metrics.Readers` is already non-empty.

**Link to tracking Issue(s):** 5416

- Resolves: #5416

**Testing:** Added a new table-driven case `"telemetry readers already configured"` to `Test0_111_0Upgrade` in `pkg/collector/upgrade/v0_111_0_test.go`, reproducing a collector with `service.telemetry.metrics.readers` already set. Confirmed the test fails without the fix (asserting `address` was wrongly added alongside `readers`) and passes with it — this is the regression test for the bug. The root cause was originally found by temporarily instrumenting `pkg/collector/upgrade/upgrade.go` with debug logging (since removed) and tracing the existing integration test `TestUpgrade/needs-upgrade` in `internal/controllers/reconcile_test.go`, which drives a collector through the full automatic-upgrade path from status version `0.109.0` to latest via envtest with real webhook admission; the trace showed the exact duplicate-reader config forming across the `0.111.0` → `0.122.0` steps. That integration test passes both before and after this fix (it only asserts on `Status.Version` and a feature-gate arg, not the telemetry config), so it doesn't itself catch the regression, but it confirmed the diagnosis end-to-end. It still passes after the fix.

Commands run and their results:
- `go build ./...` — passes
- `go test ./pkg/collector/upgrade/... ./internal/controllers/... ./internal/otelconfig/... ./internal/webhook/...` — all pass
- `go test ./pkg/... ./internal/...` (full tree) — all pass
- `golangci-lint run ./...` — 0 issues
- `make chlog-validate` — passes

I was not able to run a live Kubernetes cluster/e2e reproduction of the reported crash-loop; the above unit/integration tests exercise the same defaulting and admission-webhook code paths the real cluster reconcile loop uses (envtest-backed), including the exact automatic-upgrade sequence that produces the duplicate readers.

Checked `gh run list --repo open-telemetry/opentelemetry-operator --branch main --limit 8`: all recent CI runs on `main` are green, so no known unrelated CI failures to disclose.

**Documentation:** No user-facing documentation changes needed; this is an internal defaulting-logic bug fix. Added a changelog entry at `.chloggen/fix-upgrade-duplicate-telemetry-readers.yaml`.

Fixes #5416

### DIFF
--- a/.chloggen/fix-upgrade-duplicate-telemetry-readers.yaml
+++ b/.chloggen/fix-upgrade-duplicate-telemetry-readers.yaml
@@ -5,7 +5,7 @@ change_type: bug_fix
 component: collector
 
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Fix the 0.111.0 upgrade step re-adding the deprecated `service.telemetry.metrics.address` field on collectors that already use `readers`, which made a later upgrade step bind a second Prometheus reader to the same port.
+note: Fix the automatic-upgrade routine binding two Prometheus readers to the same port when a collector already uses `service.telemetry.metrics.readers`.
 
 # One or more tracking issues related to the change
 issues: [5416]
@@ -16,7 +16,8 @@ issues: [5416]
 subtext: |
   When an OpenTelemetryCollector already configured `service.telemetry.metrics.readers`
   (added by earlier defaulting), the automatic version-upgrade routine still backfilled the
-  older `address` field for it. The 0.122.0 upgrade step later migrates `address` into a new
-  reader without checking whether an equivalent one already exists, leaving two readers bound
-  to the same host:port. The collector then failed to start with "address already in use".
-  The 0.111.0 step now skips backfilling `address` when `readers` is already configured.
+  older, deprecated `address` field for it. The 0.122.0 upgrade step then migrated `address`
+  into a new reader, leaving two readers bound to the same host:port. The collector then
+  failed to start with "address already in use". The 0.111.0 step now skips backfilling
+  `address` when `readers` is already configured, and the 0.122.0 step now skips adding a
+  reader for `address` if an equivalent one already exists.
--- a/.chloggen/fix-upgrade-duplicate-telemetry-readers.yaml
+++ b/.chloggen/fix-upgrade-duplicate-telemetry-readers.yaml
@@ -1,0 +1,22 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. collector, target allocator, auto-instrumentation, opamp, github action)
+component: collector
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix the 0.111.0 upgrade step re-adding the deprecated `service.telemetry.metrics.address` field on collectors that already use `readers`, which made a later upgrade step bind a second Prometheus reader to the same port.
+
+# One or more tracking issues related to the change
+issues: [5416]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  When an OpenTelemetryCollector already configured `service.telemetry.metrics.readers`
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: .chloggen/fix-upgrade-duplicate-telemetry-readers.yaml [hunk 1]
```diff
@@ -0,0 +1,22 @@
...
+issues: [5416]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  When an OpenTelemetryCollector already configured `service.telemetry.metrics.readers`
+  (added by earlier defaulting), the automatic version-upgrade routine still backfilled the
+  older `address` field for it. The 0.122.0 upgrade step later migrates `address` into a new
+  reader without checking whether an equivalent one already exists, leaving two readers bound
```
iblancasa: Wouldn't it make more sense to check if there is already one in 0.122.0?

swiatekm (approved):
LGTM

pujitha24:
Good point, done — added the same check in `upgrade0_122_0` via a `hasPrometheusReader` helper that compares host/port before appending a reader, so it only migrates `address` into a new reader if an equivalent one isn't already there. Kept the 0.111.0 guard too since it stops the deprecated `address` field from being reintroduced in the first place, while the 0.122.0 check now also covers configs that reach 0.122.0 with both fields already set. Added a test for the duplicate-reader case and it passes.

iblancasa (approved):

pujitha24:
@swiatekm — aside from this PR: I'd like to apply for OpenTelemetry org membership, which needs sponsorship from two approvers. You've reviewed several of my operator changes (#5364, #5388, #5425, #5459). Would you be willing to sponsor? I have 9 merged PRs across opentelemetry-operator, collector-contrib, collector-releases and go-contrib. Entirely fine to say no.

PR #5462, opened 2026-08-14
Fix disabling target filtering for target allocator

### LINKED ISSUE #5444: Impossible to set `filterStrategy: ""`
### Component(s)

_No response_

## Description

Hi,

For quite some time now we have the problem that the TargetAllocator scrapes all serviceMonitors, but due to the filterStrategy half of the targets are not properly ingested in our Mimir instance. We tried setting `filterStrategy` to `""`, but everytime we try this the operator sets it back to `relabel-config`.

## Steps to Reproduce

Set `filterStrategy: ""`

## Expected Result

`filterStrategy:""` should set the `filterStrategy` to `""` and not to `relabel-config`

## Actual Result

The operator always assumes `filterStrategy:""` is unset and sets it back to `relabel-config`.

### Kubernetes Version

v1.36.0-gke.3070003

### Operator version

0.115.1

### Collector version

v0.156.0

### Environment information

_No response_

### Log output

```shell

```

### Additional context

_No response_

### Tip

swiatekm:
**Description:**

The value intended to disable target filtering - `filterStrategy: ""` was lost during reconciliation. When the operator instantiates the TargetAllocator CR as a Go struct, the empty string is treated as no value, which falls back to the default after admission. 

Add a new, explicit `none` value while keeping the empty one for compatibility and fix the serialization. Fixing requires making the field a pointer and take `nil` as "no value".  This is a breaking change in the Go API, but not the CRD.

I've also made the target allocator config file parsing stricter. It now only accepts `""`, `none` and `relabel-config` as values and fails validation otherwise.

- Resolves: https://github.com/open-telemetry/opentelemetry-operator/issues/5444

**Testing:**

Unit tests and an addition to the smoke e2e test.

### DIFF
--- a/cmd/otel-allocator/internal/target/discovery_test.go
+++ b/cmd/otel-allocator/internal/target/discovery_test.go
@@ -772,7 +772,7 @@ func TestReloadAppliesRelabelFilteringWhenEnabled(t *testing.T) {
 // disables filtering during discovery: targets are kept even if relabel_configs would drop them.
 func TestReloadSkipsRelabelFilteringWhenDisabled(t *testing.T) {
 	var got []*Item
-	d := newTestDiscoverer(t, "", func(targets []*Item) { got = targets })
+	d := newTestDiscoverer(t, "none", func(targets []*Item) { got = targets })
 
 	scrapeConfigs := []*promconfig.ScrapeConfig{
 		{
--- a/cmd/otel-allocator/internal/target/discovery_test.go
+++ b/cmd/otel-allocator/internal/target/discovery_test.go
@@ -64,7 +64,7 @@ func TestDiscovery(t *testing.T) {
 	require.NoError(t, err)
 	d := discovery.NewManager(ctx, config.NopLogger, registry, sdMetrics)
 	results := make(chan []string)
-	manager, err := NewDiscoverer(ctrl.Log.WithName("test"), d, RelabelConfigFilterStrategy, scu, func(targets []*Item) {
+	manager, err := NewDiscoverer(ctrl.Log.WithName("test"), d, config.FilterStrategyRelabelConfig, scu, func(targets []*Item) {
 		var result []string
 		for _, t := range targets {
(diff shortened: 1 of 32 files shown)


cmd/otel-allocator/internal/target/discovery_test.go
```diff
@@ -772,7 +772,7 @@ func TestReloadAppliesRelabelFilteringWhenEnabled(t *testing.T) {
 // disables filtering during discovery: targets are kept even if relabel_configs would drop them.
 func TestReloadSkipsRelabelFilteringWhenDisabled(t *testing.T) {
 	var got []*Item
-	d := newTestDiscoverer(t, "", func(targets []*Item) { got = targets })
```
iypetrov, line 775: Does it makes sense to keep a scenario for `""` as well as for `"none"` to be sure that the operator is backwards compatible. Probably can add a comment that says that `"none"` is preferred option.

iypetrov (commented):
This change looks good to me and it is a nice enhancement.

cmd/otel-allocator/internal/target/discovery_test.go
swiatekm: Honestly, what I should do is properly normalize this at config loading. Right now, this field accepts any string and only does anything if the value is `relabel-config`. The CRD rejects other values, but the raw target allocator config doesn't, and I can easily fix this here. Thanks for pointing it out!

iypetrov: I don't know if it makes sense to you, but I think if we make `""` enum value and mark it as deprecated would be cleaner. Later on we could remove it as an option. wdyt

swiatekm: It effectively is after my most recent commit. I'm not sure it's actually worth it to drop the old value, though.

TylerHelmuth (approved):

PR #5484, opened 2026-08-17
chore: add revision tagging for auto-instrumentation images

TylerHelmuth:
**Description:**
Adds a new tag to all our published auto-instrumenation image that is `<sdk-version>-<revision>`. This allows us to change the image without having to change the referenced sdk number. Revisioning follows these rules, which are documented in a new README for the `autoinstrumentation` folder:

1. if the sdk version ever changes, `revision` is set to `1`.
2. if anything about the image changes unrelated to the sdk version, `revision` is incremented.

The original tags are left in place so that existing pins are honored. 

**Testing:**

Local testing of the script 

Added a new readme to further document our auto-instrumentation images and the new tagging.


jaronoff97 (approved):

.github/workflows/scripts/check-autoinstrumentation-revision.sh
```diff
@@ -0,0 +1,99 @@
...
+
+source_file() {
+  case "$1" in
+    python) echo "autoinstrumentation/python/requirements.txt" ;;
+    nodejs) echo "autoinstrumentation/nodejs/package.json" ;;
+    *) echo "autoinstrumentation/$1/version.txt" ;;
+  esac
+}
+
+read_sdk_version() {
```
swiatekm: This is just an observation, but if we're adding a canonical way to get the sdk version from an instrumentation directory, then we can also use it further automate releases. That's an argument for this logic to live in a script elsewhere, maybe ./hack?

swiatekm (commented):
Can we automate bumping the revision number? As is, every renovate PR changing the busybox image will fail until we manually bump the revision.

.github/workflows/scripts/check-autoinstrumentation-revision.sh
TylerHelmuth: Ya I can move it to hack

iblancasa (approved):

swiatekm (approved):

.github/workflows/bump-autoinstrumentation-revision.yaml
```diff
@@ -0,0 +1,96 @@
...
+    steps:
+    - name: Generate otelbot token
+      uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+      id: otelbot-token
+      with:
+        client-id: ${{ vars.OTELBOT_OPERATOR_CLIENT_ID }}
+        private-key: ${{ secrets.OTELBOT_OPERATOR_PRIVATE_KEY }}
+        permission-contents: write
+
+    - name: Check out the pull request branch
```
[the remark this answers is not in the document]
TylerHelmuth: The purpose of this action is to update PR branches that originate from this repo from renovate

Issue #5183, opened 2026-06-03
Allow allocation strategies to be configured


swiatekm:
### Component(s)

target-allocator

### Is your feature request related to a problem? Please describe.

Allocation strategies employed by the target allocator don't permit any configuration. The only exception is the `per-node` strategy, which is configured via a feature gate - a state far from ideal. We have a fair amount of issues which come down to wanting to configure the strategies in some specific way: 

- #3128
- #4923
- #5156
- #3759
- #3892
- #3530
- #4133

We should allow strategies to be configured via the Collector and TargetAllocator CRDs. The specifics will be up to the individual issues, but as a starting point, the fallback strategy feature flag should become a CRD field.

### Describe the solution you'd like

In the target allocator config, we can do the following:

```yaml
allocation_strategy: per_node
allocation_strategy_config:
  per_node:
    fallback_strategy: consistent_hashing
```

And in the CRD:

```yaml
allocationStrategy: per-node
allocationStrategyConfg:
  perNode:
    fallBackStrategy: consistent-hashing
```

The reason I'd like to have a separate section per strategy is that they will accept different configuration options, and this makes it clear what can go where. There might eventually be a common section, but right now I don't see a need.

### Describe alternatives you've considered

There isn't really an alternative, other than simply not doing it and rejecting all the enhancements that require it. The config format is up for discussion.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

iypetrov:
@swiatekm can I try working on this one?

swiatekm:
@iypetrov I'm working on it, and I'll start submitting PRs to this repo's upstream very soon. If you'd like to see my progress, I've got a few PRs open in my fork: https://github.com/swiatekm/opentelemetry-operator/pull/151. I'm currently finalizing how partitioning would work, and I might end up needing an RFC before we're done, since it's a pretty substantial change.

PR #5492, opened 2026-08-20
Add ability to configure allocation strategies in the target allocator

swiatekm:
**Description:**

Add an `allocation_strategy_config` section to the target allocator configuration. This section is keyed by strategy name and allows the strategy to be configured. Right now, the only configuration option is setting a fallback strategy for `per_node`:

```yaml
allocation_strategy: per_node
allocation_strategy_config:
  per_node:
    fallback_strategy:
      name: consistent-hashing
```

The current way of configuring this still works, but is marked deprecated. The intent is to do the same thing at the CRD level in a follow-up and retire the current Alpha feature flag.

This PR adds the internal wiring for configuring strategies and is pretty straightforward in its logic. There's a generic way to build strategies based on a config, and it's called in builder functions so an actual strategy constructor can simply take a strategy object as a parameter without worrying about how it was built.

- Relates https://github.com/open-telemetry/opentelemetry-operator/issues/5183

**Testing:**

Added unit tests.

**Documentation:**

Documented the new option in the README.


jaronoff97 (approved):
all looks good to me, glad to see this simplified. Removing the providers is fine with me, IMO we had more dreams for it that never came to be that now we don't need in this way.

PR #5499, opened 2026-08-20
Bump Go toolchain to 1.26.5

frzifus:
## Summary
- Bumps Go toolchain from 1.26.3 to 1.26.5 in `go.mod` and `cmd/otel-allocator/integrationtest/go.mod` (CVE-2026-39822, CVE-2026-42505)
- Adds a Renovate custom manager to automatically open PRs for future Go patch version bumps in `go.mod` files. Minor version bumps are left for manual upgrade.

### DIFF
--- a/cmd/otel-allocator/integrationtest/go.mod
+++ b/cmd/otel-allocator/integrationtest/go.mod
@@ -1,6 +1,6 @@
 module github.com/open-telemetry/opentelemetry-operator/cmd/otel-allocator/integrationtest
 
-go 1.26.3
+go 1.26.5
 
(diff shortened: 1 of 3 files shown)


swiatekm:
Why do we need to bump this? We build binaries with the latest version anyway, and CVEs should be covered by that.

frzifus:
Checking for recent comments...

swiatekm:
I don't get why we need to do this. The toolchain version in the go.mod is just the floor, and we already build with the latest patch. If we bump minors too aggressively, it will hurt anyone using us as a library or shipping their own distribution.

PR #5500, opened 2026-08-20
[chore] Add tooling to auto-update instrumentation examples

TylerHelmuth:
**Description:**

Updates all Instrumentation examples to specify a revisioned image. Additionally adds new make targets to update all the examples and to check if examples are updated. Also update the renovate-bump workflow to update all the examples when it changes an autoinstrumentation. 

The new go script is generated with AI, but I prefer it over `awk` and I like being able to unit test it. 


hack/update-example-images/main.go
```diff
@@ -0,0 +1,182 @@
...
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+)
+
+const revisionScript = "hack/autoinstrumentation-revision.sh"
+
+// langKeys are the Instrumentation CR spec field names that carry an image.
+var langKeys = []string{"java", "nodejs", "python", "dotnet", "go", "apacheHttpd", "nginx"}
```
swiatekm: Can't we get this from the struct directly, rather than hardcode it? Or from the CRD manifest? Once we add ruby, php, and other languages, it'd be nice to have this work automatically.

hack/update-example-images/main.go
```diff
@@ -0,0 +1,182 @@
...
+	n := 0
+	for n < len(s) && s[n] == ' ' {
+		n++
+	}
+	return n
+}
+
+// resolveRefs asks the canonical revision script for each language's image so
+// there is a single source of truth for the pinned versions.
+func resolveRefs() (map[string]string, error) {
```
swiatekm: To me this indicates that the revision logic should be written in Go and possibly part of this program, too.

hack/update-example-images/main.go
```diff
@@ -0,0 +1,182 @@
...
+// direct-child image line in the block is replaced. Everything else is
+// preserved verbatim, including comments, ordering, and trailing-newline state.
+func pinImages(content string, refs map[string]string) string {
+	lines := strings.Split(content, "\n")
+	out := make([]string, 0, len(lines)+len(langKeys))
+
+	inBlock := false
+	blockIndent, childIndent := 0, 0
+
+	for _, line := range lines {
```
swiatekm: Correct me if I'm wrong, but this looks like it's significantly more complex because it wants to handle the case where the image field doesn't exist. We could instead only replace existing declarations in the program and add missing ones manually. In principle this might cause us to skip some missing ones, but after we start requiring the image, those will immediately be found via tests. What do you think?

hack/update-example-images/main.go
```diff
@@ -0,0 +1,182 @@
...
+			return nil, err
+		}
+	}
+
+	for _, f := range []string{
+		"tests/e2e-openshift/must-gather/install-instrumentation.yaml",
+		"config/samples/instrumentation_v1alpha1_instrumentation.yaml",
+		"docs/auto-instrumentation/README.md",
+		"docs/auto-instrumentation/languages/apache-httpd.md",
+		"docs/auto-instrumentation/languages/nginx.md",
```
swiatekm: Why not just `docs/auto-instrumentation/*`?

hack/update-example-images/main.go
TylerHelmuth: I can switch it to go, I'll do that in a separate PR

TylerHelmuth: Here is the refactor: https://github.com/open-telemetry/opentelemetry-operator/pull/5504. I like this a lot better, it'll make adding this extra automation a lot nicer.

hack/update-example-images/main.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+	}
+	if len(keys) == 0 {
+		return nil, errors.New("no language fields with an Image found on InstrumentationSpec")
+	}
+	slices.Sort(keys)
+	return keys, nil
+}
+
+func jsonName(tag string) string {
```
swiatekm: It would help to have docstrings for these. At a glance, I'm not sure what this is supposed to do.

swiatekm (approved):
LGTM. I'd like to see docstrings for the added functions, because for some of them it's not really obvious at a glance what they do and why, but I'm not going to block the PR over that.

hack/update-example-images/main.go
TylerHelmuth: I simplified the implementation a bit. The `f.Tag.Get("json")` logic is using reflection to find the tag of the struct that contains an `Image` field. This lets the script find all the languages auto-instrumentation supports without having to keep a hard-coded list up-to-date. If we add a new language to the struct the script will start trying to keep its images up-to-date without needed changes.

swiatekm (approved):

jaronoff97 (approved):

jaronoff97:
@TylerHelmuth appears the new unit tests are failing

PR #5526, opened 2026-08-25
[instrumentation]: make image required on v1beta1

TylerHelmuth:
**Description:** 
Make `Image` required on the new v1beta1 instrumentation struct.

I looks like there are not e2e tests for v1beta1 yet but once there are, we need to add some for this.


TylerHelmuth:
Do I need a changelog for this?

swiatekm:
> Do I need a changelog for this?

We haven't released anything, so no. It would just confuse users.

But your PR definitely needs to change the conversion logic as well.

swiatekm:
Does this change force the user to set *all* of the images? I don't think we want that, it sounds very annoying. If it's not set, so we want to treat it as disabled? I'm not sure this was discussed during the RFC process. @open-telemetry/operator-approvers

TylerHelmuth:
@swiatekm not intentionally. A language section in v1beta1 is allowed to be nil, so the whole block doesn't need to be set. If it is set, Image is required. 

I used the fact that the operator would insert a default image for any unspecified language in the v1alpha1 spec to set an image for the v1beta1 upgrade. That will result in every language being set in a v1alpha1 -> v1beta1 upgrade, but considering that everything was set for v1alpha1 I think that is ok. Any creating a fresh v1beta1 Instrumentation object does not need to set every language.

swiatekm:
> @swiatekm not intentionally. A language section in v1beta1 is allowed to be nil, so the whole block doesn't need to be set. If it is set, Image is required.
> 
> I used the fact that the operator would insert a default image for any unspecified language in the v1alpha1 spec to set an image for the v1beta1 upgrade. That will result in every language being set in a v1alpha1 -> v1beta1 upgrade, but considering that everything was set for v1alpha1 I think that is ok. Any creating a fresh v1beta1 Instrumentation object does not need to set every language.

My bad, I misread the diff to as doing the opposite of what it's actually doing.

I think the question remains, but it doesn't need to be answered here.

swiatekm (approved):

internal/webhook/instrumentation_conversion.go
```diff
@@ -126,76 +126,216 @@ func tov1beta1Instrumentation(in v1alpha1.Instrumentation) v1beta1.Instrumentati
...
+			Nginx:                        toV1beta1Nginx(c.Spec.Nginx),
 			ImagePullPolicy:              c.Spec.ImagePullPolicy,
 			InitContainerSecurityContext: c.Spec.InitContainerSecurityContext,
 		},
 	}
 }
 
+// v1beta1 requires an image whenever a language block is set, so a language is
+// only carried over when the v1alpha1 source pins one.
+func toV1beta1Java(in v1alpha1.Java) *v1beta1.Java {
```
jaronoff97, line 147: are we able to make the caller error here? or is that not an option for the conversion? Changing the typing to nilable implies that an image need not be set which isn't what the type should be doing IMO

TylerHelmuth: Since these are in the conversion I dont think we want to error. That'll create a strange upgrade situation where the user has dont nothing wrong but got an error. Any freshly created v1beta1 Instrumentation Object will error if image isn't set since it is required on `CommonLanguageSpec`.

Also what switched to nilable is the Language Block, not the Image. In the new Instrumentation structs all the languages reuse the same `CommonLanguageSpec` struct and that struct requires `Image` to be set. But now all the individual Language fields on `InstrumentationSpec` are nilable, which means the user doesn't have to supply it. So when adding a fresh v1beta1 Instrumentation object a user is free to define Java but not define Python. When defining Java, they must supply a Image.

For conversions from v1alpha1 to v1beta1, since the v1alpha1 logic results in a Instrumentation object that has every language set with a default image, the conversion takes all those set languages fields and sets them on the v1beta1 object.

jaronoff97 (approved):

pavolloffay (approved):

PR #5559, opened 2026-09-02
fix: reconcile operator NetworkPolicy when API server IPs change

### LINKED ISSUE #5551: Network Policy for TargetAllocator (and possibly Operator) Will Not Work In Real Life Scenarios (with Dynamic IPs)
### Component(s)

opentelemetry operator, specifically, target allocator

## Description

When using NetworkPolicy enabled via OpentelemetryCollectorSpec for TargetAllocator we observe that NetworkPolicy being generated
1. uses only a list of EndpointSlices for kubernetes API fetched only once on startup of operator
2. does not incorporate knowledge of dynamic ip addresses of Services/Endpoints
3. invalid networkpolicy is not reconciled when networkPolicy.enabled=false is set for OpentelemetryCollector, which effectively requires an operator to do manual cleanup

This effectively means that TA with network policies is unusable for many scenarios, whenever kubernetes.svc.default is a ClusterIP service that is translated to backing addresses, and when EndpointSlices can point to IP addresses that can change during lifetime.

See here - it is not allowed do add even a preconfigured service IP address (which is common for many k8s distros, as for example the first IP from the Service Network CIDR)
There is now watch subscription and no tracking of the actual service IP.

https://github.com/open-telemetry/opentelemetry-operator/blob/main/cmd/operator/operator.go#L296
https://github.com/open-telemetry/opentelemetry-operator/blob/main/cmd/operator/operator.go#L308
## Steps to Reproduce
example: Install 0.158.0 onto EKS, enable networkPolicy.enabled=true in OpentelemetryCollectorSpec and observe, how the networkpolicy for targetallocator misses:
1. Service IP of kubernetes.svc.default
2. Contains hardcoded list of EndpointSlices that may change during EKS lifetime

This leads in fact to the situation when targetallocator cannot start and ends in a crashloop.
## Expected Result
Policy contains valid addresses, including endpoints and service IP address, and it updates dynamically whenever Service or its EndpointSlices are updated.

### Kubernetes Version

1.34

### Operator version

0.158.0

### Collector version

does not matter

## Environment
EKS

### Log output

```shell

```

### Additional context

_No response_

### Tip

### LINKED ISSUE #5558: Operator NetworkPolicy does not reconcile ipBlock entries after API server IP changes
## Description

When the `operator.networkpolicy` feature gate is enabled, the operator creates a `NetworkPolicy` with egress rules that whitelist API server IPs (as `/32` ipBlock CIDRs on port 6443). These IPs are discovered once at startup from the `kubernetes` EndpointSlice in the `default` namespace and are never updated.

If a control plane node is replaced (e.g. via `ControlPlaneMachineSet` on OpenShift, or any node replacement that changes API server endpoint IPs), the `NetworkPolicy` retains the **stale IPs**. The operator pod can no longer reach the Kubernetes API server and enters `CrashLoopBackOff`.

Even restarting the operator does not fix the problem, because `Start()` ignores `AlreadyExists` errors — the stale policy persists.

## Environment

- OpenShift 4.21.24
- OpenTelemetry Operator 0.152.0-1
- Feature gate: `operator.networkpolicy` enabled

## Steps to Reproduce

1. Deploy an OpenShift cluster (or any Kubernetes cluster with the operator network policy feature gate enabled).
2. Install the OpenTelemetry Operator with `--feature-gates=+operator.networkpolicy`.
3. Verify the operator is running and the `opentelemetry-operator` NetworkPolicy exists in the operator namespace with the current control plane IPs.
4. Replace a control plane node (e.g. delete a control plane Machine and let the MachineSet replace it with a new node/IP).
5. Observe that the `NetworkPolicy` still contains the old IP, and the operator pod enters `CrashLoopBackOff` because it cannot reach the API server.

## Root Cause

In [`internal/operatornetworkpolicy/operatornetworkpolicy.go`](https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/operatornetworkpolicy/operatornetworkpolicy.go), the `Start()` method:

1. Builds the `NetworkPolicy` with IPs discovered once at startup.
2. Calls `Create()` and silently ignores `AlreadyExists` (so stale policies survive restarts).
3. Blocks on `<-ctx.Done()` — there is no watch or reconciliation loop.

```go
_, err = n.clientset.NetworkingV1().NetworkPolicies(n.operatorNamespace).Create(ctx, np, metav1.CreateOptions{})
if err != nil && !apierrors.IsAlreadyExists(err) {
    return err
}
<-ctx.Done()
return nil
```

## Current Workaround

1. Delete the stale NetworkPolicy: `kubectl delete networkpolicy opentelemetry-operator -n <operator-namespace>`
2. Restart the operator pod — it recreates the NetworkPolicy with current IPs.

## Expected Behavior

The operator should watch for changes to the `kubernetes` service EndpointSlice and automatically update the `NetworkPolicy` ipBlock entries when API server IPs change.

## Additional Context

The same stale-IP pattern also affects the Target Allocator NetworkPolicy (in `internal/manifests/targetallocator/networkpolicy.go`), though that follows a different reconciliation path via the CR reconcile loop and would be a separate fix.

atpatil-sketch:
## Description

Fixes #5558
Fixes https://github.com/open-telemetry/opentelemetry-operator/issues/5551 

The operator `NetworkPolicy` (created when the `operator.networkpolicy` feature gate is enabled) was built once at startup with hardcoded API server IPs and never updated. When control plane nodes are replaced and their IPs change, the policy becomes stale — the operator pod can no longer reach the API server and enters `CrashLoopBackOff`.

This PR adds an EndpointSlice informer inside the existing `manager.Runnable` so that the `NetworkPolicy` is automatically reconciled whenever the `kubernetes` service endpoints change.

### `internal/operatornetworkpolicy/operatornetworkpolicy.go`

- **`buildNetworkPolicy()`** — Extracted the NetworkPolicy spec-building logic into a reusable helper method.
- **`createOrUpdateNetworkPolicy()`** — Replaces the one-shot `Create` + silent `AlreadyExists` ignore. Now does `Create`, and if the policy already exists, does `Get` + `Update` to overwrite stale entries.
- **`watchEndpointSlices()`** — Sets up a `SharedInformerFactory` watching EndpointSlices in the `default` namespace filtered by label `kubernetes.io/service-name=kubernetes`.
- **`handleEndpointSliceEvent()`** — On any Add/Update/Delete event, re-discovers API server IPs via `discoverAPIServerIPs()`, compares with the current set using order-independent comparison, and updates the NetworkPolicy only if IPs have actually changed.
- **`ipsEqual()`** — Order-independent IP slice comparison to avoid spurious updates.
- **`WithLogger()` option** — Adds structured logging for IP change events and errors.
- **Refactored `Start()`** — Now: get owner ref → create/update policy → start EndpointSlice watcher (blocks until `ctx.Done()`).

### `internal/controllers/opentelemetrycollector_controller.go`

- Added `watch` verb to the EndpointSlice RBAC marker (`get;list` → `get;list;watch`), required for the informer.

### `cmd/operator/operator.go`

- Passes a named logger (`operator-networkpolicy`) via the new `WithLogger` option.

## Testing

All 13 tests pass, covering:

| Test | What it verifies |
|------|-----------------|
| `TestBuild_IPBlockPeersOnly` | Build with IP-only egress peers |
| `TestBuild_SelectorPeersOnly` | Build with pod/namespace selector peers |
| `TestBuild_CombinedIPBlockAndSelectors` | Build with both IP and selector peers |
| `TestBuild_WithIngressPorts` | Build with webhook + metrics ingress ports |
| `TestBuild_FullOpenShiftConfig` | Full OpenShift config (IPs + selectors + ports) |
| `TestNeedLeaderElection` | Leader election requirement |
| `TestCreateOrUpdate_CreatesWhenNotExist` | Create path for new NetworkPolicy |
| `TestCreateOrUpdate_UpdatesExistingPolicy` | Update path overwrites stale IPs (`10.0.0.99` → `10.0.0.1, 10.0.0.2`) |
| `TestIpsEqual` | 6 sub-tests: same order, different order, different IPs, different lengths, both empty, one empty |
| `TestHandleEndpointSliceEvent_UpdatesOnIPChange` | Full reconciliation flow: old IPs → new IPs from EndpointSlice |
| `TestHandleEndpointSliceEvent_NoUpdateWhenIPsSame` | No spurious update (ResourceVersion unchanged) |
| `TestStart_CreatesAndWatches` | Integration: Start creates policy and runs informer |
| `TestWithLogger` | Logger option wiring |

## Design Decisions

- **Informer over polling**: Event-driven via client-go informer — instant, efficient, and idiomatic Kubernetes.
- **Kept as a Runnable**: The operator network policy is fundamentally different from operand policies (not CR-driven). Converting to a full controller-runtime reconciler would be over-engineering for this use case.
- **Order-independent IP comparison**: Sorted before comparing to avoid false positives from EndpointSlice ordering differences.

## How to test manually

1. Enable the feature gate: `--feature-gates=+operator.networkpolicy`
2. Deploy the operator and verify the `opentelemetry-operator` NetworkPolicy exists with correct API server IPs
3. Simulate a control plane IP change (e.g. scale/replace a control plane node)
4. Verify the NetworkPolicy is automatically updated with the new IPs (check `kubectl get networkpolicy opentelemetry-operator -o yaml`)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/operatornetworkpolicy/operatornetworkpolicy.go
+++ b/internal/operatornetworkpolicy/operatornetworkpolicy.go
@@ -12,7 +12,6 @@ import (
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	discoveryv1 "k8s.io/api/discovery/v1"
 	networkingv1 "k8s.io/api/networking/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -194,10 +193,14 @@ func (n *networkPolicy) handleEndpointSliceEvent(ctx context.Context, ownerRef [
 	}
 
 	n.logger.Info("API server IPs changed, updating NetworkPolicy", "oldIPs", n.apiServerIPs, "newIPs", newIPs)
+
+	oldIPs := n.apiServerIPs
 	n.apiServerIPs = newIPs
 
 	if err := n.createOrUpdateNetworkPolicy(ctx, ownerRef); err != nil {
 		n.logger.Error(err, "Failed to update NetworkPolicy after IP change")
+		// Revert in-memory state so the next event retries the update.
+		n.apiServerIPs = oldIPs
 	}
 }
 
@@ -385,19 +388,6 @@ func (n *networkPolicy) operatorDeployment(ctx context.Context) (*appsv1.Deploym
 		return nil, fmt.Errorf("failed to get operator Deployment %q: %w", depRef.Name, err)
 	}
 	return dep, nil
-}
-
-// extractIPsFromEndpointSlice extracts endpoint addresses from an EndpointSlice.
-func extractIPsFromEndpointSlice(obj interface{}) []string {
-	es, ok := obj.(*discoveryv1.EndpointSlice)
-	if !ok {
-		return nil
-	}
-	var ips []string
-	for _, endpoint := range es.Endpoints {
-		ips = append(ips, endpoint.Addresses...)
-	}
-	return ips
 }
 
 func (*networkPolicy) NeedLeaderElection() bool {
--- a/internal/operatornetworkpolicy/operatornetworkpolicy.go
+++ b/internal/operatornetworkpolicy/operatornetworkpolicy.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"fmt"
 	"slices"
-	"sort"
 
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
@@ -160,13 +159,13 @@ func (n *networkPolicy) watchEndpointSlices(ctx context.Context, ownerRef []meta
 	informer := factory.Discovery().V1().EndpointSlices().Informer()
 
 	handler := cache.ResourceEventHandlerFuncs{
-		AddFunc: func(obj interface{}) {
+		AddFunc: func(_ any) {
 			n.handleEndpointSliceEvent(ctx, ownerRef)
 		},
-		UpdateFunc: func(_, _ interface{}) {
+		UpdateFunc: func(_, _ any) {
 			n.handleEndpointSliceEvent(ctx, ownerRef)
 		},
-		DeleteFunc: func(obj interface{}) {
+		DeleteFunc: func(_ any) {
 			n.handleEndpointSliceEvent(ctx, ownerRef)
 		},
 	}
@@ -219,7 +218,7 @@ func (n *networkPolicy) discoverAPIServerIPs(ctx context.Context) ([]string, err
 			ips = append(ips, endpoint.Addresses...)
 		}
 	}
-	sort.Strings(ips)
+	slices.Sort(ips)
 	return ips, nil
 }
 
@@ -227,11 +226,11 @@ func (n *networkPolicy) discoverAPIServerIPs(ctx context.Context) ([]string, err
 func ipsEqual(a, b []string) bool {
 	sortedA := make([]string, len(a))
 	copy(sortedA, a)
-	sort.Strings(sortedA)
+	slices.Sort(sortedA)
 
 	sortedB := make([]string, len(b))
 	copy(sortedB, b)
-	sort.Strings(sortedB)
+	slices.Sort(sortedB)
 
 	return slices.Equal(sortedA, sortedB)
 }
--- a/internal/operatornetworkpolicy/operatornetworkpolicy.go
+++ b/internal/operatornetworkpolicy/operatornetworkpolicy.go
@@ -6,22 +6,34 @@ package operatornetworkpolicy
 import (
 	"context"
 	"fmt"
+	"slices"
+	"sort"
 
+	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	discoveryv1 "k8s.io/api/discovery/v1"
 	networkingv1 "k8s.io/api/networking/v1"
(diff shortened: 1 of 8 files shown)


pavolloffay:
@atpatil-sketch the PR needs to be rebased. I can also take over if you would like.

internal/operatornetworkpolicy/operatornetworkpolicy.go
```diff
@@ -111,19 +131,140 @@ func WithAPISererNamespaceLabelSelector(selector *metav1.LabelSelector) Option {
...
+
+	informer.Run(ctx.Done())
+	return nil
+}
+
+// handleEndpointSliceEvent is called on any EndpointSlice event. It re-reads all matching
+// EndpointSlices, extracts the current API server IPs, and updates the NetworkPolicy if changed.
+func (n *networkPolicy) handleEndpointSliceEvent(ctx context.Context, ownerRef []metav1.OwnerReference) {
+	newIPs, err := n.discoverAPIServerIPs(ctx)
+	if err != nil {
```
frzifus, line 185: nit, but just to be safe, we may want to check `newIPs` too.
```suggestion
	if err != nil || len(newIPs) == 0 {
```

frzifus (commented):
I wonder if it can lead to issues when upgrading your cluster to quickly and the Operator is not running in HA.

There might be a little chance to lockout the operator itself.
Even if that chance might be low, we should should at least document that on the release notes. wdyt?

swiatekm:
There's a similar problem in https://github.com/open-telemetry/opentelemetry-operator/issues/5551, so we might need to set this informer up at the top level in main.go and pass it in to whichever components need it to register their callbacks. At the very least, we'd need to have it hooked into the reconciliation loops as well.

swiatekm:
See https://github.com/open-telemetry/opentelemetry-operator/pull/5625 for an implementation of API server endpoint tracking that should be easily usable for the operator NetworkPolicy too.

PR #5574, opened 2026-09-10
fix: only require permissions when the restart capability is enabled

### LINKED ISSUE #5573: Cannot start OpAMP bridge: missing patch permission for apps/deployments: access denied
### Component(s)

opamp

## Description

0.157.0 introduced a check for permissions and it refuses to start without patch permission for apps/deployments.

I believe this requirement was added by https://github.com/open-telemetry/opentelemetry-operator/pull/5306

## Steps to Reproduce

1. Use a ClusterRole without patch permission for apps/deployments
2. Deploy the bridge version 0.157.0+ with or without AcceptsRestartCommand enabled
3. The bridge will fail to start

## Expected Result

The bridge should start if AcceptsRestartCommand is not enabled

## Actual Result

The bridge requires the permission and fails to start

### Kubernetes Version

x

### Operator version

x

### Collector version

x

## Environment
OS: (e.g., "Ubuntu 20.04")

### Log output

```shell
Cannot start OpAMP bridge: missing patch permission for apps/deployments: access denied
```

### Additional context

_No response_

### Tip

jaronoff97:
**Description:**

0.157.0 started requiring patch permissions on deployments/daemonsets/statefulsets for the bridge to start up, even if the bridge was never going to restart anything. This makes those permissions conditional:

* patch on workloads is only required when the `AcceptsRestartCommand` capability is set
* if the capability isn't set and the server sends a restart command anyway, the bridge rejects it instead of trying to patch

Also moved the `Permission` type and the operator mode permission list onto the bridge config so we don't need a closure in main.

- Resolves: #5573

**Testing:**

* unit test for the operator mode permission list with and without the capability
* unit test for rejecting the restart command when the capability is off
* existing restart tests updated to enable the capability

**Documentation:**

Changelog entry added. The `AcceptsRestartCommand` capability isn't a default and there's no existing RBAC doc for the bridge, so no docs changes.


swiatekm:
I'm not against this, but is this distinction really worth making? We require CRUD on the collector resource by default, so patch doesn't meaningfully widen the surface. I'm not sure if making this conditional isn't going to make this more difficult for users to understand in the long term.

jaronoff97:
yes, this CRUD on all deployments which is a wider surface area and was the purpose of the original issue. This triggered some security flags. Also most bridge deployments do not need these permissions so we should not be asking for them.

swiatekm:
> yes, this CRUD on all deployments which is a wider surface area and was the purpose of the original issue. This triggered some security flags. Also most bridge deployments do not need these permissions so we should not be asking for them.

What security flags? Unless I'm misunderstanding something, if you already have CRUD, then adding patch doesn't change the effective permission level at all.

jaronoff97:
the problem arose from the bridge itself having permissions to patch a deployment which it only needs to restart a collector deployment. however, it requires the person installing the bridge to give it that permission which is not something all users want to do. This pr is to make it so that we don't require that if we don't need that.

jaronoff97:
really the point of this change is to ensure we don't fail to startup if the bridge doesn't have these permissions which it will never need if this capability is not set. We shouldn't be requiring permissions for a feature that will never be used.

swiatekm:
Allright, but in that case we should have documentation explaining this.

swiatekm (approved):

PR #5575, opened 2026-09-10
[autoinstrumentation]: add automation for language changelogs

TylerHelmuth:
**Description:**
As part of maturing the auto-instrumentation images we decided to have some per-language changelogs so that it is easier to figure out what is changing with each new tag. This PR adds automation the to existing revision bump logic to also generate a changelog entry.


autoinstrumentation/php/CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+# Changelog — PHP auto-instrumentation image
+
```
swiatekm: Can we exclude PHP? It looks like this image will be maintained by the PHP SIG outside this repository, much like the Go one.

hack/autoinstrumentation-revision/revision/revision.go
```diff
@@ -121,6 +121,10 @@ func revisionFile(lang string) string {
 	return filepath.Join(autoinstrumentationDir, lang, "revision.txt")
 }
 
+func changelogFile(lang string) string {
```
swiatekm:
```suggestion
func changelogFilePath(lang string) string {
```

hack/autoinstrumentation-revision/revision/changelog.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	return fmt.Sprintf("https://github.com/%s/releases/tag/v%s", repo, sdkVersion)
+}
+
+// changelogChange returns the published tag, changelog note, and upstream release
+// URL for a language's change against the base commit, or ok=false when there is
+// nothing to log. The release URL is set only for SDK bumps. The tag uses the
+// desired revision, so it is correct whether or not Apply has run.
+func (s languageState) changelogChange() (tag, note, releaseURL string, ok bool) {
+	if s.headSDK == "" {
```
swiatekm: Is this actually possible?

hack/autoinstrumentation-revision/revision/changelog.go
```diff
@@ -0,0 +1,194 @@
...
+		return "", "", "", false
+	}
+	desired, _ := s.desiredRevision()
+	tag = s.headSDK + "-" + strconv.Itoa(desired)
+	name := languageName(s.lang)
+	switch {
+	case s.baseSDK == "":
```
swiatekm: Have you actually tested this case?

hack/autoinstrumentation-revision/revision/changelog.go
```diff
@@ -0,0 +1,194 @@
...
+func renderChangelogEntry(tag, note, prRef, releaseURL string) string {
+	line := "- " + note
+	if releaseURL != "" {
+		line += " See [release notes](" + releaseURL + ")."
+	}
+	if prRef != "" {
+		line += " (#" + prRef + ")"
+	}
+	return fmt.Sprintf("## %s\n\n%s\n", tag, line)
+}
```
swiatekm: This looks like it might be better off as a template.

hack/autoinstrumentation-revision/revision/changelog.go
```diff
@@ -0,0 +1,194 @@
...
+func prependChangelogEntry(content, entry string) string {
+	entry = strings.TrimRight(entry, "\n") + "\n\n"
+	lines := strings.Split(content, "\n")
+	for i, l := range lines {
+		if strings.HasPrefix(l, "## ") {
+			head := strings.TrimRight(strings.Join(lines[:i], "\n"), "\n")
+			return head + "\n\n" + entry + strings.Join(lines[i:], "\n")
+		}
+	}
+	return strings.TrimRight(content, "\n") + "\n\n" + entry
```
swiatekm, line 106: I kind of hate how this looks, but I can't think of a way to make it nicer without pulling in a full markdown parser.

TylerHelmuth: Ya 😭

hack/autoinstrumentation-revision/revision/changelog.go
TylerHelmuth: It exists as a defensive check. if `extractVersion` (which sets headSDK) encountered something like a bad entry and couldn't find what it expected then headSDK would be `""`.

hack/autoinstrumentation-revision/revision/changelog.go
```diff
@@ -9,8 +9,13 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"text/template"
 )
 
+// changelogExcluded lists languages built here whose changelog is maintained
+// outside this repository (php by the PHP SIG, like go's upstream image).
+var changelogExcluded = map[string]bool{"php": true}
```
swiatekm, line 17: Is there a reason this map doesn't contain Go as well?

TylerHelmuth: Ya we don't have a go folder in `autoinstrumentation` like we do for PHP. Assuming we just use the upstream image for PHP we can delete that folder and all their exclusion logic.

swiatekm: Allright, we should do that as part of the PHP rollout, but not in this PR. I'm fine with leaving this exclusion here.

swiatekm (approved):

iblancasa (approved):

PR #5579, opened 2026-09-11
feat(target allocator) Support periodic re-checks for CRDs after startup

### LINKED ISSUE #5577: Target Allocator silently ignores Prometheus CRs when CRDs are installed after startup
### Component(s)

target allocator

The Target Allocator checks Prometheus CRD availability (ServiceMonitor, PodMonitor, Probe, ScrapeConfig) exactly once at startup. If those CRDs are not installed at that moment, no informers are created for them. The TA continues running normally but silently ignores all CRs of those types even if the CRDs and CRs are applied later. The only recovery is a pod restart.

This is a real operational problem in GitOps or ordered-rollout environments where prometheus-operator CRDs are applied after the TA is already running.

Add a periodic re-check (aligned with the existing resyncPeriod = 5min constant) that uses the already-present `discovery.DiscoveryInterface` to check whether previously-absent CRDs have appeared. When one is found, create and start its informer and wire up event handlers.

Two new fields under prometheusCR in the TA config:

prometheusCR:
```
  enabled: true

  # Enable periodic retry for CRDs absent at startup. Default: true.
  retryMissingCRDs: true

  # Scope which absent CRDs are retried. Empty = retry all absent ones.
  # Valid values: servicemonitors, podmonitors, probes, scrapeconfigs
  waitForCRDs: []
```

Behaviour:
- retryMissingCRDs: true, waitForCRDs: [] (default) — retry all absent CRDs every resync period. Zero config change needed for existing deployments.
- retryMissingCRDs: true, waitForCRDs: [servicemonitors, podmonitors] - only retry those two; treat probes/scrapeconfigs as intentionally absent.
- retryMissingCRDs: false (default) - disable retry entirely, preserve current behaviour.

- Exponential backoff at startup only: narrows the startup race window but does not handle CRDs installed minutes or hours later while the TA is running.
- Dynamic `apiextensions.k8s.io` CRD watch: reacts in seconds instead of up to 5 minutes, but requires adding customresourcedefinitions watch RBAC for the TA
- Fail fast on missing CRDs - k8s restart: simplest code change, but results in CrashLoopBackOff with long backoff intervals.

### Additional context

_No response_

### Tip

odubajDT:

- Resolves: https://github.com/open-telemetry/opentelemetry-operator/issues/5577


swiatekm:
Generally speaking, this kind of config reloading can be achieved in two ways: either the component handles it internally, or the watch is external and the component gets recreated. You've chosen the former, which has the unfortunate effect of making the component itself more complex and adding a bunch of synchronization. Could you try keeping the watcher as-is, and instead recreating it in `main.go`? New CRDs being installed happens rarely, so this doesn't need to be very performant, and the watcher only writes to a channel, so replacing it at that level is easy enough. What do you think?

I'm not sure we need this to be configurable at the CRD level. It sounds like we could have it be enabled unconditionally, minus whatever bugs we introduce in the process.

PR #5585, opened 2026-09-13
Add enableServiceLinks to collector and target allocator pod specs

### LINKED ISSUE #5569: Support enableServiceLinks in the pod spec of collector and target allocator workloads
### Component(s)

collector, target allocator

The operator does not expose enableServiceLinks, so every pod it creates inherits the Kubernetes default of true. The kubelet then injects a <SVC>_SERVICE_HOST / <SVC>_SERVICE_PORT pair (plus per-port variants) into every container of the collector and target allocator pods, for every Service active in the namespace.

The size of that environment block scales with the number of Services in the namespace, not with anything the collector itself declares. In large shared namespaces this leads to:

  - Large environment blocks in collector and target allocator containers, which slows pod start and, in namespaces with a few hundred Services, has contributed to probe timeouts and restarts for us.
  - Every collector container learning the name, cluster IP and ports of every unrelated Service in the namespace. Some deployment policies do not permit this.
  - A Service name potentially shadowing an environment variable the collector, or an operator-injected sidecar, relies on.

  There is currently no way to opt out:

  - enableServiceLinks is absent from OpenTelemetryCommonFields (apis/v1beta1/common.go) and from the v1alpha1 OpenTelemetryCollectorSpec.
  - None of the pod builders set it: internal/manifests/collector/deployment.go, statefulset.go, daemonset.go, internal/manifests/targetallocator/deployment.go.
  - Because CRD schemas prune unknown fields, setting spec.enableServiceLinks on the CR is silently discarded — no error, no warning, no effect. A user cannot discover that it is unsupported except by inspecting the resulting pod spec.

  Patching the generated Deployment/StatefulSet/DaemonSet directly is reverted on the next reconcile, so the field genuinely has to come from the CR. Checked against operator v0.152.0.

Add an optional EnableServiceLinks *bool to OpenTelemetryCommonFields in apis/v1beta1/common.go, and pass it through in the
  pod builders.

  Because TargetAllocatorSpec embeds v1beta1.OpenTelemetryCommonFields inline, a single addition to the common struct covers
  both the OpenTelemetryCollector and TargetAllocator CRDs — no separate field is needed for the target allocator.

  Suggested scope:

  1. EnableServiceLinks *bool on OpenTelemetryCommonFields (v1beta1). A pointer rather than a plain bool, so that unset means
  "leave the pod spec untouched" and existing CRs keep their current behaviour across an operator upgrade.
  2. Pass into corev1.PodSpec.EnableServiceLinks in all four builders listed above.
  3. Mirror the field on the v1alpha1 OpenTelemetryCollectorSpec and handle it in apis/v1alpha1/convert.go so v1alpha1 ↔
  v1beta1 conversion round-trips.
  4. Regenerate deepcopy and CRD manifests; add builder unit tests per workload kind.

  kind: OpenTelemetryCollector
  spec:
    mode: statefulset          # or deployment / daemonset
    enableServiceLinks: false
  ---
  kind: TargetAllocator
  spec:
    enableServiceLinks: false

_No response_

### Additional context

   
This is a straightforward pass-through of a standard `PodSpec` field, and the operator already passes through several neighbours on the same struct: `hostNetwork`, `hostPID`, `hostUsers`, `hostAliases`, `shareProcessNamespace`, `dnsPolicy` and `podDnsConfig`. As an
 optional `*bool` the change is additive and behaviour-preserving for existing users.

Prior art on disabling service links:

- kubernetes/kubernetes#121787 — community discussion on disabling service lin   environment variables by default 
- AWS EKS Best Practices Guide, Scalability / Workloads — "Disable enableServiceLinks by default"
- Kubernetes docs — Connecting Applications with Services, "Accessing the Service" via environment variables

### Tip

KR-Ravindra:
**Description:**

#### Problem
There is no way to opt out of Kubernetes service links for collector and target allocator pods: the field is pruned from the CR, so the kubelet injects `<SVC>_SERVICE_HOST` / `<SVC>_SERVICE_PORT` variables for every Service in the namespace into those containers (#5569).

#### Fix
Adds `enableServiceLinks` to `OpenTelemetryCommonFields` (v1beta1), so the field is available on both the `OpenTelemetryCollector` and `TargetAllocator` CRs, and passes it through to the pod spec of the generated Deployment, StatefulSet and DaemonSet (collector) and Deployment (target allocator).

The field is shaped like `corev1.PodSpec.EnableServiceLinks`: a `*bool` that defaults to `true` (via `+kubebuilder:default`), so existing CRs keep today's behaviour. Setting `enableServiceLinks: false` stops the kubelet from injecting `<SVC>_SERVICE_HOST` / `<SVC>_SERVICE_PORT` environment variables for every Service in the namespace into the collector and target allocator containers. Until now the field was pruned from the CR and there was no way to opt out.

The deprecated v1alpha1 `OpenTelemetryCollector` is left untouched.

- Resolves: #5569

**Testing:**

#### How tested

- New unit tests `TestDeploymentEnableServiceLinks`, `TestStatefulSetEnableServiceLinks`, `TestDaemonsetEnableServiceLinks` (collector) and `TestDeploymentEnableServiceLinks` (target allocator). Each asserts the pod spec field is unset by default and `false` when the CR sets `enableServiceLinks: false`.
- Before the change they fail to compile with `unknown field EnableServiceLinks in struct literal of type v1beta1.OpenTelemetryCommonFields`; after the change all pass.
- The `smoke-collector` e2e tests (deployment, statefulset, daemonset) now set `enableServiceLinks: false` on the CR and assert it on the reconciled workload's pod template.
- `go test ./internal/manifests/... ./internal/webhook/...` and `go vet` on the changed packages pass; `make generate manifests api-docs bundle` was run and the generated deepcopy, CRDs, bundle and `docs/api` are included.

**Documentation:** `docs/api/opentelemetrycollectors.md` and `docs/api/targetallocators.md` regenerated; changelog entry added in `.chloggen/enable-service-links.yaml`.

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.


KR-Ravindra:
Round 1 self-review.

Verified against `main`:
1. Neither `OpenTelemetryCommonFields` (v1beta1) nor the v1alpha1 collector spec carries `enableServiceLinks`, and none of the four builders (collector deployment/statefulset/daemonset, target allocator deployment) set it, so the field is pruned today as #5569 describes.
2. The field is `*bool` with `+kubebuilder:default:=true` and the core/v1 doc comment, which is the Pod-identical shape swiatekm asked for on the issue; the pointer comes from `corev1.PodSpec`, not from this change. `TargetAllocatorSpec` inlines the common fields, so the standalone TA CR gets it too; the collector-embedded `spec.targetAllocator` is unchanged, consistent with `hostNetwork` / `shareProcessNamespace`.
3. Conversion (`internal/webhook/collector_conversion.go`) round-trips the field in both directions; the five new tests fail to compile without the change and assert nil-by-default / false-when-set at the builder level.
4. `.chloggen/enable-service-links.yaml` matches the template; generated deepcopy, CRDs, bundles and API docs are consistent. CI: unit tests, lint, security, changelog, scorecard, govulncheck all green; e2e still running.

No duplicate: #5569 cross-references only this PR. Marking ready.

apis/v1beta1/common.go
```diff
@@ -206,6 +206,12 @@ type OpenTelemetryCommonFields struct {
 	// ShareProcessNamespace indicates if the pod's containers should share process namespace.
 	// +optional
 	ShareProcessNamespace bool `json:"shareProcessNamespace,omitempty"`
+	// EnableServiceLinks indicates whether information about services should be injected into pod's
+	// environment variables, matching the syntax of Docker links.
```
swiatekm: Let's just state this sets `enableServiceLinks` on the Pod, rather than restate K8s documentation.

swiatekm (commented):
Looks good to me with a few small asks:

- Let's not modify v1alpha1 at all, it's on its way to removal.
- There are e2e tests where we set all the optional fields to verify they get propagated correctly - could you add this field there?

KR-Ravindra:
@swiatekm done in 84aa122: dropped the v1alpha1 field, its conversion and generated entries; the field doc now just says it sets `enableServiceLinks` on the generated pods; the three `smoke-collector` e2e tests (deployment/statefulset/daemonset) set `enableServiceLinks: false` and assert it on the workload's pod template. Could you take another look when you have a moment?

swiatekm (approved):
LGTM, thank you for the contribution! Since we have a release PR up, I'll merge this after the release.

PR #3248, opened 2024-08-30
Rollback 3187

foolusion:
**Description:** Rollback #3187 

Fix #3242 with rollback of 3187. Please make sure this change works for clusters with custom domains before re submitting.

**Link to tracking Issue(s):** #3242 
- Resolves: #issue-number

**Testing:** This rolls back to previously working code.

**Documentation:** None

PR #3187, opened 2024-07-31
Fix TA in clusters with global proxy

pavolloffay:
Operator image: docker.io/pavolloffay/opentelemetry-operator:dev-b67d063e-1722432504 

**Description:** <Describe what has changed.>

See the changelog for more info.

On clusters with global proxy following error was happening on the collector
```
2024-07-22T21:11:58.189Z       info    [service@v0.102.1/service.go:113](mailto:service@v0.102.1/service.go:113)       Setting up own telemetry...

2024-07-22T21:11:58.189Z       info    [service@v0.102.1/telemetry.go:96](mailto:service@v0.102.1/telemetry.go:96)      Serving metrics {"address": ":8888", "level": "Normal"}

2024-07-22T21:11:58.189Z       info    [exporter@v0.102.1/exporter.go:275](mailto:exporter@v0.102.1/exporter.go:275)     Development component. May change in the future.  {"kind": "exporter", "data_type": "metrics", "name": "debug"}

2024-07-22T21:11:58.191Z       info    [service@v0.102.1/service.go:180](mailto:service@v0.102.1/service.go:180)       Starting otelcol...     {"Version": "0.102.1", "NumCPU": 12}

2024-07-22T21:11:58.191Z       info    extensions/extensions.go:34    Starting extensions...

2024-07-22T21:11:58.191Z       info        [prometheusreceiver@v0.102.0/metrics_receiver.go:279](mailto:prometheusreceiver@v0.102.0/metrics_receiver.go:279)   Starting discovery manager        {"kind": "receiver", "name": "prometheus", "data_type": "metrics"}

2024-07-22T21:11:58.192Z       info        [prometheusreceiver@v0.102.0/metrics_receiver.go:121](mailto:prometheusreceiver@v0.102.0/metrics_receiver.go:121)   Starting target allocator discovery      {"kind": "receiver", "name": "prometheus", "data_type": "metrics"}

2024-07-22T21:11:58.265Z       error        [prometheusreceiver@v0.102.0/metrics_receiver.go:154](mailto:prometheusreceiver@v0.102.0/metrics_receiver.go:154)   Failed to retrieve job list        {"kind": "receiver", "name": "prometheus", "data_type": "metrics", "error": "yaml: line 6: mapping values are not allowed in this context"}

[...]

PR #5586, opened 2026-09-13
Use .svc FQDN for TargetAllocator endpoint to work behind HTTP proxy

frzifus:
#### Description

The collector's Prometheus receiver reached the TargetAllocator using a bare service name (`http://<name>-targetallocator:80`). On clusters behind a corporate HTTP proxy, the collector and TargetAllocator inherit the cluster-wide `HTTP_PROXY`/`HTTPS_PROXY` env vars. Because the short hostname has no dots and no `.svc` suffix, it does **not** match the standard `NO_PROXY` entries (`.svc`, `.cluster.local`). Go's `net/http` evaluates the original hostname against `NO_PROXY` *before* DNS resolution, so the request is routed through the proxy. The proxy returns an HTML auth page, which the collector cannot parse as JSON (`Failed to retrieve the job list`), breaking the entire metrics scraping pipeline.

This changes the injected `target_allocator.endpoint` to the `.svc` FQDN:

```
http://<name>-targetallocator.<namespace>.svc:80
```

so it matches the `NO_PROXY=.svc` pattern and the collector reaches the TA directly.

The cluster domain (`.cluster.local`) is **intentionally omitted**. A full `.svc.cluster.local` endpoint was previously added in #3187 and reverted in #3248 because it breaks clusters that use a custom cluster domain. The `.<namespace>.svc` form is domain-agnostic: the cluster domain is appended automatically from the pod's resolv.conf search path (`ndots:5`), so it resolves on any cluster domain while still matching `NO_PROXY`.

The mTLS TargetAllocator certificate already includes the `.svc` SAN (`internal/manifests/targetallocator/certificate.go`), so the HTTPS (`:443`) endpoint stays valid without cert changes.

#### Backward compatibility / RBAC

- The TargetAllocator `Service` object name is unchanged (`naming.TAService` still returns the bare label); only the client-side endpoint string the collector dials changes.
- No RBAC changes — this is a network call to a Service, not an API-server operation.
- On upgrade, the collector configmap hash changes, triggering a one-time collector restart that picks up the new endpoint.

#### Testing

- Unit tests updated (`config_replace`, `configmap`, controller builder golden files, new `naming.TAServiceFQDN`).
- e2e assertions updated to the `.svc` FQDN; affected chainsaw tests pin their namespace so the rendered endpoint can be asserted deterministically (same approach used in #3187).

Red Hat issue tracker: [TRACING-6784](https://redhat.atlassian.net/browse/TRACING-6784)


docs/target-allocator/README.md
```diff
@@ -221,7 +221,7 @@ Behind the scenes, the OpenTelemetry Operator will convert the Collector’s con
 receivers:
   prometheus:
     target_allocator:
-      endpoint: http://collector-with-ta-targetallocator:80
+      endpoint: http://collector-with-ta-targetallocator.<namespace>.svc:80
```
pavolloffay, line 224: is this PR a breaking change for deployments that do not specify the namespace?

pavolloffay (approved):

docs/target-allocator/README.md
frzifus: nono, the old endpoint will continue to work. It has no effect if the namespace svc part was already set. Regarding the operator upgrading the path oh other deployments, it results in a configmap hash change and a collector restart.

PR #5358, opened 2026-07-23
[fix] Operator crash when Gateway API CRDs are not installed

callegarimattia:
Move `gatewayv1.Install(scheme)` from `main.go` `init()` to `operator.go`, gated on the autodetect result, so the scheme registration only happens when Gateway API CRDs are present, consistent with how Prometheus, OpenShift, and cert-manager CRDs are handled.

**Link to tracking Issue(s):** <Issue number if applicable>

- Resolves: #5357

PR #5591, opened 2026-09-14
fix(autodetect): require HTTPRoute in gateway.networking.k8s.io/v1 for Gateway API availability

### LINKED ISSUE #5571: Operator crash-loops when only some Gateway API CRDs are installed (GatewayClass present, HTTPRoute absent)
### Component(s)

collector

## Description

Searched existing bugs first: #5357 is the closest and its fix (#5358) does not cover this case, see below.

`GatewayAPIsAvailability()` in `internal/autodetect/main.go` reports Gateway API as available when the API **group** `gateway.networking.k8s.io` exists. It never checks for the kind the operator actually uses. When the group is served but `httproutes` is not, the OpenTelemetryCollector controller adds `HTTPRoute` to its owned resources, `SetupCaches` calls `IndexField(&gatewayv1.HTTPRoute{})`, informer creation fails with a `NoKindMatchError`, and the operator exits 1. The pod restarts into the same state forever.

This is the same class of bug as #5357, but #5358 does not cover it. #5358 gated scheme registration on the autodetect result, and the autodetect result is what is wrong here.

We hit this on EKS Auto Mode clusters that had `gatewayclasses.gateway.networking.k8s.io` (bundle v1.6.0, standard channel) and no other Gateway API CRD, reproduced on multiple clusters with v0.151.0 and v0.158.0. The same state arises whenever a provider, a platform team, or an in-progress bundle upgrade installs a subset of the Gateway API CRDs, which the Gateway API CRD-management guidance allows.

## Steps to Reproduce

1. Start with a cluster that has no Gateway API CRDs.
2. Install only the GatewayClass CRD from the standard bundle:

   ```sh
   curl -sSL https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.6.0/standard-install.yaml \
     | yq 'select(.kind == "CustomResourceDefinition" and .metadata.name == "gatewayclasses.gateway.networking.k8s.io")' \
     | kubectl apply -f -
   kubectl api-resources --api-group=gateway.networking.k8s.io
   # NAME             SHORTNAMES   APIVERSION                     NAMESPACED   KIND
   # gatewayclasses   gc           gateway.networking.k8s.io/v1   false        GatewayClass
   ```

3. Install the operator (Helm chart or manifests) and watch the pod.

## Expected Result

The operator starts. Gateway API HTTPRoute support is disabled, with a log line saying the `HTTPRoute` kind was not found, in the same way the Prometheus autodetect disables `ServiceMonitor`/`PodMonitor` support when those kinds are absent.

## Actual Result

```
{"level":"INFO","logger":"setup","message":"Gateway API CRDs are installed, adding to scheme."}
{"level":"INFO","logger":"setup","message":"OpenTelemetryCollectorCRDSs are available to the operator"}
{"level":"ERROR","logger":"setup","message":"unable to create controller","controller":"OpenTelemetryCollector",
 "error":"no matches for kind \"HTTPRoute\" in version \"gateway.networking.k8s.io/v1\""}
```

Pod goes to `CrashLoopBackOff`. The first line is the one added by #5358, so the fix from that PR is present and does not help.

There is no way to opt out. `gateway-apis-availability` in the config is overwritten unconditionally by `ApplyAutoDetect`, there is no feature gate for it, and the Helm chart exposes no value.

### Kubernetes Version

1.35 and 1.36 (EKS Auto Mode)

### Operator version

Reproduced on v0.151.0 and v0.158.0. The HTTPRoute feature and this probe first shipped in v0.149.0 (#4510), and the probe is unchanged on `main`, so every release from v0.149.0 onward is affected.

### Collector version

Not reached: the operator exits before reconciling any collector. Collector image pinned in the affected clusters is opentelemetry-collector-contrib 0.152.1.

## Environment
OS: EKS Auto Mode managed nodes (Bottlerocket), linux/amd64
Compiler(if manually compiled): go1.26.7. The operator binary is compiled by us from the unmodified v0.151.0 and v0.158.0 source tags (build-date 2026-08-28 in the v0.151.0 startup log); no source patches.

EKS Auto Mode, us-east-1. The affected clusters served only `gatewayclasses.gateway.networking.k8s.io`, annotated `gateway.networking.k8s.io/bundle-version: v1.6.0` and `gateway.networking.k8s.io/channel: standard`; no other Gateway API CRDs were present.

### Log output

```shell
{"level":"INFO","timestamp":"2026-09-08T04:20:08Z","logger":"setup","message":"Cert-Manager is available to the operator, adding to scheme."}
{"level":"INFO","timestamp":"2026-09-08T04:20:08Z","logger":"setup","message":"Gateway API CRDs are installed, adding to scheme."}
{"level":"INFO","timestamp":"2026-09-08T04:20:08Z","logger":"setup","message":"OpenTelemetryCollectorCRDSs are available to the operator"}
{"level":"ERROR","timestamp":"2026-09-08T04:20:08Z","logger":"setup","message":"unable to create controller","controller":"OpenTelemetryCollector","error":"no matches for kind \"HTTPRoute\" in version \"gateway.networking.k8s.io/v1\"","stacktrace":"github.com/open-telemetry/opentelemetry-operator/cmd/operator.runOperator\n\tgithub.com/open-telemetry/opentelemetry-operator/cmd/operator/operator.go:184\nmain.main\n\tgithub.com/open-telemetry/opentelemetry-operator/cmd/operator/main.go"}

$ kubectl api-resources --api-group=gateway.networking.k8s.io
NAME             SHORTNAMES   APIVERSION                     NAMESPACED   KIND
gatewayclasses   gc           gateway.networking.k8s.io/v1   false        GatewayClass
```

### Proposed fix

Check for the kind the controller binds, at the version it binds, instead of the group:

```go
func (a *autoDetect) GatewayAPIsAvailability() (gatewayapi.ApiAvailability, error) {
	resources, err := a.dcl.ServerResourcesForGroupVersion(gatewayv1.GroupVersion.String())
	if err != nil {
		if apierrors.IsNotFound(err) {
			return gatewayapi.ApiNotAvailable, nil
		}
		return gatewayapi.ApiNotAvailable, err
	}
	for _, r := range resources.APIResources {
		if r.Kind == "HTTPRoute" {
			return gatewayapi.ApiAvailable, nil
		}
	}
	return gatewayapi.ApiNotAvailable, nil
}
```

This mirrors the repo's own `PrometheusCRsAvailability` idiom, which checks for the `ServiceMonitor` and `PodMonitor` kinds rather than the `monitoring.coreos.com` group. Treating only not-found as "not available" and propagating other discovery errors also stops the probe from flapping to `ApiNotAvailable` on transient discovery failures.

### Reference in other projects

The same bug shape has been fixed with per-kind checks elsewhere:

- kumahq/kuma#10074: enabled Gateway API after finding only the Gateway CRD, failed on missing ReferenceGrant; fixed by checking every required CRD.
- envoyproxy/gateway#8991 / #9583: GKE's managed bundle lacked ListenerSet and GRPCRoute; fixed with per-kind existence probes in front of each watch.
- kubernetes-sigs/aws-load-balancer-controller#4602: added per-kind detection via `ServerResourcesForGroupVersion` and disables the feature if any required kind is missing.
- Cilium and Istio both check individual CRDs by name before enabling Gateway API support.

Happy to open a PR with the change above plus a unit test covering the GatewayClass-only case.

### Tip

alexeymoskalev-devops:
**Description:** `GatewayAPIsAvailability()` reported the Gateway API as available as soon as the `gateway.networking.k8s.io` group was served, without checking the kind the operator uses. With a partial bundle (for example only `GatewayClass`, which is what EKS Auto Mode ships), `cmd/operator` added `HTTPRoute` to the scheme and the collector controller tried to own it, so the operator crash-looped with `no matches for kind "HTTPRoute" in version "gateway.networking.k8s.io/v1"`. #5358 gated scheme registration on the autodetect result; this fixes the result itself.

The check now iterates the group's versions and requires `HTTPRoute` to be served in `gateway.networking.k8s.io/v1` — the only Gateway API kind the operator creates and watches, and the version it installs into the scheme — mirroring how `PrometheusCRsAvailability` checks for `ServiceMonitor`/`PodMonitor`. A partial bundle, or one serving `HTTPRoute` only in `v1beta1`, now disables HTTPRoute support with the existing "Gateway API CRDs are not installed, skipping adding to scheme" log line instead of crashing.

- Resolves: #5571

**Testing:**

- `TestDetectPlatformBasedOnAvailableAPIGroupsGatewayAPI` in `internal/autodetect/main_test.go`, on the same httptest discovery server as the Prometheus test: no group / group with only `GatewayClass` / `HTTPRoute` only in `v1beta1` → NotAvailable; `HTTPRoute` in `v1` → Available. The two partial-bundle cases fail on `main`. The test uses named `t.Run` subtests (each with its own `httptest.Server`), unlike the two older flat-loop tests next to it; deliberate, not an oversight.
- Reproduced on a k3s v1.33 cluster with only the `GatewayClass` CRD from the gateway-api v1.6.0 standard bundle:
  - `main`: `Gateway API CRDs are installed, adding to scheme.` → `unable to create controller ... no matches for kind "HTTPRoute"`, exit 1.
  - this branch: `Gateway API CRDs are not installed, skipping adding to scheme.` → manager and the `opentelemetrycollector` controller start.
  - this branch after also installing the `HTTPRoute` CRD: `Gateway API CRDs are installed, adding to scheme.`
- `make test` (`gotestsum -race ./...`): 2557 tests, 3 skipped, plus 18 target-allocator integration tests — pass. `go vet`, `golangci-lint run ./internal/autodetect/...` (0 issues), `make chlog-validate`, `make ensure-update-is-noop` — pass.

**Documentation:** No user-facing docs change; changelog entry `.chloggen/fix_gatewayapi-autodetect-httproute.yaml` (bug_fix, collector, #5571).


swiatekm (approved):

PR #5592, opened 2026-09-14
Fix NetworkPolicy port protocols

sugaf1204:
**Description:**

Preserve the transport protocol of collector and Target Allocator container ports when generating NetworkPolicy ingress rules.

Previously, generated NetworkPolicies always allowed TCP, even when a receiver or a custom port was configured to use UDP. This caused UDP traffic to remain blocked when operand NetworkPolicies were enabled.

This change:

- Converts container ports to NetworkPolicy ports while preserving TCP, UDP, or SCTP.
- Defaults an omitted protocol to TCP, matching Kubernetes behavior.
- Applies the shared conversion to both OpenTelemetry Collector and Target Allocator NetworkPolicies.
- Leaves the Operator NetworkPolicy unchanged because its API server, webhook, and metrics ports are inherently TCP.

N/A

**Testing:**

Added tests covering:

- A UDP receiver inferred from Collector configuration.
- Target Allocator custom TCP and UDP ports using the same port number.
- TCP defaulting when the protocol is omitted.

Executed:

- `go test ./internal/manifests/...`
- `go vet ./internal/manifests/...`
- `go test ./internal/operatornetworkpolicy`

**Documentation:**

No documentation changes are required because this fixes generated NetworkPolicy behavior without changing the API.


pavolloffay:
@sugaf1204 could you please add a changelog entry? `make chlog-new`

pavolloffay (approved):

PR #5594, opened 2026-09-14
fix(target-allocator): do not assign targets to terminating collector pods

### LINKED ISSUE #5576: Target Allocator retains targets on terminating collectors
### Component(s)

target allocator, collector

### Description

When a collector’s node is unreachable, deleting the pod sets `metadata.deletionTimestamp`, but `PodReady=True` remains and the Target Allocator continues treating the pod as eligible and assigning targets to it.

`collectorNotReadyGracePeriod` does not apply because the pod is still considered ready.

### Steps to reproduce

1. Run the Target Allocator with at least two collector replicas and:
```
collector_not_ready_grace_period: 2m
```
2. Wait for a target to be assigned.
3. Make the assigned collector’s node unreachable.
4. Delete the collector pod, or wait for Kubernetes to evict it.
5. Wait longer than the grace period and check the assignments.

### Expected

The terminating collector has its targets reassigned by the Target Allocator. 

### Actual

The terminating collector keeps its targets. Restarting the Target Allocator does not resolve this issue.

### Kubernetes Version

1.35.3

### Operator version

0.149.0

### Collector version

0.145.0

### Log output

```shell

```

### Additional context

_No response_

### Tip

alexeymoskalev-devops:
**Description:** When a collector's node becomes unreachable and the pod is deleted or evicted, the API server sets `metadata.deletionTimestamp` but the pod keeps reporting `PodReady=True`, because the kubelet that would update its status is gone. The collector watcher only filtered on phase and readiness, so the terminating pod stayed eligible and kept its targets indefinitely; `collector_not_ready_grace_period` never applied.

`runOnCollectors` now treats a pod with a `deletionTimestamp` as no longer a collector, next to the existing readiness check, so its targets are reassigned on the next reconcile. This is the single point that feeds `allocator.SetCollectors`, so every allocation strategy and the `/jobs` API pick it up.

On the watch question from the issue: the deletion goes through the API server, which sets `deletionTimestamp` and emits an update event regardless of node state; the informer store the watcher reads reflects it. Verified in the apiserver: `podStrategy.CheckGracefulDelete` only shortens the grace period for unscheduled or already-terminated pods, and `Store.updateForGracefulDeletionAndFinalizers` sets `deletionTimestamp` through `GuaranteedUpdate`, so the write and the watch event happen before any kubelet involvement. Confirmed on a cluster with a pod held in that state (below).

Behaviour note: this also applies to ordinary graceful terminations. A collector pod that is terminating during a rolling update loses its targets as soon as it is marked for deletion, rather than after `collector_not_ready_grace_period` (or never, when the grace period is 0). That seems to be the right thing for a pod on its way out (it will stop scraping and disappear either way, so moving its targets earlier shortens the gap rather than adding churn), but it is a slightly wider change than the unreachable-node case, and #3807 deliberately made target shifting on readiness opt-in via the grace period. If you would rather keep that knob in control here too, the check can be gated on `collector_not_ready_grace_period != 0` as well; I left it unconditional because the issue's case is exactly the one where readiness never changes.

- Resolves: #5576

**Testing:**

- `Test_terminatingPodIsNotACollector` in `cmd/otel-allocator/internal/collector/collector_test.go`: two Running/Ready pods, one with `deletionTimestamp` set; only the other one is a collector, for grace periods 0 and 30s. Fails on `main` (both pods reported), passes on this branch; race-clean and stable across repeated runs.
- On a k3s v1.33 cluster with the allocator running against two label-matched pods (`busybox sleep`, `terminationGracePeriodSeconds: 600`, so a deleted pod stays `Running`/`Ready=True` with `deletionTimestamp` set — the same state as a pod on an unreachable node), six static targets, `collector_not_ready_grace_period: 20s`:
  - `main`: `{col-a: 2, col-b: 4}` before deleting `col-b`; still `{col-a: 2, col-b: 4}` 12s and 32s after deletion (grace period elapsed, pod still `Ready=True`).
  - this branch: `{col-a: 2, col-b: 4}` before; `{col-a: 6}` within seconds of the deletion.
- `go test -race ./cmd/otel-allocator/...` — pass; `go vet`, `golangci-lint run ./cmd/otel-allocator/internal/collector/...` (0 issues), `make chlog-validate` — pass.

**Documentation:** No user-facing docs change; changelog entry `.chloggen/fix_targetallocator-skip-terminating-collectors.yaml` (bug_fix, target allocator, #5576).


cmd/otel-allocator/internal/collector/collector_test.go
```diff
@@ -488,6 +488,57 @@ func Test_gracePeriodWithNonReadyPodCondition(t *testing.T) {
...
+	terminating.DeletionGracePeriodSeconds = new(int64)
+
+	for _, gracePeriod := range []time.Duration{0, 30 * time.Second} {
+		t.Run(gracePeriod.String(), func(t *testing.T) {
+			synctest.Test(t, func(t *testing.T) {
+				podWatcher := getTestPodWatcher(gracePeriod)
+				var actual map[string]*allocation.Collector
+				mapMutex := sync.Mutex{}
+				for _, p := range []*v1.Pod{pod("test-pod-ready"), terminating} {
+					_, err := podWatcher.k8sClient.CoreV1().Pods(namespace).Create(context.Background(), p, metav1.CreateOptions{})
```
swiatekm:
```suggestion
					_, err := podWatcher.k8sClient.CoreV1().Pods(namespace).Create(t.Context(), p, metav1.CreateOptions{})
```

swiatekm (commented):
Looks good to me, one nitpick about context handling in the added test.

swiatekm (approved):

PR #5181, opened 2026-06-03
Revert Go upgrade in e2e test app

swiatekm:
This is currently broken after https://github.com/open-telemetry/opentelemetry-operator/pull/5173.

PR #5610, opened 2026-09-15
[Chore] Stop stripping the Go e2e test app and re-enable the instrumentation-go e2e test

IshwarKanse:
**Description:** The `instrumentation-go` e2e test has been skipped since the Go test app moved to Go 1.26 (#5173 / #5181). The `autoinstrumentation-go` agent (v0.24.0) cannot instrument the stripped `e2e-test-app-golang` binary, for two independent reasons:

1. Go 1.26 no longer stores `textStart` in the `.gopclntab` header ([runtime/symtab.go `pcHeader`](https://github.com/golang/go/blob/go1.26.0/src/runtime/symtab.go) now marks the field unused). The agent's stripped-binary lookup (`internal/pkg/process/binary/funcs_stripped.go`) reads it as 0, so on amd64 every function offset wraps and the sidecar exits with `overflow in offset to read in the text section: 18446744073707215360`. On arm64 the subtraction does not wrap and the agent silently computes wrong uprobe offsets. When the ELF symbol table is present the agent uses `FindFunctionsUnStripped` instead, which works on Go 1.26+.
2. The agent only ships cached struct offsets up to Go 1.26.2. For newer Go versions it logs `Offset not cached, analyzing directly` and reads DWARF, which `-w` strips (`decoding dwarf section info at offset 0x0: too short`).

This PR builds the test app without `-ldflags="-w -s"` so the symbol table and DWARF are kept (binary grows from ~5.6 MB to ~8.3 MB), documents why in the Dockerfile, and re-enables the `instrumentation-go` chainsaw test. This keeps working across future Go bumps by Renovate, unlike pinning the Go version (#5180), because the DWARF fallback covers versions the agent has no cached offsets for.

**Link to tracking Issue(s):** follow-up to #5181 (test was skipped "temporarily" there)

**Testing:** Ran the `instrumentation-go` scenario (collector + `Instrumentation` + `my-golang` deployment from `tests/e2e-instrumentation/instrumentation-go`) against agent `v0.24.0` with the app image rebuilt from this Dockerfile, plus a matrix of variants:

| App build | Result |
|---|---|
| current `e2e-test-app-golang:main` (Go 1.27.1, `-w -s`) | sidecar CrashLoopBackOff, text-section overflow |
| Go 1.25.14, `-w -s` | fails, DWARF fallback needed (floating `1.25-alpine` is past the cached offsets) |
| Go 1.26.1, `-w` | works (cached offsets) |
| Go 1.26.6, no ldflags | works via DWARF fallback |
| Go 1.27.1, no ldflags (this PR) | works via DWARF fallback, `GET /rolldice` server spans reach the collector |

Note for reviewers: the `e2e-instrumentation` job on this PR still pulls the previously published `e2e-test-app-golang:main` image (test-app images are only pushed on merge to `main`), so `instrumentation-go` will only go green once the image is republished from `main`. If preferred, the Dockerfile change can be merged first and the un-skip as a follow-up.

The underlying agent limitation will be reported to opentelemetry-go-instrumentation separately.

**Documentation:** None (test-only change).


IshwarKanse:
The underlying agent limitation is now tracked in open-telemetry/opentelemetry-go-instrumentation#3869.

IshwarKanse:
cc @pavolloffay

iblancasa (approved):

swiatekm (approved):

IshwarKanse:
The agent-side fix for the underlying issue is up as open-telemetry/opentelemetry-go-instrumentation#3870. This PR is still worthwhile on its own so the test app keeps working with the released agent versions.

PR #5613, opened 2026-09-16
[chore] Fix e2e-test-app-nodejs arm64 image containing amd64 binaries

IshwarKanse:
## Summary

The `nodejs` e2e test app image is built for `linux/arm64,linux/amd64,linux/s390x,linux/ppc64le` via Buildx, but its base image (`node:lts-trixie-slim`) is pinned to a single-platform **amd64-only** manifest digest instead of the multi-platform image index digest. Because a digest reference always resolves to exactly one manifest, Buildx used the amd64 image contents for every requested platform, producing an `arm64`-tagged image that actually contains amd64 binaries.

This breaks every container running the image on arm64 nodes with `exec format error` — even `/bin/sh` fails, since the base image layers themselves are the wrong architecture. All 6 Node.js auto-instrumentation e2e tests fail on arm64 clusters as a result.

The bad digest was introduced by the renovate base-image bump in #5537 (merged 2026-08-31). All other `tests/test-e2e-apps/*/Dockerfile` base images (python, java, httpd, golang) already use the correct multi-platform index digest — `nodejs` was the only outlier.

## Fix

Pin `node:lts-trixie-slim` to its current multi-platform image index digest, matching the pattern used by the other test-e2e-apps base images.

## Verification

Reproduced and verified locally on arm64 (podman):

- **Before fix**, pulling the old digest for `linux/arm64` warns: `WARNING: image platform (linux/amd64) does not match the expected platform (linux/arm64)`.
- **After fix**, `podman build --platform linux/arm64` builds natively (no emulation), and the resulting image reports `process.arch: arm64` at runtime.

## Test plan

- [x] Built `tests/test-e2e-apps/nodejs` for `linux/arm64` locally and confirmed it runs natively without `exec format error`
- [x] CI: `Publish Test E2E images` workflow rebuilds and republishes the multi-arch `e2e-test-app-nodejs:main` image on merge


jaronoff97 (approved):

swiatekm (approved):

swiatekm:
If renovate broke this, how can we prevent it from breaking it again? I'm not sure if it's aware of which digests are multiplatform.

iblancasa (approved):

IshwarKanse:
@swiatekm  Good question — I dug into the history and it turns out renovate isn't actually the root cause here.

The digest was already amd64-only when this Dockerfile switched to `node:lts-trixie-slim` in #5324 (`sha256:5301bbf5...`) — that digest resolves to a single-platform manifest, not the index. Every renovate bump since then (#5399, #5431, #5537) correctly rolled the digest forward, but stayed pinned to amd64, because renovate's docker datasource deliberately preserves the "scope" of whatever digest is currently pinned: if the current digest is a single-platform manifest, it detects that architecture and resolves the newest digest *for that same architecture* rather than promoting back to the manifest-list digest. That's documented/intentional behavior on renovate's end, not a bug — it assumes a single-arch pin was deliberate.

So renovate has been faithfully preserving a mistake that was introduced manually, most likely from copying a digest off a local `docker pull` + `docker inspect` (which resolves to your machine's platform) instead of the manifest-list digest. `node:lts-bookworm` never had this problem because it was multi-arch from the first pin.

To actually prevent this class of bug from recurring, I don't think there's a renovate.json knob for it — but we could add a small CI check that verifies each `tests/test-e2e-apps/*/Dockerfile` base-image digest resolves to a manifest index covering all the platforms declared for it in `publish-test-e2e-images.yaml`. That would catch both a bad manual pin and a future renovate hiccup at PR time. Happy to add that in a follow-up if it sounds useful — didn't want to scope-creep this fix.

swiatekm:
> @swiatekm Good question — I dug into the history and it turns out renovate isn't actually the root cause here.
> 
> The digest was already amd64-only when this Dockerfile switched to `node:lts-trixie-slim` in #5324 (`sha256:5301bbf5...`) — that digest resolves to a single-platform manifest, not the index. Every renovate bump since then (#5399, #5431, #5537) correctly rolled the digest forward, but stayed pinned to amd64, because renovate's docker datasource deliberately preserves the "scope" of whatever digest is currently pinned: if the current digest is a single-platform manifest, it detects that architecture and resolves the newest digest _for that same architecture_ rather than promoting back to the manifest-list digest. That's documented/intentional behavior on renovate's end, not a bug — it assumes a single-arch pin was deliberate.
> 
> So renovate has been faithfully preserving a mistake that was introduced manually, most likely from copying a digest off a local `docker pull` + `docker inspect` (which resolves to your machine's platform) instead of the manifest-list digest. `node:lts-bookworm` never had this problem because it was multi-arch from the first pin.
> 
> To actually prevent this class of bug from recurring, I don't think there's a renovate.json knob for it — but we could add a small CI check that verifies each `tests/test-e2e-apps/*/Dockerfile` base-image digest resolves to a manifest index covering all the platforms declared for it in `publish-test-e2e-images.yaml`. That would catch both a bad manual pin and a future renovate hiccup at PR time. Happy to add that in a follow-up if it sounds useful — didn't want to scope-creep this fix.

Honestly, I think what we should do is run some subset of e2e tests on an ARM runner on changes to image definitions. I'll open an issue for it.

PR #5615, opened 2026-09-16
Normalize collector config between the CR and ConfigMap

### LINKED ISSUE #4314: Operator misinterprets strings as numbers when converting configuration

swiatekm:
## Description

An `OpenTelemetryCollector` CR containing a string like `"0e12"` (e.g. the transform processor's `statements`) produced a ConfigMap in which that value was written unquoted. The collector's config parser reads unquoted `0e12` as a float, so the collector crashed on startup with `expected type 'string', got unconvertible type 'float64'` (#4314).

The root cause is broader than that one value. The operator serialized `spec.config` with one YAML library (`goccy/go-yaml`) while the collector parses the ConfigMap with another (`go.yaml.in/yaml/v3`, through confmap), and the two disagree about which unquoted scalars are strings - exponent notation without a decimal point (`0e12`, `1E5`), special floats (`.inf`, `.NaN`), and more. Any such disagreement silently changes a value's type between the CR and the collector. When a target allocator was enabled, the rendered YAML was additionally re-parsed with `gopkg.in/yaml.v2` (a YAML 1.1 parser) and re-encoded, adding a second opportunity for reinterpretation.

Rather than patching individual cases (a previous attempt special-cased the exponent pattern with a regex), this PR aligns YAML encoding with the collector's parser and checks every rendered document for equivalence with the structured config. The check uses documented normalization rules: numbers are compared by value, and null and empty collections are treated as equivalent. For TA-enabled collectors, verification applies to the config after the intended Prometheus receiver rewrite.

### Changes

* Encode with the collector's own YAML library. We moved off of it because it was unmaintained, that has since changed.
* Verify rendered documents against the CR JSON representation and error on differences - with well defined exceptions.
* Preserve readable integer formatting and work around a multiline-string bug.
* The target allocator rewrite no longer re-parses rendered YAML.

User-visible side effects: ConfigMaps for TA-enabled collectors now use the same 2-space indentation and top-level ordering as all others, and integral numbers within the lossless `int64` conversion range are no longer written in exponent notation. Known unsupported map keys, including `<<` and multiline keys starting with a tab, cause rendering to return an error; the multiline-string workaround applies to values, not keys.

### Link to tracking Issue(s)

- Resolves #4314
- Relates: #3988

### Testing

* Unit tests: a scalar-fidelity table asserting the collector reads back the same type and value from every position, a fuzz target over arbitrary strings
* A new chainsaw test (`tests/e2e/collector-config-scalars`) applies a CR exercising each conversion
* Existing e2e asserts for TA-enabled collectors regenerated for the unified formatting


jaronoff97 (approved):

apis/v1beta1/config_yaml.go
```diff
@@ -0,0 +1,249 @@
...
+// Equivalence is judged by what the collector's config loader observes: every scalar must have the same type
+// and value, and every collection the same members. Two differences are tolerated because they are artifacts of
+// the Go serialization on either side and invisible to the collector:
+//
+//   - Numbers are compared by value. JSON has no integer/float distinction and encoding/json decodes every
+//     number as float64, whereas the YAML decoder produces int for integer literals; the collector converts
+//     either to the field's numeric type.
+//   - null and an empty collection are interchangeable. encoding/json renders a nil map or slice as null, the
+//     YAML encoder renders it as {} or []; the collector treats a null section like an empty one.
+func (c *Config) VerifyYAMLEquivalence(doc []byte) error {
```
iblancasa: I think this is exported only so tests in `internal/otelconfig` can reach it, right?

apis/v1beta1/config_yaml.go
```diff
@@ -0,0 +1,249 @@
...
+// A multi-line string whose first line starts with a tab is turned into an explicitly double-quoted scalar to
+// work around https://github.com/yaml/go-yaml/issues/383: the encoder writes multi-line strings as literal block
+// scalars, but adds the indentation indicator the block needs only when the string starts with a space or a line
+// break; with a leading tab it emits a block that its own parser rejects ("found a tab character where an
+// indentation space is expected"). Only the first line matters: the block's indentation is fixed by it, after
+// which a tab is content.
+func renderableValue(v any) any {
+	switch val := v.(type) {
+	case float64:
+		if val == math.Trunc(val) && val >= math.MinInt64 && val < math.MaxInt64 && float64(int64(val)) == val {
```
iblancasa: I think this can be simplified to
```go
if val == math.Trunc(val) && val >= math.MinInt64 && val < math.MaxInt64 {
```

iblancasa (commented):
The PR removes the YAML round trip from `ReplaceConfig` but some callers still render with `Config.Yaml()` and immediately parse the result back with `gopkg.in/yaml.v2`:

- internal/manifests/targetallocator/configmap.go, getScrapeConfigs (around line 305) and getScrapeConfigsFromOtelConfig (around line 339)
- internal/webhook/collector_webhook.go, validateTargetAllocatorConfig (around line 334)

apis/v1beta1/config_yaml.go
swiatekm: I think so, yeah.

apis/v1beta1/config_yaml.go
swiatekm: Kind of. The boundary between the config handling in the API package and in the operator root package is a bit ambiguous. I'll see if I can move it out of here completely. Users of the api package shouldn't care about this, it's really more of a webhook implementation detail.

apis/v1beta1/config_yaml.go
swiatekm: Done.

apis/v1beta1/config_yaml.go
swiatekm: Allright, I moved it out of the module back into `internal/otelconfig`. There's now a new `RenderYaml` function internal to the operator which contains this check. Wherever we want to render yaml to write it to a ConfigMap, we use that.

iblancasa (approved):

PR #5631, opened 2026-09-21
Register the webhook_event receiver parser

IshwarKanse:
**Description:**

Register a `webhook_event` receiver parser (single port, default 8088), with the deprecated name `webhookevent` as an alias.

The receiver was renamed from `webhookevent` to `webhook_event` in open-telemetry/opentelemetry-collector-contrib#47517. Today the operator handles it through the generic fallback parser, which already derives a Service port from `endpoint`. Registering the parser adds:
- a default endpoint (`0.0.0.0:8088`) when `endpoint` is omitted. The receiver itself requires `endpoint` and has no default, so the default comes from the operator.
- the OpenShift TLS profile defaults for a `tls:` block
- resolution of the deprecated name `webhookevent`

The default port is the one open question: 8088 is also the `splunk_hec` default. If maintainers prefer, I can register the parser without a default port instead (like `tcp_log` and `udp_log`).

Part of https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45339

Similar to #5317

**Testing:**

Added `webhook_event` and `webhookevent` rows to `TestDownstreamParsers` (parser name, default port, port override, default config); they fail without the parser. `make lint` reports 0 issues and `make chlog-validate` passes.

Also verified on OpenShift 4.22 with an operator built from this branch. A collector with `webhook_event: {}` and no `endpoint` failed validation before (`missing receiver server endpoint from config`, no Service). With the parser it gets `endpoint: 0.0.0.0:8088` injected, runs, and gets a Service port 8088. The `webhookevent` name works the same way.

**Documentation:**

None needed; the operator does not document per-receiver parsers. Changelog entry added.


IshwarKanse:
cc @pavolloffay

jaronoff97 (approved):

internal/components/receivers/helpers.go
```diff
@@ -125,6 +125,12 @@ var componentParsers = []components.Parser{
 	components.NewSinglePortParserBuilder("wavefront", 2003).
 		WithTargetPort(2003).
 		MustBuild(),
+	// webhook_event, formerly webhookevent
+	// (open-telemetry/opentelemetry-collector-contrib#47517).
+	components.NewSinglePortParserBuilder("webhook_event", 8088).
```
iblancasa: I think we should use `UnsetPort` here instead of `8088`.
The receiver itself has no default port. If you leave out endpoint, the collector refuses to start. All the other parsers in this file only set a default when the receiver already has one upstream, so this would be the first case where the operator makes one up.

There is also a practical problem: `8088` is already the `splunk_hec` default. If someone defines both receivers and leaves out both endpoints, the operator would generate two Service ports with the same number and Kubernetes rejects that.

With `UnsetPort` we still get the alias for the old name and the port parsing from endpoint, same as `tcp_log` does today

IshwarKanse: This is exactly what the PR description asked reviewers about the port: "8088 is also the `splunk_hec` default. If maintainers prefer, I can register the parser without a default port instead (like `tcp_log` and `udp_log`)." So this is the direction we hoped for.

Done in b246abcf: `webhook_event` now uses `UnsetPort` like `tcp_log` and `udp_log`. The `webhookevent` alias stays and the Service port still comes from `endpoint`.

I reproduced the clash you describe. With `splunk_hec` and `webhook_event` both missing `endpoint`, the old registration produced two Service ports on 8088. Omitting `endpoint` is now an error, and a new test (`TestWebhookEventHasNoDefaultPort`) covers it. It fails with the 8088 registration.

One side effect: with an unset port the operator also skips the OpenShift TLS profile defaults for a `tls:` block, the same as `tcp_log` and `udp_log` today. That seems fine here, since the receiver already requires the user to set `endpoint`.

iblancasa: > This is exactly what the PR description asked reviewers about the port: "8088 is also the `splunk_hec` default....

🤦🏻 oh sorry

iblancasa (approved):

PR #5635, opened 2026-09-21
[chore] Fix flaky collector-config-scalars e2e test

swiatekm:
The `collector-config-scalars` e2e test, added in #5615, configures a `transform` processor. The e2e tests run the core collector distribution, which does not include it, so the collector exits on startup:

```
'processors' unknown type: "transform" for id: "transform" (valid values: [span probabilistic_sampler filter batch memory_limiter attributes resource])
```

When the test passes, it's mostly by accident.

This change sets the `"0e12"` value from #4314 with the `attributes` processor, which is in the core distribution. The test checks how scalars are rendered from the CR to the ConfigMap, so the choice of processor does not matter.


swiatekm:
Closed in favor of https://github.com/open-telemetry/opentelemetry-operator/pull/5632

